PR #611, opened 2016-01-08
[ZEPPELIN-554] Streaming interpreter output to front-end

Leemoonsoo:
### What is this PR for?
Output from interpreter is displayed after completion of paragraph execution.
It'll be useful if output can be streamed to front-end during execution.

Previous work #593 injects InterpreterOutput stream object to Interpreter.
This PR is based on #593 and stream the data from InterpreterOutput to front-end.

This implementation only streams output is %text. Other output type (%html, %angular, %table) is not streamed to the front end.

While this PR keeps backward compatibility, Interpreter who want to use this feature will need to modify code to write output into `InterpreterOutput` instead of return with `InterpreterResult`.

This PR includes modification of SparkInterpreter to use InterpreterOutput.

### What type of PR is it?
Feature

### Is there a relevant Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-554

### How should this be tested?
Run such code using Spark interpreter
```
(1 to 10).foreach{ i=>
    Thread.sleep(1000)
    println("Hello " + i)
}
```

### Screenshots (if appropriate)
stream_output

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #683, opened 2016-02-02
ZEPPELIN-646: Shell interpreter output streaming

AhyoungRyu:
### What is this PR for?
After #611 merged, Zeppelin provides streaming output for **spark** and **pyspark** interpreter. For the further improvement, I changed a few code lines using <code>[InterpreterContext](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterContext.java#L66)</code> so that **sh** interpreter can be available too.

### What type of PR is it?
Improvement

[ZEPPELIN-646: Shell interpreter output streaming](https://issues.apache.org/jira/browse/ZEPPELIN-646)
[ZEPPELIN-554: Streaming interpreter output to front-end]()

After applying this PR, run this below code with `sh` interpreter in Zeppelin.
```
date && sleep 3 &&  date
```

Then you can see two timestamps which have 3 seconds gap. 

shell_interpreter

### Questions:

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -72,7 +72,7 @@ public class ShellInterpreter extends Interpreter {
     DefaultExecutor executor = new DefaultExecutor();
     ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
     ByteArrayOutputStream errorStream = new ByteArrayOutputStream();
-    executor.setStreamHandler(new PumpStreamHandler(outputStream, errorStream));
+    executor.setStreamHandler(new PumpStreamHandler(contextInterpreter.out, errorStream));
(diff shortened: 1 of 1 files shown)


corneadoug:
After some search, it seems that the streaming output wasn't documented anywhere.
That could be valuable information for those making interpreter, but also to show possible streaming ability of Zeppelin.

AhyoungRyu:
@corneadoug Make sense. Then I'll write a new docs for ability of Zeppelin streaming output.

felixcheung:
cool @dArKmAn1

Leemoonsoo:
Yes it would be nice to document how to use InterpreterOutput when writing interpreter.
Maybe, the document can even link this PR's file diff to show how to change interpreter to leverage InterpreterOutput, as an example.

Anyway, LGTM

AhyoungRyu:
@Leemoonsoo  Make sense. Then i'll add that content you mentioned in the docs : )

Leemoonsoo:

PR #685, opened 2016-02-02
[ZEPPELIN-595] Allow displaying decimal format in d3

minahlee:
### What is this PR for?
This PR enables displaying decimal format in built-in nvd3 chart

### What type of PR is it?
Bug Fix

[ZEPPELIN-595](https://issues.apache.org/jira/browse/ZEPPELIN-595)

Before
screen shot 2016-02-02 at 2 12 38 pm
After
screen shot 2016-02-02 at 2 13 17 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1238,7 +1238,7 @@ angular.module('zeppelinWebApp')
 
   };
 
-  var groupedThousandsWith3DigitsFormater = function(x){
+  var groupedThousandsWith3DigitsFormatter = function(x){
     return d3.format(',')(d3.round(x, 3));
   };
 
@@ -1262,7 +1262,7 @@ angular.module('zeppelinWebApp')
     if(d >= Math.pow(10,6)){
       return customAbbrevFormatter(d);
     }
-    return groupedThousandsWith3DigitsFormater(d);
+    return groupedThousandsWith3DigitsFormatter(d);
   };
 
   var setD3Chart = function(type, data, refresh) {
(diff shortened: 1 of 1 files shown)


felixcheung:
looks good

minahlee:
For reviewer's information, please check [this link](https://travis-ci.org/apache/incubator-zeppelin/builds/106608051) to see CI passed. I triggered CI twice, first one failed, second one suceeded. But the first CI test finished later than second one that's why build indicated as failed.

vgmartinez:
I tested...LGTM

Leemoonsoo:
LGTM

corneadoug:
This doesn't work well without a floating point
screen shot 2016-02-03 at 11 34 06 am

I would suggest to do `return d3.round(d, 3);` instead.
which would round to 3 digits maximum (when needed)
screen shot 2016-02-03 at 11 36 34 am

minahlee:
@corneadoug Thank you for review. Just pushed a fix

corneadoug:
Since its the only problem I found, I would say LGTM

prabhjyotsingh:
I have slightly different view, what I think original idea behind using 
`var integerFormatter = d3.format(',.1d');` 
was to display number like `123123123` to `123,123,123`, but by 
`d3.round(d, 3)` 
it looses its formatting, and in my suggestion if you use 
`var integerFormatter = function(x){return d3.format(',')(d3.round(x, 3));};` 
it retains formatting and have decimal precision as well, kind of hacky but it works.

Also its integerFormatter is referenced in paragraph.controller.js.html as well, do you think this should be changed as well ?

corneadoug:
@prabhjyotsingh Doesn't look that hacky to me and preserve a nice formatting.

> Also its integerFormatter is referenced in paragraph.controller.js.html as well, do you think this should be changed as well ?

Didn't find any other references, could you post a link?

prabhjyotsingh:
My bad, searched into a wrong directory; it was under reports. Please ignore.

felixcheung:
+1 on format approach, but I think that shouldn't be called `integerFormatter` now since it works with decimal numbers.

minahlee:
@prabhjyotsingh your suggestion makes perfect sense. It's ready now.

prabhjyotsingh:
Awesome!!! 
LGTM :+1:

minahlee:
My bad. I just pushed a fix.

felixcheung:
LGTM. thanks

PR #695, opened 2016-02-04
ZEPPELiN-654 Improvement to SparkSqlInterpreter

karup1990:
### What is this PR for?
Improve performance of Sparksqlinterpreter
Use StringBuilder instead of String when building the results returned by SparksqlInterpreter 

### What type of PR is it?
Improvement

### Todos
NA

Try to run a Sparksql command that returns large number of rows(make sure `zeppelin.spark.maxResult` is set to a larger value) with and without fix.
Keep note of the time taken in each case.
The time taken with this change  is significantly lesser.

NA

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -17,15 +17,11 @@
 
 package org.apache.zeppelin.spark;
 
-import static scala.collection.JavaConversions.asJavaCollection;
 import static scala.collection.JavaConversions.asJavaIterable;
-import static scala.collection.JavaConversions.collectionAsScalaIterable;
 
 import java.io.IOException;
-import java.io.PrintStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


karup1990:
Removing some unused imports in `ZeppelinContext.java` in this PR

karup1990:
When SparkSqlinterpreter returns the result , ZeppelinContext constructs a String out of it for displaying on UI.
The logic to construct result string uses String Class for each column value in each row. 
If result has x rows and each row has y columns, it creates x * y String objects for constructing the result string. 
For large values of x and y, the number of String objects created is enormous and is time consuming.
Instead we could use StringBuilder, which will reduce number of objects created and will be faster.

corneadoug:
@karup1990 Could you provide a table showing multiple time differences depending on results size?
That would be quite useful for us to review without having to run every cases possible.

Leemoonsoo:
I have tested following code

```
%spark
case class Test(n:Int, s:String)
val data = sc.parallelize(1 to 10000).map(i=>Test(i, "aaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddd")).toDF
data.registerTempTable("test")
```

and

```
%sql
select * from test
```

took 20 sec before this patch but only 1 sec after this patch in my machine.
significant performance improvement.

It looks good to me.

Leemoonsoo:

felixcheung:
looks good

PR #699, opened 2016-02-05
Simple improvement to showDF() to use StringBuilder

markncooper:
**What is this PR for?
The way tables of results are currently getting built within
the ZeppelinContext is pretty slow due to way strings are
getting constructed.

https://issues.apache.org/jira/browse/ZEPPELIN-659

**What type of PR is it?
Performance improvement.

**How should this be tested?
The current tests should catch any issues introduced by it.

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -322,16 +322,18 @@ public class ZeppelinContext extends HashMap<String, Object> {
       throw new InterpreterException(e);
     }
 
-    String msg = null;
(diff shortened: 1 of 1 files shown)


markncooper:
Oops, I see this was just fixed with this change:

https://github.com/apache/incubator-zeppelin/pull/695

corneadoug:
@markncooper Actually you are not touching the same code lines, so maybe both would give better performances

felixcheung:
looks good - it seems we have some duplications here, perhaps @markncooper you would like to refactor this after #695 is merged to a util method?

felixcheung:
@markncooper #695 is merged.

jongyoul:
@markncooper Could you please close this PR?

PR #703, opened 2016-02-08
[ZEPPELIN-513] Dedicated interpreter session per notebook

Leemoonsoo:
### What is this PR for?

Currently, all notebooks that binded to the same interpreter setting shares the same interpreter instance.
For example, let's say there're `spark-production` interpreter setting and all notebooks are using it.
Then all notebooks share a single instance of SparkInterpreter, and that result all the variables are being shared across the notebook.

But sometimes isolated name space for variables per notebook is preferred. In this case we can create per notebook interpreter instance.

This PR provides generalized facility to provides per notebook interpreter instance, not only for SparkInterpreter but for any other existing interpreter without modification.

**Design**

It provides toggle switch that user turn on / turn off this mode per interpreter setting.
When per notebook interpreter is turned on, it still creates a single remote interpreter process per interpreter setting. But inside of the process, each notebook will have it's own interpreter instance and the instance is terminated automatically on removal of notebook or unbinding note from interpreter setting.

**SparkInterpreter**

Because of the design keeps single remote interpreter process and creates multiple interpreter instance inside of it, SparkInterpreter become a problem. While multiple SparkContext can not be created on a single process. So this PR try to, create per notebook scala compiler instance, but share single SparkContext instance. In this way, each notebook can run job concurrently by leveraging fairscheduler of SparkContext without running multiple Spark application (SparkContext).

And this approach is inspired by https://github.com/piyush-mukati/incubator-zeppelin/tree/parallel_scheduler_support_spark, this PR trying to generalize it for all other interpreters.

### What type of PR is it?
Feature

### Todos
* [x] - Per note interpreter session 
* [x] - Scala compiler instance per note and share a single SparkContext (SparkInterpreter)
* [x] - Handle Spark class server correctly
* [x] - Handle SparkInterpreter Progressbar correctly

https://issues.apache.org/jira/browse/ZEPPELIN-513

Check "Per note session" of your Spark interpreter setting. 
And create two different notebook and check scala variables are not shared.

Checkbox to enable
image

How it works for SparkInterpreter
per_note_session

### Questions:


AhyoungRyu:
@Leemoonsoo Cooool!

Leemoonsoo:
Documentation added. And add gif image in this PR description to show how it works in SparkInterpreter.
Please review this PR.

felixcheung:
Is this only applicable to Spark interpreter group?
Instead of "per note session" should we say "separate interpreter for each notebook" or similar?

Leemoonsoo:
It's available in general. So all other interpreter groups can use.
"Separate interpreter for each notebook" Sounds easier to understand.

AhyoungRyu:
:+1: for @felixcheung !

echarles:
Or "Separate interpreter for each note" ?

cfr. https://issues.apache.org/jira/browse/ZEPPELIN-616 Zeppelin UI consistency: should use Note or Notebook term

Leemoonsoo:
@felixcheung @AhyoungRyu @echarles I have changed to "Separate interpreter for each note"

khalidhuseynov:
Thanks for the great improvement! 
What do you think would be the default use case for the interpreter context sharing? Currently the checkbox is unmarked by default, meaning the context is shared for notes by default. Other way would be to have it marked by default (or more preferably, have unmarked checkbox saying `Share interpreter for notes`). Just wanted to know opinion from UX and performance ( per note interpreter maybe more resourceful) points of views.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -484,20 +488,40 @@ public void open() {
     intp.setContextClassLoader();
     intp.initializeSynchronous();
 
-    completor = new SparkJLineCompletion(intp);
+    numReferenceOfSparkContext.incrementAndGet();
```
felixcheung: should this be under try {} catch {}? if someone fails between line 493-523, shouldn't this revert the ref count?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
```diff
@@ -73,10 +95,22 @@ public String getId() {
...
-    for (Interpreter intp : this) {
-      p.putAll(intp.getProperty());
+
+    Collection<List<Interpreter>> intpGroupForANote = this.values();
+    if (intpGroupForANote != null && intpGroupForANote.size() > 0) {
+      for (List<Interpreter> intpGroup : intpGroupForANote) {
+        for (Interpreter intp : intpGroup) {
+          p.putAll(intp.getProperty());
+        }
+        // it's okay to break here while vevery List<Interpreters> will have the same property set
```
felixcheung: minor typo: vevery -> every

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -192,23 +201,35 @@ private Interpreter getInterpreter(String className) throws TException {
...
+    synchronized (interpreterGroup) {
+      List<Interpreter> interpreters = interpreterGroup.get(noteId);
+      if (interpreters != null) {
+        Iterator<Interpreter> it = interpreters.iterator();
+        while (it.hasNext()) {
+          Interpreter inp = it.next();
+          if (inp.getClassName().equals(className)) {
+            inp.close();
+            it.remove();
+            break;
```
felixcheung: shouldn't it continue the while loop to remove all interpreters with the same classname?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -430,19 +437,58 @@ private InterpreterGroup createInterpreterGroup(String id,
...
+    InterpreterGroup interpreterGroup = interpreterSetting.getInterpreterGroup();
+    String groupName = interpreterSetting.getGroup();
+    InterpreterOption option = interpreterSetting.getOption();
+    Properties properties = interpreterSetting.getProperties();
+
+    // if interpreters are already there, wait until they're being removed
+    synchronized (interpreterGroup) {
+      while (interpreterGroup.containsKey(noteId)) {
+        try {
+          interpreterGroup.wait(1000);
```
felixcheung: should this retry a finite number of times?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -34,27 +34,59 @@
   private String group;
   private String description;
   private Properties properties;
-  private InterpreterGroup interpreterGroup;
+
+  // use 'interpreterGroup' as a field name to keep backward compativility of
```
felixcheung: compatibility?

felixcheung:
big change - only a few minor comments.

Leemoonsoo:
@khalidhuseynov Thanks for the opinion. Both UX and Performance point of view, which make sense to be default will be totally depends on the interpreters.

For example, much higher resource will be required to use "separate interpreter per note" for SparkInterpreter, while it creates separate scala compiler instance per note.
But for markdown interpreter or shell interpreter, resource requirement will be pretty much the same.

And UX side, i was more focused on the providing the same default behavior to the previous version.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
Leemoonsoo: continue the while loop doesn't harm. But it is guaranteed that only single interpreter exists for specific className in the `List<Interpreter> interpreters = interpreterGroup.get(noteId)`. So breaking here saves tiny bit cpu cycle.

Leemoonsoo:
@felixcheung Thanks for the review. I have addressed the comments.

khalidhuseynov:
I see, that makes sense. CI is green, LGTM

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -140,10 +136,9 @@ public static void main(String[] args)
 
 
   @Override
-  public void createInterpreter(String interpreterGroupId, String className, Map<String, String>
-          properties)
-      throws TException {
-
+  public void createInterpreter(String interpreterGroupId, String noteId, String
```
jongyoul: @Leemoonsoo How about passing InterpreterContext which already has a noteId and more information? I think It would be flexible and extendable.

jongyoul:
@Leemoonsoo As I mentioned in some codes, how about passing InterpreterContext to the RemoteInterpreterServer? It would be useful and we don't any any other information when we add some features.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
Leemoonsoo: InterpreterContext always has actual noteId, regardless of "Separate interpreter for each note" is true or false. However, the `noteId` passed to `createInterpreter` is little bit different. 

When "Separate interpreter for each note" true, it is actual noteId. 
However, when "Separate interpreter for each note" is false, the value becomes NoteInterpreterLoader.SHARED_SESSION. So all note shares the same interpreter.

Leemoonsoo:
@felixcheung Thanks for letting me know. I have updated to use System.nanoTime().

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
jongyoul: @Leemoonsoo Thanks for letting me know. Actually, it's one of points to use noteId - but I mis-point it - and I really want to pass InterpreterContext in case of open and close. how about it?

Leemoonsoo: open() / close() can be invoked even without notebook creation. but InterpreterContext has some notebook / paragraph related values. except for that i don't see big problem for now (didn't have much time to have think thoughtfully).

I was also thinking make `Interpreter.completion()` return `List<Completion>` instead of `List<String>`. so completion can show more informations. not only method name, but also parameters, etc.

How about start a new discussion for updating Interpreter api?

jongyoul: Sure, there's no problem in this implementation. And I agree with yours of updating interpreter api. Go forward first.

jongyoul:
LGTM

Leemoonsoo:
Thanks @jongyoul @AhyoungRyu @felixcheung @echarles @khalidhuseynov for taking a look and review.

Merging it if there're no more discussions.

sourav-mazumder:
Hi,

I was just trying  some of these. Looks like with this solution there would be queue built up for running multiple paragraphs in parallel for Spark. And that essentially causes huge slowness when multiple paragraphs are running in parallel. It is a big road block for multi-tenancy.

My suggestion is to create multiple processes for Spark Interpreter but limit it/link it to user instead of notebook. So for every user there should be a separate interpreter process instead of for every notebook. As we are also having PR on user authentication and authentication in place achieving this should not be a problem.

Regards,
Sourav

EronWright:
@sourav-mazumder @Leemoonsoo There are two good reasons why Zeppelin should run each note as a separate Spark app instance.
1. Dependencies - each note may have different dependencies (expressed as %dep, as a spark package, etc.)
2. Scale - it is reasonable that each note has unique scale requirements; one might be memory-intensive, etc.

I don't think a per-user instance would address those concerns.

@Leemoonsoo how hard would it be to achieve that, and how can I help?

Leemoonsoo:
It's still an open issue. https://issues.apache.org/jira/browse/ZEPPELIN-183

While trying to answer the question 'how hard would it be', i realized that i can quickly make a patch.
I'll create a pullrequest soon.

Leemoonsoo:
@EronWright @sourav-mazumder. I just created a pullrequest https://github.com/apache/incubator-zeppelin/pull/795, Please take a look.

PR #705, opened 2016-02-09
User in interpreter context

prabhjyotsingh:
### What is this PR for?
The goal of the PR is to pass userName/userInfo from front-end to interpreters.
The PR uses Shiro authentication(https://github.com/apache/incubator-zeppelin/pull/586).

### What type of PR is it?
Improvement

### Todos
* [x] - Change RemoteInterpreterService.thrift signature
* [x] - Change InterpreterContext/RemoteInterpreterContext signature
* [x] - Modify all existing interpreters to above signature

N/A


rconline:
This is a requisite first step for getting to multi-user scenarios. I believe this should get discussed a little bit in terms of approach.

prabhjyotsingh:
This is ready for review.

felixcheung:
@rconline how would you want to discuss? Should we open a JIRA - or a JIRA parent with subtests for this multi-user auth use cases?

felixcheung:
@prabhjyotsingh could you explain more on the reason for this change?
Also this seems like a somewhat big change - could you open a JIRA on this so it could be tracked?

rconline:
@felixcheung @prabhjyotsingh  The purpose of this discussion should be to get to an approach that allows us zeppelin to do a couple of things - pass user identity to the underlying spark cluster, and thereby support multi-tenancy. There is JIRA open - https://issues.apache.org/jira/browse/ZEPPELIN-645 which we could use to discuss this further. Does that make sense?

Leemoonsoo:
I think it make sense to let access authentication info through InterpreterContext.
But i think authentication info will be more than a user name in the future. It can have some credentials/token stuff from Authentication system (shiro), so interpreter process can reuse it without asking authentication again, when they're required.

So my suggestion is, create such class `AuthenticationInfo`, and pass it into InterpreterContext instead of `String userName`, for further extension in the future.

prabhjyotsingh:
Thank you for the suggestion @Leemoonsoo. Have added AuthenticationInfo class, and made relevant changes.

Leemoonsoo:
I have tried and it works well.
One thing is, `org.apache.zeppelin.display` package supposed to have Display System related codes.
I think `AuthenticationInfo` is less related to Display System. Could you suggest other package name?

prabhjyotsingh:
Thanks you @Leemoonsoo, that does make a lot of sense.

Leemoonsoo:
Thanks @prabhjyotsingh.
It looks good to me.

jongyoul:
LGTM

felixcheung:
Let's rebase and merge this?

prabhjyotsingh:
Thank you for reviewing this. Have merged this with master.

PR #706, opened 2016-02-09
Create seperate matrix for selenium test case

prabhjyotsingh:
### What is this PR for?
A lot of time travis CI breaks because of selenium test case, and at that point it becomes trivial to know the exact reason. Hence, in this PR creating a separate matrix for testing selenium test cases.

### What type of PR is it?
Improvement

### Is there a relevant Jira issue?
N/A

### How should this be tested?
Check travis UI. It should have a new matrix added.

PR #709, opened 2016-02-10
testSparkInterpreterDependencyLoading does not check for endToEndTestEnabled

prabhjyotsingh:
### What is this PR for?
This is a fix for selenium test case `testSparkInterpreterDependencyLoading` that does not check for `endToEndTestEnabled` causing build to fail on local. (`mvn $TEST_FLAG $PROFILE -B`)

Used to get this error in file `./zeppelin-server/target/failsafe-reports/org.apache.zeppelin.ZeppelinIT.txt`

    -------------------------------------------------------------------------------
    Test set: org.apache.zeppelin.ZeppelinIT
    -------------------------------------------------------------------------------
    Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.008 sec <<< FAILURE! - in org.apache.zeppelin.ZeppelinIT
    testSparkInterpreterDependencyLoading(org.apache.zeppelin.ZeppelinIT)  Time elapsed: 0.008 sec      <<< ERROR!
    java.lang.NullPointerException: null
	    at org.apache.zeppelin.ZeppelinIT.testSparkInterpreterDependencyLoading(ZeppelinIT.java:210)

### What type of PR is it?
Hot Fix

### Todos
* [x] - Fix for testSparkInterpreterDependencyLoading
* [x] - delete created notebook for cleanup.

N/A

`mvn $TEST_FLAG $PROFILE -B` should not fail on local when env variable "CI" is not set.


zeppelin-server/src/test/java/org/apache/zeppelin/ZeppelinIT.java
```diff
@@ -202,46 +202,59 @@ public void testAngularDisplay() throws InterruptedException{
...
+      sparkEditBtn = pollingWait(By.xpath("//div[h3[text()[contains(.,'spark')]]]//button[contains(.,'edit')]"),
+          MAX_BROWSER_TIMEOUT_SEC);
+      sparkEditBtn.click();
+      WebElement testDepRemoveBtn = driver.findElement(By.xpath("//tr[descendant::text()[contains(.,'" +
+          artifact + "')]]/td[3]/div"));
+      sleep(5000, true);
+      testDepRemoveBtn.click();
+      driver.findElement(By.xpath("//button[contains(.,'Save')]")).submit();
+      driver.switchTo().alert().accept();
+    } catch (ElementNotVisibleException e) {
```
felixcheung: would there be any other error? like xpath returning no match (say the element has been removed)

felixcheung:
I think it's fine keeping this in #706 - either way is ok with me.
Much easier to see diff this way: https://github.com/apache/incubator-zeppelin/pull/709/files?w=1

zeppelin-server/src/test/java/org/apache/zeppelin/ZeppelinIT.java
prabhjyotsingh: Sure, what you say make sense, let me change all other catch from ElementNotVisibleException to just Exception, so, it catches every thing in https://github.com/apache/incubator-zeppelin/pull/706

prabhjyotsingh:
Yes, agreed, this is much cleaner.

prabhjyotsingh:
@felixcheung have replaced all "ElementNotVisibleException" to "Exception", and then log the same.

felixcheung:
looks good.

zeppelin-server/src/test/java/org/apache/zeppelin/ZeppelinIT.java
```diff
@@ -202,46 +202,59 @@ public void testAngularDisplay() throws InterruptedException{
...
+      WebElement testDepRemoveBtn = driver.findElement(By.xpath("//tr[descendant::text()[contains(.,'" +
+          artifact + "')]]/td[3]/div"));
+      sleep(5000, true);
+      testDepRemoveBtn.click();
+      driver.findElement(By.xpath("//button[contains(.,'Save')]")).submit();
+      driver.switchTo().alert().accept();
+    } catch (Exception e) {
+      LOG.error("Exception in ZeppelinIT while testSparkInterpreterDependencyLoading ", e);
+      File scrFile = ((TakesScreenshot) driver).getScreenshotAs(OutputType.FILE);
+
```
Leemoonsoo: Shouldn't we throw `e` after taking screenshot, if getting exception is considered as a test failure?

prabhjyotsingh: Sure did the relevant change.

felixcheung: You might want to just `throw e` - exception stack should already have the test name?

prabhjyotsingh: Thanks a lot @felixcheung, made the required change.

prabhjyotsingh:
@felixcheung  @Leemoonsoo Thanks for the review. I have addressed the comment.

zeppelin-server/src/test/java/org/apache/zeppelin/ZeppelinIT.java
```diff
@@ -193,55 +194,68 @@ public void testAngularDisplay() throws InterruptedException{
       sleep(100, true);
 
       System.out.println("testCreateNotebook Test executed");
-    } catch (ElementNotVisibleException e) {
+    } catch (Exception e) {
       LOG.error("Exception in ZeppelinIT while testAngularDisplay ", e);
       File scrFile = ((TakesScreenshot)driver).getScreenshotAs(OutputType.FILE);
-
+      throw new Exception(e);
```
felixcheung: can we instead say `throw e;` here? e is already an Exception
http://stackoverflow.com/questions/1097527/rethrowing-exceptions-in-java-without-losing-the-stack-trace

prabhjyotsingh: got it!!! missed it. Thank you. :)

felixcheung:
looks good, thanks.
merging if no more comments.

Leemoonsoo:
LGTM

PR #713, opened 2016-02-13
Add checkbox as a type of dynamic forms

zhongneu:
### What is this PR for?
1. Add checkbox support in dynamic forms
2. Fix ZEPPELIN-699: Cannot select the first item of dynamic forms when it is just created

### What type of PR is it?
Feature & Bug Fix

### Todos
* [x] Compatibility test with notes from previous versions
* [x] Documentation
* [x] Nice CSS layout
* [x] Support checkbox creation by string substitution
* [x] Support pyspark

https://issues.apache.org/jira/browse/ZEPPELIN-671
https://issues.apache.org/jira/browse/ZEPPELIN-669

1. Create a %spark paragraph: `val a = z.checkbox("my_check", Seq(("a", "a"), ("b", "b"), ("c", "c")))`
2. Toggle the checkboxes and see the outcome

igojnqh0mz

### Questions:

NO

Sort of. I am not sure whether Input.type (form.type) has already been used in other places

YES


AhyoungRyu:
Hi @zhongneu, thanks for your contribution!
For the quick review of this PR, could you apply our [PR template](https://github.com/apache/incubator-zeppelin/blob/master/CONTRIBUTING.md#creating-a-pull-request) to your description? : )

zhongneu:
@AhyoungRyu updated! Shall I create a JIRA for this?

AhyoungRyu:
@zhongneu Really appreciate for your quick response! It's not a mandatory, but it will be good if you create a new one : )

doanduyhai:
@zhongneu What is the syntax for this dynamic checkbox ?

For example, for input text, the syntax is `${input_text_label=default value}`

For dynamic drop down menu, it is `${drop_down_label=default value|value 1|value 2| ....|value N}`

zhongneu:
@doanduyhai The current parsing code supports syntax like: ${label:type=...}. I do have concer about this, because from the code I cannot see what does this type refer to. Is it the type of the value, or the type of the form?

I don't see the type is used anywhere in the current code though, but I am not sure whether it is used outside of the project (3rd-party interpreters)

doanduyhai:
@zhongneu me either, I didn't know that such syntax ${label:type=...} exists.

 But you did not answer my question. How can I create a dynamic form with checkbox without using ZeppelinContext ?

zhongneu:
@doanduyhai Sorry that my last reply is a little bit obscure. I mean we can use syntax like this: `${mycheck:checkbox=default1|default2,value1|value2|value3}`

The query substitution implementation is still in the TODO list. I just want to make sure it won't break anything by using the type syntax here.

doanduyhai:
Ok great, thanks for the answer

zhongneu:
I've submitted one more commit with better layout and supporting show/hide options. I do run into an issue: the show/hide option will not be saved. Currently, I use formulaire.hidden to store the status, but it won't be committed because it is not part of the COMMIT_PARAGRAPH protocol.

I am not sure about what is the best way to add the form hidden status into the protocol. Any suggestions?

corneadoug:
I'm not sure show/hide option is really necessary in general, since the goal of the dynamic form is to interact with it. (We don't even hide it in report mode)

However if we want that type of behavior, it would be better to have it implemented in a more general way for all dynamic forms, and therefore in a different PR

zhongneu:
I think checkbox is a little bit different from other types of forms, because we may have many checkboxes  in a form, and my not look nice if we cannot hide them. 

How about provide an option to the users to control the default value of hidden for checkboxes?

corneadoug:
You can already add a lot of different dynamic forms to one single paragraph and it won't look good.
Hiding better be at the form level (dynamic forms container), than for one specific element.
Also it depends what you want to achieve with a checkbox, originally it is the kind of component you use to activate an option, not to make multiple selections.
If the goal is to have a multi-select, we better work on a multi-select dropdown.
(I already shared an example using angular display: https://www.zeppelinhub.com/viewer/notebooks/aHR0cHM6Ly9yYXcuZ2l0aHVidXNlcmNvbnRlbnQuY29tL2Nvcm5lYWRvdWcvWmVwcGVsaW4tTm90ZWJvb2tzL21hc3Rlci9BdXRvLUNvbXBsZXRlLU11bHRpU2VsZWN0L25vdGUuanNvbg)

zhongneu:
Agreed. I will revoke the changes of the hidden behavior. And thanks for your example of multi-select using angular. It looks really cool! I also thought about using angular, but I really want the UI elements and charts can be in the same paragraph, that's why I created this pull request.

zhongneu:
I found that it is difficult to support multi-selection using string substitution: ${var=value}, because we do not know how to expand it properly in a query. We  cannot simply expand it into comma-separated string, because the users may want to use them in several other places (such as filtering conditions). This will make the substitution syntax more complex.

Could we support checkboxes in ZeppelinContext exclusively?

Leemoonsoo:
Right, multi-selection using string substitution can be complex, considering target language syntax.
How about just concat selections and let user take care target language syntax? 

Let's say syntax is
`${mycheck:checkbox=default1|default2,value1(label1)|value2(label2)|value3(label3)}`

And if target language is SQL, filtering can be written like

```
select * from table where 1=1 ${color:checkbox=or color=yellow,or color=yellow(yellow)|or color=red(red)}
```

zhongneu:
I was a little bit worried about this approach may not be very robust: we have several reserved characters: `${}|(),`. Because the users will put more syntactic elements inside, they may conflict with the reserved characters. However, this is definitely a promising solution, if we document it appropriately.

zhongneu:
I've refactored the variable parsing form a little bit to make the parsing/substation logic consistent, as well as added some unit tests for input parsing.

The checkbox form now can support custom delimiter using syntax like:
`${checkbox( or ):mycheck=a|b,a|b|c}`

This example will be expanded to `a or b`. See the updated screenshots for demo

Leemoonsoo:
I tried and it works really well :-)

corneadoug:
So, from what I saw in the front-end code, a checkbox is considered as a group of checkbox by default?
I'm going look into it a bit more, in the mean time @zhongneu could you take care of the last point of your todo list (documentation), so that we can proceed to merge once its reviewed?

zhongneu:
@corneadoug The naming of the css class is a little bit confusing. I cannot use "checkbox" directly, because it seems it will inherit some styles from other classes also named "checkbox". Shall I use a better name, such as "checkbox-input" or is there anyway to prevent it inheriting from other checkbox classes? 

I've updated the documentation. Added another task to support pyspark interpreter.

zhongneu:
All tasks are done. Ready for final review

Leemoonsoo:
z.checkbox() works well in Spark and PySpark. also string substitution works well.
Thanks @zhongneu for great work.

Looks good to me.

Leemoonsoo:

felixcheung:
looks good - great contribution!

PR #714, opened 2016-02-14
Exporting to TSV

zhongneu:
### What is this PR for?
Allow users to export data in a paragraph to a TSV file.

There is already a [PR](https://github.com/apache/incubator-zeppelin/pull/6) for this, but it uses DataTables, which requires Flash.

### What type of PR is it?
Feature

### Todos
* [ ] - Hide/disable the button if it is not in table/chart view
* [ ] - Find an icon for the new button
* [ ] - Saved file name is not correct in Safari
* [ ] - Test with large data set

[ZEPPELIN-672](https://issues.apache.org/jira/browse/ZEPPELIN-672)

1. create a paragraph with data in %table view
2. select the "Export to TSV" item in the paragraph settings list

vvzhsyciev

### Questions:
NO

NO

NO

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -83,6 +83,9 @@ limitations under the License.
         <a ng-click="goToSingleParagraph()"><span class="icon-share-alt"></span> Link this paragraph</a>
       </li>
       <li>
+        <a ng-click="exportToTSV()"><span class="icon-share-alt"></span> Export to TSV</a>
+      </li>
+      <li>
         <a ng-click="clearParagraphOutput()"><span class="fa fa-eraser"></span> Clear output</a>
       </li>
       <li>
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -2187,19 +2187,19 @@ angular.module('zeppelinWebApp')
   };
 
   $scope.exportToTSV = function () {
-    var data = $scope.paragraph.result
+    var data = $scope.paragraph.result;
     var tsv = '';
     for (var titleIndex in $scope.paragraph.result.columnNames) {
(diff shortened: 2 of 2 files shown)


corneadoug:
Tested after rebasing with master, no conflicts.
Let's rebase it and Merge it :)

corneadoug:
@zhongneu Could you rebase?

johnnyws:
@corneadoug are you sure you want to merge this? I thought we were working on #761?

There are still pending TODOs in this PR. I don't think we should merge it before finish them.

corneadoug:
It was already discussed there.
Regarding the TODO, there is only one that is useful: `Hide/disable the button if it is not in table/chart view`, but it can always be done later. (not critical for now)
The icon is fine, the Safari problem is not related to this feature (it affect also the existing export notebook), the test with large dataset doesn't matter since if it was too big, the browser will crash.

If we get this done with, then we can work on improving it, and build around it.
Otherwise it just become a blocker

ankurmitujjain:
@corneadoug I think its not right/user friendly way to have download data functionality.

 You liked approach of including download buttons in "settings".
But I believe **"Settings" is something to provide configuration/setting options and not to provide any functionality**.

Also @adejanovski [#6](https://github.com/apache/incubator-zeppelin/pull/6#issuecomment-185086161) commented rightly that 

> viewers have to find out that they need to click on the gear, which is hidden in report mode and simply absent when you link directly to the paragraph as an iframe

So I believe it's not a good idea to include buttons in settings as many community member don't liked it.

Secondly if DataTables provides export functionality which works and is extensible, than we should use it instead of writing our own way to do data export, this I mentioned earlier as well... [#6](https://github.com/apache/incubator-zeppelin/pull/6#issuecomment-185081090)

Also about extra UI space, its not an issue for using datatable or any of its functionality, but its general issue and to fix it we need to have a different PR...

If I look different notebook implementation, than the way they handle this issue is like below...
They are providing all buttons at bottom of table, and zeppelin provides same on top of table.

image

corneadoug:
Merging this if no more discussions.
I will take care of moving the button as suggested next to the graph buttons with a download icon, right after.

PR #720, opened 2016-02-16
Bringing compiledScope for backward compatibility

Leemoonsoo:
### What is this PR for?
compiledScope was not exposed API for users.
However, some users were using compiledScope in their notebook and [ZEPPELIN-551](http://issues.apache.org/jira/browse/ZEPPELIN-551) break this notebook work by removing compiledScope.

https://gist.github.com/granturing/a09aed4a302a7367be92 is an example notebook that uses compiledScope.

This PR bringing compiledScope back for backward compatibility.

### What type of PR is it?
Improvement

### Todos
* [x] - Bringing compiledScope back for backward compatibility

compiledScope was removed by http://issues.apache.org/jira/browse/ZEPPELIN-551 

Test https://gist.github.com/granturing/a09aed4a302a7367be92

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -21,8 +21,13 @@ angular.module('zeppelinWebApp')
   $scope.paragraph = null;
   $scope.originalText = '';
   $scope.editor = null;
+
   var paragraphScope = $rootScope.$new(true, $rootScope);
+  // to keep backward compatibility
+  $scope.compiledScope = paragraphScope;
+
   var angularObjectRegistry = {};
+
 
   var editorModes = {
     'ace/mode/scala': /^%spark/,


felixcheung:
LGTM

corneadoug:
Its not backward compatible for some of my Notebooks.
We should ask some AngularDisplay power user to test it against theirs, so that they can share how they use that feature, this would allow us to also write some good practices on how to use it.

granturing:
The notebook is code I wrote a while ago in order to show an example of "reactive" client-side JavaScript charts. Is there some other way to bind an event handler to an Angular model? Using the compiledScope was the only way I was able to subscribe to change notifications and handle them in a function, vs simply having a UI template updated.

corneadoug:
@granturing Do you use any other method recently?

There used to be 2 compiledScope in the past (App level, and Notebook level), this PR solve the case if you try to access Notebook compiled scope through the paragraph controller selector.

I think that this backward compatibility is not necessary since its doesn't handle both of the previous use cases. We should provide a better API to access it, and document how to do proper selector until then.

Leemoonsoo:
I think accessing compiledScope is more like a hack, because of it wasn't an officially exposed api for notebook users, in the end, providing some api for javascript, such as `z.scope()` to get current scope object and documenting it is better way to go.

We used to have compiledScope in Notebook level, and App level. This PR brings one of compileScope. So some notebook might fixed, some notebook might still have problem.

I'd suggest, we merge it as a effort for minimizing internal API change. 
Whatever user use, we need to minimize Zeppelin breaks notebook working version by version.
And this PR definitely reduce notebook breaking.

Once we develop officially supported API, such as `z.scope()` then, reference to compiledScope can be deprecated and removed.

Leemoonsoo:

PR #722, opened 2016-02-17
allows zeppelin to be run and managed as a service.  ZEPPELIN-641

jeffsteinmetz:
### What is this PR for?
allows zeppelin to be run and managed as a service, does not start in background via nohup
the service manager handles process instead

### What type of PR is it?
Improvement

### Todos
* None, should work as is

ZEPPELIN-641

bin/zeppelin-daemon.sh upstart

### Questions:
* Does this needs documentation? updated

### DIFF
--- a/bin/zeppelin-daemon.sh
+++ b/bin/zeppelin-daemon.sh
@@ -21,7 +21,7 @@
 # description: Start and stop daemon script for.
 #
 
-USAGE="Usage: zeppelin-daemon.sh [--config <conf-dir>] {start|stop|restart|reload|status}"
+USAGE="Usage: zeppelin-daemon.sh [--config <conf-dir>] {start|stop|upstart|restart|reload|status}"
 
 if [[ "$1" == "--config" ]]; then
   shift
@@ -159,6 +159,16 @@ function check_if_process_is_alive() {
   fi
 }
 
+function upstart() {
+
+  # upstart() allows zeppelin to be run and managed as a service
+  # for example, this could be called from an upstart script in /etc/init
+  # where the service manager starts and stops the process
+  initialize_default_directories
+
+  $ZEPPELIN_RUNNER $JAVA_OPTS -cp $ZEPPELIN_CLASSPATH_OVERRIDES:$CLASSPATH $ZEPPELIN_MAIN >> "${ZEPPELIN_OUTFILE}"
+}
+
 function start() {
   local pid
 
@@ -240,6 +250,9 @@ case "${1}" in
     ;;
   stop)
     stop
+    ;;
+  upstart)
+    upstart
     ;;
   reload)
     stop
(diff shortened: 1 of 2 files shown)


bin/zeppelin-daemon.sh
```diff
@@ -159,6 +159,16 @@ function check_if_process_is_alive() {
...
 }
 
+function upstart() {
+
+  # upstart() allows zeppelin to be run and managed as a service
+  # for example, this could be called from an upstart script in /etc/init
+  # where the service manager starts and stops the process
+  initialize_default_directories
+
+  $ZEPPELIN_RUNNER $JAVA_OPTS -cp $ZEPPELIN_CLASSPATH_OVERRIDES:$CLASSPATH $ZEPPELIN_MAIN >> "${ZEPPELIN_OUTFILE}"
```
felixcheung: would it be able to detect that it is already running like start() does?

felixcheung: or we are saying this should only be called during boot from upstart script?

jeffsteinmetz: This is purely for usage with a service manager such as upstart when found in ```/etc/init```
Its actually how I am running this in production on Ubuntu.  Upstart already detects a running process.  If you call
`sudo service zeppelin start`
then call again
`sudo service zeppelin start`
the service manager return a message that it is already started.


for example, here is a `zeppelin.conf`

```
description	"zeppelin"

start on (local-filesystems and net-device-up IFACE!=lo)
stop on shutdown

# Respawn the process on unexpected termination
respawn

# respawn the job up to 7 times within a 5 second period.
# If the job exceeds these values, it will be stopped and marked as failed.
respawn limit 7 5

# zeppelin installed in /usr/share/zeppelin in this example
chdir /usr/share/zeppelin
exec bin/zeppelin-daemon.sh upstart
```

felixcheung: cool, maybe a help doc link or this code snippet in our doc would be helpful
this would be everything more reliable running as a service too.

jeffsteinmetz: recent commit now has expanded example in docs/install/install.md

felixcheung:
cool, looks good.

Leemoonsoo:
bin/zeppelin.sh does the similar job.
maybe, we want to keep only one of these two in the future.

Anyway, this PR looks good to me especially with nice documentation.

felixcheung:
Agreed on the overlap.
Let's merge this for now if there is no more comments.

jeffsteinmetz:
Agreed as well.  
It appears `bin/zeppelin.sh` is a subset of zeppelin-daemon.sh.  
Only one script would need to be maintained if `bin/zeppelin.sh` was depreciated in favor of a single zeppelin runner script.

felixcheung:
@jeffsteinmetz you are welcome to look into how the script should be merged - we would likely keep both of them functional for a release but one script could simply wrap the other.

PR #152, opened 2015-07-12
ZEPPELIN-157: Adding Map Visualization for Zeppelin

Madhuka:
- [x] Listing Map charting Libraries
- [x] Checking compatible license 
- [x] Adding chart library (leafletjs)
- [x] Adding new button for chart mapping
- [x] Loading GIS Mapp in Zeppelin
- [x] Removing the setting for mapping chart
- [x] Adding sample data set (loading)
  - [x] Get the sample GIS data and read it 
  - [x] Add it as in tutorail, where it can be run from the clean environment
- [x] Generalization data validation model
  - [x] Add 'validator' for the Map graph input format as service
- [x] Test for this

PR #728, opened 2016-02-19
Leaflet support, based PR 152

eddies:
### What is this PR for?
An update/follow-on for PR 152 that is limited to Leaflet support. There doesn't appear to have been any new movement on PR 152 since August 2015, so this PR updates Madhuka's work against master, but without any of the data validation or other non-Leaflet work.

### What type of PR is it?
Feature

### Todos
N/A

ZEPPELIN-157

Recreate Madhuka's example, e.g. use the following to populate a cell:
  https://gist.github.com/eddies/f20241e161aa2bbbb788

Then, in a new cell
```
%sql
select * from myMap
```
There should a new chart selector rightmost that will render the data w/ leaflet. You may need to clean your browser cache initially.

See Madhuka's original PR: https://github.com/apache/incubator-zeppelin/pull/152

### Questions:
* Does this needs documentation? Probably. It wasn't clear to me if PR 152 was going to move ahead in a new, Leaflet-only incarnation as suggested in the comment thread. But I needed Leaflet support, so went ahead with my own fork. I'm happy to work on documentation for this, but not sure if a sample notebook tutorial is what's desired (e.g. maybe something under the Display System docs is preferable?). If so, I'd just stick with something like the sample notebook in PR 152.


eddies:
FYI, 18b4f55 is failing CI with the following:

```
[ERROR] bower leaflet.heat#v0.2.0                        ECMDERR Failed to execute "git ls-remote --tags --heads git@github.com:Leaflet/Leaflet.heat.git", exit code of #128 Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts. Permission denied (publickey). fatal: Could not read from remote repository.  Please make sure you have the correct access rights and the repository exists.
[ERROR] 
[ERROR] Additional error details:
[ERROR] Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts.
[ERROR] Permission denied (publickey).
[ERROR] fatal: Could not read from remote repository.
[ERROR] 
[ERROR] Please make sure you have the correct access rights
[ERROR] and the repository exists.
```
Unfortunately, leaflet.heat does not support bower (the maintainer has steadfastly rejected any pull requests to include a bower.json). So I explicitly added the git url. But it looks like the host resolution on Travis is unhappy with the address it's getting for github.com.

randerzander:
this looks awesome. maps/heatmaps are one of the most asked for features whenever I demo Zeppelin.

With this as a base, how difficult would it be to also include geojson, or have query results auto-translated into geojson map features?

eddies:
@randerzander I spent a chunk of time last week playing w/ geojson support with pretty mixed results.
I probably sank the most time tracking down an annoying CSS issue that was preventing the SVG layers from rendering properly within Zeppelin. But the real difficulty for me was finding a way to pass in arbitrary geojson. I think supporting the transformation of query results into geojson features is doable, but I still think you need the ability to pass in your own geojson (e.g. a shapefile you've converted to geojson) for it to be a useful feature in Zeppelin and it was the latter that I ultimately gave up on (for the time being).

randerzander:
@corneadoug not sure about status of #152 , but this PR adds significant additional functionality (heatmaps). since this PR is based on @Madhuka's work, perhaps both he and @eddies can be given credit for it, if they're both agreeable to that approach?

The CI failure looks unrelated to the changes in this PR. Who can help resolve them and get these features moving?

Personally I would like to begin working on adding geojson and more geo features, but need one of either this PR, or #152 to be completed first.

corneadoug:
I can take a look at the CI failure.

I think the heatmap is a nice addition to the basic map. Maybe it could be toggled as an option instead of a different visualization button (just like the options in line chart)
If #152 gets done this week, I think we can rebase @eddies PR on top of it.
If #152 doesn't get done, we would go with @eddies all together.
So, If you want to get started on geojson, it could be best to base your branch on this PR

eddies:
@corneadoug thank you for offering to look into the CI failure. I just successfully re-ran `mvn clean package -Pspark-1.6 -Phadoop-2.6` on this branch locally. If I'm reading the travis log correctly, the CI errors are occurring because of:

```
ERROR org.apache.zeppelin.AbstractZeppelinIT:173 - Exception in SparkParagraphIT while testSpark
org.openqa.selenium.TimeoutException: Timed out after 60 seconds waiting for org.apache.zeppelin.AbstractZeppelinIT$1@76a123fb
```

eddies:
> I think the heatmap is a nice addition to the basic map. Maybe it could be toggled as an option instead of a different visualization button (just like the options in line chart)

I agree. In fact, what I really would want to see is an implementation of graphOptions that would, at a minimum, let the user select the lat/lon columns rather than the hard-coded column index assumptions.

But I figured we could call that a future enhancement request, rather than let this (or the original PR) languish any longer....

Madhuka:
@corneadoug : #765  is contain the basic map feature which is in #152. #765 all checks have passed and no build failures.
@eddies : Heatmaps is nice addition for the basic map visualization, This PR ( #728 after fixing this build failure) can easily lay on top of #765

randerzander:
@eddies what CSS tricks did you have to do to get GeoJson to work with Leaflet in Zeppelin?

I'm trying to follow the simple example here http://leafletjs.com/examples/geojson.html . It works in a standalone webpage, but the polygons never show in Zeppelin.

Any help would be much appreciated!

PR #730, opened 2016-02-20
[ZEPPELIN-680] Cron job will run cells that have had their run disabled

Leemoonsoo:
### What is this PR for?
When I run a cron job, cells that I have selected the "disable run" option will still run.
This PR fixes the problem.

### What type of PR is it?
Bug Fix

### Todos
* [x] - exclude disabled paragraph from runAll

https://issues.apache.org/jira/browse/ZEPPELIN-680?jql=project%20%3D%20ZEPPELIN

Create paragraph and disable it.
Then enable cron scheduling a notebook.
The paragraph supposed to not run.

N/A

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -91,6 +91,11 @@ public class Paragraph extends Job implements Serializable, Cloneable {
     return note;
   }
 
+  public boolean isEnabled() {
+    Boolean enabled = (Boolean) config.get("enabled");
+    return enabled == null || enabled.booleanValue();
+  }
+
   public String getRequiredReplName() {
     return getRequiredReplName(text);
   }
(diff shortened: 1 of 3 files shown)


Leemoonsoo:

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -91,6 +91,11 @@ public Note getNote() {
     return note;
   }
 
+  public boolean isEnabled() {
+    Boolean enabled = (Boolean) config.get("enabled");
+    return enabled == null || enabled.booleanValue();
```
felixcheung: shouldn't `== null` mean false?

Leemoonsoo: The default behavior (when the value "enabled" is undefined) need to be 'enabled'.
Some notebook that created before this "enabled" option introduced to Zeppelin, will have null value for it. And it think it should be considered as "enabled" in this case.

Make sense?

felixcheung: ok make sense

felixcheung:
looks good

Leemoonsoo:
Thanks @felixcheung for taking a look.
I'm merging it.

PR #463, opened 2015-11-24
[ZEPPELIN-395] Support Spark 1.6

felixcheung:
Adding support for Spark 1.6

Status:
- [x] Spark/Scala
- [x] PySpark
- [x] Spark SQL - broken - fixed
image
image

TODO:
- [x] update pom when the artifacts are on central repo
- [x] update travis to build 1.6
- [x] update doc (updated readme)

PR #735, opened 2016-02-20
ZEPPELIN-631 Allow user to configure spark.yarn.dist.files

jongyoul:
### What is this PR for?
This enables to append values into spark.yarn.dist.files

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix the logic to update values of spark.yarn.dis.files

### Is there a relevant Jira issue?
ZEPPELIN-631

### How should this be tested?
TBD

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #736, opened 2016-02-21
Fix pyspark to work on yarn mode when spark version is lower than or equal to 1.4.x

minahlee:
### What is this PR for?
pyspark.zip, py4j-\*.zip should be distributed to yarn nodes to make pyspark function but this hasn't been working after #463 because [`if (pythonLibs.length == pythonLibUris.size())`](https://github.com/apache/incubator-zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java#L329) condition will never be true. This PR fixes this issue by changing this if condition to be  `pythonlibUris.size() == 2`, while integer 2 refers pyspark.zip and py4j-\*.zip.

In addition, yarn-install documentation has been updated.

### What type of PR is it?
Bug Fix

No. But the issue has reported via [user mailing list](http://apache-zeppelin-users-incubating-mailing-list.75479.x6.nabble.com/Can-t-get-Pyspark-1-4-1-interpreter-to-work-on-Zeppelin-0-6-td2229.html#a2259) by Ian Maloney

### Questions:


spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -326,7 +326,10 @@ public SparkContext createSparkContext() {
       }
     }
     pythonLibUris.trimToSize();
-    if (pythonLibs.length == pythonLibUris.size()) {
+
+    // Distribute two libraries(pyspark.zip and py4j-*.zip) to workers
+    // when spark version is less than or equal to 1.4.1
+    if (pythonLibUris.size() == 2) {
       conf.set("spark.yarn.dist.files", Joiner.on(",").join(pythonLibUris));
```
Leemoonsoo: I think it always discard user provided `spark.yarn.dist.files`. Is it okay to do?

minahlee: I think #735 handles that issue.

Leemoonsoo:
Looks good to me. and merge if there're no more discussions

PR #745, opened 2016-02-24
[ZEPPELIN-697] Replace dynamic form with angular object from registry if exists

doanduyhai:
### What is this PR for?
Replace dynamic form with angular object from registry if exists

I updated the `Paragraph.jobRun()` method to look for existing variable from the Angular Object Registry first before displaying the dynamic form.

We look for Angular object having same name:

* first at paragraph scope (note id + paragraph id)
* then at note scope (note id only)

I did not look at **global** scope because @Leemoonsoo  was mentioning somewhere that we may likely remove the global scope some day

_This is a sub-task of epic **[ZEPPELIN-635]**_

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Code Review
* [ ] - Simple Test

**[ZEPPELIN-697]**

* `git fetch origin pull/745/head:AngularObjectReplaceDynamicFormVar`
* `git checkout AngularObjectReplaceDynamicFormVar`
* Create a new note
* In the first paragraph, put the following code

```html
%angular

<form class="form-inline">
  <div class="form-group">
    <label for="superheroId">Super Hero: </label>
    <input type="text" class="form-control" id="superheroId" placeholder="Superhero name ..." ng-model="superhero"></input>
  </div>
  <button type="submit" class="btn btn-primary" ng-click="z.angularBind('superhero', superhero, 'PUT_HERE_PARAGRAPH_ID'); z.runParagraph('PUT_HERE_PARAGRAPH_ID')"> Bind</button>
    <button type="submit" class="btn btn-primary" ng-click="z.angularUnbind('superhero','PUT_HERE_PARAGRAPH_ID'); z.runParagraph('PUT_HERE_PARAGRAPH_ID')"> Unbind</button>
</form>
</form>
```
* Create a second paragraph with the following code:
```scala
%md

### The superhero is : **${superhero}**
```
* In the first paragraph, replace the text PUT_HERE_PARAGRAPH_ID by the paragraph id of the second paragraph
* Execute first the second paragraph to see that the dynamic form system is working as usual
* Now put **Batman** in the input text of the first paragraph and click alternatively on **Bind** to see that dynamic form in the second paragraph is removed
* Click on **Unbind** to see that the dynamic form system is back

replacedynamicform

### Questions:
* Does the licenses files need update? --> **No**
* Is there breaking changes for older versions? --> **Yes**
* Does this needs documentation? --> **Yes**

[ZEPPELIN-635]: https://issues.apache.org/jira/browse/ZEPPELIN-635
[ZEPPELIN-697]: https://issues.apache.org/jira/browse/ZEPPELIN-697

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.notebook;
 
+import org.apache.zeppelin.display.AngularObject;
 import org.apache.zeppelin.display.AngularObjectRegistry;
 import org.apache.zeppelin.user.AuthenticationInfo;
 import org.apache.zeppelin.display.GUI;
@@ -232,6 +233,12 @@ public class Paragraph extends Job implements Serializable, Cloneable {
       String scriptBody = getScriptBody();
       Map<String, Input> inputs = Input.extractSimpleQueryParam(scriptBody); // inputs will be built
                                                                              // from script body
+
+      final AngularObjectRegistry angularRegistry = repl.getInterpreterGroup()
+              .getAngularObjectRegistry();
+
+      scriptBody = extractVariablesFromAngularRegistry(scriptBody, inputs, angularRegistry);
+
       settings.setForms(inputs);
       script = Input.getSimpleQuery(settings.getParams(), scriptBody);
     }
@@ -389,5 +396,26 @@ public class Paragraph extends Job implements Serializable, Cloneable {
   public Object clone() throws CloneNotSupportedException {
     Paragraph paraClone = (Paragraph) this.clone();
     return paraClone;
+  }
+
+  String extractVariablesFromAngularRegistry(String scriptBody, Map<String, Input> inputs,
+                                             AngularObjectRegistry angularRegistry) {
+
+    final String noteId = this.getNote().getId();
+    final String paragraphId = this.getId();
+
+    final Set<String> keys = new HashSet<>(inputs.keySet());
+
+    for (String varName : keys) {
+      final AngularObject paragraphScoped = angularRegistry.get(varName, noteId, paragraphId);
+      final AngularObject noteScoped = angularRegistry.get(varName, noteId, null);
+      final AngularObject angularObject = paragraphScoped != null ? paragraphScoped : noteScoped;
+      if (angularObject != null) {
+        inputs.remove(varName);
+        final String pattern = "[$][{]\\s*" + varName + "\\s*(?:=[^}]+)?[}]";
+        scriptBody = scriptBody.replaceAll(pattern, angularObject.get().toString());
+      }
+    }
+    return scriptBody;
   }
 }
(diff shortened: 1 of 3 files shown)


doanduyhai:
ping @Leemoonsoo for code review

doanduyhai:
ping @Leemoonsoo

Leemoonsoo:
The implementation itself looks good to me.

While this PR links 'dynamic form' and 'angular display system', I'd like to discuss little bit about the direction.

Zeppelin has 'dynamic form' and 'angular display system' for frontend - backend interaction.
And 'resource pool' for interpreter-interpreter interaction. I can summarize characteristics of these three, like

| | Dynamic Form  | Angular Display System | ResourcePool |
| ------------- | ------------- | ------------- | ------------- |
| Propagate value to the font-end  | yes  | yes | no  |
| Persisted in note.json | yes | yes | no |
| Trigger actions on change | yes | yes | no |
| Store arbitrary java object | no | yes | yes |
| Note / Paragraph scope | no | yes | yes |
| Access value between interpreters | no | no | yes |

'Dynamic Form' and 'Angular Display System' are both does frontend-backend interaction.
'Angular Display System' and 'Resource pool' are both object store.

While they're sharing similar characteristics, do you see any possibilities of consolidate these 3 into 1?
If it is, how this PR can be aligned?

doanduyhai:
Hello @Leemoonsoo 

Thank for raising question about the 3 types of variables. 

My though is that Angular Variables should be merged/replaced by ResourcePool. Indeed the Angular Object scope is not very intuitive. It is first scoped by **interpreter group** then by note/paragraph.

Removing Angular Object to use Resource Pool is a good idea but there is still one pending issue: angular value watcher. As you know people can assign a watcher function in the back-end on Angular value. If we want to replace Angular objects by Resource Pool, we need to provide similar feature, unless you think we can completely drop this watcher feature ...

Another point to take into account is the **migration path**. For all people that have saved their notes as JSON with saved Angular objects, we need to handle the migration to Resource pool when the re-import those notes.

About the Dynamic Form system I don't think we should merge it with Angular Object/Resource Pool.  Dynamic Form values are just an alias to automatically generate a form input. And the fact that its value change triggers the execution of the current paragraph can be implemented using `z.runParagraph()` Angular function that is now available.

All in all I think we should ultimately have enhanced Resource Pool and Dynamic Form only.

But those changes are very big and deserve another JIRA epic. For now, we can just implement this overriding system for this PR and then when we refactor the Angular JS to Resource Pool, we will merge the behavior together

WDYT ?

Leemoonsoo:
I agree, consolidate Angular Object / Resource Pool into Resource Pool really make sense. I think watcher also can be handled.
Also make sense Dynamic form doesn't really need to be consolidated. 

I think the way how this links PR Resource Pool and Dynamic Form make sense and aligned with the future plan.

Thanks for the patch and sharing the idea. LGTM

doanduyhai:
@Leemoonsoo Merge if ok ? Next 2 PR will concentrate on keeping same paragraph Id when re-importing node from JSON and add comprehensive documentation for the new AngularJS `z` functions

Leemoonsoo:
Merge into master if there're no more discussions.

PR #749, opened 2016-02-26
Remove duplicated java option concats in common.sh

zhongneu:
### What is this PR for?
There are some java option concats in common.sh, which are executed twice when start an interpreter. This makes some of the options invalid, such as remote debugging options

### What type of PR is it?
Bug Fix

[ZEPPELIN-686](https://issues.apache.org/jira/browse/ZEPPELIN-686)

Set remote debug options:
```
export ZEPPELIN_INTP_JAVA_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,address=8000,suspend=n
```

Start the Zeppelin daemon, then create & run a job to trigger starting an interpreter. The job should fail without the fix.

### Questions:
NO

NO

NO

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -96,7 +96,7 @@ if [[ -z "$ZEPPELIN_MEM" ]]; then
   export ZEPPELIN_MEM="-Xms1024m -Xmx1024m -XX:MaxPermSize=512m"
 fi
 
-JAVA_OPTS=" ${ZEPPELIN_JAVA_OPTS} -Dfile.encoding=${ZEPPELIN_ENCODING} ${ZEPPELIN_MEM}"
+JAVA_OPTS+=" ${ZEPPELIN_JAVA_OPTS} -Dfile.encoding=${ZEPPELIN_ENCODING} ${ZEPPELIN_MEM}"
 export JAVA_OPTS
 
 # jvm options for interpreter process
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @zhongneu for the patch.
Looks good to me.

Leemoonsoo:
Merge if there're no more discsussions

felixcheung:
what if someone has set JAVA_OPTS in the env? shouldn't we be additive?

zhongneu:
@felixcheung I am not sure about this, but I would like to make it additive if we have such case.

Leemoonsoo:
I think @felixcheung got valid point. It's incompatible change to people who has JAVA_OPTS.
One alternative is document changes in 
https://github.com/apache/incubator-zeppelin/blob/master/docs/install/upgrade.md

felixcheung:
That sounds like a reasonable approach, we could say something like
"if you are setting JAVA_OPT, set ZEPPELIN_JAVA_OPTS instead for Zeppelin, set ZEPPELIN_INTP_JAVA_OPTS instead for interpreter process" in more words ;)

zhongneu:
I've reverted the change for JAVA_OPTS, which is fine, because it is not used by bin/interpreter.sh

Leemoonsoo:
LGTM

felixcheung:
looks good. I think we might want to deprecate JAVA_INTP_OPTS and replace with something more Zeppelin specific for clarify, and allow for admin to override it.
merging if no more comment.

PR #754, opened 2016-02-28
[ZEPPELIN-679] only print the error message for SQL exceptions

zhongneu:
### What is this PR for?

Most of the SQL exceptions are syntax errors. Printing the full stack trace is not necessary. We should only print the error message to avoid distracting the users

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-679](https://issues.apache.org/jira/browse/ZEPPELIN-679)

screen shot 2016-02-26 at 11 44 30 pm

### Questions:
NO

NO

NO

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
@@ -137,6 +137,7 @@ public class SparkSqlInterpreter extends Interpreter {
       Method sqlMethod = sqlc.getClass().getMethod("sql", String.class);
       rdd = sqlMethod.invoke(sqlc, st);
     } catch (InvocationTargetException ite) {
+      logger.error("Invocation target exception", ite);
       return new InterpreterResult(Code.ERROR, ite.getTargetException().getMessage());
     } catch (NoSuchMethodException | SecurityException | IllegalAccessException
         | IllegalArgumentException e) {
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
@@ -136,12 +136,10 @@ public class SparkSqlInterpreter extends Interpreter {
       // Therefore need to use reflection to keep binary compatibility for all spark versions.
       Method sqlMethod = sqlc.getClass().getMethod("sql", String.class);
       rdd = sqlMethod.invoke(sqlc, st);
+    } catch (InvocationTargetException ite) {
+      return new InterpreterResult(Code.ERROR, ite.getTargetException().getMessage());
     } catch (NoSuchMethodException | SecurityException | IllegalAccessException
-        | IllegalArgumentException | InvocationTargetException e) {
-      if (e instanceof InvocationTargetException) {
-        return new InterpreterResult(Code.ERROR,
-                ((InvocationTargetException) e).getTargetException().getMessage());
-      }
+        | IllegalArgumentException e) {
       throw new InterpreterException(e);
     }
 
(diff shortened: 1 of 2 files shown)


spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -138,6 +138,10 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
       rdd = sqlMethod.invoke(sqlc, st);
     } catch (NoSuchMethodException | SecurityException | IllegalAccessException
         | IllegalArgumentException | InvocationTargetException e) {
+      if (e instanceof InvocationTargetException) {
+        return new InterpreterResult(Code.ERROR,
+                ((InvocationTargetException) e).getTargetException().getMessage());
+      }
```
felixcheung: would it be clearer if it is
```
 } catch (NoSuchMethodException | SecurityException | IllegalAccessException
          | IllegalArgumentException e) {
         throw new InterpreterException(e);
 } catch (InvocationTargetException e) {
         return new InterpreterResult(Code.ERROR,
                 ((InvocationTargetException) e).getTargetException().getMessage());
 }
```

zhongneu:
@felixcheung definitely. updated based on your comment

felixcheung:
thx - let's add logger to capture the full stack before returning with InterpreterResult?
```
logger.error("Invocation target exception", e);
```
?

zhongneu:
@felixcheung great idea. updated & tested. full stack trace will be logged in the interpreter log

felixcheung:
LGTM

felixcheung:
btw, it looks like this might be a legitimate test failure, could you take a look?
https://s3.amazonaws.com/archive.travis-ci.org/jobs/112358128/log.txt
```
Failed tests: 
  SparkSqlInterpreterTest.test:110 Exception not catched
```

zhongneu:
fixed the test

echarles:
I often use the stacktraces to debug what's going wrong. With this, I expect I will miss information. Can we make this configurable (default being "not verbose")?

zhongneu:
@echarles I thought about made this configurable, but I thought the stacktraces might not be helpful for SQL queries, because typically it is difficult for the users to dig into the internal Spark implementation. Could you give me examples of using stacktraces to debug SQL queries?

echarles:
@zhongneu I will rather give you a message a user could face running the SQL: `java.lang.IllegalArgumentException: requirement failed` What will the guy responsible to troubleshoot this ticket will do without the stacktrace?

zhongneu:
@echarles Thanks, Eric! It is good to know we have such cases. We should definitely leave it as configurable if Spark doesn't provide informative error message. I am not sure, but probably in older versions of Spark, there may be more such cases.

Just curious, what kind of queries can result in such exception?

nemccarthy:
this seems a little misguided. hiding errors is great until you need to stack trace. 

id vote to have it configurable where the default is always full stack traces are visible.

joelvanveluwen:
+1 for making it configurable. I find the stack trace invaluable!

echarles:
So we all agree to make it configurable on the interpreter settings leve

`zeppelin.spark.sql.stacktrace = false` by default ?

What about when you catch and retrhow the expection to append a message like "Configure zeppelin.spark.sql.stacktrace = true to see the stacktrace, or use sqlc.sql(query)"

We would have best of both worlds.

@ zhongneu don't have the query at hand, anyway, the stacktrace helped me to identify an bug unrelated to sql but related to storage access.

zhongneu:
I agree to make it configurable, but show the shortened message by default

@echarles I think that's a good suggestion.

zhongneu:
Added a configuration to show full stacktrace

Leemoonsoo:
LGTM

felixcheung:
LGTM. @echarles what do you think?

echarles:
Tested with zeppelin.spark.sql.stacktrace being false and true, and it works fine.

Good to merge.

PS: With HEAD source, I have issues settings any property via the UI (When I restart, the changed values are not taken into account) - Had to set them changing the interpreter.json. Is this only me on my env?

AhyoungRyu:
@echarles I faced the same issue. I think it's a bug..

echarles:
@AhyoungRyu I have opened ZEPPELIN-725 for the interpreter update issue.

@felixcheung @Leemoonsoo This PR #754 is good to merge.

AhyoungRyu:
@echarles Thanks for creating the issue!

Leemoonsoo:
Thanks i'm merging it into master.

PR #755, opened 2016-03-01
ZEPPELIN-711 Load dependencies at startup

karup1990:
### What is this PR for?
Load  dependencies at startup.
When i try to re-use `interpreter.json` in a different zeppelin instance, the dependencies are not loaded and my paras fail.
If there are some deps configured in the interpretersetting, loading them at startup.

### What type of PR is it?
Improvement

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-711

Add some dependencies to interpreter settings.
Stop zeppelin daemon
Clean up local-repo folder
Start zeppelin server.
The dependencies should have been downloaded.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -243,7 +243,14 @@ public class InterpreterFactory {
           setting.getProperties(),
           setting.getDependencies(),
           setting.getOption());
-
+      try {
+        logger.info("Loading dependencies for " + "interpreter setting "
+            + setting.getName());
+        loadInterpreterDependencies(intpSetting);
+      } catch (RepositoryException e) {
+        logger.error("Failed to load dependencies for interpretersetting "
+            + intpSetting.getName() + ":" + e.getMessage());
+      }
       InterpreterGroup interpreterGroup = createInterpreterGroup(setting.id(), setting.getOption());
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -243,7 +243,14 @@ private void loadFromFile() throws IOException {
           setting.getProperties(),
           setting.getDependencies(),
           setting.getOption());
-
+      try {
+        logger.info("Loading dependencies for " + "interpreter setting "
+            + setting.getName());
+        loadInterpreterDependencies(intpSetting);
+      } catch (RepositoryException e) {
+        logger.error("Failed to load dependencies for interpretersetting "
```
felixcheung: do `logger.error("message", e)` instead?

minahlee:
Tested, looks good to me

Leemoonsoo:
@minahlee One question, isn't it supposed to load dependency library when the time interpreter creation?

minahlee:
@Leemoonsoo that sounds like more appropriate approach. @karup1990 what do you think?

karup1990:
Yeah, it makes sense to load dependencies when the interpreter process starts.

felixcheung:
@karup1990 would you be updating this PR?

karup1990:
Will try to fix this over the weekend @felixcheung

karup1990:
Closing the PR for now. Will fix this rightly and reopen.

PR #760, opened 2016-03-04
[ZEPPELIN-717] protect the whole spark repl init process

zhongneu:
### What is this PR for?
ZeppelinContext may not be initialized properly in concurrent mode:

When I create & run multiple notebooks using rest API concurrently, I can see such errors if the jobs trying to use ZeppelinContext:
```
<console>:23: error: not found: value z
```

I think this issue can be reproduced by:
1. create 4 - 5 new notebooks with content: `println(z)`
2. use rest API to run the newly created notebooks concurrently

It seems the issue is gone after I expand the lock to protect the whole process of spark REPL initializing / binding. 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-717](https://issues.apache.org/jira/browse/ZEPPELIN-717)

1. create 4 - 5 new notebooks with content: `println(z)`
2. use rest API to run the newly created notebooks concurrently

### Questions:
NO

NO

NO

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -481,18 +481,18 @@ public class SparkInterpreter extends Interpreter {
 
     System.setProperty("scala.repl.name.line", "line" + this.hashCode() + "$");
 
-    /* create scala repl */
-    this.interpreter = new SparkILoop(null, new PrintWriter(out));
-
-    interpreter.settings_$eq(settings);
-
-    interpreter.createInterpreter();
-
-    intp = interpreter.intp();
-    intp.setContextClassLoader();
-    intp.initializeSynchronous();
-
     synchronized (sharedInterpreterLock) {
+      /* create scala repl */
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @zhongneu for finding issue and provide a fix.
LGTM

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -481,18 +481,18 @@ public void open() {
 
     System.setProperty("scala.repl.name.line", "line" + this.hashCode() + "$");
 
-    /* create scala repl */
-    this.interpreter = new SparkILoop(null, new PrintWriter(out));
+    synchronized (sharedInterpreterLock) {
+      /* create scala repl */
+      this.interpreter = new SparkILoop(null, new PrintWriter(out));
```
felixcheung: should we check if this.interpreter != null if another thread has initialized it?

zhongneu: @felixcheung I don't think that's necessary since interpreter is a read-only field

Leemoonsoo:

PR #764, opened 2016-03-07
[ZEPPELIN-688] Giving an option to hide REPL output in Spark interpreter 

AhyoungRyu:
### What is this PR for?
When a user runs Spark interpreter, the result will come out with result number message like :
```
res0: Int = 250
```
Someone might want to print this REPL output with their result, but others may want to see the result only since sometimes this output is too verbose. So, I just want to give an option to hide this REPL output to those users. 

The default value of `zeppelin.spark.printREPLOutput` is `true`. This status is as-is. Users can hide REPL output only when they changes this property `true` to `false`. In this case, they can check the result by specifying such as `print(some_variable)`.

### What type of PR is it?
Improvement

### Todos
* [x] - Add a property `zeppelin.spark.printREPLOutput` 
* [x] - Add Spark interpreter property table to `docs/spark.md`

### What is the Jira issue?
[ZEPPELIN-688](https://issues.apache.org/jira/browse/ZEPPELIN-688#)

After applying this PR,
  1. Create spark interpreter for this test and change `zeppelin.spark.printREPLOutput` property value `true` -> `false`
  2. Create a notebook and bind interpreter what you made.
  3. Write `val a = 250` down and run this paragraph. Then you can check the any output is not shown although paragraph status is **FINISHED** (This is the result of this PR).
  4. Run `print(a)` in the next paragraph. Then finally you can get a result `250`.

spark-result

### Questions:
* Does this needs documentation? I added Spark interpreter property table to `docs/spark.md`.

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -40,6 +40,73 @@ Spark Interpreter group, which consisted of 4 interpreters.
 </table>
 
 ## Configuration
+Zeppelin provides the below properties for Spark interpreter. 
+<table class="table-configuration">
+  <tr>
+    <th>Property</th>
+    <th>Default</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>args</td>
+    <td></td>
+    <td>Spark commandline args</td>
+  </tr>
+    <td>master</td>
+    <td>local[*]</td>
+    <td>Spark master uri. <br/> ex) spark://masterhost:7077</td>
+  <tr>
+    <td>spark.app.name</td>
+    <td>Zeppelin</td>
+    <td>The name of spark application.</td>
+  </tr>
+  <tr>
+    <td>spark.cores.max</td>
+    <td></td>
+    <td>Total number of cores to use. <br/> Empty value uses all available core.</td>
+  </tr>
+  <tr>
+    <td>spark.executor.memory </td>
+    <td>512m</td>
+    <td>Executor memory per worker instance. <br/> ex) 512m, 32g</td>
+  </tr>
+  <tr>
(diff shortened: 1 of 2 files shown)


docs/interpreter/spark.md
```diff
@@ -40,6 +40,73 @@ Spark Interpreter group, which consisted of 4 interpreters.
 </table>
 
 ## Configuration
+Zeppelin provides the below properties for Spark interpreter.
```
Leemoonsoo: What do you think mentioning that user can add any spark properties even though they're not listed by Zeppelin?

AhyoungRyu:
@Leemoonsoo Thanks for your suggestion. I've just added a sentence as you said : )
Plus, I rebased since #760 and this PR both changed `sparkInterpreter.java`.

docs/interpreter/spark.md
```diff
@@ -40,6 +40,74 @@ Spark Interpreter group, which consisted of 4 interpreters.
 </table>
 
 ## Configuration
+Zeppelin provides the below properties for Spark interpreter. 
+You can also set other Spark properties which are not listed in the table. If so, please refer to [Spark Available Properties](http://spark.apache.org/docs/latest/configuration.html#available-properties).
```
AhyoungRyu: @Leemoonsoo I added this line. Is it okay with you ?

Leemoonsoo: Great. Thanks!

Leemoonsoo:
Thanks. it looks good to me!

Leemoonsoo:

felixcheung:
LGTM.

PR #768, opened 2016-03-08
Interpeter settings update must set the properties

echarles:
### What is this PR for?

If you change the intrepreter setting via the UI, the updated values are sent to the REST API but the JSON response coming back from the server contains the previous values (not the updated ones).

This is a regression probably introduced recently.

### What type of PR is it?
[Hot Fix]

### Todos
Nil

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-725

### Questions:


AhyoungRyu:
@echarles I tested and it works well. Thanks for the fixing ! : )

echarles:
@AhyoungRyu Thx. I am now facing another issue when registering an additional interpreter (nothing related to this PR, but I wonder if recent #703 could be the cause). Will investigate and report here.

Leemoonsoo:
@echarles Thanks for the fix. @AhyoungRyu Thanks for verifying it.
I also tested this PR and works well.

I'm merging it into master as a hotfix.

PR #771, opened 2016-03-09
Fix Selenium test url

r-kamath:
### What is this PR for?
use env variable `url` instead of system property

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* ZEPPELIN-729

Change the domain name or port and set a new env variable called `url` with the new url


prabhjyotsingh:
Good find, LGTM.

felixcheung:
LGTM
merging if no more comment

PR #775, opened 2016-03-14
Add documentation about setup HTTP basic auth via NGINX

yoonjs2:
### What is this PR for?
Provide a guide to setup basic authentication using NGINX with detailed instruction for users who wants to use NGINX as a separate authentication server. Shiro Security is preferred way to provide a basic authentication for future releases, but we may need to provide a way to workaround when user need other options.

### What type of PR is it?
Documentation

### Questions:


docs/security/authentication.md
```diff
@@ -23,6 +23,122 @@ Authentication is company-specific.
...
+    ```
+    Then check HTTP Basic Authentication works in browser. If you can see regular basic auth popup and then able to login with credential you entered into `.htpasswd` you are good to go.
+
+    <img src="/assets/themes/zeppelin/img/screenshots/authentication-basic-auth-nginx-request.png" />
+    <img src="/assets/themes/zeppelin/img/screenshots/authentication-basic-auth-nginx-https.png" />
+
+1. More security consideration
+
+* Using HTTPS connection with Basic Authentication is highly recommended since basic auth without encryption may expose your important credential information over the network.
+* Using [Shiro Security feature built-into Zeppelin](https://github.com/apache/incubator-zeppelin/pull/53) is recommended if you prefer all-in-one solution for authentication but NGINX may provides ad-hoc solution for re-use authentication served by your system's NGINX server or in case of you need to separate authentication from zeppelin server.
```
Leemoonsoo: Thanks for very useful document.
I think `https://github.com/apache/incubator-zeppelin/blob/master/SECURITY-README.md` is better address for link to shiro security feature.

yoonjs2:
@Leemoonsoo Thanks for the comment. I fixed link to the address you've suggested to me. :)

Leemoonsoo:
Thanks, Looks good to me!

Leemoonsoo:

AhyoungRyu:
Looks good :+1:

PR #776, opened 2016-03-14
Allow custom location for interpreter.json and die if not writeable

michrawson:
### What is this PR for?
This will allow for custom location for interpreter.json.
This will try to create dir if needed.
This will die if file or file's dir not writeable. 

### What type of PR is it?
Bug Fix: Zeppelin can't manipulate notes when deployed to a write only directory.

1. Test default: when not using a custom file, everything should work as usual.
2. Specify any file via zeppelin-site.xml AND zeppelin-env.sh and that file should be used.
3. If the file exists and is not writeable, zeppelin shouldn't start.
4. If the file's dir doesn't exit, try to create it.
5. If the file doesn't exist and the file's dir is not writeable, zeppelin shouldn't start.

### Questions:
No
No
* Does this need documentation?
Unknown

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -32,6 +32,10 @@ if [[ -z "${ZEPPELIN_CONF_DIR}" ]]; then
   export ZEPPELIN_CONF_DIR="${ZEPPELIN_HOME}/conf"
 fi
 
+if [[ -z "${ZEPPELIN_INTERPRETER_CONF}" ]]; then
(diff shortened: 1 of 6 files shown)


Leemoonsoo:
Thanks for the patch.

I think `bin/interpreter.sh` doesn't need to be modified while interpreter process does not access `conf/interpreter.json`.

And It would be better If new configuration can be seen in
https://github.com/apache/incubator-zeppelin/blob/master/docs/install/install.md#zeppelin-configuration

michrawson:
Fixed/Re-sent.

PR #780, opened 2016-03-15
[ZEPPELIN-658] Scala: not accepting companion objects if defined in different lines

felixcheung:
### What is this PR for?
In Scala, a singleton companion object can be defined along the class. Because of the way code lines are split, the interpreter is not passing the companion object along with the class code, or indicating any possible line continuation (eg. leaving out '}'). This causes an error with the Scala REPL/compiler.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix line splitting
* [x] - Add tests

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-658
https://issues.apache.org/jira/browse/ZEPPELIN-747

Add a new paragraph with Scala/Spark interpreter and define a class with its companion object.
See bug for more details

N/A

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -774,13 +774,34 @@ public class SparkInterpreter extends Interpreter {
     context.out.clear();
     Code r = null;
     String incomplete = "";
+    boolean inComment = false;
 
     for (int l = 0; l < linesToRun.length; l++) {
       String s = linesToRun[l];
       // check if next line starts with "." (but not ".." or "./") it is treated as an invocation
       if (l + 1 < linesToRun.length) {
         String nextLine = linesToRun[l + 1].trim();
-        if (nextLine.startsWith(".") && !nextLine.startsWith("..") && !nextLine.startsWith("./")) {
+        boolean continuation = false;
+        if (nextLine.isEmpty()
+           || nextLine.startsWith("//")         // skip empty line or comment
+           || nextLine.startsWith("}")
+           || nextLine.startsWith("object")) {  // include "} object" for Scala companion object
+          continuation = true;
+        } else if (!inComment && nextLine.startsWith("/*")) {
+          inComment = true;
+          continuation = true;
+        } else if (inComment && nextLine.lastIndexOf("*/") >= 0) {
+          inComment = false;
+          continuation = true;
+        } else if (nextLine.length() > 1
+                && nextLine.charAt(0) == '.'
+                && nextLine.charAt(1) != '.'     // ".."
+                && nextLine.charAt(1) != '/') {  // "./"
+          continuation = true;
+        } else if (inComment) {
+          continuation = true;
+        }
+        if (continuation) {
           incomplete += s + "\n";
           continue;
         }
--- a/spark/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java
+++ b/spark/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java
@@ -142,6 +142,17 @@ public class SparkInterpreterTest {
   }
 
   @Test
+  public void testNextLineComments() {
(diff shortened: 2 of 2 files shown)


spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -774,13 +774,34 @@ public InterpreterResult interpretInput(String[] lines, InterpreterContext conte
...
+        } else if (!inComment && nextLine.startsWith("/*")) {
+          inComment = true;
+          continuation = true;
+        } else if (inComment && nextLine.lastIndexOf("*/") >= 0) {
+          inComment = false;
+          continuation = true;
+        } else if (nextLine.length() > 1
+                && nextLine.charAt(0) == '.'
+                && nextLine.charAt(1) != '.'     // ".."
+                && nextLine.charAt(1) != '/') {  // "./"
```
prabhjyotsingh: Earlier repl.interpret("\"123\"\n.\ntoInt", context) used to work as well, now it doesn't.

felixcheung: you mean like
```
"123"
.
toInt
```
?
That looks rather odd...

prabhjyotsingh: Yes, I meant that, and agree looks odd. I think we can leave this case.

felixcheung:
Any more comment?

Leemoonsoo:
Looks good to me.

prabhjyotsingh:
LGTM

saj1th:
Still getting this issue with  Zeppelin 0.6.2 while trying to use companion object

```
class Event(val nuid: String) extends Product {
  override def productElement(n: Int): Any = {
    n match {
      case 0 => nuid
    }
  }
  override def productArity: Int = 1
  override def canEqual(that: Any): Boolean = that.isInstanceOf[Event]
}
object Event {
  def apply(row: Row): Event = {
    if (row.isNullAt(16)) {
      new Event("")
    } else {
      new Event(row.getString(16))
    }
  }
}
```

Any pointers ?

rodbs:
As @saj1th says, I'm still getting this issue with Zeppelin 0.7 while trying to use companion object; even placing the object definition in the same line as the class definition:

`
class Account {
  val id = Account.newUniqueNumber()
  private var balance = 0.0
  def deposit(amount: Double) { balance += amount }
  def description = "Account " + id + " with balance " + balance
};  object Account { // The companion object
  private var lastNumber = 0
  private def newUniqueNumber() = { lastNumber += 1; lastNumber }
}


val acct1 = new Account
val acct2 = new Account
acct1.deposit(1000)
val d1 = acct1.description
val d2 = acct2.description`

felixcheung:
thanks for reporting this. let me try to test in out in a couple of days

GichanLee:
version 0.7.3 same issue, It can't find case class, I think the Interpreter support :paste mode.
```
case class SensorRow(rowkey: String, sector: Double, steam: Double, usage: Double, Log: String, Status: String); 
object SensorRow extends Serializable { 
    def parseSensorRow(result: Result): SensorRow = {
    val rowkey = Bytes.toString(result.getRow())
      // remove time from rowKey, stats row key is for day
      val p0 = rowkey.split(" ")(0)
      val p1 = Bytes.toDouble(result.getValue(cfDataBytes, Bytes.toBytes("sector")))
      val p2 = Bytes.toDouble(result.getValue(cfDataBytes, Bytes.toBytes("steam")))
      val p3 = Bytes.toDouble(result.getValue(cfDataBytes, Bytes.toBytes("usage")))
      val p4 = Bytes.toString(result.getValue(cfDataBytes, Bytes.toBytes("Log")))
      val p5 = Bytes.toString(result.getValue(cfDataBytes, Bytes.toBytes("Status")))
      SensorRow(p0, p1, p2, p3, p4, p5)
    }
}
```
ERROR console message
```
defined class SensorRow
<console>:183: error: SensorRow.type does not take parameters
             SensorRow(p0, p1, p2, p3, p4, p5)

```

PR #788, opened 2016-03-21
Add username in place of Connected

ravicodder:
### What is this PR for?
To show more information on the navigation bar about user.
 - show 'Connected' if it is an anonymous user.
 - show the username if it is not an anonymous user.

### What type of PR is it?
Improvement

### Todos
* [x] - Change HTML

Case 1 :   Anonymous User
                to be an anonymous user, make the following changes in /conf/shiro.ini  file
               1. Uncomment                    /** = anon
               2. Comment                       /** = authcBasic
               3. Restart Zeppelin.

case 2:    Non-anonymous User
               to be a non-anonymous user, make the following changes in /conf/shiro.ini  file
               1. Comment                     /** = anon
               2. Uncomment                 /** = authcBasic
               3. Restart Zeppelin
               4. On Zeppelin Web page it will ask for username and password. Use one of the username and    password from the list  given in conf/shiro.ini file to login.

For example, if anonymous user it will show 'Connected'.
screen shot 2016-03-21 at 12 14 53 pm
And for non -anonymous user for example a user with userid 'admin'  will show the username as below
screen shot 2016-03-21 at 12 16 28 pm

### Questions:


ravicodder:
Ready for review.

corneadoug:
LGTM, however we might want to limit the size of the username with a truncate

ravicodder:
Added limit to size and we can see the complete username on tool tip 
screen shot 2016-03-22 at 4 41 29 pm

Leemoonsoo:
LGTM

Leemoonsoo:
I'm merging if there're no more discussions

PR #795, opened 2016-03-25
[ZEPPELIN-183] Interpreter process per note

Leemoonsoo:
### What is this PR for?

This PR give an option to create interpreter process per notebook.

### What type of PR is it?
Feature

### Todos
* [x] - Interpreter process per note
* [x] - Interpreter menu gui change
* [x] - unittest
* [x] - doc update

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-183

Change interpreter setting option to 'process' and then create two notebooks.
Try to use both notebook with the same interpreter setting.
You'll see each notebook creates different interpreter process.

For example spark interpreter, if you run 
```
%spark
sc
```
in two notebooks, then you'll see different object id.

Interpreter setting option gui looks like

image

If you hover the pointer on each option, you'll see tooltip help about each mode.

### Questions:


EronWright:
The 'process' option should be the default option for Spark in my opinion, since it solves so many issues.    A note with dependencies cannot be reliably loaded in any other way.

Have you tested that dependencies work independently in the 'process' mode?   

Again thanks, this is great.

Leemoonsoo:
@EronWright Thanks for the opinion. I prefer leave 'shared' as a default value for now, and make default value configurable in a separate issue. would it okay for you?

And yes it handles dependencies (%dep) independently in the 'process' mode.

echarles:
I don't in the src any change on the SparkInterpeter to deal with this new conf.

About default, sharing variables between notes is sometimes not the goal, but sometimes it is exactly what you want (I often work in the latter way than the former) - leaving defauilt to `shared` sounds good to me.

Not all intrerpreter would support all those modes, especially the `process` one. If we want a full solution, we should be able to configure a list of supported mode per interpreter...

Btw, to understand what `process` means, I had to RTFM, which is not the best. Maybe we should have a look to other interpreter cases and see if there are other ideas.

Things will still get harder when Zeppelin will support true multi user: there would be options to share (or not) context between users.

echarles:
Also, should non `shared` modes be marked as `experimental` (see ZEPPELIN-761). There could be unexpected side effects while using the Spark REPL in non standard ways.

Leemoonsoo:
@echarles Thanks for the feedbacks.

Regarding ZEPPELIN-761, it is a problem of way SparkInterpreter implements 'instance' mode rather than a problem of 'instance' mode itself. So i'd like to mark experimental it in a spark interpreter documentation. But 'process' mode is simply creating multiple processes of spark interpreter, so it's still standard way of using Spark REPL.

EronWright:
@Leemoonsoo leaving the default as-is sounds fine.   I agree with @echarles that the terminology of 'process' confusing.   "Isolated" or "dedicated" maybe.

felixcheung:
and what does "instance" mean?

Leemoonsoo:
I think 'isolated', 'dedicated' can be applied to both 'instance' and 'process'. isn't it?

Here is what i was intended

'shared' : single interpreter instance on single interpreter process serves all notes.
'instance' : separate interpreter instances on single interpreter process serves each notes.
'process' : separate interpreter processes serves each notes.

What could be the best name and description of those 3 to display in interpreter page?

EronWright:
How about the options are renamed as follows:

- "shared" -> "shared"
- "instance" -> "scoped"
- "process" -> "isolated"

Leemoonsoo:
@EronWright sounds reasonable

jongyoul:
@Leemoonsoo I think 'process' and 'instance' are a little bit ambiguous to understand. How about dividing 'shared' and 'isolated'?

Leemoonsoo:
Changed to

* shared
* scoped
* isolated

EronWright:
@Leemoonsoo one of the paragraph tests failed with a timeout.   Please re-execute the CI build.

vijikarthi:
I am not sure if the build is in working state or not? The dropdown value selection does not work and it always defaults to "shared"

Leemoonsoo:
@vijikarthi To enable dropdown selection, you need to click 'edit' button.

vijikarthi:
@Leemoonsoo I was trying to change the option in "edit" mode but still the selected item was not picked up by the list control.

vijikarthi:
Thoughts on multi-user support? - By extending this implementation, if we offer "per note process" for every user account ("isolated" per user session) and by passing the user identity (logged in user) to SPARK_SUBMIT command (SPARK_SUBMIT_OPTIONS="--proxy-user <USER>"), we could leverage the underlying "proxy" user support provided by the "spark-submit" doe that takes care of running the job (notes) in user context (yarn mode). This should work for both secure (kerberos) and insecure environment and it addresses multi-tenancy support (in addition to ZEPPELIN-773 which uses livy interpreter). The only caveat is the number of interpretor process increases as user base increases?

Leemoonsoo:
@vijikarthi I have pushed more commits with fix. Could you try it again?

Zeppelin already passes [AuthenticationInfo](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java) to interpreter process through [InterpreterContext](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterContext.java#L52). I think Spark interpreter can use this information and set proxy-user programatically.

Leemoonsoo:
Now CI is green and ready to merge

vijikarthi:
@Leemoonsoo The latest commits seems to be working. I could see two different SC getting created in "isolated" mode. However when the second SparkContext is getting created, I am hitting an exception which is related to "SPARK-10872"

Regarding proxy user support, I am not sure if we can handle it inside "SparkInterpreter". The way spark-submit handles proxy-user is by running the main process itself as the proxy user (see https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L160). In our case if we run the "RemoteInterpreterServer" as the proxy user, the rest of the flow is kind of taken care automatically. I have tested with/without Kerberos (by adding  SPARK_SUBMIT_OPTIONS="--proxy-user vijay -v" in bin/zeppelin-env.sh) and it works fine.

Leemoonsoo:
@vijikarthi Thanks for testing. 'SPARK-10872' is expected while default configuration of sparkcontext creates 'metastore_db' in the same location. If you configure hive-site.xml to use mysql or something else  for it's metastore, the errors will gone. Alternatively, you can set 'zeppelin.spark.useHiveContext' to 'false' in Interpreter menu, if you're not relying on HiveQL.

Regarding proxy user support, If passing "--proxy-user" argument in SPARK_SUBMIT_OPTIONS is the only way, Zeppelin needs extra mechanism to pass this argument. Thanks for sharing the details and user test result!

vijikarthi:
+1

zeppelin-web/src/app/interpreter/interpreter-create/interpreter-create.html
```diff
@@ -36,9 +36,39 @@
...
+              <li>
+                <a style="cursor:pointer"
+                   tooltip="Single interpreter instance are shared across notes"
+                   ng-click="setSessionOption(setting.id, 'shared')">
+                  shared
+                </a>
+              </li>
+              <li>
+                <a style="cursor:pointer"
+                   tooltip="Separate Interpreter instance for each note"
```
felixcheung: should this tooltip be updated? it looks the same as the isolated one

felixcheung: ah, never mind

felixcheung:
LGTM

JHowix:
So if I am understanding correctly, with this merged in we can now get one interpreter per notebook when we set the interpreter option to "isolated".  Did this also include functionality to restart the interpreter at the notebook level?  If so how does one do that?  Right now I can go to the interpreter page and click restart but I think that restarts ALL interpreters of that type.

Leemoonsoo:
@JHowix It does not include functionality to restart interpreter at the notebook level.
I think it really make sense restart interpreter at the notebook level in case of 'scoped' or 'isolated'.
Do you mind file a jira issue?

JHowix:
Done, thank you for the quick response!  https://issues.apache.org/jira/browse/ZEPPELIN-1049

PR #798, opened 2016-03-27
Fix removing notes without specified interpreter settings

khalidhuseynov:
### What is this PR for?
This PR fixes incorrect behavior when trying to remove note without saving its interpreter binding settings.

### What type of PR is it?
Bug Fix

### Todos
* [x] - don't let null return value

### What is the Jira issue?
[ZEPPELIN-765] (https://issues.apache.org/jira/browse/ZEPPELIN-765)

Try to remove some notebook without saving its interpreter bindings when first entering it

Before: 
remove_before

After:
remove_after

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -627,7 +627,9 @@ public class InterpreterFactory {
 
   public void removeNoteInterpreterSettingBinding(String noteId) {
     synchronized (interpreterSettings) {
-      List<String> settingIds = interpreterBindings.remove(noteId);
+      @SuppressWarnings("unchecked")
+      List<String> settingIds = interpreterBindings.containsKey(noteId) ?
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -627,7 +627,9 @@ public void putNoteInterpreterSettingBinding(String noteId,
 
   public void removeNoteInterpreterSettingBinding(String noteId) {
     synchronized (interpreterSettings) {
-      List<String> settingIds = interpreterBindings.remove(noteId);
+      @SuppressWarnings("unchecked")
+      List<String> settingIds = interpreterBindings.containsKey(noteId) ?
+          interpreterBindings.remove(noteId) : Collections.EMPTY_LIST;
```
anthonycorbacho: I think `Collections,emptyList();` is better here (will remove the `@SuppressWarnings("unchecked")`)

bzz:
Looks good to me. 

PR #810, opened 2016-04-01
ZEPPELIN-783 fix CI failure on Spark download

bzz:
### What is this PR for?
Improve CI by hard-ending spark download failures that are responsible for recent CI red on `master`.

### What type of PR is it?
Bug Fix | Hot Fix

### Todos
 - [x] cleanup on spark download attempts
 - [x] leverage Travis CI [cacheing](https://docs.travis-ci.com/user/caching) for spark and pyspark binaries under `.spark-dist`

### What is the Jira issue?
[ZEPPELIN-783](https://issues.apache.org/jira/browse/ZEPPELIN-783)

CI must be green

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ language: java
 sudo: false
 cache:
   directories:
-    - .spark
+    - .spark-dist
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ matrix:
       env: TEST_SELENIUM="true" SPARK_VER="1.6.0" HADOOP_VER="2.3" PROFILE="-Pspark-1.6 -Phadoop-2.3 -Ppyspark" BUILD_FLAG="package -DskipTests" TEST_FLAG="verify" TEST_PROJECTS="-pl zeppelin-interpreter,zeppelin-zengine,zeppelin-server,zeppelin-display,spark-dependencies,spark -Dtest=org.apache.zeppelin.AbstractFunctionalSuite -DfailIfNoTests=false"
 
 before_install:
+  - "ls -la .spark-dist"
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 
@@ -61,6 +62,7 @@ before_script:
   - travis_retry ./testing/downloadSpark.sh $SPARK_VER $HADOOP_VER
   - ./testing/startSparkCluster.sh $SPARK_VER $HADOOP_VER
   - echo "export SPARK_HOME=`pwd`/spark-$SPARK_VER-bin-hadoop$HADOOP_VER" > conf/zeppelin-env.sh
+  - tail conf/zeppelin-env.sh
 
 script:
   - mvn $TEST_FLAG $PROFILE -B $TEST_PROJECTS
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@
 language: java
 
 sudo: false
+cache:
+  directories:
+    - .spark
 
 matrix:
   include:
--- a/testing/downloadSpark.sh
+++ b/testing/downloadSpark.sh
@@ -43,7 +43,7 @@ set -xe
 FWDIR="$(dirname "${BASH_SOURCE-$0}")"
 ZEPPELIN_HOME="$(cd "${FWDIR}/.."; pwd)"
 
-SPARK_CACHE=".spark"
+SPARK_CACHE=".spark-dist"
 SPARK_ARCHIVE="spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}"
 export SPARK_HOME="${ZEPPELIN_HOME}/${SPARK_CACHE}/${SPARK_ARCHIVE}"
 echo "SPARK_HOME is ${SPARK_HOME}"
--- a/testing/downloadSpark.sh
+++ b/testing/downloadSpark.sh
@@ -52,12 +52,15 @@ if [[ ! -d "${SPARK_HOME}" ]]; then
     mkdir -p "${SPARK_CACHE}"
     cd "${SPARK_CACHE}"
     if [[ ! -f "${SPARK_ARCHIVE}.tgz" ]]; then
-        echo "Cache does not have ${SPARK_ARCHIVE} downloading ..."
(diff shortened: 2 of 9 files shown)


bzz:
Ready to be merged \cc @felixcheung @jongyoul @Leemoonsoo for a review

echarles:
That's a nice improvement which will remove some failures, but download failures also occur on simple jars (writing this down while thinking to those erratic failures).

bzz:
@echarles So true. I plan to take care of that as a next step under a separate PR, after dealing with Spark downloads here first.

testing/downloadSpark.sh
```diff
@@ -40,31 +40,53 @@ fi
...
+    if [[ ! -f "${SPARK_ARCHIVE}.tgz" ]]; then
+        pwd
+        ls -la .
+        echo "${SPARK_CACHE} does not have ${SPARK_ARCHIVE} downloading ..."
+        # download archive if not cached
+        if [[ "${SPARK_VER_RANGE}" == "<=1.2" ]]; then
+            # spark 1.1.x and spark 1.2.x can be downloaded from archive
+            STARTTIME=`date +%s`
+            #timeout -s KILL 300
+            wget -T 300 "http://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_ARCHIVE}.tgz"
```
felixcheung: The problem with `-T` (timeout) is that `-t` (retry) defaults to 20, so when it timeout, it might retry
Also, read might be going very slowly but still meeting the 300 timeout (--read-timeout defaults to 900) - we have seen frequently download is going on for 2 hours+.

So that's why it seems `timeout -s KILL 300` might be preferred, to put a hard timeout if download is going on. Also `300` or `590` (~9min) is chosen because Travis itself timeout in 10min, so we need to retry before then.

.travis.yml
```diff
@@ -16,6 +16,9 @@
 language: java
 
 sudo: false
+cache:
+  directories:
+    - .spark-dist
```
felixcheung: shouldn't we just add `.m2` to cache the local maven repo?

felixcheung: Please see https://github.com/apache/flink/blob/master/.travis.yml#L8

felixcheung:
Right, dependencies download is another frequent source of download timeout issues. Added a comment on that.

testing/downloadSpark.sh
```diff
@@ -40,31 +40,53 @@ fi
...
+            ENDTIME=`date +%s`
+            DOWNLOADTIME="$((ENDTIME-STARTTIME))"
+        else
+            # spark 1.3.x and later can be downloaded from mirror
+            # get download address from mirror
+            MIRROR_INFO=$(curl -s "http://www.apache.org/dyn/closer.cgi/spark/spark-${SPARK_VERSION}/${SPARK_ARCHIVE}.tgz?asjson=1")
```
felixcheung: `spark-${SPARK_VERSION}` --> `${SPARK_ARCHIVE}`?

felixcheung: same in above..

testing/downloadSpark.sh
```diff
@@ -40,31 +40,53 @@ fi
...
+            DOWNLOADTIME="$((ENDTIME-STARTTIME))"
+        fi
+    fi
 
-        STARTTIME=`date +%s`
-        timeout -s KILL 590 wget -q "${PREFFERED}${PATHINFO}"
-        ENDTIME=`date +%s`
-        DOWNLOADTIME=$((ENDTIME-STARTTIME))
+    # extract archive in un-cached root, clean-up on failure
+    cp "${SPARK_ARCHIVE}.tgz" ..
```
felixcheung: could we `rm "${SPARK_ARCHIVE}.tgz"` after the `cp`?

testing/downloadSpark.sh
```diff
@@ -26,10 +26,10 @@ fi
 SPARK_VERSION="${1}"
 HADOOP_VERSION="${2}"
 
-echo ${SPARK_VERSION} | grep "^1.[123].[0-9]" > /dev/null
-if [ $? -eq 0 ]; then
+echo "${SPARK_VERSION}" | grep "^1.[123].[0-9]" > /dev/null
+if [[ "$?" -eq 0 ]]; then
```
felixcheung: if `"$?"` shouldn't be put `0` in quote?

testing/downloadSpark.sh
bzz: Thank you for the explanation! Totally makes sence, will reset this to the the prev. impl.

.travis.yml
bzz: Yes, as I mentioned above - that would be the next thing to do. But I would prefer doing it in separate PR. 

The way it works (tar everything to s3 at the beginning\end of the buid) - I'm not sure it will work for our size of `.m2`, so some experiments will be needed.

And cache is alrady polluted by prev experiments so I was waiting for [INFRA-11580](https://issues.apache.org/jira/browse/INFRA-11580) to reset it, before moving to m2.

testing/downloadSpark.sh
bzz: I see what you mean. But from the [syleguide on quoting](https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting)

>Never quote literal integers.

bzz:
@felixcheung feedback addressed in 039513b

bzz:
Merging, as this is a hotfix and there is no more discussion

bzz:
[ZEPPELIN-786](https://issues.apache.org/jira/browse/ZEPPELIN-786) was created for caching `.m2`

Looks like current CI on master is failing to populate cache and download 1.1 and 1.2 from slow mirrors..

felixcheung:
yea, unfortunately it seems some of the apache mirrors are very slow..

bzz:
After at least 2 more PRs failed on downloading Spark distr again I have re-open [ZEPPELIN-783](https://issues.apache.org/jira/browse/ZEPPELIN-783) and thinking about increasing a default timeout

felixcheung:
This indicates a corrupted file (which seems fairly common) so would not seem to be download timeout related:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/120500704/log.txt

PR #812, opened 2016-04-04
[ZEPPELIN-789] Skip execution of a paragraph in case of error 

springml:
### What is this PR for?
New option in paragraphs to skip execution, when notebook is executed as a whole and there were some errors in execution of the previous paragraphs.
More details in jira ticket https://issues.apache.org/jira/browse/ZEPPELIN-789

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-789

* Build this PR
* Create a notebook and add three paragraphs in it
* Make sure the first paragraph results in error. Easy way is to have syntax error
* Add some print statement in second paragraph and set skipOnError to true for second paragraph
* Add some more print in third paragraph
* Execute the notebook. Second paragraph execution should be skipped.
* Now set skipOnError to false in second paragraph and execute the notebook. It should be executed as well.

skiponerror

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
@@ -68,6 +68,14 @@ public abstract class Job {
     public boolean isPending() {
       return this == PENDING;
     }
+
+    public boolean isError() {
+      return this == ERROR;
+    }
+
+    public boolean isFinished() {
+      return this == FINISHED;
+    }
   }
 
   private String jobName;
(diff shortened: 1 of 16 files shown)


springml:
We have some tests failing in this PR. But we are not getting any details of the failure. 

Below is the error that we see


```

The build has been terminated
```

How can we find more information to solve these test issues? Are these test failures caused by the code changes present in this PR?

felixcheung:
Does not seem like it, could you please close and reopen this PR to kick off Travis CI again?

springml:
Reopening as requested

springml:
Again same error. There might be some  issue with this PR. But how can I find it?

khalidhuseynov:
Could you reopen PR to trigger CI? The build failure doesn't seem to be related

corneadoug:
@springml If we still want to continue this PR, there might be a few things to consider.
When running the whole paragraph, there is no specific paragraph running order. Depending on the interpreter used, some will start before others. Because of that, I'm not sure if this feature would really be functional.

PR #816, opened 2016-04-05
Pluggable resource pools

fireboy1919:
### What is this PR for?
This is a step towards creating a REST endpoint from resource pools  https://issues.apache.org/jira/browse/ZEPPELIN-699

It allows resources to be saved to disk so that they can be reloaded.

### What type of PR is it?
Improvement

### Todos
* [X] - Add ability to use different kinds of resource pools, and show a second type.

### What is the Jira issue?
Partially implements this:  https://issues.apache.org/jira/browse/ZEPPELIN-699

Note, that this pull request is dependent upon this one:  
https://github.com/apache/incubator-zeppelin/pull/782

1.  Edit an interpreter
2.  Add a new property to the interpreter - the "ResourcePoolClass" - and assign it to "org.apache.zeppelin.resource.VFSResourcePool" 
3.  Run the interpreter

You should see new resource files created in notebook/zeppelin_resources

### Questions:
* Does this needs documentation?  No.


fireboy1919:
@felixcheung, @Leemoonsoo:  
This pull request, like the one it is based upon, is ready to be reviewed and merged.

Leemoonsoo:
@fireboy1919 Thanks for the contribution.

It looks like individual interpreter setting can configure ResourcePool implementation through it's property `ResourcePoolClass`. How about add some prefix on the property name, to distinguish from any other user defined properties? For example, like [zeppelin.interpreter.localRepo](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java#L150), i think `zeppelin.interpreter.resourcePoolClass` make sense. 

And how about add new configuration into [ZeppelinConfiguration](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L440) to set default value of resourcePoolClass globally through conf/zeppelin-env.sh and conf/zeppelin-site.xml ?

fireboy1919:
Great ideas!  Will add.

fireboy1919:
The implementation that I suggested has been implemented.  I did it by actually including all of the things that are in zeppelinconfig that start with zeppelin.interpreter into the properties list.

Leemoonsoo:
Somehow, the PR includes commits from other contributions. @fireboy1919 could you manage it to not includes them?

PR #818, opened 2016-04-06
ZEPPELIN-783: fix Spark download on CI

bzz:
### What is this PR for?
Fix Spark download on CI

### What type of PR is it?
Hot Fix

### What is the Jira issue?
[ZEPPELIN-783](https://issues.apache.org/jira/browse/ZEPPELIN-783)

CI must be green

### Questions:

### DIFF
--- a/testing/downloadSpark.sh
+++ b/testing/downloadSpark.sh
@@ -40,9 +40,27 @@ fi
 
 set -xe
 
-TIMEOUT_SEC=590
+MAX_DOWNLOAD_TIME_SEC=590
 FWDIR="$(dirname "${BASH_SOURCE-$0}")"
 ZEPPELIN_HOME="$(cd "${FWDIR}/.."; pwd)"
+
+#######################################
+# Downloads file from the givrn URL.
+# Ties 3 times with 1s delay, 20s read and 15s connection timeouts.
+# Globals:
+#   None
+# Arguments:
+#   url - source URL
+# Returns:
+#   None
+#######################################
+download_with_retry() {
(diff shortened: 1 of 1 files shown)


bzz:
\cc @felixcheung for review

bzz:
This is fixes 2 CI https://travis-ci.org/bzz/incubator-zeppelin and https://travis-ci.org/apache/incubator-zeppelin and is ready to be merged.

\cc @felixcheung @Leemoonsoo @jongyoul @prabhjyotsingh for a review

bzz:

lresende:
LGTM

testing/downloadSpark.sh
```diff
@@ -40,10 +40,28 @@ fi
 
 set -xe
 
-TIMEOUT_SEC=590
+MAX_DOWNLOAD_TIME_SEC=590
```
jongyoul: Trivial: This isn't used any longer.

jongyoul:
LGTM

testing/downloadSpark.sh
bzz: Good catch, but I thought about keeping it for a while, reminding about great use of `timeout` that we have discarded for now. We possibly might what get back to it one day (i.e to reduce build time), once we get the build stability.

bzz:
Thank you for reviews!

felixcheung:
Looks good - sorry for the delay.
My only thought was that we should probably echo the exit code from wget if it is not zero in download_with_retry()

PR #823, opened 2016-04-07
Generate unique temporary files for Python and R Spark Interpreters.

mudsam:
### What is this PR for?
Generate unique temporary files for Python and R Spark interpreters using java.io.File.createTempFile.
Remove the temporary file for the Python interpreter when the interpreter is closed (is already being done in the R interpreter).
Without this fix it is not possible to run more than one Zeppelin interpreter on the same server due to file owner conflicts between instances.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
No Jira opened for this

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
@@ -91,7 +91,12 @@ public class PySparkInterpreter extends Interpreter implements ExecuteResultHand
   public PySparkInterpreter(Properties property) {
     super(property);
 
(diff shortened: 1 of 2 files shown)


elbamos:
The patch shouldn't be necessary for the R interpreter from PR 208 right?

mudsam:
As far as I can tell it would not be needed for the R interpreter from PR 208.

Leemoonsoo:
Thanks for the fix. LGTM and CI failure is irrelevant to this change.

jongyoul:
LGTM

Leemoonsoo:
Merging it into master

PR #825, opened 2016-04-08
[ZEPPELIN-799] Add python edit mode.

astroshim:
### What is this PR for?
support python edit mode in paragraph.

### What type of PR is it?
Improvement

### Todos
* [x] - support python edit mode.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-799

you can refer to screenshot.

- before
image

- after 
image

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -79,8 +79,8 @@ angular.module('zeppelinWebApp')
   var angularObjectRegistry = {};
 
   var editorModes = {
-    'ace/mode/python': /^%(spark\.)?pyspark\s*$/,
-    'ace/mode/scala': /^%spark\s*$/,
+    'ace/mode/python': /^%(\w*\.)?pyspark\s*$/,
+    'ace/mode/scala': /^%(\w*\.)?spark\s*$/,
     'ace/mode/sql': /^%(\w*\.)?\wql/,
     'ace/mode/markdown': /^%md/,
     'ace/mode/sh': /^%sh/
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -80,6 +80,7 @@ angular.module('zeppelinWebApp')
 
   var editorModes = {
     'ace/mode/scala': /^%spark/,
+    'ace/mode/python': /^%pyspark/,
```
felixcheung: I think we need to account for cases like
```
%spark.spark
%spark.pyspark
%customgroup.pyspark
```

and so on.

astroshim: I changed the mode rule so Could you review?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -79,8 +79,8 @@ angular.module('zeppelinWebApp')
   var angularObjectRegistry = {};
 
   var editorModes = {
-    'ace/mode/scala': /^%spark/,
-    'ace/mode/python': /^%pyspark/,
+    'ace/mode/python': /^%(spark\.)?pyspark\s*$/,
+    'ace/mode/scala': /^%spark\s*$/,
```
felixcheung: I think it's possible to create a new interpreter group /setting so it might not be called "spark"
how about changing it like the sql one:
```
/^%(\w*\.)?pyspark/
/^%(\w*\.)?spark/
```

btw, it looks like the scala one would not work if the user has it as the default and has not put "%spark" in the paragraph

astroshim: I agree with you about creating a new interpreter group and thanks for pointing that out. 
I'll fix it.
and Regarding the running paragraph without "%spark" when it's default is no problem to work.

felixcheung:
cool thanks, looks good.

jongyoul:
LGTM

bzz:
Looks great to me! Merging if there is no other discussion.

PR #826, opened 2016-04-08
update VM readme and VM install docs to reference R interpreter adds

jeffsteinmetz:
### What is this PR for?
Updates Virtual Machine documentation to include SparkR information.

### What type of PR is it?
Documentation

### Questions:

### DIFF
--- a/docs/install/virtual_machine.md
+++ b/docs/install/virtual_machine.md
@@ -102,7 +102,7 @@ This assumes you've already cloned the project either on the host machine in the
 
 ```
 cd /incubator-zeppelin
-mvn clean package -Pspark-1.6 -Ppyspark -Phadoop-2.4 -Psparkr -DskipTests'
+mvn clean package -Pspark-1.6 -Ppyspark -Phadoop-2.4 -Psparkr -DskipTests
 ./bin/zeppelin-daemon.sh start
 ```
 
--- a/docs/install/virtual_machine.md
+++ b/docs/install/virtual_machine.md
@@ -26,7 +26,9 @@ Apache Zeppelin distribution includes a scripts directory
 
  `scripts/vagrant/zeppelin-dev`
  
-This script creates a virtual machine that launches a repeatable, known set of core dependencies required for developing Zeppelin.  It can also be used to run an existing Zeppelin build if you don't plan to build from source.  For pyspark users, this script also includes several helpful [Python Libraries](#pythonextras).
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
Looks good !

docs/install/virtual_machine.md
```diff
@@ -175,10 +178,6 @@ plt.title('How fast do you want to go today?')
 show(plt)
 ``` 
 
+### R Extras
```
felixcheung: should there be a `id:r-extras` for the anchor link above `#r-extras` ?

docs/install/virtual_machine.md
```diff
@@ -175,10 +178,6 @@ plt.title('How fast do you want to go today?')
...
 
-
-
-
-
-
-
+With zeppelin running, an R Tutorial notebook will be available.  The R packages required to run the examples and graphs in this tutorial notebook were installed by this virtual machine.
```
felixcheung: hmm, aren't we listing the packages like in Python?

docs/install/virtual_machine.md
jeffsteinmetz: The issue is how github markdown works vs the doc generator.
Github (readme) markdown uses the name of the header, in lower case, with spaces converted to dashes.
Is there an example of how the doc md -> html compiler interprets anchors?  I looked around at all of the other zeppelin docs, and couldn't find any doc examples that used anchors to headers within the "current" document.  I can update for the docs once I confirm how the doc generator works.  Does it use `id`?  Any examples?

docs/install/virtual_machine.md
jeffsteinmetz: yah, we could do that for sure.
The package names for Python are listed on line 96 (earlier), not toward the end where you commented.
We can include in the same location if that makes sense.

docs/install/virtual_machine.md
felixcheung: ah, got it. hmm, I'm not sure - perhaps someone on https://jekyllrb.com/docs/permalinks/

jeffsteinmetz: Which pre-processor do we use for markdown (prior to Jekyll)?
I could dig around, but figured you might be able to point it out quickly.

felixcheung:
LGTM

docs/install/virtual_machine.md
felixcheung: It's just Jekyll, i guess - you can find the extensions here: https://github.com/apache/incubator-zeppelin/blob/master/docs/_config.yml

PR #830, opened 2016-04-12
[MINOR] Set log4j.configuration into JAVA_OPTS and JAVA_INTP_OPTS explicitly

jongyoul:
### What is this PR for?
Set log4j into JVM option for enforcing logging configuration.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Set "-Dlog4j.configuration=..." into JVM option

### What is the Jira issue?
Minor issue

1. Run spark with default log4j setting and check logs/zeppelin*.out. you can see logs from Spark.
1. Apply this patch.
1. Run spark with default log4j setting and check logs/zeppelin-interpreter*.log. you can see logs from Spark.

### Questions:

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -105,6 +105,7 @@ if [[ -z "$ZEPPELIN_MEM" ]]; then
 fi
 
 JAVA_OPTS+=" ${ZEPPELIN_JAVA_OPTS} -Dfile.encoding=${ZEPPELIN_ENCODING} ${ZEPPELIN_MEM}"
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
LGTM

jongyoul:

aarcro:
I had to remove this to get my %sql interpreter running again (on EMR). From what I can tell on the yarn resource manager, something owned by the hadoop user was trying to write to zeppelin's log file. I assume zeppelin's log4j config, shouldn't get used by the spark proc behind it, right?

PR #833, opened 2016-04-13
[ZEPPELIN-705]Search should  aware notebook permission

ravicodder:
### What is this PR for?
Make search aware of notebook permissions and allow only those search results for which user has read permission. 

### What type of PR is it?
Bug Fix

### Todos
NA

### What is the Jira issue?
[ZEPPELIN-705]( https://issues.apache.org/jira/browse/ZEPPELIN-705?jql=project%20%3D%20ZEPPELIN%20AND%20status%20in%20(Open%2C%20Resolved)%20AND%20resolution%20%3D%20Unresolved%20AND%20fixVersion%20%3D%200.6.0%20ORDER%20BY%20due%20ASC%2C%20priority%20DESC%2C%20created%20ASC)

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -644,7 +644,7 @@ public class NotebookRestApi {
   }  
 
   /**
-   * Search for a Notes
+   * Search for a Notes with permissions
    */
   @GET
   @Path("search")
@@ -656,8 +656,7 @@ public class NotebookRestApi {
     userAndRoles.add(principal);
     userAndRoles.addAll(roles);
     List<Map<String, String>> notebooksFound = notebookIndex.query(queryTerm);
-    for (int i = 0; i < notebooksFound.size(); i++)
-    {
+    for (int i = 0; i < notebooksFound.size(); i++) {
       String[] Id = notebooksFound.get(i).get("id").split("/", 2);
       String noteId = Id[0];
       if (!notebookAuthorization.isOwner(noteId, userAndRoles) &&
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
@ravicodder 
Thanks for the patch. Would it be difficult to make corresponding rest api test?

prabhjyotsingh:
LGTM, +1 for adding test.

ravicodder:
@Leemoonsoo @prabhjyotsingh  Added test . Please have a look.

Leemoonsoo:
@ravicodder No worries. Thanks for the fix and tests. Looks good to me.

bzz:
@ravicodder thank you for fixing it.

Could you also please make sure that coded formatting is consistent [project's style guide](https://github.com/apache/incubator-zeppelin/blob/master/CONTRIBUTING.md#code-convention)?

ravicodder:
@bzz  Thanks for the review,
 I have tried to make test more readable. Please have a look.

Leemoonsoo:
LGTM

bzz:
Looks great, thank you for taking care.

I have one more question: in case of no auth configured, can somebody please help me to understand what is the performance implication or price per-API call to `/serach/`, that is added by this changeset? 

We want search to be as fast as possible in a common case, so just want to make sure that we keep an eye on it. Would that be reasonable to add [something like this](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/search/LuceneSearch.java#L285) to the query API code and at least manually compare results once (post here), to make sure preformance does not degraded, or see if any optimization are needed for case when no auth is configuted?

Please let me know what do you think!

ravicodder:
@bzz 
At a time we have set the search results to show at most 20 results. I feel, for such a small value there should not be any performance issues.

screen shot 2016-05-02 at 5 35 31 pm

screen shot 2016-05-02 at 5 40 59 pm

I noted the time to search in both the cases, with permission check and without permission check.
From manual test, We can see there is no performance degradation.

bzz:
@ravicodder looks great, thank you for double-checking!

PR #845, opened 2016-04-20
[ZEPPELIN-817] Fix issue for queries containing a 'fields' parameter

bbonnin:
### What is this PR for?
It is a bug fix for search queries that use a fields parameter (cf. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-fields.html).
An error was displayed.

### What type of PR is it?
Bug Fix

### Todos
* [X ] - ElasticsearchInterpreter : check if the result is in the '_source' part or in the 'fields' part of the response
* [X] - Add a unit test for this case

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-817

Use the elasticsearch interpreter and type a search query such as : search / {"fields":["date", "request.headers"],"query":{"match": {"status":404}

Before the fix, there was an error : "Error : string is null"

### Questions:


felixcheung:
looks good, thanks for the fix. Is this something that can be added to the interpreter doc, an example on how to query fields?

Leemoonsoo:
Looks good. Could you also update json-flattener version in [bin_license/LICENSE](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-distribution/src/bin_license/LICENSE#L76) file?

bbonnin:
Doc and LICENCE updated in the last commits.

Leemoonsoo:
Thanks for the contribution. LGTM

felixcheung:
merging if no more comment

PR #847, opened 2016-04-21
Remove files under /notebook/ 

cyhsutw:
### What is this PR for?

We ignore files under `/notebook/` but there are some existing files under that directory.

I execute the command to remove all existing files under `/notebook/` so the ignoring function would work properly.

```sh
$ git ls-files -i --exclude-from=.gitignore | xargs git rm --cached 
```

### What type of PR is it?
Improvement

### Todos
None

### What is the Jira issue?
N/A

Check the creating or modifying files under `/notebook/` will not change the output of `git status`.

N/A

### Questions:
* Does the licenses files need update? *no*
* Is there breaking changes for older versions? *no*
* Does this needs documentation? *no*


bzz:
Zeppelin by default includes `Welcome to Zeppelin` tutorial notebook, which is expected to be a part of the source tree\release as it is very convenient for a new users to have a reference point to start with.

Please make sure that this PR does not remove it.

cyhsutw:
Thanks for the information, I'll check it later today.

cyhsutw:
Just checked and it *does* delete the template tutorial. Sorry for the misunderstanding.

I'm closing this now, but I think maybe we can create the template notes under other directory and copy the templates into `/notes/`.

PR #854, opened 2016-04-25
ZEPPELIN-824 - Make CI Crash when there is jshint errors

corneadoug:
### What is this PR for?
This PR has for goal to fix the CI build regarding the front-end.
The CI is currently not failing due to jshint, even though it fails locally with a `grunt build`
There was two issues making this behaviour:
* Jshint was called only on newer files in build
* Mvn is using --force on grunt

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-824

Make a jshint error (!= instead of !== for example)
then run `mvn package` inside the zeppelin-web forlder

### Questions:

### DIFF
--- a/zeppelin-web/Gruntfile.js
+++ b/zeppelin-web/Gruntfile.js
@@ -434,7 +434,7 @@ module.exports = function (grunt) {
   ]);
 
   grunt.registerTask('build', [
-    'newer:jshint',
+    'jshint:all',
     'clean:dist',
     'wiredep',
     'useminPrepare',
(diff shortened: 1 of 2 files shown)


corneadoug:
Triggering the CI again, previous fail seems unrelated

felixcheung:
LGTM

prabhjyotsingh:
LGTM. Big +1 for removing `--force`

corneadoug:

PR #864, opened 2016-04-28
[ZEPPELIN-779] Dynamic form doesn't work with python3

Leemoonsoo:
### What is this PR for?

When pyspark interpreter configured to use python3, dynamic form doesn't work.
https://issues.apache.org/jira/browse/ZEPPELIN-779

### What type of PR is it?
Bug Fix

### Todos
* [x] - Convert map to list

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-779

configure pyspark interpreter to use python3 and run
```
%pyspark
print("Hello "+z.select("day", [("1","mon"),
                                ("2","tue"),
                                ("3","wed"),
                                ("4","thurs"),
                                ("5","fri"),
                                ("6","sat"),
                                ("7","sun")]))
```

### Questions:

### DIFF
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -83,14 +83,14 @@ class PyZeppelinContext(dict):
 
   def select(self, name, options, defaultValue = ""):
     # auto_convert to ArrayList doesn't match the method signature on JVM side
-    tuples = map(lambda items: self.__tupleToScalaTuple2(items), options)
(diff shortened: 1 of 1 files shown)


felixcheung:
I didn't know in python 3, map(list()) returns a map!
http://stackoverflow.com/questions/1303347/getting-a-map-to-return-a-list-in-python-3-x

LGTM.

Leemoonsoo:
I'm merging it into master if there're no more discussions.

PR #863, opened 2016-04-28
Prevent duplicated response header insertion

Leemoonsoo:
### What is this PR for?
Headers are inserted twice in http response message after https://github.com/apache/incubator-zeppelin/pull/831.

it prevents zeppelin-web development mode working (cd zeppelin-web; grunt serve)

image

This PR provides (dirty) fix for the problem. If anyone knows better way to prevent duplicated header insertion, please let me know.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Prevent duplicated header insertion

### How should this be tested?
Check if dev mode works. ie. 'cd zeppelin-web; grunt serve', browse http://localhost:9000

or

run `curl -vv http://localhost:8080/api/security/ticket` and see if there're duplicated headers

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #831, opened 2016-04-12
[ZEPPELIN-798] Migrate to Jetty version 9

echarles:
### What is this PR for?
Upgrade from Jetty 8 to Jetty 9 mainly for deadlock bug fix https://bugs.eclipse.org/bugs/show_bug.cgi?id=389645

Jetty 9 has changed the maven names and context building (for web and websocket).

### What type of PR is it?
[Bug Fix]

### Todos
* [x ] - Adapt the unit tests

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-798

### How should this be tested?
Outline the steps to test the PR here.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #867, opened 2016-04-29
Fix duplicate CORS header problem

khalidhuseynov:
### What is this PR for?
It attempts to solve cause of the duplicated CORS headers problem that was introduced by #831 and was tried to be solved as hotfix in #863 

### What type of PR is it?
Bug Fix

### Todos
* [x] - remove second `addFilter` to the same context

same as in #863 

screenshot_duplicate_cors2

### Questions:


Leemoonsoo:
It looks like better solution than #863 👍

echarles:
Yes, better tha #863.
To make it clear that the filter is applied to both the `api` and `ws`, I would move `webapp.addFilter(new FilterHolder...` somewhere at line 273.

Leemoonsoo:
I'm merging it into master as a hotfix, while it breaks development mode of zeppelin-web.
@echarles 's suggestion is very valid and it can be applied in the future PR.

PR #883, opened 2016-05-11
Fix notebook folder structure name match bug

minahlee:
### What is this PR for?
Current regex match rule for notebook folder takes not only `/` as delimiter but also `[` and `]`.

### What type of PR is it?
Bug Fix

Try to create notebook that contains `[` or `]` in name and see if the folder structure is correct.

With notebook name `/A/B[C]D`
**Before**
screen shot 2016-05-11 at 3 00 42 pm

**After**
screen shot 2016-05-11 at 3 01 31 pm

### Questions:


Leemoonsoo:
Thanks for the fix. How about add a test case to https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-web/test/spec/factory/notebookList.js#L15 ?

minahlee:
@Leemoonsoo Just added test. Please review

Leemoonsoo:
LGTM

Leemoonsoo:
Merge it into master if there're no more discussions

PR #888, opened 2016-05-12
ZEPPELIN-850 update error message

prabhjyotsingh:
### What is this PR for?
If there is an error on a paragraph, and user tries to run that paragraph again, which results in a different error message. In this case paragraph does not show up new error message.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-850](https://issues.apache.org/jira/browse/ZEPPELIN-850)

 - create a new notebook
 - on the try running paragraph by using a wrong value for interpreter like %spark.error
 - on try changing the same with say %spark.error2
observe the paragraph output it should changed.

Before:
error

After:
fixed

### Questions:


prabhjyotsingh:
CI; Job #4165.6 fails for `'npm install --color=false'` which is unrelated.

prabhjyotsingh:
Ready for review.

Leemoonsoo:
LGTM

prabhjyotsingh:
CI fails at `Job #4165.6` for `npm install --color=false` which is unrelated.

PR #769, opened 2016-03-09
Fix interpreter.sh to get Spark interpreter log file

AhyoungRyu:
### What is this PR for?
Currently, if users set their own `SPARK_HOME`, they can not get `zeppelin-interpreter-spark-xxxx.log` file. This PR is for fixing this issue.
(This issue is reported by @weipuz)

### What type of PR is it?
Hot Fix

### What is the Jira issue?
None 

### How should this be tested?
After applying this PR, 
1. Set your own `SPARK_HOME`.
2. Run `sc.version`(or whatever you want) with Spark interpreter.
3. Check under your `ZEPPELIN_HOME/logs/` directory, then you can find `zeppelin-interpreter-spark-xxx.log` file.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #889, opened 2016-05-12
Fix interpreter.sh classpath

Leemoonsoo:
### What is this PR for?
This PR apply fix #769 again, which is reverted by #208.
Also removing unnecessary code from interpreter.sh

### What type of PR is it?
Bug Fix

### Todos
* [x] - Apply #769 again
* [x] - Remove unnecessary code

### Questions:


felixcheung:
LGTM

elbamos:
It takes out a change I'd out there because the r interpreter jar wasn't getting loaded in time in some circumstances. I need to test this. I will try to do that in the next few days.

Leemoonsoo:
Every jar under interpreter/spark/ supposed to be included in the classpath by [interpreter.sh#67](https://github.com/apache/incubator-zeppelin/blob/master/bin/interpreter.sh#L67). So i expect this PR does not change any behavior. @elbamos if you find any case that jar is not loaded in this way, please file an issue.

PR #899, opened 2016-05-18
Fix a few image not rendering

corneadoug:
### What is this PR for?
There was a recent PR made by @doanduyhai that was changing some images URL to include `../assets`. However, I found a few more cases in the deployed website.
It is actually a bit hard to understand why this happens in some cases only, maybe it could be better to replace all `/assets` to `../assets`, however this PR is only fixing the current issues.

### What type of PR is it?
Bug Fix

You can build the doc and run it locally

screen shot 2016-05-18 at 3 20 53 pm
screen shot 2016-05-18 at 3 21 09 pm

### Questions:


minahlee:
LGTM

PR #900, opened 2016-05-19
Refactor CORS filter into webapp context handler

khalidhuseynov:
### What is this PR for?
#867 was a hotfix and merged immediately. However we can refactor CORS filter into more appropriate place, which is webapp context handler instead of keeping it in rest api handler (since it also applies to websocket api).

### What type of PR is it?
Refactoring

### Todos
* [x] - move from rest to webapp

Basically this is refactoring, and original flow holds. 
For testing, can use same routine as in #867.

### Questions:


bzz:
Looks good to me. Merging if there is no further discussion.

PR #911, opened 2016-05-23
ZEPPELIN-866] Hotfix  -  does not working search box on navbar.

cloverhearts:
### What is this PR for?
When you type in the search box and start the search,
The result is not exposed.
Previously, it was working well.

### What type of PR is it?
Hot Fix

### Todos
* [x] - fixed bug on navbar.html and controller

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-866
try to search action on navbar.

#### before 
test02

#### after
test01

### Questions:

### DIFF
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -16,6 +16,7 @@
 
 angular.module('zeppelinWebApp').controller('NavCtrl', function($scope, $rootScope, $http, $routeParams,
     $location, notebookListDataFactory, baseUrlSrv, websocketMsgSrv, arrayOrderingSrv) {
+
   /** Current list of notes (ids) */
 
   $scope.showLoginWindow = function() {
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -92,7 +92,6 @@ angular.module('zeppelinWebApp').controller('NavCtrl', function($scope, $rootSco
   };
 
   $scope.search = function(searchTerm) {
-    console.log('search term ', searchTerm);
     $location.url(/search/ + searchTerm);
   };
 
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -46,15 +46,6 @@ angular.module('zeppelinWebApp').controller('NavCtrl', function($scope, $rootSco
     vm.connected = param;
   });
 
-  $rootScope.$on('$locationChangeSuccess', function () {
-    var path = $location.path();
-    // hacky solution to clear search bar
-    // TODO(felizbear): figure out how to make ng-click work in navbar
-    if (path === '/') {
-      $scope.searchTerm = '';
-    }
-  });
-
   $scope.checkUsername = function () {
     if ($rootScope.ticket) {
       if ($rootScope.ticket.principal.length <= MAX_USERNAME_LENGTH) {
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -91,8 +91,9 @@ angular.module('zeppelinWebApp').controller('NavCtrl', function($scope, $rootSco
       });
   };
 
-  $scope.search = function() {
-    $location.url(/search/ + $scope.searchTerm);
+  $scope.search = function(searchTerm) {
+    console.log('search term ', searchTerm);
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/components/navbar/navbar.controller.js
```diff
@@ -91,8 +91,9 @@ angular.module('zeppelinWebApp').controller('NavCtrl', function($scope, $rootSco
       });
   };
 
-  $scope.search = function() {
-    $location.url(/search/ + $scope.searchTerm);
+  $scope.search = function(searchTerm) {
+    console.log('search term ', searchTerm);
```
prabhjyotsingh: Can you remove console.log from here.

cloverhearts: @prabhjyotsingh oh, Thank you!

prabhjyotsingh:
If we use `$scope.search = function(searchTerm)` then we can [this](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-web/src/components/navbar/navbar.controller.js#L49-56) function becomes redundant. Can you remove this as well.



    $rootScope.$on('$locationChangeSuccess', function () {
     var path = $location.path();
     // hacky solution to clear search bar
     // TODO(felizbear): figure out how to make ng-click work in navbar
     if (path === '/') {
       $scope.searchTerm = '';
     }
    });

cloverhearts:
@prabhjyotsingh 
I think that code would clean up the search box.
It should not be deleted.

prabhjyotsingh:
Or that is also broken for `$locationChangeSuccess`, its not cleaning up. Can you fix this as well ?

cloverhearts:
@prabhjyotsingh Okay, i will look to this issue.

cloverhearts:
@prabhjyotsingh  You're all right.
you're alright.
this is code is bug and dead code. If active to code, the search becomes impossible.
So I have deleted that code.

cloverhearts:
@prabhjyotsingh 
fix done.
Please, Can you feedback for this pr for me?

bzz:
Thanks for hotfix, but do we know why it stopped working in the first place?
Let me take a look.

\cc @felizbear for review as well

astroshim:
I tesed and working well. but it seems need a "initial statement".

zeppelin-web/src/components/navbar/navbar.html
```diff
@@ -65,7 +65,7 @@
         <!--TODO(bzz): move to Typeahead https://angular-ui.github.io/bootstrap  -->
           <form role="search"
             style="width: 300px; display: inline-block; margin: 0 10px"
-             ng-submit="search()">
+             ng-submit="search(searchTerm)">
```
felizbear: i definitely like passing the argument to `search` instead of taking it from the `$scope` 👍

cloverhearts:
@astroshim Thank you for feed back.
are you mean 'initial statement' is 'clean up on search box'?

AhyoungRyu:
@cloverhearts I agree with @astroshim. I think he mean when the result is nothing, Zeppelin needs to show some statement like "We couldn’t find any notebook matching 'something_you_typed' "

astroshim:
Thanks for the detail explanation AhyoungRyu.

cloverhearts:
Thank you for explaining, @astroshim @AhyoungRyu 
I agree too.
But, This is must exist only for bug fixes at this PR.
I will make it in the next PR.

AhyoungRyu:
@cloverhearts Yeah you're right. LGTM 👍

prabhjyotsingh:
I think this sounds good, have created a jira for the same https://issues.apache.org/jira/browse/ZEPPELIN-869. LGTM. 
I'll Merge this if no more discussion.

bzz:
> Let me take a look.

@prabhjyotsingh could you please wait, as there was an issue, but not addressed yet?

> Thanks for hotfix, but do we know why it stopped working in the first place?
It would be really cool, if we could know

prabhjyotsingh:
Surething.

cloverhearts:
@bzz 
can not access to '$ scope.searchTerm variable at navbar.html' in navbar.controller
Code and nav are not currently operating normally.
I did not find the PR of the problem, which seems to be a problem at some point.
This can cause a lot of problems in the current navbar.
For example
https://github.com/apache/incubator-zeppelin/pull/912

I think we need to recover the ability to hotfix first.
And it should retouching the navbar.

bzz:
@cloverhearts fix looks great to me. Pity that we can not identify the source of the bug though.

Let's merge as a hotfix!

prabhjyotsingh:
Sure, I'll merge this.

PR #919, opened 2016-05-26
support changing tabsize of paragraph

astroshim:
### What is this PR for?
The code editor tab size is fixed as 4.
It would be better to support changing tabsize to users.

### What type of PR is it?
Improvement

### Todos
* [ ] - Suppport REST API.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-887

1. change the tab size in the righ-menu on paragraph.
2. type tab on paragraph.

image

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -76,6 +76,16 @@ limitations under the License.
            ng-show="!paragraph.config.lineNumbers"><span class="fa fa-list-ol"></span> Show line numbers</a>
       </li>
       <li>
+        <a ng-click="$event.stopPropagation()" class="dropdown"><span class="fa fa-ellipsis-h"></span> TabSize
+          <form style="display:inline; margin-left:5px;">
+            <select ng-model="paragraph.config.tabSize"
+                    class="selectpicker"
+                    ng-change="changeTabSize()"
+                    ng-options="col for col in tabSizeOption"></select>
+          </form>
+        </a>
+      </li>
+      <li>
         <a ng-click="toggleEnableDisable()"><span class="icon-control-play"></span>
           {{paragraph.config.enabled ? "Disable" : "Enable"}} run</a>
       </li>
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -93,6 +93,7 @@ angular.module('zeppelinWebApp')
     $scope.originalText = angular.copy(newParagraph.text);
     $scope.chart = {};
     $scope.colWidthOption = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 ];
+    $scope.tabSizeOption = [ 1, 2, 3, 4, 5, 6, 7, 8 ];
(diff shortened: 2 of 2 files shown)


Leemoonsoo:
Thanks for addressing the issue. I think it'll be more useful if tab size can be configured globally or per interpreter, rater than individual paragraph. What do you think?

astroshim:
@Leemoonsoo Good point. I agree with you that changing tab size in the interpreter or notebook is better than paragraph. so I think Configuring tab size of notebook level makes more sense than interpreter level, what do you think?

Leemoonsoo:
I think tab size is more like per interpreter rather than notebook or paragraph.


Because 
appropriate tabsize might different from each interpreter. e.g. 2 for scala, 4 for sparksql and they can be used in the same notebook.

Does it make sense?

astroshim:
@Leemoonsoo It really makes sense and I thought that too. 
The reason why I thought about configuring tab size in notebook is better was just because easier for the users to set up.
but your example might more make sense.
Thanks.

corneadoug:
@astroshim Would you then create a new PR, or modify this one?

astroshim:
@corneadoug Thank you for your concern! I'm thinking to find right way how to handle this issue.

bzz:
To be able to change indention is very reasonable option, esp. as soon as you start working with Python.

@astroshim It could be hard to implement per-interpreter as @Leemoonsoo suggests for now, so how about changing this implementat to per-notebook settings? I think it would be useful.


My only concern is whether it actually changes the paragraph code, or not.

I have checked, and Ace editor we use in soft-tab configuration (although I can not find `useSoftTabs ` anywhere), but the user-input `\t` gets replaced by 4 spaces now.

Then, how does [changing this setting](https://ace.c9.io/api/edit_session.html#EditSession.setTabSize) affect the paragraph (that already has no `\t`)? Will it re-format whole paragraph by adding more whitespace chars (affects the diff)?

astroshim:
@bzz Thank you for your detail review! 
I was worring about changing tab size per interpreter as @Leemoonsoo suggests is more hard to config than per notebook or per paragraph.
How about implementing per notebook first and then let's talk it's good or bad?
and This changing is not affect to paragraphs of existing.

Leemoonsoo:
Like comment https://github.com/apache/zeppelin/pull/966#issuecomment-224127925 from @felixcheung, it'll be great if interpreter-setting.json can configure code-editor.

For example, add "editor" field, 
```
[
  {
    "group": "jdbc",
    "name": "sql",
    "className": "org.apache.zeppelin.jdbc.JDBCInterpreter",
    "properties": {
      ...
    },
    "editor" : {
      "tabSize" : 4,
      "syntaxHighlighting" : "sql",
      "autoCompletionOnTab" : true,
      "tabToWhitespace" : true,
      ...
    },
```

I think this approach can be used not only for 'tabsize', but also for many other configuration values as example above.

Implementing notebook level first would be an option, but if it is not very urgent issue, then it's worth to consider configure through interpreter-setting.json instead. Then many other features (syntaxHighlighting, autoCompletionOnTab, etc) can easily benefit by it.

astroshim:
Thank you for your detailed explanation @Leemoonsoo and I agree with moving interpreter-setting.json.

astroshim:
https://github.com/apache/zeppelin/pull/1218 is better way.

PR #930, opened 2016-05-30
Make path for .spark-dist folder always accessible [ZEPPELIN-902]

echarles:
### What is this PR for?
Current Path for .spark-dist is outside the reactor, which can breaks build.
[INFO] BUILD FAILURE
[INFO] Total time: 30.337 s
[INFO] Finished at: 2016-05-30T15:34:30+02:00
[INFO] Final Memory: 73M/469M
[ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.2.1:wget (download-pyspark-files) on project zeppelin-spark-dependencies: IO Error: Destination '/opt/zeppelin-master/spark-dependencies/target/../../.spark-dist/spark-1.6.1.tgz' directory cannot be created -> [Help 1]

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN/ZEPPELIN-902

Build with pyspark profile

### Questions:


bzz:
Hi, thanks for taking care of the build stability!

`.spark-dist` was introduced as our CI build cache dir, and is used in [multiple places](https://github.com/apache/incubator-zeppelin/search?utf8=%E2%9C%93&q=.spark-dist)

As far as I understand, you suggest using different dirs for bin and src downloads, and src dir will not be part of Travis CI cache any more, so it will be actually dowloaded from Apache mirror on every build again, which is from our previous experience make build very unstable and error-prone.

I do not think it's possible to cache things in the paths, that are going to be overridden by CI's `git clone ...`. Do you have suggestions on how to make it cache friendly and do not break the build?

> .spark-dist is outside the reactor, which can breaks build
Also, could you please help me to understand why the fact of being outside of reactor can break the build?

echarles:
@bzz Just commented on ZEPPELIN-902 - My concern was on my local build, not on Travis stability.
As mentionned onthe JIRA, the target `target` folder is not present after a clean, because the download-pyspark-files is binded to the `validate` phase.

I guess using ${basedir} could do the job, but then we are not 100% safe because building in the spark-dependencies folder would put the spark-dist at a different location... 

Maybe ${session.executionRootDirectory} ? (see http://stackoverflow.com/questions/3084629/finding-the-root-directory-of-a-multi-module-maven-reactor-project).

Another approach would be to bind to the `compile` phase, but I guess it will have side-effects.

bzz:
Thanks for explaining! 

So as far as I have understood, the pain point is: after `mvn clean`, on the same repo `mvn validate` faild due to  `maven-download-plugin` exception
```
iO Error: Destination '/opt/zeppelin-master/spark-dependencies/target/../../.spark-dist/spark-1.6.1.tgz' directory cannot be created 
```
who presumably complaining about non-existant `/target/` part of it's destination path?

and what you suggest is that we put there some different path, that is not target-related, and what I suggest is that we better make it still pointing to the same `.spark-dist` dir, one level above the repo root.

`${session.executionRootDirectory}/../.spark-dist` sounds as a good solution to me. 

Or another, harder way could be: if you could provide the stack trace of the same error (running `mvn -X ...`) we could look into the implementation of [maven-download-plugin](https://github.com/maven-download-plugin/maven-download-plugin) and see if we could improve it to behave like `mkdir -p` in this case.

echarles:
Tried with the ${session.executionRootDirectory} and got  [ERROR] around Ant part ...<untar src="${session.executionRootDirectory}/.spark-dist/spark-1.6.1.tgz" 

Sounds like maven-antrun-plugin (even last version) is not happy with  ${session.executionRootDirectory}

Hacking maven-antrun-plugin or maven-download-plugin is not on my list. I guess we can simply live with this for now.

jongyoul:
ping

echarles:
pong :) I don't see any solution for this. Does this issue bother anyone else?

jongyoul:
No. I just wonder if this is not merged even if it passes CI for a long time ago.

corneadoug:
@echarles @bzz So is this Ready for merge?

echarles:
No, not ready. None of the envisaged solutions are practicable.

echarles:
Closing this PR, I can not reproduce the issue anymore.

PR #933, opened 2016-05-31
[ZEPPELIN-905] fix failed notebook import bug

swkimme:
### What is this PR for?
When notebook has a result with Exceptions, it always fail to import. This PR fix it.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix this bug by ignore the result with error.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-905

Create notebook with error, then try to import. 

### Questions:
* Does the licenses files need update? X
* Is there breaking changes for older versions? X
* Does this needs documentation? X

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -20,7 +20,6 @@ package org.apache.zeppelin.notebook;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.*;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -38,7 +37,6 @@ import org.apache.zeppelin.scheduler.Job.Status;
 import org.apache.zeppelin.scheduler.JobListener;
 import org.apache.zeppelin.search.SearchService;
 
-import com.google.gson.Gson;
 import org.apache.zeppelin.user.AuthenticationInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -183,11 +181,8 @@ public class Note implements Serializable, JobListener {
     newParagraph.setTitle(srcParagraph.getTitle());
 
     try {
-      Gson gson = new Gson();
-      String resultJson = gson.toJson(srcParagraph.getReturn());
-      InterpreterResult result = gson.fromJson(resultJson, InterpreterResult.class);
-      newParagraph.setReturn(result, null);
-    } catch (Exception e) { /*ignore*/ }
+      newParagraph.setReturn((InterpreterResult) srcParagraph.getReturn(), null);
+    } catch (ClassCastException e) { /*ignore*/ }
 
     synchronized (paragraphs) {
       paragraphs.add(newParagraph);
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -189,7 +189,7 @@ public class Note implements Serializable, JobListener {
       newParagraph.setReturn(result, null);
     } catch (Exception e) {
       // 'result' part of Note consists of exception, instead of actual interpreter results
-      logger.info("Paragraph " + srcParagraph.getId() + " has a result with exception.");
+      logger.warn("Paragraph " + srcParagraph.getId() + " has a result with exception. " + e.getMessage());
     }
 
     synchronized (paragraphs) {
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
Looks good to me.
How about adding an unittest for it in [NotebookTest.java](https://github.com/apache/incubator-zeppelin/blob/18fa33a9b1127efd7b05980427fc61871b41f593/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java#L312)? Let me know if you need help!

swkimme:
@Leemoonsoo  
I've added some test!

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -309,6 +309,28 @@ public void testAutoRestartInterpreterAfterSchedule() throws InterruptedExceptio
...
 
   @Test
+  public void testExportAndImportNote() throws IOException, CloneNotSupportedException,
+          InterruptedException {
+    Note note = notebook.createNote();
+    note.getNoteReplLoader().setInterpreters(factory.getDefaultInterpreterSettingList());
+
+    final Paragraph p = note.addParagraph();
+    String simpleText = "hello world";
+    p.setText(simpleText);
```
Leemoonsoo: How about make `p.getReturn()` returns non InterpreterResult type object here? Then this test can simulate exception when doing `InterpreterResult result = gson.fromJson(resultJson, InterpreterResult.class);`

I think easiest way to set arbitrary object to [Job.result](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java#L75) is making [Job.setResult()](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java#L258) public.

swkimme: @Leemoonsoo I cannot agree that's the best way to do that.. 
What if we change type of Job.result into InterpreterResult ?

swkimme:
@Leemoonsoo  Do you know why the CI test failed btw?

prabhjyotsingh:
@swkimme I think this is failing because you are trying to get paragraph result i.e. `p.getResult()` or `p2.getResult()` without running any of the paragraph.

try having 

    note.runAll();
    while(p.isTerminated()==false || p.getResult()==null) Thread.yield();

    importedNote.runAll();
    while(p2.isTerminated()==false || p2.getResult()==null) Thread.yield();

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -175,17 +175,19 @@ public void addCloneParagraph(Paragraph srcParagraph) {
...
     newParagraph.setText(srcParagraph.getText());
     newParagraph.setTitle(srcParagraph.getTitle());
-    newParagraph.setReturn(result, null);
+
+    try {
+      Gson gson = new Gson();
+      String resultJson = gson.toJson(srcParagraph.getReturn());
+      InterpreterResult result = gson.fromJson(resultJson, InterpreterResult.class);
+      newParagraph.setReturn(result, null);
+    } catch (Exception e) { /* ignore */ }
```
bzz: Let's leave some clues for the person who is going to read it next time, shall we?

I.e `// ignore the case when result part of Note consists of exception, instead of actual interpreter result`

felixcheung: and, I think we should log the exception

swkimme:
@bzz @felixcheung Good idea, updated log and comment!

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -187,7 +187,10 @@ public void addCloneParagraph(Paragraph srcParagraph) {
       String resultJson = gson.toJson(srcParagraph.getReturn());
       InterpreterResult result = gson.fromJson(resultJson, InterpreterResult.class);
       newParagraph.setReturn(result, null);
-    } catch (Exception e) { /* ignore */ }
+    } catch (Exception e) {
+      // 'result' part of Note consists of exception, instead of actual interpreter results
+      logger.info("Paragraph " + srcParagraph.getId() + " has a result with exception.");
```
felixcheung: could you ether logger.error() or getStackTrace like [here](https://github.com/apache/incubator-zeppelin/blob/391ceab074c068d9deaeea80e357e739fd193fac/zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java#L114)?

swkimme: The exception and message is quite clear so I don't think we do not need trace, then I'm adding the error message!

bzz:
@swakrish do you think it's ready to be merged?

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
Leemoonsoo: @swkimme 
I was trying to say, the test doesn't test import note when p.getReturn() is `String`. 
However this unittest is still good, too. Anyway we didn't have test for import/export function.

About changing Job.result into InterpreterResult, we'll need to consider import note.json created from previous version of Zeppelin. They'll still have Job.result with string type.

bzz:
@swkimme ping on the status of this patch

PR #938, opened 2016-05-31
ZEPPELIN-930 ] fix width and not close li tag in searchbar

cloverhearts:
### What is this PR for?

closed li tag.
and Add a tooltip, and reduced the size of the horizontal search bar.

### What type of PR is it?
Bug Fix and Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-930

It changes the browser horizontal size.
Use the search feature.

#### before
searchbarori

#### after
searchbar

### Questions:


zeppelin-web/src/components/navbar/navbar.html
```diff
@@ -62,15 +62,17 @@
...
              ng-submit="search(searchTerm)">
             <div class="input-group">
               <input
                 type="text"
                 ng-model="searchTerm"
                 ng-disabled="!navbar.connected"
                 class="form-control"
-                placeholder="Search in your notebooks"
+                tooltip-placement="bottom"
+                tooltip="Search for your Notebooks"
```
minahlee: Is there any special reason for this change? I personally prefer to have placeholder than tooltip to keep the consistency with filter input box

AhyoungRyu:
@cloverhearts Good catch to close `<li>` tag 👍  Is there any particular reason that you changed `placeholder` to `tooltip` ?

zeppelin-web/src/components/navbar/navbar.html
AhyoungRyu: I agree with @minhlee

cloverhearts: @minahlee @AhyoungRyu Thank you for you are feedback.
Reasons are as follows. Reducing the width of the search box will show all the text in the placeholder. Since you already have the search icon, you can not misunderstand the function of the search box. For detailed function descriptions are replaced by the tooltip.

bzz: Huge +1 for finding not closed `<li>`.

But I also would prefer having a placeholder text, as it makes it immediately clear to the users the purpose of this UI element.

Could we revert the placeholder as it was and keep only `<li>` fix in scope of this PR?

bzz:
Looks great to me. 
Let's merge after reverting placeholder text back.

cloverhearts:
@bzz Thank you for you're review.
i will reverting placeholder text.
@minahlee @AhyoungRyu

cloverhearts:
Modification was completed.
Please review for this PR.

AhyoungRyu:
@cloverhearts Thanks for addressing that! LGTM 👍

astroshim:
LGTM

minahlee:

zeppelin-web/src/components/navbar/navbar.html
```diff
@@ -62,15 +62,17 @@
...
              ng-submit="search(searchTerm)">
             <div class="input-group">
               <input
                 type="text"
+                style="min-width:300px;"
                 ng-model="searchTerm"
                 ng-disabled="!navbar.connected"
                 class="form-control"
-                placeholder="Search in your notebooks"
+                placeholder="Search for your Notebooks"
```
felixcheung: I think, this is actually "in" - we searching for text matches in text, name, title and so on.
Perhaps say "Search your notebooks"?

cloverhearts: It is a good idea.
Fixed completed.

PR #940, opened 2016-05-31
[ZEPPELIN-809] Add annotation to public API

Leemoonsoo:
### What is this PR for?
Implement @ZeppelinApi @Experimental annotations and apply them in the public api.
Also mark @deprecated global scope angular object methods

### What type of PR is it?
Feature

### Todos
* [x] - Implement @ZeppelinApi, @Experimental annotations
* [x] - Apply annotation to public Api
* [x] - Mark deprecated global scope angular object related methods

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-809

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -32,6 +32,7 @@ import org.apache.spark.SparkContext;
 import org.apache.spark.sql.SQLContext;
 import org.apache.spark.sql.catalyst.expressions.Attribute;
 import org.apache.spark.sql.hive.HiveContext;
+import org.apache.zeppelin.annotation.ZeppelinApi;
 import org.apache.zeppelin.display.AngularObject;
 import org.apache.zeppelin.display.AngularObjectRegistry;
 import org.apache.zeppelin.display.AngularObjectWatcher;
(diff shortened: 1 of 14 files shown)


cloverhearts:
Looks goot to me +1

felixcheung:
what's our guideline on what is experimental?
should that be added to https://zeppelin.incubator.apache.org/docs/0.6.0-incubating-SNAPSHOT/rest-api/rest-interpreter.html?

bzz:
Documenting stable API with annotations looks great to me.

Leemoonsoo:
@felixcheung 

I have added @ZeppelinApi annotation to REST api endpoint.

Here's javadoc comment about two annotations. 

ZeppelinApi
```
APIs exposed to extends pluggable components or exposed to enduser
```

Experimental
```
Might change or be removed at anytime, or be adopted as ZeppelinApi
```

What do you think?

felixcheung:
sounds good, but how does a contributor/developer decides whether something new should be annotated as "Experimental" or "ZeppelinApi"?

Leemoonsoo:
I think it can be decided by consensus. If community can not sure API will stay unchanged until next major release (at least), it should be annotated as 'Experimental' otherwise we can annotate 'ZeppelinApi'. Once we get enough experiences of these annotation, then we can build more detailed guideline. 

Do you have any good idea about how does a contributor decides?

felixcheung:
Agreed. I think the model used in Spark is that if an API is:
- new
- expecting (significant) discussions/changes in this or the next minor release
Then it should be "Experimental".

felixcheung:
LGTM

Leemoonsoo:
Thanks @cloverhearts @felixcheung @bzz taking a look.

PR #948, opened 2016-06-02
Remove doap.rdf from gh-page branch

minahlee:
### What is this PR for?
Keep doap.rdf file only in master branch to reduce redundancy since we moved gh-page contents to master long time ago.


bzz:
I have included this file to `gh-pages` branch in order for it to be part of our website http://zeppelin.incubator.apache.org/doap.rdf

It was linked using this url to https://projects-new.apache.org so in case you want to change that setup - I think we need to update that resource as well (By editing https://svn.apache.org/repos/asf/comdev/projects.apache.org/data/projects.xml as TLP)

minahlee:
We don't update our website with `gh-pages` but with `master` so it will be still accessible via http://zeppelin.incubator.apache.org/doap.rdf
Thanks for pointing out svn update, I will do it for sure

minahlee:
Sorry I was confused, let's keep it in `gh-pages`

PR #953, opened 2016-06-02
ZEPPELIN-947: Change docs link to point the same Zeppelin version

AhyoungRyu:
### What is this PR for?
Currently, Zeppelin documentation link in Zeppelin home is pointing to http://zeppelin.apache.org/docs/latest/index.html. But maybe some people is not using the latest version. So the link should point to the same Zeppelin version that user is using now.

I just change the `latest` -> `{{zeppelinVersion}}`

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-947](https://issues.apache.org/jira/browse/ZEPPELIN-947)

1. After applying this patch and browse Zeppelin home. 
2. Just click `Zeppelin Documentation` link in here 
screen shot 2016-06-02 at 1 11 17 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/home/home.html
+++ b/zeppelin-web/src/app/home/home.html
@@ -78,7 +78,7 @@ limitations under the License.
         <div class="col-md-6">
           <h4>Help</h4>
           Get started with <a style="text-decoration: none;" target="_blank"
-                              href="http://zeppelin.apache.org/docs/latest/index.html">Zeppelin documentation</a><br>
+                              href="http://zeppelin.apache.org/docs/{{zeppelinVersion}}/index.html">Zeppelin documentation</a><br>
 
           <h4>Community</h4>
           Please feel free to help us to improve Zeppelin, <br>


corneadoug:
The CI failure is unrelated,
Tested, LGTM

prabhjyotsingh:
I have seen in some of the organisations where the build/release team edits pom.xml and truncates `-SNAPSHOT` e.g. 0.6.0-SNAPSHOT becomes 0.6.0, for their own cache purposes. 

So, IMO if we are going ahead with this, then may be we should have a redirect of http://zeppelin.apache.org/docs/{anything-else}/index.html to http://zeppelin.apache.org/docs/latest/index.html. Any thoughts ?

AhyoungRyu:
@prabhjyotsingh Appreciate for sharing your thought! and 
@Leemoonsoo Thanks for taking care that :)

corneadoug:
@Leemoonsoo not sure your PR is doing it, the idea was to use `latest` instead of `0.6.0-SNAPSHOT`

Leemoonsoo:
@corneadoug updated #961, please check!

prabhjyotsingh:
LGTM. We can merge this if no more discussion.

felixcheung:
but latest is generally older than the unreleased snapshot build?
for instance, https://zeppelin.incubator.apache.org/docs/latest/ now points to 0.5.6 release doc, not 0.6.0-snapshot doc.

AhyoungRyu:
@felixcheung Yeah right. As you can see, that's why I changed the link `http://zeppelin.apache.org/docs/latest/index.html` -> `http://zeppelin.apache.org/docs/{{zeppelinVersion}}/index.html`. Did I understand your comment incorrectly ?

felixcheung:
Right. This is good. I was not sure about the redirect comment above but I didn't really have a better solution.



		_____________________________

PR #955, opened 2016-06-03
[ZEPPELIN-940] Allow zeppelin server to connect to already executing Remote Interpreter

SachinJanani:
### What is this PR for?
Currenlty zeppelin server starts interpreter on localhost and with random port.The purpose of this pull request is to allow zeppelin server to connect to remotely executing zeppelin interpreter that user might have started in his service.This feature will be further helpful while cluster manager is to be implemented.(https://cwiki.apache.org/confluence/display/ZEPPELIN/Cluster+Manager+Proposal) 

### What type of PR is it?
Improvement

### Todos
[ ] -Add documentation

### What is the Jira issue?
* [ZEPPELIN-940] https://issues.apache.org/jira/browse/ZEPPELIN-940

Added Junit test in RemoteInterpreterProcessTest and it passes

### Questions:
* Does this needs documentation? Yes for the new properties


Leemoonsoo:
Thanks @SachinJanani for the contribution.

Ideally configure host and port can be moved from Property to [InterpreterOption](https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java) class.

That'll requires some frontend work, too. It's basically add UI component for host, port under 'Option' section, in interpreter setting page (e.g. https://github.com/apache/incubator-zeppelin/blob/master/zeppelin-web/src/app/interpreter/interpreter-create/interpreter-create.html#L38)

@SachinJanani Could you manage to do it? Let me know if you need any help!

SachinJanani:
@Leemoonsoo You mean that there should be one more option under option dropdown something like "isexecuting" and when this option is selected we have to show 2 properties Host and port?

Leemoonsoo:
@SachinJanani Exactly!

SachinJanani:
@Leemoonsoo As suggested i have added UI component for the host and port in the interpreter settings.Please see the attached image of how it will look.

Leemoonsoo:
@SachinJanani Thanks for the update!

In 'option' section, 'shard | scope | isolated' dropdown is more for how interpreter instances/processes are mapped to notebooks. 'executing' is little bit different from them i think. So how about separate from dropdown list? I think something like a check box could work to toggle 'executing'.

And what do you think about use little bit more specific name for label and variable name, instead of 'executing'? for example, 'connectToExistingProcess' or any name that can explain that it is not spawning new interpreter process but connect to existing process.

Lastly, this branch somehow includes commits not belongs to this contribution, for example afb1214, c4170b5, etc. Could you make sure that this branch does not include commits from other contributions?

SachinJanani:
@Leemoonsoo Even I think adding a checkbox will make more sense.As TravisCI was failing I have merged master in my fork which caused those commits to appear as a part of this contribution I will remove those commits from this contribution

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Constants.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.interpreter;
+/**
+ * Interpreter related constants
+ * 
+ *
+ */
+public class Constants {
```
jongyoul: How about moving these settings into ZeppelinConfiguration?

jongyoul:
@SachinJanani Thanks for the contribution. BTW, could you please rebase your change from master? If you merge something from master, it makes review on your PR hard.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Constants.java
SachinJanani: @jongyoul Please correct me if  I am wrong but I think moving it to ZeppelinConfiguration will make these settings global to all the interpreters.

jongyoul: @SachinJanani You are not wrong, definitely. I think it's better to place all of constants' variables in a same class if we can. How about you?

SachinJanani: @jongyoul Ok so you mean just the constants should be moved to ZeppelinConfiguration.But the issue is Zeppelinconfiguration is a part of zeppelin-zengine and adding it as a dependency in zeppelin-interpreter for accessing these constants will cause cyclic dependency.So I think Constants.java should be used to maintain the constants related to interpreter.

SachinJanani:
@Leemoonsoo @jongyoul This is what will look now:


Please let me know if there are any further changes need to be done

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Constants.java
jongyoul: @SachinJanani Yes, you are correct, and that's the reason why I ask you. I also think what you implement is utmost solution.

SachinJanani:
@Leemoonsoo @jongyoul   One of the TravisCI task is failing but I think its not related to my changes, is there any way I can rerun the travisCI without pushing changes?

jongyoul:
That failure is related to https://issues.apache.org/jira/browse/ZEPPELIN-878, but I have a question. What kind of procedure do you do while you run Zeppelin? Generally, when you stop Zeppelin, all interpreters also stop. Could you please share your cases? This PR will work, but because Zeppelin's interpreter will not be started by itself yet.

SachinJanani:
@jongyoul There can be scenarios where user is starting RemoteInterpreter embedded in his service by calling
`RemoteInterpreterServer server = new RemoteInterpreterServer(port)
server.start()` 
So this changes will help user to specify the host and port on which interpreter is executing and zeppelin will connect to that interpreter.This changes will also be helpful when ClusterManager(https://cwiki.apache.org/confluence/display/ZEPPELIN/Cluster+Manager+Proposal) will be implemented.We are starting ZeppelinInterpreter on remote host along with our service.
You are right that when we stop Zeppelin all interpreter will be stop but it can be handled by overriding the RemoteInterpreterServer

jongyoul:
@SachinJanani Thanks for sharing. I also think we should consider running Zeppelin's interpreters in a different machine. This would be a starting point to do this.

SachinJanani:
@jongyoul Yes you are right.But this changes will allow user to connect to existing remote process instead of zeppelin running the interpreter.I think we should create a separate jira for starting zeppelin interpreter on different machine.

jongyoul:
@Leemoonsoo Who's proper reviewer for UI part?

Leemoonsoo:
@jongyoul We have some front-end expert in the community like @corneadoug. But basically anyone in the community can participate review.

Thanks @SachinJanani for the contribution. It looks good to me.

jongyoul:
@Leemoonsoo Yes, I just think there're too many requests for him, so I'm asking you to have another candidate.

corneadoug:
@jongyoul Tested, LGTM.
What about the doc? Should it be part of this PR or a separate one?

jongyoul:
@corneadoug Thanks, and about review, I have been worried that you have been pressed with all of review requests of front-side issues.

SachinJanani:
@jongyoul Should I add section in "docs/manual/interpreters.md" something like "Connecting to the Existing Remote Interpreter"?

jongyoul:
@SachinJanani Yes, it's one of the best place.

SachinJanani:
@jongyoul @corneadoug Done with the docs can you please have a look.

docs/manual/interpreters.md
```diff
@@ -60,3 +60,16 @@ Each Interpreter Setting can choose one of 'shared', 'scoped', 'isolated' interp
 In 'shared' mode, every notebook bound to the Interpreter Setting will share the single Interpreter instance. In 'scoped' mode, each notebook will create new Interpreter instance in the same interpreter process. In 'isolated' mode, each notebook will create new Interpreter process.
 
 <img src="../assets/themes/zeppelin/img/screenshots/interpreter_persession.png" width="400px">
+
+
+## Connecting to the Existing Remote Interpreter
+
+Zeppelin users can start interpreter thread embedded in their service.This will provide flexibility to user to start interpreter on remote host. To start interpreter along with your service you have to create an instance of ``RemoteInterpreterServer`` and start it as follows:
```
jongyoul: Please add whitespace after "."

docs/manual/interpreters.md
```diff
@@ -60,3 +60,16 @@ Each Interpreter Setting can choose one of 'shared', 'scoped', 'isolated' interp
...
+
+``RemoteInterpreterServer interpreter=new RemoteInterpreterServer(3789); ``  
+``// Here, 3678 is the port on which interpreter will listen.``   
+``interpreter.start()``  
+
+The above code will start interpreter thread inside your process.Once the interpreter is started you can configure zeppelin to connect to RemoteInterpreter by checking **Connect to existing process** checkbox and then provide **Host** and **Port** on which interpreter porocess is listening as shown in the image below:
```
jongyoul: here, too.

jongyoul:
LGTM except minor comments.

docs/manual/interpreters.md
```diff
@@ -60,3 +60,16 @@ Each Interpreter Setting can choose one of 'shared', 'scoped', 'isolated' interp
 In 'shared' mode, every notebook bound to the Interpreter Setting will share the single Interpreter instance. In 'scoped' mode, each notebook will create new Interpreter instance in the same interpreter process. In 'isolated' mode, each notebook will create new Interpreter process.
 
 <img src="../assets/themes/zeppelin/img/screenshots/interpreter_persession.png" width="400px">
+
+
+## Connecting to the Existing Remote Interpreter
+
+Zeppelin users can start interpreter thread embedded in their service. This will provide flexibility to user to start interpreter on remote host. To start interpreter along with your service you have to create an instance of ``RemoteInterpreterServer`` and start it as follows:
+
+``RemoteInterpreterServer interpreter=new RemoteInterpreterServer(3789); ``
```
AhyoungRyu: @SachinJanani 
Currently this part is shown like below 
screen shot 2016-06-09 at 11 28 25 pm

How about changing these 3 code block to like this ?
screen shot 2016-06-09 at 11 33 33 pm


Then it will be like this 
screen shot 2016-06-09 at 11 28 07 pm

SachinJanani: @AhyoungRyu Thanks for review,Yes I think that will look better.Uploaded the changes please have a look

AhyoungRyu: @SachinJanani Yeah thanks for your quick response :) Looks good.

SachinJanani:
Can we merge this if there are no more comments?

Leemoonsoo:
Thanks @SachinJanani  for the contribution.

SachinJanani:
Thanks @Leemoonsoo @jongyoul @AhyoungRyu @corneadoug For reviewing.Can someone please merge this pull request?

PR #957, opened 2016-06-03
ZEPPELIN-900 Show interpreter logs in front side's result area while processing

jongyoul:
### What is this PR for?
Showing logs while running paragraph. This is POC level now.

### What type of PR is it?
[Improvement]

### Todos
* [x] - PoC
* [ ] - TBD

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-900

Just build and run long running command
```
%sh
sleep 10 && echo "success"
```
```
sc.version
sc.parallelize(1 to 100000).map(p => p/100 -> p).reduceByKey(_+_).collect().foreach(println)
Thread.sleep(10000)
```

### Questions:


felixcheung:
WIP?

bzz:
@felixcheung It looks like it's in PoC stage now. 

AFAIK the [gist of this proposal](https://github.com/apache/incubator-zeppelin/pull/957/files#diff-f9d22f7302ae356454cdcc637942856fR140) is to make interpreter's Log4J in-memory writeAppender  available to the frontend, which sounds awesome to me.

@jongyoul some quick feedback 
 - usually adding screenshots helps to get people more excited and provide better feedback and review changes faster
 - AFAIK interpreter logs can grow at least O(N) where N is the number of job executions. Will that work, given that interpreter logs (parts?) are going to be transmitted to potentially remote machine?

Alos, could you please help me to understand, will `InterpreterResult` return *full interpreter logs* from every time it is asked for (i.e by `getProgress()`) and send them?

jongyoul:
@bzz Thanks for the review. Quick answer for last question. InterpreterResult include new logs only after `getProgress()` called before. See [here](https://github.com/apache/incubator-zeppelin/pull/957/files#diff-f9d22f7302ae356454cdcc637942856fR145). Thus logs actually doesn't grow O(N) if getProgress is called periodically. But I'm also finding a better way to reduce memory usage.

Another problem with this implementation, it cannot covers multi-tenancy at all because that are thread-unsafe and unaware users from interpreter context.

And I also think of finding a better name of Remote*.

@felixcheung I'm very glad of you to review this PR, and give some idea. Thanks.

bzz:
@jongyoul how do you think, shall we close this one for now and re-open it later or do you still work on it?

jongyoul:
@bzz I'll continue to work with it after interpreter impersonation.

PR #965, opened 2016-06-06
[ZEPPELIN-957] Normalize note name.

astroshim:
### What is this PR for?
It's better if notebook name can be normalized.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-957

Try create notebook name with many slashes and no slashes.
You can refer to screen shot.

  - before
be

  - after
af

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -514,7 +514,7 @@ public class NotebookServer extends WebSocketServlet implements
 
     return cronUpdated;
   }
-  private String normalizeNoteName(String path){
+  private String normalizePath(String path){
     path = path.trim();
     path = path.replace("\\", "/");
     while (path.indexOf("///") >= 0) {
@@ -541,7 +541,7 @@ public class NotebookServer extends WebSocketServlet implements
         noteName = "Note " + note.getId();
       }
       if (noteName.indexOf('/') >= 0) {
-        noteName = normalizeNoteName(noteName);
+        noteName = normalizePath(noteName);
       }
       note.setName(noteName);
     }
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -525,7 +525,7 @@ public class NotebookServer extends WebSocketServlet implements
       if (noteName == null || noteName.isEmpty()){
         noteName = "Note " + note.getId();
       }
-      if (noteName.indexOf('/') >= 0) {
+      if (noteName.indexOf('/') >= 0 || noteName.indexOf('\\') >= 0) {
         noteName = Util.normalizePath(noteName);
       }
       note.setName(noteName);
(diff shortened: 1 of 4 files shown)


cloverhearts:
LGTM

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -514,6 +514,22 @@ private boolean isCronUpdated(Map<String, Object> configA,
 
     return cronUpdated;
   }
+  private String normalizeNoteName(String path){
+    path = path.trim();
+    path = path.replace("\\", "/");
```
felixcheung: does this change windows path like `mynotebooks\notebook1`?

astroshim: Yes, That line change for the windows path to '/'.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -514,6 +514,22 @@ private boolean isCronUpdated(Map<String, Object> configA,
...
+    while (path.indexOf("///") >= 0) {
+      path = path.replaceAll("///", "/");
+    }
+    path = path.replaceAll("//", "/");
+    if (path.length() == 0) {
+      path = "/";
+    } else if (path.charAt(0) != '/'){
+      path = "/" + path;
```
Leemoonsoo: Sometimes i want to display my note name not like "/My note" but like "My note". So It's bit odd to me force add '/' in front. What do you think?

astroshim: I think the method name is odd, This method is for the path not the notename. How about the method name is 'normalizeNotePath'?

corneadoug:
Tested

felixcheung:
should this check for both `/` and `\` then for windows?
https://github.com/apache/zeppelin/pull/965/files#diff-bcbdda87aa29314ed459e8212067e313R528

astroshim:
@felixcheung Thank you for pointing out that i missed. I just fixed it.

Leemoonsoo:
Is normalization applied on note name change?

Leemoonsoo:
Thanks @astroshim for the improvement. I've tested and although it's working well, i think there're little more things we can think

1. Normalization is not applied when change note name.
2. Normalization routine need to be placed in zeppelin-zengine, rather than zeppelin-server
    zeppelin-server is place for the code that enables front-end / back-end communication.
    business logic handles notebook should be in zeppelin-zengine module.
3. It would be great to have a unittest for normalize note name
4. Note name display is bit inconsistent. 

with this branch

| Given name when create note  | note name displayed on screen |
| ------------- | ------------- |
| MyNote  | MyNote  |
| /MyNote  | /MyNote  |
| MyNote/sub | **/MyNote/sub** |
| /MyNote/sub | /MyNote/sub |

but it should be like

| Given name when create note  | note name displayed on screen |
| ------------- | ------------- |
| MyNote  | MyNote  |
| /MyNote  | /MyNote  |
| MyNote/sub | **MyNote/sub** |
| /MyNote/sub | /MyNote/sub |

astroshim:
I really appreciate your detailed review. @Leemoonsoo 
I will fix 1,2,3 but in case 4, 
I thought ```/MyNote/sub``` and ```MyNote/sub``` are the same from directory point of view.
Do you think they should be treated differently?

Leemoonsoo:
@astroshim Thanks!
`MyNote` and `/MyNote` are treated differently. And i think it's right. (compatible to previous version)
Therefore I think it's more consistent to treat differently `MyNote/sub` and `/MyNote/sub`, too.

astroshim:
@Leemoonsoo Thank you for kindly explaination. I understand about name.

astroshim:
@Leemoonsoo I fixed 1,2,3,4. please review.

Leemoonsoo:
Thanks for quick response. LGTM

Leemoonsoo:
Merge it into master if there're no more discussions

PR #966, opened 2016-06-06
[ZEPPELIN-958] Support syntax highlight for python and r interpreter

minahlee:
### What is this PR for?
Support syntax highlight for python and r interpreter

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-958](https://issues.apache.org/jira/browse/ZEPPELIN-958)

**Before**
screen shot 2016-06-06 at 12 21 56 am

**After**
screen shot 2016-06-05 at 7 42 55 pm

**Before**
screen shot 2016-06-05 at 7 30 53 pm

**After**
screen shot 2016-06-05 at 7 30 31 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -79,8 +79,9 @@ angular.module('zeppelinWebApp')
   var angularObjectRegistry = {};
 
   var editorModes = {
-    'ace/mode/python': /^%(\w*\.)?pyspark\s*$/,
+    'ace/mode/python': /^%(\w*\.)?(pyspark|python)\s*$/,
     'ace/mode/scala': /^%(\w*\.)?spark\s*$/,
+    'ace/mode/r': /^%(\w*\.)?(r|sparkr|knitr)\s*$/,
     'ace/mode/sql': /^%(\w*\.)?\wql/,
     'ace/mode/markdown': /^%md/,
     'ace/mode/sh': /^%sh/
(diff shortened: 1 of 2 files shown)


felixcheung:
LGTM
btw, this is getting fragile - would be great if interpreter can have an extension mechanism to control what it is mapping to on the js side.

jongyoul:
@felixcheung Could you please tell me your idea in details? Basically I agree with you, but I'm not good at frontend, I don't know whether the feature can be adopted dynamically or not. If it's possible, you could that feature into Interpreter scope. It seems nice.

jongyoul:
BTW, this PR looks good to me. We can discuss @felixcheung's issue after merging this PR.

bzz:
Looks great to me, let's merge as a short term solution. 

And yes, as @felixcheung mentioned - in a mid\long term, it would be great for interpreter to encapsulate this knowledge!

bzz:
Merging if there is no further discussion

felixcheung:
right, I think that can be added to interpreterSettings and checked on the frontend similar to this:
https://github.com/apache/zeppelin/blob/09bc8a36a437a2a37736449e330a7594c89f495d/zeppelin-web/src/app/interpreter/interpreter.html#L90

bzz:
Yes.. and this method does not work if the default interpreter is set to Python for the notebook
screen shot 2016-06-17 at 01 03 50

felixcheung:
That was only to map the list of interpreter to the language though.

To figure out which interpreter is active in the interpreter right now we have regex to match "%something" at the beginning which won't work for the default interpreter at all.

PR #971, opened 2016-06-07
[ZEPPELIN-961] Longer names to the notebook make the name-text and buttons float outside the border.

sagarkulkarni3592:
### What is this PR for?
To correct the styling of notebook header which has notebook name, in case when the name is longer.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
[ZEPPELIN-961](https://issues.apache.org/jira/browse/ZEPPELIN-961)

- Open any notebook.
- Give the long name.

### Screenshots
screen shot 2016-06-07 at 11 16 14 am
screen shot 2016-06-07 at 11 17 15 am
screen shot 2016-06-07 at 11 17 31 am

### Questions:


minahlee:
It would be better if we make it to be responsive to window width size.
Attaching my screenshot
screen shot 2016-06-08 at 10 35 40 am

sagarkulkarni3592:
@minahlee Thank you for that suggestion.
What if we push all the buttons to the new div present on next line, like :
screen shot 2016-06-09 at 3 44 20 pm
This will change the look.
If not, then we could adjust the shown name size.

minahlee:
I would prefer to keep `noteAction` size as default height. As a Zeppelin user I prefer to see more of my work space than notebook name and buttons.

sagarkulkarni3592:
I have adjusted the noteAction size by adjusting name component of the noteAction bar.
screen shot 2016-06-16 at 11 46 44 am
But, still there is a problem if the window size is reduced below that. It is because, all the buttons can not be fit in that div. So they overflow. Below is the screenshot if we don't show the name at all.
screen shot 2016-06-16 at 11 53 09 am

What should be the exact requirement in this case?

minahlee:
@sagarkulkarni3592 Could you rebase and resolve conflicts?

sagarkulkarni3592:
@minahlee Resolved the conflicts. Build is failing for non-related reasons.

minahlee:
Thanks for the update. I see one more thing needs to be handled.
Could you make btn group not to cover input box in below screenshot?
screen shot 2016-06-21 at 11 03 56 pm

sagarkulkarni3592:
@minahlee Done the changes necessary to make the overlap go away.

screen shot 2016-06-22 at 12 45 34 pm

sagarkulkarni3592:
Also, observed this. When the notebook name is shorter, buttons appear far from the name or box as shown.

screen shot 2016-06-22 at 12 48 00 pm
screen shot 2016-06-22 at 12 48 16 pm

After the commit, it is made more dynamic.

screen shot 2016-06-22 at 12 48 34 pm
screen shot 2016-06-22 at 12 48 46 pm

sagarkulkarni3592:
@minahlee Also merged it with the master branch.

minahlee:
LGTM merging it into master and branch-0.6

PR #978, opened 2016-06-09
[ZEPPELIN-869] Search notebook to handle empty result.

ravicodder:
### What is this PR for?
Search notebook to handle empty result.

### What type of PR is it?
Bug Fix
### What is the Jira issue?
[ZEPPELIN-869](https://issues.apache.org/jira/browse/ZEPPELIN-869?jql=project%20%3D%20ZEPPELIN%20AND%20text%20~%20%22empty%22)

1. Try to search something which should return empty result, you should get proper message for that.
2. search box should get cleared when clicked on any other link.

Before:
screen shot 2016-06-09 at 3 44 53 pm

After:
screen shot 2016-06-09 at 3 34 47 pm
### Questions:


bzz:
@ravicodder thank you for improvement!

\cc @felizbear as original author of search frontend for a review

prabhjyotsingh:
LGTM, I noticed you are clearing search box on click on any link 👍

felizbear:
Also, I would recommend modify `ng-disabled` of search form in navbar like this:
```
ng-disabled="!navbar.connected || !searchTerm"
```
so it doesn't fire when there is an empty field

felizbear:
Also, I think you will have to clean up scope after you clear the search field value. To see what I mean do the following:
1. Search for something, e.g. `bankText`
2. Navigate away from the search result page (search field now is empty)
3. Search again with the empty search field
4. See `bankText` results again (expected to see an empty page or even have search field disabled like I suggested above)

felizbear:
Thanks for contribution by the way :)

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -34,6 +36,18 @@ angular
...
+    if ($scope.notes.length === 0) {
+      $scope.isResult = false;
+    } else {
+      $scope.isResult = true;
+    }
+
+    var routeChangeEvent = $rootScope.$on('$routeChangeStart', function (event, next, current) {
+      if (next.originalPath !== '/search/:searchTerm') {
+        angular.element('#searchTermId').val('');
+        routeChangeEvent();
```
ravicodder: @felizbear by calling routeChangeEvent I am explicitly unbinding the event handler.
$routeScope.$on returns a deregistration function whose reference we are storing in routeChangeEvent.

ravicodder:
Hi @felizbear thanks for the review.
I have done the required changes, Please have a look.

zeppelin-web/src/app/search/result-list.controller.js
felizbear: Maybe it makes sense to rename `routeChangeEvent` into something like `unbindRouteChangeEvent`? with such name no questions would arise about why it is invoked

felizbear:
@ravicodder except for [this remark](https://github.com/apache/zeppelin/pull/978#discussion_r66751128), LGTM

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -34,6 +36,19 @@ angular
 
       return note;
     });
+    if ($scope.notes.length === 0) {
+      $scope.isResult = false;
+    } else {
+      $scope.isResult = true;
+    }
+
+    var routeChangeEvent = $rootScope.$on('$routeChangeStart', function (event, next, current) {
```
corneadoug: This should be $scope.$on

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -34,6 +36,19 @@ angular
...
+      if (next.originalPath !== '/search/:searchTerm') {
+        angular.element('#searchTermId').val('');
```
corneadoug: unnecessary line

corneadoug:
The `searchTerm ` would probably be better inside the searchService instead of the $rootScope

zeppelin-web/src/app/search/result-list.controller.js
ravicodder: @felizbear ya, it makes sense.  Renamed routeChangeEvent to unbindRouteChangeEvent.

ravicodder:
@corneadoug Thanks for the review. Done required changes, kindly review it.

zeppelin-web/src/components/searchService/search.service.js
```diff
@@ -26,4 +26,6 @@ angular.module('zeppelinWebApp').service('searchService', function($resource, ba
     });
   };
 
+  this.searchTerm = '';
```
corneadoug: You will need to do a `this.searchTerm = term.q;` at the beginning of the `this.search` function, otherwise there is no search field when you reload the page

zeppelin-web/src/app/search/result-list.html
```diff
@@ -12,7 +12,7 @@
 limitations under the License.
 -->
 <div ng-controller="SearchResultCtrl" class="searchResults">
```
corneadoug: `ng-controller="SearchResultCtrl"` is not needed here, it is already set in the routes, right now it is instantiating the controller twice

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -34,6 +36,18 @@ angular
...
+    if ($scope.notes.length === 0) {
+      $scope.isResult = false;
+    } else {
+      $scope.isResult = true;
+    }
+
+    var unbindRouteChangeEvent = $scope.$on('$routeChangeStart', function (event, next, current) {
+      if (next.originalPath !== '/search/:searchTerm') {
+        searchService.searchTerm = '';
+        unbindRouteChangeEvent();
```
corneadoug: This is also not needed since the event handler is linked to the controller's scope.
The event handler is destroyed automatically

corneadoug:
@ravicodder Thank you for rebasing, I left a few more comments

ravicodder:
@corneadoug done required changes.
Please, have a look.

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -16,8 +16,10 @@
 
 angular
   .module('zeppelinWebApp')
-  .controller('SearchResultCtrl', function($scope, $routeParams, searchService) {
+  .controller('SearchResultCtrl', function($scope, $routeParams, searchService, $rootScope) {
```
prabhjyotsingh: I think now $rootScope can be removed, its unused.

ravicodder: @prabhjyotsingh removed $rootScope in latest commit.

corneadoug:

PR #979, opened 2016-06-09
Alert message when trying to delete last paragraph.

astroshim:
### What is this PR for?
This PR is for giving proper information to users and blocking unnecessary communication with zeppelin server when trying to delete last paragraph because the last paragraph can't be deleted.

### What type of PR is it?
Improvement 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-979

You can try to remove last paragraph of a notebook.

image

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -54,10 +54,10 @@ limitations under the License.
         </a>
       </li>
       <li>
-        <a ng-click="moveUp()"><span class="icon-arrow-up"></span> Move Up</a>
+        <a ng-click="moveUp()" ng-hide="$first"><span class="icon-arrow-up"></span> Move Up</a>
       </li>
       <li>
-        <a ng-click="moveDown()"><span class="icon-arrow-down"></span> Move Down</a>
+        <a ng-click="moveDown()" ng-hide="$last"><span class="icon-arrow-down"></span> Move Down</a>
       </li>
       <li>
         <a ng-click="insertNew()"><span class="icon-plus"></span> Insert New</a>
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -90,7 +90,7 @@ limitations under the License.
       </li>
       <li>
         <!-- remove paragraph -->
-        <a ng-click="removeParagraph()"><span class="fa fa-times"></span> Remove</a>
+        <a ng-click="removeParagraph()" ng-hide="$last"><span class="fa fa-times"></span> Remove</a>
       </li>
     </ul>
   </span>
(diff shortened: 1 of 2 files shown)


cloverhearts:
tested.
LGTM +1

corneadoug:
I would like to suggest a different approach:
Not showing the delete button if the paragraph is the last one.

It is also quite easy to do:
Just need to add `ng-hide="$last"` to the remove link in paragraph-control.html

screen shot 2016-06-10 at 3 43 12 pm

Since angular has both `$first` and `$last`, we could even remove 'Move Up' or 'Move Down' option depending on whether the paragraph is the first or the last.

astroshim:
@corneadoug Thank you very much for your review and opinion. I'll follow your advice.

minahlee:
@corneadoug I like the idea, but how do we want to handle paragraph deletion by keyboard shortcut in this case?

corneadoug:
@minahlee Maybe then in that case it would be needed

astroshim:
@minahlee @corneadoug Then we need two ways to solve this issue?

corneadoug:
I guess both methods can cohabitate then

astroshim:
Thank you for clarification.

astroshim:
@corneadoug @minahlee I fixed the code to hide "Remove" menu on the last paragraph menu. please review.

minahlee:
Tested, looks good. How about applying @corneadoug's opinion making the first paragraph not to show `Move Up`, last paragraph not to show `Move Down`?

astroshim:
@minahlee I'll take care about that you mentioned. thanks.

astroshim:
@minahlee Could you tell me how can i get your test failure?

corneadoug:
@astroshim She posted the test failure.
My guess is that the Notebook has only one paragraph, therefore the xpath is on the last paragraph.
Since we removed the delete element from the last paragraph, the test fail.

The exact failing tests and the error lines:
```
ParagraphActionsIT.testRemoveButton:153
ParagraphActionsIT.testCreateNewButton:86
```

astroshim:
@corneadoug Thank you for your kind explanation. 
I know the reason about the error but I couldn't reproduce the error even if i build like ```mvn clean package```. 
Did you build also like i did?

corneadoug:
You need to trigger the selenium test locally, it doesn't by default without some configuration.
(TEST_SELENIUM="true" as env)
Otherwise you can just push your changes and let the CI do it

astroshim:
@corneadoug Okay. Thank you!!

corneadoug:
Tested Selenium Tests locally from original commit:
https://github.com/apache/zeppelin/pull/979/commits/6f181efa09f4d616ac66eec5641df954883af432

And it's working well.

I will create a flacky test issue on JIRA,
@astroshim Could you rewrite history back to the commit I just Mentioned? After that LGTM

astroshim:
@corneadoug I really appreciate for your effort. Thank you. I reset my commits.

corneadoug:
Merging if there's no more discussions

PR #984, opened 2016-06-10
[ZEPPELIN-982] Improve interpreter completion API

AhyoungRyu:
### What is this PR for?
When people implement a new interpreter, they extend [interpreter.java](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java) as described in [here](https://zeppelin.apache.org/docs/0.6.0-SNAPSHOT/development/writingzeppelininterpreter.html). Among the several methods in [interpreter.java](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java), [completion API](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java#L109) enables auto-completion.

However this API is too simple compared to other project's auto-completion and hard to add more at the moment. So for the aspect of further expansion, it would be better to separate and restructure this API before the this release( 0.6.0 ).

### What type of PR is it?
Improvement

### Todos
* [x] - Create new structure : `InterpreterCompletion` in `RemoteInterpreterService.thrift` and regenerate `zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/thrift/*` files 
* [x] - Change all existing `List<String> completion` -> `List<InterpreterCompletion> completion` 
* [x] - Change `paragraph.controller.js` to point real `name` and `value`

### What is the Jira issue?
[ZEPPELIN-982](https://issues.apache.org/jira/browse/ZEPPELIN-982)

Since this improvement is just API change, it should work same as before. So after applying this patch, and check whether auto-completion works well or not. 

Use `. + ctrl` for auto-completion. For example,

```
%spark 
sc.version
```

When after typing `sc.` and pushing  `. + ctrl` down, `version` should be shown in the auto-completion list. 

auto_completion

### Questions:


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -416,10 +413,12 @@ public String getFormType(String noteId, String className) throws TException {
   }
 
   @Override
-  public List<String> completion(String noteId, String className, String buf, int cursor)
+  public List<InterpreterCompletion> completion(String noteId, String className, String buf,
+                                                int cursor)
```
jongyoul: Could you please turn off vertical alignment? If you are using Intellij, you can turn it off where 'Editor' -> 'Code Style' -> 'java' -> 'Wrapping and Braces' -> 'Method declaration parameters' -> 'Align when multiline'.

jongyoul:
@AhyoungRyu What the `value` of InterpreterCompletion means? And I suggest we remove the mark abstract from Interpreter class because almost interpreter returns empty list and it's better to offer that behaviour by default. How about it?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
AhyoungRyu: Yeah I did as you said and repushed. If I did wrong, please let me know :) 
The reason why I did like this is this line is over 100 line length. So I cut this line on purpose.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -413,8 +413,8 @@ public String getFormType(String noteId, String className) throws TException {
   }
 
   @Override
-  public List<InterpreterCompletion> completion(String noteId, String className, String buf,
-                                                int cursor)
+  public List<InterpreterCompletion> completion(String noteId,
+                                                String className, String buf, int cursor)
```
minahlee: @AhyoungRyu Zeppelin follows google code style. See https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent

AhyoungRyu:
@jongyoul Makes sense. I addressed your feedback. Please see my last two commit :)

Regarding `value`, 
as you can see, I used two variables(`name` and `value`) for `InterpreterCompletion`. When you use the auto-completion in the paragraph, you can see the several `names` in the suggested list. And if you select something in this list, the `value` that you choose will be returned. Actually these two variables are already existed in [paragraph.controller.js](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js#L786). I just connected these two.

jongyoul:
@AhyoungRyu I see. Thanks.

bzz:
@AhyoungRyu there are 2 different test failures on CI here:

**Integration test** fails https://s3.amazonaws.com/archive.travis-ci.org/jobs/137185986/log.txt
```
Tests in error: 
  ZeppelinIT.testSparkInterpreterDependencyLoading:210 Â» NoSuchElement Unable to...

Caused by: org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"//div[@class='modal-dialog'][contains(.,'Do you want to update this interpreter and restart with new settings?')]//div[@class='modal-footer']//button[contains(.,'OK')]"}
```

**Cassandra interpreter test** failure https://api.travis-ci.org/jobs/137185978/log.txt?deansi=true
```
.apache.zeppelin.cassandra.CassandraInterpreterTest  Time elapsed: 29.999 sec  <<< ERROR!
java.lang.ExceptionInInitializerError: null
	at com.datastax.driver.core.exceptions.NoHostAvailableException.copy(NoHostAvailableException.java:84)
	at com.datastax.driver.core.exceptions.NoHostAvailableException.copy(NoHostAvailableException.java:37)
	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:37)
	at com.datastax.driver.core.DefaultResultSetFuture.getUninterruptibly(DefaultResultSetFuture.java:245)
	at com.datastax.driver.core.AbstractSession.execute(AbstractSession.java:63)
	at info.archinnov.achilles.script.ScriptExecutor.executeScript(ScriptExecutor.java:61)
	at info.archinnov.achilles.embedded.AchillesInitializer.executeStartupScripts(AchillesInitializer.java:204)
	at info.archinnov.achilles.embedded.AchillesInitializer.initialize(AchillesInitializer.java:98)
	at info.archinnov.achilles.embedded.AchillesInitializer.initializeFromParameters(AchillesInitializer.java:61)
	at info.archinnov.achilles.embedded.CassandraEmbeddedServer.<init>(CassandraEmbeddedServer.java:76)
	at info.archinnov.achilles.embedded.CassandraEmbeddedServerBuilder.buildNativeSessionOnly(CassandraEmbeddedServerBuilder.java:436)
	at org.apache.zeppelin.cassandra.CassandraInterpreterTest.<clinit>(CassandraInterpreterTest.java:59)
Caused by: com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9227
```

If you believe neither of them are related to this changes - the best way to handle it is to file a JIRA issues (with the name of the test failing in the title), label `flaky-test` and attach the links to CI logs.

zeppelin-interpreter/src/main/thrift/RemoteInterpreterService.thrift
```diff
@@ -56,6 +56,11 @@ struct RemoteInterpreterEvent {
   2: string data      // json serialized data
 }
 
+struct InterpreterCompletion {
+  1: string name,
+  2: string value
```
bzz: They PR is huge so sorry if I have missed it, but could you please point to the place where this API is used?
Right now, looking at this it is not clear what exactly is `name` and `value` here

AhyoungRyu: @bzz Regarding the two value `name` and `value`, can [this comment](https://github.com/apache/zeppelin/pull/984#issuecomment-225249183) be an answer for your question ?

AhyoungRyu:
@bzz Yeah thanks for pointing it. I'll take a look at it.

zeppelin-interpreter/src/main/thrift/RemoteInterpreterService.thrift
bzz: Do you think we should add a link to this comment inside the source code then? 
Because the next person reading it will have exactly the same question as two others had here before. 

Or how do you think, is there a way we could improve things here, without referencing legacy naming from the frontend code? 

I.e ones the come to my mind and seems reasonable for a completion suggestion to have are: 
 - `text` and `code` or 
 - `visible_suggestion` and `source_code`
 - etc

zeppelin-interpreter/src/main/thrift/RemoteInterpreterService.thrift
```diff
@@ -56,6 +56,17 @@ struct RemoteInterpreterEvent {
...
+ * The below variables(name, value) will be connected to getCompletions in paragraph.controller.js
+ *
+ * name: which is shown in the suggestion list
+ * value: actual return value what you selected
+ */
+struct InterpreterCompletion {
+  1: string name,
+  2: string value
+}
+
```
AhyoungRyu: @bzz I added some descriptions instead of changing overall variable names. Since `name` & `value` is already existed  variable name in `paragraph.controll.js`. What do you think? :)

Leemoonsoo:
Thanks @AhyoungRyu for the improvement and taking care of comments.

LGTM. Merge it into master and 0.6 branch if there're no more discussions.

PR #987, opened 2016-06-10
ZEPPELIN-694 Remove remote option on InterpreterOption

jongyoul:
### What is this PR for?
Removing some redundant codes for simplifying Zeppelin.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Remove remote option from InterpreterOption

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-694

Users must be able to use Zeppelin without any changes

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -312,7 +312,6 @@ public class InterpreterFactory implements InterpreterGroupFactory {
       // While we decided to turn this feature on always (without providing
       // enable/disable option on GUI).
       // previously created setting should turn this feature on here.
-//      setting.getOption().setRemote(true);
 
       InterpreterSetting intpSetting = new InterpreterSetting(
           setting.id(),
@@ -499,19 +498,11 @@ public class InterpreterFactory implements InterpreterGroupFactory {
     AngularObjectRegistry angularObjectRegistry;
 
     InterpreterGroup interpreterGroup = new InterpreterGroup(id);
-//    if (option.isRemote()) {
     angularObjectRegistry = new RemoteAngularObjectRegistry(
         id,
         angularObjectRegistryListener,
         interpreterGroup
     );
-//    } else {
-//      angularObjectRegistry = new AngularObjectRegistry(
-//          id,
-//          angularObjectRegistryListener);
-//
-//      // TODO(moon) : create distributed resource pool for local interpreters and set
-//    }
 
     interpreterGroup.setAngularObjectRegistry(angularObjectRegistry);
     return interpreterGroup;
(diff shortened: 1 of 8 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -499,19 +499,19 @@ public InterpreterGroup createInterpreterGroup(String id, InterpreterOption opti
...
+        angularObjectRegistryListener,
+        interpreterGroup
+    );
+//    } else {
+//      angularObjectRegistry = new AngularObjectRegistry(
+//          id,
+//          angularObjectRegistryListener);
+//
+//      // TODO(moon) : create distributed resource pool for local interpreters and set
+//    }
```
felixcheung: instead of commenting out, why not remove them?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
```diff
@@ -21,25 +21,25 @@
  *
  */
 public class InterpreterOption {
-  boolean remote;
+//  boolean remote;
```
Leemoonsoo: I think it's better keep this field for a while (with mark @deprecated or something)
Without this field exist with 'true' value assigned, when previous version of Zeppelin read notebook created from newer version, previous version of Zeppelin will assume remote = 'false' and some interpreters will not work. (e.g. spark)

Leemoonsoo: There're some corresponding parts in front-end side. for example https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/interpreter/interpreter.controller.js#L123

jongyoul: @Leemoonsoo Makes sense. I'll mark @deprecated instead of removing them.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
jongyoul: @felixcheung Even though github shows the changes but I thought it's easy to find out what the changes are. I'll remove them after review.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
```diff
@@ -21,7 +21,7 @@
  *
  */
 public class InterpreterOption {
-  boolean remote;
+  boolean remote = false;
```
Leemoonsoo: I think default value should be 'true' to allow previous version of zeppelin work with it.

jongyoul: @Leemoonsoo My mistake.

jongyoul:
I'll make a new PR after I finish some preceding tasks.

PR #374, opened 2015-10-28
ZEPPELIN-371 : Import a notebook

prabhjyotsingh:
Import a notebook by either an external URL or uploading a valid JSON.

screen shot 2015-10-28 at 4 53 58 pm

screen shot 2015-10-28 at 4 56 30 pm

screen shot 2015-10-28 at 4 56 55 pm

screen shot 2015-10-28 at 4 55 29 pm

PR #989, opened 2016-06-10
 Added ipfsnotebookrepo [ Zeppelin-683 ]

onkarshedge:
### What is this PR for?
IPFSNoteBookRepo IPFS storage and sharing zeppelin notes.

### What type of PR is it?
 Feature 

### Todos
* [x] checkpoint revisions, get particular revision
* [x] List of revision history
* [x] remove note and also a particular revision
* [x] support to import the Note when hash is given as input
* [x] Documentation 

### What is the Jira issue?
* [Jira issue](https://issues.apache.org/jira/browse/ZEPPELIN-683)

You need the [ipfs jar](https://github.com/ipfs/java-ipfs-api/blob/master/pom.xml) in the local .m2 repository. `mvn install:install-file` hers is complete [command](https://maven.apache.org/guides/mini/guide-3rd-party-jars-local.html)
Also please refer the `storage document` for more help.

[Successful import](https://gifyu.com/image/52t)
[Failure import](https://gifyu.com/image/52Y)

### Questions:
* Does the licenses files need update? Yes


onkarshedge:
@bzz Please review

onkarshedge:
@bzz please review

pom.xml
```diff
@@ -227,8 +227,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
```
jongyoul: Zeppelin uses jdk 1.7 by default. could you please revert this value?

onkarshedge:
@bzz please review

pom.xml
onkarshedge: @jongyoul  the ipfs jar I use requires jsk 1.8    I  also mentioned this in the google document linked above.

onkarshedge: @jongyoul  Yes I will revert that value to 1.7 with the next commit. But for this ipfs code to execute the mvn clean install must be with jsk 1.8.

bzz: I think the best bet right now is try to re-build ipfs jar with target 1.7 locally

bzz:
@onkarshedge I understand that it's WIP, just wanted to remind that eventually your `NotebookRepo` implementation should user `NotebookRepoVersioned`, see Git one as an example.

Could you also point to some document with high-level description of the approach how notebooks are stored in IPFS, what IPFS features are used and how to view them without Zeppelin, to make sure they are saved for people who are not very familiar with IPFS? 

Also, we need to know what is overall current implementation status, like what are TODOs, what is done, what is left, etc? This could be added i.e to the GDoc that you have linked

onkarshedge:
@bzz I had a doubt about versioning approach.
Currently I am saving(ipfs add 'note') the note in IPFS for every call to save function. This gives me a new multihash on saving. So for versioning I have to save the multihash with time. The `GitNotebookRepo` has a checkpoint which corresponds to git commit. Here in ipfs each call to ipfs add is like git commit.  So the List<Rev> from history function will be large because save() method is called multiple times.
So  the List<Rev> should contain only those few multihashes   which user wants to choose.

onkarshedge:
Should i save to Filesystem like `VFSNotebookRepo` and only when user chooses to version it add it to ipfs that way the ipfs will contain only the ipfs object corresponding to versioned notes and not all through the save method ?

bzz:
Hi @onkarshedge, as mid-term evaluation starts soon, I think it's crucial to get answers on 2 previous questions now.

>Could you also point to some document with high-level description of the approach how notebooks are stored in IPFS, what IPFS features are used and how to view them without Zeppelin, to make sure they are saved for people who are not very familiar with IPFS?

and

>Also, we need to know what is overall current implementation status, like what are TODOs, what is done, what is left, etc? This could be added i.e to the GDoc that you have linked

onkarshedge:
@bzz Here is the link to document for first question.
[High level description](https://docs.google.com/document/d/1OZOowc6aX3BrRkgjkUKwD19DshNXMWE2M07LMpXepx4/edit?usp=sharing)
TODO's is the frontend api : To enter the ipfs hash to download the note.

bzz:
Thank you very much @onkarshedge ! I will go though it and let you know

bzz:
@onkarshedge sorry, people were busy releasing 0.6.0 recently

Could you please:
 - [ ] rebase this code on latest `master` branch
 - [ ] update title (add WIP) and PR description to contain TODO list of things that are missing
 - [ ] post here your current status - like what is in-prgress right now or is there anything that blocks your progress, etc

bzz:
ping, @onkarshedge are there any updates for the past week?

onkarshedge:
@bzz right now I am working on Bittorrent support. I am using the [jlibtorrent](https://github.com/frostwire/frostwire-jlibtorrent) which uses libtorrent. The past week I spent time on using this library and some examples.I will post a PR in 2-3 days. Sorry for late response.

bzz:
Got it, thanks! Please submit another WIP pr with Bittorents support.

Could you still follow https://github.com/apache/zeppelin/pull/989#issuecomment-230972882 and update this PR as soon as you get back to IPFS one?

Thanks again!

bzz:
ping @onkarshedge 

Could you please follow https://github.com/apache/zeppelin/pull/989#issuecomment-230972882 and update this PR

onkarshedge:
@bzz To import the note by hash url . I will have to use [javascript-api](https://github.com/ipfs/js-ipfs-api) in this [notenameImport.controller.js](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/components/noteName-import/notenameImport.controller.js) file.
How do you suggest to use this?.

bzz:
Do you want to do that on the client-side, right in the browser? 

I see what you mean - our existing impor function indeed fetches from URL and parse JSON right in the browser, but I do not think that such architecture will suit our case, as then the whole IPFS stack need to be run on the client. And it is not something that we want add here (we want  self-contained `IpfsNotebookRepo` implementation)  

I think what you could do here is - create a new REST API endpoint, pass this hash (which is like an URL in our case) to this REST API, and make it delegate actual fetching logic to a method in `IpfsNotebookRepo` and then, on the client, just rear the response, as a JSON.

This logic can be later used for BitTorrent notebooks storage support as well as simple URL case could be refactored to use it (but that is outside of the scope of this PR of course) 

Let me introduce you to @corneadoug - he is known Zeppelin front-end ninja, whose knowledge of the client side codebase is by far greater than mine.

@corneadoug Could you please help @onkarshedge and guide him on refactoring\updating `notenameImport.controller.js` in such way, that it could delegate hash->JSON resolution in case of notebook import to the backend API, for the case of importing notes from new IPFS storage by IPFS hash (which is like a bittorent address)?

corneadoug:
@onkarshedge just like @bzz said, it would be better to have that import function in the back-end.
This is the [PR](https://github.com/apache/zeppelin/pull/374) that first implemented the feature so you should be able to find where are the classes currently taking care of the import.

We currently have 2 options for import in the front-end: **URL** or **File Import**
You could probably create a 3rd one for IPFS.

Right now, the [websocket call](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/components/noteName-import/notenameImport.controller.js#L97) is sending a Notebook, so you might want to refactor that to have a choice between sending a Notebook or a URL/Hash/Other.
And send the Hash to the backend so it can do the import.

bzz:
@onkarshedge what do you think?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/IPFSNotebookRepo.java
```diff
@@ -0,0 +1,312 @@
...
+    Callable<Note> task = new Callable<Note>() {
+      @Override
+      public Note call() throws Exception {
+        return getNote(multihash);
+      }
+    };
+    Future<Note> noteFuture = executorService.submit(task);
+    Note note = null;
+    try {
+      note = noteFuture.get(15, TimeUnit.SECONDS);
```
onkarshedge: @bzz When the note is not available from peer it should not wait indefinitely so I added timeout in backend. 
And the exception is logged. Should this be propagated to front-end ?
The front-end does not know about the TimeoutException.
@corneadoug should there be another timeout function in front-end ?

onkarshedge:
@bzz @corneadoug  please review,it is complete. I have added the screenshot.

corneadoug:
@onkarshedge I would need a scenario/setup to be able to test it, could you add it to the PR description?
Regarding the timeout function in front-end, I think it would be better to send back a websocket event message with import result to that client (not everybody), and then show it in a ng-toast. (green: The Notebook XXX was imported, red: We couldn't import Notebook XXX)

bzz:
@onkarshedge great progress, thank you for keeping us posted! 

Before moving further, on top of great suggestions from @corneadoug, there are few more things you also need to take care of:

 - Please make sure that `zeppelin-distribution/src/bin_license/LICENSE` includes all dependencies _AND all transitive dependencies_ that this changes bring to the project. You can see them with `mvn dependency:tree`

 - Please make sure that ALL your Java code follows the conventions from [project style guide](https://github.com/apache/zeppelin/blob/master/CONTRIBUTING.md#code-convention)

 - I noticed that CI is failing here. Could you, as an author of the PR, look into that and post there the reason of the failures here please? This would speed up the review and enable the reviewers to help you identify the reason faster.

onkarshedge:
@bzz there are no transitive dependencies, just Junit and hamcrest
The reason of CI failure is 

>[ERROR] Failed to execute goal on project zeppelin-zengine: Could not resolve dependencies for project org.apache.zeppelin:zeppelin-zengine:jar:0.7.0-SNAPSHOT: Failed to collect dependencies at org.ipfs:api:jar:0.0.1-SNAPSHOT: Failed to read artifact descriptor for org.ipfs:api:jar:0.0.1->SNAPSHOT: Could not transfer artifact org.ipfs:api:pom:0.0.1-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org: Name or service not known: Unknown host nexus.codehaus.org: Name or service not known -> [Help 1]

It has to be added  in local .m2 repository

onkarshedge:
@corneadoug  feedback addressed.

bzz:
@onkarshedge so you say it's not published on Maven Central? 

According to ipfs/java-ipfs-api#1 there is an option fetch it from `https://jitpack.io`. Can you please add new repository to `pom.xml` and try?  Making sure that new code does not break the CI is quite important.

onkarshedge:
@bzz I found this [documentation](http://docs.ipfs.apiary.io/#reference/version/version/version) and there would be no need for dependency. I have used apache-httpcomponents httpclient.

onkarshedge:
@bzz CI is green please review.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -183,6 +183,34 @@ public String exportNote(String noteId) throws IOException, IllegalArgumentExcep
...
+      try {
+        newNote = createNote(subject);
+        newNote.setName(oldNote.getName());
+        List<Paragraph> paragraphs = oldNote.getParagraphs();
+        for (Paragraph p : paragraphs) {
+          newNote.addCloneParagraph(p);
+        }
+        newNote.persist(subject);
+      } catch (IOException e) {
+        logger.error(e.toString(), e);
```
bzz: Instead of `e.toString()` we could have something like "Importing note from " + url+ " failed". The error message will still be there, as we propaget `e`.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -728,6 +731,27 @@ protected Note importNote(NotebookSocket conn, HashSet<String> userAndRoles,
...
+    if (fromMessage != null) {
+      String urlHash = (String) fromMessage.get("hash");
+      AuthenticationInfo subject = new AuthenticationInfo(fromMessage.principal);
+      note = notebook.importNote(urlHash, subject);
+      Message message = new Message(OP.IMPORT_NOTE_STATUS);
+      if (note != null) {
+        note.persist(subject);
+        message.put("importStatus", "success");
+      }
+      else {
```
bzz: Formatting is off here: `else` should be 1 line above

zeppelin-web/src/components/websocketEvents/websocketMsg.service.js
```diff
@@ -151,6 +151,15 @@ angular.module('zeppelinWebApp').service('websocketMsgSrv', function($rootScope,
       });
     },
 
+    importNoteFromIpfs: function(hashUrl) {
```
bzz: I think this might be called more generic, somethin like `importNoteFromBackend` as it may be another NotebookRepo implementation configured that actuallys upports same backend-style import.
How do you think?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/ipfs/Ipfs.java
```diff
@@ -0,0 +1,220 @@
...
+    baseUrl = url;
+    client = HttpClients.createDefault();
+  }
+
+  public String pinLs() {
+    String pinnedJsonString = null;
+    HttpGet request = new HttpGet(baseUrl + "pin/ls?type=recursive&quiet=true");
+    try {
+      CloseableHttpResponse response = client.execute(request);
+      try {
```
bzz: can we avoid having embedded `try` here by keeping only one and then multiple `catch` close please?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/ipfs/IPFSNotebookRepo.java
```diff
@@ -0,0 +1,345 @@
...
+    return versionHistory;
+  }
+
+  @Override
+  public void close() {
+    super.close();
+    try {
+      ipfs.getClient().close();
+    } catch (IOException e) {
+      LOG.info("Couldn't successfully close the ipfsClient");
```
bzz: It's good practice to propagate throwables to logs everywhere, like `LOG.info("Couldn't successfully close the ipfsClient", e)`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/ipfs/IPFSNotebookRepo.java
```diff
@@ -0,0 +1,345 @@
...
+    Note note = null;
+    try {
+      note = noteFuture.get(15, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      LOG.error("Failed to download Note, Interrupted", e);
+    } catch (ExecutionException e) {
+      LOG.error("Failed to get note", e);
+      Throwable cause = e.getCause();
+      if (cause instanceof IOException)
+        throw (IOException) cause;
```
bzz: Please make sure all your code follows project coding conventions https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/ipfs/IPFSNotebookRepo.java
```diff
@@ -0,0 +1,345 @@
...
+
+
+/**
+ * Share and get versioned notebooks with IPFS. Make sure IPFS daemon is running.
+ */
+public class IPFSNotebookRepo extends VFSNotebookRepo implements NotebookRepo {
+  private static final String API_SERVER_PROPERTY_NAME = "zeppelin.notebook.ipfs.apiServer";
+  private static final String DEFAULT_API_SERVER = "http://localhost:5001/api/v0/";
+  private static final Logger LOG = LoggerFactory.getLogger(IPFSNotebookRepo.class);
+  Type simpleType = new TypeToken<Map<String, String>>() {
```
bzz: Could you please explain, why does this:
 -  require to be a field? Vs local variable somewhere
 -  require a default visibility?

bzz:
Looks great, @onkarshedge! I posted a few comments above.

\cc @khalidhuseynov for extra back-end code review and @corneadoug for front-end one

zeppelin-web/src/components/websocketEvents/websocketMsg.service.js
corneadoug: I agree with @bzz 
The parameters structure could be more generic too
```
websocketEvents.sendNewEvent({
   op: 'IMPORT_NOTE_URL',
   data: {
      type: "IPFS"
      options: {
         hash: hashURL
      }
   }
 });
```

zeppelin-web/src/components/noteName-import/notenameImport.controller.js
```diff
@@ -110,4 +115,27 @@ angular.module('zeppelinWebApp').controller('NoteImportCtrl', function($scope, $
     vm.resetFlags();
     angular.element('#noteImportModal').modal('hide');
   });
+
+  $scope.$on('importNoteFail', function(event, data) {
```
corneadoug: 'importNoteFail' isn't a great name, how about 'importNoteResult'

zeppelin-web/src/components/noteName-import/notenameImport.controller.js
```diff
@@ -110,4 +115,27 @@ angular.module('zeppelinWebApp').controller('NoteImportCtrl', function($scope, $
...
+    vm.resetFlags();
```
corneadoug: This line isn't needed

zeppelin-web/src/components/noteName-import/notenameImport.controller.js
```diff
@@ -110,4 +115,27 @@ angular.module('zeppelinWebApp').controller('NoteImportCtrl', function($scope, $
...
+    angular.element('#noteImportModal').modal('hide');
```
corneadoug: This line isn't needed either (already hidden)

zeppelin-web/src/components/noteName-import/notenameImport.controller.js
```diff
@@ -110,4 +115,27 @@ angular.module('zeppelinWebApp').controller('NoteImportCtrl', function($scope, $
...
+    var status = data.importStatus;
```
corneadoug: One comment for all the code below:
* Since status is used once in a if, I don't think there is a need to assign its value to a variable
* There is a bit of repetition in the ngToast configuration, creating a config object before the if/else would limit the code:

```
var ngToastConf = {
  content: 'Failed to import Note with hash',
  verticalPosition: 'bottom',
  horizontalPosition: 'center',
  timeout: '5000'
};

if (data.importStatus === 'success') {
  ngToastConf.content = 'Successfully imported Note';
  ngToast.success(ngToastConf);
} else {
  ngToast.danger(ngToastConf);
}
```

* The Note name should probably be shown in the ngToast Message

zeppelin-web/src/components/noteName-import/notenameImport.controller.js
```diff
@@ -110,4 +115,27 @@ angular.module('zeppelinWebApp').controller('NoteImportCtrl', function($scope, $
...
+    console.log('Data is %o', data);
```
corneadoug: Let's remove that log

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
khalidhuseynov: +1 for having more detailed error message as mentioned by @bzz

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepo.java
```diff
@@ -112,6 +121,22 @@ public Revision(String revId, String message, int time) {
...
     public String message;
     public int time;
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (!(o instanceof Revision)) return false;
+
+      Revision revision = (Revision) o;
+
```
khalidhuseynov: may need to remove this and next empty line

conf/zeppelin-site.xml.template
```diff
@@ -158,6 +158,21 @@
...
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.ipfs.IPFSNotebookRepo</value>
+  <description>notebook persistence layer implementation for sharing notebooks</description>
+</property>
+
+<property>
+  <name>zeppelin.notebook.ipfs.apiServer</name>
+  <value>http://localhost:5001/api/v0/</value>
+  <description>ipfs api Server Address</description>
+  </property>
```
corneadoug: wrong indent here

corneadoug:
I'm a bit lost with this feature, is it really considered as a NotebookRepo or does it only keep the revisions? Does importing the Hash mean that it import a Revision as Notebook?

Why do we need to comment the VFS while it still save in local file system?

Also I think the ipfs installation steps shouldn't be in the NotebookRepo doc since not every user should be having a server running. It would be more fitting in this PR description on how to test this PR.

#### Import form
I would have classic URL by default, and the hash option only as a checkbox.

#### Import Note event listener
I left comments in the code

corneadoug:
Considering https://github.com/apache/zeppelin/pull/1304 is the import really needed?

khalidhuseynov:
in a sense i agree with @corneadoug and i'll write more details on that later, but for the beginning wouldn't it make sense to save normal notes without commit in ipfs as well?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/ipfs/Ipfs.java
onkarshedge: In the inner try `response.close()` throws a IOException. If I have a single try block,then the finally block will be outside and there will be try catch in finally block for `reponse.close()`.
```
finally {
      try {
        response.close();
      } catch (IOException | NullPointerException e) {
        LOG.error("Failed to close response", e);
      }
```
Is this the right way ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
onkarshedge: I referred the previous import [function](https://github.com/apache/zeppelin/blob/d11221fb8af5568416ef5041fc2da8b6fa08598b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java#L217).

onkarshedge:
@bzz  @corneadoug addressed the feedback, updated PR description and also images.
@khalidhuseynov saving normal notes would create an ipfs object for each save , we would have to discard the previous object and hash and will need to store the latest hash. It's similar to git, it would not be suitable to commit on every small change as save() function is called on timeout.

bzz:
@onkarshedge thank you for the great work!

How do you think, how hard it would be to rebase it on latest master?

PR #993, opened 2016-06-10
[ZEPPELIN-987] Enable user to secure interpreter setting, credentials and configurations info

AhyoungRyu:
### What is this PR for?
For some user cases, people might want to hide **Interpreter Setting**, **Credentials** and **Configurations** information to other users (who are defined in `conf/shiro.ini`). So I added 

```
#/api/interpreter/** = authc, roles[admin]
#/api/configurations/** = authc, roles[admin]
#/api/credential/** = authc, roles[admin]
``` 
below the [ [urls] ](https://github.com/apache/zeppelin/blob/master/conf/shiro.ini#L38) section. 

This issue was originally suggested at [Zeppelin user mailing list](https://mail-archives.apache.org/mod_mbox/zeppelin-users/201606.mbox/%3CCAPgU7Y%3DBJrXQ_P0ond4PTukoya0FEjwoPuUb31iN3qwo8iyM1Q%40mail.gmail.com%3E) by @TomNorden

### What type of PR is it?
Improvement | Documentation

### Todos
* [x] - Add `interpreter`, `credential` and `configuration` url to `conf/shiro.ini`
* [x] - Update `shiroauthentication.md` for this change
* [x] - Redirect to home with ngToast error message when status is `401`
* [x] - Rebase after #1100 merged and add error message to `Credential` menu as well

### What is the Jira issue?
[ZEPPELIN-987](https://issues.apache.org/jira/browse/ZEPPELIN-987)

1. Apply this patch and restart Zeppelin
2. Login with `admin` and `password1`
3. Go to interpreter, credential and configuration tab -> You can see all of the information in each tabs 
4. Logout -> Login again with `user1` and `password2`
5. Go to interpreter, credential and configuration tab -> In this time, you can't see all of the information in each tabs 

- When you login with `user1` (doesn't have permission to see the interpreter, credential and cofiguration info)
  - interpreter menu
interpreters
  - configuration menu
configurations
  - credential menu
credential-after

- `shiroauthentication.md`
screen shot 2016-06-10 at 12 25 02 pm

### Questions:
* Does this needs documentation? Yes, so I updated.

### DIFF
--- a/conf/shiro.ini
+++ b/conf/shiro.ini
@@ -58,6 +58,7 @@ shiro.loginUrl = /api/login
 role1 = *
 role2 = *
 role3 = *
+admin = *
 
 [urls]
 # This section is used for url-based security.
@@ -66,8 +67,8 @@ role3 = *
 # authc means Form based Auth Security
 # To enfore security, comment the line below and uncomment the next one
 /api/version = anon
-#/api/interpreter/** = authc, roles[admin]
-#/api/configurations/** = authc, roles[admin]
-#/api/credential/** = authc, roles[admin]
-/** = anon
-#/** = authc
+/api/interpreter/** = authc, roles[admin]
+/api/configurations/** = authc, roles[admin]
+/api/credential/** = authc, roles[admin]
+#/** = anon
+/** = authc
--- a/conf/shiro.ini
+++ b/conf/shiro.ini
@@ -18,7 +18,7 @@
 [users]
 # List of users with their password allowed to access Zeppelin.
 # To use a different strategy (LDAP / Database / ...) check the shiro doc at http://shiro.apache.org/configuration.html#Configuration-INISections
-admin = password1
+admin = password1, admin
 user1 = password2, role1, role2
 user2 = password3, role3
 user3 = password4, role2
@@ -60,10 +60,14 @@ role2 = *
 role3 = *
 
 [urls]
+# This section is used for url-based security.
+# You can secure interpreter, configuration and credential information by urls. Comment or uncomment the below urls that you want to hide. 
 # anon means the access is anonymous.
-# authcBasic means Basic Auth Security
 # authc means Form based Auth Security
 # To enfore security, comment the line below and uncomment the next one
 /api/version = anon
+#/api/interpreter/** = authc, roles[admin]
+#/api/configurations/** = authc, roles[admin]
+#/api/credential/** = authc, roles[admin]
 /** = anon
 #/** = authc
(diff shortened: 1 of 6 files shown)


prabhjyotsingh:
@AhyoungRyu  is this still a work in progress ? 
Is there a way in which we can go away with seeing all of these three munu i.e. Interpreter setting, Credentials and Configurations if user is not authorize  to ?

Otherwise LGTM.

AhyoungRyu:
@prabhjyotsingh No, I'm just waiting some reviews. I'm not sure that i understood your comment correctly,
>Is there a way in which we can go away with seeing all of these three menu i.e. Interpreter setting, Credentials and Configurations if user is not authorize to ?

I just wanted to make even if some users are already defined in `[users]` section, the other user ( maybe `admin`) can hide the three information. This is the scope of this PR :)

prabhjyotsingh:
I was thinking if all of these 3 Interpreter, Credentials, and Configuration menu can be hidden, like the way they are hidden in case of user not loggedin.

AhyoungRyu:
@prabhjyotsingh Yeah it makes sense. It would be better. Let me figure out it then :)

echarles:
@AhyoungRyu If you can come to an implementation that fits @prabhjyotsingh good idea, it would be great if you could make it `generic`, with a configurable list of functions that reflect in to the menu.

AhyoungRyu:
@echarles Yes. Thanks for your suggestion ! :)

cquptEthan:
nice！

corneadoug:
@AhyoungRyu Still working on this?
Could you rebase the branch with master?

AhyoungRyu:
@prabhjyotsingh Sorry for my late response. Since i think they need to know the existence of those menus(`Interpretes`, `Configurations` and `Credentials`), I added `ngToast` error message instead of totally hiding the menus. See the updated gif images in the PR description. What do you think?

prabhjyotsingh:
Sure this will work. Thanks for the fix.

conf/shiro.ini
```diff
@@ -60,10 +60,14 @@ role2 = *
 role3 = *
```
prabhjyotsingh: Can you add one more role admin i.e. `admin = *` here, so that other user can also use that role.

AhyoungRyu:
@prabhjyotsingh Thanks for your feedback. I addressed it in the last commit :)

prabhjyotsingh:
LGTM!

AhyoungRyu:
@prabhjyotsingh i'll rebase after #1100 merged and add the alert message to credential page as well :)

AhyoungRyu:
@prabhjyotsingh Since #1100 was merged into master, I updated `credential.controler.js`.

prabhjyotsingh:
Tested on both firefox and chrome. Works well. 👍

prabhjyotsingh:

PR #994, opened 2016-06-10
[ZEPPELIN-986] Create publish release script

minahlee:
### What is this PR for?
This PR is to automate release publish to maven repository.
We used to use maven-deploy-plugin and maven-release-plugin for release but somehow it didn't work well with Zeppelin so 0.5.5 and 0.5.6 haven't been published to maven repository.

Publishing release to maven repository will eventually help zeppelin to reduce binary package size by leading users to use Dynamic interpreter loading(#908).
Originally below modules were skipped for maven release
 - all interpreters(except spark)
 - zeppelin-display
 - zeppelin-server 
 - zeppelin-distribution

on the other hand this pr will skip only:
 - zeppelin-distribution

### What type of PR is it?
Infra

### Todos
- [x] Include SparkR/R interpreter in release
- [x] Create common_release.sh to remove build configuration duplication
- [x] Check curl networking failure

### What is the Jira issue?
[ZEPPELIN-986](https://issues.apache.org/jira/browse/ZEPPELIN-986)

### Questions:
* Does this needs documentation? Yes, https://cwiki.apache.org/confluence/display/ZEPPELIN/Preparing+Zeppelin+Release will be updated accordingly once this pr is merged.

### DIFF
--- a/dev/publish_release.sh
+++ b/dev/publish_release.sh
@@ -1,0 +1,149 @@
+#!/bin/bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The script helps publishing release to mavene.
+# You need to specify release version and tag name.
+#
+# Here are some helpful documents for the release.
+# http://www.apache.org/dev/release.html
+# http://www.apache.org/dev/release-publishing
+# http://www.apache.org/dev/release-signing.html
+# http://www.apache.org/dev/publishing-maven-artifacts.html
+
+if [[ -z "${SHASUM}" ]]; then
+  SHASUM="/usr/bin/shasum -a 1"
+fi
+
+if [[ -z "${WORKING_DIR}" ]]; then
+  WORKING_DIR=/tmp/zeppelin-release
+fi
+
+if [ $# -ne 2 ]; then
+  echo "usage) $0 [Release version] [Tag]"
+  echo "   ex. $0 0.6.0 v0.6.0"
+  exit 1
+fi
+
+for var in GPG_PASSPHRASE ASF_USERID ASF_PASSWORD
+do
+  if [[ -z "${!var}" ]]; then
+    echo "You need ${var} variable set"
+    exit 1
+  fi
+done
+
+export MAVEN_OPTS="-Xmx2g -XX:MaxPermSize=512m"
+
+RELEASE_VERSION="${1}"
+GIT_TAG="${2}"
+
+PUBLISH_PROFILES="-Pspark-1.6 -Phadoop-2.4 -Pyarn -Ppyspark -Psparkr -Pr"
+PROJECT_OPTIONS="-pl !zeppelin-server,!zeppelin-distribution"
+NEXUS_STAGING="https://repository.apache.org/service/local/staging"
+NEXUS_PROFILE="153446d1ac37c4"
+
+if [[ -d "${WORKING_DIR}" ]]; then
+  echo "Dir ${WORKING_DIR} already exists"
+  exit 1
+fi
+
+mkdir ${WORKING_DIR}
+
+echo "Cloning the source"
+# clone source
+git clone https://git-wip-us.apache.org/repos/asf/zeppelin.git ${WORKING_DIR}/zeppelin
+
+if [[ $? -ne 0 ]]; then
+  echo "Can not clone source repository"
+  exit 1
+fi
+
+cd ${WORKING_DIR}/zeppelin
+git checkout ${GIT_TAG}
+echo "Checked out ${GIT_TAG}"
+
+# remove unnecessary files
+rm ${WORKING_DIR}/zeppelin/.gitignore
+rm -rf ${WORKING_DIR}/zeppelin/.git
+
+function publish_to_maven() {
+  cd ${WORKING_DIR}/zeppelin
+
+  # Force release version
+  mvn versions:set -DnewVersion=$RELEASE_VERSION
+
+  # Using Nexus API documented here:
+  # https://support.sonatype.com/hc/en-us/articles/213465868-Uploading-to-a-Staging-Repository-via-REST-API
+  echo "Creating Nexus staging repository"
+  repo_request="<promoteRequest><data><description>Apache Zeppelin $RELEASE_VERSION</description></data></promoteRequest>"
+  out=$(curl -X POST -d "$repo_request" -u $ASF_USERID:$ASF_PASSWORD \
+    -H "Content-Type:application/xml" -v \
+    $NEXUS_STAGING/profiles/$NEXUS_PROFILE/start)
+  staged_repo_id=$(echo $out | sed -e "s/.*\(orgapachezeppelin-[0-9]\{4\}\).*/\1/")
+  echo "Created Nexus staging repository: $staged_repo_id"
+
+  tmp_repo=$(mktemp -d /tmp/zeppelin-repo-XXXXX)
+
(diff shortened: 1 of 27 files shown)


Leemoonsoo:
CI failiure is unrelated. Looks good to me

dev/publish_release.sh
```diff
@@ -0,0 +1,149 @@
...
+
+function publish_to_maven() {
+  cd ${WORKING_DIR}/zeppelin
+
+  # Force release version
+  mvn versions:set -DnewVersion=$RELEASE_VERSION
+
+  # Using Nexus API documented here:
+  # https://support.sonatype.com/hc/en-us/articles/213465868-Uploading-to-a-Staging-Repository-via-REST-API
+  echo "Creating Nexus staging repository"
```
bzz: Shall this be extracted to a `function`, how do you think?

dev/publish_release.sh
```diff
@@ -0,0 +1,149 @@
...
+  echo "Creating hash and signature files"
+  for file in $(find . -type f)
+  do
+    echo $GPG_PASSPHRASE | gpg --passphrase-fd 0 --output $file.asc \
+      --detach-sig --armor $file;
+    md5 -q $file > $file.md5
+    $SHASUM $file | cut -f1 -d' ' > $file.sha1
+  done
+
+  nexus_upload=$NEXUS_STAGING/deployByRepositoryId/$staged_repo_id
```
bzz: We might want to follow the [conventions from our styleguide](https://github.com/apache/zeppelin/blob/master/CONTRIBUTING.md#code-convention) in particular on [Quoting](https://google.github.io/styleguide/shell.xml#Quoting) here

bzz:
Looks great, I think there are just few things left:
 - rebased on latest master
 - check that code conforms the [conventions from our style guide](https://github.com/apache/zeppelin/blob/master/CONTRIBUTING.md#code-convention)


@minahlee if we want to remove `maven release plugin`, could you also point to the place that does automate the version switch in POM files (as AFAIK that is the only purpose we used that plugin so far)

dev/publish_release.sh
minahlee: If we make it as function, this function should return value `staged_repo_id` which is not supported in bash. I know there is tricky way to handle it, but not sure how beneficial it would be since this will be used only once in a script. What do you think?

dev/publish_release.sh
```diff
@@ -0,0 +1,146 @@
...
+
+# remove unnecessary files
+rm "${WORKING_DIR}/zeppelin/.gitignore"
+rm -rf "${WORKING_DIR}/zeppelin/.git"
+
+function publish_to_maven() {
+  cd "${WORKING_DIR}/zeppelin"
+
+  # Force release version
+  mvn versions:set -DnewVersion="${RELEASE_VERSION}"
```
minahlee: @bzz this is where pom.xml is changed :)

minahlee:
@bzz Thank you for the review, I handled style in 21ca610. Please let me know if you have any other suggestions :)

dev/publish_release.sh
bzz: Thanks for pointing it out!

echarles:
why spark is not published. is it related to the assembly?

Btw, spark 2 doesn't use assembly any more and favor dependencies in separate jars. See https://issues.apache.org/jira/browse/SPARK-11157 for more details.

dev/publish_release.sh
```diff
@@ -0,0 +1,146 @@
...
+  done
+
+  nexus_upload="${NEXUS_STAGING}/deployByRepositoryId/${staged_repo_id}"
+  echo "Uplading files to ${nexus_upload}"
+  for file in $(find . -type f); do
+    # strip leading ./
+    file_short="$(echo "${file}" | sed -e 's/\.\///')"
+    dest_url="${nexus_upload}/org/apache/zeppelin/$file_short"
+    echo "  Uploading ${file_short}"
+    curl -u "${ASF_USERID}:${ASF_PASSWORD}" --upload-file "${file_short}" "${dest_url}"
```
bzz: As we do not use -e flag for this scripts, shall not we check return of, at least, network-related commands? 

This way we could make sure that if it have failed due to networking reasons, we print to the user meaningful error conditions and let him then re-run the script?

zeppelin-distribution/pom.xml
```diff
@@ -72,24 +72,6 @@
           </descriptors>
         </configuration>
       </plugin>
-
-      <plugin>
-        <!--assembly does all the job here-->
-        <artifactId>maven-dependency-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
```
bzz: You think it's not required any more?

bzz:
Great, I think if checking error conditions could be added to network-related commands - it would be even more awesome, as this is quite common failure pattern.

Other then that, looks great to me and thank you for automating this.

minahlee:
@echarles I think my description caused confusion. Spark will be included in release since most of Zeppelin user is based on Spark. The modules I wrote in description is skipped modules for publish and Spark always hasn't been skipped from release.

echarles:
Didn't look at the code... just saw in the description (and @minahlee  just confirmed it would be published).

Thinking more about this PR: 
- Why the deploy and release plugins were not working? Using scrips instead of existing plugins is not ideal to me.
- I understand distribution should not be published, but server should be. It contains the repo components that can be used elsewhere. Apart from the fact we could have a separate module for the repo, all components (excpect distribution) should be published. That's a rule of thumb and if we don't follow it, someone is gonna come sooner or later and ask for the missing bits.

bzz:
Thank you for the feedback @echarles !

I think it should be fine to add missing bits, as soon as they are identified and there is somebody with actual use case for it later on. Would you be willing to log it as a JIRA issue?

As person who did release management for Zeppelin before - I had no troubles with release plugin, although calling it multiple times with different arguments, instead of single script was quite error prone.

If a manager for a current release though wants to try another solution and provides a robust one (as in this PR) - it should be fine, especially given that we can always get back to the plugins if desired.

echarles:
Good to hear the release plugin is working fine.

The release manager is responsible for the tools he wants. Fine to me if it is a script or a plugin, but at the end the artifacts must find their way to maven central.

For this PR, is it possible to ensure `zeppelin-server` will be also published?

minahlee:
@echarles 
- This script is written exactly because of the reason that @bzz explained above.

> calling it multiple times with different arguments, instead of single script was quite error prone.

- AFAIK, repo is under zeppelin-zengine module, not zeppelin-server. But if there is any other use case for zeppelin-server other than notebook repo I don't see the reason not to publish. If you come up with other use case, please let me know

echarles:
@minahlee thx for this precision and context.
True, I  just realized after sending the comment that repo is not part of zeppelin-server.
However, I need zeppelin-server in my project (I kind off embed the web and websocket server) - Publishing all artifcats is also in my understanding a common practice. WDYT. Is there any issue you face to publish the server?

minahlee:
@echarles If there are such needs, nothing stops us from publishing it. I will make change accordingly. Thank you for sharing your situation.

bzz:
@minahlee Great! 

I think it should be ready to merge as soon as issues highlighted above 

  a. failing networking `curl` 
  b. build configuration duplication

are addressed

minahlee:
@bzz Sorry for late response, I am working on it and going to push commits by today. Thanks for heads up :)

bzz:
Looks awesome to me, thank you @minahlee !

2 minor things are a bit confusing:
 - the purpose of `dev/publish_release.sh.hehe` file 
 - some artifacts (namely, integration for Geode and Scalding) we publish in maven, but AFAIK they are not part of the convenience binary for the release

  ```
publish: -Pspark-1.6 -Phadoop-2.4 -Pyarn -Ppyspark -Psparkr -Pr -Pgeode -Pscalding
release: -Pspark-1.6 -Phadoop-2.4 -Pyarn -Ppyspark -Psparkr -Pr
  ```

minahlee:
@bzz Thanks for the review. I removed mistakenly pushed file.
I included geode, scalding interpreter in maven publish script so that geode, scalding users can take advantage of #1042. 
To be honest, I don't know why these two interpreters were not included in last release. Can someone explain?

Leemoonsoo:
@minahlee Here's related issues why geode and scalding is not in maven publish script.

Geode - https://issues.apache.org/jira/browse/ZEPPELIN-375
Scalding - https://issues.apache.org/jira/browse/ZEPPELIN-972

Please take a look. To include them, bin_license/LICENSE file need to be updated with their transitive dependencies(geode, scalding interpreter), and remove 3rd party repository (scalding interpreter)

minahlee:
@Leemoonsoo Thanks for pointing it out! I reverted commit of adding geode, scalding

bzz:
Got it, thank for reminding! Let's merge this guy then

minahlee:

PR #1005, opened 2016-06-13
ZEPPELIN-995 Change scheduler for JDBC interpreter to use concurrent execution

jongyoul:
### What is this PR for?
Changed scheduler from FIFO to Parallels in JdbcInterpreter. This is a default behaviour of HiveInterpreter. When we merge all JDBC-like interpreter into JDBC, we need to change default behaviour of JdbcInterpreter.

### What type of PR is it?
[Feature]

### Todos
* [x] - Changed scheduler

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-995

You can run multiple queries simultaneously.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -98,6 +98,9 @@ public class JDBCInterpreter extends Interpreter {
 
   static final String EMPTY_COLUMN_VALUE = "";
 
+  private final String CONCURRENT_EXECUTION_KEY = "zeppelin.jdbc.concurrent.use";
+  private final String CONCURRENT_EXECUTION_COUNT = "zeppelin.jdbc.concurrent.max_connection";
+
   private final HashMap<String, Properties> propertiesMap;
   private final Map<String, Statement> paragraphIdStatementMap;
 
@@ -433,8 +436,10 @@ public class JDBCInterpreter extends Interpreter {
 
   @Override
   public Scheduler getScheduler() {
-    return SchedulerFactory.singleton().createOrGetParallelScheduler(
-        JDBCInterpreter.class.getName() + this.hashCode(), 10);
+    String schedulerName = JDBCInterpreter.class.getName() + this.hashCode();
+    return isConcurrentExecution() ?
+            SchedulerFactory.singleton().createOrGetParallelScheduler(schedulerName, 10)
+            : SchedulerFactory.singleton().createOrGetFIFOScheduler(schedulerName);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


prabhjyotsingh:
@jongyoul Thank you for taking care of this. I agree this should be ParallelScheduler. 👍 
LGTM.

Leemoonsoo:
How about make it configurable and set parallel scheduler by default?

Some user might want to run query in parallel, but some might want to avoid run queries in parallel.
Scheduler of SparkSql interpreter can be configurable through 'zeppelin.spark.concurrentSQL'.

jongyoul:
@Leemoonsoo I also agree to makes this configurable. I'll follow up this with another PR. I also think we need to replace using getScheduler to another way.

bzz:
Looks great, thank you for prompt update!

@jongyoul is there a reason to make it configurable from another PR instead of this one? Just curious

jongyoul:
@bzz This is because this is related to zeppelin-server. AFAIK, getScheduler is used by zeppelin-server, not zeppelin-interpreter, thus we cannot make it configurable easily. For now, if you make it configurable, we should restart zeppelin-daemon whenever we change the value. I think it's not a good way.

Leemoonsoo:
I was thinking just creating multiple interpreter settings if user want to use different scheduler. Then just returning appropriate scheduler from getScheduler() by configuration would be enough. Isn't it?

jongyoul:
@Leemoonsoo I've missed something. I've known it was not configurable dynamically. Let me check.

Leemoonsoo:
@jongyoul Right, it's not dynamically configurable. Interpreter need to be restarted to reconfigure. 

I think it's possibly related to ZEPPELIN-999 and it's long term plan. Let's say a user want to use jdbc interpreter for hive and mysql. Currently, 

A. user can create single interpreter setting and create two configuration for both hive and mysql connection. And select connection via %jdbc(hive) or %jdbc(mysql).
B. or user can create two interpreter setting and each interpreter setting have configuration for hive and mysql connection respectively. In this case, select jdbc connection limited and annoying, because a notebook can not use two or more same type of interpreter settings at the same time. So user have to constantly bind/unbind interpreter setting to switch hive <-> mysql.

If our long term plan is generalize ZEPPELIN-999 and allow user use interpreter setting alias for interpreter selection, like %hive, %mysql, then all disadvantages of approach **B** will be eliminated.

Managing only single jdbc connection in single interpreter setting, with generalized ZEPPELIN-999 will give advantages, like leverage [interpreter authorization](https://issues.apache.org/jira/browse/ZEPPELIN-945). And in this case, i was thinking simple returning appropriate scheduler from getScheduler() by configuration would be enough.

jongyoul:
@Leemoonsoo I've totally understood your idea and agree with you. I'll patch for it and push it again

jongyoul:
Merging it into master and branch-0.6

PR #1007, opened 2016-06-13
Update and refactor NotebookRepo versioning API

khalidhuseynov:
### What is this PR for?
This is firstly to refactor API for versioning and keep everthing inside of one interface (NotebookRepo) instead of two different interfaces (NotebookRepoVersioned). Secondly, there're modifications to existing versioning api, with considerations of future complete implementation of versioning. Note that this PR doesn't implement all suggested interfaces, but lays foundation for their implementation.

### What type of PR is it?
Improvement && Refactoring

### Todos
* [x] - move versioning api (get, history) from NotebookRepoVersioned to NotebookRepo
* [x] - refactor and naming changes
* [x] - modify checkpoint api (add return value) and modify NotebookRepoSync to deal with it

Basically it doesn't add new functionality, so the only requirement is for tests to pass.

### Questions:
* Is there breaking changes for older versions? not breaking, but some api changes


khalidhuseynov:
ready for review

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/AzureNotebookRepo.java
```diff
@@ -205,8 +205,21 @@ public void close() {
   }
 
   @Override
-  public void checkpoint(String noteId, String checkPointName) throws IOException {
+  public Rev checkpoint(String noteId, String checkpointMsg) throws IOException {
```
bzz: How do you think, is there a reason why we do not want a readable and concise class name like `Revision` to be used here? Or something like `Checkpoint` as the same concept is referred in API

bzz:
Great improvement! 

Could you please explain the rationale behind removing `NotebookRepoVersioned` and making a lot of boilerplate methods returning `null` in all other notebook storages that do not support versioning?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/AzureNotebookRepo.java
khalidhuseynov: good point, so I was expecting to revise and improve naming during PR review. Actually originally it was as `Rev` in the `NotebookRepoVersioned` class, and although I modified the class itself but name wasn't changed. I agree that having full name as `Revision` for that class makes more sense :+1:

khalidhuseynov:
the reason is quite simple, that's to have only one interface to be implemented (instead of choosing from two, given that documentation doesn't mention `NotebookRepoVersioned` so far). Also I believe that would encourage to implement versioning on more storages. @bzz what do you think?

bzz:
Thank you for addressing feedback promptly.

Well, in my oppinion, if by "encourage more versioned implementation" you mean having this code duplicated 5 times around the code base
```java
  @Override
  public Note get(String noteId, Revision rev) throws IOException {
    // Auto-generated method stub
    return null;
  }
```
it looks like not very elegant solution. 

It looks more meaningful and elegant to having 2 separate interfaces (types) documented, 1 for non-versioned backends and 1 for versioned ones to allow authors to choose from.

Otherwise, if we extend the suggested logic - why do not just to type everything as `Object` and keep only 1 interface for everything? It kind of defeats the benefits of having a type system.

It's not a strong opinion though, and if after this arguments, you still feel that having single one is better - please let me know.

Other that this issues - it looks great to me.

khalidhuseynov:
yeah, I agree that having two interfaces is more elegant, but let's see some more detailed downsides. if you see [here](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java#L49) repoSync consists of `NotebookRepo` classes, thus when having any versioning related operation we will need to check  everytime if repo instance is `instanceof NotebooRepoVersioned` (or some other flag) and then do cast and call corresponding method, which is an overhead @bzz @Leemoonsoo WDYT, any other opinions? 

Moreover, `NotebookRepoVersioned` haven't been really integrated into the workflow so far, so it's good time to think on either integrating or removing it to move on. And the only existing versioning related `checkpoint` method was in `NotebookRepo` instead of `NotebookRepoVersioned`.

bzz:
Good point, if asked, I would say that having `NotebookRepoSync implements NotebookRepo` is not an elegant design either - NotebookRepoSync is not a NoteboksRepo! :)

It's up to you after all, I see that you feel it's the best way to integrate for now and is as good as it can be - so let's keep it like this. It can always be refactored later on.

Looks good to me, will merge if there is no other discussion.

khalidhuseynov:
@bzz thanks for the review, I'll be glad to hear for the elegant solutions of the mentioned issues, even can help with implementation :)

PR #1008, opened 2016-06-13
[ZEPPELIN-997] Export data to csv

minahlee:
### What is this PR for?
Add csv download from front-end leveraging #714

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-997](https://issues.apache.org/jira/browse/ZEPPELIN-997)

**Before**
screen shot 2016-06-13 at 4 55 02 pm

**After**
screen shot 2016-06-14 at 12 42 01 am

screen shot 2016-06-14 at 12 41 49 am

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
@@ -41,11 +41,17 @@ limitations under the License.
             ng-click="setGraphMode('scatterChart', true)"><i class="cf cf-scatter-chart"></i>
     </button>
   </div>
-  <span>
-    <button type="button" class="btn btn-default btn-sm" style="margin-left:10px"
-            tooltip="Download Data as TSV" tooltip-placement="bottom"
-            ng-click="exportToTSV()"><i class="fa fa-download"></i>
+  <span class="dropdown">
+    <button type="button" class="btn btn-default btn-sm"
+            style="margin-left:10px"
+            data-toggle="dropdown">
+      <i class="fa fa-download"></i>
+      <span class="caret"></span>
     </button>
+    <ul class="dropdown-menu" style="margin: 10px; min-width: 70px;">
+      <li ng-click="exportToDSV('\t')"><a>TSV</a></li>
+      <li ng-click="exportToDSV(',')"><a>CSV</a></li>
+    </ul>
   </span>
   <span ng-if="getGraphMode()!='table'"
         style="margin-left:5px; cursor:pointer; display: inline-block; vertical-align:top; position: relative; line-height:30px;">
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
@@ -41,16 +41,20 @@ limitations under the License.
             ng-click="setGraphMode('scatterChart', true)"><i class="cf cf-scatter-chart"></i>
     </button>
   </div>
-  <span class="dropdown">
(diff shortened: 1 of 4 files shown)


corneadoug:
Could we use a separator between the button and the carret?
http://getbootstrap.com/components/#btn-dropdowns-split

Default download could be CSV

minahlee:
@corneadoug I just changed, please check and let me know your opinion.

zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
```diff
@@ -41,11 +41,21 @@
             ng-click="setGraphMode('scatterChart', true)"><i class="cf cf-scatter-chart"></i>
     </button>
   </div>
-  <span>
-    <button type="button" class="btn btn-default btn-sm" style="margin-left:10px"
-            tooltip="Download Data as TSV" tooltip-placement="bottom"
```
corneadoug: Could you restore the tooltip on the main button? (with CSV instead of TSV)

zeppelin-web/src/app/notebook/paragraph/paragraph-chart-selector.html
```diff
@@ -41,11 +41,21 @@
...
-            ng-click="exportToTSV()"><i class="fa fa-download"></i>
+  <span class="btn-group">
+    <button type="button" class="btn btn-default btn-sm"
+            style="margin-left:10px"
+            ng-click="exportToDSV(',')">
+      <i class="fa fa-download"></i>
+    </button>
+    <button type="button" class="btn btn-default btn-sm dropdown-toggle"
```
corneadoug: The Button size change when the tooltip comes out because of some bootstrap CSS.

you might need to add a new class to this element, something like that would do:

```
.caretBtn {
    padding-right: 4px !important;
    padding-left: 4px !important;
    width: 20px;
}
```

It will also reduce the width of the caret button
screen shot 2016-06-14 at 4 28 11 pm

corneadoug:
Unrelated CI failure (timeout)
LGTM

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -2145,21 +2145,27 @@ angular.module('zeppelinWebApp')
...
-        tsvRow += row[index].value + '\t';
+        dsvRow += row[index].value + delimiter;
       }
-      tsv += tsvRow.substring(0, tsvRow.length - 1) + '\n';
+      dsv += dsvRow.substring(0, dsvRow.length - 1) + '\n';
     }
-    SaveAsService.SaveAs(tsv, 'data', 'tsv');
+    var extension = '';
+    if (delimiter === '\t') {
+      extension = 'tsv';
```
felixcheung: this is likely out of scope for this change, for csv (likely tsv), strictly speaking we would need to escape any `,` in the data:
https://en.wikipedia.org/wiki/Comma-separated_values#General_functionality
otherwise when the data is read it will be misaligned.

minahlee:
@felixcheung Thanks for coming up with the issue, let me handle it in another PR. I created [new issue](https://issues.apache.org/jira/browse/ZEPPELIN-1001) for that.

Leemoonsoo:
LGTM

corneadoug:
Merging in Master if there is no more discussions

qinzl1:
if i can export more than  1000 rows

corneadoug:
@qinzl1 It might be independent from the download button.
Queries are usually limited to 1000 rows (it is set in the interpreter's setting).
Can you check that?

PR #1012, opened 2016-06-14
ZEPPELIN-999 Support alias for JDBC properties

jongyoul:
### What is this PR for?
In case of using JdbcInterpreter, you should use %jdbc(prefix) if you set multiple configurations. This PR makes you use %prefix only. 

### What type of PR is it?
[Improvement]

### Todos
* [x] - Change %prefix to %jdbc(prefix) during running paragraph

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-999

screen shot 2016-06-15 at 12 42 32 am
screen shot 2016-06-15 at 12 42 49 am

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -71,7 +71,7 @@ public class ZeppelinConfiguration extends XMLConfiguration {
    *url = ZeppelinConfiguration.class.getResource(ZEPPELIN_SITE_XML);
    * @throws ConfigurationException
    */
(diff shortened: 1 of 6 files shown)


felixcheung:
Nice! Can this be generalized to not just jdbc?

jongyoul:
@felixcheung I also have a plan to generalize this feature after 0.6.0 releases. :-)

bzz:
Thank you @jongyoul, 👍  for having tests!
Improving this further by generalizing to other interpreters after a release sounds like a very good plan.

Looks great to me.

jongyoul:
@bzz I'll merge it into master and branch-0.6

jongyoul:
I also made a PR as a generalized version. https://issues.apache.org/jira/browse/ZEPPELIN-1012

PR #1013, opened 2016-06-15
[ZEPPELIN-1002] Move configuration menus under dropdown

minahlee:
## What is this PR for?
- Move configuration menus under dropdown menu
- Change dropdown menu style
- Change `Login` button style

### What type of PR is it?
Improvement

### Todos
- [x] Fix selenium test

### What is the Jira issue?
[ZEPPELIN-1002](https://issues.apache.org/jira/browse/ZEPPELIN-1002)

#### As anonymous
Before
screen shot 2016-06-14 at 9 43 35 pm

After
screen shot 2016-06-16 at 10 15 21 am

#### With shiro authc
Before
screen shot 2016-06-16 at 10 18 45 am
screen shot 2016-06-16 at 10 21 52 am

After
screen shot 2016-06-16 at 10 17 46 am
screen shot 2016-06-20 at 11 30 04 am

screen shot 2016-06-16 at 10 20 15 am

### Questions:
* Does this needs documentation? It needs image updates

### DIFF
--- a/zeppelin-web/src/components/navbar/navbar.html
+++ b/zeppelin-web/src/components/navbar/navbar.html
@@ -81,18 +81,26 @@ limitations under the License.
             </div>
           </form>
         <li class="nav-component">
+          <i ng-if="navbar.connected" class="fa fa-circle server-connected"
+             tooltip="WebSocket Connected" tooltip-placement="bottom"></i>
+          <i ng-if="!navbar.connected" class="fa fa-circle server-disconnected"
+             tooltip="WebSocket Disconnected" tooltip-placement="bottom"></i>
+        </li>
+        <li>
           <div class="dropdown">
-            <button class="btn dropdown-toggle username" type="button" data-toggle="dropdown" style="margin-right: 5px;"
-                    ng-class="{'nav-status-btn': !ticket.principal || ticket.principal === 'anonymous',
-                             'nav-btn': ticket.principal && ticket.principal !== 'anonymous'}">
-              <i class="fa fa-circle" ng-class="{'server-connected':navbar.connected, 'server-disconnected':!navbar.connected}"></i>
-              <span ng-show="navbar.connected" ng-if="!ticket.principal || ticket.principal === 'anonymous'">Connected</span>
-              <span ng-show="navbar.connected" ng-if="ticket.principal && ticket.principal !== 'anonymous'">{{ticket.principal}}</span>
-              <span ng-show="!navbar.connected">Disconnected</span>
-              <span class="caret" ng-if="ticket.principal && ticket.principal !== 'anonymous'"></span>
+            <button ng-if="ticket" class="nav-btn dropdown-toggle" type="button" data-toggle="dropdown" style="margin:11px 5px 0 0;">
+              <span class="username">{{ticket.principal}}</span>
+              <span class="caret" style="margin-bottom: 8px"></span>
             </button>
-            <ul class="dropdown-menu" ng-if="ticket.principal && ticket.principal !== 'anonymous'">
-              <li><a ng-click="logout()">Logout</a></li>
+            <span ng-if="!ticket" style="margin: 5px;"></span>
+            <ul class="dropdown-menu">
+              <li><a href="" data-toggle="modal" data-target="#aboutModal">About Zeppelin</a>
+              <li role="separator" style="margin: 5px 0;" class="divider"></li>
+              <li><a href="#/interpreter">Interpreter</a></li>
+              <li><a href="#/credential">Credential</a></li>
+              <li><a href="#/configuration">Configuration</a></li>
+              <li ng-if="ticket.principal && ticket.principal !== 'anonymous'" role="separator" style="margin: 5px 0;" class="divider"></li>
+              <li ng-if="ticket.principal && ticket.principal !== 'anonymous'"><a ng-click="logout()">Logout</a></li>
             </ul>
           </div>
         </li>
@@ -100,20 +108,6 @@ limitations under the License.
           <button class="btn nav-login-btn" data-toggle="modal" data-target="#loginModal"
                   ng-click="showLoginWindow()">Login
           </button>
-        </li>
-        <li class="nav-component" ng-if="ticket">
-          <div class="dropdown">
-            <button class="btn nav-btn dropdown-toggle" type="button" data-toggle="dropdown" style="margin-right: 5px;">
-              <i class="fa fa-cog"></i>
-            </button>
-            <ul class="dropdown-menu">
-              <li><a href="#/interpreter">Interpreter</a></li>
-              <li><a href="#/credential">Credential</a></li>
-              <li><a href="#/configuration">Configuration</a></li>
-              <li role="separator" class="divider"></li>
-              <li><a href="" data-toggle="modal" data-target="#aboutModal">About</a>
-            </ul>
-          </div>
         </li>
       </ul>
     </div>
--- a/zeppelin-web/src/components/navbar/navbar.html
+++ b/zeppelin-web/src/components/navbar/navbar.html
@@ -86,16 +86,12 @@ limitations under the License.
                     ng-class="{'nav-status-btn': !ticket.principal || ticket.principal === 'anonymous',
                              'nav-btn': ticket.principal && ticket.principal !== 'anonymous'}">
               <i class="fa fa-circle" ng-class="{'server-connected':navbar.connected, 'server-disconnected':!navbar.connected}"></i>
-              <span ng-show="navbar.connected" ng-if="ticket.principal === 'anonymous'">Connected</span>
-              <span ng-show="navbar.connected" ng-if="ticket.principal !== 'anonymous'">{{truncatedUsername}}</span>
+              <span ng-show="navbar.connected" ng-if="!ticket.principal || ticket.principal === 'anonymous'">Connected</span>
+              <span ng-show="navbar.connected" ng-if="ticket.principal && ticket.principal !== 'anonymous'">{{ticket.principal}}</span>
               <span ng-show="!navbar.connected">Disconnected</span>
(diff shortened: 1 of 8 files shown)


corneadoug:
I think this kind of menu should be under a setting icon instead.
And keep the admin dropdown element only for logout if there is auth

Leemoonsoo:
Thanks @minahlee for the update. I think it's less confusing and more organized way to manage menubar. Do you think it'll be merged into branch-0.6, too?

On documentation, I think there're some screenshots related to this update.
for example [this](http://zeppelin.incubator.apache.org/docs/0.6.0-SNAPSHOT/quickstart/explorezeppelinui.html#2-interpreter), [this](http://zeppelin.incubator.apache.org/docs/0.6.0-SNAPSHOT/quickstart/explorezeppelinui.html#3-configuration), and so on. Could you handle them here or create an issue for it?

minahlee:
I think it is common practice to have settings under profile dropdown menu.

### Examples
**Github**
screen shot 2016-06-15 at 2 01 55 pm

**Trello**
screen shot 2016-06-15 at 2 04 30 pm


### Candidates for managing setting menus
which one do you prefer among below two?
- Have separator dividing settings menu
screen shot 2016-06-15 at 2 42 18 pm

- Have separate icon for settings

   screen shot 2016-06-15 at 2 56 08 pm  screen shot 2016-06-15 at 2 54 18 pm


Plus, I am going to put username under menu to take care of long user name. This is current UI which seems not pleasant experience for users:

screen shot 2016-06-15 at 3 09 38 pm
screen shot 2016-06-15 at 3 06 26 pm


@Leemoonsoo yes I think it's possible to push this into branch-0.6 if we decided how do we want to display it in time.

bzz:
Just FYI - please mind that second menu should play nice with mobile layout. 

As first one is collapsed on mobile, even right now with just a search search - it sometimes is not enough space there.

corneadoug:
@bzz actually, last time I tested, the about dropdown was doing great on mobile

@minahlee I still prefer separated one, mainly because you won't always have a user, and
having a dropdown under Connected/Disconnected is misleading and doesn't give much indications.
The thing about settings is that it describes well what you will find inside: settings

In the case of website separating them, we can list JIRA and Gmail:
screen shot 2016-06-16 at 11 11 40 am
screen shot 2016-06-16 at 11 12 03 am

For the content for both menu, it would be best to have a better separation in their role: What is under user (if there is one), should be related to the user: Logout. And what is under setting should be related to the instance/software: Interpreter, Credentials, Configuration, About

bzz:
> @bzz actually, last time I tested, the about dropdown was doing great on mobile

Have you already tested 2 proposed dropdowns on mobile? If possible, plz share the screenshots

minahlee:
@corneadoug Having separate menu for the anonymous user totally makes sense.
Then the next question would be where do we want to place it?
Next to Notebook? - will give less confusion who is used to current layout, but there won't be big problem once they recognize it
On the right side of nav bar? - will take less space since it will use gear icon instead of `Settings` text.

I have no special preference on setting menu location.
@corneadoug, @bzz, @Leemoonsoo what is your opinion here?

Leemoonsoo:
'Credentials' is per user. So if we want to divide 'user' and 'setting', it'll be

user - Credentials, Logout
setting - Interpreter, Configuration, About

Or how about simply create a button,

| o Admin (gear icon) |
-------------------

and put every items under this single button?
Either way, i think it's better than now.

corneadoug:
@bzz Here is a screenshot of example showing a second dropdown, technically its 3 dropdown if you count the Notebooks one

screen shot 2016-06-16 at 3 32 08 pm

@mina I think a setting icon would be better as it doesn't take much space in the navbar
Style in mobile can always be modified to be more verbose in different PR. (I already found a few bugs)

@Leemoonsoo What is the Credential menu btw?

Leemoonsoo:
@corneadoug Credential menu, see https://issues.apache.org/jira/browse/ZEPPELIN-828

corneadoug:
Maybe we can even rename that menu: 'My credentials' then :)

Leemoonsoo:
If 'Credential' menu is under 'User', i think it already imply '`My` credential'. isn't it?

minahlee:
One question, if `Confidential` is per user menu, should it be shown when only user logged in?

Leemoonsoo:
I think 'Credential' still useful without login. It just becomes (anonymous user's) credential.

cloverhearts:
Confidential default value is "anonymous"

bzz:
@corneadoug thanks for posting screen! It looks reasonable

> and put every items under this single button?
> Either way, i think it's better than now.

True, as soon as it works on mobile as well and Integration tests are updated there is not that much difference and we can change it again later.

minahlee:
Thanks for the feedback guys! I've just pushed some commits and updated description with new screenshot.

corneadoug:
For the username inside the dropdown, is it not possible to have the long username tooltip disappear if the menu is out? Except for that LGTM

minahlee:
What about showing user name under dropdown only when username length exceeds `MAX_USERNAME_LENGTH`?

corneadoug:
@minahlee Actually, the long name would normally be truncated by CSS instead of javascript, so it wouldn't be good. (Since it was done elsewhere, no need to fix it here)

Is it necessary to show long name though? Maybe this can be handled in a different PR, so that we can think of other way of doing it, and not block this one.

minahlee:
I also don't feel the necessity of showing full user name since logged user would know his own username. Let me remove username under dropdown then.

bzz:
Now we have 3 drop-down menus :) Are you guys sure that we can not get away with just 2 of them i.e 'logout' instead of 'about' in under the gear icon?

Also, for the sake of consistency, should not it be "Anonymous" instead of "Connected" near the green\red circle when the user is not logged in?

corneadoug:
@bzz In the case of Anonymous, I would really prefer keeping anything to do with User Management out of Zeppelin if it's not using Auth. Since there is one single user, there is no need for Anonymous anywhere.

bzz:
Ok, thanks for explanation, this was not a strong opinion, just a question. Sounds good to me.
 @minahlee  what do you think?

minahlee:
I am ok with either approach, seems like @corneadoug has strong opinion on having two separate menu so I will leave it as it is. I will need to fix selenium test to make it to be merged though.

minahlee:
Updated selenium test and CI passes

zeppelin-web/src/components/navbar/navbar.html
```diff
@@ -89,43 +80,41 @@
               </span>
             </div>
           </form>
+        <li class="nav-component">
+          <div class="dropdown">
+            <button class="btn dropdown-toggle" type="button" data-toggle="dropdown" style="margin-right: 5px;"
```
corneadoug: @minahlee You removed the truncated text, but didn't implement it in css.
Could you add a class with this inside to this element?
```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
max-width: 180px;
```

corneadoug:
LGTM

Leemoonsoo:
CI failure is irrelevant to this change.
Tested and it looks good to me. +1 for merge it into master and branch-0.6

minahlee:
Merging it into master and branch-0.6

khalidhuseynov:
I've just went through the conversation here, and i think that having `userName` instead of `connected` status in authenticated mode with additional `logout` menu is quite misleading. first of all `connected` and `disconnected` serve different purpose and not related to logged in user, so they have to be separated. Secondly, i would be expecting most of the `action` features like `logout` under main dropdown menu, and wouldn't expect to click on my name to see `logout` button. So maybe need to rethink it a bit :)

minahlee:
@khalidhuseynov Thanks for the feedback.
I totally agree that showing `Connected` message for non-logged user can be confusing.
I made several changes based on your opinion.
- show user name as `anonymous` when shiro is configured as anonymous mode.(suggested by @bzz in [previous comment](https://github.com/apache/zeppelin/pull/1013#issuecomment-227034217))
- move websocket connected/disconnected text to tooltip.
- merge two dropdown menu to one. I am also skeptical about having one more dropdown menu only for logout action. I believe with caret icon, users will easily know where to click if they wants to do some action like change settings, logout, etc.

**anonymous**
screen shot 2016-06-22 at 7 17 45 pm
screen shot 2016-06-22 at 7 17 21 pm

**authc**
screen shot 2016-06-22 at 7 16 29 pm
screen shot 2016-06-22 at 7 16 34 pm

AhyoungRyu:
@khalidhuseynov Right. I totally agree with you. 
@minahlee It's more intuitive than before 👍

corneadoug:
I agree with @khalidhuseynov that having Connected/Disconnected being replace by the username isn't great.

However I still think that showing a username cf: 'anonymous' for an app without any Auth system activated doesn't make any sense.

Regarding not having two dropdowns, I don't mind grouping it in one, or having a separate icon next to the username for logout.

Leemoonsoo:
I think showing 'anonymous' without auth system activated still make sense.
All authorization features (i.e. notebook permission, credential) works normally as user name 'anonymous' when auth system is not activated. 

So, showing 'anonymous' is actually helping, i think.

khalidhuseynov:
@minahlee thanks for quick mockup! looks good and as @corneadoug said we can omit `anonymous` when no one logged in, and only show username if authentication is enabled. One more thing is about tooltip, not sure whether we need to give details of `WebSocket` there and whether it's going to be helpful, that's just an opinion :)

minahlee:
Thanks for the feedback all! 
@Leemoonsoo After your explanation, it makes sense to display `anonymous`
@khalidhuseynov I added `WebSocket` because logged user can misunderstand that he is connected/disconnected to session.
Integration test passes, I think I will merge this both to master, branch-0.6 if there is no more discussion

khalidhuseynov:
LGTM

PR #1020, opened 2016-06-16
[HOTFIX][DOC] Fix image path

minahlee:
### What is this PR for?
Use relative path instead of absolute path so doc can find image under proper version folder

### What type of PR is it?
Documentation

Current 0.6.0-SNAPSHOT main page looks like this:
screen shot 2016-06-15 at 5 24 25 pm

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -33,7 +33,7 @@ limitations under the License.
     </ul>
   </div>
   <div class="col-md-6" style="padding:0">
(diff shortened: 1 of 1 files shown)


bzz:
👍  looks great, let's merge!

corneadoug:
I feel like we battled against the opposite case before:
https://github.com/apache/zeppelin/pull/899

PR #1043, opened 2016-06-18
[ZEPPELIN-905] Fix import notebook with error result

minahlee:
### What is this PR for?
This PR is fixing import/clone notebook with error result. This PR adds test based on #933. 
> Note: This issue is one of the [blockers](https://issues.apache.org/jira/browse/ZEPPELIN-889) of 0.6.0 release so should be merged into branch-0.6 before release.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-905](https://issues.apache.org/jira/browse/ZEPPELIN-905)

When you try to import or clone notebook with error result, it should work.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -221,7 +221,8 @@ public class Note implements Serializable, JobListener {
       newParagraph.setReturn(result, null);
     } catch (Exception e) {
       // 'result' part of Note consists of exception, instead of actual interpreter results
-      logger.warn("Paragraph " + srcParagraph.getId() + " has a result with exception. " + e.getMessage());
+      logger.warn("Paragraph " + srcParagraph.getId() + " has a result with exception. "
+              + e.getMessage());
     }
 
     synchronized (paragraphs) {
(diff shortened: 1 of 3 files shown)


minahlee:
@swkimme @Leemoonsoo Ready for review

bzz:
Looks good to me

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -207,17 +207,23 @@ public void addCloneParagraph(Paragraph srcParagraph) {
...
-    newParagraph.setReturn(result, null);
+
+    try {
+      Gson gson = new Gson();
+      String resultJson = gson.toJson(srcParagraph.getReturn());
+      InterpreterResult result = gson.fromJson(resultJson, InterpreterResult.class);
+      newParagraph.setReturn(result, null);
+    } catch (Exception e) {
+      // 'result' part of Note consists of exception, instead of actual interpreter results
+      logger.warn("Paragraph " + srcParagraph.getId() + " has a result with exception. "
```
felixcheung: Should this be a logger.error and re-throw the exception?

minahlee: In my understanding, if we re-throw error, import/clone notebook with `Object` result(for example `String`) other than `InterpreterResult` will fail as current master is.

bzz: @felixcheung good question, but I think this just happens to be an exception here and it does not represent an error condition in our logic, as it's fine for current notebook structure to have paragraph with an exception stack trace in  the output.

bzz:
I think we should merge it ASAP as per [ZEPPELIN-889](https://issues.apache.org/jira/browse/ZEPPELIN-889) this blocks us from makeing a release.

AhyoungRyu:
@bzz Let's merge this one 👍

Leemoonsoo:
LGTM

minahlee:
Merging it into master and branch-0.6 if there is no more discussion

PR #1048, opened 2016-06-20
[ZEPPELIN-1033] HotFixing of paragraph deletion and re-ordering broken

astroshim:
### What is this PR for?
This PR is for fixing [this issue](https://issues.apache.org/jira/browse/ZEPPELIN-1033)

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1033

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
@@ -21,6 +21,9 @@ import org.apache.shiro.realm.text.IniRealm;
 import org.apache.shiro.subject.Subject;
 import org.apache.shiro.util.ThreadContext;
 import org.apache.shiro.web.mgt.DefaultWebSecurityManager;
+import org.apache.shiro.mgt.SecurityManager;
+import org.apache.shiro.config.Ini;
+import org.apache.shiro.config.IniSecurityManagerFactory;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 
 import java.net.InetAddress;
@@ -33,6 +36,12 @@ import java.util.*;
  * Tools for securing Zeppelin
  */
 public class SecurityUtils {
+
+  public static void initSecurityManager() {
+    IniSecurityManagerFactory factory = new IniSecurityManagerFactory();
+    SecurityManager securityManager = factory.getInstance();
+    org.apache.shiro.SecurityUtils.setSecurityManager( securityManager );
+  }
 
   public static Boolean isValidOrigin(String sourceHost, ZeppelinConfiguration conf)
       throws UnknownHostException, URISyntaxException {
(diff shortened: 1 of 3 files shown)


randerzander:
@astroshim still getting same errors. After trying to "move up" and then "delete" a cell, I get these errors in the log: http://pastebin.com/Sr4WveXF

astroshim:
@randerzander Thank you for your review. Let me check.

astroshim:
@randerzander Did you build this PR? 
You can get this PR from ```git clone https://github.com/astroshim/zeppelin -b ZEPPELIN-1033```.
My zeppelin is working well as following screenshot.
test

Could you please check one more?

AhyoungRyu:
It works as before 👍

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -21,6 +21,9 @@
 import org.apache.shiro.subject.Subject;
 import org.apache.shiro.util.ThreadContext;
 import org.apache.shiro.web.mgt.DefaultWebSecurityManager;
+import org.apache.shiro.mgt.SecurityManager;
+import org.apache.shiro.config.Ini;
```
khalidhuseynov: may need to remove import of `Ini`, doesn't seem to be used

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -34,6 +37,12 @@
  */
 public class SecurityUtils {
 
+  public static void initSecurityManager(String shiroPath) {
+    IniSecurityManagerFactory factory = new IniSecurityManagerFactory("file:" + shiroPath);
+    SecurityManager securityManager = factory.getInstance();
+    org.apache.shiro.SecurityUtils.setSecurityManager( securityManager );
```
khalidhuseynov: maybe remove spaces

khalidhuseynov:
I checked it, and behavior with paragraph reordering fixed. So basically, previously shiro was initialized only when needed from `shiro.ini`, and now it's initialized always when starting the server explicitly. to be honest, i'm not exactly sure at the moment about the best way to solve the issue with `SecurityManager` initialization, since previously it was also initialized (automatically through the classpath) but wasn't available (possibly because of not injecting correctly). But definitely this one way to solve the issue at the moment, which can be further improved

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
astroshim: Thank you for your review. I removed importing Ini.

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
astroshim: Thank you for your review. I removed spaces.

randerzander:
I tried again from your branch and it worked. I must have applied your commit incorrectly last time.. my bad! Thanks for the fix =)

astroshim:
nice @randerzander thanks for addressing this issue and review.

Leemoonsoo:
+1 for merge

minahlee:
Merging it as hotfix

PR #1050, opened 2016-06-20
[MINOR] Show properties ordered alphabetically

jongyoul:
### What is this PR for?
Showing properties sorted in the interpreter tab 

### What type of PR is it?
[Improvement]

### Todos
* [x] - Change the np-repeat block

### What is the Jira issue?
N/A

screen shot 2016-06-20 at 6 10 38 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.controller.js
+++ b/zeppelin-web/src/app/interpreter/interpreter.controller.js
@@ -335,15 +335,6 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl', function($scope, 
     }
   };
 
-  $scope.sortProperties = function(properties) {
-    var arr = new Array();
-    properties.map(function(key, value) {
-      arr.push(key, value);
-    });
-    arr.sort();
-    return arr;
-  };
-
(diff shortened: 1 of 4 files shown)


jongyoul:
@corneadoug Could you please review this PR? It has very simple change but I don't know where the right position of that function is.

corneadoug:
@jongyoul seems to be working, the gracious way would have been to use an orderBy filter (https://docs.angularjs.org/api/ng/filter/orderBy), but it only applies to Array, and we are using an Object for some reason.

Maybe you could do a custom filter

jongyoul:
@corneadoug Thanks for the quick reply. I've googled `orderBy` and tried it first but this was not applied by map as you told me. I'll try to do with custom filter with keys. I have another question about the location of filter. Where is a proper position to create filter? May I create `interpreter.filter.js`? or just add `interpreter.controller.js`?

corneadoug:
Would be better in a separated file

jongyoul:
@corneadoug How about it?

corneadoug:
Tested, LGTM

jongyoul:
Will merge if there's no more discussion.

PR #1052, opened 2016-06-20
ZEPPELIN-1037 Enable Kerberos support in Livy Interpreter

r-kamath:
### What is this PR for?
This PR is for enabling kerberos support in Livy interpreter. Also introduced two new Livy interpreter properties called keytab and principal to configure. 

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-1037

Kerberize the Livy server and configure the Livy interpreter with appropriate keytab and principal.

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
@@ -21,22 +21,21 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.HttpClient;
-import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
-import org.apache.http.entity.StringEntity;
-import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.zeppelin.interpreter.InterpreterContext;
 import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.InterpreterResult.Code;
 import org.apache.zeppelin.interpreter.InterpreterUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
```diff
@@ -69,15 +68,16 @@ public Integer createSession(InterpreterContext context, String kind) throws Exc
...
 
       String json = executeHTTP(property.getProperty("zeppelin.livy.url") + "/sessions",
-          "POST",
-          "{" +
-              "\"kind\": \"" + kind + "\", " +
-              "\"conf\": " + confData + ", " +
-              "\"proxyUser\": \"" + context.getAuthenticationInfo().getUser() +
-              "\"}",
-          context.getParagraphId()
+              "POST",
```
jongyoul: First of all, it seems to fix your indentation.

jongyoul:
@r-kamath There's are two kind of http-client libraries. Can you replace one to another? I don't know if it's good that we use Spring framework or not, but it looks like a problem that we have two different libraries with same function. How about it?

livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
```diff
@@ -69,15 +68,16 @@ public Integer createSession(InterpreterContext context, String kind) throws Exc
...
+              "{" +
+                "\"kind\": \"" + kind + "\", " +
+                "\"conf\": " + confData + ", " +
+                "\"proxyUser\": " + (user == null ? null : "\"" + user + "\"") +
```
felixcheung: should we check if user is an empty string here?

r-kamath: fixed

r-kamath:
@jongyoul fixed the indentation and removed apache http client dep.

r-kamath:
CI failed at one of the build matrix

jongyoul:
@r-kamath Thanks for fixing it.

prabhjyotsingh:
Have tested it, it works on a Kerberos enabled system.

prabhjyotsingh:

PR #1059, opened 2016-06-21
[ZEPPELIN-954] Fix table cell selection problem on second run by properly destroying hot.

jasonxh:
### What is this PR for?
* Fix table cell selection problem on second run by properly destroying hot.
* Also make cells readonly. Previously one were able to paste into them.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-954]

Execute the following paragraph multiple times, and verify the table cells are still selectable.
```
%sh
echo %table
echo -e "col1\tcol2\tcol3"
echo -e "1\t2.1\tabcdefg"
```
Also try to paste anything into a cell to no avail.

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1221,48 +1221,56 @@ angular.module('zeppelinWebApp')
     websocketMsgSrv.commitParagraph($scope.paragraph.id, title, text, config, params);
   };
 
-  var setTable = function(type, data, refresh) {
+  var setTable = function(data, refresh) {
     var renderTable = function() {
       var height = $scope.paragraph.config.graph.height;
       var container = angular.element('#p' + $scope.paragraph.id + '_table').css('height', height).get(0);
-      var resultRows = $scope.paragraph.result.rows;
+      var resultRows = data.rows;
       var columnNames = _.pluck($scope.paragraph.result.columnNames, 'name');
 
-      if ($scope.hot) {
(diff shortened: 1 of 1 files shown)


minahlee:
Tested both cases, looks good to me 👍

minahlee:
\cc @corneadoug for review

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1224,12 +1224,16 @@ angular.module('zeppelinWebApp')
   var setTable = function(type, data, refresh) {
     var renderTable = function() {
       var height = $scope.paragraph.config.graph.height;
-      angular.element('#p' + $scope.paragraph.id + '_table').css('height', height);
+      var container = angular.element('#p' + $scope.paragraph.id + '_table').css('height', height).get(0);
```
corneadoug: Tested, works great, but could we keep those two actions separated?

corneadoug:
Tested, No more handsontable duplicates.
Was there no other way to just update the data instead of destroying the table and creating a new one?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
jasonxh: i figured we could just start using the return value and save an extra lookup for the same selector. pls let me know if you have strong preference and i can revert this line.

jasonxh:
@corneadoug good suggestion. i updated the pr to reuse the table when possible.

minahlee:
Merge it there is no more discussion

PR #1064, opened 2016-06-22
BugFix-blocking of blank values insertion on the Credential page.

astroshim:
### What is this PR for?
This PR blocks the blank values insertion on the Credential page and 
changes the success message box to zeppelin's dialog box.

### What type of PR is it?
Bug Fix

Try to save with blank values on the Credential page.

  - before 
out

  - after
image

image

### Questions:

### DIFF
--- a/zeppelin-web/src/app/credential/credential.controller.js
+++ b/zeppelin-web/src/app/credential/credential.controller.js
@@ -24,11 +24,10 @@ angular.module('zeppelinWebApp').controller('CredentialCtrl', function($scope, $
 
   $scope.updateCredentials = function() {
     if (_.isEmpty($scope.credentialEntity.trim()) ||
-        _.isEmpty($scope.credentialUsername.trim()) ||
-        _.isEmpty($scope.credentialPassword.trim())) {
+        _.isEmpty($scope.credentialUsername.trim())) {
       BootstrapDialog.alert({
         closable: true,
-        message: 'The values must be filled.'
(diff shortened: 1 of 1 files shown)


zeppelin-web/src/app/credential/credential.controller.js
```diff
@@ -21,15 +21,28 @@ angular.module('zeppelinWebApp').controller('CredentialCtrl', function($scope, $
...
   $scope.credentialPassword = '';
-  
+
   $scope.updateCredentials = function() {
+    if (_.isEmpty($scope.credentialEntity.trim()) ||
+        _.isEmpty($scope.credentialUsername.trim()) ||
+        _.isEmpty($scope.credentialPassword.trim())) {
+      BootstrapDialog.alert({
+        closable: true,
+        message: 'The values must be filled.'
```
bzz: How about "Username\Entity can not be empty"?

Also, this will not allow user to have `" "` whitespace as a password, right? For username that make sense, but I think we should not have such restriction on the password

astroshim: @bzz Thank you for your review!
I totally agree with you. Let me fix it.

astroshim:
@bzz I fixed as your comments. please review.

AhyoungRyu:
LGTM 👍

bzz:
Thank you!
Looks good to me, mering if there is no further discussion

PR #1067, opened 2016-06-22
ZEPPELIN-1048: Pandas support for python interpreter

bzz:
### What is this PR for?
Display Pandas DataFrame using Zeppelin's Table Display system.

### What type of PR is it?
Feature

### Todos
* [x] fix NPE in logs on empty paragraph execution
* [x] matplotlib: refactor `zeppelin_show(plt)` -> `z.show(plt)`
* [x] pandas: support `z.show(df)`
* [x] update docs

### What is the Jira issue?
[ZEPPELIN-1048](https://issues.apache.org/jira/browse/ZEPPELIN-1048)

### How should this be tested?
"Zeppelin Tutorial: Python - matplotlib basic" should work, and

```python
import pandas as pd
rates = pd.read_csv("bank.csv", sep=";")
z.show(rates)
```
### Screenshots (if appropriate)
screen shot 2016-06-23 at 10 29 00

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? Yes

PR #1073, opened 2016-06-23
[HOTFIX] Fixed PythonInterpreterTest

jongyoul:
### What is this PR for?
Returning back to pass the CI

### What type of PR is it?
[Hot Fix]

### Todos
* [x] - Fix test cases

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1048

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1074, opened 2016-06-23
[HOTFIX] Fix python test case

minahlee:
### What is this PR for?
Fix python test case caused by #1067 

### What type of PR is it?
Hot Fix

run
```
mvn -pl 'zeppelin-interpreter,python' -Dtest=org.apache.zeppelin.python.PythonInterpreterTest -Denforcer.skip=true test -DfailIfNoTests=false
```
and see if build success


minahlee:
Closes since it is duplicated with #1073

PR #1088, opened 2016-06-26
[ZEPPELIN-1062] Get original InterpreterSetting when there is no change

AhyoungRyu:
### What is this PR for?
When we update interpreter property values, we edit some values and click "save" button. Then a bootstrap dialog is show up. In this step, if we click "cancel", the updated value is still existed in the table. This new value is not stored actually. After refresh, the original value will be shown. 

I just get original InterpreterSetting(`getInterpreterSettings()`) when there is no change(`result`) for fixing this bug.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1062](https://issues.apache.org/jira/browse/ZEPPELIN-1062)

It's really simple. Just see the below attached screenshots. 

 - Before
interpreter_setting

- After
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.controller.js
+++ b/zeppelin-web/src/app/interpreter/interpreter.controller.js
@@ -150,6 +150,8 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl', function($scope, 
               thisConfirm.close();
             });
           return false;
+        } else {
+          getInterpreterSettings();
         }
       }
     });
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
IMO, when user clicks on cancel after clicking on save; user should see the edit view back.

prabhjyotsingh:
Also what happens when user clicks on close ("x") button, instead of "OK" or "Close" ?

AhyoungRyu:
@prabhjyotsingh Sorry for my late response. You're right. After  just clicking `cancel`, it should focus on textbox again as you said. And regarding your second comment, I blocked ESC back drop & `x` closing. Could you check this again?

prabhjyotsingh:
Thank you for taking care of it. LGTM.

PR #1091, opened 2016-06-27
[ZEPPELIN-1065] Flaky Test - ParagraphActionsIT.testRemoveButton

prabhjyotsingh:
### What is this PR for?
This is fix for fixing flaky CI failing test

```
testRemoveButton(org.apache.zeppelin.integration.ParagraphActionsIT)  Time elapsed: 9.641 sec  <<< FAILURE!
java.lang.AssertionError: After Remove : Number of paragraphs are
Expected: <2>
     but: was <1>
	at org.apache.zeppelin.integration.ParagraphActionsIT.testRemoveButton(ParagraphActionsIT.java:161)
```

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Add delay after paragraph delete.

### What is the Jira issue?
* [ZEPPELIN-1065](https://issues.apache.org/jira/browse/ZEPPELIN-1065)

CI should be green

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
@@ -38,9 +38,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
-import static org.openqa.selenium.Keys.ENTER;
-import static org.openqa.selenium.Keys.SHIFT;
-
 abstract public class AbstractZeppelinIT {
   protected static WebDriver driver;
 
@@ -48,21 +45,6 @@ abstract public class AbstractZeppelinIT {
   protected static final long MAX_IMPLICIT_WAIT = 30;
   protected static final long MAX_BROWSER_TIMEOUT_SEC = 30;
   protected static final long MAX_PARAGRAPH_TIMEOUT_SEC = 60;
-
-  protected void sleep(long millis, boolean logOutput) {
-    if (logOutput) {
-      LOG.info("Starting sleeping for " + (millis / 1000) + " seconds...");
-      LOG.info("Caller: " + Thread.currentThread().getStackTrace()[2]);
-    }
-    try {
-      Thread.sleep(millis);
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
-    if (logOutput) {
-      LOG.info("Finished.");
-    }
-  }
 
   protected void setTextOfParagraph(int paragraphNo, String text) {
     String editorId = driver.findElement(By.xpath(getParagraphXPath(paragraphNo) + "//div[contains(@class, 'editor')]")).getAttribute("id");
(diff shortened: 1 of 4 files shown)


zeppelin-server/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
```diff
@@ -157,10 +157,11 @@ public void testRemoveButton() throws Exception {
       sleep(1000, true);
       driver.findElement(By.xpath("//div[@class='modal-dialog'][contains(.,'delete this paragraph')]" +
           "//div[@class='modal-footer']//button[contains(.,'OK')]")).click();
+      sleep(1000, true);
```
bzz: Is there a reason to use this method vs `ZeppelinITUtils.sleep(1000, false);` below in the same method?

If they are the same, may be while we are here it can be refactored to use the same impl?

bzz:
@prabhjyotsingh thank you for fixing this one!

What do you think, if would it make sense to reduce total wait time and get this test more stable with a wait polling implementation to locate such elements, instead of explicit wait everywhere?

I.e there are already one for text and paragraph here [AbstractZeppelinIT.waitForText]( https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java#L99) wich uses [AbstractZeppelinIT.pollingWait](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java#L108) underneath.

Would it make sense to add something like `AbstractZeppelinIT.waitAndClick()`?

prabhjyotsingh:
@bzz fair point;
 - have refactored all AbstractZeppelinIT's sleep calls into ZeppelinITUtils.
 - added another function clickAndWait, which will click and wait for 1s.

prabhjyotsingh:
CI fails for [ZEPPELIN-1063](https://issues.apache.org/jira/browse/ZEPPELIN-1063)

zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
```diff
@@ -147,10 +129,15 @@ protected void createNewNote() {
...
+    ZeppelinITUtils.sleep(1000, true);
     driver.findElement(By.xpath("//div[@class='modal-dialog'][contains(.,'delete this notebook')]" +
         "//div[@class='modal-footer']//button[contains(.,'OK')]")).click();
-    sleep(100, true);
+    ZeppelinITUtils.sleep(100, true);
+  }
+
+  protected void clickAndWait(final By locator) {
+    driver.findElement(locator).click();
+    ZeppelinITUtils.sleep(1000, true);
```
bzz: Looks great! 
Are there any reason not to use polling facilities of [pollingWait](https://github.com/prabhjyotsingh/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java#L108) ?

bzz:
Thank you for pointing out [ZEPPELIN-1063](https://issues.apache.org/jira/browse/ZEPPELIN-1063) - have updated it's description.

zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
prabhjyotsingh: Yes, just waiting for any animation, show/hide or any (JavaScript) binding to the element that might be happening after it is visible.

prabhjyotsingh:
I'll merge this, and make `clickAndWait` related changes in https://github.com/apache/zeppelin/pull/1092.

bzz:
Sounds good, and this PR looks great to me!

bzz:
@prabhjyotsingh on closing the JIRA issues, please do not forget to set correct `fix version` field. 
It's  0.7.0 now if changes were merged only to master, I have updated 1065

PR #1085, opened 2016-06-24
[MINOR]Removed some codes using interpreterClassList

jongyoul:
### What is this PR for?
Removing dependency from interpreterClassList because this prevent new interpreter without interpreterClassList from registering in Zeppelin.

### What type of PR is it?
[Bug Fix | Refactoring]

### Todos
* [x] - Removed codes

### What is the Jira issue?
N/A

### How should this be tested?
It doesn't have any change on user level.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1095, opened 2016-06-27
ZEPPELIN-1067 Change ordering mechanism for interpreters

jongyoul:
### What is this PR for?
Removing dependency of interpreterClassList in order to add custom interpreter easily. For now, we should add some configuration for loading custom interpreter. This will enable users to add custom interpreter without changing interpreterClassList

### What type of PR is it?
[Improvement]

### Todos
* [x] - Remove dependency of interprterClassList

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1067

That shouldn't change any thing on user level.

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
@@ -249,15 +249,22 @@ public abstract class Interpreter {
     private String name;
     //@SerializedName("interpreterClassName")
     private String className;
+    private boolean defaultInterpreter;
     private Map<String, InterpreterProperty> properties;
     private String path;
 
     public RegisteredInterpreter(String name, String group, String className,
         Map<String, InterpreterProperty> properties) {
+      this(name, group, className, false, properties);
+    }
+
+    public RegisteredInterpreter(String name, String group, String className,
+        boolean defaultInterpreter, Map<String, InterpreterProperty> properties) {
       super();
       this.name = name;
       this.group = group;
       this.className = className;
+      this.defaultInterpreter = defaultInterpreter;
       this.properties = properties;
     }
 
@@ -271,6 +278,14 @@ public abstract class Interpreter {
 
     public String getClassName() {
       return className;
+    }
+
+    public boolean isDefaultInterpreter() {
+      return defaultInterpreter;
+    }
+
+    public void setDefaultInterpreter(boolean defaultInterpreter) {
+      this.defaultInterpreter = defaultInterpreter;
     }
 
     public Map<String, InterpreterProperty> getProperties() {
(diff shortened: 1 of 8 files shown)


jongyoul:
This is a newer version of #1085

jongyoul:
`interpreterClassList` would be deleted when adopting new mechanism for interpreter setting finishes.

jongyoul:

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
```diff
@@ -249,15 +249,22 @@ public Interpreter getInterpreterInTheSameSessionByClassName(String className) {
     private String name;
     //@SerializedName("interpreterClassName")
     private String className;
+    private boolean defaultInterpreter;
```
Leemoonsoo: @jongyoul Could you explain what 'defaultInterpreter' means?

jongyoul: The original way to order interpreters is matching from `zeppelin.interpreters`. I, However, think we don't build with some of interpreters of same interpreter group. For example, We don't build Zeppelin with `spark.sql` excluding `spark.spark`. Then, we don't need to have full list of interpreters, and it would be ok to have a interpreter group orders and the information of which interpreter is used by default. `defaultInterpreter` indicates if this interpreter is default interpreter in a same interpreter group or not. If it sets, its interpreter will be located in front of the list of interpreter settings and interpreter gourp info.

Leemoonsoo: Okay, thanks for explanation.

But previously, order of interpreter(default interpreter) in the interpreter group as well as order of interpreter group(default interpreter group) were controlled by `zeppelin.interpreters`. So user were able to select which interpreter in which group will be the default.

Now order of interpreter(default interpreter) is defined in interpreter-setting.json and can not be changed by user.
And order of interpreter group (default interpreter group) is defined in `zeppelin.interpreter.group.order`.

Anyway, after [alias for interpreter](https://issues.apache.org/jira/browse/ZEPPELIN-1012), i think all notebook can always binded to every interpreter settings (we can remove interpreter binding steps). then with combination of https://issues.apache.org/jira/browse/ZEPPELIN-707, i think default interpreter / interpretergroup will become no longer necessary.

jongyoul: Sure. I also think it makes users feel comfortable if we adopt aliases and change of UI to show binded interpreter on a paragraph. Actually, in case of ordering of groups, I hope we support a menu to select default group and interpreter per user easily. It also makes backend simple.

Finally, the first impression of this work is that we should make it easily to insert a new custom interpreter not having in a `zeppelin.interpreters`. And because I also don't want to break any current user experiences, it results making a different configuration values. For the future, I'll simplify these ones.

Leemoonsoo: Okay. Thanks for explanation.

PR #1103, opened 2016-06-28
[ZEPPELIN-1076] Set hbase.client.retries.number for JDBC

prabhjyotsingh:
### What is this PR for?
If a user has "org.apache.phoenix:phoenix-core:4.x.x" jar added as a dependency in JDBC interpreter, and for some reason phoenix was not accessible or not properly configured; then the phoenix tries to for 35 times (which is default for hbase.client.retries.number) and each retires is 8 second apart, before it finally fails.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Set phoenix.hbase.client.retries.number for JDBC

### What is the Jira issue?
* [ZEPPELIN-1076](https://issues.apache.org/jira/browse/ZEPPELIN-1076)

In JDBC interpreter add `org.apache.phoenix:phoenix-core:4.4.0-HBase-1.0` as dependency, but don't configure phoenix setting.
Then try to run any sql query with any of the configured JDBC driver (like `show tables`)

 - Without this it will take slightly more than about 5 mins
 - With this it should fetch result sooner (in less than a minute)

### Questions:

### DIFF
--- a/jdbc/src/main/resources/interpreter-setting.json
+++ b/jdbc/src/main/resources/interpreter-setting.json
@@ -94,6 +94,12 @@
         "defaultValue": "org.apache.phoenix.jdbc.PhoenixDriver",
         "description": ""
       },
+      "phoenix.hbase.client.retries.number": {
+        "envName": null,
+        "propertyName": "phoenix.hbase.client.retries.number",
+        "defaultValue": "1",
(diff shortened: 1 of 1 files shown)


jongyoul:
LGTM. That property `hbase.client.retries.number` will be passed to phoenix jdbc, right?

prabhjyotsingh:
Yes, `phoenix.hbase.client.retries.number`, and have tested with and without this string.

prabhjyotsingh:

PR #1084, opened 2016-06-24
[HOTFIX] Fix flaky AuthenticationIT

prabhjyotsingh:
### What is this PR for?

This is hotfix for CI failing for 

Tests in error: 
  AuthenticationIT.testGroupPermission:177->authenticationUser:109->AbstractZeppelinIT.pollingWait:114 Â» Timeout

### What type of PR is it?
[Hot Fix]

### Todos
* [x] -  Add check for request.status === 500, while logout

### What is the Jira issue?
 - [ZEPPELIN-1009](https://issues.apache.org/jira/browse/ZEPPELIN-1009)

### How should this be tested?
Check for CI green 

### Questions:
* Does the licenses files need update? n/a
* Is there breaking changes for older versions? n/a
* Does this needs documentation? n/a

PR #1104, opened 2016-06-29
[ZEPPELIN-1078] Does not show the text result in the paragraph sometimes.

astroshim:
### What is this PR for?
This PR fixes the problem that does not show the text result in the paragraph sometimes.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1078

Try run python code constantly like screenshot.

  - before
before

### Questions:


astroshim:
It seems that the CI fail is related to https://github.com/apache/zeppelin/pull/1084.

astroshim:
re-trigger CI

astroshim:
re-trigger CI

bzz:
So you think `resultRefreshed ` condition is not useful check any more?

\cc @felizbear @corneadoug for review

astroshim:
@bzz Thank you for reviewing.
I think ```resultRefreshed``` condition causes bug and I don't know why it needs.
@corneadoug @felizbear please review.

corneadoug:
I'm currently writing a giant review as I'm testing things

corneadoug:
@bzz @astroshim 
Took me quite some time to dig into this.

Removing `resultRefreshed` would have negative performances effect, its original goal is to not re-render things if there was no change. however its condition isn't that great and a better one would be: `!angular.equals(data.paragraph.result, $scope.paragraph.result);`

A quick fix is indeed to remove `resultRefreshed`, but from this condition only `else if (newType === 'TEXT' && resultRefreshed) {`

The overall issue is that the implementation of the streaming feature in interpreters for the TEXT result type is not consistent.

And since the issue is complex and need of lot of explanations, I will create a separate issue.

We can keep this PR as a quick fix

astroshim:
@corneadoug 
I thought it might be impacted but I didn't find a problem so I was waiting these great review! 

The implementation of the streaming feature in interpreters for the TEXT result type is already a problem because the paragraph results are cleared before rendering result in the ```$scope.renderText = function() { ``` function.

Then in this PR, can I just remove ```resultRefreshed``` from ```else if (newType === 'TEXT' && resultRefreshed) {```?

corneadoug:
@astroshim Yes, its the best way for now

astroshim:
@corneadoug Thank you for making clear. I fixed the code as you advised.

astroshim:
re-trigger CI

corneadoug:
LGTM

corneadoug:

bzz:
Thank you for great feedback @corneadoug and prompt fixes @astroshim 

Looks great to me!

corneadoug:
@astroshim Can you rebase with master?

astroshim:
@corneadoug I rebased. thanks.

PR #389, opened 2015-11-04
ZEPPELIN-385 Read-only mode for zeppelin notebooks, for analysis sharing

babokim:
This PR is for read-only mode.

PR #1105, opened 2016-06-29
[ZEPPELIN-385] Notebook offline view mode

khalidhuseynov:
### What is this PR for?
This PR is to address so called `offline view` (view only + offline) mode. In this mode you open notebook in `only view` mode meaning can't do any editions + notebook and paragraph bars are disabled. Further, it's called `offline`, in case someone else does any changes in the notebook while you have it open, the changes will not be rendered in your screen until you reload it. There was initial work under #389, however some notable differences exist. For example, same functionality here available from frontend menu without need to restart Zeppelin. Also it provides `offline` view which was absent in initial PR.

### What type of PR is it?
Feature

### Todos
* [x] - basic frontend implementation
* [ ] - address comments + green CI

### What is the Jira issue?
related to [ZEPPELIN-385](https://issues.apache.org/jira/browse/ZEPPELIN-385)

1. open any notebook
2. click `offline view` from the dropdown menu of notebook `looknfeel` bar on the  top right
3. try to edit note or do any changes

view only
offlineview

offline view
offline4

### Questions:
* Does this needs documentation? Possibly

### DIFF
--- a/zeppelin-web/src/app/app.js
+++ b/zeppelin-web/src/app/app.js
@@ -62,6 +62,10 @@
               templateUrl: 'app/notebook/notebook.html',
               controller: 'NotebookCtrl'
             })
+            .when('/notebook/:noteId/offlineView', {
+                templateUrl: 'app/notebook/notebook.html',
+                controller: 'NotebookCtrl'
+            })
             .when('/jobmanager', {
               templateUrl: 'app/jobmanager/jobmanager.html',
               controller: 'JobmanagerCtrl'
--- a/zeppelin-web/src/app/notebook/notebook-actionBar.html
+++ b/zeppelin-web/src/app/notebook/notebook-actionBar.html
@@ -15,8 +15,8 @@ limitations under the License.
   <h3>
     <div style="float: left; width: auto; max-width: 40%">
       <input type="text" pu-elastic-input class="form-control2" placeholder="{{noteName(note)}}" style="min-width: 0px; max-width: 95%;"
-           ng-show="showEditor" ng-model="note.name" ng-blur="sendNewName();showEditor = false;" ng-enter="sendNewName();showEditor = false;" ng-escape="note.name = oldName; showEditor = false" focus-if="showEditor" />
-      <p class="form-control-static2" ng-click="showEditor = true; oldName = note.name" ng-show="!showEditor">{{noteName(note)}}</p>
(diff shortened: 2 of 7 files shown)


r-kamath:
@khalidhuseynov may be we should consider enabling show/hide editor and output buttons

corneadoug:
I'm not sure what the usage of this feature as if would be.
#389 was aiming at having a view only experience at a time when auth was not implemented.
In our case it would make more sense to link this feature to the notebook user access settings, especially since only changing the url would still allow to access it in edit mode

bzz:
@khalidhuseynov one thing: as soon as this changes the URL -  you might want to double-check and make sure that URLs to each separate paragraph (+IFrame ones) and a Search work with new URL.

khalidhuseynov:
@r-kamath that's good point, but that differs according to use cases (also as a workaround, possible to set those before switching to `offline link`). from point of view having `offline view`, no changes on notebook should be displayed and basically any paragraph updates are blocked to render in [here](https://github.com/apache/zeppelin/pull/1105/files#diff-52fa6628f0b6793d22ffe1f3e4763635R350), which includes updates from show/hide editor and output forms. However it can be reconsidered depending on more feedback on use cases.

@corneadoug it makes some sense, however `view only` isn't the only feature. i updated the description with second screenshot on `offline view`. anyways, I'll look more into how it goes along with the authentication and get back here.

@bzz thanks for feedback, i'll look into that.

ashishr-wizni:
Copying from Zeppelin mailing list:

It's not clear to me if offline mode would allow changing input controls (dropdowns/input box etc) and view the new result. Our requirement is for a dashboard mode, where multiple users can play with the data using the input controls but don't interfere with each other's work. This is a very basic requirement for making Dashboards using Zeppelin. 

I think isolating a notebook copy in a web socket/user session (in Dashboard mode), can achieve the desired results.

dceashish:
Copying from Zeppelin Dev List:

It's not clear to me if offline mode would allow changing input controls (dropdowns/input box etc) and view the new result. Our requirement is for a dashboard mode, where multiple users can play with the data using the input controls but don't interfere with each other's work. This is a very basic requirement for making Dashboards using Zeppelin. 

I think isolating a notebook copy in a web socket/user session (in Dashboard mode), can achieve the desired results.

bzz:
@khalidhuseynov do you think we can close this one now, until you look into it more again? Or is it WIP?

bzz:
@khalidhuseynov 1 week ping

khalidhuseynov:
@bzz thanks for ping, i'm closing it for now

PR #1111, opened 2016-07-01
[ZEPPELIN-1094] Run all paragraphs keeps appending empty paragraphs

prabhjyotsingh:
### What is this PR for?
Every time user clicks on "Run all paragraphs" button system keeps appending an empty paragraph. Ideally content of paragraph should be checked before adding any empty paragraph.
This started happening after [ZEPPELIN-707](https://issues.apache.org/jira/browse/ZEPPELIN-707) was merged.

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [x] - add more condition before calling note.addParagraph()

### What is the Jira issue?
* [ZEPPELIN-1094](https://issues.apache.org/jira/browse/ZEPPELIN-1094)

Before
before

After
after

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -1075,7 +1075,8 @@ public class NotebookServer extends WebSocketServlet implements
     boolean isTheLastParagraph = note.getLastParagraph().getId()
         .equals(p.getId());
     note.setLastReplName(paragraphId);
-    if (!Strings.isNullOrEmpty(text) && isTheLastParagraph) {
+    if (!(text.equals(note.getLastInterpreterName() + " ") || Strings.isNullOrEmpty(text)) &&
+        isTheLastParagraph) {
       note.addParagraph();
     }
 


r-kamath:
LGTM

corneadoug:
@jongyoul Since you were reviewing the original PR can you check this one out?

prabhjyotsingh:

jongyoul:
I'm testing it. Just wait.

jongyoul:
@prabhjyotsingh After run all paragraphs, note adds new paragraph at the end of last paragraph. And then, if the last paragraph is empty, note don't create paragraph anymore. Is it normal? I think it's better not to create empty paragraph. How about you?

prabhjyotsingh:
Yes, its working for me, if you refer to the gif attached above; if paragraph content is only interpreter name (like in this example "%md ") it wouldn't add a paragraph below, but if we append anything after it, on running that should append a paragraph below.

IMO, I too am in a favour of not adding an empty paragraph on running notebook.

jongyoul:
@prabhjyotsingh I only said that case of "running all paragraph". I think "running all paragraphs" is not interactive mode. But in a normal case, I like to add empty paragraph at the end of paragraph. What I told is that how about not adding empty paragraph when you click "running all paragraphs".

prabhjyotsingh:
@jongyoul Sure, now I understand what you mean, but before that what do we call an empty paragraph; should it be `Strings.isNullOrEmpty(text)` or `"%getInterpreterName()".equals(text.trim())` or `both`. 
And if we say `both` then this is kind of taking care of it. However I can try and tune it.

jongyoul:
@prabhjyotsingh You're right for the before behaviour. I thank you for taking care of it. In case of your PR, I've tested and worked well. I hope you can deal with my suggestion.

prabhjyotsingh:
Sure, thank you @jongyoul I'll open up a separate PR for fixing that.

PR #1117, opened 2016-07-03
[ZEPPELIN-1086] Auto Completion for Interpreter.

astroshim:
### What is this PR for?
This PR is to support auto-completion for interpreter.

### What type of PR is it?
Improvement 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1086

Type ```% + 'Ctrl+.' ``` in your paragraph.
please refer to screen shot.

o

### Questions:


astroshim:
re-trigger CI

AhyoungRyu:
@astroshim Awesome! :+1:  
I'm not sure where can be the proper place, it would be better users can know this feature's existence since it's really useful. How about adding this info in [here](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/components/modal-shortcut/modal-shortcut.html#L211)? This is just my personal opinion :)

astroshim:
@AhyoungRyu Thank you for your review!
I think it's better to make zeppelin manual on zeppelin homepage than adding shortcut info page. What do you think?

AhyoungRyu:
@astroshim Oh sorry for my late response. Yeah that would be better 👍

astroshim:
@bzz @Leemoonsoo please review.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -202,11 +201,26 @@ public Interpreter getCurrentRepl() {
...
+    for (InterpreterSetting intp: factory.getInterpreterSettings(note.getId())){
+      completion.add(new InterpreterCompletion(intp.getGroup(), intp.getGroup()));
+    }
+    return completion;
+  }
+
   public List<InterpreterCompletion> completion(String buffer, int cursor) {
+    String lines[] = buffer.split(System.getProperty("line.separator"));
+    if (lines.length > 0
+      && (lines[0].startsWith("%") && cursor <= lines[0].trim().length())) {
```
Leemoonsoo: Code text can follow after interpreter selection without line.separator. For example

```sql
%sql select * from table
```
In this case, `getInterpreterCompletion()` will be invoked even if cursor is in somewhere `select * from table`. isn't it?

astroshim: You're right. I missed that point. Thank you pointing out.

felixcheung: space or new line could be separator in this case?

astroshim: yes. It should be separated space or new line.

astroshim:
@Leemoonsoo As you can see the following screenshot, I fixed bug and changed completion results.

o

bzz:
@astroshim looks awesome, thank you for implementing!

How do you think, would it improve user experience with this feature if we could remove from suggestion list things, that are not interpreters? Like function names, etc. And how hard would that be to implement?

astroshim:
@bzz Thank you for reviewing. 
You mean that blocking the keyword and local meta values?

bzz:
What I mean is - as a user, when I auto-complete the interpreter name, I would expect a suggestion list that contains only interpreter names and nothing else.

Right now list of suggestion includes things like "NoSuchMethod" and `AbstractMethodError`.

If it is hard to remove such things from the suggestions list in this case - let's keep them. But if that is something we could implement in this PR - that would be nice.

How do you think?

astroshim:
@bzz I totally agree with you so I did try to remove things you mentioned but it's not easy.
Let me try one more. Thanks!

bzz:
Got it! Thanks a lot for looking into it.

Let me know when you think its ready and I will try again. You did great job already, making it auto-complete names so in case it's too hard - we could merge as it is now, its already very useful!

astroshim:
@bzz Thank you kindly for telling like that. 
What do you think about tagging meta value than removing keywords like following image? 
image

I think it looks more nicer than removing keywords.
If it is okay, let me make it at the another PR.

bzz:
Sounds as a great plan to me!

Merging this one then, if there is no further discussion

astroshim:
Ping

bzz:
Sorry, it took me a while to get back. Done!

minahlee:
@astroshim this PR causes compile error in current master. Do you mind to create another PR to fix it?

astroshim:
@minahlee Thank you for addressing that. I'll take a look.

astroshim:
@minahlee Sorry for bothering you. I just made PR for fixing compile error.

PR #1124, opened 2016-07-04
ZEPPELIN-1105: Python - add paragraph ERROR status

bzz:
### What is this PR for?
Implement paragraph ERROR status for Python interpreter in case of Error or Exception in the output.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-1105](https://issues.apache.org/jira/browse/ZEPPELIN-1105)

CI should pass, or

```
mvn "-Dtest=org.apache.zeppelin.python.PythonInterpreterWithPythonInstalledTest" test -pl python
```

should pass, or paragraph status should be ERROR for something like

```
import some-thing
```

screen shot 2016-07-04 at 21 30 23

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -24,6 +24,8 @@ import java.net.ServerSocket;
 import java.util.Collection;
 import java.util.List;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.zeppelin.display.GUI;
 import org.apache.zeppelin.interpreter.Interpreter;
@@ -34,7 +36,6 @@ import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
 import org.apache.zeppelin.scheduler.Job;
 import org.apache.zeppelin.scheduler.Scheduler;
 import org.apache.zeppelin.scheduler.SchedulerFactory;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


bzz:
Ready to merge, \cc @minahlee @jongyoul for review

bzz:
re-trigger CI

bzz:
reabsed on latest master
ping @khalidhuseynov @minahlee @jongyoul  for a review

minahlee:
Tested basic functionality. Looks good to me.

CI failure is irrelevant and being taken care of under #1129.

bzz:
Thank you @minahlee for prompt review and notice on CI fix, I really appreciate!

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -54,8 +55,9 @@
 
   private Integer port;
   private GatewayServer gatewayServer;
-  private Boolean py4J = false;
+  private Boolean py4JisInstalled = false;
   private InterpreterContext context;
+  private Pattern errorInLastLine = Pattern.compile(".*(Error|Exception): .*$");
```
felixcheung: this seems rather fragile, it would match something like:
```
there is no Error: ok
```

felixcheung: Maybe we should put user code under `try ... except` or check for non-empty `sys.stderr`?, as an alternative?

felixcheung:
I think this is a good improvement - left a comment.

bzz:
CI is green now

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
bzz: Good points, thank you @felixcheung !

There is definitely room for improvement - this is more like a foundation of this feature, as previous impl always returned `SUCCESS`.

Your example of `print("there is no Error: ok")` will indeed trigger the the false positive ERROR status.

I have filed [ZEPPELIN-1114](https://issues.apache.org/jira/browse/ZEPPELIN-1114) and logged this feedback there. I plan to take care of it in a subsequent PRs.

PR #1130, opened 2016-07-05
[ZEPPELIN-1113] Don't save Paragraph ColWidth if not necessary

corneadoug:
### What is this PR for?
We currently are saving the ColWidth changes all the time, while it is not necessary if the value is the same.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1113

* Select a Col-12 Paragraph and use the shortcut Ctr+Shift+ = to make it bigger
* You shouldn't see any `Send >> "COMMIT_PARAGRAPH"` in your browser console

### Questions:


felixcheung:
LGTM

corneadoug:
Re-Trigger CI, Tests are good locally

corneadoug:
CI is green at last :) Merging if there is no more discussions

PR #1071, opened 2016-06-23
[ZEPPELIN-1052] Application does not logout user when authcBasic is used

prabhjyotsingh:
### What is this PR for?
This PR is WRT to [this](http://apache-zeppelin-users-incubating-mailing-list.75479.x6.nabble.com/Fwd-Authentication-in-zeppelin-td3354.html) mail thread (Authentication in zeppelin)
Where in if authcBasic mechanisim is used then on clicking logout, the user doesn't gets logout.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] -  set username and password false on logout

### What is the Jira issue?
* [ZEPPELIN-533](https://issues.apache.org/jira/browse/ZEPPELIN-1052)

### How should this be tested?
In shiro.ini conf set `/** = authcBasic`, then start the zeppelin server.
 - try login as admin/password1
 - now try to logout (this should work)

### Questions:
* Does the licenses files need update? n/a
* Is there breaking changes for older versions? n/a
* Does this needs documentation? n/a

PR #1140, opened 2016-07-06
[ZEPPELIN-1125] Application does not logout user when authcBasic and `./grunt serve` is used

prabhjyotsingh:
### What is this PR for?
Creating this issue from [this](https://github.com/apache/zeppelin/pull/1071#issuecomment-230720461) comment, Application does not logout user when authcBasic is used and process was running with `grunt serve`

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1125](https://issues.apache.org/jira/browse/ZEPPELIN-1125)

Run web-app as `grunt serve` and configure shiro auth to use `authcBasic`, and then try to logout.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
@@ -17,7 +17,6 @@
 package org.apache.zeppelin.rest;
 
 import org.apache.shiro.authc.*;
-import org.apache.shiro.session.Session;
 import org.apache.shiro.subject.Subject;
 import org.apache.zeppelin.annotation.ZeppelinApi;
 import org.apache.zeppelin.server.JsonResponse;
@@ -26,7 +25,10 @@ import org.apache.zeppelin.utils.SecurityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.*;
+import javax.ws.rs.FormParam;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -109,23 +111,15 @@ public class LoginRestApi {
     LOG.warn(response.toString());
     return response.build();
   }
-  
-  @GET
+
+  @POST
   @Path("logout")
   @ZeppelinApi
   public Response logout() {
     JsonResponse response;
-    
     Subject currentUser = org.apache.shiro.SecurityUtils.getSubject();
     currentUser.logout();
-
-    Map<String, String> data = new HashMap<>();
-    data.put("principal", "anonymous");
-    data.put("roles", "");
-    data.put("ticket", "anonymous");
-    data.put("WWW-Authenticate", "Basic realm=\"Login required\"");
-   
-    response = new JsonResponse(Response.Status.UNAUTHORIZED, "", data);
+    response = new JsonResponse(Response.Status.UNAUTHORIZED, "", "");
     LOG.warn(response.toString());
     return response.build();
   }
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
@@ -26,10 +26,7 @@ import org.apache.zeppelin.utils.SecurityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.FormParam;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.Produces;
+import javax.ws.rs.*;
 import javax.ws.rs.core.Response;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -113,7 +110,7 @@ public class LoginRestApi {
     return response.build();
   }
   
-  @POST
+  @GET
   @Path("logout")
   @ZeppelinApi
   public Response logout() {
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
@corneadoug, have made a fix for https://github.com/apache/zeppelin/pull/1071#issuecomment-230736598. Can you please review.

corneadoug:
@prabhjyotsingh Thanks, I will test that

r-kamath:
LGTM

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -113,7 +110,7 @@ public Response postLogin(@FormParam("userName") String userName,
     return response.build();
   }
   
-  @POST
+  @GET
```
anthonycorbacho: Any reason to change logout from `POST` to `GET`? I dont think its aligned why HTTP standard.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -126,8 +123,9 @@ public Response logout() {
     data.put("principal", "anonymous");
     data.put("roles", "");
     data.put("ticket", "anonymous");
+    data.put("WWW-Authenticate", "Basic realm=\"Login required\"");
```
anthonycorbacho: Having `WWW-Authenticate` in the body of the response, will not solve the logout issue, it will be ignored by the browser, you have to put this information in the [response header](https://tools.ietf.org/html/rfc2617#section-3.2.1).

Plus after reading this method, I think something is not right here. In my understanding if you want to logout a Subject you should use the getSecurityManager.

somthing like
```
Subject subject = SecurityUtils.getSubject();
SecurityUtils.getSecurityManager().logout(subject);
```

zeppelin-web/src/components/navbar/navbar.controller.js
```diff
@@ -54,30 +54,22 @@ angular.module('zeppelinWebApp')
...
-          request.open('post', logoutURL, true, 'false', 'false');
-          request.send();
-        }
-      }
-    };
-    request.send();
+
+    $http.get(logoutURL).error(function() {
+      //force authcBasic (if configured) to logout
+      $http.get(logoutURL).error(function() {
```
anthonycorbacho: Why are your doing 2 logout request?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -126,8 +123,9 @@ public Response logout() {
     data.put("principal", "anonymous");
```
corneadoug: It seems we also send back data, but are never using it in the front

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
prabhjyotsingh: Yes, you are right, let me put `WWW-Authenticate` in response header.

I was already doing this, but it was not solving the case for basic-authentication. Let me try your recommendation, I'll update in a bit.
```
Subject currentUser = org.apache.shiro.SecurityUtils.getSubject();
currentUser.logout();
```

corneadoug:
@prabhjyotsingh I got it fixed by only replacing the logout function in the front-end with:
```
$http.post(logoutURL)
    .error(function(response) {
      console.log('logout error %o', response);
    })
    .success(function(response) {
      $rootScope.userName = response.body.principal;
      $rootScope.ticket.principal = response.body.principal;
      $rootScope.ticket.ticket = response.body.ticket;
      $rootScope.ticket.roles = response.body.role;
      BootstrapDialog.show({
        message: 'Logout Success'
      });
      setTimeout(function() {
        window.location.replace('/');
      }, 1000);
    });
```

This could probably use some ngToast when error

zeppelin-web/src/components/navbar/navbar.controller.js
prabhjyotsingh: This seems to be the browser behaviour (have tried it on Chrome, Firefox and Safari), and in all of the browsers that I've tested requires logout to be called twice.
http://stackoverflow.com/a/233551/2749712

prabhjyotsingh:
@corneadoug I tried your approach, it didn't worked for me.

corneadoug:
@prabhjyotsingh Let me try again

corneadoug:
@prabhjyotsingh You are calling `$http.post` twice instead of once.
Also, could you reverse the response message UNAUTHORIZED to OK, so that we can keep a success catch event when logout succeed?

prabhjyotsingh:
Yes, basic-auth was not getting logout on first call, hence I had to call it twice. 
Advantage of keeping `UNAUTHORIZED` is that we don't have to bother about $http.post's success, since just after that we are redirecting page to "/" where all the init variables will be re-set.

prabhjyotsingh:
@corneadoug  Shall I merge this, if it works for you ?

corneadoug:
@prabhjyotsingh Basic Auth is getting logout after only one call.

> Advantage of keeping UNAUTHORIZED is that we don't have to bother about $http.post's success, since just after that we are redirecting page to "/" where all the init variables will be re-set.

I don't see your point, you still have to handle an action after getting the message from the API, whether it is UNAUTHORIZED or OK. 

If the API action is successful, it should be OK.

prabhjyotsingh:
> Basic Auth is getting logout after only one call.

Kind of strange it, doesn't work for me. Have attached gif's where in console, first I'm trying to call it once, then twice, and on calling it twice, it logout.

[chrome-gif](https://cloud.githubusercontent.com/assets/674497/16712451/5cbe7b0e-46a5-11e6-9ef7-90344433dc5c.gif)

[firefox-gif](https://cloud.githubusercontent.com/assets/674497/16712450/5cbe4d1e-46a5-11e6-9524-821fc37744d0.gif)


Also after making API return OK, `./grunt serve` stops working, guess this would be this reason. http://stackoverflow.com/questions/233507/how-to-log-out-user-from-web-site-using-basic-authentication/233551#233551

corneadoug:
@prabhjyotsingh I made a test PR with gif of Chrome, Firefox, Safari https://github.com/apache/zeppelin/pull/1159

prabhjyotsingh:
@corneadoug should we merge this, so that, `grunt serve` starts working, and tune this later ?

corneadoug:
@prabhjyotsingh LGTM

PR #1004, opened 2016-06-13
[ZEPPELIN-996] Improve first page and dropdown menu in documentation site

AhyoungRyu:
### What is this PR for?
Current Zeppelin documentation site is little bit hard to find a way for Zeppelin beginners. It will not easy  to improve this at a time, but I did the below as a start of this work.

1. Restructured dropdown menu and added each category names
2. Added a overview list(with short description) to first page of website (index.md) so that people can look through the overall contents in Zeppelin website at a glance. (as [Apache Spark](http://spark.apache.org/docs/latest/#where-to-go-from-here) and [Apache Mesos](http://mesos.apache.org/documentation/latest/) does)

Please see the attached screenshot images :)

### What type of PR is it?
Improvement & Documentation

### Todos
* [x] - Change outdated screenshot images 
* [x] - Combine `text.md`, `table.md` and `html.md` to `basicdisplaysystem.md`
* [x] - Fix dead link in `virtual_machine.md`
* [x] - Improve dropdown menu and reorder
* [x] - Improve first page(`index.md`)
* [x] - Combine with #995 after it is merged into master

### What is the Jira issue?
[ZEPPELIN-996](https://issues.apache.org/jira/browse/ZEPPELIN-996)

### How should this be tested?
1. Apply this patch and [build the docs website with jekyll](https://github.com/apache/zeppelin/tree/master/docs#build-documentation)
2. Check the first page(index.html) and dropdown menu

### Screenshots (if appropriate)
 - Dropdown menu
dropdown

 - First page
screen shot 2016-06-14 at 1 28 58 pm
screen shot 2016-06-14 at 1 29 11 pm
screen shot 2016-06-14 at 1 29 39 pm
screen shot 2016-06-14 at 1 29 57 pm
screen shot 2016-06-14 at 1 30 12 pm
screen shot 2016-06-14 at 1 30 22 pm

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1143, opened 2016-07-07
[ZEPPELIN-1085] Make dropdown menu scrollbar always visible

AhyoungRyu:
### What is this PR for?
In Zeppelin docs website, there are 4 dropdown menus(Quick Start, Interpreters, Display System and More) in the navbar. Especially **Interpreters** and **More** menu have lots of lists so I made this dropdown menu scrollable in #1004.
Originally the scroll bar is shown up when  a user is trying to scroll down. That's why it's hard to notice there are more menus in the dropdown menu.  See [this issue](https://issues.apache.org/jira/browse/ZEPPELIN-1085).

So I made this scroll bar always visible. But only for chrome & safari. Unfortunately, Firefox doesn't support custom scroll bar. See [here](http://stackoverflow.com/questions/18317634/force-visible-scrollbar-in-firefox-on-mac-os-x).
In short, it says
>There is no way to actually "force" Firefox render the old-style scrollbar since the default scrollbar used in the system is predefined by the OS itself (note that you can modify which scrollbar you want in System Preferences).

### What type of PR is it?
Improvement

### TODO
- [x] : test this in major browsers

### What is the Jira issue?
[ZEPPELIN-1085](https://issues.apache.org/jira/browse/ZEPPELIN-1085)

[Build the Zeppelin docs website](https://github.com/apache/zeppelin/tree/master/docs#build-documentation) in your local and check the dropdown menu in navbar. The scrollbae has to be shown right after you click **Interpreters** or **More** menu.

 - Before : scrollbar is only shown up when you are trying to scroll up & down
before

 - After : always visible
after

 - In case of IE, you can always see the scrollbar. So it's not a problem originally. 
screen shot 2016-07-07 at 5 41 27 pm

### Questions:

### DIFF
--- a/docs/assets/themes/zeppelin/css/style.css
+++ b/docs/assets/themes/zeppelin/css/style.css
@@ -141,6 +141,16 @@ body {
   overflow: auto;
 }
 
+.scrollable-menu::-webkit-scrollbar {
+  -webkit-appearance: none;
+  width: 7px;        
+}  
+
+.scrollable-menu::-webkit-scrollbar-thumb {
+  border-radius: 3px;
+  background-color: gray; 
+}
+
 .index-header {
   font-size: 16px;
   font-style: italic;


AhyoungRyu:
@jeffsteinmetz It will be appreciate if you can review this PR as a reporter of this issue :)

Leemoonsoo:
Thanks @AhyoungRyu for improvement.
Tested with Chrome and Safari. LGTM

prabhjyotsingh:
LGTM

jeffsteinmetz:
LGTM
At least in Firefox, the scroll bar now displays for a few seconds (doesn't stay, but its an improvement).
Chrome and Safari work as expected.

Tested with:
 - Chrome on OSX: Version 51.0.2704.103 (64-bit)
 - Safari: Version 9.1.1 (10601.6.17)
 - Firefox on OSX: 47.0.1

Leemoonsoo:
Merge it into master if there is no more discussion

AhyoungRyu:
@jeffsteinmetz Right. So Firefox users can know they can scroll down and there is more menus even though this PR couldn't be applied to Firefox. 
Thanks for testing this @Leemoonsoo @jeffsteinmetz :)

PR #1146, opened 2016-07-07
ZEPPELIN-1129 Zeppelin fails to load notebooks with old date format

r-kamath:
### What is this PR for?
Old notebook fails to load due to date parse exception

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1129

Keep a notebook with old date format in the notebook dir and reload notebooks on zeppelin home page

### Questions:


prabhjyotsingh:
Nice, good catch. 
LGTM!

jongyoul:
Test case would help to understand this change.

Leemoonsoo:
Thanks for the fix.
I have marked fix version 0.6.1 for issue ZEPPELIN-1129. So we can support old date format in the next minor release.

Looks good to me and merge it into master and branch-0.6 if there're no more discussions.

PR #1148, opened 2016-07-07
[ZEPPELIN-1026] set syntax highlight based on default bound interpreter

minahlee:
### What is this PR for?
- Set syntax highlight based on default interpreter if %magic is not set
- Get syntax highlight mode from backend

### What type of PR is it?
Improvement

### Todos
* [x] Add test
* [x] Update [writingzeppelininterpreter.md](https://github.com/apache/zeppelin/blob/master/docs/development/writingzeppelininterpreter.md)
* [ ] Fix test
* [ ] Test after #1145 merged, modify if necessary

### What is the Jira issue?
[ZEPPELIN-1026](https://issues.apache.org/jira/browse/ZEPPELIN-1026)

This is screenshot when the default interpreter set to python interpreter.
**Before**
Has `scala` as syntax highlight language when %python is not set.
screen shot 2016-07-07 at 10 46 20 pm

**After**
Has `python` as syntax highlight language even when %python is not set.
screen shot 2016-07-07 at 10 44 39 pm

### Questions:

### DIFF
--- a/docs/development/writingzeppelininterpreter.md
+++ b/docs/development/writingzeppelininterpreter.md
@@ -71,7 +71,7 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
       },...
     },
     "editor": {
-      "mode": "your-syntax-highlight-language"
+      "language": "your-syntax-highlight-language"
     }
   },
   {
@@ -110,7 +110,7 @@ If you want to add a new set of syntax highlighting,
 
   ```
   "editor": {
-      "mode": "java"
+      "language": "java"
   }
   ```
 
--- a/docs/development/writingzeppelininterpreter.md
+++ b/docs/development/writingzeppelininterpreter.md
@@ -42,7 +42,7 @@ In 'Separate Interpreter(scoped / isolated) for each note' mode which you can se
 Creating a new interpreter is quite simple. Just extend [org.apache.zeppelin.interpreter](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java) abstract class and implement some methods.
 You can include `org.apache.zeppelin:zeppelin-interpreter:[VERSION]` artifact in your build system. And you should put your jars under your interpreter directory with a specific directory name. Zeppelin server reads interpreter directories recursively and initializes interpreters including your own interpreter.
 
-There are three locations where you can store your interpreter group, name and other information. Zeppelin server tries to find the location below. Next, Zeppelin tries to find `interpreter-setting.json` in your interpreter jar. 
+There are three locations where you can store your interpreter group, name and other information. Zeppelin server tries to find the location below. Next, Zeppelin tries to find `interpreter-setting.json` in your interpreter jar.
 
 ```
 {ZEPPELIN_INTERPRETER_DIR}/{YOUR_OWN_INTERPRETER_DIR}/interpreter-setting.json
@@ -68,12 +68,15 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
         "propertyName": null,
         "defaultValue": "property2DefaultValue",
         "description": "Property 2 description"
-      }, ...
+      },...
+    },
+    "editor": {
+      "mode": "your-syntax-highlight-language"
     }
   },
   {
     ...
-  } 
+  }
 ]
 ```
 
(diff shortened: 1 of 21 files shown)


zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -79,12 +79,8 @@ angular.module('zeppelinWebApp')
   var angularObjectRegistry = {};
 
   var editorModes = {
-    'ace/mode/python': /^%(\w*\.)?(pyspark|python)\s*$/,
-    'ace/mode/scala': /^%(\w*\.)?spark\s*$/,
-    'ace/mode/r': /^%(\w*\.)?(r|sparkr|knitr)\s*$/,
     'ace/mode/sql': /^%(\w*\.)?\wql/,
-    'ace/mode/markdown': /^%md/,
-    'ace/mode/sh': /^%sh/
+    'ace/mode/markdown': /^%md/
```
felixcheung: should we get rid of these then?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -950,6 +916,70 @@ angular.module('zeppelinWebApp')
...
+      }
+    ), 1000);
+    return deferred.promise;
+  };
+
+  var setParagraphMode = function(session, paragraphText, pos) {
+    // Evaluate the mode only if the the position is undefined
+    // or the first 30 characters of the paragraph have been modified
+    // or cursor position is at beginning of second line.(in case user hit enter after typing %magic)
+    if ((typeof pos === 'undefined') || (pos.row === 0 && pos.column < 30) || (pos.row === 1 && pos.column === 0)) {
```
felixcheung: similarly discussed here, should we just look for the first space after `%`?
https://github.com/apache/zeppelin/pull/1117

I think it will be good to get to be consistent in how we look for interpreter

minahlee: Thanks for reviewing. https://github.com/apache/zeppelin/pull/1148/files#diff-52fa6628f0b6793d22ffe1f3e4763635R943 this line does look for white space after `%` as you suggested.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
minahlee: This PR reads syntax highlight language from `interpreter-setting.json` introduced by #835. I didn't remove this line because `%md`, `%*ql` is not following this new interpreter registration mechanism at the moment, in other words `%md`, `%*ql` interpreter doesn't have `interpreter-setting.json` file.
You can see the list of interpretesr that can apply new syntax highlight mechanism in ecb0af2

flink/src/main/resources/interpreter-setting.json
```diff
@@ -16,6 +16,9 @@
         "defaultValue": "6123",
         "description": "port of running JobManager."
       }
+    },
+    "editor": {
+      "mode": "scala"
```
cloverhearts: In my personal opinion,
"Mode" is not a bad name. But better "language" right.
mode stands for 'launguage mode' of ui-ace.
However, Interpreter shall not directly related to the ui-ace.
In addition, its value is "kind of language."
I believe that the future of Zeppelin may not use the ace editor.
If so, the name "mode" can cause confusion.
What do you think?

minahlee: Valid point indeed. let me update it

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -949,6 +919,70 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
...
+      // If paragraph loading, use config value if exists
+      if ((typeof pos === 'undefined') && $scope.paragraph.config.editorMode) {
+        session.setMode($scope.paragraph.config.editorMode);
+      } else {
+        var magic;
+        // set editor mode to default interpreter syntax if paragraph text doesn't start with '%'
+        if (!paragraphText.startsWith('%')) {
+          magic = $scope.$parent.interpreterBindings[0].group;
+        } else {
+          var replNameRegexp = /%(.+?)\s/g;
```
felixcheung: This seems like capture the whitespace after any character as well, eg.
it matches `%1 ` for capture group 1 - notice the space after `1`.

docs/development/writingzeppelininterpreter.md
```diff
@@ -96,15 +99,20 @@ some interpreter specific code...
...
 If you want to add a new set of syntax highlighting,  
 
 1. Add the `mode-*.js` file to <code>[zeppelin-web/bower.json](https://github.com/apache/zeppelin/blob/master/zeppelin-web/bower.json)</code> ( when built, <code>[zeppelin-web/src/index.html](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/index.html)</code> will be changed automatically. ).  
-2. Add to the list of `editorMode` in <code>[zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js)</code> - it follows the pattern 'ace/mode/x' where x is the name.  
-3. Add to the code that checks for `%` prefix and calls `session.setMode(editorMode.x)` in `setParagraphMode` located in <code>[zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js)</code>.  
+2. Add `editor` object to `interpreter-setting.json` file. If you want to set your language to java, add:
+
+  ```
+  "editor": {
+      "mode": "java"
```
felixcheung: should we say this mode/language should match the list supported by "ace"?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
felixcheung: Not sure how this would work with `jdbc`'s alias magic style too, eg. `%jdbc(hive)`, or `%hive`?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
felixcheung: should we consider non-whitespace only for the "first 30 characters", eg. `      %test`?
I guess we just need some consistent way to look for this here and PR #1117

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -226,6 +226,9 @@ public void onMessage(NotebookSocket conn, String msg) {
           case LIST_UPDATE_NOTEBOOK_JOBS:
             unicastUpdateNotebookJobInfo(conn, messagereceived);
             break;
+          case EDITOR_SETTING:
+            getEditorSetting(conn, notebook, messagereceived);
```
bzz: As far as I understood we introduce this new message type to current WS api in order just to pass this information to the front-end, right?

I wonder if there is another, simpler way to do it? As we are not changing it from the front, there is much less value in broadcasting it though WS (that's what we use WS mostly for), so it makes me think that there must be a simpler way to just pass this information to the front..or do I miss something here?

What do you think?

bzz:
@minahlee do you think we should close it for now or rename it to WIP?

minahlee:
I think I will create another PR for this so closing it.

PR #1153, opened 2016-07-08
Block all routes except Home when user isn't logged

corneadoug:
### What is this PR for?
When user isn't logged while the auth is activated, we are hiding the notebooks and most of the menus.
However you can still access them using the url routes.
This PR is blocking the url routes except the home route ('/') in that case.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1123

* Activate the auth in shiro.ini
* Launch Zeppelin, and try to visit `localhost:8080/#/interpreter` by typing it in your browser
* You should be redirected to '/'

### Questions:

### DIFF
--- a/zeppelin-web/src/app/app.js
+++ b/zeppelin-web/src/app/app.js
@@ -110,8 +110,7 @@
   function bootstrapApplication() {
     zeppelinWebApp.run(function($rootScope, $location) {
       $rootScope.$on('$routeChangeStart', function(event, next, current) {
-        if (!$rootScope.ticket && !next.$$route.publicAccess) {
-          event.preventDefault();
+        if (!$rootScope.ticket && next.$$route && !next.$$route.publicAccess) {
           $location.path('/');
         }
       });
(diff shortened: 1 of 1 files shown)


zeppelin-web/src/app/app.js
```diff
@@ -109,11 +108,18 @@
   }
 
   function bootstrapApplication() {
+    zeppelinWebApp.run(function($rootScope, $location) {
+      $rootScope.$on('$routeChangeStart', function(event, next, current) {
+        if (!$rootScope.ticket && !next.$$route.publicAccess) {
+          event.preventDefault();
```
prabhjyotsingh: `event.preventDefault();` is not letting my page to render. On removing this works as expected.

prabhjyotsingh:
LGTM! minor suggestion.

prabhjyotsingh:
Tried new changes works well.

corneadoug:
CI doesn't seem related, but still gonna retrigger it

corneadoug:
Ci is green, merging if there is no more discussions

PR #1155, opened 2016-07-09
[ZEPPELIN-1143] Interpreter dependencies are not downloaded on zeppelin start

prabhjyotsingh:
### What is this PR for?
While saving interpreter setting, if zeppelin server crashed/killed/restarted by user or there was some internet/download related (intermittent issue) because of which, or any other issue because of which dependencies were not downloaded.
In any such condition, when zeppelin is started/restarted, server should try and download dependencies.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-1143](https://issues.apache.org/jira/browse/ZEPPELIN-1143)

### SCREENSHOT
zeppelin-1143

 - Put a dependency (say "org.apache.commons:commons-csv:1.1") in any of the interpreter. 
 - from command line delete `local-repo` directory
 - restart zeppelin server 

expectation is `local-repo` should be recreated with all the dependencies that were mentioned in any of the interpreters.

### Questions:


jongyoul:
@prabhjyotsingh How to deal with it in case of getting error while downloading deps? Just edit and save it again?

prabhjyotsingh:
Yes, so far that is the only way we can download the dependency again.

rconlinehwx:
LGTM. @jongyoul  Typically I use an ambari created 3 cluster setup and then follow these instructions to setup my kerberos - https://docs.hortonworks.com/HDPDocuments/Ambari-2.2.1.1/bk_Ambari_Security_Guide/content/ch_configuring_amb_hdp_for_kerberos.html

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -332,10 +332,26 @@ private void loadFromFile() throws IOException {
...
 
+      Thread t = new Thread() {
+        public void run() {
+          try {
+            loadInterpreterDependencies(intpSetting);
+          } catch (RepositoryException e) {
+            logger.error("Error while downloading repos for interpreter group :" +
+                intpSetting.getGroup(), e);
+          } catch (IOException e) {
+            logger.error("Error while downloading repos for interpreter group :" +
```
felixcheung: you could combine the two catch clauses? or I would suggest having a different text log message for the exception

prabhjyotsingh:

jongyoul:
LGTM

karup1990:
A similar fix was made with https://github.com/apache/zeppelin/pull/755/files
Should we attempt to try to download dependencies when the interpreter process starts(as mentioned in the conversation).?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -332,10 +332,25 @@ private void loadFromFile() throws IOException {
...
+            loadInterpreterDependencies(intpSetting);
+          } catch (Exception e) {
+            logger.error(String.format("Error while downloading repos for interpreter group : %s," +
+                " go to interpreter setting page click on edit and save it again to make this " +
+                "interpreter work properly." ,
+                intpSetting.getGroup()), e);
+          }
+        }
+      };
+      t.start();
```
karup1990: Should this be synchronous?
If there are number of deps, the download might take a longer time. But the interpreter will be ready to take the paras.
If the code executed depends on these deps, it will fail I guess(until the time it is downloaded). And pass again after it is downloaded. Which will be kind of unexpected behaviour.

jongyoul:
I also think @karup1990's way is proper.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -332,10 +332,22 @@ private void loadFromFile() throws IOException {
...
       setting.getOption().setRemote(true);
 
-      InterpreterSetting intpSetting = new InterpreterSetting(setting.id(), setting.getName(),
+      final InterpreterSetting intpSetting = new InterpreterSetting(setting.id(), setting.getName(),
           setting.getGroup(), setting.getInterpreterInfos(), setting.getProperties(),
           setting.getDependencies(), setting.getOption());
 
+      try {
+        synchronized (interpreterSettings) {
+          loadInterpreterDependencies(intpSetting);
```
karup1990: The issue that i faced when loading interpreters synchronously:
The dependencies can take some time(time to download from mvn , erroneous scenarios etc). This will block loading other interpreters(since it is load method).

But the Notebook websocket server is running. The notebooks web UI will render properly. But when a para is run, the UI will seems to not respond(the para status will not change), because the interpreter might not have got loaded or in the process of being loaded.

It is good to show some UI feedback in these cases.
@prabhjyotsingh

prabhjyotsingh:
@karup1990 have tried to implement the behaviour that you were asking about, can you take a look. Have added a GIF in PR description.

karup1990:
Thanks for addressing the feedbacks @prabhjyotsingh .. 
I cannot see the GIF In the description of the PR  .

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -41,6 +42,21 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+      for (var setting = 0; setting < $scope.interpreterSettings.length; setting++) {
+        if ($scope.interpreterSettings[setting].status === 'DOWNLOADING_DEPENDENCIES') {
+          isDownloading = true;
+          break;
+        }
+      }
+      if (isDownloading) {
+        $timeout(function() {
+          getInterpreterSettings();
+        }, 2000);
```
karup1990: This timeout period will vary based on the number of dependencies.
Should we have Zeppelin broadcast a message when it is done with the downloading?

prabhjyotsingh: Yes I thought about it, but I'll take care of this in another PR, where instead of doing a get call, migrate this web-sockets.

prabhjyotsingh:
Have uploaded the GIF again, I think it would have failed last time.

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -26,6 +26,7 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
     var getInterpreterSettings = function() {
       $http.get(baseUrlSrv.getRestApiBase() + '/interpreter/setting').success(function(data, status, headers, config) {
         $scope.interpreterSettings = data.body;
+        checkDownloadingDependencies();
```
karup1990: This loop `getInterpreterSettings`<> `checkDownloadingDependencies`(this in turn iterates through setting) will be executed until the dependencies are downloaded.
Should be fine?
Otherwise, lgtm

prabhjyotsingh: Have added another condition that will check for route url, and if it is not `/interpreter` stop requesting for getInterpreterSettings().

prabhjyotsingh:

PR #1158, opened 2016-07-09
[ZEPPELIN-1131] Does not initialize login page values.

astroshim:
### What is this PR for?
This PR is for initialization of login page values(id, password).
The login id and password does not initialize even if login page closed.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1131

1. click login.
2. put id and password.
3. just close the modal window.
4. reopen login modal.
5. id and password should be initialized.

- before
before

- after
after

### Questions:

### DIFF
--- a/zeppelin-web/src/components/login/login.controller.js
+++ b/zeppelin-web/src/components/login/login.controller.js
@@ -39,5 +39,15 @@ angular.module('zeppelinWebApp').controller('LoginCtrl',
       });
 
     };
+
+    $rootScope.$on('initLoginValues', function() {
+      initLoginValues();
+    });
+    var initLoginValues = function() {
+      $scope.loginParams = {
+        userName: '',
+        password: ''
+      };
+    };
   }
 );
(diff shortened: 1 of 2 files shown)


r-kamath:
LGTM

zeppelin-web/src/components/login/login.controller.js
```diff
@@ -39,5 +39,15 @@ angular.module('zeppelinWebApp').controller('LoginCtrl',
       });
 
     };
+
+    $rootScope.$on('initLoginValues', function() {
```
corneadoug: $scope.$on, not $rootScope.$on

Can we also move the content of initValues function inside this block since this is the only one calling that function?

astroshim: @r-kamath @corneadoug Thank you for reviewing.
and I totally agree with @corneadoug. I just updated it.

corneadoug:
@astroshim, CI failure seems unrelated LGTM

prabhjyotsingh:
LGTM. Merging this if no more discussion.

PR #1159, opened 2016-07-11
[Test branch] Logout

corneadoug:
### What is this PR for?
This is a test branch related to #1140 
This won't be merged

### What type of PR is it?
Bug Fix

Login and Logout

Firefox - Grunt Serve
firefox-serve

Firefox - Grunt Build
firefox-classic

Chrome - Grunt Serve
chrome-serve

Chrome - Grunt Build
chrome-classic

Safari - Grunt Serve
safari-serve

Safari - Grunt Build
safari-classic

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
@@ -131,5 +131,4 @@ public class LoginRestApi {
     LOG.warn(response.toString());
     return response.build();
   }
-
 }
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -54,30 +54,18 @@ angular.module('zeppelinWebApp')
 
   $scope.logout = function() {
     var logoutURL = baseUrlSrv.getRestApiBase() + '/login/logout';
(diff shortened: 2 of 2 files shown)


prabhjyotsingh:
This works perfectly with `/** = authc`, but #1140 was for fixing `/** = authcBasic`

corneadoug:
@prabhjyotsingh Oh, that's tricky, my initial comment was about authc.

The fun thing is that authcBasic isn't even in the doc, nor as choice in the shiro.ini anymore. Do we still need it? Even pressing cancel on that alert send you to the authc home page.

After testing them all here is the result:

#1140 - Both authc and authcBasic works
Master - Both `authc grunt serve` and `authcBasic grunt serve` do not work
This PR - `authcBasic` and `authcBasic grunt serve` do not work

prabhjyotsingh:
Agreed @corneadoug. Thought we don't have `authcBasic` in our doc, but whoever knows or had used shiro in past can choose to use it, and IMO whatever features shiro provides we should be able to leverage them as well.

corneadoug:
@prabhjyotsingh If we can remove it, then maybe it would be nice to do so in the future.
Especially since it makes the code more complicated.
For now, let's go with #1140

prabhjyotsingh:
I don't have a strong view on any, fine either way. Just that I thought authcBasic should work.

PR #1162, opened 2016-07-11
[ZEPPELIN-909] Apply new mechanism to Elasticsearch interpreter

AhyoungRyu:
### What is this PR for?
This PR applies the [new interpreter registration mechanism](https://issues.apache.org/jira/browse/ZEPPELIN-804) to Elasticsearch interpreter.

### What type of PR is it?
Improvement

### Todos
* [x] - Remove static property definition code lines in `ElasticsearchInterpreter.java` and create `interpreter-setting.json` under `elasticsearch/src/main/resources`
* [ ] - fix CI build error

### What is the Jira issue?
[ZEPPELIN-909](https://issues.apache.org/jira/browse/ZEPPELIN-909)

1. apply this patch
2. rm -r `interpreter/elasticsearch` and rm `conf/interpreter.json`
3. build source
4. bin/zeppelin-daemon.sh start
5. Run a paragraph with elasticsearch commands 

### Questions:

### DIFF
--- a/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/ElasticsearchInterpreter.java
+++ b/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/ElasticsearchInterpreter.java
@@ -103,19 +103,6 @@ public class ElasticsearchInterpreter extends Interpreter {
   public static final String ELASTICSEARCH_CLUSTER_NAME = "elasticsearch.cluster.name";
   public static final String ELASTICSEARCH_RESULT_SIZE = "elasticsearch.result.size";
 
-  /*static {
-    Interpreter.register(
-      "elasticsearch",
-      "elasticsearch",
-      ElasticsearchInterpreter.class.getName(),
-        new InterpreterPropertyBuilder()
-          .add(ELASTICSEARCH_HOST, "localhost", "The host for Elasticsearch")
-          .add(ELASTICSEARCH_PORT, "9300", "The port for Elasticsearch")
-          .add(ELASTICSEARCH_CLUSTER_NAME, "elasticsearch", "The cluster name for Elasticsearch")
-          .add(ELASTICSEARCH_RESULT_SIZE, "10", "The size of the result set of a search query")
-          .build());
-  }*/
-
   private final Gson gson = new GsonBuilder().setPrettyPrinting().create();
   private Client client;
   private String host = "localhost";
--- a/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/ElasticsearchInterpreter.java
+++ b/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/ElasticsearchInterpreter.java
@@ -108,14 +108,20 @@ public class ElasticsearchInterpreter extends Interpreter {
   private String host = "localhost";
   private int port = 9300;
   private String clusterName = "elasticsearch";
-  private int resultSize = 10;
+  private int resultSize;
 
   public ElasticsearchInterpreter(Properties property) {
     super(property);
     this.host = getProperty(ELASTICSEARCH_HOST);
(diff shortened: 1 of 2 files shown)


bzz:
Current CI failure:

```
org.apache.zeppelin.elasticsearch.ElasticsearchInterpreterTest
org.apache.zeppelin.elasticsearch.ElasticsearchInterpreterTest  Time elapsed: 8.15 sec  <<< ERROR!
java.lang.NumberFormatException: null
	at java.lang.Integer.parseInt(Integer.java:454)
	at java.lang.Integer.parseInt(Integer.java:527)
	at org.apache.zeppelin.elasticsearch.ElasticsearchInterpreter.<init>(ElasticsearchInterpreter.java:118)
	at org.apache.zeppelin.elasticsearch.ElasticsearchInterpreterTest.populate(ElasticsearchInterpreterTest.java:99)


Results :

Tests in error: 
  ElasticsearchInterpreterTest.populate:99 Â» NumberFormat null

Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

AhyoungRyu:
@bzz CI is green now :)

elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/ElasticsearchInterpreter.java
```diff
@@ -103,32 +103,25 @@
...
 
   public ElasticsearchInterpreter(Properties property) {
     super(property);
     this.host = getProperty(ELASTICSEARCH_HOST);
     this.port = Integer.parseInt(getProperty(ELASTICSEARCH_PORT));
     this.clusterName = getProperty(ELASTICSEARCH_CLUSTER_NAME);
-    this.resultSize = Integer.parseInt(getProperty(ELASTICSEARCH_RESULT_SIZE));
+    try {
+      this.resultSize = Integer.parseInt(getProperty(ELASTICSEARCH_RESULT_SIZE));
+    } catch (NumberFormatException e) {
```
bzz: Is this `try..catch` is a common code idiom we use in other interpreters for parsing properties? It looks very verbose

I think this have to be consistent, i.e [python one code](https://github.com/apache/zeppelin/blob/master/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L74) looks better, why it's not the case there?

AhyoungRyu: @bzz I just saw [this one](https://github.com/apache/zeppelin/blob/master/lens/src/main/java/org/apache/zeppelin/lens/LensInterpreter.java#L115) and didn't thought about `valueOf`. It's better indeed and addressed it in the last commit. Thanks!

AhyoungRyu: @bzz Yeah you're right. I just tried to escape the CI build failure like [LensInterpreter.java](https://github.com/apache/zeppelin/blob/master/lens/src/main/java/org/apache/zeppelin/lens/LensInterpreter.java#L115). I addressed your suggestion in the last commit :)

bzz:
Thank you for pointing out Lens case!

May be we could extract this error handling logic to the method and re-use it everywhere, so the client code would look like:

```java
this.resultSize = parseIntegerOrDefault(getProperty(ELASTICSEARCH_RESULT_SIZE), 10);
```

and the API

```java
class Interpreter {
  protected Integer parseIntegerOrDefault(String str, Integer defaultValue) { }
}
```

This can be done in a separate PR though and include refactoring of all existing interpreters (this, Lense, Python, etc). 

What do you think?

AhyoungRyu:
@bzz Sure that make sense and thanks for clarifying it. Let me do that in other PR then :)

PR #1163, opened 2016-07-11
[ZEPPELIN-1149] %sh interpreter kerberos support

rconline:
### What is this PR for?
Zeppelin Shell interpreter should work in a Kerberos environment

### What type of PR is it?
[Feature]

### What is the Jira issue?
* [ZEPPELIN-1149](https://issues.apache.org/jira/browse/ZEPPELIN-1149)

 - shell.auth.type = KERBEROS
 - shell.principal = principal value
 - shell.keytab.location = keytab location

Now try and run any shell command (example: hdfs dfs -ls) it should return with valid results.

### Questions:

### DIFF
--- a/docs/interpreter/shell.md
+++ b/docs/interpreter/shell.md
@@ -30,6 +30,21 @@ At the "Interpreters" menu in Zeppelin dropdown menu, you can set the property v
     <td>60000</td>
     <td>Shell command time out in millisecs</td>
   </tr>
+  <tr>
+    <td>zeppelin.shell.auth.type</td>
+    <td></td>
+    <td>Types of authentications' methods supported are SIMPLE, and KERBEROS</td>
+  </tr>
+  <tr>
+    <td>zeppelin.shell.principal</td>
+    <td></td>
+    <td>The principal name to load from the keytab</td>
+  </tr>
+  <tr>
+    <td>zeppelin.shell.keytab.location</td>
+    <td></td>
+    <td>The path to the keytab file</td>
+  </tr>
 </table>
 
 ## Example
--- a/docs/interpreter/shell.md
+++ b/docs/interpreter/shell.md
@@ -14,6 +14,29 @@ Shell interpreter uses [Apache Commons Exec](https://commons.apache.org/proper/c
 In Zeppelin notebook, you can use ` %sh ` in the beginning of a paragraph to invoke system shell and run commands.
 Note: Currently each command runs as Zeppelin user.
 
+
+## Properties
+You can modify the interpreter configuration in the `Interpreter` section. The most common properties are as follows, but you can specify other properties that need to be connected.
+
+<table class="table-configuration">
+  <tr>
+    <td>shell.command.timeout.millisecs</td>
+    <td>Shell command time out in millisecs. Default = 60000</td>
+  </tr>
+  <tr>
+   <td>shell.auth.type</td>
+   <td>Types of authentications' methods supported are SIMPLE, and KERBEROS</td>
+  </tr>
+  <tr>
+   <td>shell.principal</td>
+   <td>The principal name to load from the keytab</td>
+  </tr>
+  <tr>
+   <td>shell.keytab.location</td>
+   <td>The path to the keytab file</td>
+  </tr>
+</table>
+
 ### Example
 The following example demonstrates the basic usage of Shell in a Zeppelin notebook.
 
--- a/docs/interpreter/shell.md
+++ b/docs/interpreter/shell.md
@@ -24,15 +24,15 @@ You can modify the interpreter configuration in the `Interpreter` section. The m
     <td>Shell command time out in millisecs. Default = 60000</td>
   </tr>
   <tr>
-   <td>shell.auth.type</td>
+   <td>zeppelin.shell.auth.type</td>
    <td>Types of authentications' methods supported are SIMPLE, and KERBEROS</td>
   </tr>
   <tr>
-   <td>shell.principal</td>
+   <td>zeppelin.shell.principal</td>
    <td>The principal name to load from the keytab</td>
   </tr>
   <tr>
-   <td>shell.keytab.location</td>
+   <td>zeppelin.shell.keytab.location</td>
    <td>The path to the keytab file</td>
   </tr>
 </table>
--- a/shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
@@ -41,11 +41,11 @@ public class ShellSecurityImpl {
     UserGroupInformation.AuthenticationMethod authType;
     try {
       authType = UserGroupInformation
-        .AuthenticationMethod.valueOf(properties.getProperty("shell.auth.type")
+        .AuthenticationMethod.valueOf(properties.getProperty("zeppelin.shell.auth.type")
           .trim().toUpperCase());
     } catch (Exception e) {
       LOGGER.error(String.format("Invalid auth.type detected with value %s, defaulting " +
-        "auth.type to SIMPLE", properties.getProperty("shell.auth.type").trim()));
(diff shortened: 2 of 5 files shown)


prabhjyotsingh:
LGTM, tested on ubuntu 14.04 server.

shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    try {
+      authType = UserGroupInformation
+        .AuthenticationMethod.valueOf(properties.getProperty("shell.auth.type")
+          .trim().toUpperCase());
+    } catch (Exception e) {
+      LOGGER.error(String.format("Invalid auth.type detected with value %s, defaulting " +
+        "auth.type to SIMPLE", properties.getProperty("shell.auth.type").trim()));
+      authType = SIMPLE;
+    }
+
```
felixcheung: extra newline

shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
```diff
@@ -0,0 +1,70 @@
...
+          String kinitCommand = String.format("kinit -k -t %s %s",
+            properties.getProperty("shell.keytab.location"),
+            properties.getProperty("shell.principal"));
+          cmdLine.addArgument(kinitCommand, false);
+          DefaultExecutor executor = new DefaultExecutor();
+
+          try {
+            int exitVal = executor.execute(cmdLine);
+          } catch (Exception e) {
+            throw new InterpreterException(e);
```
felixcheung: add logger.error()?

shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public static void createSecureCinfiguration(Properties properties, String shell) {
+    UserGroupInformation.AuthenticationMethod authType;
+    try {
+      authType = UserGroupInformation
+        .AuthenticationMethod.valueOf(properties.getProperty("shell.auth.type")
+          .trim().toUpperCase());
+    } catch (Exception e) {
+      LOGGER.error(String.format("Invalid auth.type detected with value %s, defaulting " +
+        "auth.type to SIMPLE", properties.getProperty("shell.auth.type").trim()));
```
felixcheung: pass exception as `LOGGER.error(...., e)`?

docs/interpreter/shell.md
```diff
@@ -14,6 +14,29 @@ Shell interpreter uses [Apache Commons Exec](https://commons.apache.org/proper/c
...
+    <td>shell.command.timeout.millisecs</td>
+    <td>Shell command time out in millisecs. Default = 60000</td>
+  </tr>
+  <tr>
+   <td>shell.auth.type</td>
+   <td>Types of authentications' methods supported are SIMPLE, and KERBEROS</td>
+  </tr>
+  <tr>
+   <td>shell.principal</td>
+   <td>The principal name to load from the keytab</td>
```
felixcheung: clarify this is KERBEROS only?

felixcheung:
If each interpreter has a separate settings for Kerberos, eg. shell.keytab.location, would it be harder to manage?

rconline:
@felixcheung over a period of time, each interpreter will start to have their own Keytab's. I think we should have separate keytabs for different interpreters. We will ofcourse have to invest on getting a better management console for interpreters themselves.

jongyoul:
@rconline What is `kinit`? I've tried to find a `kinit` in my CentOS, but I couldn't find it. Is it fine?

rconline:
@jongyoul here are some reference - http://linux.die.net/man/1/kinit & http://www.tutorialspoint.com/unix_commands/kinit.htm

jongyoul:
@rconline In case of not having `kinit`, there isn't a way to use kerberos, right?

rconline:
@jongyoul no there isn't. Here's the snippet from UserGroupInformation, which is essentially from Hadoop security. In future we will implement multiple methods of security. At present for Kerberos security Kinit is a necessary step. I intend to generalize this across other interpreters as well subsequently.

`public static enum AuthenticationMethod {
        SIMPLE,
        KERBEROS,
        TOKEN,
        CERTIFICATE,
        KERBEROS_SSL,
        PROXY;
`

shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+
+/***
+ * Shell security helper
+ */
+public class ShellSecurityImpl {
+
+  private static Logger LOGGER = LoggerFactory.getLogger(ShellSecurityImpl.class);
+
+  public static void createSecureCinfiguration(Properties properties, String shell) {
```
felixcheung: do you mean for this to be `createSecureConfiguration`?

shell/src/main/java/org/apache/zeppelin/shell/security/ShellSecurityImpl.java
```diff
@@ -0,0 +1,59 @@
...
+          CommandLine cmdLine = CommandLine.parse(shell);
+          cmdLine.addArgument("-c", false);
+          String kinitCommand = String.format("kinit -k -t %s %s",
+            properties.getProperty("zeppelin.shell.keytab.location"),
+            properties.getProperty("zeppelin.shell.principal"));
+          cmdLine.addArgument(kinitCommand, false);
+          DefaultExecutor executor = new DefaultExecutor();
+
+          try {
+            int exitVal = executor.execute(cmdLine);
```
felixcheung: it looks like this is launching kinit as a separate process. Shouldn't this run in the same process where we run the shell? I think the problem we have here is that the %sh interpreter is launching each %sh paragraph in a separate shell process? This is a problem because by running kinit like this it authenticate Zeppelin server, not individual interpreter, I think.

Also, keytab based Kerberos auth is generally for service account? 
This can be problematic for multi-user or mutli-tenant scenario - where the shell command should be authenticated as the user running the %sh paragraph?

rconline:
@felixcheung as of now the interpreters are reusing zeppelin keytab. Typically this is case in the hadoop security world, where the keytab is generated at a process level, whereas the authorization is done through impersonation. This is then mapped to the user permissions setup on the underlying data system via posix users. Mature systems which implement full support for security + impersonation, can provide this ability. Right now with this change we get to the first stage which is security. To get to impersonation we will have to start the shell with the user's identity. I think that can be followup patch, in the next couple of weeks.

prabhjyotsingh:

rconline:
Re-trigger CI

zjffdu:
agree with @felixcheung, use the zeppelin service keytab/principal might bring some other potential security problems. I think user A would expect to run shell script as himself, but actually it run as user zeppelin. And seems all the users are sharing the same principal if I understand correctly. 

Besides that I have one question about shiro of zeppelin, does shiro will map the username to hadoop side ? Say I login zeppelin as user A, would that guaranteed there should also be a user A in the hadoop ecosystem?

zjffdu:
Sorry, I mistake it as hdfs interpreter. For shell interpreter, do we must have kerberos support ? I think here kinit is just for the hadoop command, so actually it is for running hdfs command in secured cluster rather than running general shell command in secured cluster.

prabhjyotsingh:
This will also, allow CLI/shell interaction to other data systems such hbase, falcon, hive, kafka, oozie, sqoop, etc. on secured clusters. Hence this is a need, particularly during data ingestion via a notebook.

zjffdu:
@prabhjyotsingh  Thanks for clarification.

prabhjyotsingh:
CI fail is unrelated, which fails for Build #6048.2 with `No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.`

felixcheung:
I think getting impersonation is an important next step that could come after.

PR #862, opened 2016-04-28
[ZEPPELIN-704] Display elapse time for long running paragraph

sagarkulkarni3592:
### What is this PR for?
For long running paragraph, it is hard to know how long it has already been running. We should either have the elapsed time displayed

### What type of PR is it?
[Bug Fix ]

### What is the Jira issue?
ZEPPELIN-704 (https://issues.apache.org/jira/browse/ZEPPELIN-704)

### How should this be tested?
-Open a note.
-Run a paragraph , It should show you the elapsed time.

### Screenshots (if appropriate)
screen shot 2016-04-28 at 3 46 22 pm

### Questions:
* Does the licenses files need update?No
* Is there breaking changes for older versions?No
* Does this needs documentation?No

PR #1172, opened 2016-07-13
[Zeppelin-1161] Revert precise execution time

corneadoug:
### What is this PR for?
It had been pointed out in the mailing list that after https://github.com/apache/zeppelin/pull/862, the execution time was changed to humanized fashion, while we would want to keep something more precise in that case.

This PR is rolling back to the old precise time for execution time.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1161

Run a paragraph, after it is finished, it should show: `Took x seconds.` instead of `Took a few seconds` (or Hours, Minutes, Sec)

### Screenshot
screen shot 2016-07-13 at 2 19 45 pm

### Questions:
* Does the licenses files need update? Yes

### DIFF
--- a/zeppelin-distribution/src/bin_license/LICENSE
+++ b/zeppelin-distribution/src/bin_license/LICENSE
@@ -136,6 +136,7 @@ The text of each license is also included at licenses/LICENSE-[project]-[version
     (The MIT License) slf4j-log4j12 v1.7.10 (org.slf4j:slf4j-log4j12:jar:1.7.10 - http://www.slf4j.org) - http://www.slf4j.org/license.html
     (The MIT License) bcprov-jdk15on v1.51 (org.bouncycastle:bcprov-jdk15on:jar:1.51 - http://www.bouncycastle.org/java.html) - http://www.bouncycastle.org/licence.html
     (The MIT License) AnchorJS (https://github.com/bryanbraun/anchorjs) - https://github.com/bryanbraun/anchorjs/blob/master/README.md#license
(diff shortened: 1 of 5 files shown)


AhyoungRyu:
@corneadoug Wasn't it intended behaviour because of [https://github.com/apache/zeppelin/pull/862#issuecomment-221474937](https://github.com/apache/zeppelin/pull/862#issuecomment-221474937)? Anyway I prefer to show the precise execution time even if it shows such as `3600 seconds` 👍

corneadoug:
@AhyoungRyu No, that discussion was about the elapsed time.
Right, for long one, maybe we should show something else than seconds. I will modify that.

corneadoug:
@AhyoungRyu Made modifications and added screenshot

AhyoungRyu:
@corneadoug more graceful! Thanks for taking care of this 👍

AhyoungRyu:
@corneadoug When I build this PR in my local, i got this jscs warnings. 

screen shot 2016-07-13 at 3 21 18 pm

corneadoug:
I'm not sure why CI is not green, triggering it once more

SachinJanani:
@corneadoug Will this changes also be available in 0.6.1 ?

corneadoug:
@SachinJanani There should be a thread about 0.6.1.
If we want it to be part of it, I guess it should be added to  https://issues.apache.org/jira/browse/ZEPPELIN-1203

SachinJanani:
@corneadoug Yes you are right.

bzz:
Looks great to me, and CI is green now!

corneadoug:
Merging it now!

PR #1179, opened 2016-07-13
[ZEPPELIN-1109] Remove bootstrap dialog fade-in/out animation

AhyoungRyu:
### What is this PR for?
This PR will fix [ZEPPELIN-1109](https://issues.apache.org/jira/browse/ZEPPELIN-1109). 
I'm not sure this approach can be the best way for fixing this issue since I would prefer to have the fade-in/out animation for bootstrap dialog. So if anyone has better idea for this, please let me know. 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1109](https://issues.apache.org/jira/browse/ZEPPELIN-1109)

1. Press `Enter` button right after clicking any bootstrap dialog (i.e. trash can icon in the notebook) 
-> Bootstrap dialog will be shown up multiple times
2. Apply this patch and [build `zeppelin-web`](https://github.com/apache/zeppelin/tree/master/zeppelin-web#configured-environment)
3. Try number 1 again -> you can see there is no fade animation so you don't have time interval to generate multi dialogs

 - Before 
multiple_dialog

 - After
remove_animation

### Questions:

### DIFF
--- a/zeppelin-web/src/app/app.controller.js
+++ b/zeppelin-web/src/app/app.controller.js
@@ -44,4 +44,7 @@ angular.module('zeppelinWebApp').controller('MainCtrl', function($scope, $rootSc
   BootstrapDialog.defaultOptions.onshown = function() {
     angular.element('#' + this.id).find('.btn:last').focus();
   };
+  
+  // Remove BootstrapDialog animation
+  BootstrapDialog.configDefaultOptions({animate: false});
 });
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @AhyoungRyu for the fix.
LGTM

Leemoonsoo:
Merge into master if there're no more discussions

PR #1180, opened 2016-07-13
[ZEPPELIN-1059] Fix the Print CSS file

corneadoug:
### What is this PR for?
While investigating the original issue, I discovered that the `printMode.css` file was never used before.
I changed it so it is taken into account when printing, and added some CSS to remove unnecessary components in the printing page.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1059

Try to print a Note with your web-browser and look at the preview.
There shouldn't be the Zeppelin Navbar or a weird Title hiding the content.

After this PR:
screen shot 2016-07-13 at 7 16 50 pm

### Questions:


Leemoonsoo:
This is cool!!
Tested and LGTM

prabhjyotsingh:
LGTM

corneadoug:
Had to change some tests because of bad behaviour of `pu-elastic-input` with printing.

Added some functions to activate/deactivate the Implicit wait, it is mainly useful in the case of testing for elements that are not in the DOM.

Also added a line to handle firebug for people with higher version of Firefox

corneadoug:
CI failure is unrelated
(re-triggering CI at the same time)

PR #1184, opened 2016-07-14
[ZEPPELIN-1159] Livy interpreter gets "404 not found" error

prabhjyotsingh:
### What is this PR for?
RestTemplate throws HttpClientErrorException, exception thrown when an HTTP 4xx is received.
http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/HttpClientErrorException.html

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1159](https://issues.apache.org/jira/browse/ZEPPELIN-1159)

Run a paragraph using livy interpreter (say sc.version), now let this session expire (or just restart livy server), then try running the same paragraph, this should result in proper error message.

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
@@ -33,6 +33,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.kerberos.client.KerberosRestTemplate;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
 import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


felixcheung:
LGTM

prabhjyotsingh:

zjffdu:
@prabhjyotsingh  
```this is result in proper error message.```
What does this mean ? Is the behavior livy interpreter give proper error message to client or just recreate a new session ?

prabhjyotsingh:
@zjffdu have updated PR's description, what I meant was "this should result in proper error message". 
So, the original issue was with the RestTemplate that we are using which throws HttpClientErrorException exception when a HTTP 4xx is received, and when ever a session is queried which is either expired or does not exist livy throws `"Session '10' not found."` with response code 404, and because of this `HttpClientErrorException` proper error message was not shown on UI.

zjffdu:
@prabhjyotsingh  understood.  
But I feel the better solution is to either recreate a new session for user implicitly or tell the user to restart the livy interpreter as the session is expired. Because I don't think it is proper to expose the concept session to users, as users don't think what livy session represent. And user don't know what to do to handle `"Session '10' not found."`

prabhjyotsingh:
@zjffdu yes, we handle it(`"Session '10' not found."`) here.
https://github.com/apache/zeppelin/pull/1184/files#diff-6590a16671e62be8da95e793c364d739L305

and change this string to `Exception: Session not found, Livy server would have restarted, or lost session.`

zjffdu:
@prabhjyotsingh Got it, I would prefer to create new session for user implicitly and don't expose any livy concept to users. Just try to make it transparent to users.  Anyway, we can do it in a follow up ticket, rest lgtm.

prabhjyotsingh:
Sure thanks.

PR #1185, opened 2016-07-14
Redirect home when requested a wrong notename path.

astroshim:
### What is this PR for?
This PR for redirecting to zeppelin home when requested wrong path.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1176

put wrong path to your browser like screenshot.

- before
b1

- after
a

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -338,7 +338,7 @@ angular.module('zeppelinWebApp').controller('NotebookCtrl', function($scope, $ro
   /** update the current note */
   $scope.$on('setNoteContent', function(event, note) {
     if (note === undefined) {
-      $location.path('/');
+      $location.path('/#');
     }
 
     $scope.paragraphUrl = $routeParams.paragraphId;
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -338,7 +338,7 @@ angular.module('zeppelinWebApp').controller('NotebookCtrl', function($scope, $ro
   /** update the current note */
   $scope.$on('setNoteContent', function(event, note) {
     if (note === undefined) {
-      $window.location.href = '/';
+      $location.path('/');
     }
 
     $scope.paragraphUrl = $routeParams.paragraphId;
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
@astroshim Good catch! How about adding 404 page (like github does) instead of just redirecting `Home` without any message. In 404 page, we can add sth like "This is not the notebook page you are looking for" with some button for redirecting to home.

astroshim:
@AhyoungRyu Thank you for your opinion. 
Your idea is great but now is redirected all to home when requested wrong path. (you can see the related code [here](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/app.js#L85))
so I think it's better to follow the present rule. What do you think?

felixcheung:
+1 for 404 - maybe useful but could be a separate PR.

AhyoungRyu:
@astroshim Right. And agree with @felixcheung. Let do that in separate PR then :)

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -338,6 +337,10 @@ angular.module('zeppelinWebApp').controller('NotebookCtrl', function($scope, $ro
 
   /** update the current note */
   $scope.$on('setNoteContent', function(event, note) {
+    if (note === undefined) {
+      $window.location.href = '/';
```
corneadoug: Let's use `$location.path('/');`

astroshim: @corneadoug Thank you for pointing out. :+1:

astroshim:
re-trigger CI

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -338,7 +338,7 @@ angular.module('zeppelinWebApp').controller('NotebookCtrl', function($scope, $ro
   /** update the current note */
   $scope.$on('setNoteContent', function(event, note) {
     if (note === undefined) {
-      $location.path('/');
+      $location.path('/#');
```
corneadoug: There is no need for the `#`

astroshim: fixed. Thanks.

corneadoug:

bzz:
Looks great to me

PR #1186, opened 2016-07-15
[ZEPPELIN-1179] Append scala version to maven artifacts

lresende:
### What is this PR for?
Now that we have support for both Scala 2.10 and 2.11 we need to append Scala version to the maven artifact id in order to properly publish artifacts for both Scala releases. 

### What type of PR is it?
[Improvement]

### What is the Jira issue?
[ZEPPELIN-1179](https://issues.apache.org/jira/browse/ZEPPELIN-1179)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,35 +35,35 @@ matrix:
   include:
     # Test all modules with scala 2.10
     - jdk: "oraclejdk7"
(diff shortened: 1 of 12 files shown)


lresende:
This assumes that #747 is merged.

minahlee:
CI failures seems irrelevant. LGTM merging if there is no more discussion

minahlee:
I have one question before merging, is there any special reason that ignite doesn't have `_2.11` suffix?

lresende:
@minahlee Good catch, let me fix that.

lresende:
@minahlee I have included the missing modules, and also updated the build to change the artifact version to _2.11 when performing the build based on Scala 2.11.

The build is still failing with the 10 minutes issue and other issue that seems not related.

minahlee:
Looks good to me. Merge if there is no more discussion

PR #1188, opened 2016-07-15
[ZEPPELIN-1183] Replace $window.location by $location

corneadoug:
### What is this PR for?
There is a native service used across the app to take care of re-routing to different page: `$location`, however there has been a few times where `$window.location` is used instead, and people tend to do it in new PR too.

This convert the `$window.location` from the code to use `$location`.

Only one was kept in `app.controller.js` for Iframe, so it doesn't flicker.

### What type of PR is it?
Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1183

You can try to reproduce the couple of cases that redirection is needed,
but overall there shouldn't be any issue from that change.

### Questions:

### DIFF
--- a/zeppelin-web/src/app/configuration/configuration.controller.js
+++ b/zeppelin-web/src/app/configuration/configuration.controller.js
@@ -13,8 +13,8 @@
  */
 'use strict';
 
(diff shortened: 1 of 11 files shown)


corneadoug:
CI is green, merging if there is no discussions

PR #1189, opened 2016-07-15
ZEPPELIN-1185. ZEPPELIN_INTP_JAVA_OPTS should not use ZEPPELIN_JAVA_OPTS

zjffdu:
### What is this PR for?

Don't use ZEPPELIN_JAVA_OPTS as the default value of ZEPPELIN_INTP_JAVA_OPTS

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1185

Tested manually. By exporting the following variable, I can debug zeppelin server correctly and remote interpreter process can ran successfully. (Before this PR, the remote  interpreter process will fail to launch because it would also listen the same debug port)
```
export ZEPPELIN_JAVA_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
```


Leemoonsoo:
Thanks for contribution!

`ZEPPELIN_JAVA_OPTS` used to set JVM options both ZeppelinServer and InterpreterProcess.
Later `ZEPPELIN_INTP_JAVA_OPTS` has been introduced. To keep compatibility with previous configuration, `ZEPPELIN_INTP_JAVA_OPTS` designed to overrides `ZEPPELIN_JAVA_OPTS` for InterpreterProcess.
So previous version of `conf/zeppelin-env.sh` can work without modification. 
That's why InterpreterProcess looks like implicitly uses `ZEPPELIN_JAVA_OPTS`.

I don't have any preference of either keeping current behavior or change it.

But if you'd like to change, I think `ZEPPELIN_INTP_MEM` also need to be changed, to make configuration consistent.

Also, this change makes `conf/zeppelin-env.sh` incompatible to previous version. So i think it's great idea that guide user how `ZEPPELIN_JAVA_OPTS` and `ZEPPELIN_INTP_JAVA_OPTS` are changed through https://github.com/apache/zeppelin/blob/master/docs/install/upgrade.md.

zjffdu:
@Leemoonsoo  I pushed another commit to add migration section in upgrade.md, please help review. Thanks.

docs/install/upgrade.md
```diff
@@ -43,4 +43,10 @@ So, copying `notebook` and `conf` directory should be enough.
 
    ```
    bin/zeppelin-daemon.sh start
-   ```
\ No newline at end of file
+   ```
+
+# Migration Guide
```
AhyoungRyu: Could you change `#` -> `##` ? Since `#` is used only for the docs title in many other pages.

docs/install/upgrade.md
```diff
@@ -43,4 +43,10 @@ So, copying `notebook` and `conf` directory should be enough.
...
+
+## Upgrading From Zeppelin 0.6 to 0.7
```
AhyoungRyu: Then this one also can be `###` :)

docs/install/upgrade.md
```diff
@@ -43,4 +43,10 @@ So, copying `notebook` and `conf` directory should be enough.
...
+
+ - From 0.7, we don't use ZEPPELIN_JAVA_OPTS as default value of ZEPPELIN_INTP_JAVA_OPTS and also the same for ZEPPELIN_MEM/ZEPPELIN_INTP_MEM. If user want to configure the jvm opts of interpreter process, please set ZEPPELIN_INTP_JAVA_OPTS and ZEPPELIN_INTP_MEM explicitly.
```
AhyoungRyu: @zjffdu 
Currently this line is rendered  as like below because of ` "_" `
screen shot 2016-07-26 at 10 37 39 am

So how about attaching back quotation mark to back & forth of each variable names? I mean 
```

 - From 0.7, we don't use `ZEPPELIN_JAVA_OPTS` as default value of `ZEPPELIN_INTP_JAVA_OPTS` and also the same for `ZEPPELIN_MEM/ZEPPELIN_INTP_MEM`. If you want to configure the jvm opts of interpreter process, please set `ZEPPELIN_INTP_JAVA_OPTS` and `ZEPPELIN_INTP_MEM` explicitly.
```

Then it will be rendered like this 
screen shot 2016-07-26 at 10 37 05 am

zjffdu:
Thanks @AhyoungRyu , I have updated the doc.

zjffdu:
BTW, I will do the windows related change after we make consensus on this approach.

zjffdu:
\\cc @Leemoonsoo  Ready for review.

Leemoonsoo:
Change looks good.
Could you trigger CI once again and see if it becomes green?

zjffdu:
ping @Leemoonsoo  in case you miss it.

bzz:
I guess @khalidhuseynov did not mean to refer this PR, but just used short-cut to name the JIRA issue.

Leemoonsoo:
Thanks @zjffdu. 
LGTM and merge if there's no more discussion.

prabhjyotsingh:
I'll merge this in sometime soon.

PR #1191, opened 2016-07-15
[ZEPPELIN-1186] Prevent NPE when unbind interpreter

minahlee:
### What is this PR for?
Unbinding interpreter from note throws NPE when paragraph repl is not set.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1186](https://issues.apache.org/jira/browse/ZEPPELIN-1186)

1. Create notebook
2. type '%fake ' including space at the end
3. Unbind spark interpreter from note binding and hit save
4. Unbind all interpreter and hit save
5. See if you can save the interpreter bind without error

**Before**
before

**After**
after

### Questions:


astroshim:
LGTM

minahlee:
Thanks for the review. Merging if there is no more discussion

PR #1192, opened 2016-07-15
[ZEPPELIN-1189] Get note revision websocket api

khalidhuseynov:
### What is this PR for?
Adds websocket api for getting note revision.

### What type of PR is it?
Improvement | Feature

### Todos
* [x] - add backend websocket handle
* [x] - add frontend call

### What is the Jira issue?
[#1189](https://issues.apache.org/jira/browse/ZEPPELIN-1189)

green CI (can be tested once frontend implemented)

### Questions:


bzz:
Looks like CI fails, could you please followup on the reason?

One more question - it looks like this PR not only add websocket API but change backend API and modify every NotebookRrpo implementation.

Is there a way to avoid such changes in a single PR? Or is there a strong reason for breaking API here?

khalidhuseynov:
@bzz fixed the issue with CI.
regarding change in api, you're right. it's better not to change api as a part of different PR, so I reverted it back. The main point in changing `Revision` object into just `String` representing `revisionId` was that `revisionId` can completely represent `Revision` object. However we don't need to address it right now, and it can be addressed as an improvement with addressing `pros` and `cons`.

bzz:
Sounds great, thank you for taking care!

CI is still failing though :\  Please feel free to followup the failure here and I will be happy to help, if needed!

khalidhuseynov:
@bzz addressed the changes and CI is green

bzz:
Looks great to me! Thank you for taking care. Megin if there is no further discussion

PR #1197, opened 2016-07-16
[ZEPPELIN-1196] Fix for bug ZEPPELIN-1196

SachinJanani:
### What is this PR for?
Fixed issue related to connecting to remote running interpreter process with multiple interpreters in interpreter group throws illegal thread state exception

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1196] https://issues.apache.org/jira/browse/ZEPPELIN-1196

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterProcessTest.java
+++ b/zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterProcessTest.java
@@ -109,11 +109,12 @@ public class RemoteInterpreterProcessTest {
   public void testRemoteInterpreterWithMultipleInterpreterInGroup() throws TException, InterruptedException {
     RemoteInterpreterServer server = new RemoteInterpreterServer(3679);
     server.start();
-    boolean running = false;
     long startTime = System.currentTimeMillis();
-    while (System.currentTimeMillis() - startTime < 10 * 1000) {
+    /*If RemoteInterpreterServer didn't start within 30 seconds than this test may fail
+     * which might be due to issue in RemoteInterpreterServer
+     */
+    while (System.currentTimeMillis() - startTime < 30 * 1000) {
       if (server.isRunning()) {
-        running = true;
         break;
       } else {
         Thread.sleep(200);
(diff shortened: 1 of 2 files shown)


SachinJanani:
One of the travis check is failing but its not related to my changes

SachinJanani:
Can someone please review this PR

jongyoul:
@SachinJanani Hi, can you add a test case for it? It would help understand and test this PR.

SachinJanani:
@jongyoul Thanks for reviewing.As suggested I have added a test case for this PR

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterProcessTest.java
```diff
@@ -103,4 +103,35 @@ public void testStartStopRemoteInterpreter() throws TException, InterruptedExcep
     assertEquals(1, rip.reference(intpGroup));
     assertEquals(true, rip.isRunning());
   }
+  
+  
+  @Test
+  public void testRemoteInterpreterWithMultipleInterpreterInGroup() throws TException, InterruptedException {
+    RemoteInterpreterServer server = new RemoteInterpreterServer(3679);
+    server.start();
+    boolean running = false;
```
jongyoul: this variable looks not used at all. Could you please remove it?

jongyoul:
@SachinJanani Thanks for the quick fix, but if it fails to launch RemoteInterpreterServer with 10 seconds, this test always fails, isn't it? how to deal with that case? I think we need to make it more concrete. How about you?

SachinJanani:
@jongyoul  Yes you are right if RemoteInterpreterServer fails to launch within 10 seconds then this test will fail,but we need to have some threshold for that because if say there is some bug in RemoteInterpreterServer which causes it to hang then our test will never complete.Should I increase it to 20 seconds.What you think?

SachinJanani:
@jongyoul  One more thing I found that as per ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT  property default timeout that we consider is 30 seconds.So I think setting it to 30 seconds will make more sense.

jongyoul:
@SachinJanani 30 seconds would be realistic. And could you please leave memo that it may occurs a potential bug when it fails launching it?

SachinJanani:
Thanks @jongyoul. Sure will make the change also will add a comment in the code about about the 30 seconds timeout

jongyoul:
LGTM. Merging if there's no more discussion.

SachinJanani:
Can someone please merge this PR as I don't have permission

jongyoul:
I'll merge it after 24 hours.

jongyoul:
@SachinJanani I tried to merge it but this is based on branch-0.6. Could you please rebase from current master? I've tried to do it but I occurs error about merge conflict. Can you handle it? And for the next time, you'd better contribute the codes to avoid this situation.

SachinJanani:
@jongyoul This changes are only for branch-0.6 as master does not have this files (even this code) anymore due to helium changes.Also i think issue might not occur for zeppelin-0.7.So this is fix will only be for zeppelin-0.6.1.I think it should be only be merged in branch-0.6.

minahlee:
@jongyoul @SachinJanani I managed it to be merged only to branch-0.6. Will close this PR manually.

SachinJanani:
Thanks @minahlee

PR #1200, opened 2016-07-18
[ZEPPELIN-1191] Supported legacy way to run paragraph with group name only

jongyoul:
### What is this PR for?
Preserving legacy way to run paragraph when users use group name only

### What type of PR is it?
[Improvement]

### Todos
* [x] - Added the way to find a interpreter with group name at last

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1191

### Questions:


jongyoul:
https://travis-ci.org/apache/zeppelin/builds/150652801

jongyoul:
Merge if there's no more discussion

PR #1204, opened 2016-07-19
[Zeppelin-1089] Replace JsHint with Eslint

corneadoug:
### What is this PR for?
This PR replace JsHint with Eslint, the original configuration was ported from jsHint using [Polyjuice](https://www.npmjs.com/package/polyjuice), then the settings were tuned to match exactly.
For this PR, the goal is not to add new rules, but only to replace the system with the same rules.

### What type of PR is it?
Improvement

### Todos
* [x] - Remove .jshintrc

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1089

You can simply build the code.
You can try to add an unused variable in the js code to see an error.

### Questions:


corneadoug:
PR done, triggering CI again

corneadoug:
CI is green!

corneadoug:
Merging if there is no discussions

PR #1205, opened 2016-07-19
[ZEPPELIN-1212] User impersonation support in JDBC

prabhjyotsingh:
### What is this PR for?
Add impersonation support to JDBC interpreters, in addition to Kerberos Authentication to improve auditability in all JDBC interpreters.

### What type of PR is it?
[Bug Fix | Improvement]

### What is the Jira issue?
* [ZEPPELIN-1212](https://issues.apache.org/jira/browse/ZEPPELIN-1212)

 - zeppelin.jdbc.auth.type = KERBEROS
 - zeppelin.jdbc.principal = principal value
 - zeppelin.jdbc.keytab.location = keytab location
 - enable shiro authentication via shiro.ini

Now try and run any of hive's query (say show tables) it should return with valid results/errors depending on user permission.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -167,19 +167,11 @@ public class JDBCInterpreter extends Interpreter {
 
     logger.debug("propertiesMap: {}", propertiesMap);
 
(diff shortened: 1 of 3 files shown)


rconline:
LGTM, at some stage we should start thinking about connection pooling. But that can be a different PR.

prabhjyotsingh:

PR #1206, opened 2016-07-20
ZEPPELIN-1199. Need to login using keytab and principal before creating SparkContext in secured cluster

zjffdu:
### What is this PR for?
Need to login using keytab and principal before creating SparkContext in secured cluster. 

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1199

Manually verified in a secured cluster. 

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -33,11 +33,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import com.google.common.base.Joiner;
 
-import com.google.common.base.Preconditions;
-import com.google.common.reflect.TypeToken;
-import com.google.gson.Gson;
 import org.apache.hadoop.security.UserGroupInformation;
-import org.apache.spark.HttpServer;
 import org.apache.spark.SparkConf;
 import org.apache.spark.SparkContext;
 import org.apache.spark.SparkEnv;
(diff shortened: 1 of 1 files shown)


zjffdu:
Might need to check whether it works in pyspark/sparkr.

zjffdu:
pyspark and sprakr both create SparkContext, so this fix also works in pyspark & sparkr

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -337,6 +339,15 @@ public SparkContext createSparkContext() {
       conf.set("spark.yarn.isPython", "true");
     }
 
+    if (conf.contains("spark.yarn.keytab") && conf.contains("spark.yarn.principal")) {
+      try {
+        String keytab = conf.get("spark.yarn.keytab");
```
bikassaha: Why are we logging on with spark keytab instead of zeppelin keytab?

zjffdu: This is not zeppelin keytab. It is just keytab specified by zeppelin. Every interpreter has its own setting, and this is just one of the setting.

prabhjyotsingh:
@zjffdu can you resolve conflict ?

zjffdu:
@prabhjyotsingh Conflicts resolved.

prabhjyotsingh:
LGTM

zjffdu:
Close it and merge it to #1237

PR #765, opened 2016-03-07
[ZEPPELIN-157] GAdding Map Visualization for Zeppelin

Madhuka:
### What is this PR for?
Adding Map Visualization for Zeppelin using Leaflet[1]

[1] http://leafletjs.com/

### What type of PR is it?
Improvement

### Todos

- [x] Checking compatible license 
- [x] Adding chart library (leafletjs)
- [x] Adding new button for chart mapping
- [x] Loading GIS Mapp in Zeppelin

### What is the Jira issue?
[ZEPPELIN-157](https://issues.apache.org/jira/browse/ZEPPELIN-157)

### How should this be tested?
`import org.apache.commons.io.IOUtils
import java.net.URL
import java.nio.charset.Charset

// Zeppelin creates and injects sc (SparkContext) and sqlContext (HiveContext or SqlContext)
// So you don't need create them manually

// load map data
val myMapText = sc.parallelize(
    IOUtils.toString(
        new URL("https://gist.githubusercontent.com/Madhuka/74cb9a6577c87aa7d2fd/raw/2f758d33d28ddc01c162293ad45dc16be2806a6b/data.csv"),
        Charset.forName("utf8")).split("\n"))
        
case class Map(Country:String, Name:String, lat : Float, lan : Float, Altitude : Float)

val myMap = myMapText.map(s=>s.split(",")).filter(s=>s(0)!="Country").map(
    s=>Map(s(0),
           s(1),
           s(2).toFloat,
           s(3).toFloat,
           s(4).toFloat
          )
)

// Below line works only in spark 1.3.0.
// For spark 1.1.x and spark 1.2.x,\n// use myMap.registerTempTable(\"myMap\") instead.\n
myMap.toDF().registerTempTable("myMap")`

### Screenshots (if appropriate)
Can found in #152 

### Questions:
* Does the licenses files need update?No
* Is there breaking changes for older versions?No
* Does this needs documentation? #152 contains documentation (I will add new doc and sample in new PR)

PR #1210, opened 2016-07-20
[ZEPPELIN-1220] Add geographical map as visualization option

mwpenny:
### What is this PR for?
Added a 2D [Esri map](https://developers.arcgis.com/javascript/) as a graph option for paragraphs. This makes it easier to visualize relations in geographical data.
Pins are plotted at points in specified longitude/latitude columns and customizable column values are shown when pins are clicked. Is this of interest?

The angular-esri-map module is used, which is released under the Apache 2.0 license, and the JavaScript API is free to use (see [https://developers.arcgis.com/terms/](https://developers.arcgis.com/terms/)).

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-1220](https://issues.apache.org/jira/browse/ZEPPELIN-1220)

Execute a query in a paragraph and select the map pin from the chart type selector.
Set the longitude/latitude (and optionally, pin info) columns, then interact with the map.

### Screenshot
screenshot

### Questions:
* Does the licenses files need update? __yes.__ For angular-esri-map.
* Is there breaking changes for older versions? __no__
* Does this needs documentation? __potentially.__ This feature follows the same layout as the other visualization modes, so users will likely be able to use it easily. However, screenshots of some supported visualizations modes are present in the current documentation. If this PR is merged, it may be desirable to display the map option as well. Interestingly the scatter chart is not included in screenshots in the documentation.

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1219,7 +1219,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       getMapPins(function(pins) {
         createPinMapLayer(pins, function(pinLayer) {
           $scope.map.map.add(pinLayer);
-          if (pins.length > 0) {
+          if (pinLayer.source.length > 0) {
             $scope.map.goTo(pinLayer.source);
           }
         });
@@ -1241,7 +1241,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
             basemap: 'streets'
           }),
           center: [-75.7325985, 45.4041593],  // Apption (lng, lat)
-          zoom: 16,
+          zoom: 14,
           pinRenderer: new SimpleRenderer({
             symbol: new SimpleMarkerSymbol({
               'color': [255, 0, 0, 0.5],
@@ -1258,7 +1258,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
             })
           })
         });
-        updateMapPins();
+        $scope.map.then(updateMapPins);
       });
     };
 
@@ -1461,10 +1461,27 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       }
     }
 
-    if (!$scope.paragraph.config.graph.map.lat && !$scope.paragraph.config.graph.map.lng &&
-            $scope.paragraph.result.columnNames.length > 1) {
-      $scope.paragraph.config.graph.map.lat = $scope.paragraph.result.columnNames[0];
-      $scope.paragraph.config.graph.map.lng = $scope.paragraph.result.columnNames[1];
+    /* try to find columns for the map logitude and latitude */
+    var i;
+    var col;
+    if (!$scope.paragraph.config.graph.map.lat) {
+      for (i = 0; i < $scope.paragraph.result.columnNames.length; ++i) {
+        col = $scope.paragraph.result.columnNames[i];
+        if (col.name.toUpperCase().indexOf('LAT') !== -1) {
+          $scope.paragraph.config.graph.map.lat = col;
+          break;
+        }
+      }
+    }
+
+    if (!$scope.paragraph.config.graph.map.lng) {
+      for (i = 0; i < $scope.paragraph.result.columnNames.length; ++i) {
+        col = $scope.paragraph.result.columnNames[i];
+        if (col.name.toUpperCase().indexOf('LONG') !== -1) {
+          $scope.paragraph.config.graph.map.lng = col;
+          break;
+        }
+      }
     }
   };
 
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -245,6 +245,12 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       config.graph.scatter = {};
     }
 
+    if (!config.graph.map) {
+      config.graph.map = {
+        pin: []
+      };
+    }
+
     if (config.enabled === undefined) {
       config.enabled = true;
     }
@@ -907,6 +913,8 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
 
       if (!type || type === 'table') {
         setTable($scope.paragraph.result, refresh);
+      } else if (type === 'map') {
+        setMap($scope.paragraph.result, refresh);
       } else {
         setD3Chart(type, $scope.paragraph.result, refresh);
       }
@@ -1131,6 +1139,9 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
     $timeout(retryRenderer);
   };
 
+  var setMap = function(data, refresh) {
+  };
+
   $scope.isGraphMode = function(graphName) {
     var activeAppId = _.get($scope.paragraph.config, 'helium.activeApp');
     if ($scope.getResultType() === 'TABLE' && $scope.getGraphMode() === graphName && !activeAppId) {
@@ -1193,6 +1204,24 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
     $scope.setGraphMode($scope.paragraph.config.graph.mode, true, false);
   };
 
+  $scope.removeMapOptionLat = function(idx) {
+    $scope.paragraph.config.graph.map.lat = null;
+    clearUnknownColsFromGraphOption();
+    $scope.setGraphMode($scope.paragraph.config.graph.mode, true, false);
+  };
+
+  $scope.removeMapOptionLng = function(idx) {
+    $scope.paragraph.config.graph.map.lng = null;
+    clearUnknownColsFromGraphOption();
+    $scope.setGraphMode($scope.paragraph.config.graph.mode, true, false);
+  };
+
+  $scope.removeMapOptionPinInfo = function(idx) {
+    $scope.paragraph.config.graph.map.pin.splice(idx, 1);
+    clearUnknownColsFromGraphOption();
+    $scope.setGraphMode($scope.paragraph.config.graph.mode, true, false);
+  };
+
   /* Clear unknown columns from graph option */
   var clearUnknownColsFromGraphOption = function() {
     var unique = function(list) {
@@ -1223,7 +1252,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       }
     };
 
-    var removeUnknownFromScatterSetting = function(fields) {
+    var removeUnknownFromFields = function(fields) {
       for (var f in fields) {
         if (fields[f]) {
           var found = false;
@@ -1235,7 +1264,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
               break;
             }
           }
-          if (!found) {
+          if (!found && !(fields[f] instanceof Array)) {
             fields[f] = null;
           }
         }
@@ -1250,7 +1279,10 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
     unique($scope.paragraph.config.graph.groups);
     removeUnknown($scope.paragraph.config.graph.groups);
 
-    removeUnknownFromScatterSetting($scope.paragraph.config.graph.scatter);
+    removeUnknownFromFields($scope.paragraph.config.graph.scatter);
+
+    removeUnknownFromFields($scope.paragraph.config.graph.map);
+    removeUnknown($scope.paragraph.config.graph.map.pin);
   };
 
   /* select default key and value if there're none selected */
@@ -1270,6 +1302,12 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       } else if ($scope.paragraph.result.columnNames.length === 1) {
         $scope.paragraph.config.graph.scatter.xAxis = $scope.paragraph.result.columnNames[0];
       }
+    }
+
+    if (!$scope.paragraph.config.graph.map.lat && !$scope.paragraph.config.graph.map.lng &&
+            $scope.paragraph.result.columnNames.length > 1) {
+      $scope.paragraph.config.graph.map.lat = $scope.paragraph.result.columnNames[0];
+      $scope.paragraph.config.graph.map.lng = $scope.paragraph.result.columnNames[1];
     }
   };
 
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -16,7 +16,7 @@
 angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $rootScope, $route, $window,
                                                                       $routeParams, $location, $timeout, $compile,
                                                                       $http, websocketMsgSrv, baseUrlSrv, ngToast,
-                                                                      saveAsService) {
+                                                                      saveAsService, esriLoader) {
   var ANGULAR_FUNCTION_OBJECT_NAME_PREFIX = '_Z_ANGULAR_FUNC_';
   $scope.parentNote = null;
   $scope.paragraph = null;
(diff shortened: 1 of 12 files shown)


corneadoug:
@mwpenny Thanks for the PR
There was already some attempts to include Leaflet before, however I think it is not yet to be finished. (No pivot to choose columns at least)
I think the community should be able to decide which map they prefer.

However, in term of license, I'm still not sure if this could be included or not.
Can you provide more infos on how the Javascript API is needed?

mwpenny:
@corneadoug, The Esri angular module is used to download and lazy load the JS API and require its various classes. These classes are then used to display the dynamically downloaded map tiles and render the different map layers (i.e., basemap + pin layer).

The angular module was developed by Esri and released under the Apache 2.0 license. They state on their website that ["The JavaScript API [downloaded by the angular module] is hosted by Esri and is available for free use"](https://developers.arcgis.com/javascript/3/jshelp/intro_accessapi.html).

Esri also has a public Apache 2.0-licensed repo of samples ([Esri/quickstart-map-js](https://github.com/Esri/quickstart-map-js)) which use the API.

ErikApption:
@corneadoug This is a great feature! ESRI is by far the most complete mapping component and the base maps are totally free (unlike gg or bing), plus their API is open source. What would be the next step to get this PR accepted?

mwpenny:
I've added basemap selection (see screenshot)
zeppelin09

corneadoug:
@mwpenny Sorry for the delay, I got a bit busy.
I will take a look this week-end. What you can do until then is to rebase with master since there is some conflicts right now.

mwpenny:
@corneadoug Done

corneadoug:
@mwpenny 

Alright, so after taking a look, here are my feedbacks:

#### Overall PR
* License of angular-esri-map will need to be added in [this file](https://github.com/apache/zeppelin/blob/master/zeppelin-distribution/src/bin_license/LICENSE) (That folder also contains single files if license needs to be copied)
* There is 2 files to commit: `index.html` and `karma.conf.js` which have changes after we build.
* I think it could be nice to have a page in the documentation, but let's make that in a different PR.

#### The Map
* The mouse scroll to zoom in the map should be deactivated, if you scroll down the notebook you might zoom out and then the map vanishes.
* Could we have the map unzoomed to show the world when loaded (Its fine to keep it centered on Canada ;) )
* The Map height do not scale when we make the paragraph bigger
* The Map is going out of the paragraph instead of being contained or fitting the size:
screen shot 2016-08-02 at 5 03 20 pm
* When I click on a Pin it moves me somewhere else
mappin


#### Options & Pivot
* Let's remove the `height:36px` style for the pivot area, I know it was copy pasted from other place in the code, but its better to have all of them with the same height.
* The BaseMap option isn't removed when we switch graphs, seems to be a mistake [here](https://github.com/mwpenny/zeppelin-esri/blob/esri/zeppelin-web/src/app/notebook/paragraph/paragraph-graphOptions.html#L31) with ng-if inside a class


I will spend more time on checking the code, and would leave inline comments if there is

mwpenny:
I have made the changes/fixes you mentioned and rebased with master. I can create the PR for the documentation after this is merged.

corneadoug:
@mwpenny I tested it again, its so much better now.

You will need to rebase your branch (conflicts, always conflicts)

We could probably improve the marker's popup by adding a bit of padding so that its not too white?
screen shot 2016-08-10 at 2 21 49 pm

Also depending on the paragraph height and the number of infos to show, it seems the popup header is hidden


screen shot 2016-08-10 at 2 24 36 pm

zeppelin-web/src/index.html
```diff
@@ -69,6 +69,7 @@
     <!-- endbuild -->
     <link rel="stylesheet" ng-href="assets/styles/looknfeel/{{looknfeel}}.css" />
     <link rel="stylesheet" href="assets/styles/printMode.css" />
+    <link rel="stylesheet" href="//js.arcgis.com/4.0/esri/css/main.css" />
```
corneadoug: Any way to have this CSS file locally? It seems it is not inside of angular-map-esri and the map is not rendering well without it.

If the license allows it, it would better to include it in the code `assets/styles/` since some people do not have access to internet when using Apache Zeppelin

corneadoug: Actually it doesn't matter, we can't use ESRI maps without internet

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1120,6 +1135,201 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
...
+      if ($scope.map && !refresh) {
+        $scope.map.map.destroy();
+        $scope.map.pinRenderer = null;
+        $scope.map = null;
+      }
+
+      // create map if not exists.
+      if (!$scope.map) {
+        if (!esriLoader.isLoaded()) {
+          esriLoader.bootstrap({
```
corneadoug: That Bootstrap function should probably be at Notebook level, otherwise it will be fetched for each paragraph

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1120,6 +1135,201 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
...
+          if ($scope.map.popup.selectedFeature) {
+            $scope.map.popup.location = $scope.map.popup.selectedFeature.geometry;
+          }
+        });
+        $scope.map.then(updateMapPins);
+      });
+    };
+
+    var renderMap = function() {
+      var mapdiv = angular.element('#p' + $scope.paragraph.id + '_map')
```
corneadoug: `angular.element('#p' + $scope.paragraph.id + '_map > div').get(0)` ?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1260,6 +1492,29 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
         $scope.paragraph.config.graph.scatter.xAxis = $scope.paragraph.result.columnNames[0];
       }
     }
+
+    /* try to find columns for the map logitude and latitude */
+    var i;
```
corneadoug: There is some duplicate code here, it could be refactored like this:

```
var findDefaultMapCol = function(settingName, keyword) {
      var col;
      if (!$scope.paragraph.config.graph.map[settingName]) {
        for (var i = 0; i < $scope.paragraph.result.columnNames.length; ++i) {
          col = $scope.paragraph.result.columnNames[i];
          if (col.name.toUpperCase().indexOf(keyword) !== -1) {
            $scope.paragraph.config.graph.map[settingName] = col;
            break;
          }
        }
      }
    };

    findDefaultMapCol('lat', 'LAT');
    findDefaultMapCol('lng', 'LONG');
```

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
mwpenny: I can do that but this way, the map JavaScript is only loaded if the user is actually using maps in their notebook (i.e., not loaded if there are no maps on the page, saving memory). It is not fetched for each paragraph since `esriLoader.isLoaded()` is called to check if the API has already been loaded.

Moving the `esriLoader.bootstrap()` call up to the notebook controller will have the same effect as it does currently.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
mwpenny: @corneadoug It is done the way it is currently so that the `#p_ID_map` div can have its height adjusted while also selecting its child div using cascading. I can do it without the cascading however that would require an additional DOM query:

e.g.,
```js
var mapdiv = angular.element('#p' + $scope.paragraph.id + '_map > div').get(0);
angular.element('#p' + $scope.paragraph.id + '_map').css('height', $scope.paragraph.config.graph.height);
```

versus (the current)

```js
var mapdiv = angular.element('#p' + $scope.paragraph.id + '_map')
                    .css('height', $scope.paragraph.config.graph.height)
                    .children('div').get(0);
```

mwpenny:
@corneadoug Good to hear!

The padding makes bottom-docked popups look a lot more clean.
zeppelin10

Good catch on the hidden popup header (before/after).
zeppelin11

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
corneadoug: Right esriLoader is a service

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
corneadoug: When I tested with `angular.element('#p' + $scope.paragraph.id + '_map > div').get(0)` only, I don't remember having any issue when resizing the paragraph.

mwpenny: I just tested it. My mistake, it isn't resizing that's the problem, it's the initial creation of the map div. Omitting the size change code results in a height of 0. The table and d3 graph modes use similar code. It should also be possible to use CSS, but I was following the convention used by the other graph modes

mwpenny: @corneadoug You were able to display the map with only `angular.element('#p' + $scope.paragraph.id + '_map > div').get(0)`?

corneadoug: Yes, I will test it again to make sure tomorrow

corneadoug:
@mwpenny latest changes I promise haha

If you resize the paragraph to the smallest height allowed, the popup is crushed and unusable, however there is plenty of space on top and bottom of the popup, maybe we could try to reduce it.
We can't support every width and height, but lets see if we can do something for that case
screen shot 2016-08-11 at 4 16 07 pm

If you don't have internet, it would be nice to show a small error message instead of the map
screen shot 2016-08-11 at 4 19 48 pm

Also without internet, the app takes a good 20-30seconds to load/show (even home page).
It is due to the page try to load the esri CSS from online.
I think this [plugin](https://github.com/GabrielDelepine/angular-css-injector) could help with that problem.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
corneadoug: Okay, I tried it again and it indeed doesn't work without the .css
I must have super power sometimes.
ysbkhg1

mwpenny:
@corneadoug I'm all to familiar with that magic when trying to reproduce bugs ;)

I've made the changes.

* __Small popup:__ The side and bottom margins of the popup are present due to yesterday's changes (differentiating the white of the popup from the white of the paragraph) and the extra-large top margin is from Esri's CSS. After decreasing the top margin the popup is still very cluttered but a bit more usable.
zeppelin13
Even with some extra CSS styling, I just don't think there's enough space to display the popup nicely in a paragraph that small. Other graph modes have a similar problem which users are likely already accustomed to:
zeppelin14



* __Internet error message:__ I went with a look similar to the other graph modes:
zeppelin12

* __Lazy loading:__ Similar to the way esriLoader loads the map JavaScript, the CSS is now lazy loaded the first time it is needed.

corneadoug:
@mwpenny With the way we include the css, there is still a risk of not being able to get the css, but having the JS file just fine.
Which end up in the map not showing, and no error message like the one we added in case there is no internet connection. I actually got that a few times with slow internet.

Except for that, LGTM, awesome work!
if anybody wants to try it, I use the test case of #765

mwpenny:
@corneadoug The CSS and JavaScript come from the same server. The CSS is timing out for you but not the JS? Or is the CSS just loading late?

corneadoug:
@mwpenny CSS loading late I guess, but there is also a possibility that the resource is not available for whatever reason.

It is not critical for now, but since there is a risk of map not showing for some users without any error message (creating more traffic in the mailing list), it would be nice to try to address  it in a different PR maybe.

karup1990:
@mwpenny Thanks for the awesome work.
I was trying out the visualizations.
I can see only the pins, the maps seems to not load.
Any idea what could be wrong?

mwpenny:
@karup1990 Can you send a screenshot of the map? If you look at the network requests in your browser's developer tools, are the requests to the ArcGIS servers successful (i.e., retrieval of the map JavaScript and CSS)? Here's a screenshot of what happens when I load the map (I'm using Firefox):
resload

Also check to make sure the map tile requests are succeeding. Screenshot from my browser:
tileload

karup1990:
screen shot 2016-08-17 at 9 13 00 pm
screen shot 2016-08-17 at 9 13 08 pm

Attaching screen capture.
Request seems fine, not sure what is preventing the browser from loading them

mwpenny:
@karup1990 Which browser are you using and what is its version?

karup1990:
I am using Google Chrome Version 52.0.2743.116 (64-bit) @mwpenny

karup1990:
BTw, I tried on Safari just now. And works great :)

mwpenny:
@karup1990 I have installed the same version of Chrome and am unable to reproduce the issue. Does clearing your browser cache and cookies help?

karup1990:
I cleared the browser cache and cookies. Still see only the pins @mwpenny

mwpenny:
@karup1990 This is very strange. I cannot reproduce the issue using the same browser version and your network doesn't seem to be the problem since it is working for you in Safari.

The only other thing I can think of is a browser setting or extension interfering somehow. Try running Chrome with the `--disable-extensions` argument. If that doesn't work, try creating a new Chrome user profile and using that temporarily.

corneadoug:
Could be related to the CSS maybe. Any chance it need the css before rendering the map?
@karup1990 no errors in the console?

mwpenny:
@corneadoug When I test it without the CSS, the tiles still load but they're not positioned properly (they extend outside of the paragraph).

karup1990:
@corneadoug There are no errors on console.
@mwpenny I disabled all extension, and tried view the page in an incognito window. Didnt help

mwpenny:
@karup1990 Are you able to test with a fresh Chrome installation (e.g,. using another computer on the same network)?

tabraiz12:
adding the following css fixed the issue for safari and chrome

```
.esri-bitmap {
   max-width: none;
}
```

karup1990:
Works fine for me after the change. Thanks @tabraiz12

Leemoonsoo:
Dear @mwpenny and Zeppelin PMC.

I have reviewed the license of ArcGIS Online https://developers.arcgis.com/terms/ and and http://www.esri.com/legal/software-license.
Although it's client javascript library is Apache 2, client does not functioning without ArcGIS online service, so the ArcGIS Online license does matter. However, the license does not look like type of license we can use. For example

> Generally, if you are not planning to monetize your app, say by charging a fee or by embedding revenue generating advertising in your app, you can make your app available to just about anyone.

and

> If you want to charge for access to apps you build using ArcGIS Online you have to have a paid subscription.

Apache 2 License allow monetize Apache Zeppelin. However ArcGIS Online license has incompatible restriction.

Also the client is loading arbitrary code (javascript) dynamically from their server on runtime and we can not guarantee what kind of code will be executed on our user's machine. 
see https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js#L1392

Therefore, unfortunately, we might need to remove this changes from our source tree.

Leemoonsoo:
Visualizations have incompatible license can be packaged as a [Helium application](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/Application.java) and loaded dynamically. Soon, there will be [highlevel api](https://issues.apache.org/jira/browse/ZEPPELIN-1003) to make visualization easier based on helium.
In the future, we'll have online registry for applications. So, please consider [use helium framework](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/development/writingzeppelinapplication.html).

volumeint:
I just published a helium visualization that uses Leaflet.  You can find it at (https://github.com/volumeint/helium-volume-leaflet).  To get it working, I first had to patch [webpack.config.js](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/resources/helium/webpack.config.js) in zengine 0.8.0-SNAPSHOT to add a test for png files.  Leaflet has two png files referenced in their CSS.


        test: /\.svg(\?\S*)?$/, loader: 'url-loader', }, {
        test: /\.png(\?\S*)?$/, loader: 'url-loader', }, {       <-- add this line
        test: /\.json$/, loader: 'json-loader' }, ],

Leemoonsoo:
@volumeint cool! Do you mind make a pullrequest with modification `test: /\.png(\?\S*)?$/, loader: 'url-loader'` ?

volumeint:
Just closing the loop.  Pull Request #2313 contained the webpack.config.js change.  It has been merged into master.

PR #1221, opened 2016-07-23
[MINOR] Change url in pom.xml files

minahlee:
### What is this PR for?
Set project url to `http://zeppelin.apache.org` in pom.xml files

### What type of PR is it?
Refactoring

### Questions:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,7 +202,7 @@ Zeppelin uses Travis for CI. In the project root there is .travis.yml that confi
 
 ```
 cd zeppelin-server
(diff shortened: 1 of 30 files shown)


minahlee:
merge if there is no more discussion

zjffdu:
lgtm

lresende:
If the URLs are all the same, and we set on the parent pom, wouldn't they be inherited ? In this case, we can actually remove all the child ones, and make future maintenance easier.

bzz:
Looks good to me. 
The idea for inheriting URL from parent to reduce duplication sounds really good.

minahlee:
I removed URLs from all child pom. Merge if there is no further comment

PR #1115, opened 2016-07-02
[ZEPPELIN-1099] Build and run Spark without spark-dependencies

jongyoul:
### What is this PR for?
When users use SPARK_HOME, they won't want to include spark-dependencies. This PR makes it possible and results reducing whole size of package.

### What type of PR is it?
[Feature]

### Todos
* [x] - Remove dependencies of spark-dependencies

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1099

### How should this be tested?
* Run `mvn clean package -DskipTests -pl '!spark-dependencies'`
* Set SPARK_HOME in zeppelin-env.sh
* Start Zeppelin
* Run `sc.version` in a paragraph and check the version
* Stop Zeppelin
* Set SPARK_HOME in a different version
* Start, check and stop

I've tested it with 1.6.1 and 1.5.2.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1195, opened 2016-07-16
[ZEPPELIN-759] Spark 2.0 support

Leemoonsoo:
### What is this PR for?
This PR implement spark 2.0 support based on #747.
This PR has approach from #980 which is reimplementing code in scala.

You can try build this branch

```
mvn clean package -Dscala-2.11 -Pspark-2.0 -Dspark.version=2.0.0-preview -Ppyspark -Psparkr -Pyarn -Phadoop-2.6 -DskipTests
```

### What type of PR is it?
Improvements

### Todos
* [x] - Spark 2.0 support
* [x] - Rebase after #747 merge
* [x] - Update LICENSE file
* [x] - Update related document (build)

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-759

### How should this be tested?

Build and try
```
mvn clean package -Dscala-2.11 -Pspark-2.0 -Dspark.version=2.0.0-preview -Ppyspark -Psparkr -Pyarn -Phadoop-2.6 -DskipTests
```

### Screenshots (if appropriate)
spark2

### Questions:
* Does the licenses files need update? yes
* Is there breaking changes for older versions? no
* Does this needs documentation? yes

PR #1222, opened 2016-07-23
[HOTFIX] add scala version to spark-dependencies artifact suffix

minahlee:
### What is this PR for?
After merging #1186, branch-0.6 build fails on travis because one dependency in `spark/pom.xml`  doesn't have valid artifact name which is `zeppelin-spark-dependencies`.(It should be `zeppelin-spark-dependencies_{scala.binary.version}`)
This dependency only exists in branch-0.6 not master after #1115.

### What type of PR is it?
Hot Fix 

Build and see if error occurs

### Questions:

### DIFF
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -66,7 +66,7 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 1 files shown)


minahlee:
Fixed by #1195

PR #1223, opened 2016-07-23
[WIP Zeppelin-945] Interpreter authorization

astroshim:
### What is this PR for?
This PR is for interpreter authorization.

### What type of PR is it?
Improvement

### Todos
* [ ] - only show the lock button if admin login.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-945

1. login as admin
2. then you can see the lock button on the right top as following image.
image
3. you can give users permission as check on the checkbox you want.
4. run paragraph job as user and see results if job has run or not properly.

image

### Questions:


Leemoonsoo:

Regarding UI, why don't we use the same UI with notebook authorization?
That can give the same user experience for all the authorization inside of Zeppelin.

Than it'll look something like

image

What do you think?

astroshim:
@Leemoonsoo Thank you for your review and I think your opinion makes sense. 
I'll change the UI.

prabhjyotsingh:
@Leemoonsoo I would say, since the use case for this is different than that of notebook, and idea for this is to map access to users with interpreters. 

While reading the title/description of this PR this is what I had in my mind



user setting

Also think about the case where user is getting auth using LDAP/AD where there are 100s of user, so this will require pagination. So, we need to have a group/profile for users that can have preset for grouping interpreters together.


user profile

And once this is set, IMO, then user who don't have permission to a particular interpreter, should not see that interpreter in the notebook page.

astroshim:
@prabhjyotsingh Thank you for share your great idea.
This PR is for interpreter authorization but your idea seems include user management.
I think it should be separated and I agree with @Leemoonsoo's idea of giving the same user experience for the authorization inside of Zeppelin.
and I also agree with @prabhjyotsingh's idea of the difference with notebook's because interpreter authorization is n:n. (it's hard to show with notebook's way.)
so What do you guys think about following way?

- when you want to grant permission of interpreters to all users 
image

- when you want to grant permission of interpreters to specific user
image

astroshim:
I'll do this at the another PR.

PR #1224, opened 2016-07-24
[ZEPPELIN-1228] Make z.show() work with Dataset

Leemoonsoo:
### What is this PR for?
z.show() does not work in spark 2.0

### What type of PR is it?
Bug Fix

### Todos
* [x] - Make z.show() work with dataset
* [x] - add a unittest

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1228

```
case class Data(n:Int)
val data = sc.parallelize(1 to 10).map(i=>Data(i)).toDF
data.registerTempTable("data")
z.show(spark.sql("select * from data"))
```

### Questions:


Leemoonsoo:
CI failure is irrelevant to this change.
Merge it to master and branch-0.6 if there's no more discussion.

PR #1225, opened 2016-07-25
[Zeppelin-1229] Add cache buster for zeppelin-web build

corneadoug:
### What is this PR for?
In order to bust the cache when there is changes to the code of zeppelin-web after build, we are adding hashes at the end of the imported .js and .css files we use.
Every time there is changes in one of those file, the hash will change and the browser will use new file instead of cached one.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1229

* Build zeppelin-web once and write down the HASH in `dist/scripts/scripts.HASH.js`
* After making a change inside a .js file of the application and building again, the hash in `dist/scripts/scripts.HASH.js` should be changed.

### Questions:

### DIFF
--- a/zeppelin-web/Gruntfile.js
+++ b/zeppelin-web/Gruntfile.js
@@ -68,6 +68,17 @@ module.exports = function(grunt) {
       src: ['src/**/*.html']
     },
 
+    cacheBust: {
+      taskName: {
+        options: {
+          baseDir: '<%= yeoman.dist %>',
+          assets: ['scripts/**.js', 'styles/**.css'],
+          deleteOriginals: true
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
+1

corneadoug:
CI Failure unrelated, Merging if there is no more discussions

Leemoonsoo:
LGTM. Can we merge it into branch-0.6 as well?

corneadoug:
@Leemoonsoo in 0.6 or 0.6.1?

PR #1226, opened 2016-07-25
Fix logger class name to correct one

khalidhuseynov:
### What is this PR for?
This is to fix logger source class name for correct logs.

### What type of PR is it?
Bug Fix

### Todos
* [x] - change class name

### What is the Jira issue?
n/a

passing CI

### Questions:


jongyoul:
LGTM

PR #1228, opened 2016-07-26
ZEPPELIN-1175. AM log is not available for yarn-client mode

zjffdu:
### What is this PR for?
For now, we share the same class path for zeppelin server and remote interpreter process. The cause the issue that AM log is not available for yarn-client mode because the yarn app also use the `ZEPPELIN_HOME/conf/log4j.properties` which is only for zeppelin server. So this PR just distinguish the CLASSPATH of zeppelin server and remote interpreter process. I use `ZEPPELIN_INTP_CLASSPATH` to represent the classpath of remote interpreter process and won't include `ZEPPELIN_HOME/conf/log4j.properties` in `ZEPPELIN_INTP_CLASSPATH`. 

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1175

Tested manually. 

### Questions:
* Is there breaking changes for older versions? Yes, if user put custom config file (hive-site.xml) under ZEPPELIN_HOME/conf, it won't take effect after this PR 

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -74,10 +74,23 @@ function addEachJarInDirRecursive(){
   fi
 }
 
+function addEachJarInDirRecursiveForIntp(){
+  if [[ -d "${1}" ]]; then
+    for jar in $(find -L "${1}" -type f -name '*jar'); do
+      ZEPPELIN_INTP_CLASSPATH="$jar:$ZEPPELIN_INTP_CLASSPATH"
+    done
+  fi
+}
 
 function addJarInDir(){
   if [[ -d "${1}" ]]; then
     ZEPPELIN_CLASSPATH="${1}/*:${ZEPPELIN_CLASSPATH}"
+  fi
+}
+
+function addJarInDirForIntp() {
+  if [[ -d "${1}" ]]; then
+    ZEPPELIN_INTP_CLASSPATH="${1}/*:${ZEPPELIN_INTP_CLASSPATH}"
   fi
 }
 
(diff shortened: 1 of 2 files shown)


zjffdu:
@Leemoonsoo  
* `$ZEPPELIN_HOME/interpreter/<interpreter_name>` is not in the classpath, actually the jars under this folder are in the classpath. Although I can put all the files under this folder on the classpath to allow user to put custom config file there, I just feel it might not be the best approach as it would introduce other issues.
 - what config file we should use if user also specify `hive-site.xml` under `$SPARK_CONF_DIR` ? 
 - what property we should use if there's conflict between these 2 `hive-site.xml` ? Do we allow override ?

So in this PR, I don't allow user to put custom config file under `$ZEPPELIN_HOME/interpreter/<interpreter_name>`, I would suggest user to put it outside ZEPPELIN ($SPARK_CONF_DIR or $SPARK_HOME/conf), seems we use this approach for hbase. 
* Windows related change and Migration doc are not done yet, will do that after we make consensus on this approach.

zjffdu:
Build fails due to some network issue. 
```
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.25:npm (npm install) on project zeppelin-web: Failed to run task: 'npm install --color=false' failed. (error code 1) -> [Help 1]
[ERROR] 
```

prabhjyotsingh:
LGTM!

prabhjyotsingh:
CI failure looks unrelated, merging this if no more discussion.

PR #1236, opened 2016-07-27
[ZEPPELIN-1237] Auto-suggestion of notebook permissions should list roles as well

prabhjyotsingh:
### What is this PR for?
Auto-suggestion of notebook permissions should list roles as well

### What type of PR is it?
[Improvement]

### Todos
* [x] - Fix test case (selenium)
* [x] - select2 in bower.json

### What is the Jira issue?
* [ZEPPELIN-1237](https://issues.apache.org/jira/browse/ZEPPELIN-1237)

Search for group/roles in notebook permission, it should get listed

screen shot 2016-07-27 at 7 13 44 pm

### Questions:


zeppelin-server/src/main/java/org/apache/zeppelin/rest/SecurityRestApi.java
```diff
@@ -29,6 +30,7 @@
 import org.apache.zeppelin.server.JsonResponse;
 import org.apache.zeppelin.ticket.TicketContainer;
 import org.apache.zeppelin.utils.SecurityUtils;
+import org.eclipse.jetty.util.StringUtil;
```
r-kamath: unused

zeppelin-server/src/main/java/org/apache/zeppelin/rest/SecurityRestApi.java
```diff
@@ -18,6 +18,7 @@
 package org.apache.zeppelin.rest;
 
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.shiro.realm.Realm;
 import org.apache.shiro.realm.jdbc.JdbcRealm;
 import org.apache.shiro.realm.ldap.AbstractLdapRealm;
```
r-kamath: unused. not a part of this pr. can you please cleanup this one too ?

r-kamath:
@prabhjyotsingh 404 for `/userlist` api call when clicked on text fields.
user: anonymous
screen shot 2016-07-27 at 10 46 03 pm

prabhjyotsingh:
@r-kamath have implement your feedback.

felixcheung:
is it a bit of an information disclosure security issue though, that we should all user name when someone type the first character?

prabhjyotsingh:
I've looked up few products of which the popular ones Outlook (app) limit to 8 results, Outlook (web i.e. owa) limit to 5 results and JIRA does not limits. 

I don't have a strong opinion on this, can change according to recommendation.

rconline:
@felixcheung i think this will continue to remain a problem, till the time we allow all user roles to setup/modify these permissions. Unless we can recognize and separate this as an admin (superuser) action, this seems okay to me. There will be cases when the LDAP user id's will end up in zeppelin/POSIX systems as 'a1890xx', based upon conversion rules, when users who setup these permissions will find it very hard to use. Hence for now, showing all users is a reasonable approach.

felixcheung:
I'm ok with it - maybe open a JIRA to track role-based access control on user names?

prabhjyotsingh:
I think a part of this problem is solved by https://github.com/apache/zeppelin/pull/993 ([ZEPPELIN-987](https://issues.apache.org/jira/browse/ZEPPELIN-987)). But, yes I agree this has scope for improvising. 

PR #1244, opened 2016-07-29
[ZEPPELIN-1251] Update interpreter artifact option list for netinst package

minahlee:
### What is this PR for?
Add scala 2.11 artifact options for netinst package

### What type of PR is it?
Documentation

### What is the Jira issue?
[ZEPPELIN-1251](https://issues.apache.org/jira/browse/ZEPPELIN-1251)

### Questions:

### DIFF
--- a/conf/interpreter-list
+++ b/conf/interpreter-list
@@ -19,15 +19,12 @@
 
 alluxio         org.apache.zeppelin:zeppelin-alluxio:0.6.1              Alluxio interpreter
 angular         org.apache.zeppelin:zeppelin-angular:0.6.1              HTML and AngularJS view rendering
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
What will happen with `./bin/install-interpreter.sh --all` ?
@minahlee Do you have any good idea to not install xxx_2.10 and xxx_2.11 together with `--all` option?

Leemoonsoo:
Instead of listing all xxx_2.10 and xxx_2.11 in `conf/interpreter_list`, i think just listing one version that is included in `zeppelin-bin-all` package would be fine, i think.

User can always manually install the other version of interpreter using `-n` and `-t` option of `bin/install-interpreter.sh` command.

minahlee:
Addressed your comment in 63509c0. Please review

Leemoonsoo:
Looks good to me

minahlee:

bzz:
Looks great to me. :shipit:

PR #1245, opened 2016-07-29
ZEPPELIN-1225. Errors before the last shell command are ignored

zjffdu:
What is this PR for?

The problem is that command "bash -c " will always return 0 as long as the last line of shell script run correctly. e.g the following command will run correctly without any error message.

hello
pwd
This PR will redirect stderr and stdout to the same place, and will display both the stderr and stdout to frontend just like what user see in the native shell terminal. So the output of above command will be as following

bash: hello: command not found
/Users/jzhang/github/zeppelin
What type of PR is it?

[Bug Fix]

What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1225
How should this be tested?

Unit test is added and also manually verify it on zeppelin notebook.

Screenshots (if appropriate)

Questions:

Does the licenses files need update? No
Is there breaking changes for older versions? No

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -79,8 +79,8 @@ public class ShellInterpreter extends Interpreter {
     }
     cmdLine.addArgument(cmd, false);
     DefaultExecutor executor = new DefaultExecutor();
-    ByteArrayOutputStream errorStream = new ByteArrayOutputStream();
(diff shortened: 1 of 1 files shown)


zjffdu:
@Leemoonsoo this is for branch-0.6, the test fail seems irrelevant.

Leemoonsoo:
@zjffdu Thanks. LGTM and merge into branch-0.6

jongyoul:
@zjffdu Could you please close this PR by yourself? It was merged 4 days ago but it's not affected into master then asfgit could not close it automatically.

PR #1247, opened 2016-07-29
ZEPPELIN-1254 Make get and save Interpreter bindings calls via websocket

r-kamath:
### What is this PR for?
This PR uses websocket to get and save interpreter bindings instead of rest api

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1254

Go to interpreter binding settings on notebook page and enable/disable interpreter or reorder and save. Simultaneously verify the network calls on browser's console.

n/a
### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -436,7 +436,8 @@ public class NotebookServer extends WebSocketServlet implements
     String noteID = (String) fromMessage.data.get("noteID");
     List<InterpreterSettingsList> settingList =
         InterpreterBindingUtils.getInterpreterBindings(notebook(), noteID);
-    broadcastInterpreterBindings(noteID, settingList);
+    conn.send(serializeMessage(new Message(OP.INTERPRETER_BINDINGS)
+        .put("interpreterBindings", settingList)));
   }
 
   public List<Map<String, String>> generateNotebooksInfo(boolean needsReload,
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -418,11 +418,14 @@ public class NotebookServer extends WebSocketServlet implements
   }
 
   public void saveInterpreterBindings(NotebookSocket conn, Message fromMessage) {
+    String noteId = (String) fromMessage.data.get("noteID");
     try {
       List<String> settingIdList = gson.fromJson(String.valueOf(
           fromMessage.data.get("selectedSettingIds")), new TypeToken<ArrayList<String>>() {
           }.getType());
-      notebook().bindInterpretersToNote((String) fromMessage.data.get("noteID"), settingIdList);
+      notebook().bindInterpretersToNote(noteId, settingIdList);
+      broadcastInterpreterBindings(noteId,
+          InterpreterBindingUtils.getInterpreterBindings(notebook(), noteId));
     } catch (Exception e) {
       LOG.error("Error while saving interpreter bindings", e);
     }
@@ -433,9 +436,7 @@ public class NotebookServer extends WebSocketServlet implements
     String noteID = (String) fromMessage.data.get("noteID");
     List<InterpreterSettingsList> settingList =
         InterpreterBindingUtils.getInterpreterBindings(notebook(), noteID);
-
-    conn.send(serializeMessage(new Message(OP.INTERPRETER_BINDINGS)
-        .put("interpreterBindings", settingList)));
+    broadcastInterpreterBindings(noteID, settingList);
   }
 
   public List<Map<String, String>> generateNotebooksInfo(boolean needsReload,
@@ -475,6 +476,12 @@ public class NotebookServer extends WebSocketServlet implements
 
   public void broadcastNote(Note note) {
     broadcast(note.id(), new Message(OP.NOTE).put("note", note));
+  }
+
+  public void broadcastInterpreterBindings(String noteId,
+                                           List settingList) {
+    broadcast(noteId, new Message(OP.INTERPRETER_BINDINGS)
+        .put("interpreterBindings", settingList));
   }
 
   public void broadcastNoteList(AuthenticationInfo subject) {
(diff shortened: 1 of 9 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -411,6 +418,48 @@ public void unicastUpdateNotebookJobInfo(NotebookSocket conn, Message fromMessag
...
+
+    List<InterpreterSetting> selectedSettings =
+        notebook().getBindedInterpreterSettings((String) fromMessage.data.get("noteID"));
+    for (InterpreterSetting setting : selectedSettings) {
+      settingList.add(new InterpreterSettingListForNoteBind(setting.getId(), setting.getName(),
+          setting.getInterpreterInfos(), true));
+    }
+
+    List<InterpreterSetting> availableSettings = notebook().getInterpreterFactory().get();
+    for (InterpreterSetting setting : availableSettings) {
```
prabhjyotsingh: I see duplicate effort in this, can you move this into a common function which both this and https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java#L197-L211 can use ?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -411,6 +418,48 @@ public void unicastUpdateNotebookJobInfo(NotebookSocket conn, Message fromMessag
             .put("notebookRunningJobs", response)));
   }
 
+  public void saveInterpreterBindings(NotebookSocket conn, Message fromMessage) {
+    try {
+      List<String> settingIdList = gson.fromJson(String.valueOf(
+          fromMessage.data.get("selectedSettingIds")), new TypeToken<ArrayList<String>>() {
+          }.getType());
+      notebook().bindInterpretersToNote((String) fromMessage.data.get("noteID"), settingIdList);
```
prabhjyotsingh: after saving can you 
`conn.send(serializeMessage(new Message(OP.INTERPRETER_BINDINGS).put("interpreterBindings", settingList)));`

jongyoul:
This PR looks like copy and paste from NotebookRestAapi. If you think websocket is a better way to do it, why don't you removing rest api? I don't think it's good to duplicate only. What's the benefit to change rest api to websocket?

jongyoul:
And please add test case if you are going further.

r-kamath:
@jongyoul Yes. this is copied from rest api. We should probably keep both options and use websocket on zeppelin UI. IMHO we all communications on the ui should happen via websockets and avoid multiple calls to server. I'll be cleaning up the duplication in my next commit. @prabhjyotsingh also had similar comment on code duplication.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -39,6 +39,7 @@
 import org.apache.zeppelin.interpreter.InterpreterGroup;
 import org.apache.zeppelin.interpreter.remote.RemoteAngularObjectRegistry;
 import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
+import org.apache.zeppelin.rest.message.InterpreterSettingListForNoteBind;
```
jongyoul: I think you'd better move this class into proper one.

jongyoul:
Thanks for the explaining it, but I think you'd better handle those changes into single PR, which is more meaningful. How do you think of it?

r-kamath:
@jongyoul I'm planning to make those changes in this PR

jongyoul:
That sounds great

bzz:
@r-kamath if you are planning to add more changes in this PR, do you think it will make sense to:
 - update TODO list in PR description with action items that you plan
 - add `WIP` prefix  (work in progress) to the PR title, and remove it when everything is done

In my experience, this will help to communicate your intent with reviewers, save time for them and overall speed up the process of merging these changes.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -411,6 +417,28 @@ public void unicastUpdateNotebookJobInfo(NotebookSocket conn, Message fromMessag
...
+      LOG.error("Error while saving interpreter bindings", e);
+    }
+  }
+
+  public void getInterpreterBindings(NotebookSocket conn, Message fromMessage)
+      throws IOException {
+    String noteID = (String) fromMessage.data.get("noteID");
+    List<InterpreterSettingsList> settingList =
+        InterpreterBindingUtils.getInterpreterBindings(notebook(), noteID);
+    broadcastInterpreterBindings(noteID, settingList);
```
prabhjyotsingh: Instead of broadcast can you just `conn.send(serializeMessage)`

r-kamath: fixed 55d95ec

prabhjyotsingh:
LGTM!

zeppelin-server/src/main/java/org/apache/zeppelin/types/InterpreterSettingsList.java
```diff
@@ -24,14 +24,14 @@
...
+public class InterpreterSettingsList {
   private String id;
   private String name;
   private boolean selected;
   private List<InterpreterInfo> interpreters;
 
-  public InterpreterSettingListForNoteBind(String id, String name,
-      List<InterpreterInfo> interpreters, boolean selected) {
+  public InterpreterSettingsList(String id, String name,
+                                 List<InterpreterInfo> interpreters, boolean selected) {
```
jongyoul: Vertial align is not recommended. Could you please revert this into original one?

jongyoul:
@r-kamath Don't you remove the rest api in this PR? If you do that, you don't have to make `getInterpreterBinding` as a static because it's only used in notebook server. I thought you would removed interpreter binding rest api entry points and changed all of them into websocket.

zeppelin-server/src/main/java/org/apache/zeppelin/types/InterpreterSettingsList.java
r-kamath: will update

r-kamath:
@jongyoul I think we have users for rest api. The change here is to make ui work via websocket while keeping the reset apis intact.

jongyoul:
@r-kamath But I thought you would removed rest api by your last comments. Do you think we should conserve both of them?

r-kamath:
@jongyoul Yes. we should keep both

jongyoul:
@r-kamath Actually, I don't know well how much efficient we change rest api to websocket because I'm not at front-end. I just wonder if we should keep two different entry point with exact same function.

jongyoul:
@r-kamath Of course, it's good to support some apis with websocket, basically.

jongyoul:
But I think this is not proper scope of this PR. We need to open discussion on it. Your PR looks good to me.

prabhjyotsingh:

PR #1249, opened 2016-07-30
[ZEPPELIN-1255] Add cast to string in z.show() for Pandas DataFrame

bustios:
### What is this PR for?
Casting data types in Pandas DataFrame to string in z.show()

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1255](https://issues.apache.org/jira/browse/ZEPPELIN-1255)

```
%python

import pandas as pd

df = pd.read_csv('https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data', header=None)
df.columns=[1, 2, 3, 'PetalWidth', 'Name']
z.show(df)

%python.sql

SELECT * FROM df  LIMIT 10
```

### Questions:

### DIFF
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -140,19 +140,19 @@ class PyZeppelinContext(object):
         """
         limit = len(df) > self.max_result
         header_buf = io.StringIO("")
-        header_buf.write(df.columns[0])
+        header_buf.write(str(df.columns[0]))
         for col in df.columns[1:]:
             header_buf.write("\t")
-            header_buf.write(col)
+            header_buf.write(str(col))
         header_buf.write("\n")
         
         body_buf = io.StringIO("")
         rows = df.head(self.max_result).values if limit else df.values
         for row in rows:
-            body_buf.write(row[0])
+            body_buf.write(str(row[0]))
             for cell in row[1:]:
                 body_buf.write("\t")
-                body_buf.write(cell)
+                body_buf.write(str(cell))
             body_buf.write("\n")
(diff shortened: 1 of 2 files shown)


felixcheung:
looks good!

felixcheung:
could you add tests for this?

bzz:
Looks great to me, @bustios could you add a simple test case that reproduce this case to the [python/src/test/java/org/apache/zeppelin/python/PythonInterpreterPandasSqlTest.java](https://github.com/apache/zeppelin/blob/master/python/src/test/java/org/apache/zeppelin/python/PythonInterpreterPandasSqlTest.java) ?

bustios:
Tranks for the revision @felixcheung, @bzz. I have already added the test case. 

I think it would be nice to show a column with the row number or index if the DataFrame had one.

bzz:
Looks great to me.

>I think it would be nice to show a column with the row number or index if the DataFrame had one.

@bustios This makes sense, please feel free to open a separate JIRA issue for such improvement. We just need to make sure that it's clearly distinguished and can not be confused by the user as a part of the dataset.

CI failure is due to flaky R integration tests, which are not related
```
Results :

Failed tests: 
  ZeppelinSparkClusterTest.sparkRTest:116 expected:<[[1] 3]> but was:<[<pre><code>Error in getSparkSession(): SparkSession not initialized
</code></pre>

<pre><code>Error in (function (classes, fdef, mtable) : unable to find an inherited method for function 'count' for signature '&quot;function&quot;'
</code></pre>]>

Tests run: 65, Failures: 1, Errors: 0, Skipped: 0
```

PR #1257, opened 2016-08-02
[Zeppelin-945] Interpreter authorization

astroshim:
### What is this PR for?
This PR is derived from https://github.com/apache/zeppelin/pull/1223

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-945

You can change user permission of interpreter at interpreter setting page.

result

### Questions:


zeppelin-web/src/app/interpreter/interpreter.html
```diff
@@ -192,6 +192,48 @@ <h3 class="interpreter-title">{{setting.name}}
...
+        </div>
+      </div>
+
+      <div class="col-md-12">
+        <!-- permissions -->
+        <div ng-show="setting.option.setPermission" class="permissionsForm">
+          <div>
+            <p>
+              Enter comma separated interpreters in the fields. <br />
+              Empty field (*) implies anyone can run the interpreters.
```
AhyoungRyu: It's a minor thing; how about `this interpreter` instead of `the interpreters`?

zeppelin-web/src/app/interpreter/interpreter.html
```diff
@@ -192,6 +192,48 @@ <h3 class="interpreter-title">{{setting.name}}
...
+          </span>
+        </div>
+      </div>
+
+      <div class="col-md-12">
+        <!-- permissions -->
+        <div ng-show="setting.option.setPermission" class="permissionsForm">
+          <div>
+            <p>
+              Enter comma separated interpreters in the fields. <br />
```
AhyoungRyu: Don't we need to type comma separated `users` list not `interpreters` in this field?

AhyoungRyu:
@astroshim Awesome. I saw many people have talked about this feature :) I tested your patch locally, it works well as expected. But I found one issue regarding user suggestion,

search_users

The suggestion list doesn't disappear after pressing `Enter` key. Could you check this out?

astroshim:
@AhyoungRyu Let me check. Thanks.

AhyoungRyu:
@astroshim 
>Does this needs documentation? no

I think we need a documentation for "Interpreter Authorization" feature like we have below docs pages.
 - [security/notebook_authorization.html](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/notebook_authorization.html)
 - [security/datasource_authorization.html](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/datasource_authorization.html)
 
But separated PR would be okay i think :)

astroshim:
@AhyoungRyu I fixed you pointed out and I'll create the document for this separated PR. 
Thanks.

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -85,7 +85,10 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
           $scope.suggestions = [];
         }
       }
-      if (event.keyCode < 37 || event.keyCode > 40) { // arrow buttons.
+      if (event.keyCode === 13) {
+        closeAutoComplete();
+      }
+      else if (event.keyCode < 37 || event.keyCode > 40) { // arrow buttons.
```
AhyoungRyu: Can we move this line to the above line? As google code style said [https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style).

AhyoungRyu:
@astroshim Tested again and checked the bug is gone. Thanks for the prompt fix 👍

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+    }
+
+    function convertToString(setting) {
+      setting.option.users = searchText.join();
+    }
+
+    var checkIfSelected = function() {
+      if (($scope.suggestions.length === 0) &&
+        ($scope.selectIndex < 0 || $scope.selectIndex >= $scope.suggestions.length) ||
+        ($scope.suggestions.length !== 0 && ($scope.selectIndex < 0 || $scope.selectIndex >= $scope.suggestions.length))
```
bzz: Long condition inside the `if` becomes quite hard to understand and follow here.

What do you think, could this be changed a bit to make it more readable i.e:

```javascript
if (suggestionSelected($scope.suggestions.length, $scope.selectIndex)) {
 //....
}

function suggestionSelected(len, i) {
  return (len === 0) && (i < 0 || i >= len) || 
         (len !== 0  && (i < 0 || i >= len))
}
```

And now it's easy to see how `suggestionSelected` can be further improved, keeping invariants.

Normalizing the parantheses according to [logical operator precedence in javascript](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence) we get

```javascript
return (len === 0 && (i < 0 || i >= len)) || 
       (len !== 0 && (i < 0 || i >= len))
}
```

which, in turn, is the same as

```javascript
return (i < 0 || i >= len) && (len === 0 || len !== 0)
```

and as the right side of the `&&` is always true for an integer, we get

```javascript
function suggestionSelected(len, i) {
 return i < 0 || i >= len
}
```

At the end, this might not worth extracting to a separate function, but this shows the benefits we can get from it.
The result is much simpler to understand and reason about.

@astroshim what do you think?

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
 
+    // auto complete related
+    $scope.suggestions = [];
+    $scope.selectIndex = -1;
+    var selectedUser = '';
+    var selectedUserIndex = 0;
+    var previousSelectedList = [];
+    var searchText = [];
+
+    function updatePreviousList() {
```
felizbear: `updatePreviousList` defined multiple times; for loop inside (or even the whole function) can be replaced with [_.clone](https://lodash.com/docs#clone)

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+      for (var i = 0; i < searchText.length; i++) {
+        previousSelectedList[i] = searchText[i];
+      }
+    }
+
+    function convertToString(setting) {
```
felizbear: the name `convertToString` implies the following signature `Array -> String`, at least in my opinion. This implementation, however as a signature `Object -> Void` which makes it hard for reader to understand what's happening. a more appropriate name would be something like `assignToUsers`; it should also at least receive stringified searchText as an argument

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+      setting.option.users = searchText.join();
+    }
+
+    var checkIfSelected = function() {
+      if (($scope.suggestions.length === 0) &&
+        ($scope.selectIndex < 0 || $scope.selectIndex >= $scope.suggestions.length) ||
+        ($scope.suggestions.length !== 0 && ($scope.selectIndex < 0 || $scope.selectIndex >= $scope.suggestions.length))
+      ) {
+        searchText[selectedUserIndex] = selectedUser;
+        $scope.suggestions = [];
```
felizbear: this is a boolean function, however it mutates two things inside its body; shall these mutations be factored out to the outer scope?

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+        return true;
+      } else {
+        return false;
+      }
+    };
+
+    $scope.checkKeyDown = function(event, setting) {
+      if (event.keyCode === 40) {
+        event.preventDefault();
+        if ($scope.selectIndex + 1 !== $scope.suggestions.length) {
```
felizbear: I think `$scope.selectIndex < $scope.suggestions.length` would be more clear

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -22,6 +22,171 @@ angular.module('zeppelinWebApp').controller('InterpreterCtrl',
...
+          $scope.selectIndex++;
+        }
+      } else if (event.keyCode === 38) {
+        event.preventDefault();
+
+        if ($scope.selectIndex - 1 !== -1) {
```
felizbear: `$scope.selectIndex > 0`

astroshim:
Thank you for detail review @bzz and @felizbear 
I'll fix the codes.

prabhjyotsingh:
@astroshim A lot of these complications is already reduced with https://github.com/apache/zeppelin/pull/1236 (which is already avail in master), take a look at it.

astroshim:
@prabhjyotsingh Great. Thank you for giving me the guidance.

astroshim:
\cc @AhyoungRyu @prabhjyotsingh @bzz @felizbear 
I fixed UI like #1236. Please review. 
Thank you!

prabhjyotsingh:
I'll try it out, in the mean while, can you update screen-shot in description.

astroshim:
@prabhjyotsingh I just updated screen-shot. Thank you for taking care of this.

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -13,7 +13,21 @@
  */
 'use strict';
 
-angular.module('zeppelinWebApp').controller('InterpreterCtrl',
+angular.module('zeppelinWebApp')
+.directive('interpreterDirective', function($timeout) {
```
prabhjyotsingh: can you move this directive to a different file `zeppelin-web/src/components/*`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -285,6 +298,24 @@ protected Object jobRun() throws Throwable {
...
+        if (replName.startsWith(intp.getName()) &&
+          intp.getOption().isSetPermission() &&
+          !hasPermission(authenticationInfo.getUser(), intp.getOption().getUsers())) {
+          logger.error("{} has no permission for {} ", authenticationInfo.getUser(), repl);
+          return new InterpreterResult(Code.ERROR, authenticationInfo.getUser() +
+            " has no permission for " + getRequiredReplName());
+/*
+          throw new RuntimeException(authenticationInfo.getUser() +
+            " has no permission for " + getRequiredReplName());
+*/
```
prabhjyotsingh: If this is not required can you remove it.

prabhjyotsingh:
Except for minor fixes, looks good.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
astroshim: I deleted it. Thanks.

zeppelin-web/src/app/interpreter/interpreter.controller.js
astroshim: I moved the directive. Thanks!

astroshim:
@prabhjyotsingh Thank you for your detail review. I fixed all that you mentioned.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
```diff
@@ -46,6 +49,17 @@ public void setHost(String host) {
     this.host = host;
   }
 
+  public boolean isSetPermission() {
```
bzz: This is quite an ambiguous naming - usually in Java `set\get` used as a convention for getters and setters, so may be here we better call it like `permissionIsSet()` or `permissionWasSet()` - in this way we keep the almos-eglish readability in case of the code like

```java
if (permissionIsSet()) {
 //....
}
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -285,6 +298,20 @@ protected Object jobRun() throws Throwable {
...
     }
 
+    if (this.user != null &&
+      !factory.getInterpreterSettings(note.getId()).isEmpty()) {
+      for (InterpreterSetting intp: factory.getInterpreterSettings(note.getId())){
+
+        if (replName.startsWith(intp.getName()) &&
+          intp.getOption().isSetPermission() &&
+          !hasPermission(authenticationInfo.getUser(), intp.getOption().getUsers())) {
+          logger.error("{} has no permission for {} ", authenticationInfo.getUser(), repl);
```
bzz: Just want to give you a perspective of somebody who reads this code: it is VERY hard to understand and reason what happens in both conditions here.

By looking at **first condition**:

```java
this.user != null && !factory.getInterpreterSettings(note.getId()).isEmpty()
```

I would assume that is some kind of pre-condition check, because we do not do anything in case it is `false`.
But if that is `true`, we do something for each `interpreter`. 

Then I would look at both conditions to understand, what they mean, starting from `this.user != null`. As a reader, I have no idea what is the case when `this.user` can be `null`. Is that because of some kind of error condition? Or is that because auth is not enabled? Or is that the case when user is not authorized? It's very hard to say, without spending a lot of time to understand all the cases

Same for `factory.getInterpreterSettings(note.getId()).isEmpty()` - what does it mean when interpreter settings are empty for a note? When can that happen? Why do we check for it here? Ah, after some thinking and looking for other code that uses it - I can see, that is some kind of collection!

And all this questions for reader could be avoided, and time saved, if we give a reader a clue in form of readable names for this conditions in english - we can extract this logic to the functions that self-describe it's purpose through the function names (you can think of it as a form of documentation):

```java
if (this.noteHasUser() && this.noteHasInterpreters()) {
  //...
}

boolean noteHasUser() {
  return this.user != null;
}

boolean noteHasInterpreters() {
  // BTW, this is used in 2 other places and looks like candidate for further simplification
  // return !getInterpreterListFor(this.note).isEmpty();
  return !factory.getInterpreyterSettings(note.getId()).isEmpty();
}
```

--------------------

By looking at **second condition**:

````java
replName.startsWith(intp.getName()) &&
 +          intp.getOption().isSetPermission() &&
 +          !hasPermission(authenticationInfo.getUser(), intp.getOption().getUsers())
```

as a reader, for me it's even harder to say, what case does this represent.

Judging by the loop and an error message - it looks this code does to things:
 - finds current interpreter by name
 - only for the current interpreter, it checks if user has access to it

So a function can be extracted here as well:

```java
boolean isUserAuthorizedToAccessInterpreter(String user,  InterpreterOption intpOpt)){
  return intpOpt.isSetPermission() &&
         !hasPermission(authenticationInfo.getUser(), intpOpt.getUsers())
}
```

But it can be even further simplified if we decouple logic for finding current `interp` in collection by `replName` to a new function `findInerpreterByName` (from the logic for checking if user is authorized to access the `interp`):

```java
InterpreterSettings findInerpreterByName(String name) {
  InterpreterSettings intp = null; // this would rather be `Option()` or `InterpreterSettings.NONE` or anything else type-safe
  for (InterpreterSetting i: factory.getInterpreterSettings(note.getId())) { // could re-use `getInterpreterListFor(this.note)` from above
    if (i.getName().startsWith(name)) {
      interp = i;
    }
 }
 return intp;
}
```

So, **final results**

*Before*

```java
if (this.user != null &&
  !factory.getInterpreterSettings(note.getId()).isEmpty()) {
  for (InterpreterSetting intp: factory.getInterpreterSettings(note.getId())){

    if (replName.startsWith(intp.getName()) &&
      intp.getOption().isSetPermission() &&
      !hasPermission(authenticationInfo.getUser(), intp.getOption().getUsers())) {
      return new InterpreterResult(...);
    }
 }
}
```

*After*

```java
if (this.noteHasUser() && this.noteHasInterpreters()) {
  InterpreterSetting intp = findInerpreterByName(replName);
  if (intp != null && 
      isUserAuthorizedToAccessInterpreter(authenticationInfo.getUser(), intp.getOption())) {
    return new InterpreterResult(...);
  }
}
```

I'm not native english speaker, but the resulting code very much looks like an english sentence. It explains the logic for the reader, who can follow the implementation details by looking at the new functions.

What do you think, does this make sense?

bzz:
@astroshim your changes look awesome to me!

There is one more improvement on Java side noted above, other than that - looks good to merge.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
astroshim: All your comments are totally makes sense and thank you very much for the detail review to spend your precious time.
I'll fix them. 
Thank you again.

astroshim:
@bzz please review and Thanks again.

bzz:
@astroshim changes looks great to me. 

There is one more comment above on `isSetPermission ` naming, please let me know if you want to address this one as well, or want to keep it like that!

Will merge to master right after that, and if there is no further discussion

astroshim:
@bzz I just fixed, I missed that.

bzz:
Thank you @astroshim !

Merging to master if there is no further discussion.

bzz:
CI fails on web-app \w networking issue :\ which is not relevant here

```
Could not transfer artifact org.codehaus.plexus:plexus-io:pom:2.1.3 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]
```

PR #1260, opened 2016-08-02
ZEPPELIN-1267. PySparkInterpreter doesn't work in spark 2.0

zjffdu:
### What is this PR for?
PySparkInterpreter doesn't work in spark 2.0 because pyspark and py4j is not distributed to executors.  This PR extract the setup staff for pyspark interpreter into method setupConfForPySpark and use it for both spark1 and spark2. But this is just a short term solution, as I think this should be handled by spark rather than zeppelin, here zeppelin duplicate part of spark's work. In the long term, I'd like to resolve it in `ZEPPELIN-1263`.

### What type of PR is it?
[Bug Fix]

### Todos
* https://issues.apache.org/jira/browse/ZEPPELIN-1263

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1267

Verify it manually. 

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -329,6 +329,7 @@ public class SparkInterpreter extends Interpreter {
       }
     }
 
+    setupConfForPySpark(conf);
     Class SparkSession = Utils.findClass("org.apache.spark.sql.SparkSession");
     Object builder = Utils.invokeStaticMethod(SparkSession, "builder");
     Utils.invokeMethod(builder, "config", new Class[]{ SparkConf.class }, new Object[]{ conf });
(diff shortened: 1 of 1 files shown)


bzz:
\cc @jongyoul for review

Leemoonsoo:
Looks good to me.

jongyoul:
LGTM.

bzz:
Let's merge if there is no further discussion!

minahlee:
@zjffdu I tried current master branch with master set to local[*], Spark standalone and yarn-client both with/without `SPARK_HOME` set. But somehow I was able to run pyspark interpreter without this patch. I don't know what I missed, could you tell me which environment should I try to make pyspark interpreter fail on master branch?

zjffdu:
I also use the latest master and build zeppelin with this command: 
```mvn clean package -Pspark-2.0 -Ppyspark -Psparkr -DskipTests
```, and then export `SPARK_HOME` to where spark-2.0 is located and run pyspark interpreter using `yarn-client` mode.

minahlee:
@zjffdu thank you for quick response, I was able to reproduce it and tested this patch fixes the issue!
Merge if there is no further discussion

PR #1272, opened 2016-08-03
Remove empty user when put permission

zhangmq:
### What is this PR for?
check empty users and delete it.  

### What type of PR is it?
[Bug Fix]

clear note permissions then your note can't read.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/RequestUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/RequestUtils.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/utils/RequestUtils.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Tools for correct request
+ */
+public class RequestUtils {
+  
+  public static void removeEmptyUser(HashSet set) {
+    for (Iterator iter = set.iterator(); iter.hasNext();) {
+      Object item = iter.next();
+      if (item == null || ((String) item).trim().isEmpty()) {
+        iter.remove();
```
jongyoul: If you assume item as String, how about assuming the first parameter as `HashSet<String>`? It's more reasonable. And it is definitely kind of utils but this is used by `NotebookRestApi` for now, I suggest this method is located `NotebookRestApi` class. How about it?

zhangmq: I really appreciate the suggestions you came up with, but seems this issue was solved inside NotebookAuthorization, I will remove my modification.

jongyoul:
I would want to see you again with nice contribution. Thanks.

PR #1273, opened 2016-08-03
[Zeppelin-1276] Fix Notebook Title Input

corneadoug:
### What is this PR for?
The Notebook Title Input would not disappear after being edited.
This is a problem made by its plugin elastic-input.
We originally used ng-if instead of ng-show so it doesn't break the Page Print, however that change made the scope value to toggle unusable.
So we added a trustworthy small controller on top of the component.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1276

* Click on the Notebook title
* Try to do: Enter or Esc or Click somewhere else
* The Input border should not be there anymore
* Do the same for the Paragraph
* If you do Print Page there shouldn't any titles duplicated at the top

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook-actionBar.html
+++ b/zeppelin-web/src/app/notebook/notebook-actionBar.html
@@ -13,10 +13,11 @@ limitations under the License.
 -->
 <div class="noteAction" ng-show="note.id && !paragraphUrl">
   <h3>
-    <div style="float: left; width: auto; max-width: 40%">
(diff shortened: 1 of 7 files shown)


bzz:
Tried locally and renaming notebook works now as expected.

Looks great to me.

corneadoug:
Trigger CI again

corneadoug:
Alright, CI is mainly green now. Only one Build Profile failing on Spark.

PR #1281, opened 2016-08-04
[ZEPPELIN-1290] Refactor Navbar Controller

corneadoug:
### What is this PR for?
This is a small refactoring to keep this Controller following the [ControllerAs with vm](https://github.com/johnpapa/angular-styleguide/tree/master/a1#controlleras-with-vm)
rules, that it was based on.

Here is a list of things that were changed and why:

* Most of the controller's $scope values & fct (except from the search q) where moved to the vm.The controller is using vm, so storing in $scope to share with the view is not needed.

* Functions or Vars that are not used in the view were removed from the vm. (kept private to the controller)

* $rootscope functions was moved to `app.js`. I think  the need for that function might need to be changed, but for the scope of this PR, we are just moving it to where the $rootScope values are declared.

* Gathering vm declaration before the functions and ordered alphabetically

* Re-order functions alphabetically

* Create `initController ` to regroup all the controller setup.

### What type of PR is it?
Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1290

You can Just verify that the below Navbar related features are still good:
* Search Form
* Connected Status
* Login button
* User Name and its dropdown menu
* Notebook list drop-down menu (and filer, folder inside of it)

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/integration/AuthenticationIT.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/integration/AuthenticationIT.java
@@ -124,7 +124,7 @@ public class AuthenticationIT extends AbstractZeppelinIT {
         userName + "')]")).click();
     ZeppelinITUtils.sleep(500, false);
     driver.findElement(By.xpath("//div[contains(@class, 'navbar-collapse')]//li[contains(.,'" +
-        userName + "')]//a[@ng-click='logout()']")).click();
+        userName + "')]//a[@ng-click='navbar.logout()']")).click();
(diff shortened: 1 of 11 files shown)


corneadoug:
Re-trigger CI

corneadoug:
@prabhjyotsingh I have Selenium test `testGroupPermission` failing on this PR, but can't find why.
Could you take a look since you made that test?

prabhjyotsingh:
Sure, let me take a look at it.

prabhjyotsingh:
LGTM! CI is also green.

corneadoug:
Awesome, Merging if there is no more discussions

PR #1283, opened 2016-08-04
Buffer append output results + fix extra incorrect results

beriaanirudh:
### What is this PR for?
There are 2 issues and their proposed fixes:
1. On a paragraph run, for every line of output, there is a broadcast of the new line from zeppelin. In case of thousands of lines of output, the browser/s would hang because of the volume of these append-output events.
2. In the above case, besides the browser-hang, another bug observed is that result data is will repeated twice (coming from append-output calls + finish-event calls).

The proposed solution for #1 is:
- Buffer the append-output event into a queue instead of sending the event immediately.
- In a separate thread, read from the queue periodically and send the append-output event.

Solution for #2 is:
- Donot append output to result if the paragraph is not runnig.

### What type of PR is it?
Improvement + Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1292

The test could be to run a simple paragraph with large result. Eg:
```
%sh
for i in {1..10000}
do
echo $i
done
```
PS: One will need to clear browser cache between running with and without this code patch since there are javascript changes as well. 

### Questions:
No
No
It could need for the design. Otherwise I have added code comments explaining behaviour.


zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -2271,7 +2271,14 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
...
   $scope.$on('appendParagraphOutput', function(event, data) {
-    if ($scope.paragraph.id === data.paragraphId) {
+    /* It has been observed that append events
+     * can be errorneously called even if paragraph
+     * execution has ended, and in that case, no append
+     * should be made. Also, it has been observed that this
+     * could be called in PENDING state as well
+     */
+    if ($scope.paragraph.id === data.paragraphId &&
+       ($scope.paragraph.status === 'RUNNING' || $scope.paragraph.status === 'PENDING')) {
```
corneadoug: Considering the Comment above, why would we accept PENDING state?

beriaanirudh: What I observed was that in between the pending and the running state, some (I observed 1) append-output events were called. If we miss those, then while the paragraph is running, we would see that some intial-result line/s is/are missing.
Also, when the para execution is finished, the complete note is sent again (with results), so we would see the correct result.

corneadoug: I see, the comment was a bit misleading, as it seems it was saying that it could be errorneously call es as well in PENDING state.

beriaanirudh: Ya, when I re-read it sounds that way.. Re-phrased it.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
```diff
@@ -0,0 +1,127 @@
...
+   * or killed due to errors.
+   */
+  public static void startRunner(RemoteInterpreterProcessListener listener) {
+    AppendOutputRunner.listener = listener;
+    if (thread == null || !thread.isAlive()) {
+      logger.info("Starting a AppendOutputRunner thread to buffer"
+          + " and send paragraph append data");
+      thread = new Thread(new AppendOutputRunner());
+      thread.start();
+    }
```
jongyoul: That's not thread-safe. You should use `synchronized` or something else

jongyoul:
I think you'd better use `ScheduledExecutorService` by running separate thread because `AppendOutputRunner` always execute `while(true)` in your current implementation.

jongyoul:
There're some `Thread.sleep(1000)` in your tests. In general, it would be nice and never be a blocker. It, however, occurs build error if CI become extremely slow. It also become a flaky test. Do you have any idea to make it more stable? And I have a question. Which part prevents updating output twice?

beriaanirudh:
It is easier to handle the twice output on the javascript side, so I have added the check in file `paragraph.controller.js`. The browser would not let results to be appended if the paragraph is completed. Also, since we are buffering events now, there would be few (1-2) events called after paragraph execution completion.
Working on your other comments..

beriaanirudh:
@jongyoul: I have incorported all feedback:
1. Using ScheduledExecutorService for ensuring AppendOutputRunner is up, instead of doing so inside the while loop.
2. Synchronizing on the thread check and creation.
3. Removed all Thread.sleep() from unit tests except for one test "testClubbedData()" (which is basically a simulation of actual run). I am out of ideas for that one... Please advice.. We could remove that test as well..

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunnerTest.java
```diff
@@ -0,0 +1,181 @@
...
+      }
+    }).when(listener).onOutputAppend(any(String.class), any(String.class), any(String.class));
+  }
+
+  private void loopForCompletingEvents(RemoteInterpreterProcessListener listener, int numTimes) {
+    numInvocations = 0;
+    prepareInvocationCounts(listener);
+    AppendOutputRunner.setListener(listener);
+    CheckAppendOutputRunner.startRunnerForUnitTests();
+    while(numInvocations != numTimes);
```
jongyoul: I suggest you add `Circuit Breaker` with few seconds to avoid running infinitely.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
```diff
@@ -0,0 +1,112 @@
...
+  private static final Long SAFE_PROCESSING_TIME = new Long(10);
+  private static final Long SAFE_PROCESSING_STRING_SIZE = new Long(100000);
+
+  private static final BlockingQueue<AppendOutputBuffer> QUEUE =
+      new LinkedBlockingQueue<AppendOutputBuffer>();
+
+  @Override
+  public void run() {
+
+    while (true) {
```
jongyoul: It looks like you are not familiar with `ScheduledExecutorService`. If you would write codes like this, only one executor will run forever. Can you check an example of `ScheduledExecutorService`? You will know this while loop is redundant.

beriaanirudh: I have a doubt. Do we want to remove this while loop and instead use ScheduledExecutorService to spawn a thread everytime, which will substitute this looping? I didn't do it because it would need over-head of a thread creation very frequently (may use lot of CPU and GC overhead). I was thinking that the job of ScheduledExecutorService (or CheckAppendOutputRunner ) is just to verify that AppendOutputRunner thread is running, and if not, then start it. That way we save overhead of spawning threads. Should I change this?

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunnerTest.java
beriaanirudh: Done.. Added a check for to fail the tests after 2 seconds.. Would have to add CircuitBreaker in dependencies to just using System times.

beriaanirudh:
My tests needed some fixes to work with other tests. I have fixed them. But the build still has some errors I think not related to my change. https://travis-ci.org/apache/zeppelin/builds/150876902 .

beriaanirudh:
@jongyoul @corneadoug I have incorporated all feedback, and added comments on queries. This PR is ready for review.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
jongyoul: ScheduledExecutorService didn't create every time. That runs `run()` method by schedule. If you write the codes with schedules like you, the pending jobs will consumes memories a lot. Please take a look.

jongyoul:
I've a few comment on your comments. Please check it.

corneadoug:
Tested a few cases to check that the front-end still handle most of the cases, and its good

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final Logger logger =
+      LoggerFactory.getLogger(AppendOutputRunner.class);
+  private static RemoteInterpreterProcessListener listener;
+
+  private static final Long BUFFER_TIME_MS = new Long(100);
+  private static final Long SAFE_PROCESSING_TIME = new Long(10);
+  private static final Long SAFE_PROCESSING_STRING_SIZE = new Long(100000);
+
+  private static final BlockingQueue<AppendOutputBuffer> QUEUE =
+      new LinkedBlockingQueue<AppendOutputBuffer>();
```
bzz: This dynamic data structure does not feel like a constant, so accouding to the [project styleguide#s5.2.4-constant-names)](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) better be rather named `queue`.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventPoller.java
```diff
@@ -72,6 +72,8 @@ public void setInterpreterGroup(InterpreterGroup interpreterGroup) {
   @Override
   public void run() {
     Client client = null;
+    AppendOutputRunner.setListener(listener);
```
bzz: Having a static setter does not look good to me - it hides the state which is hard to mock and test.

Would it be possible to have an instance of `AppendOutputRunner` here instead please?

beriaanirudh:
@corneadoug thanks for verifying.
@jongyoul Got it fixed now. I had mis-understood previously.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
beriaanirudh: Done..

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventPoller.java
beriaanirudh: Changed the class AppendOutputRunner to non-static. It has made code and tests simpler too. This is ready for review now.

beriaanirudh:
@corneadoug @jongyoul @bzz this is ready for review.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/AppendOutputRunner.java
```diff
@@ -0,0 +1,120 @@
...
+
+  public AppendOutputRunner(RemoteInterpreterProcessListener listener) {
+    this.listener = listener;
+  }
+
+  @Override
+  public void run() {
+
+    Map<String, Map<String, StringBuilder> > noteMap =
+        new HashMap<String, Map<String, StringBuilder> >();
```
bzz: Probably a nitpick, but 'diamond operator' should come handy

```java
Map<String, Map<String, StringBuilder> > noteMap = new HashMap<>();
```

beriaanirudh: Done..

beriaanirudh:
@corneadoug @jongyoul @bzz this is ready for review.

jongyoul:
@beriaanirudh You need to change more where you don't use 'diamond operator'. Can you fix them?

beriaanirudh:
@jongyoul done. Used diamond operator at all places except for 1 where compiler wouldn't allow (AppendOutputRunner.java:78)

bzz:
@beriaanirudh thank you for prompt fixes!

vipul1409:
+1 Looking forward to this

jongyoul:
LGTM

beriaanirudh:
@corneadoug @bzz this is ready for review...

Leemoonsoo:
Text output streaming used to work in following way (branch-0.6)
pr1283_before

But somehow, this PR (and master branch) works like
pr1283_after

Shell we address this problem first, before apply this patch?

corneadoug:
@Leemoonsoo 

The behaviour is still the same with:
```
(1 to 40).foreach{ i=>
    Thread.sleep(1000)
    println("Hello " + i)
}
```

Leemoonsoo:
@corneadoug yes, that streams output every one seconds if i try the master branch.

However, I did some more experiment and found

Output streaming : master branch (X), this pullrequest (X)
```
%sh
for i in {1..3}
do
    date
    sleep 2
done
```

Output streaming : master branch (O), this pullrequest (X)
```
%spark
(1 to 3).foreach{i=>
    Thread.sleep(2000)
    println(new java.util.Date)
}
```

@corneadoug @beriaanirudh Could you verify these cases, too?

corneadoug:
I tested both cases:
__Case1:__ master (X), this branch (X)
__Case2:__ master (O), this branch(X)

However the first time I tried this branch and commented, the Case 2 was fine in this Branch

Leemoonsoo:
According to the test results,

%sh interpreter output streaming might be other issue.
but it looks like %spark interpreter output streaming in this branch is not working as expected.

@beriaanirudh Could you take care of it?

beriaanirudh:
Hey,
I had done the testing on streaming cases for shell and spark earlier while raising this PR, and they did work then. I just did the testing on top of `d11221fb8af5568416ef5041fc2da8b6fa08598b` (currently latest master) and I found these results:
Case1: master (X), this branch (X)
Case2: master(O), this branch (O)

I think SparkInterpreter could have been broken which got fixed today in `5f1208bd` (ZEPPELIN-1284). Perhaps thats why Case2 was failing 2-3 days back. Could you please confirm my test-results?
cc @Leemoonsoo @corneadoug

corneadoug:
Don't know why sometimes we get some failure in that branch,
Tried again and got:
 
__Case1:__ master (X), this branch (X)
__Case2:__ master (O), this branch(O)

beriaanirudh:
Thanks for the verification @corneadoug . Even I am not able to reproduce it. My best guess is that it had to do something with the fix I mentioned above.

beriaanirudh:
@corneadoug @Leemoonsoo  please let me know if any action-item on my side.

Leemoonsoo:
Tested again and got  **Case2** works fine on this branch.
LGTM. CI failure is irrelevant to this change. 

Thanks @beriaanirudh for the contribution!

PR #1284, opened 2016-08-04
[ZEPPELIN-1291]Tagging meta value on interpreter completion.

astroshim:
### What is this PR for?

This PR is for showing the meta tag on auto completion.
### What type of PR is it?

Improvement
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1291

Please refer to screenshot.
- before 1 
  image
- before 2
  image
- after
  after
### Questions:


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/thrift/InterpreterCompletion.java
```diff
@@ -1,22 +1,5 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-/**
- * Autogenerated by Thrift Compiler (0.9.2)
+ * Autogenerated by Thrift Compiler (0.9.1)
```
AhyoungRyu: @astroshim 
I'm not sure what side-effect can be existed because of the difference between Thrift version `0.9.2` and `0.9.1`, but it would be better compile with `0.9.2` version for keeping consistency i think. What do you think?

astroshim: @AhyoungRyu you're right! I'll use 0.9.2. Thank you.

zeppelin-interpreter/src/main/thrift/RemoteInterpreterService.thrift
```diff
@@ -70,7 +70,8 @@ struct RemoteApplicationResult {
  */
```
AhyoungRyu: Actually I wrote this description to address [this feedback](https://github.com/apache/zeppelin/pull/984#discussion_r66837016) by @bzz. So why don't you add some description about what `meta` variable do for the future contributors? :)

astroshim: okay. I'll do that.Thanks.

astroshim:
@AhyoungRyu I fixed. Please review. :)

AhyoungRyu:
@astroshim Thank you for the prompt fix!

felixcheung:
great improvement - would it be better if the tag matches the interpreter name - it could be `%spark1` `%spark2` for example?

astroshim:
@felixcheung Thank you for your review. 
I think The meta name of auto completion should be original meta name because the user interpreters name that you mentioned(spark1,spark2) is not related to meta name since the meta name comes from remote interpreter.
And If the meta name of spark is defined ```spark```, auto completion meta name should be ```spark``` even if interpreter group name is ```spark1 or spark2```.
What do you think?

felixcheung:
hmm, shouldn't we match the group name? that's what the user type though (eg. `%spark1`) and auto complete is to help with the user when she's typing?

astroshim:
Even though the interpreter name is ```spark1``` or ```myintp```, the auto completion is spark's.
so This way(just spark) is clear i think.

astroshim:
ping

astroshim:
re-trigger CI

felixcheung:
As before, I think it will be more intuitive if in a paragraph, the user types:
```
%hadoop1
spark.
```
And to autocomplete and the popups says `hadoop1`?
Yes, I agree the completion is for "Spark stuff" but completion result might include items that are interpreter instance/state specific things, like local variables and so on?

That said, I don't feel strongly about this so I'll let others to comment.

PR #1286, opened 2016-08-05
[ZEPPELIN-1294] Implement one-way sync for notebook repos

jasonxh:
### What is this PR for?
Currently, if there are two notebook storages configured, the sync process treats them equally and can pull changes from either storage. This can cause confusions if there is a clear distinction between primary and secondary storages, and the primary storage is managed outside of Zeppelin itself, such as a local folder that's git controlled. If a notebook is deleted or reverted to an older version in the primary storage behind Zeppelin's back, the sync process will overwrite the change with data from the secondary storage, likely against the user's intention.

The proposal is to enable the scenario with a one-way sync flag, where the primary storage is treated as the only source of truth and the secondary storage is merely a mirror of the primary.

### What type of PR is it?
Improvement

### What is the Jira issue?
<https://issues.apache.org/jira/browse/ZEPPELIN-1294>

1. Setup two notebook storages through `ZEPPELIN_NOTEBOOK_STORAGE`, with the first is local file storage, i.e. `org.apache.zeppelin.notebook.repo.VFSNotebookRepo`. The second storage can be anything, e.g. `org.apache.zeppelin.notebook.repo.zeppelinhub.ZeppelinHubRepo`.
2. Add `export ZEPPELIN_NOTEBOOK_ONE_WAY_SYNC=true` to `zeppelin-env.sh`.
3. Start Zeppelin server.
4. Delete a notebook directly from the notebook folder.
5. Refresh notebooks from Zeppelin UI.
6. Observe from the secondary storage that the notebook has been deleted.

### Questions:
* Does this needs documentation? Documentation for the new flag has been added to `install.md`

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -271,7 +271,8 @@ public class NotebookRepoSync implements NotebookRepo {
 
         if (sdate.compareTo(ddate) != 0) {
           if (sdate.after(ddate) || oneWaySync) {
-            /* source contains more up to date note - push */
+            /* if source contains more up to date note - push
+             * if oneWaySync is enabled, always push no matter who's newer */
             pushIDs.add(snote.getId());
             LOG.info("Modified note is added to push list : " + sdate);
           } else {
@@ -293,9 +294,11 @@ public class NotebookRepoSync implements NotebookRepo {
       if (dnote == null) {
         /* note exists in destination storage, and absent in source */
         if (oneWaySync) {
+          /* if oneWaySync is enabled, delete the note from destination */
           LOG.info("Extraneous note is added to delete dest list : " + note.getId());
           delDstIDs.add(note.getId());
         } else {
+          /* if oneWaySync is disabled, pull the note from destination */
           LOG.info("Missing note is added to pull list : " + note.getId());
           pullIDs.add(note.getId());
         }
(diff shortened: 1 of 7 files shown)


jongyoul:
@khalidhuseynov Could you please take a look into it? Code implementation is simple but we need to confirm the concept actually.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -246,14 +268,17 @@ NotebookRepo getRepo(int repoIndex) throws IOException {
...
-          pushIDs.add(snote.getId());
-          LOG.info("Modified note is added to push list : " + sdate);
-        } else if (sdate.compareTo(ddate) != 0) {
-          /* destination contains more up to date note - pull */
-          LOG.info("Modified note is added to pull list : " + ddate);
-          pullIDs.add(snote.getId());
+
+        if (sdate.compareTo(ddate) != 0) {
+          if (sdate.after(ddate) || oneWaySync) {
+            /* source contains more up to date note - push */
```
khalidhuseynov: I think it's better to add comment here that basically everything will be pushed in case of `oneWaySync`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -266,14 +291,21 @@ NotebookRepo getRepo(int repoIndex) throws IOException {
     for (NoteInfo note : destNotes) {
       dnote = containsID(sourceNotes, note.getId());
       if (dnote == null) {
-        /* note exists in destination storage, and absent in source - pull*/
-        pullIDs.add(note.getId());
+        /* note exists in destination storage, and absent in source */
+        if (oneWaySync) {
+          LOG.info("Extraneous note is added to delete dest list : " + note.getId());
```
khalidhuseynov: may add comment here on deleting from secondary storage

khalidhuseynov:
@jasonxh as you mentioned it makes sense in cases when you control your primary storage by yourself and want your secondary storage be exactly the reflection of it. may need to add some descriptive comments as mentioned above. LGTM

jasonxh:
Thanks @khalidhuseynov @jongyoul. Comments added.

khalidhuseynov:
LGTM

jongyoul:
Thanks @khalidhuseynov for the review. Merging if there's no more discussion.

PR #1293, opened 2016-08-06
[ZEPPELIN- 1298] Log instead of throwing trace for ping messages

khalidhuseynov:
### What is this PR for?
When non authenticated user in non-anonymous mode tries to send any websocket api message it results in throwing a traceback. However on PING message we don't need to show the whole traceback and we can just log in order not to pollute the logs file.

### What type of PR is it?
Hot Fix

### Todos
* [x] - log and return on PING

### What is the Jira issue?
[Zeppelin-1298](https://issues.apache.org/jira/browse/ZEPPELIN-1298)

Follow the steps in issue and shouldn't get repeating traceback.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -20,6 +20,8 @@ import com.google.common.base.Strings;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
+
+import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.conf.ZeppelinConfiguration.ConfVars;
 import org.apache.zeppelin.display.AngularObject;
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @khalidhuseynov for the fix. I tried this patch and it does not print stacktrace on PING message anymore when user not logged in.

However, it still prints stacktrace when user browse zeppelin gui. And i think it's also the case we don't need to print stacktrace.
Could you handle this case, too?

And IMO, i think it's okay to adjust log message level to DEBUG for both case because user browse Zeppelin gui without valid ticket is fairly normal behavior, when they're not logged in.

khalidhuseynov:
@Leemoonsoo thanks for review. i addressed your comments by changing all the invalid cases to log in debug mode instead of throwing exception. let me know if there's something else left.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -131,8 +131,12 @@ public void onMessage(NotebookSocket conn, String msg) {
       }
       
       String ticket = TicketContainer.instance.getTicket(messagereceived.principal);
-      if (ticket != null && !ticket.equals(messagereceived.ticket))
-        throw new Exception("Invalid ticket " + messagereceived.ticket + " != " + ticket);
+      if (ticket != null && !ticket.equals(messagereceived.ticket)){
+        /* not to pollute logs, log in debug instead of exception */
+        LOG.debug("{} message: invalid ticket {} != {}", messagereceived.op,
```
felixcheung: LOG.warning instead?

Leemoonsoo: When messagereceived.ticket is null or empty string, which is the case user not logged in and browse Zeppelin GUI, i think we can minimize the log in this case (log.debug?).

When messagereceived.ticket is not empty, but not equals to ticket, that means something wrong, and in this it's better print some logs.

bzz:
`LOG.debug` sounds like a great idea to me to avoid filling up the logs in default configuration.

LGTM

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
khalidhuseynov: @felixcheung @Leemoonsoo addressed in 3fa5403058b2d258180ecf23e8e7e1c82169dc84

minahlee:
@khalidhuseynov Could you resolve conflicts?

khalidhuseynov:
@minahlee done

Leemoonsoo:
LGTM

Leemoonsoo:
Merge into master and branch-0.6 if there're no further discussions.

PR #1301, opened 2016-08-08
[MINOR] Removed unused profiles from spark/pom.xml

jongyoul:
### What is this PR for?
Making spark/pom.xml simple

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Simplified pom.xml in spark

### What is the Jira issue?
N/A

No test. CI should be green

### Questions:

### DIFF
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -38,8 +38,6 @@
     <mockito.version>1.10.19</mockito.version>
     <powermock.version>1.6.4</powermock.version>
     <spark.version>2.0.0</spark.version>
-    <hadoop.version>2.3.0</hadoop.version>
-    <yarn.version>${hadoop.version}</yarn.version>
   </properties>
 
   <dependencies>
@@ -83,13 +81,6 @@
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-yarn_${scala.binary.version}</artifactId>
       <version>${spark.version}</version>
-      <scope>provided</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-yarn-api</artifactId>
-      <version>${yarn.version}</version>
       <scope>provided</scope>
     </dependency>
 
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -38,6 +38,8 @@
     <mockito.version>1.10.19</mockito.version>
     <powermock.version>1.6.4</powermock.version>
     <spark.version>2.0.0</spark.version>
+    <hadoop.version>2.3.0</hadoop.version>
+    <yarn.version>${hadoop.version}</yarn.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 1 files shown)


jongyoul:
https://travis-ci.org/apache/zeppelin/builds/150548730

jongyoul:
https://travis-ci.org/apache/zeppelin/builds/150645554

bzz:
Looks like candidate for `flaky-test` labeled JIRA issue in zeppelin-server:
```
Failed tests: 
  ZeppelinSparkClusterTest.zRunTest:204 expected:<FINISHED> but was:<ERROR>

Tests run: 65, Failures: 1, Errors: 0, Skipped: 0
```

jongyoul:
@bzz Basically, Spark module doesn't have to need hadoop version because we only use Spark repl, core and so on not including specific hadoop version. spark-dependencies only needs specific hadoop version. Thus, specifying hadoop version is not necessary for it.

bzz:
Thank you for explanation! Makes sense.

Looks great to me.

May be it's worth to `git rebase master` as CI seems to be recently fixed.

zjffdu:
Definitely should remove profile yarn. I got classpath issue when enabling yarn profile. 

And we need to update the docs accordingly.

bzz:
@jongyoul do you want to take of docs update that @zjffdu mention in this PR as well?

jongyoul:
@zjffdu @bzz I've already removed yarn profile. Have I missed something?

jongyoul:
And about docs, it never change some behaviour. I think another issue linked above is a better point to update docs.

spark/pom.xml
```diff
@@ -658,292 +547,6 @@
...
-          <groupId>org.apache.hadoop</groupId>
-          <artifactId>hadoop-yarn-api</artifactId>
-          <version>${yarn.version}</version>
-        </dependency>
-      </dependencies>
-    </profile>
-    
     <!-- include sparkr in the build -->
     <profile>
       <id>sparkr</id>
```
zjffdu: Is sparkr needed here ? Because I notice there's another sparkr profile in spark-dependencies/pom.xml

jongyoul: This is because of selecting interpreter-setting.json if/if not we use sparkR and this is because we have two different SparkR versions.

felixcheung:
I thought we do need `cassandra-spark-*` @doanduyhai ?

jongyoul:
@felixcheung What I want to remove is duplicated from spark-dependencies/pom.xml. This deletion never hurt current features. It just removes duplicated codes.

bzz:
Quick question: just to double-check - this should not affect users following published build instructions like https://www.mapr.com/blog/building-apache-zeppelin-mapr-using-spark-under-yarn , right?

In a way like in [ZEPPELIN-1353](https://issues.apache.org/jira/browse/ZEPPELIN-1353)

jongyoul:
@bzz Yes, what you mentioned is related to #1339 which will make a change to build custom Spark.

PR #1302, opened 2016-08-08
[ZEPPELIN-1306] Interpreter restarts on a note.

astroshim:
### What is this PR for?

This PR is for usability of restarting interpreter. 
### What type of PR is it?

Improvement
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1306

You can restart interpreter on the interpreter binding page on a note.
and can only restart `scoped` and `isolated` interpreter.
Please refer to screen shot.

restartintp
### Questions:


astroshim:
ping

astroshim:
ping

AhyoungRyu:
@astroshim I've also thought the interpreter `restart` button in separated page quite uncomfortable. This PR can be a good start for the better UX i think :)
I tested your patch and it almost works well except one thing. 

>can only restart `scoped` and `isolated` interpreter.

Since you said like above in the description,  I edited `Spark` interpreter as `shared` mode in the interpreter menu and restarted with your new interpreter binding UI. I expected it failed with some error messages, but it restarted. Did I miss sth?

AhyoungRyu:
@astroshim Oh that was my bad. Please ignore my above comment. I only build `zeppelin-web` not backend side.
Tested again and it works well as expected. Let's wait the other reviewers :)

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -856,6 +856,23 @@ public void setPropertyAndRestart(String id, InterpreterOption option, Propertie
     }
   }
 
+  private boolean interpreterIsShared(InterpreterSetting setting) {
+    if (setting.getOption().isPerNoteProcess() == false &&
+      setting.getOption().isPerNoteSession() == false){
+      return true;
+    }
+    return false;
```
bzz: Two things noted here:

1. Method body uses only `setting.getOption()` but not the `setting`, so it's better to just pass `setting.getOption()` as an argument to the function. This a general prinicple that is called ["Law of Demeter"](https://en.wikipedia.org/wiki/Law_of_Demeter) and it generally [helps to make code more testable](http://googletesting.blogspot.kr/2008/08/by-miko-hevery-so-you-decided-to.html).

2. Usually, returning just `true` or `false` after evaluating an exception means that the code can be simplified (given that 1 is implemented)

  ```java
return interpreterSetting.isPerNoteProcess() == false && interpreterSetting.isPerNoteSession() == false
  ```

  Which can be event further simplified to more readable and understandable:

 ```java
return interpreterSetting.isPerNoteProcess() || interpreterSetting.isPerNoteSession()
  ```

What do you think?

astroshim:
Thank you very much about reviewing this PR. :smile:

Leemoonsoo:

How about make UI little bit more consistent?

On interpreter setting page, we have interpreter restart button look like
image

And interpreter binding page now have different style of restart button
image

How about use the similar style and similar icon for interpreter restart button? so user can easily guess what this button is for. For example,

image

Leemoonsoo:
And i think it's more make sense and useful if we can make restart button like,

restart button on interpreter setting page : restart all interpreter instances/processes related to the setting.

restart button on interpreter binding page : restart a interpreter instance/process for the note when "per Note" is enabled. Otherwise restart all interpreter instances/processes.

What do you think?

astroshim:
@Leemoonsoo Your suggestion about UI is definitely better!
and also makes sense about restarting action on interpreter binding page.
I will fix them.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
astroshim: Thank you for the review.

1. I totally agree with your opinion! but lots of zeppelin codes have same problems. 
Do you think it should be refactored the other codes too?

2. The interpreterIsShared() method returns `true` only when two variables are same `false`, Is it same with `return interpreterSetting.isPerNoteProcess() || interpreterSetting.isPerNoteSession()` ?

astroshim:
\cc @Leemoonsoo I fixed UI like following screenshot and process. please review.
o

AhyoungRyu:
@astroshim Definitely looks better! I like @Leemoonsoo 's idea 👍 
It's just my personal opinion, how about putting `refresh` icon to the left side of the interpreter name? I mean

 - Before
screen shot 2016-08-28 at 12 30 26 pm

 - After 
screen shot 2016-08-28 at 12 29 11 pm

And while we are familiar with the feature; "Restart Interpreter", some ppl may not. So what about showing some message with tooltip like this?
screen shot 2016-08-28 at 12 37 23 pm

What do you think?

AhyoungRyu:
@astroshim Great! Personally I would prefer "just icon" like the reload notes icon in Zeppelin home. 
But it's just my personal opinion. Final decision is up to you :)

astroshim:
I just changed to 'icon'.
Thank you.

astroshim:
ping

AhyoungRyu:
@astroshim Looks good!
I think it would be better the activated and deactivated interpreter buttons can be in line like below images.

 - Before (with this patch)

screen shot 2016-08-31 at 2 15 43 pm

 -  After 

screen shot 2016-08-31 at 2 15 43 pm

Or I'd like to suggest another option. How about simply changing `Restart` button color to grey for the deactivated interpreter button?(usually the grey color stands for `deactivated` I guess)
screen shot 2016-08-31 at 2 29 40 pm


What do you think? :D

astroshim:
@AhyoungRyu I think `deactivated` button is better so I updated it. Thanks.

AhyoungRyu:
@astroshim Tested again and it works well. I like it :) 
Here is the new UI (for the other reviewers)
new

@Leemoonsoo If you have some time, could you please check this PR again?

corneadoug:
My only feedback is that right now, it's hard to know if the interpreter has finished restarting or not.
My guess is that it takes some time, so you shouldn't be able to restart it again and you should be warned somehow when it's done

astroshim:
@corneadoug Actually it's very short time to restart interpreter job because it's just kill interpreter process or thread.
How about alerting message when request is success?

bzz:
Looks great to me, thank you @astroshim !

On @corneadoug point - it should be possible to animate the icon button \w arrows to be i.e rotating while "restart is in progress" so it stops spinning, as soon as restart is done. This might give user a nice visual feedback.

But @astroshim is righ, restart is pretty fast due to the nature of implementation - interpreter process just gets killed and then it's is lazy-started only on next time user runs the paragraph where it is binded.

corneadoug:
Maybe a spinner just like the save button after editing an interpreter would be enough then.
Its just a matter of having a small feedback.
screen shot 2016-09-05 at 3 21 01 pm

astroshim:
@corneadoug I fixed it. please review. Thank you!

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -627,6 +627,44 @@ angular.module('zeppelinWebApp').controller('NotebookCtrl', function($scope, $ro
...
+          }).error(function(data, status, headers, config) {
+            thisConfirm.close();
+            console.log('Error %o %o', status, data.message);
+            BootstrapDialog.show({
+              title: 'Error restart interpreter.',
+              message: data.message
+            });
+          });
+          return false;
+        } else {
```
corneadoug: You can remove that else statement, `form` do not exist i this case and we don't need to do anything

astroshim: I fixed. Thank you.

corneadoug:
Tested LGTM
@astroshim can you rebase? It should make the CI green.

astroshim:
@corneadoug I just re-based thanks.

corneadoug:

Leemoonsoo:
@astroshim @corneadoug, Shell we have some unittests for `public void restart(String settingId, String noteId)` before we merge? Restarting only a interpreter instance/process for a note is new ability. So i think it make sense to have a unittest.

Also, shell we have https://github.com/apache/zeppelin/blob/master/docs/rest-api/rest-interpreter.md#restart-an-interpreter updated, too?

astroshim:
@Leemoonsoo I agree with your opinion. Let me fix them.

corneadoug:
@astroshim Can you rebase so that we can check the CI?
I will merge after it's green

zeppelin-server/src/test/java/org/apache/zeppelin/rest/InterpreterRestApiTest.java
```diff
@@ -191,6 +192,60 @@ public void testInterpreterRestart() throws IOException, InterruptedException {
...
+    }
+
+    String jsonRequest = "{\"noteId\":\"" + note.getId() + "\"}";
+    LOG.info("=========>{}", jsonRequest);
+
+    // Restart isolated mode of Interpreter for note.
+    mdIntpSetting.getOption().setPerNoteProcess(true);
+    mdIntpSetting.getOption().setPerNoteSession(false);
+    PutMethod put = httpPut("/interpreter/setting/restart/" + mdIntpSetting.getId(), jsonRequest);
+    assertThat("isolated interpreter restart:", put, isAllowed());
```
Leemoonsoo: Thanks for adding test.

Could you add some other tests that not only check success of REST API call, but also check actual restart of interpreter that binded to specific note?

Let me know if you need help.

corneadoug:
@astroshim Do you some time to rebase, and add more tests?

astroshim:
@corneadoug I'm going to add test case soon. :laughing:

zeppelin-server/src/test/java/org/apache/zeppelin/rest/InterpreterRestApiTest.java
astroshim: Could you give me some information about checking actual restart of interpreter that binded to specific note?

astroshim:
@corneadoug, @Leemoonsoo  I re-based and added testcase. please review.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/InterpreterRestApiTest.java
Leemoonsoo: @astroshim 

You can get some idea from 
https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java#L718,  how interpreter restart can be checked in the test case.

I've seen you have added "restart interpreter with isolated(per note process) mode enabled" there.
But still, it does not test `public void restart(String settingId, String noteId)`. Could you add test cases for it?

astroshim: @Leemoonsoo I added testcase already. please review the NotebookTest.java.

Leemoonsoo: I've seen the test case. but doesn't look like test case for new feature introduced in this PR `public void restart(String settingId, String noteId)`.

astroshim:
re-trigger CI

jongyoul:
Can you add consideration for the multi-tenancy? If Shiro is enabled, current implementation will disable all of users' interpreters

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
```diff
@@ -150,10 +151,15 @@ public Response removeSetting(@PathParam("settingId") String settingId) throws I
...
-  public Response restartSetting(@PathParam("settingId") String settingId) {
-    logger.info("Restart interpreterSetting {}", settingId);
+  public Response restartSetting(String message, @PathParam("settingId") String settingId) {
+    logger.info("Restart interpreterSetting {}, msg={}", settingId, message);
+
     try {
-      interpreterFactory.restart(settingId);
+      RestartInterpreterRequest request = gson.fromJson(message, RestartInterpreterRequest.class);
+
+      String noteId = request == null ? null : request.getNoteId();
```
astroshim: If `noteId` is null, execute `restart(String id)` in the `interpreterFactory`.

jongyoul: But, I don't think it's not a good implementation because it doesn't look intended. Can you tell me the case that don't have `noteId`?

astroshim:
@jongyoul I totally agree with you adding multi-tenancy but how about separating that to another PR because we also already have restarting interpreter function in the `Interpreter Setting page` .

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -947,13 +947,29 @@ public void setPropertyAndRestart(String id, InterpreterOption option, Propertie
...
 
+  private boolean noteIdIsExist(String noteId) {
+    return noteId == null ? false : true;
+  }
+
+  public void restart(String settingId, String noteId) {
+    InterpreterSetting intpsetting = interpreterSettings.get(settingId);
+    Preconditions.checkNotNull(intpsetting);
+    
+    if (noteIdIsExist(noteId) &&
```
jongyoul: You'd better adopt `noteId`, too. :-)

astroshim: Thank you pointing out properly!

toughrogrammer:
How it's going on? I want to restart *note specific interpreter* when cron is executed or refreshing spark context, but current version(0.7.2), it's impossible. It's really unproductive experience.

Leemoonsoo:
@toughrogrammer Could you explain little more about the problem? How cron execution is related with restart interpreter?

toughrogrammer:
screen shot 2017-10-14 at 2 39 53 am

I'm using spark interpreter per note isolated mode and I wrote a notebook for batch job with cron execution. (Please refer to the attached image.)
When this notebook is executed by cron, interpreter is restarted. But problem is other notebook's spark context also to be expired. Every hour I need to re-run all paragraph and read large data set from storage.

Leemoonsoo:
Ah i see. That definitely looks like a problem. I filed an issue https://issues.apache.org/jira/browse/ZEPPELIN-2995.

toughrogrammer:
But isn't this original issue also important? I usually want to restart specific notebook. Because spark executors aren't returned automatically, I must kill zeppelin application from YARN web UI. And then, zeppelin notebook lost spark context and I need to restart spark interpreter.

Leemoonsoo:
Yeah, i think we can make "auto-restart interpreter on cron execution" checkbox restart specific interpreter instance to the notebook.

kjmrknsn:
The bug that "auto-restart interpreter on cron execution" restarts all interpreters will be fixed by https://github.com/apache/zeppelin/pull/2681.

PR #1305, opened 2016-08-09
[ZEPPELIN-1310] [WIP] Deactivate angular debugging classes

corneadoug:
### What is this PR for?
This PR removes the angular debug classes added to HTML elements (that plugins like batarang are using), it follows advices from the angular [documentation](https://code.angularjs.org/1.5.5/docs/guide/production) and should improve performances.

### What type of PR is it?
Improvement

### Todos
* [x] - Add documentation on how to easily activate it for people wanted to debug
* [ ] - Have the `angular.element().scope().controllerCall()` refactored
* [ ] - Fix Selenium tests

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1310

Do a `./grunt build` or launch Zeppelin normally
Check in the console there should not be any `ng-scope` or such in the html classes

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/integration/ZeppelinIT.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/integration/ZeppelinIT.java
@@ -199,7 +199,7 @@ public class ZeppelinIT extends AbstractZeppelinIT {
     }
     try {
       // navigate to interpreter page
-      WebElement settingButton = driver.findElement(By.xpath("//button[@class='nav-btn dropdown-toggle ng-scope']"));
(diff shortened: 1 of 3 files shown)


prabhjyotsingh:
👍

corneadoug:
This PR is going to have to wait,
When activating this option, things like:
`angular.element().scope().functionIWantToCall()` are not possible anymore.
We are relying on it to call 5 different functions in our code, and bypassing it would mean some refactoring.

corneadoug:
Let's close this PR, this would also break most of the custom angular display notebooks.
Too much changes need to be done before this can be applied.

PR #1311, opened 2016-08-10
[ZEPPELIN-1316] Zeppelin can not start due to an incorrect Interpreter Setting	

cloverhearts:
### What is this PR for?
If there are problems with the information defined in the Interpreter Setting (conf / interpreter.json), Interpreter Group during the defined file does not exist, can not run Zeppelin.
Thus, it was an exception handling for that part.

In order to address this issue in the current version it was restarted after deleting Interpreter.json.

### What type of PR is it?
Bug Fix

### Todos
- [x] - check to Interpreter Setting is null

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1316

1. Modify the following part of the Interpreter.json file.
```json
{
      "interpreterGroup": [
        {    
          "name": "spark",
          "class": "fake and invalid class path more more...", // invalid value
          "defaultInterpreter": false
        }
}
```

2. 
After restarting the Zeppelin, Zeppelin Confirm that normally works.
Version of the PR is to be normally started up.
It should also be display all possible Interpreter Web UI.

#### before
error

#### after
correct

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -370,7 +370,7 @@ public class InterpreterFactory implements InterpreterGroupFactory {
       // Update transient information from InterpreterSettingRef
       interpreterSettingObject = interpreterSettingsRef.get(setting.getGroup());
       if (interpreterSettingObject == null) {
-        logger.info("can't get InterpreterSetting " +
+        logger.warn("can't get InterpreterSetting " +
           "Information From loaded Interpreter Setting Ref - {} ", setting.getGroup());
         continue;
       }
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -55,6 +55,7 @@ import com.google.gson.reflect.TypeToken;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.NullArgumentException;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.sonatype.aether.RepositoryException;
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -365,9 +368,14 @@ private void loadFromFile() throws IOException {
       setting.getOption().setRemote(true);
 
       // Update transient information from InterpreterSettingRef
-      // TODO(jl): Check if reference of setting is null
-
-      setting.setPath(interpreterSettingsRef.get(setting.getGroup()).getPath());
+      interpreterSettingObject = interpreterSettingsRef.get(setting.getGroup());
+      if (interpreterSettingObject == null) {
+        logger.info("can't get InterpreterSetting " +
+          "Information From loaded Interrpeter Setting Ref - {} ", setting.getGroup());
```
AhyoungRyu: There is typo :)
`Interrpeter` -> `Interpreter`

AhyoungRyu:
@cloverhearts Thanks for the patch! I faced this wired 503 error several times.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
cloverhearts: oh.. Thank you!
my mistake. :)

bzz:
Thank you @cloverhearts ! That happened to me as well.

Looks great to me.

\cc @jongyoul for review

jongyoul:
This will ignore non-exist setting if it doesn't exist anymore in `interpreter/*`. I think it looks realistic in production level because admin usually delete that interpreter setting if it doesn't exists. LGTM

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -365,9 +368,14 @@ private void loadFromFile() throws IOException {
       setting.getOption().setRemote(true);
 
       // Update transient information from InterpreterSettingRef
-      // TODO(jl): Check if reference of setting is null
-
-      setting.setPath(interpreterSettingsRef.get(setting.getGroup()).getPath());
+      interpreterSettingObject = interpreterSettingsRef.get(setting.getGroup());
+      if (interpreterSettingObject == null) {
+        logger.info("can't get InterpreterSetting " +
```
felixcheung: logger.warn?

cloverhearts: warning log?
Okay, i will modify it.

bzz:
Thank you for prompt review!

Merging _to master_ if there is no further discussion

bzz:
CI is green now, so merging it in

PR #1315, opened 2016-08-10
[ZEPPELIN-530] Added changes for Credential Provider, using hadoop commons Credential apis

rconline:
### What is this PR for?
This is the first step in order to ensure clear text passwords are not stored in the configuration files. To start with this PR will take care of getting AD system password from the .jceks file, configured by the user specified in the shiro.ini file. Going forward the same keystore can be used to read passwords for other systems as well. 

If the hadoopSecurityCredentialPath path is present and not empty in the shiro.ini, then the password is read from the keystore file and it need not be stored inside the shiro.ini file. 

### What type of PR is it?
[ Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-530

Create a keystore file using the hadoop credential commandline, for this the hadoop commons should be in the classpath

`hadoop credential create activeDirectoryRealm.systempassword -provider jceks://file/user/zeppelin/conf/zeppelin.jceks`

Change the following values in the Shiro.ini file, and uncomment the line:

`activeDirectoryRealm.hadoopSecurityCredentialPath = jceks://file/user/zeppelin/conf/zeppelin.jceks`

### Questions:
No
No. This is an additional option. 
Yes

### Tasks
* Documentation

### DIFF
--- a/conf/shiro.ini
+++ b/conf/shiro.ini
@@ -28,7 +28,10 @@ user3 = password4, role2
 ### A sample for configuring Active Directory Realm
 #activeDirectoryRealm = org.apache.zeppelin.server.ActiveDirectoryGroupRealm
 #activeDirectoryRealm.systemUsername = userNameA
+
+#use either systemPassword or hadoopSecurityCredentialPath, more details in http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/shiroauthentication.html
 #activeDirectoryRealm.systemPassword = passwordA
+#activeDirectoryRealm.hadoopSecurityCredentialPath = jceks://file/user/zeppelin/zeppelin.jceks
 #activeDirectoryRealm.searchBase = CN=Users,DC=SOME_GROUP,DC=COMPANY,DC=COM
 #activeDirectoryRealm.url = ldap://ldap.test.com:389
 #activeDirectoryRealm.groupRolesMap = "CN=admin,OU=groups,DC=SOME_GROUP,DC=COMPANY,DC=COM":"admin","CN=finance,OU=groups,DC=SOME_GROUP,DC=COMPANY,DC=COM":"finance","CN=hr,OU=groups,DC=SOME_GROUP,DC=COMPANY,DC=COM":"hr"
--- a/conf/shiro.ini
+++ b/conf/shiro.ini
@@ -29,7 +29,7 @@ user3 = password4, role2
 #activeDirectoryRealm = org.apache.zeppelin.server.ActiveDirectoryGroupRealm
 #activeDirectoryRealm.systemUsername = userNameA
 
-#use either systemPassword or hadoopSecurityCredentialPath, more details in http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/shiroauthentication.html
+#use either systemPassword or hadoopSecurityCredentialPath, more details in http://zeppelin.apache.org/docs/latest/security/shiroauthentication.html
 #activeDirectoryRealm.systemPassword = passwordA
 #activeDirectoryRealm.hadoopSecurityCredentialPath = jceks://file/user/zeppelin/zeppelin.jceks
 #activeDirectoryRealm.searchBase = CN=Users,DC=SOME_GROUP,DC=COMPANY,DC=COM
--- a/docs/security/shiroauthentication.md
+++ b/docs/security/shiroauthentication.md
@@ -1,3 +1,4 @@
+
 ---
 layout: page
 title: "Apache Shiro Authentication for Apache Zeppelin"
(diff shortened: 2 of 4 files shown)


prabhjyotsingh:
Thank you @rconline for taking care of this. LGTM!

docs/security/shiroauthentication.md
```diff
@@ -1,3 +1,4 @@
+
```
AhyoungRyu: Did you add this new line on purpose? If not, you need to remove this empty line. Since Zeppelin docs website is using Jekyll as a site builder, it needs to recognize `___`(triple-dash lines).

prabhjyotsingh:

jongyoul:
@rconline This PR makes Zeppelin server module include Hadoop-common 2.6. How can I change the version of Hadoop? Or don't we need to change it even we use Hadoop 1.x?

rconline:
@jongyoul It does not matter, because we are just depending upon hadoop 2.6 credential api's. So if the user is on Hadoop 1.x, then he just needs to add Hadoop common - 2.6 jars to the classpath to create the jceks file. At the time of authentication, this jar is already present inside the zeppelin artifact, so we are good there.

jongyoul:
@rconline Do you mean it's fine if some users don't use Hadoop2 and don't want to use jceks features? Isn't there any side effect at all?

rconline:
@jongyoul  in that case, they will not be able to take advantage of storing AD passwords in an encrypted fashion. That said, there will be no functionality loss.

jongyoul:
@rconline Thanks for explaining it. I've researched the `jceks`. AFAIK, it doesn't need hadoop-common dependencies. It's included by java.security by default. Can you remove hadoop-common and support `jceks`? I've googled it and found some examples. Basically, `jceks` would be a good candidate for making password secured.

rconline:
@jongyoul this work is currently solving for AD, but is not limited to it. Going forward users may choose to store their Hive, Hbase and other data system passwords. These passwords as conventions get stored on HDFS, where we will need a simplified, credential API, and therefore the choice is to use the credential API. Also this is pretty standard implementation in the Big Data world.

jongyoul:
@rconline It's not clear for me. Do you tell me that users store their Hive, Hbase passwords in HDFS? Or does this support reading `jceks` from HDFS? In your description, this PR is for encrypting the plain text password. I don't know what the relationship is from HDFS. I think you should describe why you must use hadoop-common in details.

rconline:
@jongyoul let me try to take a step back and try to explain.

Zeppelin is going to be used for various use cases, some of which will involve HDFS - Hive/Spark/Phoenix/Hbase etc, some use cases will also need support for non-HDFS such as Postgres/Mysql etc. 

**Problem** - all of these end systems may require users to store passwords. Currently in zeppelin there are two locations for storing these passwords - 1. shiro.ini for AD passwords and 2. interpreter.json for the rest of the data systems. These passwords are stored in clear text as of now. 

**Solution** - Encrypt the password and store in a file that can be read only at runtime, by zeppelin process to connect successfully. Question is where? Either on the zeppelin host system or hdfs where big-data users are akin to storing passwords. JCEKS is a java supported concept and has worked well for most users, and therefore can be used. Creating a .jceks file is possible on a host, 
`jceks://file/tmp/test.jceks`, whereas on HDFS the user may have to connect to hdfs and then create the file such as `jceks://hdfs@nn1.example.com/my/path/test.jceks`, when the password is being stored on hdfs. 

At this point in time we have solved the problem for not storing passwords in Shiro.ini, which can be stored at the zeppelin host itself. However we have to improve this solution and make it work for the rest of the use cases and that's where Credential API comes into play. 

Credential API is a generic solution which allows users to create password files for both of the cases - on the host and hdfs. Needless to point out that this has been used across Knox, which is a good standard for security. 

Please let me know if this makes sense, or if you have any more questions.

conf/shiro.ini
```diff
@@ -28,7 +28,10 @@ user3 = password4, role2
 ### A sample for configuring Active Directory Realm
 #activeDirectoryRealm = org.apache.zeppelin.server.ActiveDirectoryGroupRealm
 #activeDirectoryRealm.systemUsername = userNameA
+
+#use either systemPassword or hadoopSecurityCredentialPath, more details in http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/shiroauthentication.html
```
AhyoungRyu: It would be better we use `latest` instead of `0.7.0-SNAPSHOT`.

rconline: @AhyoungRyu taken care. Thanks!

jongyoul:
@rconline I know what you focus on and try to solve the problem. I, actually, cannot accept whether your solution is perfect or not, but I agree the problem you told. Thanks for the explaining it. your codes looks good to me.

prabhjyotsingh:
CI fails for  #6786.9 - Which looks unrelated.

PR #1316, opened 2016-08-10
[ZEPPELIN-1118] Migrate Zeppelin Live Demo feature from zeppelin-project.org to zeppelin.apache.org

Leemoonsoo:
### What is this PR for?
Migrate Zeppelin Live Demo feature from zeppelin-project.org to zeppelin.apache.org
It adds

1) "Demo" button on main page to redirect demos.html
2) "demos.html" to list all the available 3rd party demos
3) Apache Zeppelin and Spark demo by Aws-emr.

Amazon EMR provided the demo instances and NFLabs helped configuring this demo.

### What type of PR is it?
Feature

### Todos
* [x] - migration of contents

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1118

image

image

image

### Questions:

### DIFF
--- a/_includes/themes/zeppelin/_navigation.html
+++ b/_includes/themes/zeppelin/_navigation.html
@@ -64,7 +64,8 @@
     </p>
     <p><a href="http://youtu.be/_PQbVH_aO5E" target="_zeppelinVideo" class="btn btn-primary btn-lg bigFingerButton" role="button">Watch the video</a>
 
-       <a href="./download.html" class="btn btn-primary btn-lg bigFingerButton" role="button">Get Zeppelin</a></p>
+       <a href="./download.html" class="btn btn-primary btn-lg bigFingerButton" role="button">Get Zeppelin</a>
+       <a href="./demos.html" class="btn btn-primary btn-lg bigFingerButton" role="button">Live Demo</a></p>
   </div>
 </div>
 {% endif %}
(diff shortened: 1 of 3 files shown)


Leemoonsoo:
I'd like to open this PR for a week at least to get more feedbacks about adding live demo features on the website while it's the first time 3rd party involves contents on the Apache Zeppelin website.

Any opinions, questions, feedbacks would be appreciated.
I can throw some questions here.

  - I'd like to add a logo of 3rd party on demos.html. Will there be any expected problem?
  - Live demo instance will needs maintenance like any other services. What could be the best way Zeppelin community communicate to the 3rd party who maintains it?
  - Shell we have guide line to add a 3rd party live demo? Such as 
    * Zeppelin shouldn't have modified in such way
    * etc
  - Shell we need notify some kind of warranty or tems of service?

zjffdu:
One suggestion is asking the 3rd party to specify their zeppelin version for this demo. If they update the demo, they should also update the version accordingly.

felixcheung:
It would probably be a good idea to have a click-through terms of service.
Should we also display the zeppelin version?

bzz:
Both, version and terms of service (to avoid liability) sounds as a good ideas to me

corneadoug:
@Leemoonsoo Could we maybe move the list of third party live demo to community instead of having an extra page?

Any update on the version and term of services?

Leemoonsoo:
@corneadoug I don't think live demo is related to contents of community page.

Any help on writing terms of services would be appreciated.

PR #1322, opened 2016-08-11
[ZEPPELIN-1320] Run zeppelin interpreter process as web front end user

prabhjyotsingh:
### What is this PR for?
While running a Notebook using shell, spark, python uses same user as which zeppelin server is running. Which means these interprets have same permission on file system as zeppelin server.
IMO users should be able to impersonate themselves as a complete security system.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Update doc
* [x] - FIX NPEs
* [x] - FIX CI

### What is the Jira issue?
* [ZEPPELIN-1320](https://issues.apache.org/jira/browse/ZEPPELIN-1320)

 - Enable shiro auth in shiro.ini
 - Add ssh key for the same user you want to try and impersonate (say user1).
```
adduser user1
ssh-keygen
ssh user1@localhost mkdir -p .ssh
cat ~/.ssh/id_rsa.pub | ssh user1@localhost 'cat >> .ssh/authorized_keys'
```
 - Start zeppelin server, try and run following in paragraph in a notebook
 - Go to interpreter setting page, and enable "User Impersonate" in any of the interpreter (in my example its shell interpreter)

```
%sh
whoami
```

Check that it should run as new user, i.e. "user1"

user impersonate

### Questions:


felixcheung:
shouldn't interpreter process be impersonating the user logging onto the web front end?

prabhjyotsingh:
@felixcheung Fair point, let me try and do it, will change the title to WIP for now.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -1094,14 +1075,15 @@ private String getInterpreterClassFromInterpreterSetting(InterpreterSetting sett
     return null;
   }
 
-  private Interpreter getInterpreter(String noteId, InterpreterSetting setting, String name) {
+  private Interpreter getInterpreter(String noteId, InterpreterSetting setting, String name,
+                                     String userName) {
```
bzz: Probably a nitpick, but horizontal alignment is controversial idea and generally [is discouraged by the styleguide](https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment) as it creates a "blast radius" of re-formatting in case of future changes i.e renaming a function.

prabhjyotsingh:
CI green! Ready for review.

bin/interpreter.sh
```diff
@@ -42,6 +42,9 @@ while getopts "hp:d:l:v" o; do
             . "${bin}/common.sh"
             getZeppelinVersion
             ;;
+        u)
+            ZEPPELIN_SSH_COMMAND="ssh ${OPTARG}@localhost "
+            ;;
         esac
```
zjffdu: This requires the login user must exist in the os account and  be able to ssh to localhost.  I am not sure whether this is a good way, but just feel the approach is a little strange compared to the impersonation implementation in hadoop.

jongyoul:
I agree that it's simple way to use ssh to support impersonation. but I'm worried about it. First, we should consider not to use ssh server in a local machine. It's disabled on Mac by default and in case of Windows users, they might not have any ssh server. Second, even if all of users can use connect their machine via ssh, all of users' name should be the same as system users. AFAIK, Some Zeppelin use cases, the system admin uses virtual users as well. Do you think of it?

prabhjyotsingh:
Yes, I thought about the usage in mac and windows, and initially started of with using `RUNAS ${userName}` for windows and `su - ${userName}` for *nix systems, but then it requires zeppelin server to run as root. Hence, implemented with `ssh ${userName}@localhost`.

Have not thought about the cases in which system admin uses virtual users.

Now since with this, we are able to propagate end web user to  RemoteInterpreterManagedProcess.start, we can choose to use some other mechanism in `interpreter.sh`/`interpreter.cmd` instead of "ssh", or may be make it configurable using some extra config in "zeppelin-env.sh"

What do you recommend, that would be a secure and all full proof mechanism by which we can run interpreter as different user ?

bin/interpreter.sh
prabhjyotsingh: @zjffdu yes, I agree, its not as implementation in hadoop, would you recommend something else ?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -36,8 +36,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.gson.Gson;
-import com.google.gson.reflect.TypeToken;
+import java.util.*;
```
bzz: Probably a nitpick but Zeppelin's [Java code conventions](https://github.com/apache/zeppelin/blob/master/CONTRIBUTING.md#code-convention) discourages [usage of wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports).

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -77,6 +40,24 @@
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreterProcessListener;
 import org.apache.zeppelin.scheduler.Job;
 import org.apache.zeppelin.scheduler.Job.Status;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sonatype.aether.RepositoryException;
+import org.sonatype.aether.repository.Authentication;
+import org.sonatype.aether.repository.RemoteRepository;
+
+import java.io.*;
```
bzz: Againa, here and below https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

jongyoul:
@prabhjyotsingh I don't know how to support different users' environments fully, actually. But I think it's better to use `RUNAS ~` and `su - ~` and using `ssh` without password make some security issues. In case Mesos, it uses that way to support restrict resources. But I never see using `ssh` without password. How do you think of it?

jongyoul:
@prabhjyotsingh Without issues above, Could you check this PR support `scoped` as well which uses multiple threads in one process?

Leemoonsoo:
If i add one more,
What do you guys think about adding an option `Impersonate` in the interpreter setting on GUI?

That'll give user flexibility of selecting current behavior (without impersonation) and new behavior. Otherwise, this PR will make incompatible user behavior change.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
prabhjyotsingh: Sure, I'll revert this, and check that my Editor (Intellij Idea) is also configured properly.

prabhjyotsingh:
> It's better to use RUNAS ~ and su - ~

@jongyoul  How about I make `use RUNAS ~ and su - ~` by default, but if in `zeppelin-env.sh` a property say `USE_SSH_IMPERSONATION` is set to true, then it will use `ssh web-user@localhost` in this way user gets to decide, what is best suited for their user case.

>Could you check this PR support `scoped` as well which uses multiple threads in one process?

Yes I've checked this with Shell and Python interpreter it was working as expected.

@Leemoonsoo, yes agreed, I too think this options should be there, and have implemented it as well. If you take a look at GIF attached in this PR description, it's doing that you are asking for :)

echarles:
Whatever `su` or `ssh` is used, I feel the main trick is the user provisioning on the host running the interpreter. Until now, the shiro authentication system had no impact on the user provisioning. This PR changes this.

I guess we all agree and are aware that adding user `foo` to shiro.ini, and enabling impersonation, will require to run `adduser foo` manually.

We should make this clear in the doc but also stress it in the UI (with a hover, or a clear text/link near the <checkbox> User Impersonate.

prabhjyotsingh:
@echarles , Yes agreed, will need to update in doc, and a extra toolbar near the check box where user can enable User Impersonate.

echarles:
To make ZEPPELIN-1337 Umbrella for multiple user support for zeppelin more readable, should we rename the following:

- ZEPPELIN-1340: "Run Hadoop-based interpreter process on Kerberos as web front end user"
- ZEPPELIN-1320:  "Run zeppelin interpreter process as web front end user"

echarles:
... and make ZEPPELIN-1320 a subtask of ZEPPELIN-1337

?

prabhjyotsingh:
Yes, you are right, let me do it right away.

jongyoul:
@prabhjyotsingh  I agree @echarles's idea. Interpreter tries to find hadoop dependencies first and if it passes, it uses `doAs`. Otherwise, let's talk about how to do it.  How do you think of it?

prabhjyotsingh:
Sure, In this PR I was only thinking about the otherwise case i.e. in the environment where hadoop dependencies where not present, and hence start interpreter as end-web-user.

echarles:
Btw, for the hadoop case (or spark on yarn case), this PR may give an issue for `doAs`.

Typically, you configure `hadoop.proxyuser.foo.hosts` and `hadoop.proxyuser.foo.group`, `foo` being the os/kerberos user under which you run your java code that calls doAs.

If we run ssh/su as the front-end user, we will not fullfill what the hadoop/yarn cluster is expecting.

We thus should have two checkboxes:
- One for the OS/kerberos impersonation (this PR only adresses OS).
- The other for Hadoop impersonation.

If you select one, I would expect the other one to be disabled.

prabhjyotsingh:
Agreed @echarles, the `doAs` part will be a problem, until [ZEPPELIN-1340](https://issues.apache.org/jira/browse/ZEPPELIN-1340) is resolved. Until then for security we may have to run half interpreter with "User Impersonate" enable from UI (for example shell, python interpreter), and for others use the standard `doAs` already implemented (like livy, spark, jdbc)

Leemoonsoo:
Instead of `USE_SSH_IMPERSONATION`, how about let user customize impersonation method?
For example,
```
ZEPPELIN_INTERPRETER_IMPERSONATION_CMD="su - ${ZEPPELIN_USER_NAME}"
```
by default. but user can override this env variable, like
```
ZEPPELIN_INTERPRETER_IMPERSONATION_CMD="ssh -p12345 ${ZEPPELIN_USER_NAME}@localhost"
```
It gives more flexibility i think. (e.g. give additional options like -p. use different command to impersonate)

prabhjyotsingh:
@Leemoonsoo yes thats a good suggestion. Let me try and do it.

astroshim:
I got following checkstyle error while building source.
```
[INFO] There are 1 checkstyle errors.
[ERROR] NotebookServer.java[1381] (sizes) LineLength: Line is longer than 100 characters (found 102).
```
@prabhjyotsingh Could you fix this?

prabhjyotsingh:
Closing this, will open a new one with merge of https://github.com/apache/zeppelin/pull/1265.

zjffdu:
Sorry for late comment. I was in vacation in the last 2 weeks.  I found this didn't work for spark interpreter. @prabhjyotsingh Did you try it for spark interpreter and other interpreters ?

prabhjyotsingh:
@zjffdu Yes, you are right, with SPARK_HOME/SPARK_SUBMIT it doesn't work.

zjffdu:
Then I think we should either revert this PR or fix it for spark interpreter as well. Because spark interpreter is the most important interpreter of zeppelin IMO.

prabhjyotsingh:
Sure make sense I'll try to fix it ASAP. https://issues.apache.org/jira/browse/ZEPPELIN-1701

PR #1329, opened 2016-08-12
ZEPPELIN-1318 - Add support for matplotlib displaying png images in python interpreter

agoodm:
### What is this PR for?
This PR adds support for plotting png images using the matplotlib helper function within a python interpreter (eg `z.show()`). The primary motivation for this is due to the overhead incurred from svg images, which can lag the notebooks if multiple, complicated images are generated (for example, multiple filled contour plots). png images are more lightweight, but of course come at a cost of image quality due to them being raster rather than vector like svg. The support for png images is incorporated through the use of a new optional argument to `z.show` called `fmt` which can be one of `'svg'` or `'png'`. The same code that is currently used in show is used for svg images while the code for png images relies on converting the image directly to a byte array and then entering the decoded byte string directly into an HTML image tag. Currently `fmt` defaults to `'png'` but I think we should consider discussing the pros and cons of each option in this PR. 

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-1318](https://issues.apache.org/jira/browse/ZEPPELIN-1318)

In a notebook cell, enter:
```python
%python
import numpy as np

plt.figure()
plt.plot(np.arange(10))
z.show(plt, fmt=fmt)
```
Where `fmt` may be one of `'svg'` or `'png'`, and any other input should result in a `ValueError`. I would also recommend testing the example in the screenshot below. 

zeppelin plot

### Questions:
* Does this needs documentation? Yes (if the changes to the `help()` docstring are not sufficient)

### DIFF
--- a/docs/interpreter/python.md
+++ b/docs/interpreter/python.md
@@ -87,7 +87,7 @@ print("".join(z.checkbox("f3", [("o1","1"), ("o2","2")],["1"])))
 
 ## Matplotlib integration
  The python interpreter can display matplotlib graph with the function `z.show()`.
- You need to have matplotlib module installed and a XServer running to use this functionality !
+ You need to have matplotlib module installed and a XServer running to use this functionality!
  
  ```python
 %python
@@ -97,12 +97,12 @@ plt.figure()
 z.show(plt)
 plt.close()
 ```
-z.show function can take optional parameters to adapt graph width and height
+The `z.show()` function can take optional parameters to adapt graph dimensions (width and height) as well as output format (png or optionally svg).
 
  ```python
 %python
 z.show(plt, width='50px')
-z.show(plt, height='150px')
+z.show(plt, height='150px', fmt='svg')
 ```
 <img class="img-responsive" src="../assets/themes/zeppelin/img/docs-img/pythonMatplotlib.png" />
 
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -19,6 +19,7 @@
 # Remove interactive mode displayhook
 import sys
 import signal
+import base64
 
 try:
     import StringIO as io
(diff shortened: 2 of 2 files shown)


felixcheung:
could you please update this PR and the JIRA to say `python` interpreter?

agoodm:
@felixcheung Done.

bzz:
@agoodm thank you for improving matplotlib integration!

Please correct me, but right now PR does 2 things:
 - add PNG support, as well as SVG
 - switches default to PNG

I see you point on performance, but do you think changing default can bring any compatibility issue for existing notebooks?

Other then that - the changes looks great to me!

agoodm:
@bzz The way I see it the pros and cons of each format can be summed up as follows:

### SVG 
Pros:
- Vector so image quality will be consistent regardless of resolution
- Displays more quickly for simple plots (eg. single line plots)

Cons:
- Very slow for complex plots involving many data points

### PNG
Pros:
- Lightweight regardless of the complexity of the plot

Cons:
- Does not scale to higher resolutions as well due to being raster
- _Slightly_ slower to display simple plots

To be honest, I am not really sure which option would be best to default to. png is the default format in most matplotlib backends so that is why I have chosen it for now, but I think it's worth thinking about. Either way though I don't think the choice of default format will matter in terms of backwards compatibility as the change should be seamless for users.  I am fine with either option honestly.

bzz:
Thank you for prompt and thoughtful comparison @agoodm !

> the choice of default format will matter in terms of backwards compatibility as the change should be seamless for users

This sounds great to me, let's keep PNG as default. 

Thank you for taking care of the docs in `help()` as well, could you also add it to [docs/interpreter/python.md#matplotlib-integration](https://github.com/apache/zeppelin/blob/master/docs/interpreter/python.md#matplotlib-integration) so it will show up on the [doc website](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/interpreter/python.html#matplotlib-integration)?

Here you can find more details on how to [update the website](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/development/howtocontributewebsite.html).

agoodm:
@bzz Done!

bzz:
Looks great to me, thank you @agoodm!

Will merge to master later today, if there is no further discussion

PR #1335, opened 2016-08-16
[ZEPPELIN-1336] hadoop library for spark interpreter is not match.

astroshim:
### What is this PR for?
This PR is for download hadoop libraries for yarn cluster.
This issue comes from https://github.com/apache/zeppelin/pull/1318.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1336

1 build zeppelin for `spark2.0&hadoop2.7`
```
mvn clean package -Pspark-2.0 -Phadoop-2.7 -Dhadoop.version=2.7.2 -Pyarn -Ppyspark -Pscala-2.11 -DskipTests 
```

2 verify hadoop library version for spark.
```
ls -al $ZEPPELIN_HOME/spark/target/lib/
```

### Questions:

### DIFF
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -954,6 +954,18 @@
           <version>${yarn.version}</version>
         </dependency>
 
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-yarn-server-common</artifactId>
+          <version>${yarn.version}</version>
+        </dependency>
+
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-yarn-server-web-proxy</artifactId>
+          <version>${yarn.version}</version>
+        </dependency>
+
 	<dependency>
 	  <groupId>org.apache.hadoop</groupId>
 	  <artifactId>hadoop-common</artifactId>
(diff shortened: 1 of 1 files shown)


spark/pom.xml
```diff
@@ -27,7 +27,7 @@
   </parent>
 
   <groupId>org.apache.zeppelin</groupId>
-  <artifactId>zeppelin-spark_2.10</artifactId>
+  <artifactId>zeppelin-spark_2.11</artifactId>
```
minahlee: Currently Zeppelin is built with Scala 2.10 by default and that's why it has `2.10` suffix. If you changed this line because of the build error, I recommend you to use `dev/change_scala_version.sh 2.11` prior to building with mvn.
FYI, there is [jira ticket](https://issues.apache.org/jira/browse/ZEPPELIN-1249) making Scala 2.11 to default so this change better to be handled there.

jongyoul:
I think spark-dependencies is a proper position for this change. How do you think?

spark/pom.xml
astroshim: Update spark2.11 is my mistake. It was for testing so I'll fix it. 
Thank you.

astroshim:
@jongyoul I will close this issue since removing yarn profile is working.
please do it at the https://github.com/apache/zeppelin/pull/1301 as @zjffdu mentioned.

PR #1339, opened 2016-08-18
[ZEPPELIN-1332] Remove spark-dependencies & suggest new way

AhyoungRyu:
### What is this PR for?
Currently, Zeppelin's embedded Spark is located under `interpreter/spark/`. 
For whom **builds Zeppelin from source**, this Spark is downloaded when they build the source with [build profiles](https://github.com/apache/zeppelin#spark-interpreter). I think this various build profiles are useful to customize the embedded Spark, but many Spark users are using their own Spark not Zeppelin's embedded one. Nowadays only Spark&Zeppelin beginners use this embedded Spark. For them, there are too many build profiles(it's so complicated i think). 
In case of **Zeppelin binary package**, it's included by default under `interpreter/spark/`. That's why Zeppelin package size is so huge. 

#### New suggestions
This PR will change the embedded Spark binary downloading mechanism like below.

1. `./bin/zeppelin-daemon.sh get-spark` or `./bin/zeppelin.sh get-spark`
2. create `ZEPPELIN_HOME/local-spark/` and will download `spark-2.0.1-hadoop2.7.bin.tgz` and untar 
3. we can use this local spark without any configuration like before (e.g. setting `SPARK_HOME`)

### What type of PR is it?
Improvement

### Todos
- [x] - trap `ctrl+c` & `ctrl+z` key interruption during downloading Spark
- [x] - test in the different OS 
- [x] - update related document pages again after get feedbacks

### What is the Jira issue?
[ZEPPELIN-1332](https://issues.apache.org/jira/browse/ZEPPELIN-1332)

1. `rm -r spark-dependencies` 
2. Apply this patch and build with `mvn clean package -DskipTests`
3. try`bin/zeppelin-daemon.sh get-spark` or `bin/zeppelin.sh get-spark`
4. should be able to run `sc.version` without setting external `SPARK_HOME`

- `./bin/zeppelin-daemon.sh get-spark`
```
$ ./bin/zeppelin-daemon.sh get-spark
Download spark-2.0.1-bin-hadoop2.7.tgz from mirror ...

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  178M  100  178M    0     0  10.4M      0  0:00:17  0:00:17 --:--:-- 10.2M

spark-2.0.1-bin-hadoop2.7 is successfully downloaded and saved under /Users/ahyoungryu/Dev/zeppelin-development/zeppelin/local-spark
```
- if `ZEPPELIN_HOME/local-spark/spark-2.0.1-hadoop2.7` already exists
```
$ ./bin/zeppelin-daemon.sh get-spark
spark-2.0.1-bin-hadoop2.7 already exists under local-spark.
```

### Questions:
- Does this needs documentation? Need to update some related documents (e.g. README.md, spark.md and install.md ?)


bzz:
@AhyoungRyu great initiative, but while making this changes, you have to think also about CI use case of zeppelin build as well.

I.e so far `/.spark-dist/` is under cache on TravisCI which is S3 bucket that gets synced automatically with the content of this folder while running a build. It you un-tar the whole archive there - it will take forever to sync \w S3 and will defeat the purpose of cache on CI side, making build times longer.

If you ask me - I would say that before doing such big changes as refactoring of the build structure we all need very clear understanding and explanation of `what is the benefit` and what problem does this change solves. 

So far I have not understood the answer to the questions above from PR description (may be my fault). But, in case of voting for such change, will make me at least `-0` for if, if not `-1` due to potential bugs that such changes will bring.

If that is reduction of convenience binary size - then we need to know how much does the size changes with the proposed changes to understand if that is worth. If that impacts CI build times - we also need to know how much.

Also regarding user experience - while running `zeppelin-demon.sh` user does not usually expect it to be network-dependant and download 100Mb archives - is there at least a user notification\progress indicator? Otherwise there going to be bug reports like "Zeppelin is not starting" as soon as such change is introduced. 
And how about Windows users of Zeppelin? How about EMR\Dataproc\Juju\BigTop users, will the proposed change affect them?

Please take it with the grain of salt, and of course I will be happy to help addressing each item addressed one by one.

AhyoungRyu:
@bzz Thank you for such precise comment! Let me break down your feedback one by one(just for making it clear) :)

1.
>/.spark-dist/ is under cache on TravisCI which is S3 bucket that gets synced automatically with the content of this folder while running a build. 

Right. That's my bad. I'll change the dir to another. Then how about `ZEPPELIN_HOME/interpreter/spark/` as like before? 

2, 3, 4.
>what is the benefit and what problem does this change solves?

Actually I also tried to describe well about the current problem & the advantage of this change in Jira issue and the PR description, but i guess i didn't. I should've explain more clearly. Let me explain more in here with actual digit. (I'll update the Jira & PR description as well)

 - **What was the problem?**

As you said in the above, yes. The main problem is the Zeppelin binary package size. The latest version of Zeppelin bin size was
```
zeppelin-0.6.1-bin-all.tgz: 517MB
zeppelin-0.6.1-bin-netinst.tgz: 236MB
```
Didn't we ask ASF infra team(?) every release because of Zeppelin's huge package size?

 - **What is the benefit?**

When I created binary package without `spark-dependencies`, the each bin package size was
```
zeppelin-0.6.1-bin-all.tgz: 344MB
zeppelin-0.6.1-bin-netinst.tgz: 64MB
```
As you can see in the above those two cases' size diff is about `170MB`!  Moreover, users don't need to type build profiles i.e. `-Pr` or `-Psparkr`. I saw many users who are trying to use `%sparkr` in Zeppelin, they hit NPE because they didn't build with `-Psparkr`. It's truly confuse maybe they don't know well about the maven build mechanism. But with this change, they don't need to know about the complicating maven build profiles. 

5.
> Also regarding user experience - while running zeppelin-demon.sh user does not usually expect it to be network-dependant and download 100Mb archives - is there at least a user notification\progress indicator

So far, I just added below line to show in console after users start `zeppelin-daemon.sh`
```
echo "There is no SPARK_HOME in your system. After successful Spark bin installation, Zeppelin will be started."
``` 
Then it starts downloading Spark binary from the mirror site. I'm planning to add some description to README as we have provided many build profiles information in there. I also agree there must be better way to notify that instead of just writing about "We will download 100MB Spark binary package if you don't set SPARK_HOME yet" on README. 

After first I came up with removing `spark-dependencies` to reduce Zeppelin bin package size, I spent long time to think about how can we substitute the preexisting way seamlessly to provide embedded Spark in Zeppelin as like before. Please regard this PR as the first initiative. And will be appreciated if you can share your awesome idea about this issue! :)

bzz:
Thank you for kind explanation and feedback. I think you proposal and implementation with recent updates makes perfect sense.

Please keep up a good work and ping me back for the final review, once you think it's ready!

AhyoungRyu:
@bzz Thank you for saying so! Then I'll continue my work in here and let you know :)

bin/common.sh
```diff
@@ -106,7 +106,13 @@ function getZeppelinVersion(){
     exit 0
 }
 
-# Text encoding for 
+function downloadSparkBinary() {
+  if [[ -z "${SPARK_HOME}" ]]; then
```
Leemoonsoo: One possible issue is
SPARK_HOME can not only be defined in `zeppelin-env.sh` but also in GUI interpreter setting page (related bug is being handled by [ZEPPELIN-1334](https://issues.apache.org/jira/browse/ZEPPELIN-1334)).

If SPARK_HOME is not defined in `zeppelin-env.sh` but in interpreter setting page on GUI, this script will not recognize SPARK_HOME at this point. Because `downloadSparkBinary()` is invoked by either `bin/zeppelin-daemon.sh` or `bin/zeppelin.sh`, but SPARK_HOME defined from GUI will be propagated to `bin/interpreter.sh` only.

Another possible issue is
Some user may doesn't even have a spark interpreter installed. (e.g. netinst package and user installed jdbc interpreter only) In this case downloading spark binary because of SPARK_HOME is not defined doesn't make any sense.

Considering these two possible issue around this conditional statement, i suggest change this condition to check installation of spark interpreter, not check SPARK_HOME. for example

```
if [[ -d "${ZEPPELIN_HOME}/interpreter/spark" ]]; then
```

And inside of `bin/download-spark.sh` script, how about just ask user explicitly whether user want to download spark binary for local mode or not? And if user answers 'N' we can create a small text file somewhere under `interpreter/spark` to remember the answer.

What do you think?

AhyoungRyu: @Leemoonsoo Thank you for pointing the possible issues! Didn't know that parts actually. Checking the existence of `${ZEPPELIN_HOME}/interpreter/spark` more makes sense than `SPARK_HOME`. And getting user's answer as well. Will update that two parts :)

bin/download-spark.sh
```diff
@@ -0,0 +1,78 @@
...
+ZEPPELIN_HOME="$(cd "${FWDIR}/.."; pwd)"
+ZEPPELIN_ENV="${ZEPPELIN_HOME}/conf/zeppelin-env.sh"
+ZEPPELIN_ENV_TEMP="${ZEPPELIN_HOME}/conf/zeppelin-env.sh.template"
+
+# Downloads file from the given URL.
+# Ties 3 times with 1s delay, 20s read and 15s connection timeouts.
+# Arguments: url - source URL
+function download_with_retry() {
+  local url="$1"
+  wget --retry-connrefused --waitretry=1 --read-timeout=20 --timeout=15 -t 3 "${url}"
```
bzz: `wget` is not installed by default on some Linux distributions, but usually `cUrl` does. 

For compatibility reasons this code needs to:
 - check if download tool of choice is installed
 - if not - report it to the user properly (and fail gracefully?)
 - output of download command must be checked (as with [any other command with side effects](https://google.github.io/styleguide/shell.xml#Checking_Return_Values)) and in case of failure - reported to user properly

AhyoungRyu:
@bzz @Leemoonsoo 
Sorry for my late response. I spent some time to test various cases on the different OS.
I think it's ready for review(CI is green at last!). 
I updated the PR description accordingly. Hope it helps you to remind the purpose of this PR :)

Here is the list of changes after my initial commits

 - Directory of Spark bin
I changed the dir of Spark bin from `interpreter/spark/` to `local-spark`. Since `mvn clean` will remove `interpreter/`, users will see “Do you want to download local Spark?” whenever they re-build and restart Zeppelin. So I think creating a new dir(`local-spark`) would be better in this case.

 - Supporting Windows users
I wanted to create `download-spark.cmd` for Windows. But sadly we can’t use many shell commands such as `curl`, `tar` and `sed` in the batch script. It was hard to find 100% compatible commands for Windows. Maybe we can guide the Windows users to install those commands by themselves, but it’s a bit overdoing i think. Actually `download-spark` script is only for downloading the latest version of Spark and set `SPARK_HOME`. So I updated some docs to explain this as a alternative way. 

 - Documentations
I think it’s quite big change that ppl need to enter “Yes/No” when they start Zeppelin. Even though it’s only one time. So I updated `README.md`, `install.md` and `spark.md`. The below screenshot is `spark.md`. 
screen shot 2016-09-13 at 8 22 34 pm


And as @bzz said before, 
>How about EMR\Dataproc\Juju\BigTop users, will the proposed change affect them? 

Do we need to provide this local Spark mode for them? Actually it's my question.. :D
And please feel free to point me for anything if it's needed.

bin/download-spark.sh
```diff
@@ -0,0 +1,144 @@
...
+function create_local_spark_dir() {
+  if [[ ! -d "${SPARK_CACHE}" ]]; then
+    mkdir -p "${SPARK_CACHE}"
+  fi
+}
+
+function save_local_spark() {
+  local answer
+
+  echo "There is no local Spark binary in ${ZEPPELIN_HOME}/${SPARK_CACHE}"
```
Leemoonsoo: I think user might not know what 'local Spark' is and why they need it.
How about explain, 'to use spark interpreter in local mode (without external spark installation), spark binary need to be downloaded', or such way that gives user enough information to understand and make a decision?

bin/download-spark.sh
```diff
@@ -0,0 +1,144 @@
...
+function check_zeppelin_env() {
+  if [[ ! -f "${ZEPPELIN_ENV}" ]]; then
+    echo -e "\n${ZEPPELIN_ENV} doesn't exist\nCreating ${ZEPPELIN_ENV} from ${ZEPPELIN_ENV_TEMP}..."
+    cp "${ZEPPELIN_HOME}/${ZEPPELIN_ENV_TEMP}" "${ZEPPELIN_HOME}/${ZEPPELIN_ENV}"
+  fi
+}
+
+function set_spark_home() {
+  local line_num
+  check_zeppelin_env
```
Leemoonsoo: After `check_zeppelin_env`, if there is already `conf/zeppelin-env.sh` exists, which means user explicitly created it, then i think we shouldn't change it.

AhyoungRyu:
@Leemoonsoo "zero configuration like before" makes sense. Let me update and will ping you again.

bin/interpreter.sh
```diff
@@ -112,34 +113,20 @@ if [[ "${INTERPRETER_ID}" == "spark" ]]; then
...
-      elif [[ -d "/etc/hadoop/conf" ]]; then
-        export HADOOP_CONF_DIR="/etc/hadoop/conf"
-      fi
-    fi
-
-    if [[ -n "${HADOOP_CONF_DIR}" ]] && [[ -d "${HADOOP_CONF_DIR}" ]]; then
-      ZEPPELIN_INTP_CLASSPATH+=":${HADOOP_CONF_DIR}"
-    fi
-
-    export SPARK_CLASSPATH+=":${ZEPPELIN_INTP_CLASSPATH}"
```
Leemoonsoo: Any special reasons removing above code blocks?

AhyoungRyu: @Leemoonsoo 
There are two type of code blocks I removed. One is for exporting `HADOOP_CONF_DIR` and the other is `SPARK_CLASSPATH`. As you know, those two are all for the order Spark version support. 
 - Regarding the setting `HADOOP_CONF_DIR` is not the problem for this change since it's embraced in the if statement(maybe we can remove this in another PR as refactoring). Anyway I reverted this part in [f5dcd04e](https://github.com/apache/zeppelin/pull/1339/commits/f5dcd04e1ec1e9575bc8d8a8e858b0b83f60f578).
 
- But `SPARK_CLASSPATH` can conflict with `SPARK_SUBMIT` that I used in [here](https://github.com/apache/zeppelin/pull/1339/files#diff-8d3bd42a5de342ebad38e1d113d9f75cR117). `SPARK_CLASSPATH` is deprecated after `spark-1.x` and  `SPARK_SUBMIT` is recommended instead. 
```
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 1.0.0
      /_/

Using Scala version 2.10.4 (IBM J9 VM, Java 1.7.0)
WARN spark.SparkConf: 
SPARK_CLASSPATH was detected (set to 'path-to-proprietary-hadoop-lib/*:/path-to-proprietary-hadoop-lib/lib/*').
This is deprecated in Spark 1.0+.

Please instead use:
 - ./spark-submit with --driver-class-path to augment the driver classpath
 - spark.executor.extraClassPath to augment the executor classpath
```
Please see [this mail thread](http://apache-spark-user-list.1001560.n3.nabble.com/SPARK-CLASSPATH-Warning-td9233.html) for the more information about this. 
Since I set the local spark version as `spark-2.0.0`, we need to use only one : `SPARK_CLASSPATH` or `SPARK_SUBMIT`. That's why I removed the `SPARK_CLASSPATH` and set `SPARK_SUBMIT`.

AhyoungRyu:
I think this PR is working well as expected(at least to me haha). So ready for review again.
@moon If you possible, could you please check this one again? :)

AhyoungRyu:
I think [ZEPPELIN-1101](https://issues.apache.org/jira/browse/ZEPPELIN-1101) can also be resolved by this change. 

>It looks related to ZEPPELIN-1099 which is about removing dependencies from Spark. I think we don't need to build spark-dependencies by ourselves. we'd better support script to download spark binary and set SPARK_HOME. How about it?

@jongyoul As you replied like above in ZEPPELIN-1101, could you please take a look this one? :)

AhyoungRyu:
ping 👯

jongyoul:
@AhyoungRyu Thanks for your effort. LGTM. But I think it would be better to support non-interactive mode for running the server because some of users launches Zeppelin as a start-up service for their server and interactive mode would break this feature.

AhyoungRyu:
@jongyoul Thanks for your feedback! Yeah I didn't try to cover that case. So you mean we need to support ppl who are using [this upstart option](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/install/install.html#optional-start-apache-zeppelin-with-a-service-manager), am I right? :)

AhyoungRyu:
@Leemoonsoo @jongyoul Sorry for my late update. 
I've just added new option: `get-spark` to [zeppelin-daemon.sh](https://github.com/apache/zeppelin/pull/1339/files#diff-bd1714fd11d1853b691468647374113dR23) and [zeppelin.sh](https://github.com/apache/zeppelin/pull/1339/files#diff-1724182f3ebaf54f5c9e202dcdf82415R46) to download local Spark binary. I think this is more simpler than getting user's answer and then separating "interactive mode" and "non-interactive mode" that @jongyoul mentioned in [here](https://github.com/apache/zeppelin/pull/1339#issuecomment-250672904).

So to sum up, ppl can download local Spark with `./bin/zeppelin-daemon.sh get-spark` or `./bin/zeppelin.sh get-spark` with my latest update. If this way is okay, i'll update related docs pages accordingly. Maybe we need to let ppl know the existence of `get-spark` option by updating documentation i think.

What do you think?

jongyoul:
@AhyoungRyu Basically, it's simple solution but you need to update docs. Except that, LGTM

AhyoungRyu:
@jongyoul Thanks. Yeah i just wanted to get feedback about the change before updating the docs. 
@Leemoonsoo What do you think?

Leemoonsoo:
@AhyoungRyu Could you rebase and see if CI test goes green?

AhyoungRyu:
@Leemoonsoo Rebased but there is an issue raised from #1564 like below 

```
16/11/05 06:04:34 ERROR PySparkInterpreter: Error
java.util.NoSuchElementException: spark.submit.pyFiles
	at org.apache.spark.SparkConf$$anonfun$get$1.apply(SparkConf.scala:235)
	at org.apache.spark.SparkConf$$anonfun$get$1.apply(SparkConf.scala:235)
	at scala.Option.getOrElse(Option.scala:121)
	at org.apache.spark.SparkConf.get(SparkConf.scala:235)
	at org.apache.zeppelin.spark.PySparkInterpreter.setupPySparkEnv(PySparkInterpreter.java:172)
	at org.apache.zeppelin.spark.PySparkInterpreter.createGatewayServerAndStartScript(PySparkInterpreter.java:209)
	at org.apache.zeppelin.spark.PySparkInterpreter.open(PySparkInterpreter.java:159)
	at org.apache.zeppelin.spark.PySparkInterpreterTest.setUp(PySparkInterpreterTest.java:98)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:264)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:153)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:124)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:200)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:153)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 15.942 sec <<< FAILURE! - in org.apache.zeppelin.spark.PySparkInterpreterTest
```

Seems #1564 and this change has conflict by removing `spark-dependency`. Let me fix this first.

astroshim:
PySparkInterpreter need a [python library for pyspark](https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java#L483) so I think we need [pyspark] (https://github.com/apache/zeppelin/blob/1cde24665180e8f10651012f53d7bcb58ea2eb44/spark-dependencies/pom.xml#L819). 
What do you think?

AhyoungRyu:
@astroshim Appreciate for your help! I've just pushed it and let's wait until it finished :)

AhyoungRyu:
All test passes(except for selenium test) in my own travis [AhyoungRyu/zeppelin/builds](https://travis-ci.org/AhyoungRyu/zeppelin/builds/174094481), but Zeppelin travis doesn't even started... [apache/zeppelin/build](https://travis-ci.org/apache/zeppelin/builds/174094499).

AhyoungRyu:
@astroshim It passed at last!! Thanks again. 

Will update the related docs if there are no further discussions about this changes :)

AhyoungRyu:
Updated the related docs pages ([README.md](https://github.com/apache/zeppelin/pull/1339/files#diff-04c6e90faac2675aa89e2176d2eec7d8), [spark.md](https://github.com/apache/zeppelin/pull/1339/files#diff-83df2e7970d5a53a9028d05098bc626d), [upgrade.md](https://github.com/apache/zeppelin/pull/1339/files#diff-f472957a611b3e4d6c1171edca51cf93) and [install.md](https://github.com/apache/zeppelin/pull/1339/files#diff-f472957a611b3e4d6c1171edca51cf93)) and CI has passed now.

bzz:
Guys, what great work here, simplifying the build! 

A quick question @AhyoungRyu as it's kind of a big change, and I'm sorry if that was explained before, but could you please recap:

```
AhyoungRyu commented on Sep 17
@Leemoonsoo Thanks for your quick feedback! 
The "zero configuration like before" makes sense. Let me update and will ping you again.
```

 1. Is the comment above is how it works now? Meaning, does on the first run of `./bin/zeppelin-deamon.sh` or `./bin/zeppelin.sh` a download of Apache Spark (100+Mb) happen, without asking a user?
 2. does this also mean that on CI it will happen on every run of SeleniumTests as well?
 3. `-Ppyspark` disappeared, but I remember it was added because we need to re-pack some files from Apache Spark to incorporate them in Zeppelin build in order for it to work on a cluster. Is it not the case any more? For Spark standalone and YARN, etc

Thanks in advance!

AhyoungRyu:
@bzz Yeah I also wanted to get more and more feedbacks for this change since it's a huge change as you said. Thanks for asking and i'm willing to explain again :)

> ** 1.**  Is the comment above is how it works now? Meaning, does on the first run of ./bin/zeppelin-deamon.sh or ./bin/zeppelin.sh a download of Apache Spark (100+Mb) happen, without asking a user?

First time, I intended to ask sth like "Do you want to download local Spark?" when user starts Zeppelin daemon. But there are lot's of things to think about more since this question will be added before Zeppelin server start. e.g. [Some ppl are using Zeppelin as a start up service](https://github.com/apache/zeppelin/pull/1339#issuecomment-250672904) with their script as @jongyoul said. This kind of interactive mode will bother their env. 
So I decided to download this local Spark with `./bin/zeppelin-daemon.sh get-spark` or `./bin/zeppelin.sh get-spark`. With `get-spark` option, users don't need to be asked and they can choose whether they download this local mode Spark or not. Also they can use this local Spark without any configuration aka `zero configuration`. But we need to notice them the existence of `get-spark` option. That's why I updated documentation pages to let them know. 

> **2.** does this also mean that on CI it will happen on every run of SeleniumTests as well?
This change won't effect to CI build. I added `./bin/download-spark.sh` to download Spark only when the user run `./bin/zeppelin-daemon.sh get-spark`.  

> **3.** -Ppyspark disappeared, but I remember it was added because we need to re-pack some files from Apache Spark to incorporate them in Zeppelin build in order for it to work on a cluster. Is it not the case any more? For Spark standalone and YARN, etc

`pyspark` profile only exists in `spark-dependency` (Please see [here](https://github.com/apache/zeppelin/blob/master/spark-dependencies/pom.xml#L820)). Since `spark-dependencies` won't be existed anymore, `-Ppyspark` needs to be removed accordingly I guess.

tae-jun:
Tested and worked as expected 😄

Fantastic work! But, there are other benefits caused by this change.

Before this change, users couldn't use `SPARK_SUBMIT_OPTIONS` env variable using embedded Spark. (am i right?)

But now it's  possible! I tested with `export SPARK_SUBMIT_OPTIONS="--driver-memory 4G"` and on Spark UI, I could check It works.

image

Therefore, I think it would be better to update `conf/zeppelin-env.sh`. There is a comment which is:

```sh
## Use provided spark installation ##
## defining SPARK_HOME makes Zeppelin run spark interpreter process using spark-submit
##
# export SPARK_HOME                             # (required) When it is defined, load it instead of Zeppelin embedded Spark libraries
# export SPARK_SUBMIT_OPTIONS                   # (optional) extra options to pass to spark submit. eg) "--driver-memory 512M --executor-memory 1G".
# export SPARK_APP_NAME                         # (optional) The name of spark application.

## Use embedded spark binaries ##
## without SPARK_HOME defined, Zeppelin still able to run spark interpreter process using embedded spark binaries.
## however, it is not encouraged when you can define SPARK_HOME
##
```

This should be updated properly. In my opinion, it doesn't need to encourage use external spark anymore :-)

And, is it possible to use embedded spark without `get-spark`? If not, I think it should be written on README clearly.

LGTM 👍

AhyoungRyu:
@tae-jun Appreciate your nice feedback! Will update again `zeppelin-env.sh` and `install.md` instead of `README.md` as you suggested(since #1615 is trying to make it simpler to deliver only key content).

AhyoungRyu:
### To whom may concern about the breaking current UX with this change

This change has many benefits comparing to current embedded Spark as I wrote in the PR description (and @tae-jun mentioned in [this comment](https://github.com/apache/zeppelin/pull/1339#issuecomment-259486249) as well. Thanks!).
But as always, this kind of big change brings downside as well (e.g. breaking current UX). So I wanna write down how we can address some major cases as below. I think it would be better to share my opinion and get more feedback before merging. :)

1. New Spark/Zeppelin user, running Zeppelin for the first time
 : Quite easy to cover and already handled by updating the related docs pages I guess.

2. Existing Spark/Zeppelin user, running new Zeppelin installation (e.g. upgrading version)
 :  Definitely this case is harder to handle than 1. As the user already has expectation, that local mode will **just works** and surely they won't read the docs. To resolve this, I’ll update `bin/download-spark.sh` to print sth like “You don’t have local-spark/, you can download embedded Spark with `get-spark` option.” When the user run  `./bin/zeppelin-daemon.sh start`. And this sentences can be removed in the future when Zeppelin users can be getting accustomed with `get-spark` option.

3. Docker user, starting `bin/zeppelin.sh` inside the container
: This one can be also hard to handle because the user might assume that Spark just works. So I would suggest start applying this change to #1538 as a first step. Since it can be a Zeppelin-provided official docker script.

4. CI issue
Since @bzz raised some concern about CI issue, let me answer again in here to make sure :)
The reason I removed `-Ppyspark` in `.travis` is `pyspark` profile is only existed in `spark-dependencies/pom.xml`. So `pyspark` profile won’t be anymore after this PR merged. Actually the Pyspark testcase that @astroshim added recently had some conflict with this change. But we solved by simply adding `export SPARK_HOME=`pwd`/spark-$SPARK_VER-bin-hadoop$HADOOP_VER` to `.travis.yml` so that travis can run it before running the script. So there are no more CI issues especially concerning about removing `spark-dependencies` related build profiles.

tae-jun:
@AhyoungRyu Thanks for taking care of my feedback 😄

AhyoungRyu:
CI is green now, so ready for review.
I updated related docs again based on #1615 and @tae-jun 's feedback as well.

@bzz Could you take a look this again? 
As I mentioned in [this comment](https://github.com/apache/zeppelin/pull/1339#issuecomment-259683752), I added `You do not have neither local-spark, nor external SPARK_HOME set up.\nIf you want to use Spark interpreter, you need to run get-spark at least one time or set SPARK_HOME.` This msg will be printed when the user starts Zeppelin if he doesn't have neither `local-spark/` yet nor set external `SPARK_HOME` in his machine. Please see [my latest commit](https://github.com/apache/zeppelin/pull/1339/commits/2747d9eec49aa04f92ac93408f4c00cb101cb23e) :) 

Maybe this msg can be removed in the future, when many Zeppelin users can get accustomed to this change.

AhyoungRyu:
ping 💃

docs/install/upgrade.md
```diff
@@ -48,9 +48,10 @@ So, copying `notebook` and `conf` directory should be enough.
 ## Migration Guide
 
 ### Upgrading from Zeppelin 0.6 to 0.7
-
- - From 0.7, we don't use `ZEPPELIN_JAVA_OPTS` as default value of `ZEPPELIN_INTP_JAVA_OPTS` and also the same for `ZEPPELIN_MEM`/`ZEPPELIN_INTP_MEM`. If user want to configure the jvm opts of interpreter process, please set `ZEPPELIN_INTP_JAVA_OPTS` and `ZEPPELIN_INTP_MEM` explicitly. If you don't set `ZEPPELIN_INTP_MEM`, Zeppelin will set it to `-Xms1024m -Xmx1024m -XX:MaxPermSize=512m` by default.
+ - From 0.7, the support on Spark 1.1.x to 1.3.x is deprecated.
+ - Zeppelin embedded Spark won't work anymore. You need to run `./bin/zeppelin-daemon.sh get-spark` or `./bin/zeppelin.sh get-spark` at least one time. Please see [local Spark mode](../interpreter/spark.html#local-spark-mode) for more detailed information.
```
bzz: As this looks like a substantial change, may be we could use a bit stronger language here - i.e

```
Apache Zeppelin releases do not come with Apache Spark build-in by default any more. In order to be able to run Apache Spark paragraphs, please either run `./bin/zeppelin-daemon.sh get-spark` or point `$SPARK_HOME` to Apache Spark installation. See .... for more details ..
```

What do you think?

bzz:
Thank you @AhyoungRyu for great job and taking care in addressing the [user experience concerns](https://github.com/apache/zeppelin/pull/1339#issuecomment-259683752)!

1ambda:
Let me also review this great PR and then give some feedbacks 👍

AhyoungRyu:
@bzz Just updated `upgrade.md` as your feedback.
@1ambda Sure. Thanks! Please do :)

Leemoonsoo:
In case of

  1) Don't have plan to use spark interpreter, just want to use other interpreters like python, big query.
  2) Set SPARK_HOME in interpreter property instead of conf/zeppelin-env.sh

User may not interested in local-spark. but user will keep seeing messages 

```
Lees-MacBook:pr1339 moon$ bin/zeppelin-daemon.sh start

You do not have neither local-spark, nor external SPARK_HOME set up.
If you want to use Spark interpreter, you need to run get-spark at least one time or set SPARK_HOME.

Zeppelin start                                             [  OK  ]
Lees-MacBook:pr1339 moon$ bin/zeppelin-daemon.sh stop
Zeppelin stop                                              [  OK  ]
Lees-MacBook:pr1339 moon$ bin/zeppelin-daemon.sh start

You do not have neither local-spark, nor external SPARK_HOME set up.
If you want to use Spark interpreter, you need to run get-spark at least one time or set SPARK_HOME.

Zeppelin start                                             [  OK  ]
```

@AhyoungRyu  What do you think?

1ambda:
Short summary and small thought about #1399 

1. Using symlink like `local-spark/master` would be safe i think. It enables user replace his local spark without renaming directories. Currently we are using hard coded name.

```
SPARK_CACHE="local-spark"
SPARK_ARCHIVE="spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}"
```

2. about UX, 

- most (experienced in zeppelin) users do not use local spark
- for new comers, we can provide embedded spark using docker will be shipped by #1538 
- and storking `get-spark` is not too hard even if new users do not use the docker images.

3. Now users need to type `get-spark`. it works as described 

```
$  zeppelin-review git:(pr/1339) ./bin/zeppelin-daemon.sh start
Log dir doesn't exist, create /Users/1ambda/github/apache-zeppelin/zeppelin-review/logs
Pid dir doesn't exist, create /Users/1ambda/github/apache-zeppelin/zeppelin-review/run

You do not have neither local-spark, nor external SPARK_HOME set up.
If you want to use Spark interpreter, you need to run get-spark at least one time or set SPARK_HOME.

Zeppelin start                                             [  OK  ]
$  zeppelin-review git:(pr/1339) ./bin/zeppelin-daemon.sh stop
Zeppelin stop                                              [  OK  ]
$  zeppelin-review git:(pr/1339) ./bin/zeppelin-daemon.sh get-spark
Download spark-2.0.1-bin-hadoop2.7.tgz from mirror ...

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  178M  100  178M    0     0  7157k      0  0:00:25  0:00:25 --:--:-- 6953k

spark-2.0.1-bin-hadoop2.7 is successfully downloaded and saved under /Users/lambda/github/apache-zeppelin/zeppelin-review/local-spark

$  zeppelin-review git:(pr/1339) ./bin/zeppelin-daemon.sh start
Zeppelin start                                             [  OK  ]
```

AhyoungRyu:
I'm closing this PR since there'll be better solution for this (e.g. [ZEPPELIN-1993](https://issues.apache.org/jira/browse/ZEPPELIN-1993)) :)

PR #1349, opened 2016-08-19
Add spark dependency jars in driver classpath also in spark-submit mode

gbloisi:
### What is this PR for?
This commit adds spark dependency jars into the driver classpath of spark-submit. It is needed to let driver app find classes distributed along with spark, such as classes required for hive support.

### What type of PR is it?
[Bug Fix]

When SPARK_HOME is set, zeppelin.spark.useHiveContext is set to true (default), and hive-site.xml is present in the conf dir Spark is expected to use the hive catalog as described in the hive-site.xml.
Without this commit spark is configured to use always the in-memory catalog, because it does not find hive support classes so it defaults to in memory embedded hive catalog.

### DIFF
--- a/bin/interpreter.cmd
+++ b/bin/interpreter.cmd
@@ -68,6 +68,8 @@ if not exist "%ZEPPELIN_LOG_DIR%" (
 )
 
 if /I "%INTERPRETER_ID%"=="spark" (
+    call "%bin%\functions.cmd" ADDJARINDIR "%INTERPRETER_DIR%\dep"
+
     if defined SPARK_HOME (
         set SPARK_SUBMIT=%SPARK_HOME%\bin\spark-submit.cmd
         for %%d in ("%ZEPPELIN_HOME%\interpreter\spark\zeppelin-spark*.jar") do (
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
Hi, @gbloisi Thanks for the contribution.

Zeppelin can be built with one version of Spark. The specific version of spark dependencies will be collected into "${INTERPRETER_DIR}/dep". When user doesn't have local spark installation, Zeppelin will load them in the classpath.
When user have spark installation, Zeppelin allows user set SPARK_HOME to a different version that Zeppelin has been built with. It supposed to work. 

So when SPARK_HOME is set, Zeppelin will run interpreter process with SPARK_HOME/bin/spark-submit command and it'll include spark dependencies under SPARK_HOME in the classpath. And the spark dependencies version can be different from dependencies in "${INTERPRETER_DIR}/dep". So we probably don't want to add "${INTERPRETER_DIR}/dep" into the classpath in this case.

gbloisi:
Thanks for the explanation. I was using https://github.com/dylanmei/docker-zeppelin docker image and now I just realize it is shipping a spark-2.0.0 version not including hive dependecies (spark-2.0.0-bin-without-hadoop.tgz) and they are not even distributed with the separate hadoop package (hadoop-2.7.2.tar.gz). This explains why I was not able to instantiate a proper hive metastore. 
The confusing part is that instead the spark distribution with hadoop (spark-2.0.0-bin-hadoop2.7.tgz) is distributed with hive jars as well.

So please discard this pull request.

PR #1351, opened 2016-08-22
[ZEPPELIN/1356] The graph legend truncates at the nearest period (.) in its grouping

Peilin-Yang:
### What is this PR for?
Fix the issue: in line graph if user uses the numbers that contains period(.), e.g. 3.14 in the groups the legend will only show 3 instead of 3.14.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1356

Localhost test with screenshot

Before
screen shot 2016-08-25 at 6 51 36 pm
After
screen shot 2016-08-25 at 6 45 19 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1835,7 +1835,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       if (groups.length === 1 && values.length === 1) {
         for (d3gIndex = 0; d3gIndex < d3g.length; d3gIndex++) {
           colName = d3g[d3gIndex].key;
-          colName = colName.split('.').slice(0, -1).join('.'); // [ZEPPELIN-1356]: in case the column contains period
+          colName = colName.split('.').slice(0, -1).join('.');
           d3g[d3gIndex].key = colName;
         }
       }
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1835,7 +1835,7 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
       if (groups.length === 1 && values.length === 1) {
         for (d3gIndex = 0; d3gIndex < d3g.length; d3gIndex++) {
           colName = d3g[d3gIndex].key;
-          colName = colName.split('.')[0];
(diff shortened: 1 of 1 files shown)


bzz:
Looks great to me, thank you for taking care @Peilin-Yang !

corneadoug:
@bzz I'd like to review it first

Peilin-Yang:
@bzz is that possible to apply this to 0.6.0-SNAPSHOT. We are still using that version in Twitter:)

bzz:
@Peilin-Yang make sense, have updated jira fix-for version.

@corneadoug absolutely, please take your time for review.

corneadoug:
Tested LGTM, @Peilin-Yang Can you just remove the comment in that change?

Peilin-Yang:
@corneadoug sure, will do and push

prabhjyotsingh:
Can you take care of the trailing decimal as well ?

screen shot 2016-08-26 at 12 08 51 pm

corneadoug:
@prabhjyotsingh which trailing decimal and what is supposed to be the result?

prabhjyotsingh:
IMO, instead of UI showing "1.41234312421341234213" it should get truncated/ceil/floor to say not more than 3 places of decimal.

corneadoug:
@prabhjyotsingh It is already, that's why there is 1.412 shown in the tooltip.
The other one is the group name.

The original purpose of this PR is that in case of the group name we should not remove what is after the coma.

To give a non numerical example:

##### Before the PR
Group Name would go from `hello.world` -> `hello` in the legend

##### After the PR
Group Name would be shown as `hello.world` in the legend

prabhjyotsingh:
IMO shouldn't it be the other way, in tool tip, show the whole number without truncate, generally not to huge number.
Otherwise LGTM.

corneadoug:
@prabhjyotsingh I guess the tooltip value that is truncated can be the subject of another issue.

Peilin-Yang:
@prabhjyotsingh I agree with @corneadoug 
truncation on numeric values in the legend to me is another improvement/feature.
What this PR does is to fix the bug of existing code.
Does that make sense to you?

Peilin-Yang:
@corneadoug should I make the CI build successful? I do not know the failure is because of my commit.

prasadwagle:
Thanks @Peilin-Yang! This problem and the one in #1363 were reported by Twitter users and I am glad we are fixing them.

The build failures are due to flaky tests, You may want to see what tests are failing and create jira issues for those if they don't exist already.

corneadoug:
Ci failure is unrelated, merging if there is no more discussions

PR #1352, opened 2016-08-22
[ZEPPELIN-1327] Fix bug in z.show for Python interpreter

bustios:
### What is this PR for?
Currently, height parameter for z.show implementation to display PNG images in Python interpreter is not working. This PR fix that bug.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1327](https://issues.apache.org/jira/browse/ZEPPELIN-1327)

```python

x = [1,2,3,4,5]
y = [6,7,8,9,0]

plt.plot(x, y, marker="o")
z.show(plt, height="200px")
plt.close()
```

### Questions:

### DIFF
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -173,7 +173,9 @@ class PyZeppelinContext(object):
             p.savefig(img, format=fmt)
             img_str = b"data:image/png;base64,"
             img_str += base64.b64encode(img.getvalue().strip())
-            img_str = "<img src={img}>".format(img=img_str.decode("ascii"))
+            # Decoding img_str is necessary for Python 3 compability
+            img_str = img_str.decode("ascii")
+            img_str = "<img src={img}>".format(img=img_str)
         elif fmt == "svg":
             img = StringIO()
             p.savefig(img, format=fmt)
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -117,7 +117,6 @@ class PyZeppelinContext(object):
     
     def __init__(self):
         self.max_result = 1000
-        self.py3 = bool(sys.version_info >= (3,))
     
     def input(self, name, defaultValue=""):
         print(self.errorMsg)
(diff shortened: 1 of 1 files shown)


bustios:
The [CI failure](https://s3.amazonaws.com/archive.travis-ci.org/jobs/154284590/log.txt) seems not to be related to this change.
```
INFO  [22:02:24,946][] org.apache.zeppelin.cassandra.CassandraInterpreter@:<clinit> Bootstrapping Cassandra Interpreter 
ERROR [22:02:24,950][] org.apache.zeppelin.interpreter.Interpreter@:register Static initialization is deprecated. You should change it to use interpreter-setting.json in your jar or interpreter/{interpreter}/interpreter-setting.json 
```

```
[INFO] Running 'npm install --color=false' in /home/travis/build/apache/zeppelin/zeppelin-web
[ERROR] npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
[ERROR] npm WARN deprecated graceful-fs@2.0.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
[ERROR] npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
[ERROR] npm WARN deprecated minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
[ERROR] npm WARN deprecated lodash@1.0.2: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^4.0.0.
[ERROR] npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
[ERROR] npm WARN deprecated tough-cookie@2.2.2: ReDoS vulnerability parsing Set-Cookie https://nodesecurity.io/advisories/130
[ERROR] npm WARN deprecated CSSselect@0.7.0: the module is now available as 'css-select'
[ERROR] npm WARN deprecated CSSwhat@0.4.7: the module is now available as 'css-what'
[ERROR] npm WARN optional dep failed, continuing fsevents@1.0.14
```
```
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.25:npm (npm install) on project zeppelin-web: Failed to run task: 'npm install --color=false' failed. (error code 1) -> [Help 1]
```

bzz:
@bustios thank you for the fix and followup on CI!

Could you please also briefly explain, why Python3 compatibility facilities are not required here any more?

bustios:
Yes @bzz, I just deleted `if self.py3:` but not `img_str = img_str.decode('ascii')` (it is in line 173), which means it will be executed for both Python 2 and 3. Decoding bytes to string it necessary for Python 3 because that version makes distinction between String and Bytes, which does not occur in Python 2.

Maybe, `img_str = img_str.decode('ascii')` should be in a separate line of code to be clear.

bzz:
Got it, makes perfect sense! 

Do you think we should add a comment in there, explaining that compatibility is the reason for such decoding? I feel like this might help future contributors to see how to deal with such compatibility issue.

Other then that - it looks great to me, thanks again for fixing! It shall be ready to merge to master right after that, if there is no further discussion.

bustios:
Done. I added a comment and put the decoding statement in a separate line for a better understanding.

agoodm:
@bustios 
Nice job on removing the need to perform the python version check completely. This certainly helps make the code look cleaner.

However I cannot get your changes to work correctly with PNG images. Instead of resizing the image itself, it resizes the background containing the image:

ex

If I use width instead of height in this example, the image actually gets resized:

ex

bustios:
Thank you @agoodm. I committed a fix, could you try again please?

agoodm:
@bustios Nice job, this works great now. 👍

bzz:
Looks great to me! Merging if there is no further discussion

felixcheung:
LGTM

PR #1360, opened 2016-08-25
[ZEPPELIN-1366] Removed legacy JDBC alias

jongyoul:
### What is this PR for?
Removing old JDBC sugar

### What type of PR is it?
[Feature]

### Todos
* [x] - Removed codes to check jdbc alias enabled

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1366

* No longer available
  * `%jdbc(mysql)` -> `%mysql`

### Questions:
* Does this needs documentation? Yes, but I don't know where the proper location is.


bzz:
LGTM.

> Is there breaking changes for older versions? Yes

Do we have a change-log or something where we can keep track of breaking changes to highlight them in the release?

>Does this needs documentation? Yes, but I don't know where the proper location is.

@minahlee @AhyoungRyu what do you think is the best location for such docs?

Leemoonsoo:
I think it changes way of configuration. From single jdbc interpreter setting manages multiple jdbc connection, to individual interpreter setting manage single jdbc connection.

How about remove non-default jdbc connection configurations from default properties and documentation?
https://github.com/apache/zeppelin/blob/master/jdbc/src/main/resources/interpreter-setting.json
https://github.com/apache/zeppelin/blob/master/docs/interpreter/jdbc.md

And incompatible changes can be guided in https://github.com/apache/zeppelin/blob/master/docs/install/upgrade.md

jongyoul:
@bzz @Leemoonsoo Thanks, I don't know the proper location to describe the changes actually. I'll also update upgrade.md. And about changing a behaviour of JDBC, in case of `%jdbc(mysql)`, I thought it's not related to this issue directly, then I'll make another PR for handling that. We don't need to multiple configurations due to interpreter aliases.

AhyoungRyu:
@jongyoul Then can we use JDBC interpreter as described in [How to use](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/interpreter/jdbc.html#how-to-use) section like before? I just wondered :D

jongyoul:
@AhyoungRyu Sure, That works for now. If I'll remove those docs after I remove the prefix feature.

jongyoul:
Finished rebase. Ci failure is unrelated. Merging if there's no more discussion.

docs/install/upgrade.md
```diff
@@ -43,4 +43,8 @@ So, copying `notebook` and `conf` directory should be enough.
 
    ```
    bin/zeppelin-daemon.sh start
-   ```
\ No newline at end of file
+   ```
+
+# Breaking Changes
+## Upgrading from 0.6.x to 0.7.0
+- Mapping from `%jdbc(prefix)` to `%prefix` is no longer available. You can set multiple interpreters with same interpreter group.
```
Leemoonsoo: nit: How about "Instead, you can use `%[interpreter alias]` with multiple interpreter setttings on GUI", instead of "You can set multiple interpreters with same interpreter group"?

jongyoul: Yeah, that looks more kind

jongyoul:
Addressed upgrade.md

Leemoonsoo:
Thanks LGTM

jongyoul:
I've rebased it because of resolving conflict. I'll merge it soon.

PR #1361, opened 2016-08-25
[ZEPPELIN-1354] [WIP] Inject Services

echarles:
### What is this PR for?

Services like authentication, authorization, security filters... should be injected rather than manually instanciated.

This will allow to give central + finer control and abstracting the functionaly in interface and allow third parties to implement their own functions in with configurable injection points.

For this, I propose to use Guice as injection framework.

### What type of PR is it?
Improvement

### Todos
* [ ] - Documentation
* [ ] - Tests

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1354

Just compile and run

NA

### Questions:


zeppelin-web/pom.xml
```diff
@@ -127,7 +127,7 @@
               <goal>grunt</goal>
             </goals>
             <configuration>
-              <arguments>build</arguments>
+              <arguments>build --force</arguments>
```
jongyoul: Is this related to this PR?

echarles: nope, bad commit, I will remove this at next push.

pom.xml
```diff
@@ -88,7 +88,9 @@
     <slf4j.version>1.7.10</slf4j.version>
     <log4j.version>1.2.17</log4j.version>
     <libthrift.version>0.9.2</libthrift.version>
+    <governator.version>1.14.2</governator.version>
```
anthonycorbacho: Seems like governator is not used (all import are commented), I think it better to remove it instead of having bunch of commented code.

zeppelin-server/src/main/java/org/apache/zeppelin/inject/ZeppelinModule.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public ZeppelinModule(ZeppelinConfiguration conf) {
+    this.conf = conf;
+  }
+
+  @Override
+  public void configure(Binder binder) {
+
+    binder.bind(WebSecurity.class)
+        .to((Class<WebSecurity>) conf.getZeppelinWebSecurityClassname());
```
anthonycorbacho: Shall we add `Singleton scope` here? what do you think?

zeppelin-server/src/main/java/org/apache/zeppelin/web/DefaultWebSecurity.java
```diff
@@ -0,0 +1,62 @@
...
+  public void addSecurityFilter(WebAppContext webApp) {
+
+    webApp.setInitParameter("shiroConfigLocations",
+            new File(conf.getShiroPath()).toURI().toString());
+
+    SecurityUtils.initSecurityManager(conf.getShiroPath());
+    webApp.addFilter(org.apache.shiro.web.servlet.ShiroFilter.class, "/api/*",
+            EnumSet.allOf(DispatcherType.class));
+
+    webApp.addEventListener(new org.apache.shiro.web.env.EnvironmentLoaderListener());
```
anthonycorbacho: Wouldnt be cleaner to import `org.apache.shiro.web.env` and change this line with
`webApp.addEventListener(new EnvironmentLoaderListener());` ?

PR #1363, opened 2016-08-25
[ZEPPELIN-1371]add text/numeric conversion support to table display

Peilin-Yang:
### What is this PR for?
We people in Twitter have great demand of adding a flexible drop down menu for the columns in the tables which enables the text/numeric conversion.
This is because people want some columns to be of string type which, for example, fits to the underlying DB definition.
The use cases of the change include:

1. For now sorting on the columns is always lexicographically because Zeppelin front end treats the data as strings. It the values in the table can be convertible between string and number then we can sort the column by numeric values. I have filed another ticket for this: https://issues.apache.org/jira/browse/ZEPPELIN-1372
2. Since the users know more about their data than us it would be nice if we let the users decide what is the real type of their data. It is annoying if user wants the column to be strings but the front end forcefully inserts commas in it. In some scenarios, users may also want to copy/paste the table to somewhere else. If people want to remove the commas before other actions then that will be a nightmare.... 

### What type of PR is it?
[Improvement | Feature]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1371

1. Click on the dropdown menu would convert the text/number of that column.
2. Other functionalities esp. the sorting function should not be affected.

image
image

### Questions:
* Is there breaking changes for older versions? Probably a noticeable change in UI


corneadoug:
@Peilin-Yang Thanks for the contribution.
Small tip: Instead of hosting your screenshot, you can drop the image directly in the input of the PR description and the image will be hosted by github and shown.

Could you add more information on what is the benefit of changing the text/numeric conversion?
There is no example on what is happening when you actually switch the type and why it is useful.

Peilin-Yang:
@corneadoug thanks for your advice!
I updated the contents of PR so that it includes the use cases of this change.
Please let me know whether they make any sense.

bzz:
So cool - this should solve sorting issue for numeric columns! Thank you @Peilin-Yang!

One question - do you guys think it could be possible to detect numeric columns automatically using some heuristics and apply proper renderer on table creation time, rather than make user choose?

Peilin-Yang:
@bzz Actually I am currently look into that and I think it is possible because numbers are automatically detected and that is why the commas are inserted.
But I think even we could have the initial sorting for numbers it is still nice to let the users change it on the fly.

bzz:
Go it! It would be really nice to have such defaults for numbers, that make them sort numerically.

As for letting user change it - it sounds good, am it looks like the main use case for it is basically, removing commas?

Peilin-Yang:
Well, consider it alone, yes it is.

Peilin-Yang:
Added another PR #1371 that may have impact on this one.
i.e. if you guys think that one is reasonable then this one should be adjusted so that when user changes the datatype of the column the sorting order is changing at the same time.

bzz:
Let's merge #1371 first, and then it makes perfect sense to adjust this one to match user expectations of sorting changes according to column type (so it lets user override auto-detection). 

@corneadoug what do you think?

Peilin-Yang:
@bzz I agree

corneadoug:
Let's try to rebase this one, and review it

Peilin-Yang:
I made several changes for this commit:

1. put the style from notebook.css to paragraph.css
2. do not make any assumption/auto-detection of the original type of the data. Basically I drop the changes made in #1371 
3.  now the data type is column based. I added the validators for numbers and dates. Cells that fail to be converted to the required type will be shown with red background.

Peilin-Yang:
@corneadoug can we get this reviewed?

corneadoug:
@Peilin-Yang Thanks for the ping, I will look into that.
Any specific reason to drop #1371?

Peilin-Yang:
@corneadoug I just found that Handsontable has its intrinsic data type detection/conversion which we can rely on for the type conversion and is thus safer. More details at http://docs.handsontable.com/0.28.0/Options.html#type 
I think it is more convenient to this and make the data type on column basis so we do not worry about each individual cell.
With the dropdown menu we provide users are more aware of the fact that the data type is based on column, not the cell.
Does this make any sense?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -966,6 +966,117 @@ angular.module('zeppelinWebApp').controller('ParagraphCtrl', function($scope, $r
   };
 
   var setTable = function(data, refresh) {
+    function addButtonMenuEvent(button, menu) {
```
corneadoug: Do you think we could separate this function and all the sub-functions of that feature inside its own service?

corneadoug:
@Peilin-Yang I tested this PR and found one bug.
* Change a column that wouldn't order properly to Numerical type (balance in bank table for example)
* Order that column in descending order
* Switch the graph type (let's say bar chart)
* Switch Back to Table Viz
* The previous column still has the Numerical type, however if you try ordering again, it won't order properly

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
Peilin-Yang: I am not sure about this.
The feature is closely related to the HandsonTable that's why I included it inside the setTable function.
It seems that the logic is hard to de-coupled from the table itself.
Anyway, I can try to de-couple if you insist.

Peilin-Yang:
@corneadoug I have fixed the bug you mentioned. Testing was done and it seems ok.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
corneadoug: @Peilin-Yang Right now, it seems there is a lot of functions that are not directly related to the controller in itself, and targeting mainly handsontable. (buildMenu, buildButton, isNumeric etc...)

So Having a service dedicated to Handsontable with all the utility functions we use, would make the pragraph controller way lighter.

I know there is pretty much everything mix inside that controller, but let's try to separate a bit more the logic when possible.

I'll be happy to help if you want a simple example to show how it can be done, or I can also do a PR to this one with the whole service refactoring

corneadoug: @Peilin-Yang Sure, let me take care of that

Peilin-Yang: @corneadoug thanks. I will catch up later and learn how to better organize the code.

Peilin-Yang:
@corneadoug merged your PR.
Do you think now it is the good time to merge this one?

corneadoug:
@Peilin-Yang Yes, you will need to rebase first though

Peilin-Yang:
@corneadoug I have trouble with the rebase:( ... can you help a little bit?

corneadoug:
@Peilin-Yang Actually it's a bit hard for me to help since I can't rebase for you (Don't have the rights on your branch), but I can see that it is a pretty messy conflict you got.

For the Rebase process, you need to do a:
```
git fetch apache-github
git rebase apache-github/master
```
(apache-github being a remote for this official repo)

At the first round of conflicts on `paragraph.controller.js`, I would advise to copy paste the file from master, and replace your local file with conflict with it.
Then to copy paste the changes of your first commit into it (you can see the first commit diff [Here](https://github.com/apache/zeppelin/pull/1492/commits/b92e92ff8821a58c47d7510818ec518414499c83))

After that first round, the other conflicts should be easier

karuppayya:
@Peilin-Yang Very useful feature!!
let me know if i can help in taking this PR forward

PR #1364, opened 2016-08-25
[MINOR] Remove unnecessary question mark

zjffdu:
### What is this PR for?
Just remove the unnecessary quotation mark. 

### What type of PR is it?
[Documentation]

### Todos
* [ ] - Task

### What is the Jira issue?
* No jira created

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -184,7 +184,7 @@
 
 <property>
   <name>zeppelin.interpreter.group.order</name>
-  <value>"spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase</value>
+  <value>spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase</value>
   <description></description>
 </property>
 
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -184,7 +184,7 @@
 
 <property>
   <name>zeppelin.interpreter.group.order</name>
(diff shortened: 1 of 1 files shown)


zjffdu:
\cc @Leemoonsoo

Leemoonsoo:
Thanks @zjffdu for taking care.
LGTM and merge to master if there's no more discussion.

conf/zeppelin-site.xml.template
```diff
@@ -184,7 +184,7 @@
 
 <property>
   <name>zeppelin.interpreter.group.order</name>
-  <value>"spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase</value>
+  <value>spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase</value>
```
AhyoungRyu: @zjffdu Actually it's not a scope of this change, I just noticed `bigquery` is not included in here. Can we add `bigquery` using this PR?

prabhjyotsingh:
LGTM

zjffdu:
Thanks @AhyoungRyu , I have added `bigquery` in the following commit.
BTW, I don't this is a good idea to define the interpreter order in this way. Because it brings extra overhead when adding new interpreter. And I think it will break when we have multiple interpreter setting for the same interpreter (e.g. 2 interpreter setting for spark, one for spark1, another for spark2).

So I am thinking that maybe we can allow user to define the ordering in the interpreter page, the available interpreter can be fetched through rest api. So that we won't miss any interpreter when new interpreter is added.  \cc @Leemoonsoo @jongyoul

Leemoonsoo:
Yes, ordering from the interpreter page would be better.

PR #1369, opened 2016-08-26
[ZEPPELIN-1376] Add proxy credentials for dependency repo for corporate firewall use-cases

doanduyhai:
### What is this PR for?
When using Zeppelin behind corporate firewall, sometimes the dependencies download just fails silently. This PR has 2 objectives:

* add proxy credentials information for dependencies repo
* raise clear error message in case of dependencies download failure

There are 3 commits.

The first one add extra inputs in the form for adding new repository

add_repo

The second commit fixes some issues and display a clear and explicit error message when download of dependencies fail.

Before that, when the download fails, we can see the below behaviour

irrelevant_double_error_message

* the error message is displayed twice because the call twice the method `checkDownloadingDependencies();`. One in the success callback of:

```javascript
 $scope.updateInterpreterSetting = function(form, settingId) {
              ...
            $http.put(baseUrlSrv.getRestApiBase() + '/interpreter/setting/' + settingId, request)
              .success(function(data, status, headers, config) {
                $scope.interpreterSettings[index] = data.body;
                removeTMPSettings(index);
                thisConfirm.close();
                checkDownloadingDependencies();
                $route.reload();
              })
              .error(function(data, status, headers, config) {
             ...
    };
```

Another call is inside success callback of `getInterpreterSettings()`

```javascript
var getInterpreterSettings = function() {
      $http.get(baseUrlSrv.getRestApiBase() + '/interpreter/setting')
      .success(function(data, status, headers, config) {
        $scope.interpreterSettings = data.body;
        checkDownloadingDependencies();
      }).error(function(data, status, headers, config) {
      ....
```

The problem is that `$route.reload();` in the success callback of `updateInterpreterSetting()` will trigger `init()` then `getInterpreterSettings()` so `checkDownloadingDependencies()` is called twice.

I remove the call to `checkDownloadingDependencies()` from success callback of `updateInterpreterSetting()` 

The second modification is on class `DependencyResolver`. In the screen capture above, we get a **cryptic** NullPointerException coming from `DefaultRepositorySystem`. I now catch this NPE to wrap it into a more sensible and clearer exception:

```java

  public List<ArtifactResult> getArtifactsWithDep(String dependency,
    Collection<String> excludes) throws RepositoryException {
    Artifact artifact = new DefaultArtifact(dependency);
    DependencyFilter classpathFilter = DependencyFilterUtils.classpathFilter(JavaScopes.COMPILE);
    PatternExclusionsDependencyFilter exclusionFilter =
            new PatternExclusionsDependencyFilter(excludes);

    CollectRequest collectRequest = new CollectRequest();
    collectRequest.setRoot(new Dependency(artifact, JavaScopes.COMPILE));

    synchronized (repos) {
      for (RemoteRepository repo : repos) {
        collectRequest.addRepository(repo);
      }
    }
    DependencyRequest dependencyRequest = new DependencyRequest(collectRequest,
            DependencyFilterUtils.andFilter(exclusionFilter, classpathFilter));
    
 //Catch NPE thrown by aether and give a proper error message 
    try {
      return system.resolveDependencies(session, dependencyRequest).getArtifactResults();
    } catch (NullPointerException ex) {
      throw new RepositoryException(String.format("Cannot fetch dependencies for %s", dependency));
    }
  }
```

The result is much more cleaner

dependencies_download_error_popup

The last commit is just doc update

updated_docs

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Code Review
* [ ] - Simple test with no Internet connection
* [ ] - Test within a corporate firewall env with a third-party dependency, requiring download

### What is the Jira issue?
**[ZEPPELIN-1376]**

##### Simple test
* `git fetch origin pull/1369/head:WebProxy`
* `git checkout WebProxy`
* disconnect from the Internet (pull out the cable, shutdown wifi ...)
* add a random dependency to the Spark interpreter (take `info.archinnov:achilles-core:4.2.2` for example)
* validate the change, you should see an error popup on the top-right corner saying that Zeppelin cannot download the dependency

##### Corporate firewall test
* follow the steps above for simple test
* create a new repository (see how to **[here]**) and set the proxy information
* retry the steps above to ensure that the download is successful

 
See above

### Questions:
* Does the licenses files need update? --> **NO**
* Is there breaking changes for older versions? --> **NO**
* Does this needs documentation?  --> **YES, DONE**

[ZEPPELIN-1376]: https://issues.apache.org/jira/browse/ZEPPELIN-1376
[here]: http://localhost:4000/manual/dependencymanagement.html


felixcheung:
LGTM - how could we have automated tests for this?

doanduyhai:
@felixcheung 

That's the main issue I'm facing now, how to **automate** test with corporate firewall ... I'm going to ask someone to test the this PR at his place (he has a corporate firewall)

But apart from manual testing the firewall proxy, I don't see an **easy** way to create integration test for this feature, unless we build a Docker image of Zeppelin with a Docker image of a HTTP Proxy server ...

jongyoul:
@doanduyhai Could you please specify the title of this PR?

doanduyhai:
@jongyoul Sorry, it was a bad copy/paste between Github & JIRA, fixed now

jongyoul:
@astroshim @minahlee Can you review this PR?

astroshim:
About the second commits, Is it intended behavior that the error message bar is stacked whenever save the other interpreter setting?

image

doanduyhai:
@astroshim 

It is explained above:

> The problem is that $route.reload(); in the success callback of updateInterpreterSetting() will trigger init() then getInterpreterSettings() so checkDownloadingDependencies() is called twice.

Every time you save an interpreter setting, the JS script calls `$route.reload()` so function `checkDownloadingDependencies()` is called again, creating a new error popup.

In a normal workflow, the user will just remove the popup by clicking on the `x` icon.

As an enhancement, I propose:

1) set the property `combineDuplications ` from ngToast (http://tamerayd.in/ngToast/) to true
2) also, we can set a timeout so that the error popup disappear after a while

What do you think ?

astroshim:
Wow very quick response!!: 
I think set `combineDuplications` true is better but I don't have much talent in UI.:)

doanduyhai:
Ok let me push a new commit with the `combineDuplications = true`

doanduyhai:
@astroshim Done, now the popup error is only displayed once even if you update many other interpreters.

astroshim:
@doanduyhai Good. I am going to check first commit(proxy) soon.

astroshim:
@doanduyhai BTW how come only spark interpreter shows `error message bar`?

astroshim:
Proxy repo tested and it works properly. :+1:

doanduyhai:
@astroshim 

> BTW how come only spark interpreter shows error message bar?

The error popup applies to all interpreters, try to add a fictive dependency to `angular` or `shell` interpreter and you'll see also the error popup. The error message mentions clearly the `<group>.<interpreter>` that has the dependency issue

astroshim:
@doanduyhai Could you check the infinitely downloading library problem(like following screen shot)? 
Strangely this problem happens all interpreter except spark. Is this problem happen only to me? 

w

AhyoungRyu:
I also checked the interpreter setting page like @astroshim  did. 
(edit interpreter -> add `one:two:1.1.1` to the dependency table -> save ) 

Two things are different with master branch. 
1. Only Spark interpreter throw the `ng-toast` error message.
2. The other interpreters' status circle is looping infinitely. It need to be reload to turn to red and also don't show the `ng-toast` error message.

doanduyhai:
Let me investigate the root cause, it should be some Javascript subtle bug

doanduyhai:
OKKKK I've found the issue, it's tricky. 

I have added a non-existing dependency to the angular interpreter and I can also see a spinning circle forever. With Web Dev Console, I can see the return JSON result to the call GET /interpreter on the REST API. Below is the JSON result for Angular:


```json

...
    "status":"DOWNLOADING_DEPENDENCIES",
    "errorReason":"Could not find artifact one:two:jar:1.1.1 in central (http://repo1.maven.org/maven2/)",
...
```

So the status is wrong, it should be "ERROR" instead of "DOWNLOADING_DEPENDENCIES". Let me debug the server code

doanduyhai:
@astroshim @AhyoungRyu it tooks me hours but I have found the root cause of the issue and it's pretty ugly. Ready to hear ?

1) First issue is we **do not** set the InterpreterSetting `status` and `errorReason` together. For example here: https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java#L398

```java
    ...
    setting.setStatus(InterpreterSetting.Status.DOWNLOADING_DEPENDENCIES);
    interpreterSettings.put(setting.getId(), setting);
    ...
``` 

 So here we set the status to `DOWNLOADING_DEPENDENCIES` but the `errorReason` stills contains the previous value, possible something like "Cannot download dependency x.y.z ....". That explains why we see 

```
...
    "status":"DOWNLOADING_DEPENDENCIES",
    "errorReason":"Could not find artifact one:two:jar:1.1.1 in central (http://repo1.maven.org/maven2/)",
...
```

2) Second issue is more serious, it is concurrency issue. We set the status of the dependency download in a separate thread: https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java#L400-L433

```java
...
catch (Exception e) {
            logger.error(String.format("Error while downloading repos for interpreter group : %s," +
                    " go to interpreter setting page click on edit and save it again to make " +
                    "this interpreter work properly.",
                setting.getGroup()), e);
            setting.setErrorReason(e.getLocalizedMessage());
            setting.setStatus(InterpreterSetting.Status.ERROR);
...
```

The problem is that the initial status is set to `DOWNLOADING_DEPENDENCIES` and then the interpreter status is returned by the caller class to the client:  https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java#L113-L135

I added extra log message to debug:

```
            ...
            logger.info("************** DEBUG_ZEPPELIN setting status ERROR");
            setting.setStatus(InterpreterSetting.Status.ERROR);
            setting.setErrorReason(e.getLocalizedMessage());
            ...

    ...
    logger.info("************** DEBUG_ZEPPELIN setting debug " + setting.toDebugString());
    return new JsonResponse<>(Status.OK, "", setting).build();
    ...
```

 And below is what is displayed in the logs:

```
INFO [2016-09-07 12:35:26,151] ({qtp240166646-20} InterpreterRestApi.java[updateSetting]:132) - ************** DEBUG_ZEPPELIN setting debug DOWNLOADING_DEPENDENCIES :
ERROR [2016-09-07 12:35:26,539] ({Thread-25} InterpreterFactory.java[run]:433) - Error while downloading repos for interpreter group : angular, go to interpreter setting page click on edit and save it again to make this interpreter work properly. : Could not find artifact one:two:jar:1.1.25 in central (http://repo1.maven.org/maven2/)
org.sonatype.aether.resolution.DependencyResolutionException: Could not find artifact one:two:jar:1.1.25 in central (http://repo1.maven.org/maven2/)
        at org.sonatype.aether.impl.internal.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:375)
        at org.apache.zeppelin.dep.DependencyResolver.getArtifactsWithDep(DependencyResolver.java:164)
        at org.apache.zeppelin.dep.DependencyResolver.loadFromMvn(DependencyResolver.java:116)
        at org.apache.zeppelin.dep.DependencyResolver.load(DependencyResolver.java:79)
        at org.apache.zeppelin.dep.DependencyResolver.load(DependencyResolver.java:96)
        at org.apache.zeppelin.dep.DependencyResolver.load(DependencyResolver.java:88)
        at org.apache.zeppelin.interpreter.InterpreterFactory$3.run(InterpreterFactory.java:425)
Caused by: org.sonatype.aether.resolution.ArtifactResolutionException: Could not find artifact one:two:jar:1.1.25 in central (http://repo1.maven.org/maven2/)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:537)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:216)
        at org.sonatype.aether.impl.internal.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:358)
        ... 6 more
Caused by: org.sonatype.aether.transfer.ArtifactNotFoundException: Could not find artifact one:two:jar:1.1.25 in central (http://repo1.maven.org/maven2/)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$4.wrap(WagonRepositoryConnector.java:971)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$4.wrap(WagonRepositoryConnector.java:966)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$GetTask.flush(WagonRepositoryConnector.java:707)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector$GetTask.flush(WagonRepositoryConnector.java:701)
        at org.sonatype.aether.connector.wagon.WagonRepositoryConnector.get(WagonRepositoryConnector.java:452)
        at org.sonatype.aether.impl.internal.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:456)
        ... 8 more
 INFO [2016-09-07 12:35:26,541] ({Thread-25} InterpreterFactory.java[run]:437) - ************** DEBUG_ZEPPELIN setting status ERROR
```

So the `DOWNLOADING_DEPENDENCIES` status is returned to the client **before** the download error status is set in a **separate** thread.

This issue is indeed already there in the master and it **not** related to this pull request

doanduyhai:
@astroshim @AhyoungRyu 

Ok I've found the issue, it's my bad:

```javascript
 for (var setting = 0; setting < $scope.interpreterSettings.length; setting++) {
        var setting = $scope.interpreterSettings[setting];
        if (setting.status === 'DOWNLOADING_DEPENDENCIES') {
          isDownloading = true;
        } else if (setting.status === 'ERROR') {
          ngToast.danger({content: 'Error setting properties for interpreter \'' +
            setting.group + '.' + setting.name + '\': ' + setting.errorReason,
            verticalPosition: 'top', dismissOnTimeout: false});
        }
      }
```

I'm using the variable `setting` as an index for the iteration of the for loop as well as the variable for the setting object itself, which breaks awfully the for loop.

Now it's fixed, see below

fixed_errorpopup

doanduyhai:
@AhyoungRyu @astroshim Is the fix ok for you now ? The build is green

astroshim:
@doanduyhai Sorry to keep you waiting. 
It works like a charm now but the error count number at the beginning of message appears a little strange like following screen shot.
st

doanduyhai:
@astroshim 

The number you see in front of each error popup is normal, it comes from the fact that now we set `combineDuplications  = true` for **ngToast**. The number corresponds to the time the error popup is duplicated

astroshim:
@doanduyhai I know that number means but as you can see the screenshot i attached, the number seems doesn't match with count.

doanduyhai:
@astroshim 

> the number seems doesn't match with count

What do you mean by this ? Which number did you expect to see ?

On the animated screencapture, the count started with 1 then 2 for both popups.

Then you close one so the count is reset to 0 but the 2<sup>nd</sup> popup is still there and continue counting to 4.

Do not forget that the `checkDownloadingDependencies()` method is called multiple times:

 - everytime an interpreter setting is updated (2 times indeed)
 - everytime the status returned by the server is **DOWNLOADING_DEPENDENCIES**

astroshim:
@doanduyhai Could you take a look at the following screenshot?
After the `alluxio` and `angular` dependency error occurs, I just saved `alluxio` interpreter setting several times but the `angular` error count is more then `alluxio`. 
I thought the error count would be same.
s

and Is it possible to close the pop-up when user move to a different page?

AhyoungRyu:
@doanduyhai Sorry for my late response and thanks for the kind reminder.
I tested again and checked the below issues I mentioned are gone :D
>(edit interpreter -> add `one:two:1.1.1` to the dependency table -> save )
Two things are different with master branch. 
1. Only Spark interpreter throw the ng-toast error message.
2. The other interpreters' status circle is looping infinitely. It need to be reload to turn to red and also don't show the ng-toast error message. 

But I found another issues in the latest update. 

1. edit Alluxio interpreter -> add `aaa:bbb:1.1.1`  to the dependency field ->  go to another page like "Credential" or "Configuration" -> 
The `ng-toast` error message is still shown as @astroshim said in the above comment. How about adding some timeout property to the message?

2. edit Alluxio -> add `aaa:111:1.1.1` & `bbb:111.2.2.2` to the dependency field -> save 
I think the number of errors and reasons need to be same. Currently, while the number of errors keeps increasing whenever I save the change, the only one reason "Could not find artifact aaa:111:jar:1.1.1 in central (http://repo1.maven.org/maven2/)" is shown. Need to show "Could not find artifact bbb:111.2.2.2 in central (http://repo1.maven.org/maven2/)" as well if we want to show the number of errors. 

screen shot 2016-09-18 at 9 37 55 pm

The number of error should be 2 in the above case I guess. But as you can see the number is 1 since `checkDownloadingDependencies()` is called only one time.

doanduyhai:
@AhyoungRyu 

> The ng-toast error message is still shown as @astroshim said in the above comment. How about adding some timeout property to the message?

I knew that there is a timeout option for `ng-toast` but I chose not use it. The simple reason is that I consider the dependency issue is a **serious** issue so we want the user to **explicitly/manually** close the error popup and acknowledge that he is aware about the download issue.

What is your opinion about that ? Adding a timeout to automatically close the error popup is very easy.

>  Need to show "Could not find artifact bbb:111.2.2.2 in central (http://repo1.maven.org/maven2/)" as well if we want to show the number of errors.

There is a simple explanation for this. `sonatye.aether` library tries to download the dependencies **one after another** and raise an exception as soon as it cannot find/download a dependency. In your example, we'll see the error popup for the 2<sup>nd</sup> dependency if the download of the 1<sup>st</sup> is successful

```java
 synchronized (repos) {
      for (RemoteRepository repo : repos) {
        collectRequest.addRepository(repo);
      }
    }
    DependencyRequest dependencyRequest = new DependencyRequest(collectRequest,
            DependencyFilterUtils.andFilter(exclusionFilter, classpathFilter));
    try {
      return system.resolveDependencies(session, dependencyRequest).getArtifactResults();
    } catch (NullPointerException ex) {
      throw new RepositoryException(String.format("Cannot fetch dependencies for %s", dependency));
    }
```

We cannot change the behavior of `sonatype.aether`. We can change our source code to download dependency **one by one** but it is very slow and require changing the interface of a lot of methods

doanduyhai:
@astroshim @AhyoungRyu 

The wrong counting on error popup is now fixed. Every time you 

* reload the interpreter page
* update an interpreter setting

all the error popup are removed and the Zeppelin re-display error popup only when necessary

Also, when you navigate to different pages, the error popups are removed.

The latest commit reset the ngToast every time you "enter" a page. With this change we don't need to set a timeout on the error popup anymore

doanduyhai:
@astroshim @AhyoungRyu 

Please see the final result, it should make you both happy :)

fixed error popups

doanduyhai:
@AhyoungRyu @astroshim 

Please tell me if it's ok so that I can merge. We need to merge this PR before I can move on the refactoring task (of changing `synchronized` blocks by appropriate datastructures)

AhyoungRyu:
@doanduyhai So sorry for my late response ㅠ_ㅠ 
Tested again and checked
>when you navigate to different pages, the error popups are removed.

the above issue is gone! 👍 

However, so sadly another issue is emerged. After removing the wrong dependency, the error message should be gone. But I think it's not. Here is the procedure how you can reproduce this.

  1. edit `Alluxio` interpreter dependency -> type `a:b:1.1.1` and save -> surely the ng-toast error message is shown up
  2. edit `Angular` interpreter dependency -> type `a:b:1.1.1` and save -> ng-toast error message is also shown up for this interpreter
  3. remove those wrong dependencies from both `Alluxio` and `Angular` -> save -> the status circles for that two interpreters are turning to green
  4. but the ng-toast error messages are still shown up even if I reload the page multiple times

doanduyhai:
@AhyoungRyu 

I followed your procedure but it works perfectly for me, see below screenshot. Did you clean your browser cache when testing the new commit ? Maybe there is some old JS code left-over

popup ok

AhyoungRyu:
@doanduyhai Yeah your screenshot is saying so. Let me test it again. Thanks for your quick response!

AhyoungRyu:
@doanduyhai Checked again and it works well! After removing the wrong dependency, the error message doesn't show up anymore.

doanduyhai:
OK I'll merge if no more remarks

WegDamit:
Those changes are only for dependencies given in the interpreter definition, right?
Using a %spark.dep section in a notebook seems to uses a SparkDependencyContext which does not inherit from AbstractDependency.* and so does not use the proxy definition at all. Right?
Wouldn't it be useful to give the %dep section access to the proxy too?

PR #1347, opened 2016-08-19
ZEPPELIN-1284. Unable to run paragraph with default interpreter

zjffdu:
### What is this PR for?
This issue happens when SPARK_HOME is not defined. In this case, you are using spark 2.0 and scala-2.10

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1284

### How should this be tested?
Run the following command, and then run the tutorial note in local mode
```
mvn package -DskipTests -Ppyspark -Psparkr -Pyarn -Phadoop-2.7 -Pspark-2.0
```

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #1373, opened 2016-08-28
[ZEPPELIN-1385] Avoid NPE when create SparkSession

lresende:
### What is this PR for?
Avoid NPE when create SparkSession is outputDir has not been set yet

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1385](https://issues.apache.org/jira/browse/ZEPPELIN-1385)


lresende:
After this fix is applied, have verified that can connect and execute a job on a remote Spark cluster.

Leemoonsoo:
Thanks @lresende for the fix and verification. But i think we already have exactly the same fix at #1347.

lresende:
@Leemoonsoo No worries, didn't see that...

PR #1374, opened 2016-08-28
[MINOR] Add new line before logging paragraph content

lresende:
### What is this PR for?
Add new line before logging paragraph content

### What type of PR is it?
[Improvement]

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
@@ -295,7 +295,9 @@ public class RemoteInterpreter extends Interpreter {
 
   @Override
   public InterpreterResult interpret(String st, InterpreterContext context) {
-    logger.debug("st: {}", st);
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/thrift/RemoteInterpreterService.java
```diff
@@ -6135,7 +6135,7 @@ public String toString() {
       }
       first = false;
       if (!first) sb.append(", ");
-      sb.append("st:");
+      sb.append("st:\n");
```
Leemoonsoo: Is it an intended change? because this file is an autogenerated file from RemoteInterpreterService.thrift

lresende: Good catch. I have updated the commit.

lresende:
@Leemoonsoo any other comments here ?

Leemoonsoo:
Thanks @lresende.
LGTM and merge if there're no more discussions!

PR #1375, opened 2016-08-29
[ZEPPELIN-1171] Only retrieve resources if client is set

lresende:
### What is this PR for?

Fix NullPointerException when trying to retrieve resources and the client is not set
### What type of PR is it?

[Bug Fix]
### What is the Jira issue?
- [ZEPPELIN-1171](https://issues.apache.org/jira/browse/ZEPPELIN-1171)


jongyoul:
I think it would be a potential error but do you have any experience where it occurs error on that line?

corneadoug:
@lresende @jongyoul Any news on this issue

PR #1378, opened 2016-08-29
[ZEPPELIN-1358] Add support to display Pandas DataFrame index using z.show()

bustios:
### What is this PR for?

Add support to display optionally Pandas DataFrame index using z.show(show_index=True) in python interpreter. By default, DataFrame index will not be displayed.
### What type of PR is it?

Improvement
### What is the Jira issue?

[ZEPPELIN-1358](https://issues.apache.org/jira/browse/ZEPPELIN-1358)

```
mvn -Dpython.test.exclude='' test -pl python -am
```

screenshot from 2016-10-09 18-18-44
### Questions:

### DIFF
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -142,17 +142,19 @@ class PyZeppelinContext(object):
         """
         limit = len(df) > self.max_result
         header_buf = StringIO("")
-        header_buf.write(str(df.columns[0]))
-        for col in df.columns[1:]:
+        idx_name = df.index.name if df.index.name is not None else ""
+        header_buf.write(idx_name)
+        for col in df.columns:
             header_buf.write("\t")
             header_buf.write(str(col))
         header_buf.write("\n")
         
         body_buf = StringIO("")
         rows = df.head(self.max_result).values if limit else df.values
-        for row in rows:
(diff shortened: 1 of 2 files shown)


bzz:
@bustios thank you for improvements! 

Could you please explain a bit the benefits of having print index behaviour enabled by default? Do you think it should not be optional? 

It's nice advanced feature but I also see the value in having default behaviour consistent between interpreters. What do you think?

bustios:
Sure @bzz, it can be optional. I just thought it was more appropriate, since a dataframe index is always printed, but I can add a parameter like `show_index=False`  to make it optional.

bzz:
I see, indeed we could change the default to always show dataframe index, if you think it's more consistent with Python behavior, but let's add a flag anyway, so later, in case we decided to switch back - it would be a trivial change.

Other then that - looks great to me, 👍  for tests!

bzz:
@bustios let me know if you still think it's possible to make it optional and I'll be happy to merge it.

bustios:
Sorry for the late response @bzz. Of course it's possible. I already made the changes to make it optional.

bzz:
Sorry for massive communication delay :\

Looks awesome to me, merging to master, if there is no further discussion!

PR #1382, opened 2016-08-29
[ZEPPELIN-1391][Interpreters] print error while existing registedInterpreter with the same key but different settings

WeichenXu123:
### What is this PR for?
print error while existing registedInterpreter with the same key but different settings.
In order to compare the property Map in the Interpreter easily,
I override the `equals` method in `InterpreterProperty` class. (and `hashCode` at the same time)

and I fix a small error in  `InterpreterProperty.toString()`, its result forgot `}` in the end. 

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1391

Existing tests.

### Questions:


WeichenXu123:
cc @jongyoul thanks!

jongyoul:
LGTM. Merging if there's no more discussion.

PR #1391, opened 2016-09-01
ZEPPELIN-1403. Should maintain the using port list

zjffdu:
### What is this PR for?

For now, in RemoteInterpreterManagedProcess we will first find one available port (By using ServerSocket) and then use this port in the interpreter process. But it is possible that 2 interpreters use the same port. Here's the scenarios.
1. Thread1 (Interpreter A) find available port 3000 
2. Thread2 (Interpreter B) find available port 3000 
3. Process for Interpreter A starts
4. Process for Interpreter B starts
### What type of PR is it?

[Bug Fix]
### Todos
- [ ] - Task
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-1403

No test added
### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterUtils.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterUtils.java
@@ -24,18 +24,27 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  *
  */
 public class RemoteInterpreterUtils {
   static Logger LOGGER = LoggerFactory.getLogger(RemoteInterpreterUtils.class);
-  public static int findRandomAvailablePortOnAllLocalInterfaces() throws IOException {
+  private static Set<Integer> portsUsing = new HashSet();
+
+  public static synchronized int findRandomAvailablePortOnAllLocalInterfaces() throws IOException {
     int port;
     try (ServerSocket socket = new ServerSocket(0);) {
       port = socket.getLocalPort();
       socket.close();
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterUtils.java
```diff
@@ -24,18 +24,27 @@
...
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  *
  */
 public class RemoteInterpreterUtils {
   static Logger LOGGER = LoggerFactory.getLogger(RemoteInterpreterUtils.class);
-  public static int findRandomAvailablePortOnAllLocalInterfaces() throws IOException {
+  private static Set<Integer> portsUsing = new HashSet();
```
bzz: How do you think, Is there another way to do that, rather than introducing a static state? 
I.e I have hard time imagining how this can be tested..

zjffdu: Sorry for late response @bzz , For the unit test, I am thinking that I can create 100 threads (increase the possibility of port conflict) and each thread will try to find a new avaliable port. And finally I can verify that size of `portsUsing` is 100 which means no port is used by 2 threads.  What do you think ?

bzz:
@zjffdu test plan sounds great. 

My only concern was - may we can find a better place for this state, rather than `static` ? Move it somewhere, where if does not require to be static, but can be an instance method\state

zjffdu:
@bzz Sorry for delay, I was in vacation in the last 2 weeks. Regarding move it to other places rather than static, I am thinking to use a PortProvider class to wrap all the logic and of course this class should be singleton. What do you think ?

bzz:
Having a `PortProvider` as injectable dependency, that incapsulates all the logic about picking the port numbers sounds great!

PR #1330, opened 2016-08-13
[ZEPPELIN-1144]Zeppelin home page should only list notebooks with read or write permission

mwkang:
### What is this PR for?
If logged in user does not have Read and Write permission for a notebook, user should not see the notebook in the zeppelin home page.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1144

### How should this be tested?
* unit test
* online test

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #1392, opened 2016-09-01
[ZEPPELIN-1437, 1438] Multi-user note management - user aware reload broadcast

khalidhuseynov:
### What is this PR for?
This PR addresses part of multi-user note management in Zeppelin. One of the tasks namely listing notes per user on Zeppelin start was addressed in #1330. However that PR didn't solve all problems, and reloading notes was incomplete as well as socket broadcast was not user aware [ZEPPELIN-1437](https://issues.apache.org/jira/browse/ZEPPELIN-1437), [ZEPPELIN-1438](https://issues.apache.org/jira/browse/ZEPPELIN-1438). This PR addresses those issue.

### What type of PR is it?
Improvement

### Todos
* [x] - list notes per user on reload
* [x] - broadcast per user (multicast)
* [x] - tests
* [x] - use authorization module to filter notes on sync
* [x] - broadcast on permissions change
* [ ] - discussion and review

### What is the Jira issue?
[Zeppelin-1437](https://issues.apache.org/jira/browse/ZEPPELIN-1437), [ZEPPELIN-1438](https://issues.apache.org/jira/browse/ZEPPELIN-1438)

1. Start Zeppelin
2. Login as user1, and user2 on different windows
3. Each user should be able to see their own note workbench
4. If note changed to private (readers, writers not empty), that note should disappear from others note workbench.

reload_broadcast

### Questions:


khalidhuseynov:
this is ready for review

khalidhuseynov:
rebased and CI is green

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
```diff
@@ -31,17 +35,22 @@
...
+  private static Map<String, Map<String, Set<String>>> authInfo = new HashMap<>();
+  private static ZeppelinConfiguration conf;
+  private static Gson gson;
+  private static String filePath;
+
+  private NotebookAuthorization() {}
 
-  public NotebookAuthorization(ZeppelinConfiguration conf) {
-    this.conf = conf;
+  public static NotebookAuthorization init(ZeppelinConfiguration config) {
```
echarles: Why moving from constructor to init method (factory builder)?
Especially, the conf object is static, so calling multiple time init will override the conf field. The state of the NotebookAuthorization will not be consistent.

khalidhuseynov: @echarles good point. actually that's more of a singleton pattern that gets instantiated once on `init` and can be retrieved by `getInstance` further on. the purpose of it was to enable other modules (e.g. storage in zengine) not having direct reference to notebookAuthorization, to be able to use its services without changing api. I believe for longer term authorization and storage modules can be refactored in a way that we don't need that pattern.

regarding overriding the conf field, you're right. I addressed it in 9cf1d8848ee02b5fafda1afd538b71befd3eb412 so that it gets instantiated only once.

khalidhuseynov:
i've addressed [zeppelin-1438](https://issues.apache.org/jira/browse/ZEPPELIN-1438) here as well in last commit a2ce26889117f364f9d2a5242f6c4daa1c25e265, so user workbench should be updating user lists without manual reload, ready for review

Leemoonsoo:
Tested and LGTM.

anthonycorbacho:
@Leemoonsoo is it possible to merge it in couple of hour? some of my work will depends on this PR :)

PR #1394, opened 2016-09-02
rename r directory to 2BWJFTXKJ

prabhjyotsingh:
### What is this PR for?
Rename "R" directory to "2BWJFTXKJ" under notebook directory to make it look like all other notebooks.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Rename

### What is the Jira issue?
* N/A

goto  $ZEPPELIN_HOME/notebook directory, under this there should be no "R" directory.
On starting zeppelin-server, the existing notebook should be listed.

N/A

### Questions:


prabhjyotsingh:
@echarles I'm assuming there was no specific reason to keep name of this directory as "R" and hence, renaming this to look like any other notebook.

echarles:
@prabhjyotsingh yes, assumption is correct. At the time of the R interpreter dev, I was used to manually rename the note folder and edit the json to set the id the same value.

This PR takes it back to the default layout (didn't test it, but checked that the id value in the json was also correct).

LGTM

prabhjyotsingh:
Sure, Thank you. Will merge this if no more discussion.

Leemoonsoo:
Shell we update https://github.com/apache/zeppelin/blob/master/notebook/r/note.json#L1011, too?

prabhjyotsingh:
Yes, I did both, rename the directory and https://github.com/prabhjyotsingh/zeppelin/blob/5993506b77be84d71ab730c2965572edd73e870e/notebook/2BWJFTXKJ/note.json#L1041

Leemoonsoo:
Ah I missed the change. LGTM

PR #1397, opened 2016-09-02
[ZEPPELIN-1406] modify to load method artifact local path in Helium application json

cloverhearts:
### What is this PR for?

Currently contained in the Zeppelin Helium application does not work.
The relative value of the configuration file path does incurring call normally.
Therefore, it sets the Helium directory you specified in the application default directory.
An example of the application was to modify the path of helium.
### What type of PR is it?

Bug Fix, Improvement
### Todos
- [x] added setter method for artifact in Helium Packakge class
- [x] if not absolute path for artifact then replace to absolute path for helium artifact local path
- [x] replace helium example artifact path for json
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1406

step 1. 
build

```
mvn clean package -Pbuild-distr -Pexamples -DskipTests
```

step 2.
go to web ui and create a new notebook.
write paragraph.
here.

```
%spark 
import java.util.Date
new Date();
```

step 3.
choice. 
helium application clock.

working well.
#### before

does not working.
noshow
#### after

clock
### Questions:

### DIFF
--- a/zeppelin-examples/zeppelin-example-clock/zeppelin-example-clock.json
+++ b/zeppelin-examples/zeppelin-example-clock/zeppelin-example-clock.json
@@ -18,7 +18,7 @@
   "type" : "APPLICATION",
   "name" : "zeppelin.clock",
   "description" : "Clock (example)",
-  "artifact" : "zeppelin-examples/zeppelin-example-clock/target/zeppelin-example-clock-0.7.0-SNAPSHOT.jar",
+  "artifact" : "../zeppelin-examples/zeppelin-example-clock/target/zeppelin-example-clock-0.7.0-SNAPSHOT.jar",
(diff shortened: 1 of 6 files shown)


AhyoungRyu:
@cloverhearts I can't reproduce this bug. Here is what I did.

###### 1. Build the latest `master` branch with below command and restart Zeppelin daemon
```
mvn clean package -Pbuild-distr -Pexamples -DskipTests
```

###### 2. Create a new notebook and run 
```
%spark

import java.util.Date
new Date();
```

###### 3. I can see the Helium icon and the clock without this patch
screen shot 2016-09-05 at 4 47 15 pm

Did I miss something?

cloverhearts:
@AhyoungRyu 
I'll check it.

PR #1251, opened 2016-07-30
[ZEPPELIN-1256][BUILD] Build distribution package with Spark 2.0 and Scala 2.11

minahlee:
### What is this PR for?
- build distribution package with Spark 2.0 and Scala 2.11
- change travis profile to use spark 2.0 instead of 2.0.0-preview by removing `-Dspark.version=2.0.0-preview` property
- Change profile activation rule 
  * current way of profile activation is not proper for supporting more than three scala version
  * -Pscala-2.11 activates both scala-2.10 and scala-2.11 profile without this patch

### What type of PR is it?
Build

### What is the Jira issue?
[ZEPPELIN-1256](https://issues.apache.org/jira/browse/ZEPPELIN-1256)

### How should this be tested?

**Before**
Run `mvn help:active-profiles -Pscala-2.11`:
   ```
Active Profiles for Project 'org.apache.zeppelin:zeppelin:pom:0.7.0-SNAPSHOT':

The following profiles are active:

 - scala-2.10 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)
 - scala-2.11 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)



Active Profiles for Project 'org.apache.zeppelin:zeppelin-interpreter:jar:0.7.0-SNAPSHOT':

The following profiles are active:

 - scala-2.10 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)
 - scala-2.11 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)



Active Profiles for Project 'org.apache.zeppelin:zeppelin-zengine:jar:0.7.0-SNAPSHOT':

The following profiles are active:

 - scala-2.10 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)
 - scala-2.11 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)



Active Profiles for Project 'org.apache.zeppelin:zeppelin-display_2.11:jar:0.7.0-SNAPSHOT':

The following profiles are active:

 - scala-2.11 (source: org.apache.zeppelin:zeppelin-display_2.11:0.7.0-SNAPSHOT)
 - scala-2.10 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)
 - scala-2.11 (source: org.apache.zeppelin:zeppelin:0.7.0-SNAPSHOT)



[...]

PR #1400, opened 2016-09-03
[ZEPPELIN-1407] Fix Scala 2.11 build

lresende:
### What is this PR for?
Avoid activating the Scala 2.10 profile when building for Scala 2.11

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* (https://issues.apache.org/jira/browse/ZEPPELIN-1407)[https://issues.apache.org/jira/browse/ZEPPELIN-1407]

Perform Scala 2.10 and 2.11 builds starting from a maven repository that does not have org.apache.zeppelin artifacts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -531,7 +531,7 @@
     <profile>
       <id>scala-2.10</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
+        <property><name>!scala-2.11</name></property>
       </activation>
       <properties>
         <scala.version>2.10.5</scala.version>
@@ -541,6 +541,9 @@
 
     <profile>
       <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
       <properties>
         <scala.version>2.11.7</scala.version>
         <scala.binary.version>2.11</scala.binary.version>
@@ -826,5 +829,6 @@
       </build>
     </profile>
   </profiles>
+
 
 </project>
--- a/zeppelin-display/pom.xml
+++ b/zeppelin-display/pom.xml
@@ -95,6 +95,9 @@
   <profiles>
     <profile>
       <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
       <dependencies>
         <dependency>
           <groupId>org.scala-lang.modules</groupId>
--- a/zeppelin-distribution/pom.xml
+++ b/zeppelin-distribution/pom.xml
@@ -114,6 +114,9 @@
   <profiles>
     <profile>
       <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
       <dependencyManagement>
         <dependencies>
           <dependency>
--- a/zeppelin-server/pom.xml
+++ b/zeppelin-server/pom.xml
@@ -448,6 +448,9 @@
   <profiles>
     <profile>
       <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
       <dependencyManagement>
         <dependencies>
           <dependency>


lresende:
@bzz @Leemoonsoo Please review, this is more like a hotfix for the build.

lresende:
This should go for both master and 0.6x release.

felixcheung:
build failed I think because of
```
[ERROR] Failed to execute goal on project zeppelin-spark-dependencies_2.10: Could not resolve dependencies for project org.apache.zeppelin:zeppelin-spark-dependencies_2.10:jar:0.7.0-SNAPSHOT: The following artifacts could not be resolved: org.apache.spark:spark-core_2.11:jar:1.1.1, org.apache.spark:spark-repl_2.11:jar:1.1.1, org.apache.spark:spark-sql_2.11:jar:1.1.1, org.apache.spark:spark-hive_2.11:jar:1.1.1, org.apache.spark:spark-streaming_2.11:jar:1.1.1, org.apache.spark:spark-catalyst_2.11:jar:1.1.1, org.spark-project.akka:akka-actor_2.11:jar:2.2.3-shaded-protobuf, org.spark-project.akka:akka-remote_2.11:jar:2.2.3-shaded-protobuf, org.spark-project.akka:akka-slf4j_2.11:jar:2.2.3-shaded-protobuf, org.spark-project.akka:akka-testkit_2.11:jar:2.2.3-shaded-protobuf, org.spark-project.akka:akka-zeromq_2.11:jar:2.2.3-shaded-protobuf: Could not find artifact org.apache.spark:spark-core_2.11:jar:1.1.1 in central (http://repo.maven.apache.org/maven2) -> [Help 1]
```

lresende:
@bzz Sorry, I had put the profiles activation as if scala 2.11 was the default, have fixed it now.
three builds still failing with timeout, I will try to force new builds

lresende:
@bzz the build errors are related to RAT issues
>>>

The build has been terminated
>>>

felixcheung:
CI green, any more comments?

lresende:
@felixcheung @Leemoonsoo @bzz Any more comments here ? Should we merge this to 0.6 branch and master, as it's kind blocking to 2.11

felixcheung:
Done. master and branch-0.6 - thanks!

Leemoonsoo:
I'm sorry my comment is bit late. code is already been merged.

But let my write it for the records

Because of all our documents and scripts uses `-Pscala-2.xx` instead of `-Dscala-2.xx`. And with `-Pscala-2.xx` we don't have any problem.

Build instruction
https://github.com/apache/zeppelin#-pscala-version-optional

travis ci script
https://github.com/apache/zeppelin/blob/master/.travis.yml#L43

release script
https://github.com/apache/zeppelin/blob/master/dev/create_release.sh#L106


If we want to use `-D` instead of `-P` to activate profile, that's different story.
I think using `-D` is actually more flexible than `-P` in maven, but while all other profiles are activated by `-P`, i'm not sure only activating scala profile by `-D` can be helpful and beneficial.

lresende:
@Leemoonsoo The issue wasn't about -P versus -D, but that the Scala 2.10 was with activation true by default, so even when you did only -Pscala2.11 the Scala 2.10 profile was being activated.

lresende:
BTW, the properties on the profile activation, only enables the "default"  to be activated by default, so mvn clean install will not have properties defined, so will start Scala 2.10...

Leemoonsoo:
@lresende You can see followings from http://maven.apache.org/guides/introduction/introduction-to-profiles.html.

```
All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.
```

So once -Pscala-2.11 is defined, `scala-2.10` should be automatically deactivated.

lresende:
But unfortunately this wasn't what maven was enforcing, and Scala 2.11 build was failing when -PScala2.11 was being used, and when invoking dependency:tree with -Pscala2.11 there were few dependencies that were being resolved to _2.10 version of the dependency and thus causing the build failure. This approach seems to not rely on that maven feature and make the build work on both scenarios.

Leemoonsoo:
@lresende In the [issue description](https://issues.apache.org/jira/browse/ZEPPELIN-1407), your build command used `-Dscala-2.11` when you get those errors.

I could reproduce the error when i use `-Dscala-2.11`, but if i use `-Pscala-2.11` i couldn't reproduce the problem.

felixcheung:
It looks like -D is in fact setting property and this change is adding such a check for the property?

minahlee:
As moon mentioned, if you use build instruction in https://github.com/apache/zeppelin#example it doesn't bring any issue.
One of the reason I changed activation rule in #1251 was that using -D won't be scalable for scala-2.12 support in the future.
@lresende if you are not strongly against to build with -P and if you can confirm that there is no issue to build using `-Pscala-2.11` not `-Dscala-2.11` in c88010fd0a744d026b75d02bc62bc6dc690f0f9a, I would like to revert this change.

lresende:
I have verified that with that by reverting my changes -PScala2.10/Scala2.11 have the proper behavior and my original issue was that indeed my build alias had -D for the profile activation. Sorry for the confusion, please go ahead and revert this change.

minahlee:
@lresende thank you for verification. I am reverting this PR then.

PR #1401, opened 2016-09-04
[ZEPPELIN-1409] Refactor RAT build on Travis.CI configuration

lresende:
### What is this PR for?
Create a specific build for checking license compliance with RAT
and avoid running these checks on every build that compose the
PR build. 
As for normal development builds, this follows  the same pattern 
used for maven tests, RAT is enabled to run by default, but now
there is support disabling it with -DskipRat. 
Travis CI will run RAT once, on the RAT build, and disable RAT 
checks on all other build profiles.

### What type of PR is it?
[Enhancement]

### What is the Jira issue?
* [ZEPPELIN-1409](https://issues.apache.org/jira/browse/ZEPPELIN-1409)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,39 +40,39 @@ matrix:
 
     # Test all modules with spark 2.0.0 and scala 2.11
     - jdk: "oraclejdk7"
-      env: SCALA_VER="2.11" SPARK_VER="2.0.0" HADOOP_VER="2.3" PROFILE="-Pspark-2.0 -Phadoop-2.3 -Ppyspark -Psparkr -Pscalding -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr" TEST_FLAG="verify -Pusing-packaged-distr" TEST_PROJECTS=""
+      env: SCALA_VER="2.11" SPARK_VER="2.0.0" HADOOP_VER="2.3" PROFILE="-Pspark-2.0 -Phadoop-2.3 -Ppyspark -Psparkr -Pscalding -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" TEST_PROJECTS=""
 
     # Test all modules with scala 2.10
     - jdk: "oraclejdk7"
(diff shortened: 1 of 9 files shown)


lresende:
One of the builds failed with the following, which seems not related

  The code passed to eventually never returned normally. Attempted 1 times over 1.164197794 seconds. Last failure message: 0 was not equal to 1. (AbstractAngularElemTest.scala:72)

lresende:
@bzz @Leemoonsoo Please review

.travis.yml
```diff
@@ -36,6 +36,10 @@ matrix:
   include:
     # Test all modules with spark 2.0.0 and scala 2.11
```
Leemoonsoo: Thanks for the contribution. Shell we have better comment here?

lresende: @Leemoonsoo updated.

zeppelin-zengine/pom.xml
```diff
@@ -129,12 +129,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.eclipse.jetty.websocket</groupId>
```
bzz: Is this related to RAT?

bzz:
I think having a separate profile it is reasonable approach for CI. Though I also think there is value in keepinv the default behaviour for local Dev workflow and make RAT check on by default.

In order to communicate the value of such fix better, I would suggest  linking to JIRA/here the logs of CI builds that have faild due to RAT issue as well.

lresende:
@bzz  this has now been updated to follow the same pattern used for maven tests, enable RAT to run by default, but support disabling it with -DskipRat. Travis CI will run RAT once, on the RAT build, and disable RAT checks on all other build profiles.

corneadoug:
Then how about we use -DskipRat on all building profiles, and have a profile that run only RAT without building?

Build time is pretty long, so it would be nice to separate some easy check that do not need compilation or specific settings separated, we would keep the CI feedback loop shorter and it would also be easier to know what failed without going to the log.

It could probably also be applied to things like Style Check or Unit Tests later

lresende:
@corneadoug This is exactly what is implemented on this pr.

corneadoug:
@lresende My bad, when I first read I understood that RAT would be included in one profile that would still build everything. (just like we do with the -DskipTests)

felixcheung:
LGTM

bzz:
Looks good to me, merging if there is no further discussion.

CI fails on single profile with 1 flaky test

```
[INFO] Zeppelin: Server ................................... FAILURE [03:39 min]

Failed tests: 
  ZeppelinSparkClusterTest.zRunTest:204 expected:<FINISHED> but was:<ERROR>
```

PR #1404, opened 2016-09-06
ZEPPELIN-1411. UDF with pyspark not working - object has no attribute 'parseDataType'

zjffdu:
### What is this PR for?
The root cause is that SQLContext's signature changes in spark 2.0. 
Spark 1.6
```
def __init__(self, sparkContext, sqlContext=None):
```
Spark 2.0
```
def __init__(self, sparkContext, sparkSession=None, jsqlContext=None):
```
So we need to create SQLContext using named parameters, otherwise it would take intp.getSQLContext() as sparkSession which cause the issue.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1411

Tested using the example code in ZEPPELIN-1411. 

image

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -118,7 +118,7 @@ public class SparkInterpreter extends Interpreter {
 
   private Map<String, Object> binder;
   private SparkVersion sparkVersion;
-  private File outputDir;          // class outputdir for scala 2.11
+  private static File outputDir;          // class outputdir for scala 2.11
   private Object classServer;      // classserver for scala 2.11
 
 
@@ -572,8 +572,11 @@ public class SparkInterpreter extends Interpreter {
         sparkReplClassDir = System.getProperty("java.io.tmpdir");
       }
 
-      outputDir = createTempDir(sparkReplClassDir);
-
+      synchronized (sharedInterpreterLock) {
+        if (outputDir == null) {
+          outputDir = createTempDir(sparkReplClassDir);
+        }
+      }
       argList.add("-Yrepl-class-based");
       argList.add("-Yrepl-outdir");
       argList.add(outputDir.getAbsolutePath());
@@ -1276,7 +1279,12 @@ public class SparkInterpreter extends Interpreter {
     logger.info("Close interpreter");
 
     if (numReferenceOfSparkContext.decrementAndGet() == 0) {
-      sc.stop();
+      if (sparkSession != null) {
+        Utils.invokeMethod(sparkSession, "stop");
+      } else if (sc != null){
+        sc.stop();
+      }
+      sparkSession = null;
       sc = null;
       if (classServer != null) {
         Utils.invokeMethod(classServer, "stop");
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
@@ -171,7 +171,12 @@ public abstract class AbstractTestRestApi {
 
         String sparkHome = getSparkHome();
         if (sparkHome != null) {
-          sparkIntpSetting.getProperties().setProperty("master", "spark://" + getHostname() + ":7071");
+          if (System.getenv("SPARK_MASTER") != null) {
+            sparkIntpSetting.getProperties().setProperty("master", System.getenv("SPARK_MASTER"));
+          } else {
+            sparkIntpSetting.getProperties()
+                    .setProperty("master", "spark://" + getHostname() + ":7071");
+          }
           sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");
           // set spark home for pyspark
           sparkIntpSetting.getProperties().setProperty("spark.home", sparkHome);
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
@@ -154,7 +154,7 @@ public abstract class AbstractTestRestApi {
         // set spark master and other properties
         sparkIntpSetting.getProperties().setProperty("master", "spark://" + getHostname() + ":7071");
         sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");
-
+        sparkIntpSetting.getProperties().setProperty("zeppelin.spark.useHiveContext", "false");
         // set spark home for pyspark
         sparkIntpSetting.getProperties().setProperty("spark.home", getSparkHome());
         pySpark = true;
@@ -180,6 +180,7 @@ public abstract class AbstractTestRestApi {
           sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");
           // set spark home for pyspark
           sparkIntpSetting.getProperties().setProperty("spark.home", sparkHome);
+          sparkIntpSetting.getProperties().setProperty("zeppelin.spark.useHiveContext", "false");
           pySpark = true;
           sparkR = true;
         }
@@ -199,7 +200,11 @@ public abstract class AbstractTestRestApi {
   }
 
   private static String getSparkHome() {
-    String sparkHome = getSparkHomeRecursively(new File(System.getProperty(ZeppelinConfiguration.ConfVars.ZEPPELIN_HOME.getVarName())));
+    String sparkHome = System.getenv("SPARK_HOME");
+    if (sparkHome != null) {
+      return sparkHome;
+    }
+    sparkHome = getSparkHomeRecursively(new File(System.getProperty(ZeppelinConfiguration.ConfVars.ZEPPELIN_HOME.getVarName())));
     System.out.println("SPARK HOME detected " + sparkHome);
     return sparkHome;
   }
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -218,7 +218,10 @@ java_import(gateway.jvm, "scala.Tuple2")
 jconf = intp.getSparkConf()
 conf = SparkConf(_jvm = gateway.jvm, _jconf = jconf)
 sc = SparkContext(jsc=jsc, gateway=gateway, conf=conf)
-sqlc = SQLContext(sc, intp.getSQLContext())
+if sparkVersion.isSpark2():
+  sqlc = SQLContext(sparkContext=sc, jsqlContext=intp.getSQLContext())
+else:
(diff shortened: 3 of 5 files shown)


zjffdu:
\cc @Leemoonsoo  Please help review, thanks

felixcheung:
LGTM
It'll be great to add some test for pyspark interpreter?

zjffdu:
It guess it is because PythonInterpreter depends on python environment, so there's no test for it yet.

felixcheung:
right. probably another PR, but I think we could use travis' addons support to install python via apt-get https://docs.travis-ci.com/user/installing-dependencies/
as we have for R.

felixcheung:
CI failed because of selenium, I think

felixcheung:
could you kick off CI again? Let's merge this after

Leemoonsoo:
Thanks @zjffdu for the contribution. Actually, we do have some tests for pyspark already.

Please see https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java#L125

If it's not too much difficult, adding unit test for this case would be really beneficial.

Leemoonsoo:
I tested this branch with given example, but it doesn't work for me.
On my machine, it hangs on `sqlContext.createDataFrame()` and end up with errors like

```
kqueue: Too many open files in system
```

I'm not sure it's problem of this patch or not.
Could someone else test this patch, too?

zjffdu:
@Leemoonsoo  Can you guide me how to run this test ? I try to run it using maven, but fails, seems it depends on something.

Leemoonsoo:
@zjffdu 

Once you build zeppelin,
```
mvn package [your profiles] -DskipTests
```

Then you can run this test, like
```
mvn package -pl 'zeppelin-interpreter,zeppelin-zengine,zeppelin-server' -Dtest=ZeppelinSparkClusterTest -DfailIfNoTests=false -Drat.skip=true
```

Let me know if it does not work.

zjffdu:
@Leemoonsoo , I follow the above command, but seems it doesn't work. I check `AbstractTestRestApi`, It seems pyspark related job would only run either in travis CI or set in spark standalone with SPARK_HOME is setup (`pyspark` needs to be set as true).  Do I understand correctly ?
```
 // ci environment runs spark cluster for testing
      // so configure zeppelin use spark cluster
      if ("true".equals(System.getenv("CI"))) {
        // assume first one is spark
        InterpreterSetting sparkIntpSetting = null;
        for(InterpreterSetting intpSetting : ZeppelinServer.notebook.getInterpreterFactory().get()) {
          if (intpSetting.getName().equals("spark")) {
            sparkIntpSetting = intpSetting;
          }
        }

        // set spark master and other properties
        sparkIntpSetting.getProperties().setProperty("master", "spark://" + getHostname() + ":7071");
        sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");

        // set spark home for pyspark
        sparkIntpSetting.getProperties().setProperty("spark.home", getSparkHome());
        pySpark = true;
        sparkR = true;
        ZeppelinServer.notebook.getInterpreterFactory().restart(sparkIntpSetting.getId());
      } else {
        // assume first one is spark
        InterpreterSetting sparkIntpSetting = null;
        for(InterpreterSetting intpSetting : ZeppelinServer.notebook.getInterpreterFactory().get()) {
          if (intpSetting.getName().equals("spark")) {
            sparkIntpSetting = intpSetting;
          }
        }

        String sparkHome = getSparkHome();
        if (sparkHome != null) {
          sparkIntpSetting.getProperties().setProperty("master", "spark://" + getHostname() + ":7071");
          sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");
          // set spark home for pyspark
          sparkIntpSetting.getProperties().setProperty("spark.home", sparkHome);
          pySpark = true;
          sparkR = true;
        }
```

Leemoonsoo:
@zjffdu Right, it looks like AbstractTestRestApi need to be improved when CI is not defined.
So far, i think you can try download and run spark standalone cluster in this way

```
./testing/downloadSpark.sh 1.6.2 2.6
./testing/startSparkCluster.sh 1.6.2 2.6
```

And then try run the test cases, so `getSparkHome()` can find sparkHome.

zjffdu:
@Leemoonsoo , I updated the unit test. And also made a little change `AbstractTestRestApi`, the standalone way doesn't work for me. So I allow user to export `SPAKP_MASTER` to run it in other modes.

Leemoonsoo:
Thanks @zjffdu. I think second ci test profile failure looks relevant.

```
ests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 120.231 sec <<< FAILURE! - in org.apache.zeppelin.rest.ZeppelinSparkClusterTest
pySparkTest(org.apache.zeppelin.rest.ZeppelinSparkClusterTest)  Time elapsed: 3.842 sec  <<< FAILURE!
java.lang.AssertionError: expected:<FINISHED> but was:<ERROR>
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.apache.zeppelin.rest.ZeppelinSparkClusterTest.pySparkTest(ZeppelinSparkClusterTest.java:150)
```

Could you check?

zjffdu:
@Leemoonsoo The test is passed in my local box both under spark 1.6.2 and spark 2.0.  Not sure why the travis CI fails, Is it possible for us to access the

zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
```diff
@@ -171,10 +171,16 @@ protected static void startUp() throws Exception {
...
+            sparkIntpSetting.getProperties().setProperty("master", System.getenv("SPARK_MASTER"));
+          } else {
+            sparkIntpSetting.getProperties()
+                    .setProperty("master", "spark://" + getHostname() + ":7071");
+          }
           sparkIntpSetting.getProperties().setProperty("spark.cores.max", "2");
           // set spark home for pyspark
           sparkIntpSetting.getProperties().setProperty("spark.home", sparkHome);
+          sparkIntpSetting.getProperties().setProperty("zeppelin.spark.useHiveContext", "false");
           pySpark = true;
```
zjffdu: Disable HiveContext, otherwise will hit the issue of multiple derby instance.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
```diff
@@ -171,10 +171,16 @@ protected static void startUp() throws Exception {
 
         String sparkHome = getSparkHome();
         if (sparkHome != null) {
-          sparkIntpSetting.getProperties().setProperty("master", "spark://" + getHostname() + ":7071");
+          if (System.getenv("SPARK_MASTER") != null) {
+            sparkIntpSetting.getProperties().setProperty("master", System.getenv("SPARK_MASTER"));
+          } else {
+            sparkIntpSetting.getProperties()
+                    .setProperty("master", "spark://" + getHostname() + ":7071");
+          }
```
zjffdu: Allow user to specify SPARK_MASTER, so that can run other modes (like yarn-client)

zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
```diff
@@ -194,7 +200,11 @@ private static String getHostname() {
   }
 
   private static String getSparkHome() {
-    String sparkHome = getSparkHomeRecursively(new File(System.getProperty(ZeppelinConfiguration.ConfVars.ZEPPELIN_HOME.getVarName())));
+    String sparkHome = System.getenv("SPARK_HOME");
+    if (sparkHome != null) {
+      return sparkHome;
+    }
+    sparkHome = getSparkHomeRecursively(new File(System.getProperty(ZeppelinConfiguration.ConfVars.ZEPPELIN_HOME.getVarName())));
     System.out.println("SPARK HOME detected " + sparkHome);
```
zjffdu: Allow user to specify SPARK_HOME, so that can use existing spark cluster

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1278,6 +1281,7 @@ public void close() {
     if (numReferenceOfSparkContext.decrementAndGet() == 0) {
       sc.stop();
       sc = null;
+      sparkSession = null;
       if (classServer != null) {
```
zjffdu: set sparkSession as null, so that it will be created again if the interpreter is scoped

zjffdu:
@Leemoonsoo Finally got the unit test passed (The remaining one failure is irrelevant).

Actually the test failure is caused by several bugs of `SparkInterpreter`.  Here's the steps to reproduce the first issue.
1. Use Spark2.0 and make SparkInterpeter as scoped
2. Open note1 and run sample code  to start `SparkInterperer`, then Open note2 and run sample code to start another `SparkInterpreter`, then you will hit the following issue.
 
image

The root cause of this issue is that the `outputDir` should be unique otherwise the second `SparkInterpreter` instance can not find the class in the `outputDir` of previous `SparkInterpeter`.

The second bug is that we should also set `sparkSession` as null. Otherwise it won't be created in the next second `SparkInterperter`. 

The  third bug is that we should disable `HiveContext` in `AbstractTestRestApi`, otherwise we will hit the issue of multiple derby instances running.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
felixcheung: `stop` should be called on sparkSession before `sc.stop()`
http://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.sql.SparkSession

(as of now this is ok since `sparkSession.stop()` simply calls `sc.stop()` but this could change)

zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
felixcheung: this is testing code only, but doesn't seem like we are using this in tests?

zjffdu: It is for local system test when user want to run it in other modes (e.g. yarn-client).

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: Good catch, when sparkSession is not null (spark 2.0), sparkSession.stop() should be called first.

Leemoonsoo:
@zjffdu Thanks for great work!
CI failure looks irrelevant. Merge into master and branch-0.6 if there're no more discussions.

PR #1356, opened 2016-08-24
[ZEPPELIN-1346] Add contribution guidelines to the website

corneadoug:
### What is this PR for?
Recreation of #1354 
This PR is bringing Contribution guidelines to the website.
We also introduce a sidemenu on those pages allowing to create additional content.
As an example, for the Web Application Guidelines, we added a page about "Defining Components"

In a different PR, we will remove the contributing markdown, and change the links in the repository to the website.

The layout of the doc was slightly modified to allow a larger documentation at max size.
It was also modified to accommodate better on smaller screens.

### What type of PR is it?
Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1346

### How should this be tested?
Build the website locally and visit the community page

### Screenshots (if appropriate)
screen shot 2016-08-23 at 3 50 41 pm

screen shot 2016-08-31 at 10 33 36 am

#### Layout update (Before / After)

Navbar Bug fix

screen shot 2016-08-31 at 10 35 14 am

screen shot 2016-08-31 at 10 31 18 am

Small Device Margin Improvement

screen shot 2016-08-31 at 10 35 34 am

screen shot 2016-08-31 at 10 31 28 am

Content Width Change

screen shot 2016-08-31 at 10 33 50 am

screen shot 2016-08-31 at 10 32 12 am

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1405, opened 2016-09-06
Update Zeppelin Website building instructions

corneadoug:
### What is this PR for?
Usually in order to push jekyll code to github pages, it will compile using the `--safe`
One drawback is that you cannot use any plugins.

After working on PR #1356 and introduction a plugin, using the build line with `--safe` specified in the README is failing.

However in our case, we are not forced to use that flag, since we are sending the compiled website directly to the Apache SVN, and I was able to confirm that fact when I sent the updated website to the SVN and it was rendered correctly

Also since plugins like Google Analytics needs to be activated only in production we are adding a special production environment option to add `JEKYLL_ENV=production`

### What type of PR is it?
Documentation

### Questions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ See https://help.github.com/articles/using-jekyll-with-pages#installing-jekyll
 ## Deploy to ASF svnpubsub infra (committers only)
  1. generate static website in `./_site`
     ```
-    bundle exec jekyll build
+    JEKYLL_ENV=production bundle exec jekyll build
     ```
 
  2. checkout ASF repo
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
@corneadoug There is one more `--safe` in here : [https://github.com/apache/zeppelin/blob/gh-pages/contribution/documentation.md#deploy-to-asf-svnpubsub-infra](https://github.com/apache/zeppelin/blob/gh-pages/contribution/documentation.md#deploy-to-asf-svnpubsub-infra)

corneadoug:
Let's Merge This since #1410 was merged too

AhyoungRyu:
@corneadoug Yeah please do :)

PR #1407, opened 2016-09-06
[ZEPPELIN-1412] add support multiline for pythonErrorIn method on python interpreter

cloverhearts:
### What is this PR for?
currently, has not support multiline exception text on python interpreter.
for example:

```
Exception: blabla
```
is error.

but

```
  File "<stdin>", line 1, in <module>
Exception: test exception
```
is sucess (now)

to resolve this issue.

### What type of PR is it?
Bug Fix

### Todos
- [x] modification pythonErrorIn method
- [x] add test case

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1412?jql=project%20%3D%20ZEPPELIN%20AND%20status%20%3D%20Open

added test case.

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -150,8 +150,17 @@ public class PythonInterpreter extends Interpreter {
    * @return true if syntax error or exception has happened
    */
   private boolean pythonErrorIn(String output) {
-    Matcher errorMatcher = errorInLastLine.matcher(output);
-    return errorMatcher.find();
+    boolean isError = false;
+    String[] outputMultiline = output.split("\n");
+    Matcher errorMatcher;
+    for (String row : outputMultiline) {
+      errorMatcher = errorInLastLine.matcher(row);
+      if (errorMatcher.find() == true) {
+        isError = true;
+        break;
+      }
+    }
+    return isError;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


bzz:
Thank you for improvements, @cloverhearts 

Looks good to me.

Although as discussed under [ZEPPELIN-1360](https://issues.apache.org/jira/browse/ZEPPELIN-1360) - this approach have performance implications and this code most probably is going to be removed under [ZEPPELIN-1325](https://issues.apache.org/jira/browse/ZEPPELIN-1325).

cloverhearts:
@bzz 
thank you.
This code works with both Python 3 and 2.
In this code, the part that related to the Python version does not exist except `print (" Exception ( "test exception")) `entry in the test code.
Thank you for your answer.

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -150,8 +150,17 @@ public InterpreterResult interpret(String cmd, InterpreterContext contextInterpr
    * @return true if syntax error or exception has happened
    */
   private boolean pythonErrorIn(String output) {
-    Matcher errorMatcher = errorInLastLine.matcher(output);
-    return errorMatcher.find();
+    boolean isError = false;
+    String[] outputMultiline = output.split("\n");
+    Matcher errorMatcher;
+    for (String row : outputMultiline) {
```
felixcheung: should it look at only the last line?

cloverhearts: @felixcheung  Sorry, Can you please explain in more detail?

bzz: Sorry guys, I missed GH notifications on this discussion and merged it :\

PR #1414, opened 2016-09-08
filter applied to graph

lobamba:
### What is this PR for?
We were looking for a Business Intelligence tools that allow to make  data analysis with nice charting.
We decided to use zeppelin that enables interactive data analytics.
We modified the source code to add more functionality:
For a first step we applied a filter to data.
User can now filter the data by clicking on graph. The filter is applied on every paragraph that have the same data(same column name).
On top you can see the filtered value. 

### What type of PR is it?
Improvement

### What is the Jira issue?
* No Jira created

<a href="https://raw.githubusercontent.com/lobamba/Test/master/graph_filter.gif" target="_blank">
graph filterd
</a>

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook-actionBar.html
+++ b/zeppelin-web/src/app/notebook/notebook-actionBar.html
@@ -109,6 +109,15 @@ limitations under the License.
           data-toggle="dropdown" id="revisionsDropdown">
           <span class="caret"></span>
         </button>
+        <span>
+        <button type="button" class="btn btn-default btn-xs"
+		  ng-class="{'active' : noFilter()}"
+		  ng-click="allParagraphFiltered(true)" tooltip-placement="bottom"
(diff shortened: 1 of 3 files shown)


Leemoonsoo:
Thanks @lobamba for the contribution.
It's interesting feature!

It'll be great if we can have more feedbacks and opinions here from the people, since this kind of feature has been never discussed in the community, as far as know. :-)

placeybordeaux:
I'd personally like to see this PR land.

pmccaffrey6:
This seems like a really cool idea and something that our group, for one, would love

AhyoungRyu:
@lobamba Great feature indeed!
As @Leemoonsoo said, I also would suggest to discuss about this newly proposed feature (maybe in `dev@zeppelin.apache.org`) first and get other ppl's opinion. It would be more helpful to proceed your work i think.

PR #1415, opened 2016-09-08
[ZEPPELIN-1026] set syntax highlight based on default bound interpreter

minahlee:
### What is this PR for?
This is complete work of #1148. Comments and tasks on #1148 has been handled in this PR.
- Add syntax language information in `interpreter-setting.json`
- When user type `%replName` in paragraph, back-end check if the interpreter name with `replName` exists, and return language information to front-end if it does
- If user doesn't specify `%replName`, default interpreter's language will be used
- Using alias name for paragraph syntax highlight

### What type of PR is it?
[Bug Fix | Improvement]

### What is the Jira issue?
[ZEPPELIN-1026](https://issues.apache.org/jira/browse/ZEPPELIN-1026)

1. Create new note and make markdown interpreter to be default.
2. See if markdown syntax is applied.

#### Case 1. When the default interpreter set to python interpreter.
**Before**
Has `scala` as syntax highlight language when %python is not set.
screen shot 2016-07-07 at 10 46 20 pm

**After**
Has `python` as syntax highlight language even when %python is not set.
screen shot 2016-07-07 at 10 44 39 pm

#### Case 2. When use alias name as repl name.
**Before**
screen shot 2016-09-08 at 4 22 39 pm

**After**
screen shot 2016-09-08 at 4 34 57 pm

### Further possible improvements
There are still several cases that Zeppelin doesn't handle syntax highlight well. These can be handled with another jira ticket/PR.
1. When default bound interpreter changes, syntax highlight is not changed accordingly
2. When copy/paste code, syntax highlight won't be applied properly since Zeppelin only checks changes when cursor is in first line.

### Questions:
* Does this needs documentation? yes(for creating new interpreter)

### DIFF
--- a/docs/development/writingzeppelininterpreter.md
+++ b/docs/development/writingzeppelininterpreter.md
@@ -69,6 +69,9 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
         "defaultValue": "property2DefaultValue",
         "description": "Property 2 description"
       }, ...
+    },
+    "editor": {
+      "language": "your-syntax-highlight-language"
     }
   },
   {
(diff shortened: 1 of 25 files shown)


minahlee:
CI is green, ready for review.

docs/development/writingzeppelininterpreter.md
```diff
@@ -69,6 +69,9 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
         "defaultValue": "property2DefaultValue",
         "description": "Property 2 description"
       }, ...
+    },
+    "editor": {
```
bzz: Quick question - it's a bit not clear if this section is mandatory and if not, what is going to be the default one? May be it could be described in docs below. 
Asking as there are i.e PRs that implement a new interpreters, before this section was introduced.

bzz:
Looks great to me.

docs/development/writingzeppelininterpreter.md
minahlee: Good point! If the language is not specified **plain text mode** will be used for syntax highlight, which will highlight nothing. I added description in c7221b2. Thanks for review :)

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -650,6 +652,40 @@ public void setInfo(Map<String, Object> info) {
     this.info = info;
   }
 
+  public Interpreter getRepl(String name) {
+    return factory.getInterpreter(getId(), name);
+  }
+
+  public Map<String, Object> getEditorSetting(String replName) {
```
Leemoonsoo: I think this method is more make sense to be placed in InterpreterFactory, while editor setting is related interpreter and interpreter setting, not each note. What do you think?

minahlee: @Leemoonsoo It totally makes sense, I made change in b2a051c. Please review.

Leemoonsoo:
@minahlee Great work! LGTM

minahlee:

PR #1419, opened 2016-09-09
ZEPPELIN-1420. java.util.ConcurrentModificationException caused by calling remove inside foreach loop

zjffdu:
### What is this PR for?
We should use Iterator to iterate the list when we want to remove items in the middle of iteration, Otherwise will get the ConcurrentModificationException

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1420

No test added

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -1017,15 +1017,16 @@ public class InterpreterFactory implements InterpreterGroupFactory {
   public List<InterpreterSetting> getInterpreterSettings(String noteId) {
     List<String> interpreterSettingIds = getNoteInterpreterSettingBinding(noteId);
     LinkedList<InterpreterSetting> settings = new LinkedList<>();
-    synchronized (interpreterSettingIds) {
-      for (String id : interpreterSettingIds) {
-        InterpreterSetting setting = get(id);
-        if (setting == null) {
-          // interpreter setting is removed from factory. remove id from here, too
-          interpreterSettingIds.remove(id);
-        } else {
-          settings.add(setting);
-        }
+
+    Iterator<String> iter = interpreterSettingIds.iterator();
+    while (iter.hasNext()) {
+      String id = iter.next();
+      InterpreterSetting setting = get(id);
+      if (setting == null) {
+        // interpreter setting is removed from factory. remove id from here, too
+        iter.remove();
+      } else {
+        settings.add(setting);
       }
     }
     return settings;


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -1017,15 +1017,16 @@ public void setInterpreters(String noteId, List<String> ids) throws IOException
...
-        InterpreterSetting setting = get(id);
-        if (setting == null) {
-          // interpreter setting is removed from factory. remove id from here, too
-          interpreterSettingIds.remove(id);
-        } else {
-          settings.add(setting);
-        }
+
+    Iterator<String> iter = interpreterSettingIds.iterator();
+    while (iter.hasNext()) {
```
zjffdu: I also remove the synchronized block, it seems not necessary.

Leemoonsoo:

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -1017,15 +1017,16 @@ public void setInterpreters(String noteId, List<String> ids) throws IOException
...
+      String id = iter.next();
+      InterpreterSetting setting = get(id);
+      if (setting == null) {
+        // interpreter setting is removed from factory. remove id from here, too
+        iter.remove();
```
prasadwagle: What is the purpose of removing the id? It seems to me that interpreterSettingIds is a temporary LinkedList and the effect of removing id is not persisted. 
  private List<String> getNoteInterpreterSettingBinding(String noteId) {
    LinkedList<String> bindings = new LinkedList<>();
    synchronized (interpreterSettings) {
      List<String> settingIds = interpreterBindings.get(noteId);
      if (settingIds != null) {
        bindings.addAll(settingIds);
      }
    }
    return bindings;
  }

zjffdu: @prasadwagle It is because the interpreter setting may be removed in other threads, so we need to check it here. I don't quite understand your concern about the persistence.

echarles: A few weeks ago, I came across the exact same question as @prasadwagle. Call remote on the temporary (not returned) list is useless (this consideration/question is separate from this PR which aims solving the concurrent modification exception (which I met also), but is worth to consider).

bzz:
Shall we merge this guy now?

khalidhuseynov:
tested locally, LGTM

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
khalidhuseynov: that's good question although not related to topic of this PR. maybe @jongyoul could bring little clarity as I believe he's the original author of that function

bzz:
Merging to master, if there is no further discussion.

PR #1421, opened 2016-09-09
[ZEPPELIN-1422][zeppelin-interpreter] fix InterpreterProperty equals methods, add null check

WeichenXu123:
### What is this PR for?
in PR https://github.com/apache/zeppelin/pull/1382,
there is a small bug that `InterpreterProperty.equals` do not check `null`.
jdk lib request the `null` check otherwise it has the risk causing java collections operation throw exception.
so I fix it.

### What type of PR is it?
Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1422

existing test.

### Questions:


Leemoonsoo:

prabhjyotsingh:
LGTM

PR #1422, opened 2016-09-09
Allow matplotlib plot display in pyspark context

hafizur-rahman:
### What is this PR for?
Similar to %python context, allow matplotlib plot display in %pyspark context

### What type of PR is it?
Improvement

The following code should display a graph:

```
%pyspark

import numpy as np
import matplotlib

matplotlib.use('Agg')

x = np.linspace(0, 1)
y = np.sin(4 * np.pi * x) * np.exp(-5 * x)

plt.figure()

plt.fill(x, y, 'r')
plt.grid(True)

z.show(plt, fmt="svg")
plt.close()
```

### DIFF
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -30,6 +30,13 @@ from pyspark.serializers import MarshalSerializer, PickleSerializer
 import ast
 import traceback
 
+import base64
+from io import BytesIO
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
 # for back compatibility
 from pyspark.sql import SQLContext, HiveContext, Row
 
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Could you try indent code with 2 spaces like other part of codes do?

hafizur-rahman:
Ok, giving a try...

Leemoonsoo:

hafizur-rahman:
Ok, please merge

bzz:
This code looks very similar to the one we have in [python interpreter](https://github.com/apache/zeppelin/blob/master/python/src/main/resources/bootstrap.py#L167), contributed by @bustios and @agoodm

FRosner:
Can we have some integration test to verify this one?

felixcheung:
where are we on this?

PR #1423, opened 2016-09-12
ZEPPELIN-1425. sparkr.zip is not distributed to executors

zjffdu:
### What is this PR for?
sparkr.zip is not distrubuted to executor, so any sparkR job that requrie R daemon in executor will fail. This PR would add sparkr.zip into `spark.yarn.dist.archives`. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1425

Run the following code
```
%spark.r

df <- createDataFrame(sqlContext, mtcars)
showDF(df)
```

image

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -329,6 +329,7 @@ public class SparkInterpreter extends Interpreter {
     }
 
     setupConfForPySpark(conf);
+    setupConfForSparkR(conf);
     Class SparkSession = Utils.findClass("org.apache.spark.sql.SparkSession");
     Object builder = Utils.invokeStaticMethod(SparkSession, "builder");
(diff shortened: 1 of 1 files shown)


zjffdu:
@Leemoonsoo  Please help review.

Leemoonsoo:
@zjffdu Do you think [rpkg.zip](https://github.com/apache/spark/blob/v1.6.2/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L394) is the something also need to be taken care?

zjffdu:
@Leemoonsoo  The `rpkg.zip` is for the case that user have R code inside of jars specified through --jars.  The reason why I didn't include it here is that I have to duplicate lots of code from spark (most of them is private, I have to copy them to zeppelin). Besides I think very few people would include R inside of jars. So I think it is acceptable to not include it for now.

BTW, I think zeppelin still duplicate lots of code of `SparkSubmit`, I plan to improve them in `ZEPPELIN-1263`.

Leemoonsoo:
Thanks for the explanation.

PR #1426, opened 2016-09-13
ZEPPELIN-1430. Display appId and webui link in LivyInterpreter's output

zjffdu:
### What is this PR for?
For now, it is hard to figure out what the yarn application of the livy session represent, it would be better to display the appId and webui link in the output of LivyInterpreter for diagnosing purpose. It can also be applied to the native SparkInterpreter, but could be done in another ticket. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1430

Tested manually

image

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivySparkInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivySparkInterpreter.java
@@ -38,11 +38,16 @@ public class LivySparkInterpreter extends Interpreter {
   private LivyOutputStream out;
 
   protected static Map<String, Integer> userSessionMap;
+  protected static Map<Integer, String> sessionId2AppIdMap;
+  protected static Map<Integer, String> sessionId2WebUIMap;
+
   private LivyHelper livyHelper;
 
   public LivySparkInterpreter(Properties property) {
     super(property);
     userSessionMap = new HashMap<>();
+    sessionId2AppIdMap = new HashMap<>();
+    sessionId2WebUIMap = new HashMap<>();
     livyHelper = new LivyHelper(property);
     out = new LivyOutputStream();
   }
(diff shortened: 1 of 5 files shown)


zjffdu:
@prabhjyotsingh @Leemoonsoo @felixcheung @bzz @minahlee Please help review.

I changed the livy output display style from plain text to angular in order to display the link. And for now it is only applicable in LivyScalaInterpreter, I can add it to Python/R interpreter as well when this approach is acceptable.

prabhjyotsingh:
Here is before and after of this PR that I observe, and I see these 2 changes

 - now it does not stream output, i.e. it waits for the entire paragraph execution to be finished, and returns result.
 - and I see empty line between "1" and "2" where I wrote "Thread.sleep()", not a big deal, but it wasn't earlier.

Before:
master
After:
with-1430

Leemoonsoo:
I think displaying result with additional information is quick and simple approach. However, i think there're some downsides like
  - can not leverage output streaming 
  - when user output want to use different zeppelin display system, like %table, %html, etc, it may be conflict with the feature this PR trying to provide.

I think better approach would be [support multiple interpreter result]( https://issues.apache.org/jira/browse/ZEPPELIN-212) and display additional informations as a one of separate interpreter result with it's own result type.

Although it'll going to be very large work, i think resolving ZEPPELIN-212 and build this feature on top of it, is right approach.

zjffdu:
Thanks for the info, I agree that the long term solution would be using [ support multiple interpreter result](https://issues.apache.org/jira/browse/ZEPPELIN-212). But for livy interpreter, it is very hard to figure out which spark app each session represent. And I think multiple user environment is common for livy users so additional info would be very helpful for livy users. So how about adding a new option to livy interpreter for this feature and by default it is turned off ? When [ support multiple interpreter result](https://issues.apache.org/jira/browse/ZEPPELIN-212) is implemented, we can turn the feature on by default.

zjffdu:
Add @felixcheung for review.

felixcheung:
I think the problem is if the output is a DataFrame now it can't be turned into a table since %angular is already there because only the first type is picked up? That would seem to me a bigger problem. 

Also, while I think it's important to support multiple output types, for interpreter output customization like this it might be an opportunity to create a new framework to allow this to plug in to the paragraph "gear" menu or similar UX decorators? This could be a great opportunity to add Spark UI web link to the Spark interpreter group too. 

I agree with @zjffdu that this is important to have in the short term, and I think it's a good idea if we change the output type if only under the switch for Livy interpreter only.

zjffdu:
@felixcheung DataFrame won't be affected, as it is handled in %livy.sql (`LivySparkSQLInterpreter`). And since `ZeppelinContext` is not available in livy, so you can not user `z.show(df)` in livy

zjffdu:
@prabhjyotsingh @felixcheung @Leemoonsoo I just update the PR to add a flag to display appInfo. By default it is false.

livy/src/main/java/org/apache/zeppelin/livy/LivySparkInterpreter.java
```diff
@@ -67,31 +74,66 @@ public void close() {
...
-                  "spark")
-          );
+          sessionId = livyHelper.createSession(interpreterContext, "spark");
+          userSessionMap.put(interpreterContext.getAuthenticationInfo().getUser(), sessionId);
+          if (displayAppInfo) {
+            String appId = extractStatementResult(
+                    livyHelper.interpret("sc.applicationId", interpreterContext, userSessionMap)
+                            .message());
+            livyHelper.interpret(
+                    "val webui=sc.getClass.getMethod(\"ui\").invoke(sc).asInstanceOf[Some[_]].get",
```
felixcheung: Is this and `appUIAddress` method exist in versions of Spark supported by this interpreter?

felixcheung:
LGTM except for one question

livy/src/main/java/org/apache/zeppelin/livy/LivySparkInterpreter.java
```diff
@@ -38,13 +38,20 @@
...
+  private boolean displayAppInfo;
 
   public LivySparkInterpreter(Properties property) {
     super(property);
     userSessionMap = new HashMap<>();
+    sessionId2AppIdMap = new HashMap<>();
+    sessionId2WebUIMap = new HashMap<>();
     livyHelper = new LivyHelper(property);
     out = new LivyOutputStream();
+    this.displayAppInfo = Boolean.parseBoolean(getProperty("zeppelin.livy.displayAppInfo"));
```
prabhjyotsingh: This key is different from `livy/src/main/resources/interpreter-setting.json` i.e. livy.spark.displayAppInfo

zjffdu: Thanks @prabhjyotsingh Typo is fixed. BTW, I only implement this feature for %livy.spark but not for other livy interpreter.

prabhjyotsingh:
Also, can you update the doc as well.
Rest LGTM.

prabhjyotsingh:

gss2002:
@zjffdu and @prabhjyotsingh the new static Maps can these be converted to ConcurrentHashMaps's to prevent possible future contention with multiple users/threads.

https://dzone.com/articles/java-7-hashmap-vs

PR #1429, opened 2016-09-14
ZEPPELIN-1440 Notebook clone: prefix name with "Copy of" and end with count

r-kamath:
### What is this PR for?
While cloning a notebook create a new name using the current notebook name by prefixing "Copy of" and end with count

e.g:
name: test
pre-filled clone name: Copy of test 1

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1440

Create a new book and clone on ui or see unit test in notename.js

Before
before

After
after

### Questions:
* Does the licenses files need update? na
* Is there breaking changes for older versions? na
* Does this needs documentation? na


felixcheung:
I think it might be easier if we don't prefix "Copy of"? We might need to localize or make this localizable at some point

r-kamath:
@felixcheung I think either we remove "Copy of" or make it localizable. Will wait comments from others before making the choice.

AhyoungRyu:
@r-kamath Thanks for the improvement. It's better than just "Untitled Note #" . 

But regarding `Copy of`, I agree with @felixcheung. If an user tries to clone a notebook in `this/is`, then `is 1` note will be created under `Copy of this` folder. It's a bit awkward in my point of view.
  

So just numbering would be enough I guess. What do you think? :)

corneadoug:
Except for the folder issue that @AhyoungRyu reported, I found another bug where the number do not increment:
* Make a clone
* Go back to the original note
* Make another clone, the number is still 1 even though 1 already exist

r-kamath:
@felixcheung removed "Copy of" prefix.
@AhyoungRyu fixed folder related issue.
@corneadoug fixed sequence bug.

AhyoungRyu:
@r-kamath Thanks! Tested again and looks good 👍

corneadoug:
@r-kamath Could you just rebase once more? I will make a final check after

prabhjyotsingh:
Tested, LGTM

zeppelin-web/test/spec/controllers/notename.js
```diff
@@ -0,0 +1,45 @@
...
+
+describe('Controller: NotenameCtrl', function() {
+  beforeEach(module('zeppelinWebApp'));
+
+  var scope;
+  var ctrl;
+  var notebookList;
+
+  beforeEach(inject(function($injector) {
+    notebookList = $injector.get('notebookListDataFactory');
```
corneadoug: Why 2 beforeEach functions?
This could be moved to the other one

corneadoug:
Except for my comment about the test, LGTM.
CI failure is irrelevant

prabhjyotsingh:
@r-kamath do you want to try and merge this ?
[This](https://cwiki.apache.org/confluence/display/ZEPPELIN/Committer+Guide) will help you.

r-kamath:
@corneadoug removed the redundant beforeEach b8b4f24
Will merge if there are no more discussion

corneadoug:
@r-kamath If you have trouble merging, tell me

PR #1434, opened 2016-09-18
[MINOR] Hide "Currently there are no properties ~" msg & "edit", "restart" and "remove" btn in edit mode

AhyoungRyu:
### What is this PR for?
"Currently there are no properties and dependencies set for this interpreter" message is shown when there are no properties & dependencies in interpreter setting page. But this message can be seen even in the edit mode. It's quite wired. This message needs to be shown only in non-editable mode.

The `edit`, `restart` and `remove` button are same. 
 - `edit` button doesn't do any work after changing to edit mode
 - `restart` button: it doesn't make sense that someone is editing and restarting interpreter at the same time
 - `remove` button: Maybe someone wants to remove the interpreter while he is editing the interpreter. But it would be better this button is shown only in non-editable mode for the consistency with the other buttons. 

For the above reasons, I changed those buttons to be shown in non-editable mode only. `Save` & `Cancel` buttons are enough to the edit page I think. Please see the attached gif image. 

### What type of PR is it?
Bug Fix

1. After applying this patch and build only `zeppelin-web` with `./grunt build` under `zeppelin-web`. 
2. You can quickly check this change in dev mode with `./grunt serve`. 
3. Go to `Interpreter` menu and click "edit" button on Angular interpreter to see "Currently there are no properties and dependencies set for this interpreter" message. Can also check the existence of `edit`, `restart` and `remove` buttons. 

 - Before
before

 - After 
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.html
+++ b/zeppelin-web/src/app/interpreter/interpreter.html
@@ -285,8 +285,8 @@ limitations under the License.
           <thead>
             <tr>
               <th style="width:40%">artifact</th>
-              <th>exclude</th>
-              <th ng-if="valueform.$visible">action</th>
+              <th style="width:40%">exclude</th>
+              <th style="width:20%" ng-if="valueform.$visible">action</th>
             </tr>
           </thead>
           <tr ng-repeat="dep in setting.dependencies">
--- a/zeppelin-web/src/app/interpreter/interpreter.html
+++ b/zeppelin-web/src/app/interpreter/interpreter.html
@@ -299,7 +299,7 @@ limitations under the License.
             <td>
               <textarea ng-if="valueform.$visible" ng-model="dep.exclusions"
                         placeholder="(Optional) comma separated groupId:artifactId list"
-                        e-form="valueform"
+                        form="valueform"
                         e-msd-elastic
                         ng-list="">
               </textarea>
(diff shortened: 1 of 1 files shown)


zeppelin-web/src/app/interpreter/interpreter.html
```diff
@@ -299,7 +299,7 @@ <h3 class="interpreter-title">{{setting.name}}
             <td>
               <textarea ng-if="valueform.$visible" ng-model="dep.exclusions"
                         placeholder="(Optional) comma separated groupId:artifactId list"
-                        form="valueform"
+                        e-form="valueform"
```
corneadoug: No `e-form` is needed here,
Actually the `e-msd-elastic` below should also be `msd-elastic`
those e-properties are only necessary in elements like `editable-text`

corneadoug:
tested, except for the comment, LGTM

zeppelin-web/src/app/interpreter/interpreter.html
AhyoungRyu: @corneadoug Ah didn't know that. I'm using IntelliJ now, it said like below for `form="valueform"`.
screen shot 2016-09-19 at 4 38 50 pm

Anyway I reverted it :)

corneadoug:
@AhyoungRyu Test, Nice Improvements
Can you just rebase to have a green CI?
After that, Merging if there is no more discussions

PR #1435, opened 2016-09-19
ZEPPELIN-1368. interpreter-setting.json may be loaded mutliple times

zjffdu:
### What is this PR for?
here're several ways to load interpreter-setting.json, but for now we will load it multiple times. It is supposed to load only once. We should load it by the following orders
         * 1. Register it from path {ZEPPELIN_HOME}/interpreter/{interpreter_name}/ interpreter-setting.json
         * 2. Register it from interpreter-setting.json in classpath {ZEPPELIN_HOME}/interpreter/{interpreter_name}
         * 3. Register it by Interpreter.register

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1368

Check the log that each interpreter is registered once. And also modify file interpreter/spark/interpreter-setting.json to make pyspark as the default interpreter and it works. Before this PR, it doesn't work, because it would be override by interpreter-setting.json in `interpreter/spark/zeppelin-spark_2.10-0.7.0-SNAPSHOT.jar`

image

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
@@ -338,9 +338,9 @@ public abstract class Interpreter {
   @Deprecated
   public static void register(String name, String group, String className,
       boolean defaultInterpreter, Map<String, InterpreterProperty> properties) {
-    logger.error("Static initialization is deprecated. You should change it to use " +
-                     "interpreter-setting.json in your jar or " +
-                     "interpreter/{interpreter}/interpreter-setting.json");
+    logger.warn("Static initialization is deprecated for interpreter {}, You should change it " +
+                     "to use interpreter-setting.json in your jar or " +
+                     "interpreter/{interpreter}/interpreter-setting.json", name);
     register(new RegisteredInterpreter(name, group, className, defaultInterpreter, properties));
   }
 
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
```diff
@@ -338,9 +338,9 @@ public static void register(String name, String group, String className,
   @Deprecated
   public static void register(String name, String group, String className,
       boolean defaultInterpreter, Map<String, InterpreterProperty> properties) {
-    logger.error("Static initialization is deprecated. You should change it to use " +
-                     "interpreter-setting.json in your jar or " +
-                     "interpreter/{interpreter}/interpreter-setting.json");
+    logger.warn("Static initialization is deprecated for interpreter {}, You should change it " +
+                     "to use interpreter-setting.json in your jar or " +
+                     "interpreter/{interpreter}/interpreter-setting.json", name);
     register(new RegisteredInterpreter(name, group, className, defaultInterpreter, properties));
```
zjffdu: Just improve the logging.

zjffdu:
@jongyoul  Could you help review it ? Thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
jongyoul: Makes sense.

jongyoul:
Thanks Jeff, this patch fixes my mistake correctly. LGTM. If CI passes, I'll merge it asap.

zjffdu:
@jongyoul The test failure seems irrelevant. I create 2 tickets for these flaky test. 
* https://issues.apache.org/jira/browse/ZEPPELIN-1455
* https://issues.apache.org/jira/browse/ZEPPELIN-1456

PR #1437, opened 2016-09-19
ZEPPELIN-1454: Wrong property value on interpreter page

prabhjyotsingh:
### What is this PR for?
If for some reason (for example permission issue in file system) while saving interpreter setting, UI shows wrong value till the next restart.
IMO interpreter.json should be the source of truth, that should always be reflected on UI

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - read from file-system after saving

### What is the Jira issue?
* [ZEPPELIN-1454](https://issues.apache.org/jira/browse/ZEPPELIN-1454)

Change file system permission of "interpreter.json", and make it readonly, then on interpreter setting page try and change any property and refresh the page, refer screenshot

Before
fix-permission-before

After
fix-permission-after

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -839,17 +839,21 @@ public class InterpreterFactory implements InterpreterGroupFactory {
     synchronized (interpreterSettings) {
       InterpreterSetting intpsetting = interpreterSettings.get(id);
       if (intpsetting != null) {
+        try {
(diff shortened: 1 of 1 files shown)


randerzander:
I believe the test failure is unrelated.

However, before merging this and implying support for the thin client, I ran into another issue with mangled JDBC properties in the URL: https://issues.apache.org/jira/browse/ZEPPELIN-1459

I think ZEPPELIN-1459 should be resolved before merging this doc update.

prabhjyotsingh:
@randerzander can you explain a bit more, how do you think this is related. As scope of this PR is only limited to interpreter property being saved (or fail to save) on file system.

prabhjyotsingh:
Will merge this in both master and 0.6.

PR #1440, opened 2016-09-20
ZEPPELIN-1427. Scala z.show() doesn't work on v.0.6.1

zjffdu:
### What is this PR for?
`ZeppelinContext.show` doesn't work for spark 1.6. The root cause is that `Dataset` is also available in spark 1.6, so the following line will be false when cls is `Dataset` while o is `Dataframe` in spark 1.6
```
if (cls.isInstance(o)) {
```
This PR create a list of supported class and make it a member of `ZeppelinContext `so that we don't need to create it every time. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1427

Tested it manually on spark 1.6 using the following sample code
```
z.show(sqlContext.sql("show tables"))
```

image

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -24,6 +24,7 @@ import static scala.collection.JavaConversions.collectionAsScalaIterable;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


zjffdu:
@Leemoonsoo  Please help review, I will add system test after #1404 is merged

Leemoonsoo:
Tested and it works well!

zjffdu:
@Leemoonsoo  I added some unit test, and also fix another bug. The bug is that exception will be throw when display `Dataset`, because we assume the type of each Record is Row while it is not true for `Dataset`, so I convert `Dataset` to `DataFrame` first when displaying Dataset. 

The failed test is irrelevant.

Leemoonsoo:
Thanks for the fix.
LGTM and merge to master and branch-0.6 if there're no more discussions.

PR #1443, opened 2016-09-20
[ZEPPELIN-1462] Zeppelin-Web Good Practices #3

corneadoug:
### What is this PR for?
This Good Practice Guide will focus on how to write a controller following the controller as vm concept, and how to provide a nice separation of the code.

### What type of PR is it?
Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1462

Check the `.md` file, or build the website to see the full rendering

### Questions:

### DIFF
--- a/contribution/zeppelinweb/goodPracticeGuide03.md
+++ b/contribution/zeppelinweb/goodPracticeGuide03.md
@@ -1,0 +1,145 @@
+---
+layout: sideMenu
+title: "3 - Making A Controller"
+description: ""
+group: nav-contrib-front
+menu: nav-contrib-front
+---
+<!--
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# How to Write a Controller
+
+<br/>
+Our main rule regarding writing a controller, is to use the `Controller as vm` style.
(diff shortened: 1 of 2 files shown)


corneadoug:
@AhyoungRyu Since you always find a few mistakes, can you make a quick spell check like always? :)

contribution/zeppelinweb/goodPracticeGuide03.md
```diff
@@ -0,0 +1,145 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# How to Write a Controller
+
+<br/>
+Our main rule regarding writing a controller, is to use the `Controller as vm` style.
+
+However we also have a few other style guidelines.
```
prabhjyotsingh: However,

corneadoug: @prabhjyotsingh You really like that comma :)

contribution/zeppelinweb/goodPracticeGuide03.md
```diff
@@ -0,0 +1,145 @@
...
+
+You can see below a full example of what the controller would look like.
+
```
AhyoungRyu: Do we need additional new lines? How about putting those three lines in one line? I mean

```
Our main rule regarding writing a controller, is to use the `Controller as vm` style.
However we also have a few other style guidelines. You can see below a full example of what the controller would look like.
```

corneadoug: I would have prefered those lines to be one under another without any additional spacing, but If I don't have that new line, the `You can` goes straight on the right side of the previous sentence

corneadoug: Eventually I could group the first two sentences:

```
Our main rule regarding writing a controller, is to use the `Controller as vm` style, however, we also have a few other style guidelines.
```

corneadoug:
Changes done.
Merging if there is no other discussions

PR #1446, opened 2016-09-21
ZEPPELIN-1263. Should specify zeppelin's spark configuration through --conf arguments of spark-submit

zjffdu:
### What is this PR for?

For now we spark configuration at runtime rather than pass them through `--conf`, it would cause several issues.
- Some configuration has to be set through --conf, otherwise we need to duplicate code in SparkSubmit.scala (spark.yarn.keytab, spark.yarn.principal)
- Some configuration would conflict with spark-defaults.conf. If you specify spark.master as yarn-client in spark-defaults.conf but specify spark.master as local in zeppelin side, you will see the spark interpreter fail to start due to this inconsistency. 
- As ZEPPELIN-1460 described, it is hard to figure what is the effective configuration. 
- We can not use yarn-cluster mode although it is not supported now, but I think it is necessary to do that as zeppelin needs to support multiple users.

This PR would pass all the spark related configuration to spark-submit through `--conf`, so that it is easy to know and guarantee that configuration on zeppelin interpreter setting take precedence over spark-defaults.conf.  And it is also good for maintenance that upstream change (any change about configuration in spark) would not affect us. 
### What type of PR is it?

[Improvement]
### Todos
- [ ] - Task
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-1263

Tested spark 1.6 spark 2.0 on both yarn-client mode and embedded mode. 

image
### Questions:

…


spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -371,18 +365,13 @@ private SparkContext createSparkContext_2() {
   }
 
   public SparkContext createSparkContext_1() {
-    logger.info("------ Create new SparkContext {} -------", getProperty("master"));
-
-    String execUri = System.getenv("SPARK_EXECUTOR_URI");
-    String[] jars = null;
-
-    if (Utils.isScala2_10()) {
```
felixcheung: Could you elaborate? seem like removing this could break some environment with jar already added to repl?

zjffdu: Do you mean enviroment variable `ADD_JARS` ? It is deprecated from spark 1.3 (SPARK-3340) and no spark document mention this variable, so I think it is safe to ignore it in zeppelin to keep the code clean (Actually I think we should drop support for old version of spark in some piont of future, otherwise the SparkInterpreter is hard to maintain ). If this is a concern, I can revert this piece of code.

zjffdu:
@Leemoonsoo @jongyoul @felixcheung Please help review.

felixcheung:
I think we still support Spark 1.2, 1.3 officially so removing this might be an issue.

jongyoul:
I didn't dig into it but have simple question. Does it support embedded mode?

zjffdu:
I haven't tried that, but I guess it won't work. Because embedded mode is special case for zeppelin, and zeppelin didn't use spark-submit to launch embedded mode, so I guess it won't work in embedded mode. The workaround is to still set the properties in java code like what we did for now.

zjffdu:
@jongyoul @Leemoonsoo @felixcheung Would you mind to take a look at this PR ? This issue blocks [ZEPPELIN-2715](https://issues.apache.org/jira/browse/ZEPPELIN-2715), [ZEPPELIN-2720](https://issues.apache.org/jira/browse/ZEPPELIN-2720) which I think is very critical.

Leemoonsoo:
I agree on the direction of this PR. But on the implementation I wish we try keep `class RemoteInterpreter` interpreter independent as much as possible.

This PR adds spark specific code to `RemoteInterpreter.getEnvFromInterpreterProperty()`.
And we used to have spark specific code in `bin/interpreter.sh`.

In the future, it'll be better if we provide generic facility in both `bin/interpreter.sh` and `class RemoteInterpreter` and move all spark specific code into spark interpreter module.

What do you think @zjffdu? This doesn't need to be addressed in this PR but if you think it's right direction, i wish we at least create a issue for it, so we can keep track on it.

Other than that, LGTM.

zjffdu:
Thanks @Leemoonsoo , Agree to move them to spark interpreter instead of `RemoteInterpreter`, it is actually on my plan of interpreter refactoring.

jongyoul:
You refactoring PR is already huge. I think creating new PR would be better.

zjffdu:
@jongyoul I won't do it in the existing interpreter refactoring PR. I will do it in other PRs, I believe there would still be several PRs for refactoring.

jongyoul:
Sounds good

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -140,15 +141,27 @@ public RemoteInterpreter(Properties property, String sessionKey, String classNam
...
-    for (Object key : property.keySet()) {
-      if (RemoteInterpreterUtils.isEnvString((String) key)) {
-        env.put((String) key, property.getProperty((String) key));
+    Map<String, String> env = new HashMap<String, String>();
+    StringBuilder sparkConfBuilder = new StringBuilder();
+    for (String key : property.stringPropertyNames()) {
+      if (RemoteInterpreterUtils.isEnvString(key)) {
+        env.put(key, property.getProperty(key));
+      }
+      if (key.equals("master")) {
```
felixcheung: I'm not sure we should special case this? we should ask use to use "spark.master"

zjffdu: This is just for compatibility for previous version where using `master`. See
https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java#L351

zjffdu: @felixcheung  Any more comments ? If not, I will merge it.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -489,13 +486,9 @@ public SparkContext createSparkContext_1() {
     if (execUri != null) {
       conf.set("spark.executor.uri", execUri);
     }
-    if (System.getenv("SPARK_HOME") != null) {
-      conf.setSparkHome(System.getenv("SPARK_HOME"));
```
felixcheung: should we remove this?

zjffdu: `SPARK_HOME` will be set by spark-submit script. Otherwise spark-submit can not launch spark app properly.

felixcheung: actually, spark-submit should be able to set SPARK_HOME itself too
https://github.com/apache/spark/blob/master/bin/spark-submit#L20

but I think Zeppelin requires SPARK_HOME to be set before calling spark-submit https://github.com/apache/zeppelin/blob/a3c78a4f49653ee66b0316c694bf57bf1c38a53b/bin/interpreter.sh#L108

zjffdu: `SparkInterpreter.java` runs in interpreter process which is launched by `interpreter.sh`. That means SPARK_HOME should already be set before `SparkInterpreter` is launched. So it is user's responsibility to set `SPARK_HOME` either in `zeppelin-env.sh` or interpreter setting.

zjffdu: ping @felixcheung

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -140,15 +141,27 @@ public RemoteInterpreter(Properties property, String sessionKey, String classNam
...
+        sparkConfBuilder.append(" --master " + property.getProperty("master"));
+      }
+      if (isSparkConf(key, property.getProperty(key))) {
+        sparkConfBuilder.append(" --conf " + key + "=" + property.getProperty(key));
```
felixcheung: for some of these values, perhaps we need to put it around quotes?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
felixcheung: there is still the "included version of spark" in Zeppelin? In that case SPARK_HOME is not set?
http://zeppelin.apache.org/docs/0.7.2/interpreter/spark.html#2-set-master-in-interpreter-menu

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -340,22 +340,19 @@ public boolean isYarnMode() {
    * Create SparkSession
    */
   public Object createSparkSession() {
-    logger.info("------ Create new SparkContext {} -------", getProperty("master"));
+    // use local mode for embedded spark mode when spark.master is not found
+    if (!conf.contains("spark.master")) {
+      conf.setMaster("local");
```
felixcheung: nit: you can also use http://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.SparkConf@setIfMissing(key:String,value:String):org.apache.spark.SparkConf

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
zjffdu: Good catch, fixed it

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: Fixed

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: Embedded spark still works

zjffdu:
I will merge it if no more comments.

PR #1458, opened 2016-09-26
[ZEPPELIN-1486] Allow configuring whether shown values will be trimmed

spektom:
### What is this PR for?
This patch improves Livy interpreter behavior: it allows configuring whether shown values will be trimmed or not.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1486

 * Set property `zeppelin.livy.spark.sql.truncateResult` to `true`
 * Run SQL query using Livy SQL interpreter
 * See whether long values are trimmed or not

image

### Questions:
* Does this needs documentation? - Yes. Documenting `zeppelin.livy.spark.sql.truncateResult` option.


AhyoungRyu:
Hi @spektom. Thanks for your contribution :)

As you said 
>Yes. Documenting zeppelin.livy.spark.sql.truncateResult option.

But it seems there is no docs file change. The below code needs to be added [here](https://github.com/apache/zeppelin/blob/master/docs/interpreter/livy.md#configuration) as well. 

```
<tr>
  <td>zeppelin.livy.spark.sql.truncateResult</td>
  <td>false</td>
  <td>Truncate long strings in SQL display</td>
</tr>
```

spektom:
@AhyoungRyu Added, thanks for pointing it out to me.

r-kamath:
@spektom thanks for the improvement. Can you also update`conf/zeppelin-env.sh.template` and `conf/zeppelin-site.xml.template` with env name/property name and description.

spektom:
@r-kamath I've added documentation to the `zeppelin-env.sh.template` file, but I couldn't find any other option related to Livy interpreter in `zeppelin-site.xml.template`, so I've left this file untouched.

jongyoul:
@spektom I agree with you not to add zeppelin-env.* because it depends on Live only.

livy/src/main/resources/interpreter-setting.json
```diff
@@ -100,6 +100,12 @@
         "defaultValue": "1000",
         "description": "Max number of Spark SQL result to display."
       },
+      "zeppelin.livy.spark.sql.truncateResult": {
+        "envName": "ZEPPELIN_LIVY_TRUNCATERESULT",
+        "propertyName": "zeppelin.livy.spark.sql.truncateResult",
+        "defaultValue": "false",
```
zjffdu: Will it make more sense to set `true` by default ?

livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
```diff
@@ -80,7 +80,8 @@ public InterpreterResult interpret(String line, InterpreterContext interpreterCo
               line.replaceAll("\"", "\\\\\"")
                   .replaceAll("\\n", " ")
               + "\").show(" +
-              property.get("zeppelin.livy.spark.sql.maxResult") + ")",
+              property.get("zeppelin.livy.spark.sql.maxResult") + "," +
+              property.get("zeppelin.livy.spark.sql.truncateResult") + ")",
           interpreterContext, userSessionMap);
```
zjffdu: What error will be displayed in frontend if I set invalid value for `zeppelin.livy.spark.sql.truncateResult` ?

spektom: There will be an exception thrown. I don't think every property must be checked for validity in an explicit way.

docs/interpreter/livy.md
```diff
@@ -61,10 +61,15 @@ Example: `spark.master` to `livy.spark.master`
...
   </tr>
   <tr>
-    <td>zeppelin.livy.spark.maxResult</td>
+    <td>zeppelin.livy.spark.sql.maxResult</td>
     <td>1000</td>
     <td>Max number of Spark SQL result to display.</td>
   </tr>
+  <tr>
+    <td>zeppelin.livy.spark.sql.truncateResult</td>
+    <td>false</td>
```
spektom: Changed

livy/src/main/resources/interpreter-setting.json
spektom: Changed

livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
zjffdu: What kind of exception do you see in frontend ? If user set an invalid value like 'F' and if the exception in frontend doesn't imply that, then it might be better to verify the property in method `open()`, otherwise it is hard to figure out what is wrong.

spektom: I don't see whether `zeppelin.livy.spark.sql.maxResult` or any other property is validated in any place. What's special about `zeppelin.livy.spark.sql.truncateResult`? Moreover, in my opinion, if user doesn't read documentation, and sets something completely wrong - it is his own problem.

gss2002: @spektom I hate to be the bearer of bad news but Zeppelin is being adopted by many Enterprises and unforuantely in most cases I agree that it's the users problem. But if you are going to be committing to software that is going to be used by large enterprises then I would expect validation to occur. I am actually going to raise these issues with my Hadoop vendor.

spektom:
Are you trying to say that people working for enterprise companies require
special input validation in addition to providing a documentation saying
that an option accepts booleans? :)
I don't think entry possible input must be validated, after all these are
Software engineers who will be seeing up Zeppelin...

gss2002:
Well we will just agree to disagree

gss2002:
Also input validation is secure java coding best practice regardless.. http://www.oracle.com/technetwork/java/seccodeguide-139067.html#5

PR #1467, opened 2016-09-28
[ZEPPELIN-1488]  JDBC Interpreter throws error while the interpreter is downloading dependencies

rajarajan-g:
### What is this PR for?
For first time, when we add dependencies for JDBC interpreter, dependencies will start getting downloaded in background. During that time, if user runs a paragraph of JDBC interpreter now user getting error , But instead paragraph execution should be put on 'PENDING' state and wait for dependencies to get downloaded and then run the paragraph 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1488] https://issues.apache.org/jira/browse/ZEPPELIN-1488

prerequisites:
1. Any DB setup.(For my testing, i considered hive)
Steps:
1. Delete the local-repo folder under zeppelin project, if it exists
2. Go to interpreter settings page, provide hive connection details under JDBC interprepreter
3. For hive interpreter to run, it needs some dependencies to be added in interpreter settings page
    For hive below dependencies needs to added
    1. org.apache.hive:hive-jdbc:0.14.0
    2. org.apache.hadoop:hadoop-common:2.6.0
4. Once the settings for JDBC interpreter is saved, dependencies will start getting downloaded in background.
5. Run any paragraph with JDBC as interpreter, paragraph should not throw error, status of the paragraph should change to 'pending' while the dependencies are getting downloaded in background.
6. Once the downloading of dependencies is done, the paragraph which were in pending will start executing in order depending on the execution mode of the interpreter (i.e Shared, Scoped, Isolated)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -289,9 +289,12 @@ public class Paragraph extends Job implements Serializable, Cloneable {
       logger.error("Can not find interpreter name " + repl);
       throw new RuntimeException("Can not find interpreter for " + getRequiredReplName());
     }
-
+    InterpreterSetting intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
+    while (intp.getStatus().equals(
+      org.apache.zeppelin.interpreter.InterpreterSetting.Status.DOWNLOADING_DEPENDENCIES)) {
+      intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
+    }
     if (this.noteHasUser() && this.noteHasInterpreters()) {
-      InterpreterSetting intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
       if (intp != null &&
         interpreterHasUser(intp) &&
         isUserAuthorizedToAccessInterpreter(intp.getOption()) == false) {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -292,7 +292,7 @@ public class Paragraph extends Job implements Serializable, Cloneable {
     InterpreterSetting intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
     while (intp.getStatus().equals(
       org.apache.zeppelin.interpreter.InterpreterSetting.Status.DOWNLOADING_DEPENDENCIES)) {
-      Thread.sleep(200);	
+      Thread.sleep(200);
       intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
     }
     if (this.noteHasUser() && this.noteHasInterpreters()) {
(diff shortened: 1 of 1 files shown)


rajarajan-g:
CI build is green.Please review

rajarajan-g:
Ping

jongyoul:
Can you answer my question?

rajarajan-g:
i didn't get the question. Can you please share it again ?

jongyoul:
I left in-line comment on your PR. Don't you see it?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -289,9 +289,12 @@ protected Object jobRun() throws Throwable {
       logger.error("Can not find interpreter name " + repl);
       throw new RuntimeException("Can not find interpreter for " + getRequiredReplName());
     }
-
+    InterpreterSetting intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
+    while (intp.getStatus().equals(
```
jongyoul: Oh.. my... Github's new review system is weird. Sorry for confusing you. My question is at this point. this loop execute too often. If an interpreter is under the downloading progress, it will take a long time, usually. I think we need to change this loop to run slowly. How do you think?

jongyoul:
I've left my review in four days ago. But I misunderstood new Github's review system. I left my comment again as a singe comment. Can you see it?

rajarajan-g:
Oh. Its' ok. now i see your comment.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
rajarajan-g: Even i was thinking about the same while making this changes. Do you think we should have a thread that should execute this function for every 'n' milliseconds. Should 1000 milliseconds be good enough ? 

Any thoughts would also be great

rajarajan-g: Please do check now

rajarajan-g:
ping

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
rajarajan-g: @jongyoul : Will changes be good enough now ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -289,9 +289,13 @@ protected Object jobRun() throws Throwable {
...
+      org.apache.zeppelin.interpreter.InterpreterSetting.Status.DOWNLOADING_DEPENDENCIES)) {
+      Thread.sleep(200);
+      intp = getInterpreterSettingById(repl.getInterpreterGroup().getId());
```
jongyoul: Why do you get `intp` every loop? AFAIK, it has same reference, doesn't it?

jongyoul:
LGTM. Merging there's no more discussion.

PR #1471, opened 2016-09-29
Closes [ZEPPELIN-1505] Add Scio interpreter

ravwojdyla:
### What is this PR for?

Closes #ZEPPELIN-1505. Adds Scio interpreter. Scio is a Scala DSL on top of Dataflow/Beam.
### What type of PR is it?

Improvement - ZEPPELIN-1505
### Todos
- [x] - test integration with zeppelin context (zeppelin context is too tightly coupled withs spark)
- [x] - what to do about code completion?
- [x] - add more tests?
- [x] - add helpers to display data
- [x] - add doc in `docs/interpreter/scio`
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1505

```
mvn -pl scio,zeppelin-display,zeppelin-interpreter -Dtest='org.apache.zeppelin.scio.*' -DfailIfNoTests=false test
```
### Screenshots

GitHub Logo
### Questions:
- Does the licenses files need update? ~~no~~ yes
- Does this needs documentation? yes (included in the PR)


scio/src/main/java/org/apache/zeppelin/scio/LogOutputStream.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.OutputStream;
+
+// Borrowed from spark interpreter.
+
+/**
+ * Minor modification of LogOutputStream of apache commons exec.
+ * LogOutputStream of apache commons exec has one issue that method flush doesn't throw IOException,
+ * so that SparkOutputStream can not extend it correctly.
+ */
+public abstract class LogOutputStream extends OutputStream {
```
zjffdu: Seems you are doing the same thing as me in (https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/LogOutputStream.java).  Maybe we can put it in module interpreter so that every interpreter implementation can use it. But could do it in a follow up ticket.

scio/src/main/scala/org/apache/zeppelin/scio/ScioInterpreter.scala
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  override def getFormType: FormType = {
+    FormType.NATIVE
+  }
+
+  override def getProgress(context: InterpreterContext): Int = {
+    // not implemented
+    42
+  }
```
zjffdu: why 42 ?

scio/src/main/resources/interpreter-setting.json
```diff
@@ -0,0 +1,19 @@
...
+  {
+    "group": "scio",
+    "name": "scio",
+    "className": "org.apache.zeppelin.scio.ScioInterpreter",
+    "defaultInterpreter": true,
+    "properties": {
+      "args": {
+        "envName": null,
+        "propertyName": null,
+        "defaultValue": "--runner=InProcessPipelineRunner",
```
zjffdu: propertyName is null ?

conf/interpreter-list
```diff
@@ -35,3 +35,4 @@ md              org.apache.zeppelin:zeppelin-markdown:0.6.1             Markdown
 postgresql      org.apache.zeppelin:zeppelin-postgresql:0.6.1           Postgresql interpreter
 python          org.apache.zeppelin:zeppelin-python:0.6.1               Python interpreter
 shell           org.apache.zeppelin:zeppelin-shell:0.6.1                Shell command
+scio            org.apache.zeppelin:zeppelin-scio:0.6.1                 Scio interpreter
```
AhyoungRyu: @ravwojdyla can we keep maintaining alphabetic order in here? need to change the position of `scio` with `shell`.

scio/README.md
```diff
@@ -0,0 +1,18 @@
+Scio interpreter for Apache Zeppelin
+====================================
+
+## Raison d'être:
```
AhyoungRyu: You mean `Purpose`? :)

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,90 @@
...
+    if (repl == null) {
+      intpGroup = new InterpreterGroup();
+      intpGroup.put("note", new LinkedList<Interpreter>());
+      repl = new ScioInterpreter(getScioTestProperties());
+      repl.setInterpreterGroup(intpGroup);
+      intpGroup.get("note").add(repl);
+      repl.open();
+    }
+
+    context = new InterpreterContext("note", "id", "title", "text",
```
AhyoungRyu: @ravwojdyla Zeppelin's java code block indentation is "2" space :)

AhyoungRyu:
@ravwojdyla Thanks for your contribution and welcome! 
I assume this PR is not finished but `Work In Process` since there are some unchecked TODOs. So how about changing the title of this PR from `Closes ZEPPELIN-1505: Add Scio interpreter ` to `[WIP][ZEPPELIN-1505] Add Scio interpreter` or sth? It would be more helpful to anyone who want to understand the purpose of this PR at a glance and start review :)

scio/README.md
ravwojdyla: I'm happy to change, that said `Raison d'être` is commonly used in English. What do you think?

scio/src/main/scala/org/apache/zeppelin/scio/ScioInterpreter.scala
ravwojdyla: There is no notion of overall progress in scio/beam, but would prefer it to be different than `0` just to see the different between execution vs interpreter startup. Is this fine?

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: This file seems to be indented with 2 spaces. Which part exactly is the problem?

ravwojdyla:
@AhyoungRyu @zjffdu thanks for prompt review. I have some questions in the comments, plus fixes coming in. I have also changed the topic of the PR.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,90 @@
...
+    assertEquals(InterpreterResult.Code.INCOMPLETE, incomplete.code());
+    assertTrue(incomplete.message().length() > 0);
+  }
+
+  @Test
+  public void testBasicPipeline() {
+    assertEquals(InterpreterResult.Code.SUCCESS,
+        repl.interpret("sc.parallelize(1 to 10).closeAndCollect().toList", context).code());
+  }
+}
```
ravwojdyla: Fixed. Added test for `ERROR`. Plus will add more as part of the PR.

conf/interpreter-list
ravwojdyla: Fixed.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
AhyoungRyu: @ravwojdyla Ah that's my bad. It's just a wrapping. Please ignore my comment for this :)

scio/src/main/resources/interpreter-setting.json
ravwojdyla: I have now changed it to fully named property.

ravwojdyla:
@AhyoungRyu @zjffdu could you please review when your have time. all the todos are ✔️

scio/src/main/scala/org/apache/zeppelin/scio/ScioInterpreter.scala
zjffdu: The convention is 0 if progress is not implemented IIRC. If the execution is started, the state would transite from `PENDING` to `RUNNING`

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void testBasicSuccess() {
+    assertEquals(InterpreterResult.Code.SUCCESS,
+        repl.interpret("val a = 1\nval b = 2", context).code());
+  }
+
+  @Test
+  public void testBasicSyntaxError() {
+    assertEquals(InterpreterResult.Code.ERROR,
+        repl.interpret("val a:Int = 'ds'", context).code());
+  }
```
zjffdu: It would be nice to verify the message as well so that user can see the right error message in frontend.

scio/README.md
AhyoungRyu: didn't know that. Good to Know :D

scio/src/main/scala/org/apache/zeppelin/scio/package.scala
```diff
@@ -0,0 +1,185 @@
...
+        } else {
+          first.getSchema.getFields.iterator.asScala.map(_.name()).toArray
+        }
+
+        val header = fieldNames.mkString(tab)
+        val firstStr = fieldNames.map(first.get).mkString(tab)
+        val content = it.take(maxResults)
+                        .map(r => fieldNames.map(r.get).mkString(tab))
+                        .mkString(newline)
+        println(s"$table $header$newline$firstStr$newline$content")
```
zjffdu: What is the purpose of println ? Is it for logging or display ? If it is for display, then does it support table format of zeppelin ?

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: Fixed, added:
 * multiline
 * throw exception

scio/src/main/scala/org/apache/zeppelin/scio/ScioInterpreter.scala
ravwojdyla: Fixed.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: Fixed.

scio/src/main/scala/org/apache/zeppelin/scio/package.scala
ravwojdyla: This is for the display system and the `%table` format specifically.

ravwojdyla:
@AhyoungRyu @zjffdu are there any more comment, how do you want to proceed? the error on travis does not seems to be related?

ravwojdyla:
ping @AhyoungRyu @zjffdu

zeppelin-zengine/pom.xml
```diff
@@ -96,6 +96,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-vfs2</artifactId>
       <version>2.0</version>
+      <exclusions>
```
jongyoul: Is this related to this PR?

zeppelin-zengine/pom.xml
```diff
@@ -107,6 +113,10 @@
            <groupId>org.apache.httpcomponents</groupId>
            <artifactId>httpclient</artifactId>
          </exclusion>
+         <exclusion>
```
jongyoul: Is this related to this PR?

zeppelin-zengine/pom.xml
ravwojdyla: It was an issue that came up while dev - fixed via 45decf5

zeppelin-zengine/pom.xml
ravwojdyla: It was an issue that came up while dev - fixed via 45decf5

AhyoungRyu:
@ravwojdyla Can you add `docs/interpreter/scio.md` path below [this line](https://github.com/apache/zeppelin/blob/master/docs/_includes/themes/zeppelin/_navigation.html#L68)?

zeppelin-zengine/pom.xml
ravwojdyla: fixed

zeppelin-zengine/pom.xml
ravwojdyla: fixed

docs/interpreter/scio.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Progress
+
+There can be only one paragraph running at a time. There is no notion of overall progress, thus progress bar will be `0`.
+
+### SCollection display helpers
+
+Scio interpreter comes with display helpers to ease working with Zeppelin notebooks. Simply use `closeAndDisplay()` on `SCollection` to close context and display the results. The number of results is limited by `zeppelin.scio.maxResult` (by default 1000).
+
+Supported `SCollection` types:
```
ravwojdyla: fixed

AhyoungRyu:
@ravwojdyla 
Here is the list of dependencies from `mvn dependency:tree`. All of them will become part of the Zeppelin binary package. What needed to be done is 

 - make sure that all these dependencies and the transitive dependencies as well
 - have an appropriate licence, and are logged under the `./zeppelin-distribution/src/bin_license/LICENSE`
 - and for non-Apache license, full text of the license need to be added to `./zeppelin-distribution/src/license/`

For your convenience I have listed all current dependencies of `scio` module below. You don't need to add all of the licenses listed in here. Many things are already existed. 

```
[INFO] Building Zeppelin: Scio 0.7.0-SNAPSHOT
[INFO]
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ zeppelin-scio_2.10 ---
[INFO] org.apache.zeppelin:zeppelin-scio_2.10:jar:0.7.0-SNAPSHOT
[INFO] +- org.slf4j:slf4j-api:jar:1.7.10:compile
[INFO] +- org.slf4j:slf4j-log4j12:jar:1.7.10:compile
[INFO] |  \- log4j:log4j:jar:1.2.17:compile
[INFO] +- org.apache.zeppelin:zeppelin-interpreter:jar:0.7.0-SNAPSHOT:compile
[INFO] |  +- org.apache.thrift:libthrift:jar:0.9.2:compile
[INFO] |  |  +- org.apache.httpcomponents:httpclient:jar:4.3.6:compile
[INFO] |  |  |  \- commons-codec:commons-codec:jar:1.5:compile
[INFO] |  |  \- org.apache.httpcomponents:httpcore:jar:4.3.3:compile
[INFO] |  +- com.google.code.gson:gson:jar:2.2:compile
[INFO] |  +- org.apache.commons:commons-pool2:jar:2.3:compile
[INFO] |  \- org.apache.commons:commons-lang3:jar:3.4:compile
[INFO] +- com.spotify:scio-repl_2.10:jar:0.2.3:compile
[INFO] |  +- com.spotify:scio-core_2.10:jar:0.2.3:compile
[INFO] |  |  +- com.spotify:scio-bigquery_2.10:jar:0.2.3:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-bigquery:jar:v2-rev317-1.22.0:compile
[INFO] |  |  |  \- org.joda:joda-convert:jar:1.8.1:compile
[INFO] |  |  +- com.google.cloud.dataflow:google-cloud-dataflow-java-sdk-all:jar:1.7.0:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-dataflow:jar:v1b3-rev36-1.22.0:compile
[INFO] |  |  |  +- io.grpc:grpc-all:jar:0.13.1:compile
[INFO] |  |  |  |  +- io.grpc:grpc-auth:jar:0.13.1:compile
[INFO] |  |  |  |  +- io.grpc:grpc-netty:jar:0.13.1:compile
[INFO] |  |  |  |  |  \- io.netty:netty-codec-http2:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  |     \- io.netty:netty-codec-http:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  +- io.grpc:grpc-protobuf:jar:0.13.1:compile
[INFO] |  |  |  |  +- io.grpc:grpc-core:jar:0.13.1:compile
[INFO] |  |  |  |  +- io.grpc:grpc-okhttp:jar:0.13.1:compile
[INFO] |  |  |  |  |  +- com.squareup.okio:okio:jar:1.6.0:compile
[INFO] |  |  |  |  |  \- com.squareup.okhttp:okhttp:jar:2.5.0:compile
[INFO] |  |  |  |  +- io.grpc:grpc-protobuf-nano:jar:0.13.1:compile
[INFO] |  |  |  |  |  \- com.google.protobuf.nano:protobuf-javanano:jar:3.0.0-alpha-5:compile
[INFO] |  |  |  |  \- io.grpc:grpc-stub:jar:0.13.1:compile
[INFO] |  |  |  +- io.netty:netty-handler:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  +- io.netty:netty-buffer:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  |  \- io.netty:netty-common:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  +- io.netty:netty-transport:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  |  \- io.netty:netty-resolver:jar:4.1.0.CR1:compile
[INFO] |  |  |  |  \- io.netty:netty-codec:jar:4.1.0.CR1:compile
[INFO] |  |  |  +- com.google.api.grpc:grpc-pubsub-v1:jar:0.0.2:compile
[INFO] |  |  |  |  \- com.google.api.grpc:grpc-core-proto:jar:0.0.3:compile
[INFO] |  |  |  +- com.google.auth:google-auth-library-oauth2-http:jar:0.4.0:compile
[INFO] |  |  |  |  \- com.google.auth:google-auth-library-credentials:jar:0.4.0:compile
[INFO] |  |  |  +- com.google.cloud.bigtable:bigtable-protos:jar:0.3.0:compile
[INFO] |  |  |  +- com.google.api-client:google-api-client:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.api-client:google-api-client-java6:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.api-client:google-api-client-jackson2:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-clouddebugger:jar:v2-rev8-1.22.0:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-pubsub:jar:v1-rev10-1.22.0:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-storage:jar:v1-rev71-1.22.0:compile
[INFO] |  |  |  +- com.google.http-client:google-http-client:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.http-client:google-http-client-jackson:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.http-client:google-http-client-jackson2:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.http-client:google-http-client-protobuf:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.oauth-client:google-oauth-client-java6:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.oauth-client:google-oauth-client:jar:1.22.0:compile
[INFO] |  |  |  +- com.google.apis:google-api-services-datastore-protobuf:jar:v1beta2-rev1-4.0.0:compile
[INFO] |  |  |  +- com.google.cloud.bigdataoss:gcsio:jar:1.4.5:compile
[INFO] |  |  |  +- com.google.cloud.datastore:datastore-v1-proto-client:jar:1.1.0:compile
[INFO] |  |  |  +- com.google.cloud.datastore:datastore-v1-protos:jar:1.0.1:compile
[INFO] |  |  |  +- com.google.cloud.bigdataoss:util:jar:1.4.5:compile
[INFO] |  |  |  +- com.google.protobuf:protobuf-java:jar:3.0.0-beta-1:compile
[INFO] |  |  |  +- com.google.code.findbugs:jsr305:jar:3.0.1:compile
[INFO] |  |  |  +- com.fasterxml.jackson.core:jackson-core:jar:2.7.0:compile
[INFO] |  |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.7.0:compile
[INFO] |  |  |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.7.0:compile
[INFO] |  |  |  +- org.apache.avro:avro:jar:1.7.7:compile
[INFO] |  |  |  |  +- org.codehaus.jackson:jackson-core-asl:jar:1.9.13:compile
[INFO] |  |  |  |  +- org.codehaus.jackson:jackson-mapper-asl:jar:1.9.13:compile
[INFO] |  |  |  |  \- com.thoughtworks.paranamer:paranamer:jar:2.3:compile
[INFO] |  |  |  +- org.xerial.snappy:snappy-java:jar:1.1.2.1:compile
[INFO] |  |  |  \- joda-time:joda-time:jar:2.4:compile
[INFO] |  |  +- com.twitter:algebird-core_2.10:jar:0.12.1:compile
[INFO] |  |  |  \- com.googlecode.javaewah:JavaEWAH:jar:0.6.6:compile
[INFO] |  |  +- com.twitter:chill_2.10:jar:0.8.0:compile
[INFO] |  |  |  +- com.twitter:chill-java:jar:0.8.0:compile
[INFO] |  |  |  \- com.esotericsoftware:kryo-shaded:jar:3.0.3:compile
[INFO] |  |  |     +- com.esotericsoftware:minlog:jar:1.3.0:compile
[INFO] |  |  |     \- org.objenesis:objenesis:jar:2.1:compile
[INFO] |  |  +- com.twitter:chill-protobuf:jar:0.8.0:compile
[INFO] |  |  +- commons-io:commons-io:jar:2.4:compile
[INFO] |  |  \- org.apache.commons:commons-math3:jar:3.6.1:compile
[INFO] |  +- com.spotify:scio-extra_2.10:jar:0.2.3:compile
[INFO] |  |  +- org.scalanlp:breeze_2.10:jar:0.12:compile
[INFO] |  |  |  +- org.scalanlp:breeze-macros_2.10:jar:0.12:compile
[INFO] |  |  |  +- com.github.fommil.netlib:core:jar:1.1.2:compile
[INFO] |  |  |  +- net.sourceforge.f2j:arpack_combined_all:jar:0.1:compile
[INFO] |  |  |  +- net.sf.opencsv:opencsv:jar:2.3:compile
[INFO] |  |  |  +- com.github.rwl:jtransforms:jar:2.4.0:compile
[INFO] |  |  |  +- org.spire-math:spire_2.10:jar:0.7.4:compile
[INFO] |  |  |  |  \- org.spire-math:spire-macros_2.10:jar:0.7.4:compile
[INFO] |  |  |  \- com.chuusai:shapeless_2.10.4:jar:2.0.0:compile
[INFO] |  |  \- info.debatty:java-lsh:jar:0.10:compile
[INFO] |  |     \- info.debatty:java-string-similarity:jar:0.12:compile
[INFO] |  +- org.slf4j:slf4j-simple:jar:1.7.21:compile
[INFO] |  +- jline:jline:jar:2.10:compile
[INFO] |  +- com.nrinaudo:kantan.csv_2.10:jar:0.1.12:compile
[INFO] |  |  \- com.nrinaudo:kantan.codecs_2.10:jar:0.1.6:compile
[INFO] |  +- org.scalamacros:paradise_2.10.6:jar:2.1.0:compile
[INFO] |  |  \- org.scalamacros:quasiquotes_2.10:jar:2.1.0:compile
[INFO] |  \- org.scala-lang:jline:jar:2.10.6:compile
[INFO] |     \- org.fusesource.jansi:jansi:jar:1.4:compile
[INFO] +- com.google.guava:guava:jar:14.0.1:compile
[INFO] +- org.apache.maven:maven-plugin-api:jar:3.0:compile
[INFO] |  \- org.apache.maven:maven-artifact:jar:3.0:compile
[INFO] +- org.sonatype.aether:aether-api:jar:1.12:compile
[INFO] +- org.sonatype.aether:aether-util:jar:1.12:compile
[INFO] +- org.sonatype.aether:aether-impl:jar:1.12:compile
[INFO] |  \- org.sonatype.aether:aether-spi:jar:1.12:compile
[INFO] +- org.apache.maven:maven-aether-provider:jar:3.0.3:compile
[INFO] |  +- org.apache.maven:maven-model:jar:3.0.3:compile
[INFO] |  +- org.apache.maven:maven-model-builder:jar:3.0.3:compile
[INFO] |  |  \- org.codehaus.plexus:plexus-interpolation:jar:1.14:compile
[INFO] |  +- org.apache.maven:maven-repository-metadata:jar:3.0.3:compile
[INFO] |  \- org.codehaus.plexus:plexus-component-annotations:jar:1.5.5:compile
[INFO] +- org.sonatype.aether:aether-connector-file:jar:1.12:compile
[INFO] +- org.sonatype.aether:aether-connector-wagon:jar:1.12:compile
[INFO] |  +- org.codehaus.plexus:plexus-classworlds:jar:2.4:compile
[INFO] |  +- org.codehaus.plexus:plexus-utils:jar:2.0.7:compile
[INFO] |  \- org.sonatype.sisu:sisu-inject-plexus:jar:2.2.2:compile
[INFO] |     \- org.sonatype.sisu:sisu-inject-bean:jar:2.2.2:compile
[INFO] |        \- org.sonatype.sisu:sisu-guice:jar:no_aop:3.0.2:compile
[INFO] +- org.apache.maven.wagon:wagon-provider-api:jar:1.0:compile
[INFO] +- org.apache.maven.wagon:wagon-http-lightweight:jar:1.0:compile
[INFO] +- org.apache.maven.wagon:wagon-http:jar:1.0:compile
[INFO] |  \- org.apache.maven.wagon:wagon-http-shared:jar:1.0:compile
[INFO] |     +- commons-httpclient:commons-httpclient:jar:3.1:compile
[INFO] |     +- commons-logging:commons-logging:jar:1.1.1:compile
[INFO] |     \- org.jsoup:jsoup:jar:1.6.1:compile
[INFO] +- org.apache.commons:commons-exec:jar:1.3:compile
[INFO] +- org.scala-lang:scala-library:jar:2.10.5:compile
[INFO] +- org.scala-lang:scala-compiler:jar:2.10.5:compile
[INFO] +- org.scala-lang:scala-reflect:jar:2.10.5:compile
[INFO] +- commons-lang:commons-lang:jar:2.5:compile
[INFO] +- org.apache.commons:commons-compress:jar:1.9:compile
[INFO] +- org.scalatest:scalatest_2.10:jar:2.2.4:test
[INFO] +- junit:junit:jar:4.11:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] \- com.spotify:scio-test_2.10:jar:0.2.3:test
[INFO]    +- com.novocode:junit-interface:jar:0.11:test
[INFO]    |  \- org.scala-sbt:test-interface:jar:1.0:test
[INFO]    \- org.hamcrest:hamcrest-all:jar:1.3:test
```

ravwojdyla:
@AhyoungRyu sorry I missed it. Will go ahead and add all the licenses. Thanks!

ravwojdyla:
@AhyoungRyu updated licenses. The travis failures do not seems to be related could you please restart the travis jobs?

AhyoungRyu:
@ravwojdyla Only the owner can re-trigger Travis job by closing & opening the PR :)

ravwojdyla:
The tests are failing for:

```
Tests run: 9, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 119.221 sec <<< FAILURE! - in org.apache.zeppelin.integration.ParagraphActionsIT
testWidth(org.apache.zeppelin.integration.ParagraphActionsIT)  Time elapsed: 21.445 sec  <<< FAILURE!
java.lang.AssertionError: New Width is : 4
Expected: <true>
     but: was <false>
	at org.apache.zeppelin.integration.ParagraphActionsIT.testWidth(ParagraphActionsIT.java:314)

testWidth(org.apache.zeppelin.integration.ParagraphActionsIT)  Time elapsed: 21.445 sec  <<< FAILURE!
java.lang.AssertionError: New Width is : 8
Expected: <true>
     but: was <false>
	at org.apache.zeppelin.integration.ParagraphActionsIT.testWidth(ParagraphActionsIT.java:314)
```

I can't see the connection of those with this PR.

docs/interpreter/scio.md
```diff
@@ -0,0 +1,150 @@
...
+
+</table>
+
+## Enabling the Scio Interpreter
+
+In a notebook, to enable the **Scio** interpreter, click the **Gear** icon and select **scio**.
+
+## Using the Scio Interpreter
+
+In a paragraph, use `%scio` to select the **Scio** interpreter. You can use it much the same way as vanilla Scala REPL and [Scio REPL](https://github.com/spotify/scio/wiki/Scio-REPL). Context is shared among all *Scio* paragraphs. There is a special variable **argz** which holds arguments from Scio interpreter settings. The easiest way to proceed is to create a context via standard `ContextAndArgs`.
```
zjffdu: Is Context shared in isolated mode ? If not, you need to highlight in which mode Context is shared.

docs/interpreter/scio.md
```diff
@@ -0,0 +1,150 @@
...
+
+```scala
+%scio
+val (sc, args) = ContextAndArgs(argz)
+```
```
zjffdu: Does `sc` need to be created explicitly ? SparkContext in spark is create implicitly, not sure the convention in scio

docs/interpreter/scio.md
```diff
@@ -0,0 +1,150 @@
...
+
+```scala
+%scio
+@BigQueryType.fromQuery("""|SELECT departure_airport,count(case when departure_delay>0 then 1 else 0 end) as no_of_delays
+                           |FROM [bigquery-samples:airline_ontime_data.flights]
+                           |group by departure_airport
+                           |order by 2 desc
+                           |limit 10""".stripMargin) class Flights
+
+val (sc, args) = ContextAndArgs(argz)
```
zjffdu: If sc is created every time in each paragraph ? How is it shared between paragraphs. ?

docs/interpreter/scio.md
```diff
@@ -0,0 +1,150 @@
...
+sc.typedBigQuery[Flights]().flatMap(_.no_of_delays).mean.closeAndDisplay()
+```
+
+#### Avro example:
+
+```scala
+%scio
+import com.spotify.data.ExampleAvro
+
```
zjffdu: Is this class available by default ? Does user need to add dependency in interpreter setting ?

docs/interpreter/scio.md
ravwojdyla: The context should be created explicitly - according to the doc. In scio/dataflow ` once "context" is closed, it has to be recreated. Single context equals single dataflow job.

docs/interpreter/scio.md
ravwojdyla: Thanks @zjffdu. The doc was ambiguous - fixed. The "REPL" state is shared, meaning variables, imports etc.

docs/interpreter/scio.md
ravwojdyla: That is just a example of using Avro class. `com.spotify.data.ExampleAvro` is not available by default.

ravwojdyla:
@zjffdu @AhyoungRyu are there any other comments?

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,118 @@
...
+      repl.open();
+    }
+
+    context = getNewContext();
+  }
+
+  @Test
+  public void testBasicSuccess() {
+    assertEquals(InterpreterResult.Code.SUCCESS,
+        repl.interpret("val a = 1" + newline + "val b = 2", context).code());
```
zjffdu: Could you also verify the `InterpreterResult.message()` ?

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Test
+  public void testBasicSyntaxError() {
+    InterpreterResult error = repl.interpret("val a:Int = 'ds'", context);
+    assertEquals(InterpreterResult.Code.ERROR, error.code());
+    assertTrue(error.message().length() > 0);
```
zjffdu: might be better to verify the error content.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,118 @@
...
+    InterpreterResult incomplete = repl.interpret("val a = \"\"\"", context);
+    assertEquals(InterpreterResult.Code.INCOMPLETE, incomplete.code());
+    assertTrue(incomplete.message().length() > 0);
+  }
+
+  @Test
+  public void testBasicPipeline() {
+    assertEquals(InterpreterResult.Code.SUCCESS,
+        repl.interpret("val (sc, _) = ContextAndArgs(argz)" + newline
+            + "sc.parallelize(1 to 10).closeAndCollect().toList", context).code());
```
zjffdu: ditto

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,118 @@
...
+  @Test
+  public void testBasicMultiStepPipeline() {
+    final StringBuilder code = new StringBuilder();
+    code.append("val (sc, _) = ContextAndArgs(argz)").append(newline)
+        .append("val numbers = sc.parallelize(1 to 10)").append(newline)
+        .append("val results = numbers.closeAndCollect().toList").append(newline)
+        .append("println(results)");
+    assertEquals(InterpreterResult.Code.SUCCESS,
+        repl.interpret(code.toString(), context).code());
+  }
```
zjffdu: ditto

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Test
+  public void testException() {
+    InterpreterResult exception = repl.interpret("val (sc, _) = ContextAndArgs(argz)" + newline
+        + "throw new Exception(\"test\")", context);
+    assertEquals(InterpreterResult.Code.ERROR, exception.code());
+    assertTrue(exception.message().length() > 0);
```
zjffdu: ditto

zjffdu:
@ravwojdyla I left some comments. And I am afraid no one in zeppelin community familiar with scio, could you ask someone else that know scio to review it ?

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: There is no message to verify.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: There is no message to verify.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: There is no message to verify.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: There is no message to verify.

scio/src/test/java/org/apache/zeppelin/scio/ScioInterpreterTest.java
ravwojdyla: Fixed.

ravwojdyla:
@nevillelyh (the creator of scio) - could you please review.

docs/interpreter/scio.md
```diff
@@ -0,0 +1,168 @@
...
+## Configuration
+<table class="table-configuration">
+  <tr>
+    <th>Name</th>
+    <th>Default Value</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>zeppelin.scio.argz</td>
```
nevillelyh: `argz` not `args`? Is that a Zeppelin convention?

nevillelyh: Scratch that just saw examples below.

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
```diff
@@ -0,0 +1,105 @@
...
+  private def materialize[T: ClassTag](self: SCollection[T]) = {
+    val f = self.materialize
+    self.context.close()
+    f
+  }
+
+  // TODO: scala 2.11
+  // implicit class ZeppelinSCollection[T: ClassTag](private val self: SCollection[T])(implicit ev: T <:< AnyVal) extends AnyVal {
+  implicit class ZeppelinSCollection[T: ClassTag](val self: SCollection[T])
+                                                 (implicit ev: T <:< AnyVal) {
```
nevillelyh: Why `AnyVal`? Should be `AnyRef` really?

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
```diff
@@ -0,0 +1,105 @@
...
+     * and display elements from Product like SCollection */
+    def closeAndDisplay(): Unit = {
+      DisplayTapImplicits.ZeppelinProductTap(materialize(self).waitForResult()).display()
+    }
+  }
+
+  // TODO: scala 2.11
+  // implicit class ZeppelinAvroSCollection[T: ClassTag](val self: SCollection[T])(implicit ev: T <:< GenericRecord) extends AnyVal {
+  implicit class ZeppelinAvroSCollection[T: ClassTag](val self: SCollection[T])
+                                                     (implicit ev: T <:< GenericRecord) {
```
nevillelyh: Both specific and generic records extend `IndexedRecord` but specific doesn't extend `GenericRecord`, is that gonna be a problem?

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
```diff
@@ -0,0 +1,105 @@
...
+import com.spotify.scio.values.SCollection
+import org.apache.avro.Schema
+import org.apache.avro.generic.GenericRecord
+
+import scala.reflect.ClassTag
+
+/**
+ * Implicit Zeppelin display helpers for SCollection.
+ */
+object DisplaySCollectionImplicits {
```
nevillelyh: One alternative to the current approach (one implicit class per type), is to have a single implicit class for all `SCollection[T]`, and peak the items in `Iterator[T]` and use reflection. It might make the code simpler but maybe less safe though.

scio/src/main/scala/org/apache/zeppelin/scio/DisplayHelpers.scala
```diff
@@ -0,0 +1,160 @@
...
+      notifyIfTruncated(it)
+    }
+  }
+
+  /**
+   * Displays Avro values from given [[Iterator]] using optional [[Schema]].
+   * @param schema optional "view" schema, otherwise schema is inferred from the first object
+   */
+  private[scio] def displayAvro[T: ClassTag](it: Iterator[T], schema: Schema = null)
+                                            (implicit ev: T <:< GenericRecord): Unit = {
```
nevillelyh: This won't work with `SpecificRecord` which doesn't extend `GenericRecord` but `IndexedRecord`.

scio/src/main/scala/org/apache/zeppelin/scio/DisplayHelpers.scala
```diff
@@ -0,0 +1,160 @@
...
+        .mkString(newline)
+      println(s"$table $header$newline$firstStr$newline$content")
+      notifyIfTruncated(it)
+    }
+  }
+
+  /**
+   * Displays [[TableRow]] values from given [[Iterator]] using specified [[TableSchema]].
+   */
+  private[scio] def displayBQTableRow[T: ClassTag](it: Iterator[T], schema: TableSchema)
```
nevillelyh: Possible to infer schema here?

scio/src/main/scala/org/apache/zeppelin/scio/DisplayTapImplicits.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.avro.Schema
+import org.apache.avro.generic.GenericRecord
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+/**
+ * Implicit Zeppelin display helpers for [[Tap]] and [[Future]] of a [[Tap]].
+ */
+object DisplayTapImplicits {
```
nevillelyh: Same question as above. Possible to use one implicit class and do reflection inside?

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
ravwojdyla: Nope - I specifically want to have just AnyVal tree over here.

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
ravwojdyla: Depending on the type of `T` in some cases I need extra info in display - which is simpler with current approach + as you say current one is more safe. It's not be overhead as most logic is in DisplayHelpers anyway. Would prefer to keep it this way, is this fine?

nevillelyh: Make sense ignore that then.

scio/src/main/scala/org/apache/zeppelin/scio/DisplayHelpers.scala
nevillelyh: 👍

scio/src/main/scala/org/apache/zeppelin/scio/DisplaySCollectionImplicits.scala
ravwojdyla: Starting from `1.7.5` `SpecificRecordBase` implements `GenericRecord` (~2013). In here we use 1.7.7. That said, it might worth supporting old versions and use `IndexedRecord`.

scio/src/main/scala/org/apache/zeppelin/scio/DisplayHelpers.scala
ravwojdyla: Starting from 1.7.5 SpecificRecordBase implements GenericRecord (~2013). In here we use 1.7.7. That said, it might worth supporting old versions and use IndexedRecord.

scio/src/main/scala/org/apache/zeppelin/scio/DisplayTapImplicits.scala
ravwojdyla: Same answer as above. Would prefer to keep it this way, same reasons:
 * difference in display methods
 * safer
 * not big overhead

ravwojdyla:
@AhyoungRyu @zjffdu code was reviewed by @nevillelyh, build looks green. Any other comments?

zjffdu:
I don't have more comments. \cc @Leemoonsoo @felixcheung if they have comments.

ravwojdyla:
ping @Leemoonsoo @felixcheung

felixcheung:
Thanks for the contribution and the ping.
If it hasn't been asked before, should this be part of the beam interpreter group? Since they are both for Apache Beam and perhaps they could share config/backend and so on?

ravwojdyla:
@felixcheung well - yes and no. scio is currently being contributed to beam, but current implementation still uses dataflow sdk. maybe we can consolidate it once scio migration to beam is done?

felixcheung:
Hmm... perhaps you could advise us, would it be easy to switch over when it becomes part of beam? would it, say, make sense to put this into the beam group (`%beam.scio`) to help the transition and/or share the configuration now?

ravwojdyla:
@felixcheung after going through the `BeamInterpreter`, I don't see immediate technical advantage of sharing the group, current implementation of `BeamInterpreter`, is actually pure Java interpreter, they would not share any configuration. That said from strictly logical point of view it makes sense for them to be in the same group - given they share the same model.

What kind of changes are needed to put scio in the same group as beam? And what is gonna be the impact on users?

felixcheung:
I think that's strictly cosmetic - you can see the Spark/R interpreter on how multiple interpreter classes are going to the same group.

I think this helps discoverability and longer term transition when Scio becomes part of Beam.

ravwojdyla:
@felixcheung I see, thanks for help, is 95dd9bb enough?

ravwojdyla:
ping @felixcheung

felixcheung:
ah, sorry, I added a reply but it didn't go through apparently.
I think that's good - I'm not sure if scio should be the default of the beam group (aren't there more Java users?)

Could you update example, documentation and so on to reflect this change? I can do a pass after that

ravwojdyla:
@felixcheung my reasoning for the change of default interpreter, is that using `BeamInterpreter` in current form lacks benefits of IDE and gives no advantages of Zeppelin itself (there are no display helpers, notebook env etc), it's pure java interpreter - it would make more sense to use Beam Java SDK inside scala interpreter - just for the sake of more friendly dev/notebook experience.

That said, it is not super important from my perspective which one is the default (?), so I have reverted that change.

ravwojdyla:
ping @felixcheung

felixcheung:
@ravwojdyla reminder on my notes https://github.com/apache/zeppelin/pull/1471#issuecomment-255930386 ?

ravwojdyla:
@felixcheung correct - i missed it - updated doc/examples.

pom.xml
```diff
@@ -586,7 +587,14 @@
         <module>beam</module>
       </modules>
     </profile>
-    
+
+     <profile>
+      <id>scio</id>
```
bzz: Looks like indentation is a bit off here

bzz:
So cool guys, sorry, I have totally missed out on this - it looks great and CI is green now!

Looks like every feedback has been addressed by now? 

Quick question on test and CI - it seems that there is `scio` build profile that is not part of any [TravisCI](https://github.com/apache/zeppelin/blob/master/.travis.yml#L42) build? Or is it part of the standard build and so test going to be executed on every build?

docs/interpreter/scio.md
felixcheung: actually, maybe I'm missing something, why is it `argz` and not `args`?

docs/interpreter/scio.md
```diff
@@ -0,0 +1,168 @@
...
+    <td>--runner=InProcessPipelineRunner</td>
+    <td>Scio interpreter wide arguments</td>
```
felixcheung: does Beam have documentation on runner and how to config it? could you add link to that here?

pom.xml
ravwojdyla: Fixed.

ravwojdyla:
@bzz `scio` profile was added to keep it consistent with other modules. Test/build should happen as part of standard build. Is it fine?

docs/interpreter/scio.md
ravwojdyla: @felixcheung `argz`, because there is a common pattern in Scio to use `val (sc, args) = ContextAndArgs(<args>)`, thus interpreter `args` would be overwritten by pipeline `args`. Does it answer your question?

ravwojdyla:
Rebased to solve conflict in `zeppelin-distribution/src/bin_license/LICENSE`. Are there any other comments?

bzz:
@ravwojdyla 
> Test/build should happen as part of standard build. Is it fine?

Yes, sounds fantastic, thank you for pointing it out.

Looks great to me, shall we merge it, if there is no further discussion?

felixcheung:
LGTM

felixcheung:
@ravwojdyla Could you rebase to resolve conflict so we could merge this?

ravwojdyla:
Those failures on travis seem unrelated - restarting build.

Leemoonsoo:
Looks great to me.
CI failure is unrelated (known flaky test https://issues.apache.org/jira/browse/ZEPPELIN-1623)

Merge to master if no further comments.

PR #1399, opened 2016-09-02
[ZEPPELIN-1182] Zeppelin should have Startup and Shutdown message

rajarajan-g:
### What is this PR for?
PR is for logging configuration details of Zeppelin server.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1182

### How should this be tested?
Please check the log if information on Server host, server path, context path, zeppelin version, class path is available in that file.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

Note:
    For configuration details such as  zeppelin version, server host, port & context path , these are available as part of zeppelin configuration. 

Logging information

```
 INFO [2016-09-02 14:35:25,218] ({main} ZeppelinConfiguration.java[create]:107) - Server Host: 0.0.0.0
 INFO [2016-09-02 14:35:25,218] ({main} ZeppelinConfiguration.java[create]:108) - Server Port: 8080
 INFO [2016-09-02 14:35:25,218] ({main} ZeppelinConfiguration.java[create]:109) - Context Path: /
 INFO [2016-09-02 14:35:25,224] ({main} ZeppelinConfiguration.java[create]:110) - Zeppelin Version: 0.7.0-SNAPSHOT
```

For Zeppelin class path and restart information,  I am logging it from Shell script to log file as in below paragraph, So the format of logging is different . I know this is not per standards, Please let me know if you have any suggestions
```
Zeppelin is restarted

ZEPPELIN_CLASSPATH: /home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-server/target/lib/*:/home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-zengine/target/lib/*:/home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-interpreter/target/lib/*:/home/rajarajang/Workspace/stsWorksapce/zeppelin/*::/home/rajarajang/Workspace/stsWorksapce/zeppelin/conf:/home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-interpreter/target/classes:/home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-zengine/target/classes:/home/rajarajang/Workspace/stsWorksapce/zeppelin/zeppelin-server/target/classes
```

PR #1472, opened 2016-09-29
Add message of localhost:port on script

djKooks:
### What is this PR for?

To show the localhost:port number info when running zeppelin.sh script.
### What type of PR is it?

[Improvement]
### Todos
- [ ] - Task
### What is the Jira issue?
- Not an issue.

Run ./zeppelin.sh or ./zeppelin_daemon.sh and see the message about localhost is appearing.

### DIFF
--- a/bin/zeppelin-daemon.sh
+++ b/bin/zeppelin-daemon.sh
@@ -93,6 +93,8 @@ function initialize_default_directories() {
     echo "Pid dir doesn't exist, create ${ZEPPELIN_PID_DIR}"
     $(mkdir -p "${ZEPPELIN_PID_DIR}")
   fi
+
+  echo "Browse localhost:8080 in your browser"
 }
 
 function wait_for_zeppelin_to_die() {
--- a/bin/zeppelin-daemon.sh
+++ b/bin/zeppelin-daemon.sh
@@ -96,15 +96,18 @@ function initialize_default_directories() {
 
   PORT_NUMBER="8080"
   ZEPPELIN_SITE="conf/zeppelin-site.xml"
+  ZEPPELIN_ENV="conf/zeppelin-env.sh"
   if [ -e "${ZEPPELIN_SITE}" ]; then
     PORT_NUMBER="$(xmllint --xpath 'string(//configuration/property[@name="zeppelin.server.port"]/value)' conf/zeppelin-site.xml)"
   fi
 
-  PORT_LINE="$(cat conf/zeppelin-env.sh | grep ZEPPELIN_PORT | xargs)"
 
-  if [[ ! ${#PORT_LINE} -lt 0 ]]; then
-    if [[ $PORT_LINE != \#* ]]; then
-      PORT_NUMBER="$(cat conf/zeppelin-env.sh | grep ZEPPELIN_PORT | cut -d '=' -f2)"
+  if [ -e "${ZEPPELIN_ENV}" ]; then
+    PORT_LINE="$(cat conf/zeppelin-env.sh | grep ZEPPELIN_PORT | xargs)"
+    if [[ ! ${#PORT_LINE} -lt 0 ]]; then
+      if [[ $PORT_LINE != \#* ]]; then
+        PORT_NUMBER="$(cat conf/zeppelin-env.sh | grep ZEPPELIN_PORT | cut -d '=' -f2)"
+      fi
     fi
   fi
   
--- a/bin/zeppelin-daemon.sh
+++ b/bin/zeppelin-daemon.sh
@@ -94,7 +94,8 @@ function initialize_default_directories() {
     $(mkdir -p "${ZEPPELIN_PID_DIR}")
   fi
 
-  echo "Browse localhost:8080 in your browser"
+  IP_ADDR="$(ifconfig en0 | grep inet | grep -v inet6 | cut -d ' ' -f2)"
+  echo -e "Browse $IP_ADDR:8080 in your browser.\nIf you are testing on your local computer, use localhost:8080"
 }
 
 function wait_for_zeppelin_to_die() {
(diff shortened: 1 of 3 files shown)


bin/zeppelin-daemon.sh
```diff
@@ -93,6 +93,8 @@ function initialize_default_directories() {
     echo "Pid dir doesn't exist, create ${ZEPPELIN_PID_DIR}"
     $(mkdir -p "${ZEPPELIN_PID_DIR}")
   fi
+
+  echo "Browse localhost:8080 in your browser"
```
AhyoungRyu: Hi @djKooks. Thanks for your contribution! 

What if the users are not using `localhost`?

djKooks:
@AhyoungRyu Oh, it was my mistake. Sorry for silly bugs.
I just updated script to find public IP.

bin/zeppelin-daemon.sh
```diff
@@ -94,7 +94,8 @@ function initialize_default_directories() {
     $(mkdir -p "${ZEPPELIN_PID_DIR}")
   fi
 
-  echo "Browse localhost:8080 in your browser"
+  IP_ADDR="$(ifconfig en0 | grep inet | grep -v inet6 | cut -d ' ' -f2)"
+  echo -e "Browse $IP_ADDR:8080 in your browser.\nIf you are testing on your local computer, use localhost:8080"
```
AhyoungRyu: @djKooks As you can see in [Apache Zeppelin Configuration](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/install/install.html#apache-zeppelin-configuration), user can also change the server port using `ZEPPELIN_PORT ` and `zeppelin.server.port` value.  Can we update the port as well? Sorry about this. I should've told about it with my last comment.

AhyoungRyu:
@djKooks Yeah I just saw the commit right after you pushed. It's not a silly bug :) But the port part should be updated as well. Some users  can use other port number not `8080`. Could you check again please?

bin/zeppelin-daemon.sh
djKooks: @AhyoungRyu Thanks! Which file do I have to refer for this value? `zeppelin-site.xml` or `zeppelin-env.sh` or else?

AhyoungRyu: @djKooks As you can see in the referred docs
>You can configure Apache Zeppelin with either environment variables in conf/zeppelin-env.sh (conf\zeppelin-env.cmd for Windows) or Java properties in conf/zeppelin-site.xml. If both are defined, then the environment variables will take priority. 

User can change the port using both `zeppelin-site.xml` and `zeppelin-env.sh`. If the user modify only `zeppelin-env.sh` or both, then you can easily refer `zeppelin-env.sh` file. But we should also cover the case that only `zeppelin-site.xml` is changed.

djKooks: @AhyoungRyu Okay. I'll check and update.

djKooks: @AhyoungRyu I updated script to read port num from zeppelin-env.sh, but still have problem on reading from zeppelin-site.xml. There are some xml parser but needs additional parser to be installed, and that seems really not good.
How do you think about adding id for each nodes? Like below...
```
<configuration>

<property>
  <name>zeppelin.server.addr</name>
  <value id="zeppelin.server.addr">0.0.0.0</value>
  <description>Server address</description>
</property>

<property>
  <name>zeppelin.server.port</name>
  <value id="zeppelin.server.port">>8080</value>
  <description>Server port.</description>
</property>
...
```
How do you think about it? This will cause no side effect as long as I know.
If you are not sure about it, I'll update this first so you could look through it carefully.

AhyoungRyu:
@djKooks Sorry for my late response. I didn't notice, but seems there is already similar feature merged few days ago. Could you please take a look at #1399?

djKooks:
@AhyoungRyu It looks different. The one in #1399 records ip address in log, but I want to show this during script is running. It seems ok to keep working on.

AhyoungRyu:
@djKooks Yeah you're right. I left the comment like that at the first time, I realized that it's different change after checking it again. Sorry for any confusion.

djKooks:
@AhyoungRyu No problem~thanks for guide! I'll keep work on it.

djKooks:
@AhyoungRyu Sorry for late reply.
I updated scripts to read port number from environment setups. Please look on and let me know if there is any problem on running this.

AhyoungRyu:
@djKooks Thanks for kind reminder :)

Currently Zeppelin works with "zero-configuration" status. As you know, even if users didn't activate `conf/zeppelin-env.sh.template` or `conf/zeppelin-site.xml.template`, Zeppelin works with default configurations. I think you missed this side. Can you test this patch without `zeppelin-env.sh` and `zeppelin-site.xml` again?

djKooks:
@AhyoungRyu you're right. Thanks for comments. I just update lines for file check.

jongyoul:
@djKooks Hi, Is there any issue for installing `xmllint`? I found it in Mac and Linux but I'm not sure  if it exists in all of linux and mac. Do you have any idea?

AhyoungRyu:
@djKooks Sorry for my late response. 

Tested again and it works as expected only when I'm using wireless interpreter.
If I connected wired internet, it doesn't print host info as below
```
$ ./bin/zeppelin-daemon.sh restart
Zeppelin stop                                              [  OK  ]
Browse :8080 in your browser.
If you are testing on your local computer, use localhost:8080
Zeppelin start                                             [  OK  ]
```
Can you check again this?

AhyoungRyu:
@djKooks And in my personal opinion, I wonder we need to print wireless network address in shell window since it changes depending on the user's network environment. What do you think?

djKooks:
@AhyoungRyu Actually, I was also thinking about whether this commit needs now. I'll drop this issue and maybe look it again after when it seems useful.
Thanks for help.

AhyoungRyu:
@djKooks Yeah right. Thanks for your consideration. 
And please do not hesitate if you want to contribute to Zeppelin next time. All ppl in the community will help you!

djKooks:
@AhyoungRyu Okay. Thanks~

PR #1477, opened 2016-09-30
[ZEPPELIN-1512] Support Kylin project name in interpreter runtime

yiming187:
### What is this PR for?

Currently, the parameter "project" is defined with "kylin.query.project" in properties. It's not convenience when query Kylin among different projects. May I propose introducing %kylin(project_name) at the interpreter runtime? If not set, the default project will work, otherwise, will use explicit project_name for the query request.
### What type of PR is it?

Improvement
### Todos

DONE
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1512

%kylin select ...
%kylin(new_project) select ....
### Questions:
- Does the licenses files need update?
  No.
- Is there breaking changes for older versions?
  No.
- Does this needs documentation?
  Will do later.

### DIFF
--- a/kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
+++ b/kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-
 package org.apache.zeppelin.kylin;
 
 import org.apache.commons.codec.binary.Base64;
@@ -41,7 +40,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * Kylin interpreter for Zeppelin. (http://kylin.io)
+ * Kylin interpreter for Zeppelin. (http://kylin.apache.org)
  */
 public class KylinInterpreter extends Interpreter {
   Logger logger = LoggerFactory.getLogger(KylinInterpreter.class);
@@ -59,6 +58,7 @@ public class KylinInterpreter extends Interpreter {
   public KylinInterpreter(Properties property) {
     super(property);
   }
+
   @Override
   public void open() {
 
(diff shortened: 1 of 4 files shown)


jongyoul:
Thanks for the contribution. I'll start review this PR.

kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
```diff
@@ -148,6 +146,22 @@ public HttpResponse prepareRequest(String sql) throws IOException {
     return httpClient.execute(postRequest);
   }
 
+  public String getProject(String cmd) {
```
jongyoul: Hi, this kind of style is used by JDBCInterpreter, but I think it's too tricky and removed from JDBCInterpreter for now. And we started to discuss on how to support this kind of property - that we called as note level properties -. I don't think your implementation makes a big problem but it can confuse existing users.

jongyoul:
I've no preference about adding some properties in a paragraph but you need to update docs for that feature in details.

kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
yiming187: Thanks Jongyoul for the review. You are right, I check the code from JDBCInterpreter and copy the similar function. Could you point me some note level properties examples?

yiming187:
Hi  @jongyoul , the doc is updated. Could you help review?

bzz:
@yiming187 thank you for improvement, looks good to me, 👍  for having tests!

Could you please rebase it on the latest master, before merging it in?

yiming187:
Thanks @bzz , the rebase is done.

bzz:
Thank you @yiming187 !

 CI failures looks not relevant to the changes

```
Results :

Failed tests: 
  HeliumApplicationFactoryTest.testLoadRunUnloadApplication:160 expected:<Hello world 1> but was:<null>

Tests run: 122, Failures: 1, Errors: 0, Skipped: 0
```

and

```
Results :

Tests in error: 
  ParagraphActionsIT.testRemoveButton:155->AbstractZeppelinIT.clickAndWait:131 » ElementNotVisible
  ParagraphActionsIT.testMoveUpAndDown:200->AbstractZeppelinIT.clickAndWait:131 » ElementNotVisible
  ParagraphActionsIT.testCreateNewButton:85 » ElementNotVisible Element is not c...

Tests run: 18, Failures: 0, Errors: 3, Skipped: 0
```

Merging to master if there is no further discussion

yiming187:
Thanks @bzz , there is no more commit from my side.

PR #1479, opened 2016-10-02
Notebook: HDFS as a backend storage (Read & Write Mode)

hayssams:
### What is this PR for?

Zeppelin currently supports many backends for storing notes through Apache Commons VFS. 
Apache Commons VFS supports HDFS in readonly mode. 
This PR makes HDFS a first class citizen by allowing users to load notes from / save notes to HDFS.
### What type of PR is it?

Improvement
### Todos
- [ ] - Task
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1515

Update zeppelin.notebook.dir property to a value like hdfs://localhost:9000/tmp/notebook and the property zeppelin.notebook.storage to the value org.apache.zeppelin.notebook.repo.HdfsNotebookRepo

check that your notes are loaded from and stored to HDFS by listing notes using the command :
hdfs dfs -ls /tmp/notebook
### Questions:
- Does this needs documentation? Yes


Leemoonsoo:
Thanks for the contribution.

How about let user set not only `HADOOP_CONF_DIR` but also `HADOOP_HOME` in conf/zeppelin-env.sh to use HDFS notebook storage?

Then hadoop dependencies in zeppelin-zengine can be provided scope, and that'll reduce binary package size. Also that'll allow user set different hadoop version without rebuild.

what do you think?

hayssams:
@Leemoonsoo  Just pushed the changes to make hadoop lib provided.

bin/zeppelin.sh
```diff
@@ -70,6 +70,7 @@ addJarInDir "${ZEPPELIN_HOME}/zeppelin-zengine/target/lib"
 addJarInDir "${ZEPPELIN_HOME}/zeppelin-server/target/lib"
 addJarInDir "${ZEPPELIN_HOME}/zeppelin-web/target/lib"
 
+
 CLASSPATH+=":${ZEPPELIN_CLASSPATH}"
```
zjffdu: unecessary change

pom.xml
```diff
@@ -91,6 +91,7 @@
     <gson.version>2.2</gson.version>
     <guava.version>15.0</guava.version>
     <jetty.version>9.2.15.v20160210</jetty.version>
+    <hadoop.hdfs.version>2.6.0</hadoop.hdfs.version>
```
zjffdu: how about using `hadoop.version` ? Since in spark-dependencies/pom.xml, we also define hadoop.verison. We can reuse the same property name. Otherwise user need to specify both `hadoop.version` and `hadoop.hdfs.version` when building with custom hadoop.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.Date;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ *
+ */
+
+public class HdfsNotebookRepo implements NotebookRepo {
+  private Logger logger = LoggerFactory.getLogger(HdfsNotebookRepo.class);
```
zjffdu: logger can be static

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,203 @@
...
+  private HdfsUtils hdfsUtils;
+  private ZeppelinConfiguration conf;
+  private Path rootDir;
+
+
+  public HdfsNotebookRepo(ZeppelinConfiguration conf) throws IOException {
+    this.conf = conf;
+    String hadoopConfDir = System.getenv("HADOOP_CONF_DIR");
+    try {
```
zjffdu: Is the right exception caught when HADOOP_CONF_DIR is not defined ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/HdfsUtils.java
```diff
@@ -0,0 +1,205 @@
...
+  protected String dataPath;
+  private Logger logger = LoggerFactory.getLogger(HdfsUtils.class);
+
+  /**
+   * @param dataPath Full hdfs path (including scheme) to notes root dir
+   * @throws URISyntaxException
+   */
+  public HdfsUtils(String dataPath, String hadoopConfDir) throws URISyntaxException {
+    logger.info("hadoopConfDir:" + hadoopConfDir);
+    if (hadoopConfDir != null && !hadoopConfDir.equals("")) {
```
zjffdu: Can be replaced with StringUtils.isBlank

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/HdfsUtils.java
```diff
@@ -0,0 +1,205 @@
...
+import java.net.URISyntaxException;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Utility class to create / move / delete / write to / read from HDFS filessytem
+ */
+public class HdfsUtils {
+  protected String dataPath;
+  private Logger logger = LoggerFactory.getLogger(HdfsUtils.class);
```
zjffdu: logger can be static

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,203 @@
...
+
+  @Override
+  public void close() {
+    //no-op
+  }
+
+  @Override
+  public Revision checkpoint(String noteId, String checkpointMsg, AuthenticationInfo subject)
+      throws IOException {
+    // Auto-generated method stub
```
zjffdu: these kinds of code can be removed.

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/HdfsUtils.java
```diff
@@ -0,0 +1,205 @@
...
+import java.io.*;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Utility class to create / move / delete / write to / read from HDFS filessytem
+ */
+public class HdfsUtils {
```
zjffdu: *Utils class usually provides static method.  Either you can change all its method to static or rename it to a more proper name

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
hayssams: Null value is valid for HADOOP_CONF_DIR. Hdfs API will  default to the file:// protocol instead of hdfs ://

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
hayssams: Won't compile if note present

hayssams:
@Leemoonsoo 
Test testClose in org.apache.zeppelin.python.PythonInterpreterTest. Can't understand why it is failing on travis CI but not failing on my laptop.

jongyoul:
@hayssams Why did you removed `provided` scope? If it doesn't, all kind of hadoop dependencies will be included in zengine.

hayssams:
@jongyoul 
When using provided and bringing in all hadoop dependencies to zengine by adding hadoop lib in HADOOP_CONF_DIR to the classpath, Zeppelin raise an exception at runtime.

However by bringing explicitely in scope to the zengine only hadoop-hdfs and hadoop-common, everything works as intended.

jongyoul:
@hayssams AFAIK, those hadoop-* have a lot of dependencies by themselves and it's possible to make a problem like version mismatch even in case that we don't use that feature. What do you think of it?

zjffdu:
Agree with @jongyoul we should make hadoop dependency as provided, @hayssams  Could you paste the errors you can see when using provided ? So that someone else can help that.

hayssams:
@jongyoul @zjffdu 
This issue is with hadoop 2.7 only. With hadoop 2.6 making hadoop dependency as provided works well.


I get the following exception
Caused by: java.lang.NoSuchMethodError: org.apache.hadoop.tracing.SpanReceiverHost.get(Lorg/apache/hadoop/conf/Configuration;Ljava/lang/String;)Lorg/apache/hadoop/tracing/SpanReceiverHost;
        at org.apache.hadoop.hdfs.DFSClient.<init>(DFSClient.java:634)
        at org.apache.hadoop.hdfs.DFSClient.<init>(DFSClient.java:619)
        at org.apache.hadoop.hdfs.DistributedFileSystem.initialize(DistributedFileSystem.java:149)
        at org.apache.hadoop.fs.FileSystem.createFileSystem(FileSystem.java:2596)
        at org.apache.hadoop.fs.FileSystem.access$200(FileSystem.java:91)
        at org.apache.hadoop.fs.FileSystem$Cache.getInternal(FileSystem.java:2630)
        at org.apache.hadoop.fs.FileSystem$Cache.get(FileSystem.java:2612)
        at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:370)
        at org.apache.hadoop.fs.FileSystem.get(FileSystem.java:169)
        at org.apache.zeppelin.util.HdfsSite.mkdirs(HdfsSite.java:98)

jongyoul:
@hayssams Do you have any clue that you mentioned? I don't know why two different jars while building Zeppelin.

aspen01:
You know, there is WebHDFS Rest API in hadoop.
If we use this, we don’t need to care about library dependency and configuration.
What do you think using WebHDFS Rest API?

HDFS interpreter already use it, so I added HDFSNotebookRepo using this function.
Please check this code : https://github.com/aspen01/zeppelin/commit/4b57a2faefdd34f4675e1acd78b25677111a5e57

aspen01:
I made a pull request : https://github.com/apache/zeppelin/pull/1600

hayssams:
Replaced by PR #2333

PR #1096, opened 2016-06-27
ZEPPELIN-1030 Added example for various jdbc-like interpreters

jongyoul:
### What is this PR for?
Adding examples settings for JdbcInterpreter to use jdbc drivers. This hasn't changed actual codes

### What type of PR is it?
[Documentation]

### Todos
* [x] - Add example

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1030

### How should this be tested?
1. mvn clean package -DskipTests -Pspark-1.6 -pl 'jdbc,zeppelin-interpreter,zeppelin-web,zeppelin-server,zeppelin-zengine,zeppelin-display'
2. Test it

### Screenshots (if appropriate)
screen shot 2016-06-27 at 11 46 00 pm
screen shot 2016-06-27 at 11 47 59 pm

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1480, opened 2016-10-03
[ZEPPELIN-1489] Remove jdbc-like interpreter example properties and improve JDBC docs

AhyoungRyu:
### What is this PR for?
Currently we can use `%jdbc(prefix)` for multi database connections(e.g. configuring both postgresql and hive in one JDBC interpreter). But after #1360 merged, Zeppelin doesn't support `%jdbc(prefix)` to `%prefix` anymore. So auto-prefix feature is not working for `%jdbc(prefix)`. The auto-prefix only works for `%prefix` now. Considering this status, it would be better we remove the JDBC connection examples in `jdbc/interpreter-setting.json` (this is come from #1096) so that users can create several JDBC interpreters instead of using multiple connections with one JDBC interpreter.  I removed the related contents in `jdbc.md` as well.

Also, the contents of [current JDBC documentation page](http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/interpreter/jdbc.html) is quite confusing and unorganized. So I updated the contents with some screenshot images to guide "How to create JDBC interpreter", "How to edit the interpreter properties for the connection", "How to use `%prefix` with the interpreter", and so on.

### What type of PR is it?
Improvement & Documentation

### What is the Jira issue?
[ZEPPELIN-1489](https://issues.apache.org/jira/browse/ZEPPELIN-1489)

* Removing example properties in `interpreter-setting.json`
  1. after applying this patch and build with `mvn clean package -DskipTests -Pspark-1.6 -pl 'jdbc,zeppelin-interpreter,zeppelin-web,zeppelin-server,zeppelin-zengine,zeppelin-display'`
  2. create JDBC interpreter and check whether the example settings are gone or not  

* JDBC docs
  1. Build only `docs/` dir as described in [here](https://github.com/apache/zeppelin/blob/master/docs/README.md#build-documentation)
  2. Go to `interpreter -> JDBC` and just compare this locally builded page with [the original JDBC page](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/interpreter/jdbc.html)

 - Before 
before

 - After 
screen shot 2016-10-04 at 1 05 00 am

 - Some parts of updated JDBC docs
Since many contents are changed, it would be better to build `docs/` locally to review all of the change. 

screen shot 2016-10-04 at 12 36 39 am

screen shot 2016-10-04 at 12 36 59 am

screen shot 2016-10-04 at 12 37 31 am

screen shot 2016-10-04 at 12 37 43 am

### Questions:


Leemoonsoo:
@AhyoungRyu Great work! It definitely helps using JDBC datasource in Zeppelin.
CI failure is not related.

LGTM and merge if there're no further comments.

AhyoungRyu:
@Leemoonsoo Thanks!

PR #1486, opened 2016-10-05
[ZEPPELIN-1518] Fix for: Lambda expressions are not working on CDH 5.7.x Spark

mvitaly:
### What is this PR for?
Lambda expressions are not working on CDH 5.7.x Spark because of the backported RpcEnv and elimination of class server.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1518

The following paragraph should work with SPARK_HOME pointing to CDH 5.7x spark:
`val rdd = sc.parallelize(Seq(1,2,3,4,5))
rdd.filter(_ > 3).count()`

### Questions:
* Does this needs documentation? n

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -381,6 +381,7 @@ public class SparkInterpreter extends Interpreter {
     }
 
     String classServerUri = null;
+    String replClassOutputDirectory = null;
 
     try { // in case of spark 1.1x, spark 1.2x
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @mvitaly for the contribution!

Looks good to me!

minahlee:
Tested, it fixes the bug. Btw do we need this patch only in branch-0.6 not to master?

Leemoonsoo:
I think master also needs this patch.

minahlee:
Merging to master and branch-0.6 if there is no more discussion

PR #1487, opened 2016-10-05
[ZEPPELIN-1520] Change test environment to use the local mode of Apache Spark

jongyoul:
### What is this PR for?
Simplify travis test to reduce usage of resources

### What type of PR is it?
[Improvement]

### Todos
* [x] - Remove start-up/stop SparkCluster

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1520

Travis will pass without any error

### Questions:


jongyoul:
Ready to review

jongyoul:

felixcheung:
LGTM

zjffdu:
Should we provide the option to allow user to use other modes ? Otherwise it seems the scripts under folder `test` can be removed as well.

zjffdu:
And think again, how about making the local mode as the default mode for local dev (non-travis enviroment), and use standalone mode for travis ? Because local mode success doesn't mean other modes also succeed, there are subtle differences between them.  For now, user can set enviroment variable `SPARK_MASTER` to control which mode to use. We can just export `SPARK_MASTER` to local for local dev.  What do you think ? @jongyoul @felixcheung

jongyoul:
@zjffdu I've missed remove some scripts in `testing`. I'll fix it. Concerning the modes, in an early stage of Spark, there're some different behaviours of different modes, thus we need to test Zeppelin with Spark in a standalone cluster, but for now, Spark become stable and if different modes have different behaviours, it's definitely Spark's problem and Spark should fix it. Thus I don't think we need to test Zeppelin a cluster mode. Furthermore, personally, I don't know what benefit we have in a different test environment between local and CI. In case of Python, I've experienced that some tests only failed in CI, not in a local with different setting between them. I think we should try to keep both environments equally.

PR #1488, opened 2016-10-05
[ZEPPELIN-1519] Fix notebook importer AM/PM omission problem

sixmen:
### What is this PR for?
When loading note from storage (or importing note), AM/PM informations are gone. So dates are always in AM.
Fix this problem.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1519

Do some changes on sample notebooks, and see file changes.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookImportDeserializer.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookImportDeserializer.java
@@ -35,7 +35,8 @@ import java.util.Locale;
 public class NotebookImportDeserializer implements JsonDeserializer<Date> {
   private static final String[] DATE_FORMATS = new String[] {
     "yyyy-MM-dd'T'HH:mm:ssZ",
-    "MMM d, yyyy h:mm:ss a"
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookImportDeserializer.java
```diff
@@ -35,7 +35,7 @@
 public class NotebookImportDeserializer implements JsonDeserializer<Date> {
   private static final String[] DATE_FORMATS = new String[] {
     "yyyy-MM-dd'T'HH:mm:ssZ",
-    "MMM dd, yyyy HH:mm:ss"
+    "MMM d, yyyy h:mm:ss a"
```
Leemoonsoo: Thanks for the contribution.

Because of deserialize() tries all formats in `DATE_FORMATS` sequentially, i think we can keep both formats. "MMM d, yyyy h:mm:ss a" and then "MMM dd, yyyy HH:mm:ss".

Leemoonsoo:
Thanks. LGTM and merge if there're no more discussions.

PR #1490, opened 2016-10-05
[ZEPPELIN-1480] Blocking message pending 10000 for BLOCKING

Leemoonsoo:
### What is this PR for?
This patch try to address problem described in ZEPPELIN-1480

### What type of PR is it?
Bug Fix

### Todos
* [x] - Make websocket send thread safe

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1480

Create multiple paragraphs (for example 10 ```%sh date``` paragraphs) and schedule it every 10sec `0/10 * * * * ?`

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
@@ -65,7 +65,7 @@ public class NotebookSocket extends WebSocketAdapter {
     return protocol;
   }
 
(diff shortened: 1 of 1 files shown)


corneadoug:
@Leemoonsoo That Jira issue is in Korean, could you update the PR description so that there is more details on what the problem is and what is the expected result of that test before and after the patch?

Leemoonsoo:
@corneadoug i commented translation on the issue.

Leemoonsoo:
Merge to master and branch-0.6 if there're no more discussions.

PR #1493, opened 2016-10-07
ZEPPELIN-1472 - Create new LdapRealm based on Apache Knox LdapRealm

gss2002:
### What is this PR for?
Provides LdapRealm Functionality similar to what Apache Knox provides. This is critical as in large enterprise environments Active Directory Global Catalogs are used for lookup with samAccountName and using a DN Template is not an option as their are multiple OUs. Also searching on "userPrincipalName" is risky in an AD environment since the explicit UPN vs Implicit UPN can be different this is definitely the case with environments using Office 365. And the LDAP userPrincipalName attribute is the explicit UPN which can be defined by the directory administrator to any value and it can be duplicated. SamAccountName is unique per domain and Microsoft states best practice is to not allow duplicate samAccountName's across the forest. 

Information about samAccountName and userPrincipalName with ActiveDirectory
http://windowsitpro.com/active-directory/q-does-samaccountname-object-have-be-unique-active-directory-domain-or-entire-fores
https://jorgequestforknowledge.wordpress.com/2010/10/12/user-principal-names-in-ad-part-1/
 
### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1472

shiro.ini
[main]
ldapRealm = org.apache.zeppelin.server.LdapRealm
ldapRealm.contextFactory.systemUsername = CN=hdpbind,OU=Svc,DC=exadc,DC=w2k,DC=example,DC=com
ldapRealm.contextFactory.systemPassword = ldapPassword
ldapRealm.searchBase = dc=w2k,dc=example,dc=com
ldapRealm.userSearchBase = dc=w2k,dc=example,dc=com
ldapRealm.groupSearchBase = dc=w2k,dc=example,dc=com
ldapRealm.contextFactory.url = ldap://exampledc1.exadc.w2k.example.com:3268
ldapRealm.userSearchAttributeName = sAMAccountName
ldapRealm.contextFactory.authenticationMechanism = simple
ldapRealm.userObjectClass = user
ldapRealm.groupObjectClass = group
ldapRealm.memberAttribute = member
securityManager.realms = $ldapRealm

### Questions:


gss2002:
@vinayshukla here is the pull request for the LdapRealm enhancement we discussed last week at HadoopWorld/Strata

khalidhuseynov:
i believe there's also `org.apache.zeppelin.realm` package under `zeppelin` and in the long run maybe need to move these realms there rather than keeping under `server`

gss2002:
@khalidhuseynov 
I'm going to make the changes and move this under realms. I will also close this pull request and open a new one

bhavintandel:
Do we have something like ldapADGCRealm.userLowerCase which will force username to change the case to lower. Currently if i set user permission to "user1" and then i login with "User1", i cannot see the notebooks.

PR #1494, opened 2016-10-07
[ZEPPELIN-451] Save codes and messages as multi-line

sixmen:
### What is this PR for?

Save Paragraph texts and result msgs as array of strings instead of one long string.
One long strings are not VCS-friendly.
### What type of PR is it?

Improvement
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-451

Modify notebooks and check saved note files
### Questions:
- Does the licenses files need update? NO
- Is there breaking changes for older versions? Notebooks from new zeppelin can not be imported in old zeppelin.
- Does this needs documentation? NO


anthonycorbacho:
@sixmen thank you for your contribution, this feature looks AWESOME.
I didnt dive into the code, but one question came in my mind;

This change is kinda *breaking change* since the notebook format is now different from what we have. So what can be the best way in your opinion to support *old* notebook format?

sixmen:
It supports old notebook format.
Old code can not read new notebook format, but it can not be helped any way.
So I think it is an endurable breaking change.

sixmen:
I worked again for this commit with recent sources.
I hope someone reviews this pull request.
As I know, there are already breaking changes in the notebook format. So it does not matter with this PR.

PR #1497, opened 2016-10-08
[ZEPPELIN-1233]  Support Redis interpreter

darionyaphet:
### What is this PR for?

### What type of PR is it?

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

Support Redis interpreter

What type of PR is it?

New Feature

What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1233


redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,98 @@
...
+import static org.apache.zeppelin.interpreter.InterpreterResult.Code.ERROR;
+import static org.apache.zeppelin.interpreter.InterpreterResult.Code.SUCCESS;
+
+public class RedisInterpreter extends Interpreter {
+
+    private Logger LOG = LoggerFactory.getLogger(RedisInterpreter.class);
+
+    private Jedis client;
+
+    private enum RedisCommand {
```
anthonycorbacho: As a redis user, i am expecting to use other redis command like `hkeys`, range and stuff

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,98 @@
...
+        super(property);
+
+        for (RedisCommand command : RedisCommand.values()) {
+            keywords.add(command.name());
+        }
+    }
+
+    @Override
+    public void open() {
+        client = new Jedis();
```
anthonycorbacho: I think here we should provide a better Redis client creation, as far as i know, this will assume that you are running redis locally, i think we can exploit more the interpreter setting creation, here an [example](https://github.com/apache/zeppelin/blob/master/jdbc/src/main/resources/interpreter-setting.json) with JDBC.
So user will be able to set host, port and credential.

Also how about using `JedisPool` instead of `Jedis`?

It will look like something like this
```
private static final int MAX_CONNECTION = 142;
private static final int MAX_IDLE = 10;
private static final int MIN_IDLE = 142;
private static final long MAX_WAIT = TimeUnit.SECONDS.toMillis(30);

private JedisPool jedisPool;

public void open() {
  jedisPool = new JedisPool(getPoolConfiguration(), property.getCacheHost(), property.getCachePort());
}

private JedisPoolConfig getPoolConfiguration() {
  JedisPoolConfig config = new JedisPoolConfig();
  config.setMaxTotal(MAX_CONNECTION);
  config.setMaxIdle(MAX_IDLE);
  config.setMinIdle(MIN_IDLE);
  config.setMaxWaitMillis(MAX_WAIT);

  return config;
}

// Utils method :)
private Jedis borrow() {
  return jedisPool.getResource();
}

// Then you can use it like
private InterpreterResult getProcess(String key) {
  Jedis jedis = borrow();
  Pipeline pipe = jedis.pipelined();
  Response<String> response = pipe.get(key);
  pipe.sync();
  jedis.close();

  String result = response.get();
  return new InterpreterResult(SUCCESS, result);
}

```

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,98 @@
+package org.apache.zeppelin.redis;
```
anthonycorbacho: ASF Header is missing here.

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,98 @@
...
+import redis.clients.jedis.Jedis;
+
+import java.util.HashSet;
+import java.util.Properties;
+import java.util.Set;
+
+import static org.apache.zeppelin.interpreter.InterpreterResult.Code.ERROR;
+import static org.apache.zeppelin.interpreter.InterpreterResult.Code.SUCCESS;
+
+public class RedisInterpreter extends Interpreter {
```
anthonycorbacho: Could you add Clean header with a description of the interpreter? that will be useful for futur commiters to understand the purpose of the class

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private Logger LOG = LoggerFactory.getLogger(RedisInterpreter.class);
```
anthonycorbacho: Can we change this field with `private static final Logger LOG = LoggerFactory.getLogger(RedisInterpreter.class);`?

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
darionyaphet: Actually I'm not sure is there any good way to  running redis command so I just simply create  `GET` and `SET` :)

darionyaphet: maybe I can use redis.clients.jedis.Protocol.Command to replacement `RedisCommand `

anthonycorbacho: yes, i think its better to use this `enum` will be enough.

i was thinking about doing something like:

```
private HashSet<String> redisCommands;

public static HashSet<String> getRedisCommands() {
  HashSet<String> hashSetOfRedisCommands = new HashSet<String>();

  for (Command c : Command.values()) {
      hashSetOfRedisCommands.add(c.name());
  }

  return hashSetOfRedisCommands;
}

// then use this like 
redisCommands.contains(tokens[1]);

```

what do you think?

darionyaphet: Maybe we don't need a redis commands set to check whether the input command is illegal because there is a default case in `switch` . Is it a better one? 

```
switch (Protocol.Command.valueOf(keyword)) {
            case GET:
                result = getProcess();
                break;
            case SET:
                result = setProcess();
                break;
            default:
                result = new InterpreterResult(ERROR, "key word not found");
                break;
        }
```

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
darionyaphet: I read Jedis source code and find a strange case . The `returnResource` will be deprecated after `3.0` and the resource shoule be close after usage (don't return into source pool ) .  Now I'm using `2.9` . I'm not sure it's a good idea .

felixcheung:
Hi - thank you for your contribution. Could you see the list of steps here:
https://zeppelin.apache.org/docs/0.6.1/development/writingzeppelininterpreter.html#contributing-a-new-interpreter-to-zeppelin-releases

Notably, could you add:
- tests
- documentation
- list licenses for dependencies

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
anthonycorbacho: I think its better to use a Pool for this job, you only need jedis instance when you actually run the jobs, that is whay i am asking for changing from `Redis` to `JedisPool`, and in the doc i didnt find the notice about removing `returnResource`.

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
anthonycorbacho: Yes it look better :)

anthonycorbacho:
@darionyaphet do you need some help for the change?

darionyaphet:
Hi @anthonycorbacho  thank you for you great help and I'm still working on it .

I have implement a lot of command include `GEO` ,`HASH`, `HYPERLOGLOG`,  `LIST`, `KEYS`, `Sets` ,`SORTED SET` and `Strings` and have updated .

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,1986 @@
...
+    clientPool = new JedisPool(config, host, port);
+  }
+
+  @Override
+  public InterpreterResult interpret(String input, InterpreterContext context) {
+    String[] tokens = splitTokens(input);
+    String keyword = tokens[0];
+
+    InterpreterResult result;
+    switch (Protocol.Command.valueOf(keyword)) {
```
anthonycorbacho: Humm, that is a very massive switch case, i think we can come up with a more elegant way to to this kind of things,

since it mainly take the same argument `String[] token`, i guess we can go with the design pattern  [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern).

This idea here is to extract the logic to smaller classes and register then to a map.

eg:
```
// In the constructor
Map<Protocol.Command, Command> redisCommand = Maps.newHashMap<>();
// register command and class
redisCommand.put(GEOADD, GeoAdd.createInstance());
... // add other command

// create a java package org.apache.zeppelin.redis.cmd and add every implementation of redis cmd.
// but first you need to create a interface
// eg:
public interface Cmd {
  public boolean execute(String[] tokens);
}

// use this interface to implement the actual cmd
public class MyRedisCmd implements Cmd {
  public static MyRedisCmd createInstance() {
    return new MyredisCmd();
  }

  @Override
  public boolean execute(String[] tokens) {
   // do the work here
  }
}

// Once you have implemented the redis cmd, you can use this like this:
redisCommand.get(Protocol.Command.valueOf(keyword)).execute(tokens);

```
let me know what you think

kavinkumarks:
I too checked the changes and they look good.The comment from @anthonycorbacho about usage of "Strategy pattern" for selecting the relevant redis command instance and executing them is the better way to implement it.

Also, could we add some test cases for this?

Thanks,
Kavin
MailTo: kavin.kumar@imaginea.com

HeartSaVioR:
Every command methods have a chance to leak Jedis instance when command is failing and throwing a kind of JedisException.
You need to replace your borrow-call-return pattern with `try-with-resource` thanks to Zeppelin requires Java 1.7.

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
anthonycorbacho: Personally, I would prefer to have classes (will make test more easy) than a massive switch case. the main concern i have about big switch case tis that it is really hard to maintain and make the parent class huge. by de coupling the logic into smaller classes will improve the maintainability and make testability more easy.

anthonycorbacho:
@HeartSaVioR thank you for pointing that out, you are right we better handle the potential `JedisException` exception to avoid leak.

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
darionyaphet: @anthonycorbacho  @HeartSaVioR  It's seems a good idea to split execute Redis Command into classes and this command classes will reused in  Redis Cluster Interpreter. I have create a abstract class `RedisCommand` as a basic class : 

```
public abstract class RedisCommand {

    private JedisPool clientPool;

    public RedisCommand(JedisPool clientPool) {
        this.clientPool = clientPool;
    }

    private Jedis borrowClient() {
        return clientPool.getResource();
    }

    private void releaseClient(Jedis client) {
        client.close();
    }

    public InterpreterResult execute(String[] tokens) {
        Jedis client = borrowClient();
        try {
            return execute(client, tokens);
        } finally {
            releaseClient(client);
        }
    }

    /**
     * @param tokens
     * @return
     * @throws Exception
     */
    protected abstract InterpreterResult execute(Jedis client, String[] tokens);

}
```

I think it's OK and each redis command could extend this one .

darionyaphet:
Hi  @anthonycorbacho  @HeartSaVioR  I have update `RedisInterpreter` with some redis commands (include APPEND ,BITCOUNT ,BITFIELD ,BITOP, BITPOS, DECR) . 

Please help review this one . If anything is OK , I will clean the commented code and completion other redis commands . Thanks ~~ :)

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,455 @@
...
+                }
+                start = index + 1;
+            }
+        }
+
+        list.add(input.substring(start, input.length()));
+        return (String[]) list.toArray();
+    }
+
+//    ////////////////////////// KEYS //////////////////////////
```
anthonycorbacho: Better to remove comment here if there are not intentions of using the code below.

redis/src/main/java/org/apache/zeppelin/redis/command/RedisCommand.java
```diff
@@ -0,0 +1,65 @@
...
+    private void releaseClient(Jedis client) {
+        client.close();
+    }
+
+    public InterpreterResult execute(String[] tokens) {
+        Jedis client = borrowClient();
+        try {
+            return new InterpreterResult(SUCCESS, execute(client, tokens));
+        } catch (JedisException exception) {
+            return new InterpreterResult(ERROR, exception.getMessage());
```
anthonycorbacho: What do you think about adding a `LOG.error` here? something like
`LOG.error("Couldn't execute redis command {}", token[0], e);`

redis/src/main/resources/interpreter-setting.json
```diff
@@ -0,0 +1,21 @@
...
+      },
+      "redis.port": {
+        "envName": null,
+        "propertyName": "redis.port",
+        "defaultValue": 6379,
+        "description": "redis server port"
+      }
+    }
+  }
+]
```
anthonycorbacho: Need a new line here :)

redis/src/test/java/org/apache/zeppelin/redis/RedisInterpreterTest.java
```diff
@@ -0,0 +1,41 @@
...
+    client = new Jedis("127.0.0.1", 6379);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    if (client != null) {
+      client.close();
+    }
+  }
+}
```
anthonycorbacho: Need a new line here :)

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
```diff
@@ -0,0 +1,455 @@
...
+    public InterpreterResult interpret(String input, InterpreterContext context) {
+        String[] tokens = splitTokens(input);
+        String keyword = tokens[0];
+
+        InterpreterResult result;
+        Protocol.Command command = Protocol.Command.valueOf(keyword);
+        if (COMMANDS.containsKey(command)) {
+            result = COMMANDS.get(command).execute(tokens);
+        } else {
+            result = new InterpreterResult(ERROR, "keyword not found");
```
anthonycorbacho: Maybe we can provide a better error to the user like
`Keyword " + tokens + " is not supportted`?

What do you think?

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
darionyaphet: Hi I still working on `Key` commands . I will clean the unnecessary code after finished this PR :)

redis/src/main/java/org/apache/zeppelin/redis/RedisInterpreter.java
darionyaphet: Yep It's a good idea 👍

redis/src/main/java/org/apache/zeppelin/redis/command/list/BLPOPCommand.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class BLPOPCommand extends RedisCommand {
+
+    public BLPOPCommand(JedisPool clientPool) {
+        super(clientPool);
+    }
+
+    @Override
+    protected String execute(Jedis client, String[] tokens) {
+        String[] keys = new String[tokens.length - 2];
```
HeartSaVioR: Please extract this logic to method of utility class or method of RedisCommand. This is duplicated from many places. The logics which manipulate keys/parameters are all candidates to extract.

anthonycorbacho:
@HeartSaVioR yeah i would love to have a class that manipulate redis cmd and handle null, ussually i am not a huge fan of returning `null`, i would prefer to return an `EMPTY` type instead.

redis/src/main/java/org/apache/zeppelin/redis/command/geo/GEOADDCommand.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class GEOADDCommand extends RedisCommand {
+
+    public GEOADDCommand(JedisPool clientPool) {
+        super(clientPool);
+    }
+
+    @Override
+    protected String execute(Jedis client, String[] tokens) {
+        String result = client.geoadd(tokens[1], Double.valueOf(tokens[2]),
```
darionyaphet: `toString` maybe better and reasonable .

darionyaphet:
Aha. A lot of files conflict and I can't resolve that . Maybe I will close this one and reopen another .

HeartSaVioR:
Follow this:
1. create a new branch which is based on current master
2. cherry-pick your commits (write down your last commit id)
3. change branch to PR's branch
4. git reset --hard `last commit id`
5. git push --force

`last commit id` might be replaced with branch name you created, but I didn't try it.

darionyaphet:
Add a Screenshot 
1539610b-0afe-4cd0-9a3a-abb208adfb7d

PR #1499, opened 2016-10-10
[ZEPPELIN-1528] Add progress information in addition to progress percentage

sadikovi:
### What is this PR for?

PR adds progress information in addition to progress percentage, allowing to provide links to Spark UI, for example. Currently done by adding `getProgressInfo()` method to the interpreter API that returns list of `InterpreterProgressInfo`, similar to completion API in this sense.
Main changes are here:
- `zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java`
- `zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java`
- `zeppelin-interpreter/src/main/thrift/RemoteInterpreterService.thrift`
- `spark/src/main/scala/org/apache/zeppelin/spark/SparkJobDetailsListener.scala`
- `zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js`
- `zeppelin-web/src/app/notebook/paragraph/paragraph-progressInfo.html`
### What type of PR is it?

Feature
### Todos
- [x] - Add unittests for `SparkJobDetailsListener`
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1528

Currently tested manually, by building distribution and launching simple Spark jobs in paragraph. 
Also inspecting that progress info is updated as jobs get executed and final result is returned correctly in stdout in paragraph (initially had problems with this since there is a difference in displaying between 0.6.x and 0.7.x). 

screenshot from 2016-10-10 17 38 02
### Questions:
- Does the licenses files need update? No.
- Is there breaking changes for older versions? Possibly.
- Does this needs documentation? Possibly.


sadikovi:
@Leemoonsoo @AhyoungRyu @felixcheung could you have a look at this PR? I am not sure about how good overall implementation is, and whether or not you guys want this feature in Zeppelin. Would appreciate the feedback, thanks.

felixcheung:
Great! Thanks for your contribution. Looks like a big change and I'll take a look shortly.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/LazyOpenInterpreter.java
```diff
@@ -112,6 +113,12 @@ public int getProgress(InterpreterContext context) {
   }
 
   @Override
+  public List<InterpreterProgressInfo> getProgressInfo(InterpreterContext context) {
+    open();
```
sadikovi: To initialize interpreter, if not initialized, similar to `getProgress()` and `completion()`.

jongyoul:
@sadikovi Thanks for the contribution. I have a question of the design. This looks focused SparkInterpreter only. In case of another interpreters doesn't have any progress info that you implemented. Do you have any idea to expand this api to adopt all other interpreters?

sadikovi:
@jongyoul yes, you are right, it was built mainly for Spark, and I have taken a naive approach of showing progress info only when non-empty array is returned, otherwise you will not see progress info link. By default, method `getProgressInfo()` returns empty list, so all interpreters will not show anything and will evaluate result as they normally do. Spark interpreter only shows link when job is being executed, otherwise it is no-op in backend and UI.

I feel like my approach is not particularly elegant. Do you have any suggestions on what way it would be better to implement this feature? Should I add flag that disables/enables it globally?

PR #1501, opened 2016-10-10
[ZEPPELIN-1532] Remove map visualization from source tree, because of incompatible license

Leemoonsoo:
### What is this PR for?
According to https://developers.arcgis.com/terms/ and and http://www.esri.com/legal/software-license, current map visualization depends on an online service that has incompatible restrictions to Apache 2 license. Please see https://github.com/apache/zeppelin/pull/1210#issuecomment-252484382

Possible alternative way is explained [here](https://github.com/apache/zeppelin/pull/1210#issuecomment-252763598).

I'm very sad to remove this very good contribution. So please anyone review the license and comment here if you have other opinions, or other way around.

### What type of PR is it?
[Task]

### Todos
* [x] - remove implementation from source tree

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1532

### Questions:
* Does the licenses files need update? yes


Leemoonsoo:

PR #1508, opened 2016-10-12
[ZEPPELIN-1537] Elasticsearch improvement for results of aggregations

bbonnin:
### What is this PR for?
The result of an aggregation query returned by the interpreter contains only "key" and "doc_count" in case of a multi-buckets aggregations.
But the result returned by Elasticsearch can contain more data according to the query.
This PR is an improvement of the result returned by the interpreter.

### What type of PR is it?
[Improvement]

### Todos
* [X] - Dev of the improvement in the interpreter
* [X] - Add a test case 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1537

In a paragraph, enter a query with multiple aggregations:
search /logs { "aggs" : { 
            "length" : { "terms": { "field": "status" }, 
            "aggs" : { "sum_length" : { "sum" : { "field" : "content_length" } } } } }

### Questions:


bzz:
Looks good to me, 👍 for tests!

PR #1514, opened 2016-10-14
ZEPPELIN-1516. NPE LivySparkSQLInterpreter thrown with %livy.sql interpreter

zjffdu:
### What is this PR for?

This PR fixed NPE issue in `LivySparkSQLIntepreter`, also it resolves the multiple thread issues discussed in #1503
### What type of PR is it?

[Bug Fix]
### Todos
- [ ] - Task
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-1516
- Enable shiro
- Log in 2 users in 2 browsers
- Start livy sql paragraph at almost the same time
- 2 livy sessions are created correctly. 
### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
@@ -69,13 +69,14 @@ public class LivyHelper {
 
       String confData = gson.toJson(conf);
       String user = context.getAuthenticationInfo().getUser();
-
+      LOGGER.debug("Try to create session for user:" + user);
       String json = executeHTTP(property.getProperty("zeppelin.livy.url") + "/sessions", "POST",
           "{" +
               "\"kind\": \"" + kind + "\", " +
               "\"conf\": " + confData + ", " +
-              "\"proxyUser\": " + (StringUtils.isEmpty(user) ? null : "\"" + user + "\"") +
-              "}",
+              "\"proxyUser\": " + (context.getAuthenticationInfo().isAnonymous() ? null : "\"" +
+              user + "\"") +
+          "}",
           context.getParagraphId()
       );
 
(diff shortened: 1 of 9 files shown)


felixcheung:
looks good. is it possible to have a test for this?

zjffdu:
I plan to do it after #1462, could you also take a look at #1462 ?

zjffdu:
@felixcheung @prabhjyotsingh Please help review.

prabhjyotsingh:
Does this happens only for livy.sql, or can it happen to others as well ?

gss2002:
@prabhjyotsingh I have only seen this issue with livy.sql as it is the only one that shares the HashMap between livy.spark and livy.sql. livy.sparkr and livy.pyspark do not share session info. In the future yes?

zjffdu:
If you start livy.sql first, you will hit this issue, because userSessionMap is null.

prabhjyotsingh:
Sure, got it. LGTM!

livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
```diff
@@ -131,6 +132,12 @@ public Integer createSession(InterpreterContext context, String kind) throws Exc
     }
   }
 
+  protected void initializeSpark(final InterpreterContext context,
+                                 final Map<String, Integer> userSessionMap) throws Exception {
+    interpret("val sqlContext = new org.apache.spark.sql.SQLContext(sc)\n" +
+        "import sqlContext.implicits._", context, userSessionMap);
```
felixcheung: isn't Livy suppose to do this internally?

zjffdu: sqlContext is not available in livy 0.2 but available for later version. So here I will create sqlContext is it is not available.

felixcheung:
Got it. Is there a way to version check Livy, like how we do Spark? Perhaps a PR on Livy to expose that?

Just thinking this could be fragile when moving to SparkSession and so on (maybe in the future)

zjffdu:
I am planing to add such rest api for livy, but for now livy don't have such api for checking version.

felixcheung:
Great, one question LGTM otherwise

livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
zjffdu: yes, as LivySQLInterpreter needs SQLContext. For other interpreters (livy.spark, livy.pyspark, livy.sparkr), if SQLContext is not available, user can create it by himself.

zjffdu:
Close this PR, as I am refactoring LivyInterpreter in ZEPPELIN-1607

PR #1530, opened 2016-10-17
[Zeppelin-1555] Eliminate prefix in PythonInterpreter exception

vectorijk:
### What is this PR for?
Solve bug metioned [here](https://github.com/apache/zeppelin/blob/3dec4d7006b8a57136f34ae330ba937d8990f2d2/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L139)

Since we launch python interpreter as a process and redirect stdin and stdout, only exception occurred (like syntax error or indentation error, etc) could give string like `...`. Thus, we don't need to determine whether syntax error happened in [`PythonProcess.sendAndGetResult`](https://github.com/apache/zeppelin/blob/3dec4d7006b8a57136f34ae330ba937d8990f2d2/python/src/main/java/org/apache/zeppelin/python/PythonProcess.java#L86) because we have detected error in [`PythonInterpreter.pythonErrorIn`](https://github.com/apache/zeppelin/blob/3dec4d7006b8a57136f34ae330ba937d8990f2d2/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L152)
### What type of PR is it?
Bug Fix

### What is the Jira issue?
Jira: https://issues.apache.org/jira/browse/ZEPPELIN-1555
Test locally.

### Screenshots
screen shot 2016-10-16 at 18 05 00

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonProcess.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonProcess.java
@@ -91,11 +91,6 @@ public class PythonProcess {
     String line = null;
     while (!(line = reader.readLine()).contains(STATEMENT_END)) {
       logger.debug("Read line from python shell : " + line);
-      if (line.equals("...")) {
-        logger.warn("Syntax error ! ");
-        output.append("Syntax error ! ");
-        break;
-      }
       output.append(line + "\n");
     }
     return output.toString();
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -68,7 +68,7 @@ public class PythonInterpreter extends Interpreter {
 
   @Override
   public void open() {
-    LOG.info("Starting Python interpreter .....");
+    LOG.info("Starting Python interpreter ---->");
     LOG.info("Python path is set to:" + property.getProperty(ZEPPELIN_PYTHON));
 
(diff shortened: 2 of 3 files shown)


python/src/main/java/org/apache/zeppelin/python/PythonProcess.java
```diff
@@ -91,11 +91,6 @@ public String sendAndGetResult(String cmd) throws IOException {
     String line = null;
     while (!(line = reader.readLine()).contains(STATEMENT_END)) {
       logger.debug("Read line from python shell : " + line);
-      if (line.equals("...")) {
-        logger.warn("Syntax error ! ");
-        output.append("Syntax error ! ");
-        break;
```
vectorijk: break here could cause cause chaos afterwards.

vectorijk:
cc @zjffdu

bzz:
Thank you for contributing! 

How do you think, how hard would be to add some tests here, to make sure the change does not break things?

vectorijk:
@bzz Thanks for the quick response! I will write and update some tests for this to make sure the changes doesn't break things before.

vectorijk:
@bzz I have updated and added unit test.

vectorijk:
cc @Leemoonsoo @jongyoul

Leemoonsoo:
Tested and Looks great to me!

Thanks @vectorijk for the contribution!

Leemoonsoo:

vectorijk:
@Leemoonsoo @bzz Thanks for the review!

PR #1536, opened 2016-10-18
[Zeppelin 796] Duplicated notebook names should not be allowed

rajarajan-g:
What is this PR for?

When a notebook is created/cloned/imported the title/name of the new notebook should not be duplicate of the existing notebook's title/name

What type of PR is it?

Improvement

Todos

 - Task
What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-796

How should this be tested?

Create a notebook with name 'Note', Notebook should allow user to create a note
When a user try to create/clone/import a notebook with name 'Note', Notebook should not allow user to create note.
Screenshots (if appropriate)

Screenshot of error message displayed during create/clone notebook
create

Screenshot of error message displayed during import notebook
import

Questions:

Does the licenses files need update? No
Is there breaking changes for older versions? No


anthonycorbacho:
I think its okay to have a duplicate name, if the noteId is different nothing bad will happen. What kind of case are you trying to solve here?

rajarajan-g:
@anthonycorbacho : Lets say if i create a note with name 'Note'. If zeppelin allows me to create a new note with name 'Note', then after sometime when i search for one of this note, I am going to get confused here.

anthonycorbacho:
I see your point, but let say i want to import a note and i have a note with the same name, it will not allow me to do that and i feel like its a pretty bad user experience here. What do you think?

rajarajan-g:
@anthonycorbacho : For import note function too, we will have same problem as i pointed out first. Anyway for import note too, any user can give a new name for importing notebook

rajarajan-g:
@All : Please review this PR and let me know your comments

AhyoungRyu:
@rajarajan-g There are some conflicts in here. Could you handle them? :)

rajarajan-g:
Ping

rajarajan-g:
Now i have resolved the conflicts. Please review

rajarajan-g:
Any updates ?

Tagar:
IMHO, automatically adding `" (1)", " (2)" ..` etc to a note name in case of duplicate name found 
might be a better solution than disallowing import at all. A variation on this idea - adding `" (import date/time)"` suffix to note name would do too (only in case if dup name found).

It'll be then at user's discretion to rename notes appropriately - so they would still be allowed to import a note.

PR #1537, opened 2016-10-18
[Zeppelin-1561] Improve sync for multiuser environment 

khalidhuseynov:
### What is this PR for?
apply multi-tenancy for storage sync mechanism

### What type of PR is it?
Bug Fix | Improvement

### Todos
* [x] - broadcast on sync
* [x] - set permissions for pulled notes
* [x] - add test

### What is the Jira issue?
 [ZEPPELIN-1561](https://issues.apache.org/jira/browse/ZEPPELIN-1561)

green CI

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -92,6 +92,10 @@ public class NotebookRepoSync implements NotebookRepo {
       LOG.info("No storages could be initialized, using default {} storage", defaultStorage);
       initializeDefaultStorage(conf);
     }
+    //syncOnStart();
+  }
+
+  private void syncOnStart() {
     if (getRepoCount() > 1) {
       try {
         AuthenticationInfo subject = new AuthenticationInfo("anonymous");
@@ -101,7 +105,6 @@ public class NotebookRepoSync implements NotebookRepo {
       }
     }
   }
-
   @SuppressWarnings("static-access")
   private void initializeDefaultStorage(ZeppelinConfiguration conf) {
     Class<?> notebookStorageClass;
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -166,6 +166,10 @@ public class NotebookRepoSync implements NotebookRepo {
     /* TODO(khalid): handle case when removing from secondary storage fails */
   }
 
+  void remove(int repoIndex, String noteId, AuthenticationInfo subject) throws IOException {
+    getRepo(repoIndex).remove(noteId, subject);
+  }
+
   /**
    * Copies new/updated notes from source to destination storage
    *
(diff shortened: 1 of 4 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -230,16 +235,34 @@ public void sync(AuthenticationInfo subject) throws IOException {
...
+          makePrivate(id, subject);
+        }
       } catch (IOException e) {
         LOG.error("Failed to push note to storage, moving onto next one", e);
       }
     }
   }
 
+  private void makePrivate(String noteId, AuthenticationInfo subject) {
+    if (subject != null && !"anonymous".equals(subject.getUser())) {
```
anthonycorbacho: How about creating in `AuthenticationInfo` a new method called `isAnonymous(AuthenticationInfo subject)` and move the check inside?

something like
```
// In AuthenticationInfo.java
 private static final String ANONYMOUS = "anonymous";

 public static boolean isAnonymous(AuthenticationInfo subject) {
   if (subject != null ) {
     LOG.info("Subject is null assuming anonymous");
     return true;
   }
   return ANONYMOUS.equals(subject.getUser(); 
 }

//for the fun
public boolean isAnonymous() {
  return ANONYMOUS.equals(user);
}

//then here you can do something like this in makePrivate

if (AuthenticationInfo.isAnonymous(subject)) {
  // Add maybe meaningful log :)
  return;
}
NotebookAuthorization notebookAuthorization = NotebookAuthorization.getInstance();
Set<String> users = notebookAuthorization.getOwners(noteId);
users.add(subject.getUser());
notebookAuthorization.setOwners(noteId, users);
users = notebookAuthorization.getReaders(noteId);
users.add(subject.getUser());
notebookAuthorization.setReaders(noteId, users);
users = notebookAuthorization.getWriters(noteId);
users.add(subject.getUser());
notebookAuthorization.setWriters(noteId, users);

```

khalidhuseynov: makes sense, can be reused to make `anonymous` consistent in usage over codebase. I'll change now

khalidhuseynov:
@anthonycorbacho addressed in d0a8c3c3c7eaf84e2796997bd85d3d565101550f with some minor changes

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -92,6 +94,10 @@ public NotebookRepoSync(ZeppelinConfiguration conf) {
       LOG.info("No storages could be initialized, using default {} storage", defaultStorage);
       initializeDefaultStorage(conf);
     }
+    //syncOnStart();
```
khalidhuseynov: @rajarajan-g thanks for pointing to it. actually initial purpose was to delete it, but it was extracted into sub-function and commented out. addressed by deleting it in 1c11c2ddbd27b222669278f22f87b65cd2292fc9

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -230,16 +224,36 @@ public void sync(AuthenticationInfo subject) throws IOException {
...
   }
 
+  private void makePrivate(String noteId, AuthenticationInfo subject) {
+    if (AuthenticationInfo.isAnonymous(subject)) {
+      LOG.info("User is anonymous, permissions are not set for pulled notes");
+      return;
+    }
+    NotebookAuthorization notebookAuthorization = NotebookAuthorization.getInstance();
+    Set<String> users = notebookAuthorization.getOwners(noteId);
+    users.add(subject.getUser());
```
Leemoonsoo: Could you elaborate why 'makePrivate()' routine, which is being used inside of sync() adds current user as a owner, reader, writer?

Leemoonsoo: Can you explain why it sets when pulling notes from secondary storage to the main one?

For example, what happen if a note permission has a user as a reader, but not as a writer and owner?

khalidhuseynov: if it's pulling, that means there was no such a note in main storage and it's pulling new note from secondary storage. Thus, that new note not supposed to have previous  owners, readers, writers. So generally it's same as creating note and only owner should be set. But I also set readers and writers in order other users don't see that new note (make it private).

khalidhuseynov:
@Leemoonsoo your comment makes sense if there's already existing acl for the note. addressed it with test under e41b9b9324971b472a9210b03c189f01ffc6a0f3

khalidhuseynov:
CI is green, ready for review

anthonycorbacho:
LGTM merging if no more discussion

PR #1540, opened 2016-10-20
[ZEPPELIN-1566] Make paragraph editable with double click

minahlee:
### What is this PR for?
This PR enables edit on double click for markdown/angular paragraph.  Users can change `editOnDblClick` field to be `false` by editting `interpreter/md/interpreter-setting.json` or  `conf/interpreter.json`. In the same context, users can set other type paragraphs to be editable on double click by setting `editOnDblClick` to be true.

This PR also fixes bug that syntax highlight doesn't work on pasted code.

### What type of PR is it?
Feature

### Todos
* [x] Create test
* [x] Update docs

### What is the Jira issue?
[ZEPPELIN-1566](https://issues.apache.org/jira/browse/ZEPPELIN-1566)

1. Create new markdown interpreter
2. Create notebook and run markdown paragraph
3. Double click markdown paragraph to edit

**Edit on double click and hide editor on paragraph run**
oct-20-2016 12-28-54

**Syntax highlight on paste**

Before
oct-20-2016 12-24-54

After
oct-20-2016 12-25-06

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -705,7 +705,7 @@
       }
     };
 
-    var getAndSetEditorSetting = function(session, interpreterName) {
+    var getEditorSetting = function(interpreterName) {
       var deferred = $q.defer();
       websocketMsgSrv.getEditorSetting($scope.paragraph.id, interpreterName);
       $timeout(
@@ -715,13 +715,14 @@
           }
         }
       ), 1000);
-      deferred.promise.then(function(editorSetting) {
-        if (!_.isEmpty(editorSetting.editor)) {
-          var mode = 'ace/mode/' + editorSetting.editor.language;
-          $scope.paragraph.config.editorMode = mode;
-          session.setMode(mode);
-        }
-      });
+      return deferred.promise;
+    };
+
+    var setEditorLanguage = function(session, language) {
+      var mode = 'ace/mode/';
+      mode += language;
+      $scope.paragraph.config.editorMode = mode;
+      session.setMode(mode);
     };
 
     var setParagraphMode = function(session, paragraphText, pos) {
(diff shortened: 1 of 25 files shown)


jongyoul:
How about making editOnDblClick false by default?

minahlee:
@jongyoul Thank you for suggestion!
As you might see I set `editOnDblClick` to be `true` for angular, markdown.

I think make markdown paragraph editable with double click is quite common behavior. Jupyter notebook and databricks cloud would be the example. Also in ux point of view, I think it would be rare to show both editor and output at the same time for markdown paragraph.

But not sure what would be better for angular paragraph.

What do you think?

jongyoul:
It looks good to hide editor on md. And I'm not good at ui, but in case of angular, I think it's better to hide it, too.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -414,6 +419,28 @@
       commitParagraph($scope.paragraph.title, $scope.paragraph.text, newConfig, newParams);
     };
 
+    var openEditorAndCloseTable = function() {
```
anthonycorbacho: The code below from L422 to L443 look very alike, what do you think about making one function that will handle both case?

I have in mind something like
```
var openEditorAndCloseTable = function() {
  manageEditorAndTableState(false, true);
}

var closeEditorAndOpenTable = function() {
  manageEditorAndTableState(true, false);
}

// Please dont get attached to the name :) put whatever you like :D
var manageEditorAndTableState(showEditor, showTable) {
  var newParams = angular.copy($scope.paragraph.settings.params);
  var newConfig = angular.copy($scope.paragraph.config);

  newConfig.editorHide = showEditor;
  newConfig.tableHide = showTable;

  commitParagraph($scope.paragraph.title, $scope.paragraph.text, newConfig, newParams);
}
```
The idea here is to keep this login in one single place so it will help for the future maintainability.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -127,6 +127,10 @@
 
   private Interpreter devInterpreter;
 
+  private static final Map<String, Object> DEFAULT_EDITOR = ImmutableMap.of(
```
anthonycorbacho: This is awesome, love `ImmutableMap`!!!

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
minahlee: Thank you for review. I handled it in the last commit.

anthonycorbacho:
Awesome, **LGTM**

minahlee:
@jongyoul @anthonycorbacho Thank you for review! Merging if there is no more discussion.

tae-jun:
Wow, this is a really cool feature for me. Thanks! 👍

PR #1549, opened 2016-10-22
ZEPPELIN-1547. Paragraph text should be trimmed

zjffdu:
### What is this PR for?

The paragraph text should be trimmed, otherwise repl name can not be correctly recognized. 
### What type of PR is it?

[Bug Fix]
### Todos
- [ ] - Task
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-1574

Unit test is added
### Questions:


felixcheung:
Could you clarify? I think you are referring to the first line with the `%interpreter` parsing, and not trimming all paragraph text?

zjffdu:
Yes, the bug is due to interpreter parsing. But I think triming the paragraph text is the easiest fix.

PR #1554, opened 2016-10-23
[ZEPPELIN-1320] Run zeppelin interpreter process as web front end user

prabhjyotsingh:
Have recreated this from https://github.com/apache/zeppelin/pull/1322
### What is this PR for?

While running a Notebook using shell, spark, python uses same user as which zeppelin server is running. Which means these interprets have same permission on file system as zeppelin server.
IMO users should be able to impersonate themselves as a complete security system.
### What type of PR is it?

[Improvement]
### Todos
- [x] - Update doc
- [x] - FIX NPEs
- [x] - FIX CI
### What is the Jira issue?
- [ZEPPELIN-1320](https://issues.apache.org/jira/browse/ZEPPELIN-1320)
- Enable shiro auth in shiro.ini
- Add ssh key for the same user you want to try and impersonate (say user1).

```
adduser user1
ssh-keygen
ssh user1@localhost mkdir -p .ssh
cat ~/.ssh/id_rsa.pub | ssh user1@localhost 'cat >> .ssh/authorized_keys'
```
- Start zeppelin server, try and run following in paragraph in a notebook
- Go to interpreter setting page, and enable "User Impersonate" in any of the interpreter (in my example its shell interpreter)

```
%sh
whoami
```

Check that it should run as new user, i.e. "user1"

user impersonate

### Questions:


prabhjyotsingh:
know issue: restart on interpreter does not work as expected.

Leemoonsoo:
@prabhjyotsingh Do you think is there a way to impersonate without adding ssh key when user logged in using PAM authentication https://github.com/apache/zeppelin/pull/1589?

prabhjyotsingh:
Sure let me check, I think it could be possible.

prabhjyotsingh:
Updated screen shot, ready for review.

@Leemoonsoo  I'll try to take care of "PAM authentication" in a different PR.

r-kamath:
@prabhjyotsingh LGTM.
Thanks for the documentation.

bin/interpreter.sh
```diff
@@ -42,6 +42,9 @@ while getopts "hp:d:l:v" o; do
             . "${bin}/common.sh"
             getZeppelinVersion
             ;;
+        u)
+            ZEPPELIN_SSH_COMMAND="ssh ${OPTARG}@localhost "
```
Leemoonsoo: How about make ZEPPELIN_SSH_COMMAND configurable? because some user may run ssh in non standard port, some user may event want to use different command, like sudo, to impersonate.

The idea is, we can make this part just get impersonate user
```bash
u)
    ZEPPELIN_IMPERSONATE_USER="${OPTARG}"
```

and then we can define ZEPPELIN_IMPERSONATE_CMD from `bin/common.sh` in configurable way

```bash
if [[ -z "$ZEPPELIN_IMPERSONATE_CMD" ]]; then
  ZEPPELIN_IMPERSONATE_CMD=`echo "ssh ${ZEPPELIN_IMPERSONATE_USER}@localhost"`
else
  ZEPPELIN_IMPERSONATE_CMD=$(eval "echo ${ZEPPELIN_IMPERSONATE_CMD}")
fi
```

So user can override ZEPPELIN_IMPERSONATE_CMD in `conf/zeppelin-env.sh` or inside of interpreter property, for example

```
ZEPPELIN_IMPERSONATE_CMD='sudo -u ${ZEPPELIN_IMPERSONATE_USER}'
```

what do you think?

prabhjyotsingh: Yes, sure, that totally make sense. I'll do it right away.

Leemoonsoo:
LGTM and merge to master if there're no further comments.

PR #1557, opened 2016-10-24
ZEPPELIN-1577. LivyInterpreter should not use FIFOScheduler

zjffdu:
### What is this PR for?

This PR create a new scheduler FIFOPerUserScheduler. So that each user's request is isolated. and it is FIFOSchduler per user. 
### What type of PR is it?

[Bug Fix | Improvement]
### Todos
- [ ] - Task
### What is the Jira issue?
- https://issues.apache.org/jira/browse/ZEPPELIN-1577

Tested manually 
### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOPerUserScheduler.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOPerUserScheduler.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.apache.zeppelin.scheduler;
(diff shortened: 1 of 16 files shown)


zjffdu:
Will add test after #1462

zjffdu:
@prabhjyotsingh @felixcheung  Please help review.

felixcheung:
Probably something @Leemoonsoo should review

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOPerUserScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class FIFOPerUserScheduler implements Scheduler {
+
+  private static Logger LOGGER = LoggerFactory.getLogger(FIFOScheduler.class);
+
+  private ExecutorService executor;
+  private SchedulerListener listener;
+  boolean terminate = false;
+  private String name;
+  private int maxUser = 10;
```
Leemoonsoo: Looks like unused var. Is it?

bikassaha:
Perhaps for this issue and other similar issues related to multiple users, the mode for the livy interpreter should be scoped instead of shared. If we run 1 interpreter per user then that matches the end to end flow wrt sessions. Then interpreters will have their own queue and not need this workaround. Also one user restarting their interpreter would not affect existing interpreter instances for others users. Is that a correct understanding?
Reply

Leemoonsoo:
@bikassaha Good point. I Agree, that approach seems better.

zjffdu:
Yes, that would be better. I think currently zeppelin hard coded the default mode (shared mode), there's no way to set the default mode for interpreter, we may add one option in interpreter-setting.json.  And also we may add one option to restrict the mode usage as for livy only scoped is valid.

bikassaha:
@zjffdu Isolated would also be valid but perhaps an overkill. So default of scoped makes sense.

Leemoonsoo:
Exactly. Adding option in interpreter-setting.json and restrict mode usage is great idea. Similar approaches has been made in ZEPPELIN-1026 for editor mode restriction.

zjffdu:
Close this one, and open #1585 to allow user to set default mode in interpreter-setting.json

PR #1565, opened 2016-10-27
[ZEPPELIN-1564] Enable note deletion and paragraph output clear from main page

minahlee:
### What is this PR for?
- Enables removing note and clear all paragraph's output from Zeppelin main page.
- Add rest api for clearing all paragraph output
  
Next possible improvement can be removing notes in folder level and rename folder.

### What type of PR is it?
Improvement

### Todos
* [x] - Merge #1567 and apply security to `clearAllParagraphOutput` rest api method

### What is the Jira issue?

[ZEPPELIN-1564](https://issues.apache.org/jira/browse/ZEPPELIN-1564)

oct-27-2016 18-26-03
### Questions:

### DIFF
--- a/docs/rest-api/rest-notebook.md
+++ b/docs/rest-api/rest-notebook.md
@@ -493,7 +493,7 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
     <col width="200">
     <tr>
       <td>Description</td>
-      <td> This ```POST``` method runs the paragraph synchronously by given note and paragraph id. This API can return SUCCESS or ERROR depending on the outcome of the paragraph execution
+      <td>This ```POST``` method runs the paragraph synchronously by given note and paragraph id. This API can return SUCCESS or ERROR depending on the outcome of the paragraph execution
       </td>
     </tr>
     <tr>
@@ -969,6 +969,34 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
   "message": "",
   "body": "2AZPHY918"
 }</pre></td>
+    </tr>
+    </tr>
+  </table>
+
+<br />
+### Clear all paragraph result
+  <table class="table-configuration">
+    <col width="200">
+    <tr>
+      <td>Description</td>
+      <td>This ```POST``` method clear all paragraph results from note of given id.
+      </td>
+    </tr>
+    <tr>
+      <td>URL</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/notebook/[noteId]/clear```</td>
+    </tr>
+    <tr>
+      <td>Success code</td>
+      <td>200</td>
+    </tr>
+    <tr>
+      <td> Fail code</td>
+      <td> 500 </td>
+    </tr>
+    <tr>
+      <td>sample JSON response</td>
+      <td><pre>{"status": "OK"}</pre></td>
     </tr>
     </tr>
   </table>
--- a/docs/rest-api/rest-notebook.md
+++ b/docs/rest-api/rest-notebook.md
@@ -991,8 +991,16 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
       <td>200</td>
     </tr>
     <tr>
-      <td> Fail code</td>
-      <td> 500 </td>
+      <td>Forbidden code</td>
+      <td>401</td>
+    </tr>
+    <tr>
+      <td>Not Found code</td>
+      <td>404</td>
+    </tr>
+    <tr>
+      <td>Fail code</td>
+      <td>500</td>
     </tr>
     <tr>
       <td>sample JSON response</td>
--- a/docs/rest-api/rest-notebook.md
+++ b/docs/rest-api/rest-notebook.md
@@ -979,7 +979,7 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
     <col width="200">
     <tr>
       <td>Description</td>
-      <td>This ```POST``` method clear all paragraph results from note of given id.
+      <td>This ```PUT``` method clear all paragraph results from note of given id.
       </td>
     </tr>
     <tr>
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -528,14 +528,11 @@ public class NotebookRestApi {
   public Response clearAllParagraphOutput(@PathParam("noteId") String noteId)
       throws IOException {
     LOG.info("clear all paragraph output of note {}", noteId);
+    checkIfUserCanWrite(noteId, "Insufficient privileges you cannot clear this note");
 
-    if (!noteId.isEmpty()) {
-      Note note = notebook.getNote(noteId);
-      if (note == null) {
-        return new JsonResponse(Status.NOT_FOUND, "note not found.").build();
-      }
-      note.clearAllParagraphOutput();
-    }
+    Note note = notebook.getNote(noteId);
+    checkIfNoteIsNotNull(note);
+    note.clearAllParagraphOutput();
 
     return new JsonResponse(Status.OK, "").build();
   }
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -43,7 +43,7 @@ import org.apache.zeppelin.notebook.Notebook;
 import org.apache.zeppelin.notebook.NotebookAuthorization;
 import org.apache.zeppelin.notebook.Paragraph;
 import org.apache.zeppelin.rest.exception.NotFoundException;
-import org.apache.zeppelin.rest.exception.UnauthorizedException;
+import org.apache.zeppelin.rest.exception.ForbiddenException;
 import org.apache.zeppelin.rest.message.CronRequest;
 import org.apache.zeppelin.rest.message.NewNoteRequest;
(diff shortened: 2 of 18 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -462,6 +462,30 @@ public Response deleteParagraph(@PathParam("noteId") String noteId,
...
+  public Response clearAllParagraphOutput(@PathParam("noteId") String noteId)
+      throws IOException {
+    LOG.info("clear all paragraph output of note {}", noteId);
+
+    if (!noteId.isEmpty()) {
+      Note note = notebook.getNote(noteId);
+      if (note == null) {
+        return new JsonResponse(Status.NOT_FOUND, "note not found.").build();
+      }
+      note.clearAllParagraphOutput();
```
khalidhuseynov: I wonder how we enforce note authorization values here, for example someone without owner/write privileges does a call via rest api.

anthonycorbacho: yeah, i think @khalidhuseynov  is right here.
But after checking the code of notebookRestApi, there are no check here. we need to fix this...

minahlee: I checked notebookRestApi class and there are a lot of method which doesn't take care of privileges like `getNote`, `getParagraph`. I think it is better to create another PR to refactor notebookRestApi.

anthonycorbacho: @minahlee thats right, i created the issue [ZEPPELIN-1586](https://issues.apache.org/jira/browse/ZEPPELIN-1586) i will create the  PR soon

minahlee:
@khalidhuseynov Thank you for review. I made "clear paragraph output" in main page for the case that note's output data is too big so it causes Zeppelin fail to load note and hang. 
Here are tickets reporting this issue: [ZEPPELIN-1564](https://issues.apache.org/jira/browse/ZEPPELIN-1564), [ZEPPELIN-1491](https://issues.apache.org/jira/browse/ZEPPELIN-1491)

The more optimal solution would be improving data loading and rendering itself. We might not need "clear paragraph output" button once we improve this performance issue.

I believe having delete button on the main page is beneficial in terms of UX in case user wants to delete multiple notes. They won't need to navigate into each note to delete them. I am planning to implement remove folder & edit folder name in main page after this PR is merged.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
minahlee: @anthonycorbacho Thank you! @khalidhuseynov Let me update this PR to handle permission issue once [ZEPPELIN-1586](https://issues.apache.org/jira/browse/ZEPPELIN-1586) is ready

tae-jun:
Hi @minahlee :) This is a cool feature! I have an experience that deleted output through editing note.json because of huge output stopping Zeppelin web.

By the way, I've wanted a renaming folder feature. So as soon as I saw this PR, I implemented renaming note, and I am almost done with implementing renaming folder! (I saw you are planning this after coding, sorry 😢) This is a really good example for implementing it!

So, the question is this:
My branch is based on your PR's branch. How should I create JIRA issue? As well as PR.
Is it easy to wait to merge this PR?

minahlee:
Hi @tae-jun, thank you for your interest. I had to rebase this branch to resolve conflict so you will need too since it is based on mine.

I am planning to finish this PR once #1567 merged so that I can handle permission issue. It will speed up merging process if you can review it (I am also going to review it as soon as all the tasks are done).

About the jira, there is nothing stops you from creating one so feel free to create a ticket on Jira. And make sure that you assign your self on that task, and put `PROGRESS` label so everyone can see that you are working on it. And for the PR,  I suggest you to do the cherry pick after #1567, #1565 merged, otherwise it will cause some confusion for the ones who don't know the context.

tae-jun:
Thanks @minahlee! I appreciate your kind comment 😄

I created a ticket on JIRA: [ZEPPELIN-1598](https://issues.apache.org/jira/browse/ZEPPELIN-1598)

But, I don't know how to set myself to assignee, and change status to `PROGRESS`. I guess I don't have a permission to do that 😢

I will check #1567. Thanks again!

minahlee:
@tae-jun I just added your name to contributor list so you will be able to set yourself as assignee from now.

tae-jun:
@minahlee Awesome! I just changed status to `IN PROGRESS` now. Thanks 👍

Leemoonsoo:
Tested and LGTM

minahlee:
@khalidhuseynov I updated rest api to do privilege check. Could you help review?

khalidhuseynov:
@minahlee looks much better now! thanks for addressing the issue and LGTM

minahlee:
@khalidhuseynov Thank you for quick response!

docs/rest-api/rest-notebook.md
```diff
@@ -972,3 +972,31 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
...
+      <td>URL</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/notebook/[noteId]/clear```</td>
+    </tr>
+    <tr>
+      <td>Success code</td>
+      <td>200</td>
+    </tr>
+    <tr>
+      <td> Fail code</td>
+      <td> 500 </td>
```
anthonycorbacho: shall we also add 404 (not found) and 401(unauthorised)? as a possible response

tae-jun:
@minahlee I have a question about REST API style. Is there any purpose that you used `PUT` method?

In my opinion, `DELETE` method is more proper since it does not `PUT` `clear` under a `noteId`. This action `DELETE`s `noteId`'s `outputs`. So, this is API style I suggest:

```
DELETE {noteId}/outputs
```

And also maybe we can implement one paragraph output deletion in the future like:

```
DELETE {noteId}/outputs/{paragraphId}
```

minahlee:
@tae-jun the reason I use `PUT` is that `clearAllParagraphOutput` method is not really deleting `result` field from paragraph, but updating `result` field to empty.
I was considering using `DELETE` but I didn't because it can confuse user that it will literally 'DELETE' result field.

What do you think?

tae-jun:
@minahlee Oh, in that context, it makes sense. LGTM.

But, after looking all around Notebook REST API, it works well now but it doesn't seem RESTful. I summarized current list of REST API:

- `GET`     /
- `POST`    /
- `GET`     /[noteId]
- `POST`    /[noteId] 
- `DELETE`  /[noteId]
- `GET`     /job/[noteId]
- `POST`    /job/[noteId]
- `DELETE`  /job/[noteId]
- `GET`     /job/[noteId]/[paragraphId]
- `POST`    /job/[noteId]/[paragraphId]
- `DELETE`  /job/[noteId]/[paragraphId]
- `POST`    /run/[noteId]/[paragraphId]
- `GET`     /cron/[noteId]
- `POST`    /cron/[noteId]
- `DELETE`  /cron/[noteId]
- `GET`     /search?q=[query]
- `POST`    /[noteId]/paragraph
- `GET`     /[noteId]/paragraph/[paragraphId]
- `DELETE`  /[noteId]/paragraph/[paragraphId]
- `POST`    /[noteId]/paragraph/[paragraphId]/move/[newIndex]
- `GET`     /export/[noteId]
- `POST`    /import

A good REST API is API that doesn't need a documentation, which means users can understands only with HTTP method and URL. So, it should be simple. However, I think some of current API may need to be changed.

REST API is UI for developers, so should be well designed. There are problems I can see now:
- Need to organize resources of Zeppelin: hierarchy of resources, and what they have; For example:
  - Need hierarchical URL design: `notebook/:noteId/paragraphs/:paragraphId/{title|text|result|job}`
  - Do jobs belong to notebook or paragraph?
- Use plural
- Use nouns as far as you can, no verbs. Make use of HTTP methods. URL should point resources.
- For in the case of a request that doesn't fit CRUD(Create(POST), Read(GET), Update(PUT|PATCH), Detele(DELETE)), we need a rule for this. Heroku deals with this by placing actions for the resource under `/actions`. For example, in our case, run all paragraphs can be represented as `POST notebook/:noteId/actions/run`
- Need API versioning

Zeppelin needs more REST API, so eventually, it will be getting bigger. I think it's time to organize REST APIs. After all of this, the REST API documentation should also be updated. Not by listing actions what you can, just by listing APIs. This is easier to read since users can know what they can do with resources. 

Would you mind if I open an issue for this?

anthonycorbacho:
@tae-jun This is a great sum of the status of the rest api in zeppelin, and i think we all agree that currently zeppelin doenst follow rest standard (just by looking at the endpoint + response (mostly 500 ...)).

But you have to consider that going on the path of refacto this logic will be a huge task, not only in backend but you need to understand that a lot of users are using this api already changing the hole api will brings a lot of breaking change in they current workflow. So we need to be careful here.

khalidhuseynov:
@tae-jun thanks for summarizing and giving feedback on REST API. Although we can't address all at once, but I think it deserves an issue with above content and feedback logged in there

tae-jun:
@anthonycorbacho @khalidhuseynov Thanks for the comment! I agree with you guys.

This is a breaking change, so yeah we must be careful. So, if we proceed this, new API URL's prefix will be `v1/api/notebook`. And keep legacy API under `api/notebook` for backward compatibility. This will be the start of versioning.

I think we should design REST API together. I will create a proposal on JIRA soon, please review it when the time comes!

It will be harder to fix later. So, let's start now! 😄

PR #1518, opened 2016-10-14
[ZEPPELIN-1549] Change NotebookID variable name to NoteID

hyonzin:
### What is this PR for?
This PR fixes wrong written NotebookID to NoteID.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1549

### Questions:
* Does the licenses files need update? No.
* Is there breaking changes for older versions? No.
* Does this needs documentation? No

PR #1567, opened 2016-10-29
[ZEPPELIN-1586] Add security check in NotebookRestApi

anthonycorbacho:
### What is this PR for?

Bring some security check in `NotebookRestApi`.
### What type of PR is it?

[Bug Fix | Improvement | Refactoring]
### Todos
- [x] - Create a proper way to throw webapp error
- [x] - Add in `NotebookAuthorization` some method to check if user is owner, reader or writer
- [x] - Add Authorization check in `NotebookRestapi` 
- [x] - Add New test for security in notebook rest api

### What is the Jira issue?
- [ZEPPELIN-1586](https://issues.apache.org/jira/browse/ZEPPELIN-1586)

First, force Zeppelin to use auth.
- In `conf/zeppelin-site.xml` change `zeppelin.anonymous.allowed` to **false**
  
  ```
  <property>
  <name>zeppelin.anonymous.allowed</name>
  <value>false</value>
  <description>Anonymous user allowed by default</description>
  </property>
  ```
- In `conf/shiro.ini` set Shiro to use `Auth` at the end of the file
  
  ```
  #/** = anon                                                                                                                                           
  /** = authc
  ```
- Start Zeppelin, login and set some permission to a note
- try to get a note from Zeppelin Rest Api `http://localhost:8080/api/notebook/{noteId}` (you can use your browser or curl (if you use curl please add shiro token to curl cookie))

note_permission_rest_api
### Questions:
- Does this needs documentation? Maybe

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -34,31 +35,33 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
-import com.google.common.collect.Sets;
-import com.google.common.reflect.TypeToken;
-import com.google.gson.Gson;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.zeppelin.interpreter.InterpreterResult;
-import org.apache.zeppelin.utils.InterpreterBindingUtils;
-import org.quartz.CronExpression;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import org.apache.zeppelin.annotation.ZeppelinApi;
+import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.notebook.Note;
 import org.apache.zeppelin.notebook.Notebook;
 import org.apache.zeppelin.notebook.NotebookAuthorization;
 import org.apache.zeppelin.notebook.Paragraph;
+import org.apache.zeppelin.rest.exception.UnauthorizedException;
 import org.apache.zeppelin.rest.message.CronRequest;
-import org.apache.zeppelin.types.InterpreterSettingsList;
 import org.apache.zeppelin.rest.message.NewNoteRequest;
 import org.apache.zeppelin.rest.message.NewParagraphRequest;
 import org.apache.zeppelin.rest.message.RunParagraphWithParametersRequest;
 import org.apache.zeppelin.search.SearchService;
 import org.apache.zeppelin.server.JsonResponse;
 import org.apache.zeppelin.socket.NotebookServer;
+import org.apache.zeppelin.types.InterpreterSettingsList;
 import org.apache.zeppelin.user.AuthenticationInfo;
+import org.apache.zeppelin.utils.InterpreterBindingUtils;
 import org.apache.zeppelin.utils.SecurityUtils;
+import org.openqa.selenium.NotFoundException;
+import org.quartz.CronExpression;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Sets;
+import com.google.common.reflect.TypeToken;
+import com.google.gson.Gson;
 
 /**
  * Rest api endpoint for the notebook.
@@ -90,6 +93,8 @@ public class NotebookRestApi {
   @Path("{noteId}/permissions")
   @ZeppelinApi
   public Response getNotePermissions(@PathParam("noteId") String noteId) {
+    checkIfUserIsMember(noteId,
+        "Insufficient privileges you cannot get the list of permissions for this note");
     HashMap<String, Set<String>> permissionsMap = new HashMap<>();
     permissionsMap.put("owners", notebookAuthorization.getOwners(noteId));
     permissionsMap.put("readers", notebookAuthorization.getReaders(noteId));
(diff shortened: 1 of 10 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -444,14 +501,12 @@ public Response deleteParagraph(@PathParam("noteId") String noteId,
     LOG.info("delete paragraph {} {}", noteId, paragraphId);
 
     Note note = notebook.getNote(noteId);
-    if (note == null) {
-      return new JsonResponse(Status.NOT_FOUND, "note not found.").build();
-    }
+    checkIfNoteIsNotNull(note);
+    checkIfUserIsOwner(noteId,
```
minahlee: Since writer has permission to insert paragraph to note, wouldn't it make sense to allow them remove paragraph? For example if someone added new paragraph by mistake, he won't be able to remove it

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -106,34 +111,87 @@ private String ownerPermissionError(Set<String> current, Set<String> allowed) th
...
+   */
+  
+  /**
+   * Check if the current user own the given note.
+   */
+  private void checkIfUserIsOwner(String noteId, String errorMsg) {
+    Set<String> userAndRoles = Sets.newHashSet();
+    userAndRoles.add(SecurityUtils.getPrincipal());
+    userAndRoles.addAll(SecurityUtils.getRoles());
+    if (!notebookAuthorization.hasReadAuthorization(userAndRoles, noteId)) {
```
anthonycorbacho: Bad copy paste :dagger:

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -106,34 +111,87 @@ private String ownerPermissionError(Set<String> current, Set<String> allowed) th
...
     HashSet<String> userAndRoles = new HashSet<>();
     userAndRoles.add(principal);
     userAndRoles.addAll(roles);
-    if (!notebookAuthorization.isOwner(noteId, userAndRoles)) {
-      return new JsonResponse<>(Status.FORBIDDEN,
-          ownerPermissionError(userAndRoles, notebookAuthorization.getOwners(noteId))).build();
-    }
+    
+    checkIfUserCanWrite(noteId,
+        ownerPermissionError(userAndRoles, notebookAuthorization.getOwners(noteId)));
```
anthonycorbacho: Good catch, i guess a copy pasted so fast :(

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
anthonycorbacho: Yes, it make sense

minahlee:
Thank you for quick response, I only went through the code and it looks good to me. It would be nice if you can add some tests. Meanwhile let me build this branch and test it out.

anthonycorbacho:
@minahlee yeah, actually i am doing this right now, I also updated to todo tasks thanks for your review!

minahlee:
I tested some of rest apis and it works well. Next step would be applying same policies to websocket. For example `reader` cannot change bound interpreter to note via rest api after this PR, but it is possible to do it via websocket(or GUI).

anthonycorbacho:
@minahlee you are right, I guess the next step will be to abstract this logic from rest api and apply to both rest and websocket.

tae-jun:
I also tested and was able to reproduce screenshot above.

When I was another user, it returned `403` status code with message: 
```json
{"status":"FORBIDDEN","message":"Insufficient privileges you cannot get this note"}
```

However, when I didn't log in (i.e. anonymous), the browser(Chrome) redirected me to `http://localhost:8080/api/login` and returned `405` status code without any message. Users can be confused when there is no error message. And I think `403` status code is more proper since it's `forbidden`, not `method not allowed`.

So in my opinion, it would be better:
- Send `403` status code with some messages when a user is not logged in. Maybe something like:
```json
{"status":"FORBIDDEN","message":"Please log in"}
```

This is a miracle feature, by the way 👍

anthonycorbacho:
@tae-jun thanks for the feedback, let me take a look tomorrow

zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java
```diff
@@ -189,6 +227,14 @@ protected static void startUp() throws Exception {
       }
     }
   }
+  
+  protected static void startUpWithAutenticationEnabled() throws Exception {
```
tae-jun: There was a typo! :)

`startUpWithAutenticationEnabled` -> `startUpWithAuthenticationEnabled `

anthonycorbacho:
@tae-jun After looking at the code base, I think this case should be handle in another PR, this is kinda out of the scope of this PR and its already becoming super big.

But your made a very valid point here and I guess it deserve a Jira ticket. i will try to find some time to handle this special case.

What do you think?

tae-jun:
@anthonycorbacho Nice! I agree with you :)

I will open the issue on JIRA. But since I don't know much about the code structure, I may need some help 😃

anthonycorbacho:
@zjffdu Hey dudy I keep failing at Livy interpreter, i dont know why but i keep getting timeout somehow,  can you take a look t it please? It also look like other pr are failing at the same stage :(

anthonycorbacho:
@tae-jun creating an issue doenst mean that you have to handle it, of course if you want you are welcome to do so, but remember we are a community so that mean we are here to help each others so if you create the PR i can spend some time and work with you on it :)

tae-jun:
@anthonycorbacho Thanks :-) Don't worry! I do this because I want to do 😄

minahlee:
Same Spark 1.3.1 test profile failure exists on master which looks like below:
```
Results :

Tests in error: 
  InterpreterRestApiTest.init:57->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  CredentialsRestApiTest.init:46->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  ZeppelinRestApiTest.init:59->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  NotebookRestApiTest.init:58->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  SecurityRestApiTest.init:44->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  NotebookRepoRestApiTest.init:52->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  ConfigurationsRestApiTest.init:39->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  ZeppelinSparkClusterTest.init:52->AbstractTestRestApi.startUp:233->AbstractTestRestApi.start:201 Â» NullPointer
  NotebookSecurityRestApiTest.init:49->AbstractTestRestApi.startUpWithAuthenticationEnable:229->AbstractTestRestApi.start:201 Â» NullPointer

Tests run: 9, Failures: 0, Errors: 9, Skipped: 0
```
 So I am going to merge this if there is no more discussion.

tae-jun:
I guess it's same CI error with #1518.

There is a log which is:
```bash
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
+echo 'Unable to extract spark-1.3.1-bin-hadoop2.3.tgz'
Unable to extract spark-1.3.1-bin-hadoop2.3.tgz
```

This is because of cache failure of Travis CI, and if it happens, it goes forever. Because caching is done only at the first time.

Because of this, if you follow the log more, you can see:
```bash
SPARK HOME detected null
```
Therefore, Zeppelin cannot find Spark, and it goes to failure.

I think it will pass if you run Travis CI on your own repository.

anthonycorbacho:
Yeaaaaaay!

PR #1573, opened 2016-10-31
[DOC] Add `/ws` for nginx websocket configuration

m30m:
### What is this PR for?
It's a bit misleading that both regular webserver location and websocket location use the exact same address. It's better to specify that websocket requests should be proxy-passed to  `/ws` subdirectory.

### What type of PR is it?
Documentation

### Questions:


Leemoonsoo:
Thanks @m30m for the contribution.
CI failure is irrelevant to this update.

LGTM and merge into master if there're no more discussions.

PR #1574, opened 2016-11-01
ZEPPELIN-1595. Make ZeppelinContext extensible

zjffdu:
### What is this PR for?
For now, ZeppelinContext only support Spark Interpreter. I'd like to make it extensible, so that it can support other interpreters as well. For now, user need to implement the following 3 methods to extend ZeppelinContext

* public List<Class> getSupportedClasses() 
* public abstract Map<String, String> getInterpreterClassMap();
* protected abstract String showData(Object obj);

### What type of PR is it?
[Feature | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1595

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
@@ -1,0 +1,280 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.spark;
+
+import com.google.common.collect.Lists;
+import org.apache.spark.SparkContext;
+import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.catalyst.expressions.Attribute;
+import org.apache.zeppelin.annotation.ZeppelinApi;
+import org.apache.zeppelin.display.AngularObjectWatcher;
+import org.apache.zeppelin.display.Input;
+import org.apache.zeppelin.display.ui.OptionInput;
+import org.apache.zeppelin.interpreter.*;
+import scala.Tuple2;
+import scala.Unit;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.*;
+
+import static scala.collection.JavaConversions.asJavaCollection;
+import static scala.collection.JavaConversions.asJavaIterable;
+import static scala.collection.JavaConversions.collectionAsScalaIterable;
+
+/**
+ * ZeppelinContext for Spark
+ */
+public class SparkZeppelinContext extends BaseZeppelinContext {
+
+
+  private SparkContext sc;
+  public SQLContext sqlContext;
+  private List<Class> supportedClasses;
+  private Map<String, String> interpreterClassMap;
+
+  public SparkZeppelinContext(
+      SparkContext sc, SQLContext sql,
+      InterpreterHookRegistry hooks,
+      int maxResult) {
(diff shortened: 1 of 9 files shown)


zjffdu:
@Leemoonsoo @jongyoul @felixcheung  Please help review.

felixcheung:
In my mind the longer term goal of Zeppelin Context is to support communication between interpreter groups. Would we be going the opposite direction somewhat by having interpreter/interpreter group specific Zeppelin Context, and so they are not longer shared?

zjffdu:
Do you mean ZeppelinContext would support communication between interpreter groups across JVM ? Currently ZeppelinContext only support spark interpreter, I don't see any case that it support communication between interpreter group. Maybe I miss something.

felixcheung:
To clarify, it doesn't right now, but I recall requests for that capability (which would be very useful for some of the more restricting interpreters to pass data around)

I will let others to comment?

zeppelin-interpreter/pom.xml
```diff
@@ -213,6 +213,13 @@
       <exclusions>
       </exclusions>
     </dependency>
+
+    <dependency>
+      <groupId>org.scala-lang</groupId>
+      <artifactId>scala-library</artifactId>
+      <version>${scala.version}</version>
+      <scope>provided</scope>
+    </dependency>
```
Leemoonsoo: Is there any special reason adding scala dependency here?

zjffdu: Because `BaseZeppelinContext` use scala api, and to make it extensible by other interpreters, I move it from `zeppelin-spark` to `zeppelin-interpreter`.

wesm:
In trying to investigate why the Apache Travis CI build queue is extremely backlogged, I found many build requests from this particular PR in https://travis-ci.org/apache/zeppelin/pull_requests

Since this project's build matrix has 11 entries, this can pretty easily jam up the build queue for other projects. I created https://issues.apache.org/jira/browse/INFRA-13469 in an attempt to mitigate this problem in the future

felixcheung:
thanks @wesm for investigating this and following up. it's odd since this PR is 3 months old? though I'll add we should streamline the test matrix a bit (as we have from time to time)

wesm:
@felixcheung it looks like these builds were triggered by a merge on a different branch. see for example https://github.com/apache/zeppelin/commit/e259feacf78770b5dc412c0642216acdce17008f

in looking through your build history, it seems there is quite a bit of merging going on, and you may have been inadvertently triggering a lot of builds because of a quirk in Travis CI and GitHub's hooks: https://github.com/apache/zeppelin/commit/c4cf1abbf06ddf4447dacc886060ea4c1238464e

I would recommend advising the Zeppelin contributors to avoid using `git merge` altogether because of this and prefer instead to rebase or cherry-pick.

zjffdu:
Thanks @wesm , I updated this PR yesterday, but didn't use `git merge`. I use `git rebase` and `git commit --amend`, does these also the issue as well ?

wesm:
Whatever you did generated a merge commit. The proof is right there -- a rebate or amend would not have two parents

zjffdu:
@Leemoonsoo @felixcheung I update this PR, please help help review that. My purpose is to make ZeppelinContext extensible is in 2 perspective. 
* All the interpreter can has its own ZeppelinContext. e.g. My next PR would to implement FlinkZeppelinContext which would display flink DataSet/Table data in zeppelin's table format
* Interpreters can also add customized features for its own ZeppelinContext. e.g. SparkZeppelinContext can add one extention point to allow user to visualize mllib model in zeppelin.

Communication between interpreter group mentioned by @felixcheung above is not implemented in this PR as I think this would bring much large change and need more sophisticated design. But we could put it in `BaseZeppelinContext` in future when we want to implement it.

zjffdu:
@felixcheung @Leemoonsoo Would you mind to review this ?

spark/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
```diff
@@ -0,0 +1,280 @@
...
+    } catch (ClassNotFoundException e) {
+    }
+
+    try {
+      supportedClasses.add(this.getClass().forName("org.apache.spark.sql.SchemaRDD"));
+    } catch (ClassNotFoundException e) {
+    }
+
+    if (supportedClasses.isEmpty()) {
+      throw new InterpreterException("Can not road Dataset/DataFrame/SchemaRDD class");
```
felixcheung: `Can not road ` -> `Cannot load `

Leemoonsoo:
Great work. Looks good to me!

spark/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
zjffdu: Good catch, fixed

zjffdu:
Thanks for review, will merge if no more comments.

PR #1578, opened 2016-11-02
ZEPPELIN-1599 Remove support on some old versions of Spark.

jongyoul:
### What is this PR for?
removing support on old versions of Spark including testing and building them.

### What type of PR is it?
[Feature]

### Todos
* [x] - Remove .travis.yml
* [x] - Remove pom.xml
* [x] - Remove some docs

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1599

No test. Check travis simplified

### Questions:
* Is there breaking changes for older versions? You cannot use spark from 1.1 to 1.3 any longer
* Does this needs documentation? Yes, should remove some docs

Removed some profiles concerning old versions of Spark

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,18 +58,6 @@ matrix:
     - jdk: "oraclejdk7"
       env: SCALA_VER="2.10" SPARK_VER="1.4.1" HADOOP_VER="2.3" PROFILE="-Pspark-1.4 -Pr -Phadoop-2.3 -Ppyspark -Psparkr" BUILD_FLAG="package -DskipTests -DskipRat" TEST_FLAG="verify -DskipRat" TEST_PROJECTS="-pl zeppelin-interpreter,zeppelin-zengine,zeppelin-server,zeppelin-display,spark-dependencies,spark,r -Dtest=org.apache.zeppelin.rest.*Test,org.apache.zeppelin.spark.* -DfailIfNoTests=false"
 
-    # Test spark module for 1.3.1
-    - jdk: "oraclejdk7"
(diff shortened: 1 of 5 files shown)


jongyoul:
@AhyoungRyu Can you take a look into it briefly? I didn't change your part to avoid conflicts.

zjffdu:
`SparkVersion.java`  `SparkInterpreter.java` also needs to be updated.

jongyoul:
@zjffdu Good catch up, but I think if users use Spark 1.3 with SPARK_HOME, we'd better support it. If we change the minimum version of Spark to 1.4 in `SparkVersion`, user cannot use spark 1.3 any more even if user set SPARK_HOME as kind of 1.3. What do you think of it?

jongyoul:
I just want to remove official Spark support by Zeppelin, but if user still want to use old version, we don't have to block their usages.

Leemoonsoo:
@jongyoul Then, I think we can update related doc and say spark 1.1~1.3 support is deprecated.

jongyoul:
@Leemoonsoo I'll remove those support from docs and send an email for this PR.

Leemoonsoo:
I was trying to say mark 'deprecated' in the doc. but it's also good idea to mention it in mailinglist :-)

jongyoul:
@Leemoonsoo We'd better leave some notes in upgrade.md in docs, too.

AhyoungRyu:
@jongyoul Yeah thanks for your consideration. Seems there is no conflict between this patch and #1339 i think. LGTM 👍

minahlee:
I think https://github.com/apache/zeppelin/blob/master/testing/downloadSpark.sh#L79 this line should be updated too.

felixcheung:
Great!

jongyoul:
@minahlee I've fixed what you mentioned.

AhyoungRyu:
@jongyoul [This part](https://github.com/jongyoul/zeppelin/blob/acf514f05eaeedd2f8de0aa4f4002aa258b5aadd/testing/downloadSpark.sh#L30) also needs to be updated accordingly I guess.

tae-jun:
👍

bzz:
Looks great to me.

Nice side-effects is that CI should become faster now.

PR #1585, opened 2016-11-03
ZEPPELIN-1606. Add interpreter option to interpreter-setting.json

zjffdu:
### What is this PR for?
For livy, the default mode should be scoped per use. This PR is trying to add interpreter option to `interpreter-setting.json` to allow user to set the default mode. 

### What type of PR is it?
[Improvement]

### Todos
This PR is just a short term solution. There're several followup work needs to be done.
* Refactor livy interpeter to adapt scope mode. 
* Option should be interpretergroup level setting rather than interpreter level setting. 
* Format of `interperter-setting.json` is not consistent with `Interpereter.json`, there's some code duplication, like `RegisteredInterpereter`,  `InterpreterInfo` and etc. 
* And this PR doesn't address the issue of mode restriction, user can still change to an invalid mode manually.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1606

### Questions:


zjffdu:
\cc @Leemoonsoo @jongyoul  Please help review, this is a short term solution for livy, I would plan to do the followup work in the Todos.

jongyoul:
LGTM. We will simplify registration mechanism by deprecating a old way.

prabhjyotsingh:
LGTM, I like this idea.

zeppelin-interpreter/pom.xml
```diff
@@ -213,6 +213,11 @@
       <exclusions>
       </exclusions>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
```
zjffdu: InterpreterOption needs it.

Leemoonsoo: I think minimizing dependency of `zeppelin-interpreter` is important.
Because every interpreter implementation will have `zeppelin-interpreter` as a dependency, and more dependency in `zeppelin-interpreter` the higher chances of dependency conflict.

For example, spark interpreter may require different `guava` library version depends on SPARK_HOME on runtime.

I think making InterpreterOption independent from `guava` is not going to be too difficult.

zjffdu: Good point, I will update the PR soon

Leemoonsoo:
Thanks @zjffdu for the contribution. LGTM and merge if there're no further discussions.

PR #1587, opened 2016-11-03
[ZEPPELIN-1609] using pyspark(python3) with livy interperter

purechoc:
### What is this PR for?
using pyspark(python3) with livy interperter

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1609

%livy.pyspark3
3/2
check result is "1.5"

### Questions:
* Is there breaking changes for older versions?  - NO
* Does this needs documentation?  - NO


felixcheung:
hi - thanks for the contribution. How is this going to start python3? could you elaborate?

purechoc:
first, you need to install livy and setup python3 environment.
second, set config. "spark-defaults.conf"
spark.yarn.appMasterEnv.PYSPARK3_DRIVER_PYTHON usr/bin/python3

and then, just using livy interpreter with "%livy.pyspark3"

felixcheung:
Hmm, can't we do that with the existing `%livy.pyspark` interpreter?

purechoc:
can't do that,
%livy.pyspark support python2 only.

Leemoonsoo:
@purechoc Thanks for the contribution.

Like @felixcheung mentioned, alternative approach could be adding a interpreter property [here](https://github.com/apache/zeppelin/blob/master/livy/src/main/resources/interpreter-setting.json#L145 ) and let LivyPySparkInterpreter be configured with the property without adding new interpreter class.

What do you think?

purechoc:
@Leemoonsoo 
thanks to review.

first. i try like that.
```
# add properties
  {
    "group": "livy",
    "name": "pyspark3",
    "className": "org.apache.zeppelin.livy.LivyPySparkInterpreter",
    "properties": {
       "zeppelin.livy.kind": {
        "propertyName": "zeppelin.livy.kind",
        "defaultValue": "pyspark3"
      },
    },
    "editor": {
      "language": "python",
      "editOnDblClick": false
    }
  },
```
it is work. but that property assign both "livy.pyspark" and "livy.pyspark3".

if user want to python2 and python3 both,
user need to make new interpreter base on livy group and set property  zeppelin.livy.kind=pyspark3.

if, this is better than adding new interpreter class, i fixed code and commit soon.

Leemoonsoo:
Thanks for explain. but i'm little bit confused.
If you are creating `LivyPySpark3Interpreter` in addition to `LivyPySparkInterpreter`, there will be no need to add property `zeppelin.livy.kind`.

But by adding a property, `LivyPySparkInterpreter` interpreter can read property and change it's behavior which version of python to use, in runtime. and no need to create `LivyPySpark3Interpreter`. 
In this case, to use both python2 and python3 at the same time, user can create multiple interpreter settings in the GUI and use them.

I have no strong opinion on any approaches, but at least `python`, `pyspark` interpreter handle python version using property instead of having separate interpreter class.

@purechoc @zjffdu @prabhjyotsingh @felixcheung What do you think?

purechoc:
@Leemoonsoo . 
thanks to comment.

you mean like this?

if user add property `zeppelin.livy.kind` = pyspark, "%livy.pyspark" working on `python2`. 
if user add property `zeppelin.livy.kind` = pyspark3, "%livy.pyspark" working on `python3`.

like a set `zeppelin.python` = python or python3 ``pyspark `interpreter`
but if user want to `python2 `and `python3 `same time `python` interpreter. user need to create new interperter in the GUI.
```
interperter name = python3
interperter group = python
zeppelin.python = python3

%python3
2/3 = 1.5
```

if live using this approch, user using livy interperter like this.
```
# run python2
%livy.pyspark

# run python3
%livy_python3.pyspark
```

add property `zeppelin.livy.kind` and set python version if user want to run. (each time. not same time)
is this right?

Leemoonsoo:
I see what you mean. Thanks for the explanation. Make sense.
LGTM

felixcheung:
I think I can review this in more details, but is there a need to run both Python2 and Python 3 at the same time in the same environment? Most likely people will have chosen one?

If one at a time is ok why not do this with an environment variable LIVY_PYTHON - this is the similar approach Spark has chosen - and must also apply here since Livy just kick off PySpark anyway. So there really only one can be set at a time.

Or am I missing something?

purechoc:
for example.
in our environment. (zeppelin + livy)
there is 3 team, (teamA, B, C)
teamA using python2, 
teamB, C using python3,
there are using same enveroment (zepplein + livy)

so. livy offer pyspark and pyspark3.  this is livy's feature.
watch this : https://github.com/cloudera/livy/commit/d3d203160436abb37b1136035d534f3d1c1a0533

i dont't understand livy.pyspark and livy.pyspar3 using same time.
(pyspark3 is livy's kind feature)

purechoc:
@zjffdu @felixcheung  @Leemoonsoo 
could you merge about this?
or need to fix something?

zjffdu:
I prefer to use `%livy.pyspark` & `$livy.pyspark3` instead of `zeppelin.livy.kind`.  Because if user change `zeppelin.livy.kind` some paragraphs may not work again, this is very confusing. So explicitly specifying python version is better IMO.  And could you wait for #1612 which is refactoring livy interpreter ? After #1612 , it should be pretty easy to add `%livy.pyspark3`

purechoc:
@zjffdu 
thanks to comment.
i am wait #1612 
 #1612 is LGTM

purechoc:
refector code based on  #1612

zjffdu:
lgtm

purechoc:
@felixcheung  @Leemoonsoo 
please review this.
thanks

livy/src/main/resources/interpreter-setting.json
```diff
@@ -160,6 +160,17 @@
   },
   {
     "group": "livy",
+    "name": "pyspark3",
+    "className": "org.apache.zeppelin.livy.LivyPySpark3Interpreter",
+    "properties": {
+    },
```
felixcheung: should this have a `option` group like `"org.apache.zeppelin.livy.LivyPySparkInterpreter` above?

felixcheung:
shouldn't this be listed here: https://github.com/apache/zeppelin/blob/master/conf/zeppelin-site.xml.template#L193

purechoc:
thank to address.

update code completed

felixcheung:
LGTM
merging if no more comment

purechoc:
i think CI failure cause is not this PR.
some other reason in other testcase.

felixcheung:
all green!
merging if no more comment

PR #1590, opened 2016-11-03
[ZEPPELIN-1612] Fix NPE when initializing Notebook

khalidhuseynov:
### What is this PR for?
Sometimes Zeppelin wasn't able to start because of empty subject when initializing Notebook class, more details in issue.

### What type of PR is it?
Bug Fix

### Todos
* [x] - add anonymous subject
* [x] - add test

### What is the Jira issue?
[ZEPPELIN-1612](https://issues.apache.org/jira/browse/ZEPPELIN-1612)

* added test passing and no relevant CI failures
* also can be starting Zeppelin in anonymous mode when you have notes with angular objects

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
@@ -209,8 +209,7 @@ public class NotebookTest implements JobListenerFactory{
       p1.setText("hello world");
       note.persist(anonymous);
     } catch (IOException fe) {
-      logger.warn("Failed to create note and paragraph. Safe to return, other tests might be failing as well", fe);
-      return;
+      logger.warn("Failed to create note and paragraph. Possible problem with persisting note, safe to ignore", fe);
     }
 
     try {
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
@@ -65,7 +65,7 @@ public class NotebookTest implements JobListenerFactory{
   private DependencyResolver depResolver;
   private NotebookAuthorization notebookAuthorization;
   private Credentials credentials;
-  private AuthenticationInfo anonymous = new AuthenticationInfo("anonymous");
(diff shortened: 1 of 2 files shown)


zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -197,6 +197,31 @@ public void testReloadAllNotes() throws IOException {
...
+      Paragraph p1 = note.addParagraph();
+      Map config = p1.getConfig();
+      config.put("enabled", true);
+      p1.setConfig(config);
+      p1.setText("hello world");
+      note.persist(anonymous);
+    } catch (IOException fe) {
+      logger.warn("Failed to create note and paragraph. Safe to return, other tests might be failing as well", fe);
+      return;
+    }
```
bzz: The error message above implies that it is intended as failing condition for this test, but `return` will actually make test pass.

What was the intention? I think it's fine to explicitly fail the test if some pre-condition fail

bzz:
Just wonder if that means that notes are always load as `anonymous` irregarding Zeppelin mode, because we do not have such permissions yet?

Curious, as I would expect that we would load all notes for a specific user, in case auth is enabled in Zeppelin..

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
khalidhuseynov: Actually if precondition fails it means the problem is in file saving routine and not related to testing this certain function. In this case we have option of being 1) strict and fail this test as well, or 2) being less strict since failing precondition in this case means that `note.persist` is failing and probably test for that function should be failing as well. I chose here second option initially, but it makes sense to make it more strict; i'll change it then.

khalidhuseynov:
@bzz that function `loadAllNotes` is called only from constructor and when loading from constructor you're either in anonymous mode, or in shiro and loading all notes by anonymous (public). For the cases when specific user logs in another function for reloading note in [here](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java#L479) is called with specific user credentials.

bzz:
Got it! Thank you for kind explanation.

CI failure looks not relevant for the change set.

Looks great to me, merging to master if there is no further discussion.

PR #1593, opened 2016-11-04
ZEPPELIN-1613. PerUser scoped mode doesn't work

zjffdu:
### What is this PR for?
For perUser scoped mode, difference users still use the same interpreter instance which is incorrect. This is due to they are using the same interpreter instance key.  This PR will fix this by using different interpreter instance key for different users. Besides, I also did some refactoring and add more log.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1613

Tested manually. 

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -50,7 +50,6 @@ import java.util.Set;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.internal.StringMap;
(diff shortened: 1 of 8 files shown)


zjffdu:
\cc @jongyoul @Leemoonsoo  Please help review.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -1172,28 +1172,28 @@ private String getInterpreterInstanceKey(String user, String noteId, Interpreter
     String key;
     if (option.isExistingProcess()) {
       key = Constants.EXISTING_PROCESS;
-    } else if (!option.perNoteShared()) {
-      key = noteId;
-      if (shiroEnabled && !option.perUserShared()) {
-        key = user + ":" + key;
-      }
+    } else if (option.perNoteScoped()) {
```
zjffdu: hmm, right, I think I miss that combination, will update the PR soon.

prabhjyotsingh:
Thanks for creating this issue, this wasn't working for me either.

zjffdu:
@jongyoul I fix the combination issue and also did some refactoring. 
* rename interpreterInstanceKey to sessionKey

zjffdu:
@jongyoul @Leemoonsoo Could you help review ?

Leemoonsoo:
@zjffdu Looks great to me!

PR #1598, opened 2016-11-04
[ZEPPELIN-1618] [WIP] Support ElasticSearch 5.0

1ambda:
### What is this PR for?

ElasticSearch was released in Oct 26 and It requires 5.x client driver but zeppelin currently use 2.x only.

- we need 2 different profiles for 2.x and 5.x (there are no 3.x, 4.x)
- 5.x client require different codebase. In this case we should keep 2 version of code at the same time like the spark code.
- 5.x client requires JAVA 8 at least

### What type of PR is it?

Improvement

### Todos
* [x] - create profiles for 2.x and 5.x
* [x] - setup JAVA 8 validation for 5.x 
* [x] - update default `elasticsearch.version` to 2.4.x (the most recent for 2.x)
* [x] - update license
* [x] - refactor
* [x] - create the 5.x interpreter
* [x] - loading dynamically
* [x] - hide unused classes using maven profile
* [ ] - write tests for 5.x
* [ ] - update docs (build, usage, upgrade)

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1618

Some tests will be included.

- `mvn clean test -pl 'elasticsearch' -Pelasticsearch-2`
- `mvn clean test -pl 'elasticsearch' -Pelasticsearch-5`

screen shot 2016-11-04 at 9 46 23 pm
elasticsearch

### Questions:
* Does the licenses files need update? - YES
* Does this needs documentation? - YES


1ambda:
Ref - https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_java_api_changes.html

1ambda:
2+ Solutions. 

#### 1. Use just reflection

- Pros: We can have only 1 interpreter `ElasticsearchInterpreter` 
- Cons: but too much reflection code :( 

#### 2. Create `Elasticsearch5Interpreter` for version 5.x 

- Pros: We can remove reflection code by ignoring the interpreter code based on the maven profile
- Cons: Code might be duplicated but we can extract them as an interface. For example, `ElasticsearchInterpeterBase`

1ambda:
Blocked until 5.1 release due to too many undocumented breaking changes

bzz:
Great work @1ambda ! Do you know if there are any ETA for Elasticsearch 5.1?

bbonnin:
I have just one comment: it's about some duplicate lines of codes. I think some lines that we can find in both ElasticSearchXConnectors could be put in the base class, in order to keep only the specific part ES v2 and V5 in the sub-classes. An example is the method `executeDeleteQuery` that is almost the same in both classes.
Otherwise, looks good to me.

1ambda:
@bbonnin Thanks for the quick review :) 

1. Duplicated code in `executeDeleteQuery` is due to incompatible API between 2.x and 5.x 

```
// 2.x
public boolean isFound() { ... } 
// 5.x
RestStatus.NOT_FOUND != response.status())
```

But we can extract duplicated validation logic to the base class as you mentioned.

2. @bzz, @bbonnin  Regarding to update docs, 

Are there more files I should update? I am asking because 

- new maven profile is added `elasticsearch-2` (activated by default), `elasticsearch-5`
- `elasticsearch-5` profile requires java 8 +

stevenmccord:
@1ambda were you still working through this one?  I was going to give it a test run if this is something that might be getting in.  Noticed it hasn't been updated with master in a while, so just wanted to check to see if it might be worthwhile to give it a try.  Thanks!

1ambda:
@stevenmccord Hi :)

I am trying to fix other urgent issue. So it's ok to close this PR if you want to implement. 
The reason I am keeping this issue is sharing info about ES 5.0

jaspinderdineout:
Is it live? Does zeppelin support elasricsearch 5.0 ???

zjffdu:
\cc @1ambda

1ambda:
@jaspinderdineout Hi :)

Zeppelin doesn't support es 5.0 currently as i know. 
I will close this PR. This issue can be easily handled after #1902.

brbrown25:
@1ambda recently elastico deprecated http and requires https now, is there any plan to upgraded this connector?

PR #1595, opened 2016-11-04
[ZEPPELIN-1615] - Zeppelin should be able to run without Shiro

anthonycorbacho:
### What is this PR for?
Right now, Zeppelin use Shiro by default even if you dont need it.
(It will use shiro.ini.template file if it doenst find shiro.ini), this behaviors is a little flacky and we should start zeppelin without shiro context if user doenst want to use it.

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [x] - Update configuration - Return empty if shiro file not found
* [x] - refactor Rest Api handler, if shiro.ini not found start a handler without shiro context
* [x] - refactor SecurityUtils to handle the case of shiro is disabled.

### What is the Jira issue?
 * [ZEPPELIN-1615](https://issues.apache.org/jira/browse/ZEPPELIN-1615)

### How should this be tested?
Start zeppelin without shiro.ini

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1599, opened 2016-11-04
[HOTFIX] Always download spark binary package from archive

minahlee:
### What is this PR for?
* Not download Spark for first profile which does license check only
* Always download Spark from archive not mirror
   - We need to check which spark versions are in mirror or not, and update [this line](https://github.com/apache/zeppelin/blob/master/testing/downloadSpark.sh#L79) which is unsustainable
    - Sometimes mirror site has problem such as `Not Found`, like we have issue in CI right now.
* Remove unused variable `SPARK_VER_RANGE` from `testing/downloadSpark.sh` (https://github.com/apache/zeppelin/pull/1578#issuecomment-258122087)

> Note: CI will still fail until #1595 is merged

### What type of PR is it?
Hot Fix

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
   include:
     # Test License compliance using RAT tool
     - jdk: "oraclejdk7"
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
LGTM. Lets merge.

minahlee:
Merging it as hotfix

PR #1608, opened 2016-11-07
[ZEPPELIN-1633] Flink: Set maxClassfileName to 128

PhilippGrulich:
### What is this PR for?
This bug was fixed for the spark interpreter in 
https://issues.apache.org/jira/browse/ZEPPELIN-926, but the bug can also occur in the Flink interpreter.
So we have to migrate the patch to the flink interpreter.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1633

### Questions:


Leemoonsoo:
Thanks @PhilippGrulich for the fix.
LGTM and merge to master if there're no further discussions.

PR #1611, opened 2016-11-07
ZEPPELIN-1625 - Add support of user specific properties

benoyantony:
### What is this PR for?
User should be able to define properties and these properties must be passed to Interpreter via Interpreter Context.
Add functionality so that Interpreter can substitute the its property value with the value specified by the user.

### What type of PR is it?
Feature

### What is the Jira issue?
ZEPPELIN-1625

unit  tests provided 

### Questions:


1ambda:
Hi 😃  @benoyantony!

I'v checked out your PR and tested, but zeppelin doesn't start rendering 503 page.

```
$ zeppelin-master git:(pr/1611) mvn clean package -DskipTests
$ ...
```

screen shot 2016-11-08 at 11 28 50 am

benoyantony:
Opened a new pull - https://github.com/apache/zeppelin/pull/1623

PR #1613, opened 2016-11-08
ZEPPELIN-1643:Make spark web UI accesible from interpreters page

karup1990:
### What is this PR for?
Make spark web UI accesible from interpreters page

### What type of PR is it?
Improvement

### Todos
NA

### What is the Jira issue?
ZEPPELIN-1643

Start sparkcontext.
Goto interpreters page. Corresponding to the spark interpreter used in notebook, there will be *spark ui* button.
Clicking the button should open a new tab with the application's spark web UI.

zscpr58qdi

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
@@ -22,6 +22,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.log4j.Logger;
 import org.apache.zeppelin.display.AngularObjectRegistry;
+import org.apache.zeppelin.interpreter.remote.RemoteInterpreterEventClient;
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreterProcess;
 import org.apache.zeppelin.resource.ResourcePool;
 import org.apache.zeppelin.scheduler.Scheduler;
@@ -49,6 +50,8 @@ public class InterpreterGroup extends ConcurrentHashMap<String, List<Interpreter
   RemoteInterpreterProcess remoteInterpreterProcess;    // attached remote interpreter process
   ResourcePool resourcePool;
   boolean angularRegistryPushed = false;
+
+  private RemoteInterpreterEventClient eventClient;
 
   // map [notebook session, Interpreters in the group], to support per note session interpreters
   //Map<String, List<Interpreter>> interpreters = new ConcurrentHashMap<String,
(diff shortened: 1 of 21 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
```diff
@@ -50,6 +51,8 @@
   ResourcePool resourcePool;
   boolean angularRegistryPushed = false;
 
+  private RemoteInterpreterEventClient eventClient;
```
Leemoonsoo: `InterpreterGroup` is not only used in RemoteInterpreter process but also in ZeppelinServer side, too.
So having reference to eventClient in `InterpreterGroup` is not optimal i think.

I think `InterpreterContext` is better place to have a such reference, and you can inject your object in when [InterpreterContext is being created](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java#L544).

InterpreterContext is not passed to interpreter in open(). So getting routines that getting SparkUI and update meta need to be moved from open() to interpret(). I think it's not a big problem in this use case.

Also i think it'll be better passing some wrapper object instead of eventClient directly. Because of we don't want to let interpreter directly access all eventClient's methods.

karup1990:
@Leemoonsoo Thanks for the suggestions.
I have refactored the code now.

karup1990:
Ready for review

karup1990:
Failure seems unrelated.
`03:55:19,906 ERROR org.apache.zeppelin.AbstractZeppelinIT:136 - Exception in ParagraphActionsIT while testEditOnDoubleClick `

Leemoonsoo:
LGTM. Merge if there're no further discussions.

zjffdu:
If I read it correctly, this PR assume each interpreter setting has one spark ui url ? But I think this is wrong assumption, as we may have multiple spark applications for single interpreter setting. @karup1990 @Leemoonsoo

Leemoonsoo:
@zjffdu Right. Isolated mode can create multiple spark applications.
@karup1990 @zjffdu how can we handle multiple spark applications?

zjffdu:
I think the right place to put spark ui link is in note page instead of interpreter setting page.

Leemoonsoo:
Make sense. @karup1990 what do you think?

karup1990:
@zjffdu @Leemoonsoo Thanks for letting me know.
  I will make the changes to handle isolated scope as well and create a PR

karup1990:
I did a code reading to see how to fix this.. Looks like I will need to handle it per user as well.
To complete this PR, I can quickly make the changes to show the link in interpreter page, only for per note scoped and global mode. 
Else , I will try to fix all modes(user, note, global) and then create a PR.
@zjffdu @Leemoonsoo  let me know your thoughts.

karup1990:
@zjffdu @Leemoonsoo I have handled the cases in https://github.com/apache/zeppelin/pull/1678

PR #1618, opened 2016-11-09
Add tests with external python dependencies to CI build

agoodm:
### What is this PR for?
Many of our python tests are excluded from the main build profile and hence, the CI builds as well because they require external dependencies. This PR modifies `.travis.yml` to download the needed python dependencies for our tests (`matplotlib`, `pandas`, and `pandasql`) using `conda`. 

### What type of PR is it?
Improvement

### Todos
* [ ] - matplotlib test for `PySparkInterpreter`

### What is the Jira issue?
[ZEPPELIN-1639](https://issues.apache.org/jira/browse/ZEPPELIN-1639)

Make sure newly added python tests (for matplotlib and pandas) are passing on CI!

### Questions:


bzz:
Looks like CI is stalled during PySpark execution somehow..

```
16/11/09 09:46:13 INFO PySparkInterpreter: File /tmp/zeppelin_pyspark-7835708002435081304.py created


The build has been terminated
```

agoodm:
This CI failure is not unrelated, it also happened on my fork. I have a hunch that changing the path might have something to do with it, which would be odd since I regularly use Zeppelin with conda-based python installations with no issues all the time. I'll keep testing this further.

PR #1623, opened 2016-11-11
ZEPPELIN-1625  add support for user specific properties

benoyantony:
### What is this PR for?
User should be able to define properties and these properties must be passed to Interpreter via Interpreter Context.
Add functionality so that Interpreter can substitute the its property value with the value specified by the user.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1625

unit tests provided
Manual testing steps:
Add a property, update the property and then delete it

Available in documentation

### Questions:
* Does the licenses files need update?  NO
* Does this needs documentation? Yes, Documentation is provided.

### DIFF
--- a/docs/_includes/themes/zeppelin/_navigation.html
+++ b/docs/_includes/themes/zeppelin/_navigation.html
@@ -98,12 +98,16 @@
                 <li><a href="{{BASE_PATH}}/rest-api/rest-notebook.html">Notebook API</a></li>
                 <li><a href="{{BASE_PATH}}/rest-api/rest-configuration.html">Configuration API</a></li>
                 <li><a href="{{BASE_PATH}}/rest-api/rest-credential.html">Credential API</a></li>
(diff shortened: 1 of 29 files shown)


1ambda:
@benoyantony

Hi, I am trying to test this feature. May I ask 2 things about this PR? 

- What is possible if this feature is merged in terms of usage? (e.g user scenario)
- Are there another features which can be implemented upon this PR?

benoyantony:
thanks @1ambda , Sorry for the delay.

1. If this feature is merged, then the properties can be referred in the interpreter properties. Each user can override the property value.

2. Once the feature is merged, I will have another pull request on the Livy Interpreter which make use of the feature for its properties. 

I have tested this feature at our company.  Attached is screenshot where the administrator setup the Livy property - livy.spark.yarn.queue to ${queue,defaultQueue}

livy-interpreter-properties


In the second screenshot, the  user override the queue as benoyqueue. If the user does not override, then the property will remain as defaultQueue.
 
user-defined-properties

PR #1628, opened 2016-11-13
[HOTFIX][ZEPPELIN-1656] z.show in Python interpreter does not work

agoodm:
### What is this PR for?
There have been reports of #1534 causing the python interpreter to always show an error because `z` is not being set. As it turns out this is a result of improperly handling the case when matplotlib isn't found when initializing the interpreter.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1656](https://issues.apache.org/jira/browse/ZEPPELIN-1656)

Run any simple python paragraph.

### Questions:

### DIFF
--- a/python/src/main/resources/bootstrap.py
+++ b/python/src/main/resources/bootstrap.py
@@ -207,7 +207,7 @@ class PyZeppelinContext(object):
         try:
             import matplotlib
         except ImportError:
-            pass
+            return
         # Make sure custom backends are available in the PYTHONPATH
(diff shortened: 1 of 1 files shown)


agoodm:
@bzz please test ASAP!

bzz:
Will test right now and post back

bzz:
Tested with and without `matplotlib` - both cases work fine!

Also

```
%python

a = 'a'
print "{}".format(a)
```

works as expected as well!

👍  thank you for prompt fix, looks awesome to me, will merge to master as hotfix asap, if there is no further discussion

PR #1629, opened 2016-11-14
[ZEPPELIN-1659] DON'T reset filter whenever each interpreter setting is updated

1ambda:
### What is this PR for?

Filter is cleared whenever each interpreter setting is updated. 
it's sometimes annoying if an user want to debug interpreter settings (I attached GIF)

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?

[ZEPPELIN-1659](https://issues.apache.org/jira/browse/ZEPPELIN-1659)

1. insert interpreter filter
2. update a interpreter setting
3. check whether the filter is cleared or not

filter_is_cleared_everytime

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.controller.js
+++ b/zeppelin-web/src/app/interpreter/interpreter.controller.js
@@ -24,6 +24,7 @@
     $scope.availableInterpreters = {};
     $scope.showAddNewSetting = false;
     $scope.showRepositoryInfo = false;
+    $scope.searchInterpreter = '';
     $scope._ = _;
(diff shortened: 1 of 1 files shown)


1ambda:
After fixed:

filter_is_now_not_cleared

1ambda:
What if we keep filter after interpreter creation?

Leemoonsoo:
That was annoying.  Thanks @1ambda  for the fix.
Tested and LGTM.

Merge if there're no more comments.

PR #1633, opened 2016-11-15
Fix formatting/typos/style in scio doc

ravwojdyla:
### What is this PR for?
Fix formatting/typos/style in scio doc

### What type of PR is it?
Documentation

### DIFF
--- a/docs/interpreter/scio.md
+++ b/docs/interpreter/scio.md
@@ -59,7 +59,7 @@ $beam.scio
 val (sc, args) = ContextAndArgs(argz)
 ```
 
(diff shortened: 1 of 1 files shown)


ravwojdyla:
Is jira ticket needed for this?

AhyoungRyu:
@ravwojdyla Thanks for taking care of Scio interpreter docs. Regarding the Jira ticket, usually it's not a mandatory but good to have one since it enables to track the issues. But I think you don't need to in this case :)
LGTM, so will merge if there are no more discussions.

PR #1640, opened 2016-11-15
Enable to search interpreters only based on the name

AhyoungRyu:
### What is this PR for?
Currently when you're trying to search some interpreters, for example "Spark", the page shows both "Livy" and "Spark". It's because Livy contains Spark related properties such as `livy.spark.driver.cores`, `livy.spark.driver.memory` and etc etc. As an user aspect, I felt a bit uncomfortable with this.
So I added one more condition so that it can search only based on interpreter name. 

### What type of PR is it?
Improvement

### What is the Jira issue?
no Jira issue for this

 - Before
before
 
 - After 
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.html
+++ b/zeppelin-web/src/app/interpreter/interpreter.html
@@ -90,7 +90,7 @@ limitations under the License.
 </div>
 
 <div class="box width-full"
-     ng-repeat="setting in interpreterSettings | orderBy: 'name' | filter: searchInterpreter" interpreter-directive>
+     ng-repeat="setting in interpreterSettings | orderBy: 'name' | filter: {name:searchInterpreter} " interpreter-directive>
   <div id="{{setting.name | lowercase}}">
     <div class="row interpreter">
 


1ambda:
I'v just tested. It works like @AhyoungRyu described. 
It was annoying before this PR because filter investigated all fields in the interpreter :(

- create 2 JDBC interpreter named `postgres` and `duplicated` that have postgres dep 
- insert filter `duplicated` 
- check whether the `duplicated` interpreter is only selected

screen shot 2016-11-15 at 11 53 26 pm
screen shot 2016-11-15 at 11 53 33 pm

1ambda:
I think this PR is cool :) @AhyoungRyu One more question, 

What do you think if we can filter out interpreters using keywords like `%dep` `%r`?

AhyoungRyu:
@1ambda Thanks for your suggestion. That would be better I think. Let me update then :)

AhyoungRyu:
@1ambda I can address your feedback by changing `filter: {name:searchInterpreter}` to `filter: {interpreterGroup: {name: searchInterpreter}}`. In this case, if we type `pyspark`, then Livy interpreter will show above Spark again. Or if we type `sql`, then many interpreters that have `sql` as interpreterGroup name(e.g. `Bigquery`, `Jdbc`, `Spark`..) will show up. So current status would be enough I guess. What do you think?

1ambda:
@AhyoungRyu 

If so, i think this is enough. Maybe we need a category dropbox to specify search area later. 

Thanks for really needed improvement :)

AhyoungRyu:
@1ambda yeah that could be available approach in the future. Thanks for your review! 
Then will merge if there are no more discussions.

PR #1641, opened 2016-11-15
[ZEPPELIN-1664] Built-in chart doesn't render and resize height doesn't work

Leemoonsoo:
### What is this PR for?
Built-in chart doesn't render and resize height doesn't work

### What type of PR is it?
Hot fix

### Todos
* [x] - Fix problem

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1664

run multiple times and see if graph is rendered correctly.

```
%spark
Thread.sleep(1000)
println(s"""%table key\tvalue\nsun\t50\nmoon\t20\n""")
```

try change height of paragraph and see if graph is updated

### Questions:


Leemoonsoo:
Merge to master as a Hotfix.

PR #1653, opened 2016-11-18
Relative path on windows getAbsolutePath() and new URI causes excepti…

kenshalo:
### What is this PR for?
Relative path on windows getAbsolutePath() and new URI causes exception and most test cases to fail. Switched to create new File() and use f.toURI() so that test cases pass on windows

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?

Run zengine JUnit test cases in on windows with relative path

### Questions:
No
No
No


Leemoonsoo:
Thanks @kenshalo for the fix.
LGTM. If you can create an issue, it will help people track when this problem has been fixed.

Leemoonsoo:
merge to master if there're no further discussions.

PR #1656, opened 2016-11-18
Fix Yarn configuration of Spark2.0

astroshim:
### What is this PR for?
`yarn-client` is deprecated since Spark 2.0, so SparkInterpreter also support `yarn` as `application master`.

### What type of PR is it?
Bug Fix | Improvement

Run spark paragraph with Spark2.0 + yarn.

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -319,6 +319,7 @@ public class SparkInterpreter extends Interpreter {
 
     conf.set("spark.scheduler.mode", "FAIR");
     conf.setMaster(getProperty("master"));
+    conf.set("master", "yarn");
     conf.set("spark.submit.deployMode", "client");
 
     Properties intpProperty = getProperty();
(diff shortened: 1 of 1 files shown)


astroshim:
review please~

Leemoonsoo:
Thanks @astroshim for the fix.
LGTM

Leemoonsoo:

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -296,6 +296,10 @@ private DepInterpreter getDepInterpreter() {
     return (DepInterpreter) p;
   }
 
+  private boolean isYarnMode() {
+    return getProperty("master").equals("yarn-client") || getProperty("master").equals("yarn");
```
felixcheung: is this because we don't support master == "yarn-cluster"?

astroshim: Yes, you're right. It's because Zeppelin don't support `yarn-cluster`yet.
You think it's better to include `yarn-cluster` too?

felixcheung:
Perhaps just check for master starts with "yarn"?

astroshim:
@felixcheung It makes sense. Thank you.

Leemoonsoo:

PR #1639, opened 2016-11-15
[ZEPPELIN-1667] add support for es6

felizbear:
### What is this PR for?
This PR adds support for the current ECMAScript standard [es6](http://www.ecma-international.org/ecma-262/6.0/) via [babel](http://babeljs.io/) ([babel-grunt](https://github.com/babel/grunt-babel) plugin)

### What type of PR is it?
Improvement

### Todos
* [x] - Add babel presets for es6 (aka es2015) and object spread operator
* [x] - Add eslint rules to lint es6 correctly
* [x] - Add `babel` task to grunt and modify `serve`, `test` and `build` tasks

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1667

### How should this be tested?
1. Run `grunt serve` and make sure it works properly (including reload on change)
2. Run `grunt test` and make sure tests pass
3. Run `grunt build` and make sure it builds and runs correctly

PR #1659, opened 2016-11-19
[ZEPPELIN-1667] Fix maven options while updating node, npm packages

1ambda:
### What is this PR for?

Update maven plugin, node vesion, npm package and fix tests to make #1639 pass CI

### What is the Jira issue?

Sub issue of #1639 

### Questions:

Does the licenses files need update? NO
Is there breaking changes for older versions? YES (front-end build tool usage changed)
Does this needs documentation? YES (front-end build tool usage changed)


1ambda:
```
// https://api.travis-ci.org/jobs/177226860/log.txt?deansi=true

[INFO] Zeppelin: Server ................................... SUCCESS [ 15.562 s]
[INFO] Zeppelin: Packaging distribution ................... SKIPPED
[INFO] Zeppelin: Scalding interpreter ..................... SKIPPED
[INFO] Zeppelin: Examples ................................. SKIPPED
[INFO] Zeppelin: Example application - Clock .............. SKIPPED
[INFO] Zeppelin: Example application - Horizontal Bar chart SKIPPED
[INFO] BUILD SUCCESS
[INFO] Total time: 10:02 min
[INFO] Finished at: 2016-11-19T08:09:15+00:00
[INFO] Final Memory: 188M/461M
[ERROR] GC overhead limit exceeded -> [Help 1]
```

zeppelin-web/bower
```diff
@@ -1,17 +0,0 @@
-#
```
bzz: Why removing these files?

So far our frontend build structure did not assume having bower or grunt installed globally, but both got installed locally though maven plugin. 

So how shall we a developer run them now?

bzz:
@1ambda great effort!

As mentioned in #1639 - could you please re-purpose (interactive rebase, basically) this PR to contain only CI improvements?

One question on Node version update: it's great to be updated, but why this particular version? It would be awesome (although, not necessary) to have similar one to other Apache projects i.e BigTop, Ambari, etc. AFAIK they are still using a bit earlier one now.

Other than that - it looks great to me, will be happy to mere it ASAP after 3 issues above are addressed, if there is no further discussion.

1ambda:
rebased

zeppelin-web/bower
1ambda: @bzz 

I think this cause breaking in term of usage for some people who already use `./grunt test`, `./bower install`

Executing those shell files, however, is not convenient. So I think many users install their own grunt-cli and bower globally so that they can type just `grunt test` and `grunt build` 

IMO, 

- Using locally installed bower and grunt through npm scripts is more natural, common way for front-end users nowdays.
- And if we are moving to webpack from grunt to get reduced build time and other useful things, this is necessary change.

In this case, we need documentation and should notice to our front-end developers

1ambda:
rebased interactively not to contain the same commits in #1639

1ambda:
@bzz Thanks for review! Things you pointed out are very important. So I'v done some research about node and npm

The above screenshot is about LTS schedule. 

- Some apache projects use 4.x. but Active LTS will be end before 2017.04. So they (should) move to 6.x If they care about what they are building.
- I used node [6.9.1](https://nodejs.org/en/blog/release/) in this PR. This is one of LTS version for 6.x that means, it (6.x) is somewhat stabilised. ([6.0.0 released in 2016.04](https://nodejs.org/en/blog/release/v6.0.0/))
- **Basically, we use node as a tool**. **not the runtime**. In other words, we need something can boost our productivity instead of old one. For example, writing to build tools easily using es2015 or supporting fast compilation. ([6.9.1 supports es2015 99%](http://node.green/))

Regard to npm

- [npm 3.0.0 allow us to our dependencies to be installed maximally flat](https://github.com/npm/npm/releases/tag/v3.0.0). (will increasing download speed + support cache functionalities)
- So we might use [node 5.x which shipped npm 3.x](https://nodejs.org/en/blog/release/v5.0.0/) but 5.x is not LTS
- So use 3.10.10 which is the most recent one. Even, we can use [4.0.3](https://github.com/npm/npm/releases) just released 2 days ago

felizbear:
@1ambda great analysis!

> Basically, we use node as a tool. not the runtime

This makes a lot of sense

1ambda:
https://travis-ci.org/apache/zeppelin/builds/177545019 (before caching bower, npm, node dc426b3)

screen shot 2016-11-21 at 11 56 51 am

bzz:
Thanks for improving CI @1ambda 

Looks great to me, 
will be happy to merege asap, right after [zeppelin-web/README.md](https://github.com/apache/zeppelin/blob/master/zeppelin-web/README.md) gets updated on new frontend build process!

1ambda:
README is updated. You can preview 

- https://github.com/1ambda/zeppelin/blob/239c74a3100031bf615b879bf589719a78e9ba71/zeppelin-web/README.md

1ambda:
README is updated. 

- (preview) https://github.com/1ambda/zeppelin/blob/ece898e982360aae05b899a02de9e8f8da095759/zeppelin-web/README.md

cloverhearts:
This is my personal opinion.
For the Node.js 6x or 7x version, there is no problem in itself.
However, I know that many of the libraries in Node.js have problems starting from version 6x or later, or are not being built.
(I think this is due to a rapid change in the 6x version.)

If you are upgrading to the 6x version first, we will greatly reduce the choice of many libraries now supported by the node.js community.
On the other hand, I also agree with you that the current version of node.js is too low.

However, we do not use node.js directly, we use it for dependency management and build management.
So, what do you think about upgrading to the 5x or 4x version first?

(Indeed, since the 6x LTS version is constantly changing over time, Zeppelin also has to take the uncertainty of Node.js along with the change)

I'm sorry
I'm not good at English.
I apologize to you if I have been rude.
Thank you.

1ambda:
@cloverhearts 

Thanks for your comment. I have some questions. 

- How we can check one of our dependencies doens't work on 6.x but works in 4.x even build passes?
- As I know, all node modules are dev dependencies and doesn't affect on frontend libraries (bower packages). They are just concated and added to HTML. Are not they? ([Gruntfile.js of es6 PR](https://github.com/felizbear/zeppelin/blob/f083492cfb22073732ee4af70d1cb1d02b743706/zeppelin-web/Gruntfile.js))

cloverhearts:
:)
I listened to your feedback, and tested the current version.
Surely your opinion is correct.

I personally have experienced problems with versioning up tools and environments.
So I wrote a comment.

I have verified the build test for node 6 and the behavior of the legacy library.
There were some problems, but it was a problem that is happening in the current version.
I think this is not a problem in upgrading node 6.
I think that's why we should upgrade to node 6 rather.
(Also for ES6 and other reasons)

Thank you for listening to my comments.
I agree to upgrade the Node 6 version.

LGTM :)

bzz:
@asanjar node version is bumped to 6.x TLS one, please let us know if that works for you you on Power arch.

1ambda:
rebased to resolve conflict with #1666

1ambda:
retrigger CI due to failures of irrelevant tests

bzz:
Great, thank you @1ambda !

Looks great to me, @asanjar if there is no more input on Node version (this PR uses v6, latest LTS) - merging it to master.

asanjar:
+1 
sorry for the delay in responding. you guys have been superb

PR #1660, opened 2016-11-20
[WIP][ZEPPELIN-1594] Personalized mode

jongyoul:
### What is this PR for?
Support personalized mode in multi user environments.

### What type of PR is it?
[Feature]

### Todos
* [x] - Add the method to send the result to specific user
* [ ] - Add the UI to toggle "personlized mode" in a note

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1594

1. Apply this patch
2. Enable shiro
3. note.config.personalizedMode = true in a note.json to be tested

### Questions:

This covers the first Todo only. @cloverhearts @1ambda Can you help me implement in a front-side?

### DIFF
--- a/zeppelin-web/src/app/jobmanager/jobmanager.controller.js
+++ b/zeppelin-web/src/app/jobmanager/jobmanager.controller.js
@@ -14,25 +14,93 @@
 'use strict';
 (function() {
 
-  angular.module('zeppelinWebApp').controller('JobmanagerCtrl', JobmanagerCtrl);
+  angular.module('zeppelinWebApp')
+    .controller('JobmanagerCtrl', JobmanagerCtrl);
 
-  JobmanagerCtrl.$inject = ['$scope', 'websocketMsgSrv', '$interval', 'ngToast'];
+  JobmanagerCtrl.$inject = ['$scope', 'websocketMsgSrv', '$interval', 'ngToast', '$q', '$timeout', 'jobManagerFilter'];
 
-  function JobmanagerCtrl($scope, websocketMsgSrv, $interval, ngToast) {
+  function JobmanagerCtrl($scope, websocketMsgSrv, $interval, ngToast, $q, $timeout, jobManagerFilter) {
     ngToast.dismiss();
-    $scope.filterValueToName = function(filterValue) {
-      var index = _.findIndex($scope.ACTIVE_INTERPRETERS, {value: filterValue});
+    var asyncNotebookJobFilter = function(jobInfomations, filterConfig) {
+      return $q(function(resolve, reject) {
+        $scope.JobInfomationsByFilter = $scope.jobTypeFilter(jobInfomations, filterConfig);
+        resolve($scope.JobInfomationsByFilter);
+      });
+    };
 
-      if ($scope.ACTIVE_INTERPRETERS[index].name !== undefined) {
-        return $scope.ACTIVE_INTERPRETERS[index].name;
+    $scope.doFiltering = function(jobInfomations, filterConfig) {
+      asyncNotebookJobFilter(jobInfomations, filterConfig).then(
+        function() {
+          // success
+          $scope.isLoadingFilter = false;
+        },
+        function() {
+          // failed
+        });
+    };
+
+    $scope.filterValueToName = function(filterValue, maxStringLength) {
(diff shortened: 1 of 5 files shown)


jongyoul:
@cloverhearts @1ambda Can you help me? :-)

cloverhearts:
@jongyoul 
Sure, What part are you helping me with?

1ambda:
I think we need to implement the second todo item.

cloverhearts:
@1ambda 
hmm.. But, this PR does not seem to have an implementation for the personalization toggle.
This PR seems to be implemented only in the context of the per user send when personalization mode.
hmm... I'm sorry.
(perhaps).. i did missing part?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -425,6 +425,20 @@ private void broadcast(String noteId, Message m) {
     }
   }
 
+  private void broadcastToUser(String noteId, String user, Message m) {
```
khalidhuseynov: i think there's already pretty similar `multicastToUser` function in [here](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java#L449) that does pretty much the same. so in longer run we may either combine them or keep the same naming with different parameters i think

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -154,6 +155,12 @@ public void setName(String name) {
     this.name = name;
   }
 
+
+  public boolean isPersonalizedMode() {
```
khalidhuseynov: so basically it means that `personalized mode` depends on the `config` field of the `note` itself, and the good part is that we have fine grained switch per note, although can't see where we'll be switching it. i guess it's left for frontend work

khalidhuseynov:
@jongyoul thanks for starting work on it! one more question. with current implementation in case of refreshing the page, user will get last run note right? so for example, if  `user2` runs note and `user1` refreshes, he'll still get results of `user1`.  any roadmap on that?

Leemoonsoo:
@jongyoul Could you describe more about the personalized mode?  I don't think current implementation is anywhere close to personalized mode discussed in https://issues.apache.org/jira/browse/ZEPPELIN-1236.

I think Personalized mode should be

* Keep boardcast every changes to every users who open the same note
* Zeppelin personalize parts of the Note, which is
    - paragraph.result (output)
    - paragraph.config.graph (graph selections)
    - paragraph.settings (dynamic form)

So, every changes, like paragraph addition, deletion, change the code, change width, 
everything is broadcasted, but only result / graph selection / dynamic form is personalized.

Current implementation will have following limitations
  - Once user refresh the page, user will see the other result.
  - Once two users changes note at the same time (i.e. update paragraph code, change paragraph order,  etc), the result is undefined.
  - Chart selection, dynamic form value is not personalized, which is major pain point that personalized mode try to resolve, i think.

What do you think?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
jongyoul: @khalidhuseynov I've removed that method. Thanks

jongyoul:
@Leemoonsoo As we discussed before, the first version of personalized mode is to avoid changing note simultaneously while two user see a same note. I think this version fits it. one user want to change a paragraph or run a paragraph, and he has a proper permission, he can change it but the others doesn't see his change until they refresh their browser. but they refresh it, the newest one is adopted. Isn't it the first version of personalized mode?

jongyoul:
@Leemoonsoo I've not tested the graph. If you mean a dynamic change of graph, I'll check it.

Leemoonsoo:
@jongyoul I think key requirement of personalized mode is two or more different user want to keep their own dynamic form input and the output of paragraph. How can this be done without let user run paragraph?

jongyoul:
@Leemoonsoo Yes, I haven't check the front-side. I'll fix it

PR #1637, opened 2016-11-15
[ZEPPELIN-1665] Z.run with external note executable and access resource for zeppelin in each interpreter

cloverhearts:
### What is this PR for?
Currently, the z.run command is restricted.
Only paragraphs in a single note can be executed.
I have modified this to allow you to freely execute paragraphs of other notes.
This PR provides the basis for the freeful use of Zeppelin's resources at each Interpreter implementation.

### What type of PR is it?
Improvement, Feature

### Todos
- [x] extends z.run
- [x] run all paragraph in external note
- [x] run paragraph for external note.
- [x] get resource for zeppelin in each interpreter.
- [x] improve test case.
- [x] how to use docuement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1665

### How should this be tested?
Currently under development.

run paragraph in same note
```
%spark
z.run("paragraphID")
```

run paragraph with external note
```
z.run("noteid", "paragraphid");
```

all note run
```
z.runNote("noteid");
```

### Screenshots (if appropriate)
- paragraph run
zrun

- noterun
runnote

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? yes

PR #1663, opened 2016-11-21
ZEPPELIN-1692: Ability to access Spark jobs UI from the paragraph

karuppayya:
### What is this PR for?
A paragraph execution may result in spark job(s).
Adding ability to access the spark job UI(corresponding to the job generated by the paragraph run), directly from the paragraph.

### What type of PR is it?
Improvement

### Todos
* [x]  Write tests

### What is the Jira issue?
ZEPPELIN-1692

Run paragraphs with spark code(scala, pyspark, sql, R).
The paragraph will display a button on the top right corner, which on click will open up the corresponding job UI

spark_jobs

### Questions:


cloverhearts:
awsome!

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -61,6 +61,7 @@
   Date dateUpdated;
   private Map<String, Object> config; // paragraph configs like isOpen, colWidth, etc
   public final GUI settings;          // form and parameter settings
+  private Map<String, Set<String>> runtimeInfos;
```
karup1990: I dont want to persist this info in json.
When i add the `transient` modifier, the serailization to JSON(for broadcasting to client browser) doesnot happen.
I  can add an Exclusion strategy for all the storage handlers. But wanted to check if there is an easier way to achieve this . @Leemoonsoo @cloverhearts

cloverhearts: Hello,
I think there is no problem using `transient`.
If you think you have a problem with using `transient`, can you comment on reason?

Actually, my english is not good.
I am sorry if I have been rude to you.
Thank you.

karup1990: @cloverhearts  Your English is not bad 👍 

If i add `transient` modifier, when `Gson.toJson(Paragraph)` is done to send message  about note/paragraph, the `runtimeInfos` wont be serialized and will not be received by browser(`runtimeInfos` wont get persisted in `note.json`), 

Whereas if i add the modifier(current implementation), it will get persisted in JSON. But we dont want it, since it a run time information, and we might not want to persist in the `note.json`

Is there a way where i can serailize this info(for sending to client browser) as well as not persist in json?
PS: `runtimeInfos` object has the paragraph url information.

karup1990:
Ready for review @Leemoonsoo

1ambda:
Let me review this PR and then give you some feedbacks. Thanks @karup1990

karup1990:
Thanks @1ambda for looking into this PR and for the suggestions.
For the 3rd point, 
I am not very sure of the use cases for hiding the output in a paragraph.
Do you see a case where the job link has to be hidden  when output is hidden?

1ambda:
@karup1990 sorry for misspelling It was **clear output**

spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
```diff
@@ -220,7 +222,8 @@ public static String showDF(SparkContext sc,
       Object df, int maxResult) {
     Object[] rows = null;
     Method take;
-    String jobGroup = "zeppelin-" + interpreterContext.getParagraphId();
+    String jobGroup = "zeppelin-" + interpreterContext.getNoteId() + "-"
+        + interpreterContext.getParagraphId();
```
zjffdu: I see several places of constructing jobGroupId from noteId and extract noteId from jobGroupId. It's better to put them together so that others can understand it easily and other places can reuse it.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventPoller.java
```diff
@@ -214,6 +222,12 @@ public void run() {
     }
   }
 
+  private String getInterpreterSettingId() {
+    String id = interpreterGroup.getId();
+    int indexOfColon = id.indexOf(":");
+    return id.substring(0, indexOfColon);
+  }
```
zjffdu: This assume the interpreterGroup id format, seems fragile.

zjffdu:
I left some comments inline and could you add test for this feature ?

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -26,6 +26,20 @@
   <span class="icon-control-play" style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"
         ng-click="runParagraph(getEditorValue())"
         ng-show="paragraph.status!='RUNNING' && paragraph.status!='PENDING' && paragraph.config.enabled"></span>
+  <span ng-show="paragraph.runtimeInfos.jobUrl.length == 1">
+    <a href="{{paragraph.runtimeInfos.jobUrl[0]}}" target="_blank"><span class="fa fa-tasks"></span> Spark job </a>
+  </span>
+  <span class="dropdown" ng-show="paragraph.runtimeInfos.jobUrl.length > 1">
+    <span class="fa fa-tasks" style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"
```
AhyoungRyu: I think `style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"` doesn't need in here :)

AhyoungRyu: I think `tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"` doesn't need in here :)

AhyoungRyu:
Hi @karup1990, really nice and useful feature indeed.
While quickly looking through this, I have some minor feedbacks about UI :)

1. `Spark Job` and `Spark jobs` have different font like below.
screen shot 2016-11-24 at 3 07 53 pm
screen shot 2016-11-24 at 3 08 04 pm

2. And I think these block that you added in this PR should be below [this line](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html#L31).
```
  <span ng-show="paragraph.runtimeInfos.jobUrl.length == 1">
    <a href="{{paragraph.runtimeInfos.jobUrl[0]}}" target="_blank"><span class="fa fa-tasks"></span> Spark Job </a>
  </span>
  <span class="dropdown" ng-show="paragraph.runtimeInfos.jobUrl.length > 1">
  <span class="fa fa-tasks" style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"
        data-toggle="dropdown"
        type="button">  Spark Jobs
  </span>
    <ul class="dropdown-menu" role="menu" style="width:200px;z-index:1002">
       <li ng-class="{'option-disabled': !noteOperationsAllowed()}" ng-repeat="url in paragraph.runtimeInfos.jobUrl">
         <a href="{{url}}" target="_blank"><span class="fa fa-external-link-square"></span> Jobs #{{$index}}</a>
       </li>
    </ul>
  </span>
```
   - current
before

   - after
after

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -26,6 +26,20 @@
   <span class="icon-control-play" style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="Run this paragraph (Shift+Enter)"
         ng-click="runParagraph(getEditorValue())"
         ng-show="paragraph.status!='RUNNING' && paragraph.status!='PENDING' && paragraph.config.enabled"></span>
+  <span ng-show="paragraph.runtimeInfos.jobUrl.length == 1">
+    <a href="{{paragraph.runtimeInfos.jobUrl[0]}}" target="_blank"><span class="fa fa-tasks"></span> Spark job </a>
```
Leemoonsoo: How about display runtime info button before paragraph status and run button?

image

image

So all other other control buttons are stay in the same position whether runtimeInfo button is displayed or not.

Leemoonsoo: And how about we generalize this feature little bit more, such as

```
runtimeInfos = {
   label: "Spark Job"
   jobUrl : [],
}
```

So other interpreters can also leverage this feature.

karup1990:
Thanks for the feedback.  i will have them handled  some time over the weekend.

spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
karup1990: I have added `Utils. buildJobGroupId ` and using it across.

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
karup1990: @1ambda Clearing the links when we clear the output 👍 
@AhyoungRyu Fixed font issue
@Leemoonsoo Added `ParagraphRuntimeInfos` class(https://github.com/apache/zeppelin/pull/1663/files#diff-a2ac97dfb3f07b2a3fd1533af701295e) to have a generic support to add more properties. Let me know how it looks.
@AhyoungRyu @Leemoonsoo @1ambda Fixed the alignment and capitalized the text(to access the jobs) 
spark

karup1990:
Failure seems not related
`04:09:12,112 ERROR org.apache.zeppelin.AbstractZeppelinIT:136 - Exception in ZeppelinIT while testAngularDisplay 
org.openqa.selenium.TimeoutException: Timed out after 60 seconds waiting for org.apache.zeppelin.AbstractZeppelinIT$1@4c7305e4
Build info: version: '2.48.2', revision: '41bccdd10cf2c0560f637404c2d96164b67d9d67', time: '2015-10-09 13:08:06'
System info: host: 'testing-docker-bcffa7c6-8593-4893-a72f-07a719f105fa', ip: '172.17.0.6', os.name: 'Linux', os.arch: 'amd64', os.version: '4.8.7-040807-generic', java.version: '1.7.0_76'
Driver info: driver.version: unknown
 `

karuppayya:
Updated the PR with test @zjffdu

karuppayya:
Ready for review

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -156,7 +159,46 @@ public boolean isSparkContextInitialized() {
...
+        int secondIndex = jobgroupId.indexOf("-", indexOf + 1);
+        return jobgroupId.substring(indexOf + 1, secondIndex);
+      }
+
+      private String getParagraphId(String jobgroupId) {
+        int indexOf = jobgroupId.indexOf("-");
+        int secondIndex = jobgroupId.indexOf("-", indexOf + 1);
+        return jobgroupId.substring(secondIndex + 1, jobgroupId.length());
+      }
+    };
```
zjffdu: Would it be better to put `getNoteId` and `getParagraphId` to Utils ? So that the constructing and parsing logic are in the same class.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventPoller.java
zjffdu: I think it is better to put `getInterpreterSettingId` into `Utils` as well because it is about how we parse interpreterGroup

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfos.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public ParagraphRuntimeInfos(String propertyName, String label, String group) {
+    this.propertyName = propertyName;
+    this.label = label;
+    this.group = group;
+  }
+
+  public void addValue(String value) {
+    if (values == null) {
+      values = new ArrayList<>();
```
zjffdu: Should we create values in the constructor ? Otherwise at least this method is not thread-safe (we might create 2 values in 2 threads)

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -61,6 +61,7 @@
   Date dateUpdated;
   private Map<String, Object> config; // paragraph configs like isOpen, colWidth, etc
   public final GUI settings;          // form and parameter settings
+  private Map<String, ParagraphRuntimeInfos> runtimeInfos;
```
zjffdu: Should we rename `ParagraphRuntimeInfos` to `ParagraphRuntimeInfo` ? Because it is a little confusing to see `runtimeInfos` also plural

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -413,6 +413,7 @@ private Note loadNoteFromRepo(String id, AuthenticationInfo subject) {
       if (p.getDateFinished() != null && lastUpdatedDate.before(p.getDateFinished())) {
         lastUpdatedDate = p.getDateFinished();
       }
+      p.clearRuntimeInfo();
```
zjffdu: Why we need to clear RuntimeInfo when we load note if the runtimeInfo is not stored ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfos.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * 
+ *
+ */
+public class ParagraphRuntimeInfos {
+
+  String propertyName;
+  String label;
+  String group;
```
zjffdu: I think here the `label` and `group` are duplicated. I don't see you use label in frontend.  Another approach I think of is that we always use interpreter group name as the prefix of `propertyName` as convention, (e.g. `spark.jobUrl`) @Leemoonsoo

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -970,6 +1015,7 @@ public void populateSparkWebUrl(InterpreterContext ctx) {
         infos.put("url", sparkUrl);
         logger.info("Sending metainfos to Zeppelin server: {}", infos.toString());
```
zjffdu: Should we put the logging in the if block ? And what about else block ? Will that happen ?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1927,4 +1927,46 @@ private void broadcastToWatchers(String noteId, String subject, Message message)
...
+        metaInfos.remove("paraId");
+        String label = metaInfos.get("label");
+        metaInfos.remove("label");
+        paragraph.updateRuntimeInfos(label, metaInfos, setting.getGroup());
+        broadcast(note.getId(), new Message(OP.PARAGRAPH).put("paragraph", paragraph));
+      }
+    }
+  }
+
+  public void clearParagraphRuntimeInfo(InterpreterSetting setting) {
```
zjffdu: I am afraid this method may some issues because the binding between `InterpreterSetting` and `Note/Paragraph` may change. e.g. At the beginning I use spark in p1, but later I use flink in p1, but the method will clear all the runtime info of p1 when I restart spark interpreter.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfos.java
karuppayya: The label is used here 
https://github.com/apache/zeppelin/blob/9612dd08b12b1b76f70d9baab1064d0207f1fd46/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html#L19
I had added group , so that two interpreters having same property name can be rendered in a different manner in UI.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
karuppayya: done

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
karuppayya: I will nedd to remove this line.
The  runtime infos is persisted currently. But I dont want to persist it. 
If i add transient modifier to `runtimeInfos`, when Gson.toJson(Paragraph) is done to send message about note/paragraph, the runtimeInfos wont be serialized and will not be received by browser(runtimeInfos wont get persisted in note.json),

Whereas if i add the modifier(current implementation), it will get persisted in JSON. But we dont want it, since it a run time information, and we might not want to persist in the note.json

Is there a way where i can serialize this info(for sending to client browser) as well as not persist in json?
PS: runtimeInfos object has the paragraph url information.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
karuppayya: done

karuppayya:
@zjffdu Addressed the feedbacks

Leemoonsoo:
Tried but no button is displayed. Maybe i'm missing something?

karuppayya:
@Leemoonsoo The buutton gets displayed if there is a spark job corresponding to the paragraph run.
Can you try this `sc.parallelize(1 to 10).map(x => x ).collect`

Leemoonsoo:
Tried but still no luck ..

karuppayya:
@Leemoonsoo Missed pushing a change to the remote, which was causing the issue.
I have fixed it and tested it. Can you please take  look.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -2021,4 +2021,46 @@ private void broadcastToWatchers(String noteId, String subject, Message message)
...
+      if (paragraph != null) {
+        InterpreterSetting setting = notebook().getInterpreterFactory()
+                                               .get(interpreterSettingId);
+        setting.addNoteToPara(noteId, paragraphId);
+        metaInfos.remove("noteId");
+        metaInfos.remove("paraId");
+        String label = metaInfos.get("label");
+        metaInfos.remove("label");
+        paragraph.updateRuntimeInfos(label, metaInfos, setting.getGroup(), setting.getId());
+        broadcast(note.getId(), new Message(OP.PARAGRAPH).put("paragraph", paragraph));
```
karuppayya: I think the paragraph need not be completely broadcasted here. A new message can be created to updated only the runtime infos. Any thoughts on this? @Leemoonsoo @zjffdu

zjffdu:
Personally, I think this feature is useful but I just feel that with adding more new features to zeppelin(especially related with interpreter). We couple many components closely much more than before. (e.g. We usually we add parameter to Constructor of class A to allow it access class B). This would add more complexities to the code base which would make it difficult to add new features and easy to break some function in future. I think maybe we should spend some time to do refactoring to make the component boundary more clean. It would not only make adding new feature easier but also make zeppelin robust and less error-prone.  This comment is not related with PR but just share my thoughts of development of zeppelin. We could merge it first and do the refactoring later. \cc @Leemoonsoo

zjffdu:
BTW @karuppayya Do you have andy integration test for it ?

karuppayya:
@zjffdu +1 on the refactor part.
I have written a [test](https://github.com/apache/zeppelin/pull/1663/files#diff-7d13c722724f9c346f38303f4b435548R326). Would this suffice?
I can add more tests if required.

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -13,7 +13,24 @@
 -->
 
 <div id="{{paragraph.id}}_control" class="control" ng-show="!asIframe">
-
+  <span ng-show="paragraph.runtimeInfos.jobUrl.values.length == 1 && paragraph.runtimeInfos.jobUrl.group == 'spark'">
```
Leemoonsoo: Shell we remove condition `paragraph.runtimeInfos.jobUrl.group == 'spark'` ? so other interpreter can use this feature, too?

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -13,7 +13,24 @@
...
+    <a href="{{paragraph.runtimeInfos.jobUrl.values[0]}}" target="_blank" style="text-decoration: none;">
+      <span class="fa fa-tasks"></span>
+      {{paragraph.runtimeInfos.jobUrl.label}}
+    </a>
+  </span>
+  <span class="dropdown" ng-show="paragraph.runtimeInfos.jobUrl.values.length > 1 && paragraph.runtimeInfos.jobUrl.group == 'spark'">
+    <span style="cursor:pointer;color:#3071A9" tooltip-placement="top" tooltip="View in Spark web UI"
```
Leemoonsoo: Shell we remove condition `paragraph.runtimeInfos.jobUrl.group == 'spark'` here, too?
Also how about display tooltip from property of `runtimeInfos` instead of hardcoded "View in Spark web UI" ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -61,6 +61,7 @@
   Date dateUpdated;
   private Map<String, Object> config; // paragraph configs like isOpen, colWidth, etc
   public final GUI settings;          // form and parameter settings
+  private Map<String, ParagraphRuntimeInfo> runtimeInfos;
```
Leemoonsoo: `runtimeInfos` doesn't need to be persisted. Can we add `transient`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -45,6 +46,7 @@
   // always be null in case of InterpreterSettingRef
   private String group;
   private transient Map<String, String> infos;
+  private transient Map<String, Set<String>> noteIdToParaIdsetMap;
```
Leemoonsoo: Because of every `Note` already has reference to its `Paragraphs`, i don't think it make sense maintaining another noteId - paragraphId map here, according to this variable name.

How about change this variable name to `runtimeInfosToCleared` or something similar to describe it's purpose better? That may help understanding the code better, imo.

spark/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java
```diff
@@ -92,14 +96,40 @@ public void setUp() throws Exception {
...
         new HashMap<String, Object>(),
         new GUI(),
         new AngularObjectRegistry(intpGroup.getId(), null),
         new LocalResourcePool("id"),
         new LinkedList<InterpreterContextRunner>(),
-        new InterpreterOutput(null));
+        new InterpreterOutput(null)) {
+        
+        @Override
+        public RemoteEventClientWrapper getClient() {
```
Leemoonsoo: #1637 also exposed one of method of RemoteEventClient to interpreter. But little different way.
#1637 created new class `RemoteWorksController` to wrap RemoveEventClient.
https://github.com/apache/zeppelin/pull/1637/files#diff-5f11952b2fbf4e41e273df20c409ec4b

This approach and #1637 's approach is basically the same, but in the future refactoring, i think we need to decide either a) create new class for every feature exported from RemoveEventClient (like RemoteWorksController), or b) create a common wrapper class RemoteEventClientWrapper and export all necessary methods from it.

zjffdu: @Leemoonsoo Regarding the refactoring, I not only mean `RemoveEventClient`, but also other interpreter related components. Like `InterpreterFactory`, `InterpreterSetting`, `InterpreterGroup`, `RemoteInterpreter` and etc. I just feel the boundary and responsibility of these classes/components is not clear for now. `InterpreterFactory` is too complicated and should delegate many work to `InterpreterSetting`. And currently there's no single component to manage the lifecycle of interpreters. I know refactoring is a little risky and may affect functions. So I think maybe we put this as part of 0.8.  If you think this is feasible and necessary. I can write a more detailed document and gather more feedback from communities.  What do you think ?

Leemoonsoo: The reason i mentioned especially `RemoveEventClient` is because its part of Interpreter development API and we want to minimize API change on every release. (doesn't mean this PR has to take care)

Agree, we can reactor overall interpreter layer once we freeze 0.7.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
karuppayya: Yes we dont need to persist `runtimeInfos`.
If we add `transient` modifier, the `runtimeInfos` will not be serialized and sent to client browser.
Any suggestions to solve this? @Leemoonsoo

karuppayya:
Thanks @Leemoonsoo @zjffdu for the comments .
 I have addressed the feedbacks. Can you please take a look.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
Leemoonsoo: I see what you mean.

Because of Note object is passed to underlying NotebookRepo layer, and each NotebookRepo layer take care of serialization, we have no control of serializer in the each NotebookRepo implementation. So i think it's better have @transient for them.

What we can change is serializer on both REST and websocket end point.

websocket
https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java#L107

rest
https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/server/JsonResponse.java#L103

Will it be possible to have exclusion strategy on websocket and rest endpoint which ignore @transient for a field "runtimeInfos" ?

karuppayya: Looks like `excludeFields` callback  of `ExcluionStrategy` is not triggered for field with `transient` modifiers.
http://grepcode.com/file/repo1.maven.org/maven2/com.google.code.gson/gson/2.3/com/google/gson/internal/Excluder.java#150

Should i remove `transient` modifier , and add the excluson strategy for all `NotebookRepo` ?? @Leemoonsoo

Leemoonsoo: I see, thanks for explain. Do you see any other possible approaches? 
Because NotebookRepo designed to allow 3rd party implementation, and i think it's not the best making every possible NotebookRepo implementation handles the `runtimeInfos` with exclusionStrategy.

karuppayya: @Leemoonsoo 
Based on [this](http://stackoverflow.com/questions/11271375/gson-custom-seralizer-for-one-variable-of-many-in-an-object-using-typeadapter) thread, I have introduced `NotebookTypeAdapterFactory.java`.
Now the notebook is not persisted, and  the information is sent over web socket connection.
Let me know how it looks

Leemoonsoo:
Sorry for coming back late.
Tested the last commit and looks good to me.
@karuppayya could you merge master and resolve the conflict and see if CI becomes green? Then i'd like to merge it to master.

I've found one behavior from last commit which can be addressed in either this PR or following PR
sparkui

karuppayya:
Thanks @Leemoonsoo for reviewing. 
I have rebased my the PR and resolved conflicts.
Also didnt get the behaviour from the GIF, can u please explain.
If its a simple change , I can fix it in this PR.

karuppayya:
Test failures seem unrelated
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.17:test (default-test) on project zeppelin-python: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.17:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd /home/travis/build/apache/zeppelin/python && /usr/lib/jvm/java-7-oracle/jre/bin/java -Xmx2g -Xms1g -Dfile.encoding=UTF-8 -jar `

Leemoonsoo:
The behavior from GIF is, 

When spark job is run in `%spark` paragraph, runtime info is displayed correctly.
After that, if the same paragraph is reused for other interpreter (e.g. %sh), i think runtime info should be reset. but currently it's still showing previous runtime info.

karuppayya:
Than @Leemoonsoo for the explanation

I am not able to repro it.

The following two lines takes care of it
1. [Note.java#567](https://github.com/apache/zeppelin/pull/1663/files#diff-0f40daff726ba29c9a004b342d422dfeR567) -  Clears the runtime info of paragraph , when we `run` the paragraph
2. [paragraph.controller.js#1054](https://github.com/apache/zeppelin/pull/1663/files#diff-52fa6628f0b6793d22ffe1f3e4763635R1054) - When the status of paragraph gets updated when we run a paragraph(`FINISHED`, `ABORTED` etc -> `PENDING`), paragraph update message is sent to UI(which has `runtimeInfos` nullified , from #1), which in turn hides the link.

If you have any specific steps, can you please let me know.

Leemoonsoo:
@karuppayya tried again with clean build and i don't see problem now. Maybe my test branch has been messed up. 

Now, two CI profiles are failing
https://travis-ci.org/apache/zeppelin/jobs/195910315 (SELENIUM)
https://travis-ci.org/apache/zeppelin/jobs/195910316 (python/pyspark)

Later one doesn't look related but i think it worth to make sure first one(SELENIUM) passes the test.
Could you trigger CI again and see if test goes green?

karuppayya:
I see the following errors in the SELENIUM test

`org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"//i[@tooltip='WebSocket Connected']"}`
`org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"(//div[@ng-controller=\"ParagraphCtrl\"])`

Reopening the PR to see it is intermittent

felixcheung:
I think it is related, right?
```
Caused by: org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"(//div[@ng-controller=\"ParagraphCtrl\"])[1]//div[contains(@class, 'control')]//span[1][contains(.,'READY')]"}
```

karuppayya:
Thanks @felixcheung . Looks like i read only part of the exception line and thought it was not due to this PR. I have fixed it now.
The exception related to `websocket connected` tooltip still remains
`
org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"//i[@tooltip='WebSocket Connected']"}
` @Leemoonsoo

Leemoonsoo:
last CI run fails with 
```
Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 228.978 sec <<< FAILURE! - in org.apache.zeppelin.integration.ParagraphActionsIT
testDisableParagraphRunButton(org.apache.zeppelin.integration.ParagraphActionsIT)  Time elapsed: 10.301 sec  <<< FAILURE!
java.lang.AssertionError: Paragraph status is 
Expected: "READY"
     but: was ""
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.apache.zeppelin.integration.ParagraphActionsIT.testDisableParagraphRunButton(ParagraphActionsIT.java:247)
```
Could you trigger CI once more?

karuppayya:
Made [this](https://github.com/apache/zeppelin/pull/1663/commits/aef095a0c0f82a13d0482584c57d85881a91dc53) fix for the test failure. But the tests are still failing. Not familiar with the selenium constructs as well.
I am not able to run the tests locally on my machine.
@Leemoonsoo can u please help me here?

Leemoonsoo:
@karuppayya sure, let me take a look selenium test in this branch.

Leemoonsoo:
@karuppayya could you try following change?

```
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
@@ -71,7 +71,7 @@ abstract public class AbstractZeppelinIT {
 
   protected String getParagraphStatus(final int paragraphNo) {
     By locator = By.xpath(getParagraphXPath(paragraphNo)
-        + "//div[contains(@class, 'control')]//span[2]");
+        + "//div[contains(@class, 'control')]/span[2]");
 
     return driver.findElement(locator).getText();
   }
```

In case of you'd like to run selenium test, please check https://github.com/apache/zeppelin/pull/1967

karuppayya:
Thanks @Leemoonsoo  for the suggestion and the pointer.
Looks like all tested have passed. Ready to merge.

Leemoonsoo:
@karuppayya Great work!

Tagar:
Just upgraded to yesterday's master snapshot.

When I click on any of these links, link is leading to 

http://hostname.domain.com:8088/proxy/application_1488384993892_0001/jobs/job/

and that error shows **HTTP ERROR 400**

Page reads 

> Problem accessing /jobs/job/. 
> 
> Reason:    requirement failed: **Missing id parameter**
> Powered by Jetty://
> 

Thanks.

karuppayya:
@Tagar to access a specific job, the id has to be part of the url .
Something like http://hostname.domain.com:8088/proxy/application_1488384993892_0001/jobs/job?id=<id of the job>
I am not sure why the id is not part of the URL. Did you find anything suspicious in logs? Also were you able to access the jobs from Spark Web UI?
Is it consistent?

mindcrusher11:
not sure but

http://sparkkafka:4041/streaming/batch/?id=1488549398000

from history server job ids can be accessed.

Tagar:
@karuppayya thank you for the follow up.

A little bit more information - the link on paragprah leads to 

> http://10.20.32.57:28009/jobs/job?id=123 

A Spark Driver UI. 
Spark Driver web server then redirects to a link like I sent in my previous post,
and it misses job id in the redirected URL, i.e.

> http://host.domain.com:8088/proxy/application_1488384993892_0044/jobs/job/ 

To your question on consistency - yes it happens in 100% cases, I never saw this new feature works for us. 
Is this Spark version dependent or something? We're running CDH Spark 2.

Tagar:
According to https://github.com/apache/spark/pull/5947 URL format is different in YARN and non-YARN modes? Was PR-1663 for ZEPPELIN-1692 tested on both of these modes? Not sure what else might break those links. We do use Spark on YARN. Created https://issues.apache.org/jira/browse/ZEPPELIN-2221 to move discussion there. Thank you.

Tagar:
Does this feature work for anyone who is using Spark on YARN?

It seems to be broken by https://issues.apache.org/jira/browse/SPARK-20772

I've updated https://issues.apache.org/jira/browse/ZEPPELIN-2221

PR #1666, opened 2016-11-21
[ZEPPELIN-1689] enable Power architecture

asanjar:
### What is this PR for?

this patch is platform independent and should not effect Intel architecture. We are simply upgrading the version of fronend-maven-plugin, nodejs, phantomjs-launcher to versions that are supported by Power architecture as well.

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1689

should test on x86 for any regression, highly unlikely.
OpenPOWER foundation team has already verified the patch on Power as part of Apache Bigtop project.

### Questions:

### DIFF
--- a/zeppelin-web/package.json
+++ b/zeppelin-web/package.json
@@ -36,7 +36,7 @@
     "karma": "~0.12.23",
     "karma-coverage": "^0.5.1",
     "karma-jasmine": "~0.1.5",
-    "karma-phantomjs-launcher": "~0.1.4",
+    "karma-phantomjs-launcher": "~1.0.0",
     "load-grunt-tasks": "^0.4.0",
     "time-grunt": "^0.3.1"
   },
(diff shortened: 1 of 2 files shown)


1ambda:
Hi, @asanjar 

Actually, we are talking about upgrading node, npm and karma related versions in https://github.com/apache/zeppelin/pull/1659
Could you check it?

bzz:
Indeed, if you could check #1659 - AFAIK it does all the same, only the node version is different there.

How about we merge this guy as it does not bring major changes and then rebase #1659 on it and see if we want to keep Node 6x there?

1ambda:
I think we can merge this first and then rebase #1659 easily :) 

This PR looks good to me and is what we want :)

bzz:
👍  Looks good to me, merging if there is no further discussion

asanjar:
+1

PR #1670, opened 2016-11-23
[ZEPPELIN-1700] ConnectException between ZeppelinServer and RemoteInterpreterProcess.

astroshim:
### What is this PR for?
The default value of interpreter connection timeout is 30s but sometimes 30 seconds may be insufficient to connect to RemoteInterpreterProcess.
I think to increase the timeout value doesn't effect of performance so it's okay to give large value.
and in the later, It's better give users to be able to edit Zeppelin Configurations on the web.

### What type of PR is it?
Bug Fix | Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1700

- before
image

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -1101,7 +1101,7 @@ public class InterpreterFactory implements InterpreterGroupFactory {
 
   private Interpreter connectToRemoteRepl(String noteId, String className, String host, int port,
       Properties property, String userName, Boolean isUserImpersonate) {
-    int connectTimeout = conf.getInt(ConfVars.ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT);
+    int connectTimeout = conf.getInt(ConfVars.ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT) * 10;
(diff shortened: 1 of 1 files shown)


Leemoonsoo:

Could you edit default value from [here](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L557) instead of multiplying timeout value when reading it?

Although it's not configurable through the web, `ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT` is  already configurable through conf/zeppelin-env.sh or conf/zeppelin-site.xml. 
So some unusual case that requires very long process startup time will still able to configure the timeout.

So i would suggest some reasonable time for the default timeout. 5min (30s*10) is too large IMO. Maybe a minute?

astroshim:
@Leemoonsoo Oh sorry, I didn't know that the value is configurable.
If modification is possible, It's better to close this PR because 30s is okay in normal case.
Thank you!

jpgbus:
@Leemoonsoo Hats off to you and the community for Zeppelin - it seems to be a masterpiece in the making. I'd like to update the document for this ConfVars.

https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L576
`ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT("zeppelin.interpreter.connect.timeout", 30000),`

I submitted [PR](https://github.com/apache/zeppelin/pull/1993)

PR #1678, opened 2016-11-24
[ZEPPELIN-1706] Make spark Ui accesible in various interpreter modes

karup1990:
### What is this PR for?
Spark UI should be accessible in various other modes(user, note) and honour the properties of the corresponding mode.
Example: In per user mode isolated mode, each logged in user should be taken to his/her respective spark UI
In per note isolated mode, each notebook will have its own application and the user should be taken the respective spark application UI
All the interpreter modes are handled with this PR

### What type of PR is it?
Bug fix

### What is the Jira issue?
ZEPPELIN-1706

Set various interpreter modes for the interpreter setting and test if the spark Application UI is accessible appropriately

zep-1702

The screen capture shows per note isolated mode. But the PR handles all interpreter modes
### Questions:


karuppayya:
@Leemoonsoo  @zjffdu Ready for review

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
```diff
@@ -243,17 +244,26 @@ public Response addRepository(String message) {
   /**
    * get the metainfo property value
    */
-  @GET
+  @POST
```
zjffdu: Why this is changed to POST ?

karuppayya: I wanted to send more amount of parameter data , hence started using `POST`
https://github.com/apache/zeppelin/pull/1678/files/6ec58e8ebf2493e5f417d5ee3c352ab3c4c793bd#diff-79927cd87b606e2efc3267cf436aa704R643
Also read this in one of the posts.
```
an important consideration when using GET for AJAX requests is that some browsers - IE in particular - will cache the results of a GET request
```
Not very sure if its applicable.
One another reason is to keep the url cleaner :)

zjffdu: POST is for request that would make state changes on server side, GET is just for read-only operation.  IIUC this api didn't change anything in server side, then we should use GET.

karuppayya:
@zjffdu I have addressed the feedbacks .

karuppayya:
@Leemoonsoo @zjffdu can u please take a look.and let me know how it looks

zeppelin-web/pom.xml
```diff
@@ -108,8 +108,8 @@
               <goal>install-node-and-npm</goal>
             </goals>
             <configuration>
-              <nodeVersion>${node.version}</nodeVersion>
-              <npmVersion>${npm.version}</npmVersion>
+              <nodeVersion>v6.9.1</nodeVersion>
```
RPCMoritz: this change makes no sense in the scope of this PR

zjffdu:
ping @karuppayya

PR #1679, opened 2016-11-25
ZEPPELIN-1707. Pass userName when creating interpreter through thrift

zjffdu:
### What is this PR for?
In ZEPPELIN-1607, I'd like refactor livy interpreter to scoped mode by default, this require username when open this interpreter. So I propose to pass username when creating interpreter through thrift.
What I did in this PR.
* update `RemoteInterpreterService.thrift` and regenerate the java thrift code.
* update `genthrift.sh`, otherwise hashCode method won't be generated correctly.
* This is one compilation issue (`PythonDockerInterpreterTest.java`) in the existing master branch, I also fix it here. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1707

### Questions:


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/thrift/RemoteInterpreterEvent.java
```diff
@@ -29,12 +47,9 @@
 import java.util.BitSet;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
-import javax.annotation.Generated;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@SuppressWarnings({"cast", "rawtypes", "serial", "unchecked"})
-@Generated(value = "Autogenerated by Thrift Compiler (0.9.2)", date = "2016-11-18")
```
prabhjyotsingh: Probably you should use the script mentioned here https://github.com/apache/zeppelin/blob/master/docs/development/howtocontribute.md#generating-thrift-code for generating thrift file, which will retain this information.

zjffdu:
@Leemoonsoo @prabhjyotsingh Please help review.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/dev/ZeppelinDevServer.java
```diff
@@ -52,7 +52,8 @@ protected Interpreter getInterpreter(String noteId, String className) throws TEx
             "dev",
             noteId,
             DevInterpreter.class.getName(),
-            new HashMap<String, String>());
+            new HashMap<String, String>(),
+            "anonymous");
```
felixcheung: maybe we shouldn't hardcode UI string in the server code?

bzz:
Looks good to me.

CI is green now, so will merge it to master if there is no further discussion.

astroshim:
Sorry for late review. 

User name is already passed to interpreter.
We can get the user id on interpreterContext.getAuthenticationInfo().getUser()

I think passing password is valuable in this PR because we can use the login account information(username + password) on interpreter like JDBC.

What do you think?

zjffdu:
@astroshim `InterpreterContext` is only available when interpreting paragraph (after Interpreter is opened).  What I need here is to get username when opening interpreter.

astroshim:
@zjffdu It's my misunderstanding. LGTM!

Leemoonsoo:
LGTM

PR #1695, opened 2016-11-28
Zeppelin 1283

PetroKostyuk:
### What is this PR for?
It's adding option to specify column in scatterPlot, that will be displayed in label

### What type of PR is it?
Feature

### Todos
I had trouble to get label displayed for most of points. Label was triggered by cursor on different position than point was. See picture, my cursor was on position of red X. But this problem was there even before my edits and I think it should be handeled as separate issue. Mostly I was able to trigger labels when I selected only Y axis and hovered on points on bottom of chart.
screenshot from 2016-11-26 13-51-10

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1283

reate scatterPlot and put some column in new 'label' field. Values from that column should appear in label corresponding to selected point.

screenshot from 2016-11-28 19-51-52
screenshot from 2016-11-28 19-51-18

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-pivot.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-pivot.html
@@ -157,7 +157,7 @@ limitations under the License.
             class="list-unstyled"
             style="height:36px">
           <li ng-if="paragraph.config.graph.scatter.size">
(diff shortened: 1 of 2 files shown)


bzz:
@PetroKostyuk thank you for improvement! 

It looks like your work need to be rebased on the latest master in order to resolve conflicts, before merging.

PR #1697, opened 2016-11-29
[ZEPPELIN-1352] Refactor Notebook repo versioning to return Revision Empty instead of NULL

anthonycorbacho:
### What is this PR for?
In Zeppelin notebookRepo, versioning (checkpoint, get, revisionHistory) use to return NULL.
It will be better to return Empty (Revision or list).

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-1352](https://issues.apache.org/jira/browse/ZEPPELIN-1352)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/AzureNotebookRepo.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/AzureNotebookRepo.java
@@ -224,20 +224,20 @@ public class AzureNotebookRepo implements NotebookRepo {
   public Revision checkpoint(String noteId, String checkpointMsg, AuthenticationInfo subject)
       throws IOException {
     // no-op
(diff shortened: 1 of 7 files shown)


anthonycorbacho:
Please review otherwise i am merging it :)

khalidhuseynov:
normally when doing this kind of changes, need to:
 1.  either override `equals` method or make explicitly static `isEmpty` method that would handle comparison cases. (e.g. `revision1.equals(Revision.EMPTY)` or `Revision.isEmpty(revision1)`)
 2. substitute all instances of that class that previously have been compared with `null`. for example, [here](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java#L1400) an instance of it was compared to `null` to see if checkpoint was successful, which wouldn't work now and should be changed.
 
Generally, this kind of changes would be nice to be reviewed, unless that's hotfix.

PR #1701, opened 2016-11-29
remove usage of non-standard string method

felizbear:
### What is this PR for?
This PR removes a non-standard string prototype method `contains` that can cause potential bugs in the future maintenance.

### What type of PR is it?
Bug Fix 

### Todos
* [x] - remove usage of non-standard string method `contains` in favor of standard `indexOf`

Download as `csv / tsv` (graph view) should work as expected


bzz:
Looks good to me.

Usually, it's better to have JIRA issue to keep author\attribution information for the release.

Merging to master if there is no further discussion

PR #1702, opened 2016-11-29
[ZEPPELIN-1727] Shift + Enter runs two paras in particular scenario

karuppayya:
### What is this PR for?
1. click *insert new* in p1(say code in p1 is %sh date -u )
2. New para p2 is created and cursor is on the p2 editor.
3. Type code say %sh date -u and press shift + enter.(without clicking on  p2's editor)
The output of p1 and p2 will be same, p2 run triggered p1's run as well.
This PR fixes the above scenario

### What type of PR is it?
Bug fix

### Todos
NA

The steps mentioned in description of PR need s to be tested

date

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -365,9 +365,9 @@
       _.each($scope.note.paragraphs, function(para) {
         if (para.id === paragraph.id) {
           para.focus = true;
+          $scope.$broadcast('focusParagraph', para.id, 0, false);
         }
       });
-      $scope.$broadcast('focusParagraph', paragraph.id, 0, false);
     };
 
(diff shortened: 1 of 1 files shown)


cloverhearts:
That's a good fix!
I have a personal oppinion.
How about moving the `focusParagraph` event` broadcast` to the link below?
https://github.com/karuppayya/incubator-zeppelin/blob/3dd7a5dac9ae47c7cdaac105b0c766b5649e844f/zeppelin-web/src/app/notebook/notebook.controller.js#L367
In my opinion, there should be a `focusParagraph` event in the matching part that checks the id of the paragraph.
It seems to be safer.
What do you think?

Leemoonsoo:
Tested and it works well.

Leemoonsoo:

PR #1705, opened 2016-11-29
[ZEPPELIN-1729] With CRLF line endings, Livy gets HTTP 500 errors

rhymoid:
### What is this PR for?
When you paste text on Windows, you may inadvertedly add some Carriage Return control characters along with it. When this happens in a Livy paragraph in Zeppelin 0.6.0, execution of that paragraph will fail with a "500 Internal Server Error" message. This is because `LivyHelper` does not properly sanitise its input when generating a JSON payload for the Livy server. This PR takes the subset for the resolution of [ZEPPELIN-1430](https://issues.apache.org/jira/browse/ZEPPELIN-1430/) that should resolve this issue.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
<https://issues.apache.org/jira/browse/ZEPPELIN-1729>

Paste the following code in a new paragraph, ensuring that lines are ended by CRLF rather than LF:

    %livy.pyspark
    import datetime
    import json

This can be easily realised on Windows, by copy-pasting this code from Notepad. Then, try to execute the paragraph. Without the fix, a "500 Internal Server Error" message should appear. With the fix, the paragraph will successfully execute (and do nothing interesting).

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivyHelper.java
@@ -20,6 +20,7 @@ package org.apache.zeppelin.livy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import org.apache.zeppelin.interpreter.InterpreterContext;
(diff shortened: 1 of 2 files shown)


rhymoid:
Looks like the builds for branch-0.6 are failing because some Spark binaries were removed from all mirrors. I think this was fixed in master with #1599 + #1709, but not for branch-0.6.

rhymoid:
I'm closing this PR, as I now understand that 0.6.3 will not happen. The only thing this PR contributes is some tests for `LivyHelper`, but that class needs a complete rewrite anyway.

zjffdu:
Thanks @rhymoid, I am refactoring LivyInterpreter now in #1612 , refactoring LivyHelp is my next step.

PR #1712, opened 2016-12-01
[ZEPPELIN-1742] send error info to popup when commit fails

khalidhuseynov:
### What is this PR for?
To notify user when commit of note revision have failed.

### What type of PR is it?
Bug Fix | Improvement

### Todos
* [x] - send error

### What is the Jira issue?
[ZEPPELIN-1742](https://issues.apache.org/jira/browse/ZEPPELIN-1742)

try to commit note with default storage

failure case before:
commit error before

failure case after:
commit error

### Questions:


Leemoonsoo:
Tested and LGTM. Thanks @khalidhuseynov for the update.

Merge to master if there're no further discussions.

PR #1713, opened 2016-12-01
[ZEPPELIN-1530] Update common scripts for Linux and Windows platforms

lresende:
### What is this PR for?
Avoid forcing set the DEBUG environment variable and also review the two common scripts for Linux and Windows platforms and update with missing environment variables.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1530](https://issues.apache.org/jira/browse/ZEPPELIN-1530)

CI Tests

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -141,7 +141,3 @@ export ZEPPELIN_RUNNER
 if [[ -z "$ZEPPELIN_IDENT_STRING" ]]; then
   export ZEPPELIN_IDENT_STRING="${USER}"
 fi
-
-if [[ -z "$DEBUG" ]]; then
-  export DEBUG=0
-fi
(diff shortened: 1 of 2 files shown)


bin/common.sh
```diff
@@ -135,13 +139,12 @@ if [[ -n "${JAVA_HOME}" ]]; then
...
 export ZEPPELIN_RUNNER
 
 if [[ -z "$ZEPPELIN_IDENT_STRING" ]]; then
   export ZEPPELIN_IDENT_STRING="${USER}"
 fi
 
-if [[ -z "$DEBUG" ]]; then
-  export DEBUG=0
+if [[ -z "$ZEPPELIN_INTERPRETER_REMOTE_RUNNER" ]]; then
+  export ZEPPELIN_INTERPRETER_REMOTE_RUNNER="bin\interpreter.sh"
```
felixcheung: shouldn't this be "bin/interpreter.sh"

lresende: Good catch, fixed.

Leemoonsoo:

PR #1720, opened 2016-12-02
Fix case where user use zeppelin in normal mode with token

anthonycorbacho:
### What is this PR for?
Fix the case where user use zeppelin in normal mode (doesnt use authentication) and set zeppelinhub token in zeppelin-env.sh.

### What type of PR is it?
[Bug Fix | Hot Fix]

### Questions:


minahlee:
Tested, LGTM

khalidhuseynov:
I see two tests failing `testGetAllNotes`, `testGetNote` under the corresponding class test. Other than that looks good.

anthonycorbacho:
@minahlee @khalidhuseynov thanks you for review, fixed the test and hopefully merging it as a hotfix soon

PR #1722, opened 2016-12-02
[ZEPPELIN-1776] substitute null check for the refactored Revision class

khalidhuseynov:
### What is this PR for?
This is to fix minor bug introduced by #1697 

### What type of PR is it?
Bug Fix

### Todos
* [x] - substitute null check

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1776

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -1453,7 +1453,7 @@ public class NotebookServer extends WebSocketServlet implements
     String commitMessage = (String) fromMessage.get("commitMessage");
     AuthenticationInfo subject = new AuthenticationInfo(fromMessage.principal);
     Revision revision = notebook.checkpointNote(noteId, commitMessage, subject);
-    if (!Revision.EMPTY.equals(revision)) {
+    if (!Revision.isEmpty(revision)) {
       List<Revision> revisions = notebook.listRevisionHistory(noteId, subject);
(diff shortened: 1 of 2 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1397,7 +1397,7 @@ private void checkpointNote(NotebookSocket conn, Notebook notebook,
     String commitMessage = (String) fromMessage.get("commitMessage");
     AuthenticationInfo subject = new AuthenticationInfo(fromMessage.principal);
     Revision revision = notebook.checkpointNote(noteId, commitMessage, subject);
-    if (revision != null) {
+    if (!Revision.EMPTY.equals(revision)) {
```
anthonycorbacho: maybe we can come up with a better implementation here like `revision.isEmpty()` or `Revision.isEmpty(revision)`

what do you think

khalidhuseynov: @anthonycorbacho yeah right, addressed in 13178aee18c0a4aed1dda0c72750e08c9ed9f53c

khalidhuseynov:
@anthonycorbacho this is done, what do you think?

anthonycorbacho:
LGTM thank for the contrib

khalidhuseynov:
ping

Leemoonsoo:
LGTM and merge to master if there're no further discussions.

PR #1724, opened 2016-12-03
[ZEPPELIN-1674] save barchart, areachart option

Leemoonsoo:
### What is this PR for?
Persist 
 - bar graph options (grouped, stacked)
 - area graph option: (stacked, stream, expanded)

### What type of PR is it?
Improvement

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1674

Change graph option and refresh the page and see if change is restored

### Questions:

### DIFF
--- a/zeppelin-web/src/app/tabledata/transformation.js
+++ b/zeppelin-web/src/app/tabledata/transformation.js
@@ -53,7 +53,6 @@ zeppelin.Transformation.prototype.renderSetting = function(targetEl) {
   // already readered
   if (this._scope) {
     var self = this;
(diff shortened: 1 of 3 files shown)


prabhjyotsingh:
Tested, it works as expected.
Should we have the same for scatterchart, and piechart as well ?

Leemoonsoo:
Thanks @prabhjyotsingh for the review.
I think other charts doesn't have these options.

Leemoonsoo:
@prabhjyotsingh Thanks.
Merge to master if there're no more comments.

PR #1733, opened 2016-12-07
[ZEPPELIN-1753] Fix blank notename issue

marchpig:
### What is this PR for?
Note ID is displayed on action bar after pressing enter key with blank notename.
This is because the scope variable for notename is used directly as the model of input field and it doesn't revert the blank notename back to its existing name even though the notename was actually not updated to the blank value. (Temporarily lost notename from scope)
To resolve this issue, I added a variable to `ElasticInputCtrl` for the model of input field.
This change can always make the notename valid.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1753

Input blank notename on action bar and..
- Check the notename on action bar
- Click 'Export this note' button and check the downloaded json filename.

blank-notename

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook-actionBar.html
+++ b/zeppelin-web/src/app/notebook/notebook-actionBar.html
@@ -16,8 +16,9 @@ limitations under the License.
     <div style="float: left; width: auto; max-width: 40%"
       ng-controller="ElasticInputCtrl as input">
       <input type="text" pu-elastic-input class="form-control2" placeholder="New name" style="min-width: 0px; max-width: 95%;"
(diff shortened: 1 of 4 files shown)


cuspymd:
@marchpig I checked it works nicely ~.
But this PR dosen't resolve another note name issue [ZEPPELIN-1763](https://issues.apache.org/jira/browse/ZEPPELIN-1763).
I reproduce ZEPPELIN-1763 after applying this PR.
It still occurs When changing note names repeatedly. 
anyway, LGTM!!

marchpig:
@cuspymd Thank you for review!
Maybe ZEPPELIN-1763 should be fixed by another PR. :smile:

Leemoonsoo:
Tested and LGTM.

PR #1740, opened 2016-12-09
ZEPPELIN-1773. Remove method destroy() in Interpreter

zjffdu:
### What is this PR for?
IMHO `destroy` is a little functional duplicated with `close`. And it seems currently only cassandra use destroy and it just calls `close()` https://github.com/apache/zeppelin/blob/master/cassandra/src/main/java/org/apache/zeppelin/cassandra/CassandraInterpreter.java#L231. So IMO, we could delete destroy method, also we can delete the destroy method in `InterpreterGroup`.  Besides, I move the terminate remoteInterpreterProcess in `close`.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1773

Just remove something so the existing test pass should be sufficient. 

### Questions:


Leemoonsoo:
LGTM. CI failure looks irrelevant.
Merge to master if there're no further discussions.

PR #1709, opened 2016-11-30
[HOTFIX] Make Spark download stable on CI

bzz:
### What is this PR for?
There has been issues with downloading\caching Spark, esp in #1689 #1696 
This is hotfix for Spark download on CI.

### What type of PR is it?
Hot Fix

### Todos
 - [x] do not use distrs.apache.org
 - [x] levirage `download-maven-plugin` cache for Spark download
 - [x] set timeout 1min and 5 re-tries on download
 - [x] un-pack them under `/target/` so `mvn clean` works as expected
 - [x] mute logs for `./testing/install_external_dependencies.sh`

### How should this be tested?
In CI logs, Spark should be downloaded by `spark-dependencies` and cached under `${HOME}/.m2/repository/.cache/maven-download-plugin`

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1741, opened 2016-12-09
[HOTFIX][ZEPPELIN-1779]Flaky test: PySparkInterpreterMatplotlibTest

agoodm:
### What is this PR for?
The matplotlib tests have stalled as a result of conda downloading version 2.0.0b1. While this was an old beta version and the issue has been fixed upstream (https://github.com/conda-forge/matplotlib-feedstock/issues/86), we still have not tested the inline backend for 2.0.0 which will be a release with a lot of major changes, hence this PR should be made anyway. I am also partially undoing one of the changes made in #1709 which suppressed the output of the dependency installation from conda. This change was very puzzling to me since dependency info isn't that verbose. I ended up determining the cause of this issue because I reverted this change and was then able to see the dependency versions in the log. I could see suppressing this being necessary for the R packages in case they aren't cached and a wall of compiler output gets printed out, but I see no real justification to mute it for python packages, especially given that they will only be installed for two short profiles.

### What type of PR is it?
Hot Fix 

### What is the Jira issue?
[ZEPPELIN-1779](https://issues.apache.org/jira/browse/ZEPPELIN-1779)

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ matrix:
 
 before_install:
   - echo "MAVEN_OPTS='-Xms1024M -Xmx2048M -XX:MaxPermSize=1024m -XX:-UseGCOverheadLimit -Dorg.slf4j.simpleLogger.defaultLogLevel=warn'" >> ~/.mavenrc
-  - ./testing/install_external_dependencies.sh > /dev/null 2>&1
+  - ./testing/install_external_dependencies.sh
   - ls -la .spark-dist ${HOME}/.m2/repository/.cache/maven-download-plugin || true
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks @agoodm lgtm

felixcheung:
LGTM

felixcheung:
Is this a hotfix? CI passed so should this be merged ASAP?

zjffdu:
+1 to merge it ASAP. It is a hotfix to fix the flaky test.

felixcheung:
merging if no more comment - I'll wait to tomorrow morning.

PR #1744, opened 2016-12-10
Better output of JDBC Interpreter.

astroshim:
### What is this PR for?
Currently, the output format of the JDBC Interpreter is the table format with all the results except EXPLAIN statement, It is better for the user experience to output only the result of the SELECT statement in table format.
This PR fixes the above issue and did some code refactoring.
Tested on mysql and aws athena.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1784

 - before

 - after

### Questions:


Leemoonsoo:
LGTM

bzz:
Looks great to me, modulo minor issue above.

Also, in my experience if CI failure reason is summarised and posted here by PR author - it makes merging/reviewing process much faster.

Leemoonsoo:

PR #1745, opened 2016-12-12
ZEPPELIN-1785. Remove ZEPPELIN_NOTEBOOK_DIR and ZEPPELIN_INTERPRETER_DIR from common.sh

zjffdu:
### What is this PR for?
We should remove `ZEPPELIN_NOTEBOOK_DIR` and `ZEPPELIN_INTERPRETER_DIR` from `common.sh` otherwise the corresponding property defined in `zeppelin-site.xml` won't take effect. I also check other properties like ZEPPELIN_CONF_DIR and ZEPPELIN_WAR. Although we define them explicitly in common.sh, we didn't expose them in document `install.md` and `zeppelin-site.xml.template`, so I think these are only for internal use and it is OK to keep them in common.sh although their correponding property in zeppelin-site.xml won't take effect too. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1785

Tested manually

### Questions:

…

### DIFF
--- a/bin/common.cmd
+++ b/bin/common.cmd
@@ -43,14 +43,6 @@ if not defined ZEPPELIN_WAR (
     )
 )
 
-if not defined ZEPPELIN_NOTEBOOK_DIR (
-    set ZEPPELIN_NOTEBOOK_DIR=%ZEPPELIN_HOME%\notebook
-)
-
-if not defined ZEPPELIN_INTERPRETER_DIR (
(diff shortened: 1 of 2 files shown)


zjffdu:
@Leemoonsoo  Please help review.

Leemoonsoo:
Thanks @zjffdu for the fix!
LGTM and merge to master if there're no more discussions

jongyoul:
LGTM

PR #1755, opened 2016-12-13
[ZEPPELIN-1801] To force end table - append %text

ravwojdyla:
### What is this PR for?
It seems there is a bug in scio introduced by recent changes. Table rendered from scio helper methods would include interpreter results as well as the data.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1801

Try:
```
%beam.scio
val (sc, args) = ContextAndArgs(argz)
sc.parallelize(Seq("foo", "foo", "bar")).countByValue.closeAndDisplay()
```
before and after the change - and notice extra meaningless columns before the fix.

### Questions:


felixcheung:
LGTM, CI good, merging if no more discussions

felixcheung:
@ravwojdyla feel free to ping me directly for scio changes

PR #1766, opened 2016-12-15
[MINOR] Set default interpreter \w first intp

AhyoungRyu:
### What is this PR for?
Currently we can select default interpreter when we create a new note like below.
screen shot 2016-12-15 at 4 41 15 pm

However, even if we don't set the default interpreter(don't select any interpreters) in this dialog and just create new note, the default interpreter is automatically set as `Spark` interpreter. This can be checked in the interpreter binding page. It doesn't make sense I think. 
So I removed `--select--` and set the default option in the select box with `Spark` as the interpreter binding page does. 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
N/A

1. Build only web \w `npm run build` under `zeppelin-web/` and run `npm run start` for dev mode. 
2. Create a new note and see the select box 

or just see the below screenshots :)

 - Before 
before

 - After
default_intp

### Questions:

### DIFF
--- a/zeppelin-web/src/components/noteName-create/note-name-dialog.html
+++ b/zeppelin-web/src/components/noteName-create/note-name-dialog.html
@@ -34,7 +34,6 @@ limitations under the License.
                       name="defaultInterpreter"
                       id="defaultInterpreter"
                       ng-options="option.name for option in interpreterSettings">
-                <option value="">--Select--</option>
               </select>
             </div>
           </div>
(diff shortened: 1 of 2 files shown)


zjffdu:
lgtm

Leemoonsoo:
LGTM!

AhyoungRyu:
@zjffdu @Leemoonsoo Thanks for your review! 
Will merge into master if there are no more comments.

PR #1771, opened 2016-12-15
[ZEPPELIN-1820] Fix action bar layout broken issue

marchpig:
### What is this PR for?
Action bar buttons and paragraph are overlapped when the browser width is small.
To fix this issue, this PR replaces `height` css property for action bar with `min-height` and adds a binding code for `window resize` to dynamically change the paragraph position based on the height of action bar.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1820

Change browser width

[Before]
actionbar-layout-broken

[After]
actionbar-layout-broken

### Questions:


Leemoonsoo:
Tested and works great.
Thanks @marchpig for the fix.

soralee:
Tested and It works well! 
LGTM :+1:

PR #1775, opened 2016-12-16
[ZEPPELIN-1745] View revisions in non-editable mode

khalidhuseynov:
### What is this PR for?
This is to make view of revisions non-editable

### What type of PR is it?
Improvement

### Todos
* [x] - disable action bar buttons
* [x] - disable per paragraph editing actions
* [x] - disable hotkeys

### What is the Jira issue?
[ZEPPELIN-1745](https://issues.apache.org/jira/browse/ZEPPELIN-1745)

1. initialize `GitNotebookRepo` by adding
```
export ZEPPELIN_NOTEBOOK_STORAGE="org.apache.zeppelin.notebook.repo.GitNotebookRepo"
```
in your `conf/zeppelin-env.sh`
2. go to any note and create few revisions using commit menu
3. then switch between revisions, during which you shouldn't be able to modify note

revision_view

### Questions:


bzz:
@khalidhuseynov Is it WIP?
 
Then please feel free to add WIP in PR title and ping back, when it's ready for reviews.

khalidhuseynov:
@bzz it's ready now, also /cc @cloverhearts @1ambda to help review

1ambda:
Glad to help reviewing this PR :) 
Let me check and then give you feedbacks :)

cloverhearts:
good feature!

1ambda:
For other reviewers 

- If you are using git 2.10+, Notebook revision feature will not work. Checkout [ZEPPELIN-1737](https://issues.apache.org/jira/browse/ZEPPELIN-1737)

khalidhuseynov:
@1ambda actually i've tried to reproduce [ZEPPELIN-1737](https://issues.apache.org/jira/browse/ZEPPELIN-1737) and even with git 2.11.0 wasn't able to do it. also from documentation [here](https://github.com/git/git/tree/master/Documentation/RelNotes) only from 2.11.0 `--allow-empty` commits was introduced, meaning only git versions 2.11+ could be a problem which is the most recent version and can be built from the source only for now. So I think reviewing this PR wouldn't much depend on git version, and [ZEPPELIN-1737](https://issues.apache.org/jira/browse/ZEPPELIN-1737) will be addressed separately.

1ambda:
FYI - you can install 2.11.0 using brew. 

@khalidhuseynov is right, that should be handled in the separate issue. Let's talk about it on ZEPPELIN-1737.

1ambda:
Works well as described. Here are few things to consider.

- It would be great if there is a signal (whatever) so let user know that this is committed version of note. It's confusing at first for me. 
- Enabling and modifying committed notes affects on the HEAD note. What about disabling **double click**  to committed paragraphs (you can check this in screenshot below)

pr1775

- the last one is not directly related with this issue. But It would be good to hide git notebook related icons if we are not using `GitNotebookRepo`

screen shot 2016-12-23 at 6 50 35 pm

Thanks for implementing great feature.

khalidhuseynov:
@1ambda thanks for review. answers in order: 
> It would be great if there is a signal (whatever) so let user know that this is committed version of note. It's confusing at first for me. (Assume that someone get the link for a committed note. He might wonder why he can not modify this note)

yeah that's good point, do you have something in mind regarding it? i was thinking whether it should be more of a mode switch (say different mode added in addition to report mode, etc)  or just represent visually (say gray out editor background like when running note), or something else.

> Enabling and modifying committed notes affects on the HEAD note. What about disabling double click to committed paragraphs (you can check this in screenshot below)

thanks for testing, addressed under 50e1472957446f7f907e2ea8b8eb19f4347eac8b. please check it out.

> the last one is not directly related with this issue. But It would be good to hide git notebook icons if we are other notebook repo rather than GitNotebookRepo. 

actually we have issue regarding this one to have `GitNotebookRepo` as default repo, in that case we don't need to hide it i believe.

Leemoonsoo:
Tested this branch and found

1. select HEAD 
2. and then switch to any revision. 
3. change height of graph or change pivot setting
4. HEAD is changed by 3)

@khalidhuseynov could you check scenario above?

khalidhuseynov:
@Leemoonsoo thanks for testing, right i could reproduce it. So in that case probably it shouldn't be allowed to change graph properties or pivot settings while viewing revisions, right? because otherwise it's changing `note.json`

1ambda:
@khalidhuseynov Thanks for reply! 

> yeah that's good point, do you have something in mind regarding it? i was thinking whether it should be more of a mode switch (say different mode added in addition to report mode, etc) or just represent visually (say gray out editor background like when running note), or something else.

We might use css transparent property to let user know this paragraphs are readonly like in running interpreters

screen shot 2016-12-28 at 12 54 15 pm

> thanks for testing, addressed under 50e1472. please check it out.

I checked and works well!

> actually we have issue regarding this one to have GitNotebookRepo as default repo, in that case we don't need to hide it i believe.

Glad to hear that! Actually i want it to be default 👍

khalidhuseynov:
@Leemoonsoo @cloverhearts @1ambda thanks for the review! I addressed all your comments and updated the screenshot above. Please check it out and let me know what you think. 

Also [CI logs](https://s3.amazonaws.com/archive.travis-ci.org/jobs/187629252/log.txt) seems to be failing only in selenium profile with flaky test from [ZEPPELIN-1836](https://issues.apache.org/jira/browse/ZEPPELIN-1836).

AhyoungRyu:
@khalidhuseynov How about disable "Interpreter binding" & "Note permission" buttons as well? Currently I can change the bound interpreters and note permission for the readonly-mode note.

khalidhuseynov:
@AhyoungRyu actually i thought about it and couldn't find a reason why not let user change "Interpreter bindings" and "Note permissions" even in read-only mode, since they're still not able to run or edit. Do you think it's not common sense and confusing? Let me know what you think and we can change that easily. I wasn't also sure about it

AhyoungRyu:
@khalidhuseynov 
>since they're still not able to run or edit

Yes you're right. I just thought it's not common sense as you told. Since the user can't run or edit anyway, but he can do sth in interpreter binding page and note permission setting page. That's why I suggested to block those features by disabling the icon buttons.

khalidhuseynov:
@AhyoungRyu I think it makes sense and addressed your feedback. please check it out and let me know if it's fine.

AhyoungRyu:
@khalidhuseynov Thanks for your consideration! Tested again and it works well as we said :)
And there is one minor thing in lookandfeel setting menu. Is this intended behavior? 
 - Head 
before


 - Read-only mode
screen shot 2016-12-31 at 3 49 53 pm

khalidhuseynov:
@AhyoungRyu thanks for review, it wasn't intended behaviour actually, fixed in bbaf286d60a07d4f5f7dd7a55e0f836ab116507d

AhyoungRyu:
Tested again and now I can see the text in the looknfeel dropdown menu of non-editable mode note. LGTM

khalidhuseynov:
merged master and finally CI is green. @Leemoonsoo ready for final check

Leemoonsoo:
Tested and LGTM. Thanks @khalidhuseynov for the improvement.

PR #1779, opened 2016-12-17
[ZEPPELIN-1835] Upgrade ElasticSearch to version 2.4.3

lresende:
### What is this PR for?
Upgrade ElasticSearch to version 2.4.3

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-1835](https://issues.apache.org/jira/browse/ZEPPELIN/ZEPPELIN-1835)

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -33,7 +33,7 @@
   <name>Zeppelin: Elasticsearch interpreter</name>
 
   <properties>
-    <elasticsearch.version>2.3.3</elasticsearch.version>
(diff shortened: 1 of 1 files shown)


1ambda:
Thanks for contribution 👍 Looks good to me.

felixcheung:
LGTM. I guess next would be to support ES 5.x? :)

merging if no more comment.

bzz:
Looks good to me.

Just FYI - there already is #1598 from @1ambda but AFAIK it is stuck and need a hand. 
That would be awesome, if we could come up with a plan to merge that one as well..

PR #1750, opened 2016-12-13
[ZEPPELIN-1796] Add 'copy to clipboard' next to the paragraph id

AhyoungRyu:
### What is this PR for?
There are lots of use cases that need to know each `Paragraph Id` in Zeppelin. 
e.g. [run paragraph using rest api](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/rest-api/rest-notebook.html#run-a-paragraph-asynchronously) / [`z.angularBind`](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/displaysystem/front-end-angular.html#basic-usage) / [export paragraphs using iframe](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/manual/publish.html) ...

So I added `Copy to clipboard` btn next the each `Paragraph Id` under paragraph setting menu using [angular-zeroclipboard](https://github.com/lisposter/angular-zeroclipboard). 
It'll be useful even if it's small feature :)

### What type of PR is it?
Improvement | Feature 

### What is the Jira issue?
[ZEPPELIN-1796](https://issues.apache.org/jira/browse/ZEPPELIN-1796)

### How should this be tested?
1. `npm run build` under `zeppelin-web/`
2. `npm run start` to start dev mode and browse `http://localhost:9000`
3. click gear icon placed top of the each paragraph -> hover the mouse on the paragraph id 
4. click it and paste the paragraph id to anywhere to check it's copied well or not

### Screenshots (if appropriate)
copy_pid

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no 
* Does this needs documentation? no

PR #1783, opened 2016-12-19
Zeppelin 727

PetroKostyuk:
### What is this PR for?
This PR adds button that allows user to copy link to paragraph in context. Then if user follows this link, it will load note and scroll to selected paragraph

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-727

1) Open note, choose one paragraph and click on 'Link to paragrapgh in context'.
screenshot from 2016-12-19 19-29-22

2) Copy link that is prompted to you.
screenshot from 2016-12-19 19-30-17

3) Open link in new tab or window.
4) Your notebook should load and scroll to linked paragraph.

### DIFF
--- a/zeppelin-web/src/app/app.controller.js
+++ b/zeppelin-web/src/app/app.controller.js
@@ -23,6 +23,24 @@
 
     var init = function() {
       $scope.asIframe = (($window.location.href.indexOf('asIframe') > -1) ? true : false);
+
+      $scope.paragrapScrollTimeout = 0;
+
+      $scope.$on('$includeContentLoaded', function() {
+        if ($scope.paragrapScrollTimeout !== 0) {
+          clearTimeout($scope.paragrapScrollTimeout);
+        }
+
+        $scope.paragrapScrollTimeout = setTimeout(function() {
+          var paraID = window.location.href.match(/paragraph=([^&#]+)/);
+          var para = jQuery('#' + paraID[1] + '_container');
+
+          if (para.length > 0) {
+            para[0].scrollIntoView(true);
+            window.scrollBy(0, -120);
+          }
+        }, 1000);
+      });
     };
 
     init();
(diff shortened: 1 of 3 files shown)


AhyoungRyu:
@PetroKostyuk Thanks for your contribution! It's really useful feature indeed. 
For the quick review, it would be better to edit PR title to sth like "[ZEPPELIN-727] Enable to jump to specific paragraph using anchor" instead of just "ZEPPELIN 727". Then other ppl can recognize what this PR implemented at a glance.

I tested and it works well as you described. Here are some my thoughts 

1. In my case, couldn't get the difference between screen shot 2016-12-20 at 11 54 50 am
 and screen shot 2016-12-20 at 11 54 16 am at the first time. Probably some users will be same with me. I know "Link this paragraph" is existed one, but "Publish this paragraph with iframe" instead of "Link this paragraph" would be more precise to prevent any confusion.

2. With this PR, we need to 
Click gear icon -> Click "Link to paragraph in context" menu  -> Ctrl+C and enter to copy the paragraph url -> Open new tab -> Paste the url 

IMPO using anchor has to be simple. Let's say we need specific paragraph url to give this location to someone. e.g. in Zeppelin official docs site,
anchor

The icon is exposed in the outside(need mouse hovering) and we can instinctively recognize what it is. 

So i think it would be better the anchor icon can be exposed in the outside not in the dropdown menu. And after we click that icon, the url will be exposed in the top header bar like I posted in the above, or maybe automatically copied right after we click it (please refer to #1750 ).

It's just my opinion in my head. So please take with grain of salt and feel free to share your thought! :D

FireArrow:
I agree with @AhyoungRyu. What she describes is the common way to link to paragraphs, so it seems reasonable to assume that is what users are used to.

PetroKostyuk:
Thank you for your replays! 
I'll try to implement what you suggested, it will indeed make this feature more user friendly.

aurely06:
Hi, 

Any update about this request ? The link to paragraph in context is available or not yet ? 
Thank you !

PetroKostyuk:
As it seems, this pull request is still open, so this feature is not in added to application. I no longer work on zeppelin project, so I don't know whether this feature will get published in future or no.

aurely06:
ok, thank you for your answser

PR #1786, opened 2016-12-20
[ZEPPELIN-1842] sync on start when anonymous allowed

khalidhuseynov:
### What is this PR for?
this is to keep backward compatibility when it's possible to sync on start in anonymous mode.

### What type of PR is it?
Bug Fix | Improvement

### Todos
* [x] - add sync call with condition

### What is the Jira issue?
[ZEPPELIN-1842](https://issues.apache.org/jira/browse/ZEPPELIN-1842)

1. enable secondary storage (e.g. s3, azure, [here](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/storage/storage.html#notebook-storage-in-s3))
2. keep anonymous mode without authentication
3. start zeppelin -> notes should be synced

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -93,6 +93,14 @@ public class NotebookRepoSync implements NotebookRepo {
       LOG.info("No storage could be initialized, using default {} storage", defaultStorage);
       initializeDefaultStorage(conf);
     }
+    // sync for anonymous mode on start
+    if (getRepoCount() > 1 && conf.getBoolean(ConfVars.ZEPPELIN_ANONYMOUS_ALLOWED)) {
+      try {
+        sync(AuthenticationInfo.ANONYMOUS);
+      } catch (IOException e) {
+        LOG.error("Couldn't sync on start ", e);
+      }
+    }
   }
 
   @SuppressWarnings("static-access")


khalidhuseynov:
this is ready for review, /cc @AhyoungRyu for help in review

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -93,6 +93,14 @@ public NotebookRepoSync(ZeppelinConfiguration conf) {
       LOG.info("No storage could be initialized, using default {} storage", defaultStorage);
       initializeDefaultStorage(conf);
     }
+    // sync for anonymous mode on start
+    if (getRepoCount() > 1 && conf.getBoolean(ConfVars.ZEPPELIN_ANONYMOUS_ALLOWED)) {
+      try {
+        sync(AuthenticationInfo.ANONYMOUS);
+      } catch (IOException e) {
+        LOG.error("Couldn't sync on start ", e);
```
felixcheung: is there a way to display this status to the user?

khalidhuseynov: @felixcheung good point, actually this is done before the server actually starts so at this time user wouldn't have his window open either, meaning it would be difficult to notify the user on that one. However next step plan on this feature is to move it to frontend and have button in `NotebookRepo` menu. Then it will be more self-explanatory and possible to display status on failure.

AhyoungRyu:
@khalidhuseynov Thanks for fixing. I test this on [ZeppelinHub](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/storage/storage.html#storage-in-zeppelinhub), seems it's synced well when I ran Zeppelin as anonymous user. So I can see my Zeppelin notebook in the connected ZeppelinHub instance. But one thing is (which is not in this PR's scope i guess),
notebookrepo

even if the anonymous user shouldn't allow to see the list of other instances in ZeppelinHub, the `No value` looks a bit unnatural. Hopefully we can show this with better way.  

Anyway looks good to me!

khalidhuseynov:
@AhyoungRyu thanks for review. you're right about `No value` field but it's not related to this issue and can be handled separately

AhyoungRyu:
@khalidhuseynov yup, hopefully it can be :)
Will merge this into master if there are no more comments.

PR #1753, opened 2016-12-13
[ZEPPELIN-1594] Support personalized mode

jongyoul:
### What is this PR for?
Enabling "Personalized Mode" for specific note

### What type of PR is it?
[Feature]

### Todos
* [x] - Backend work for supporting that feature
* [x] - Develop front-end UI

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1594

### How should this be tested?
1. config.personalizedMode = true in note.json
1. Navigate some jobs in two different browser

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1793, opened 2016-12-22
[MINOR] Refactoring Job and Paragraph

jongyoul:
### What is this PR for?
Job class has only two sub classes. I don't think we need Job abstraction class anymore, then I'll move all function of Job into Paragraph and InterpretJob. Paragraph is used into zeppelin-server and InterpretJob is done by remoteInterpreterServer. I think what we disconnect those two different classes is better to maintain each side of codes. This PR moves two variable - result and results - from Job to Paragraph in order to remove Job class. It also based on ZEPPELIN-1594(#1753), thus all reviewers click the last commit and review it.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Moved these variables into Paragraph

### What is the Jira issue?
N/A

All things are same as before

N/A

### Questions:


jongyoul:
```
Failed tests: 
```

This test passed in my local

jongyoul:
I've rebased after merging #1753

jongyoul:
Will merge it.

PR #1797, opened 2016-12-23
[ZEPPELIN-1854] make Per User mode in interpreter setting visible even if there is no role assigned to a user

prabhjyotsingh:
### What is this PR for?
There can be a case where user may have enabled shiro authentication, but may not have configured role based mapping.
In such a scenarios "Per User" mode in interpreter setting should be visible even if there are no role assigned to a user.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1854](https://issues.apache.org/jira/browse/ZEPPELIN-1854)

Enable authentication via shiro.ini, and in that remove "[roles]" block, now after restarting zeppelin server "Per User" should be visible when user edits any interpreter setting.

Before:
screen shot 2016-12-23 at 12 10 58 pm

After:
screen shot 2016-12-23 at 12 11 28 pm

### Questions:

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter-create/interpreter-create.html
+++ b/zeppelin-web/src/app/interpreter/interpreter-create/interpreter-create.html
@@ -59,7 +59,7 @@ limitations under the License.
                     Per Note
                   </a>
                 </li>
-                <li ng-if="ticket.ticket !== 'anonymous'">
+                <li ng-if="ticket.principal !== 'anonymous'">
                   <a style="cursor:pointer"
                      ng-click="setInterpreterRunningOption(setting.id, 'shared', 'scoped')">
                     Per User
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
@jongyoul can you help review this one.

jongyoul:
@cloverhearts Can you review it?

cloverhearts:
When shiro is enabled, an anonymous user may exist.
If so, this code stops all functionality for the user named anonymous.
Therefore, I have to additionally validate the role assigned to the user.
I think there are many opinions about this.
Whether a non-role user is a normal user or is able to change settings.
@prabhjyotsingh @jongyoul If you do not mind, can you tell us your opinion?

prabhjyotsingh:
Sure ` ticket.principal` sounds good to me. I'll do that.

r-kamath:
LGTM

prabhjyotsingh:

jongyoul:
@prabhjyotsingh When I design this feature, I've researched all possible combination for it and I decide to choose "Per note" mode first because I thought we should have the first criterium both of "Per note" and "Per user". Thus if you want to choose "Per user" at first, I think we should assume "Per note" is set to "Shared" and users cannot add "Per note" more anymore. Do you agree? If you also think like it, This PR looks good to me.

prabhjyotsingh:
@jongyoul I didn't get you on this. I haven't changed the order, only  `ng-if="ticket.ticket !== 'anonymous' && ticket.roles !== '[]'"` changed to `ng-if="ticket.principal !== 'anonymous'"`. So as, even if there is no role assigned to user, it can have option to instantiate an interpreter "Per user".

jongyoul:
@prabhjyotsingh I've misunderstood something. LGTM

jongyoul:
Will merge it.

PR #1807, opened 2016-12-27
ZEPPELIN-1861 Support custom interpreter.sh script to run interpreter

jongyoul:
### What is this PR for?
Supporting custom interpreter.sh for an individual interpreter. Some interpreters doesn't need current complicated interpreter.sh and want to use custom RemoteInterpreterServer class to launch their interpreters.

### What type of PR is it?
[Feature]

### Todos
* [x] - Add new way to be able to run custom interpreter.sh

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1861

Users can add 
```
"runner": {
      "linux": "interpreter.sh",
      "win": "interpreter.cmd"
    }
```
in their interpreter-setting.json. This path is relative from `${INTERPRETER_DIR}/${INTERPRETER_NAME}`

N/A

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.interpreter;
 
+import com.google.common.base.Joiner;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
@@ -237,7 +238,8 @@ public class InterpreterFactory implements InterpreterGroupFactory {
       interpreterInfo =
           new InterpreterInfo(r.getClassName(), r.getName(), r.isDefaultInterpreter(),
               r.getEditor());
-      add(r.getGroup(), interpreterInfo, r.getProperties(), defaultOption, r.getPath());
+      add(r.getGroup(), interpreterInfo, r.getProperties(), defaultOption, r.getPath(),
+          r.getRunner());
     }
 
     for (String settingId : interpreterSettingsRef.keySet()) {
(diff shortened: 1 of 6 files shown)


jongyoul:
I have tested md for this new way but I didn't add that code because I didn't think it's good to add that example.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -1130,15 +1135,21 @@ private Interpreter connectToRemoteRepl(String interpreterSessionKey, String cla
...
       String className, Properties property, String interpreterSettingId,
-      String userName, Boolean isUserImpersonate) {
+      String userName, Boolean isUserImpersonate, InterpreterRunner interpreterRunner) {
     int connectTimeout = conf.getInt(ConfVars.ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT);
     String localRepoPath = conf.getInterpreterLocalRepoPath() + "/" + interpreterSettingId;
     int maxPoolSize = conf.getInt(ConfVars.ZEPPELIN_INTERPRETER_MAX_POOL_SIZE);
+    String interpreterRunnerPath;
+    if (null != interpreterRunner) {
+      interpreterRunnerPath = Joiner.on(File.separator)
+          .join(interpreterPath, interpreterRunner.getPath());
```
Leemoonsoo: Can  we make it accept absolute path when path starts with `/` ?

felixcheung:
interesting. should we have documentation for this?

jongyoul:
I made a PR for editing those values in GUI. @felixcheung I think it would be better to make a new PR to describe interpreter-setting.json fully.

jongyoul:
Flaky test. Will merge it if there're no more discussion.

PR #1809, opened 2016-12-28
ZEPPELIN-1866. No meaningful exception for pig interpreter under tez engine

zjffdu:
### What is this PR for?
This PR is to fix the bug that there's no meaningful exception when using tez engine. 

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1866

Unit test is added and also manually test it. 

image

### Questions:

### DIFF
--- a/pig/src/main/java/org/apache/zeppelin/pig/PigUtils.java
+++ b/pig/src/main/java/org/apache/zeppelin/pig/PigUtils.java
@@ -171,6 +171,15 @@ public class PigUtils {
   private static String extractFromTezPigStats(TezPigScriptStats stats) {
 
     try {
+      if (stats.getReturnCode() == PigRunner.ReturnCode.UNKNOWN) {
+        LOGGER.warn("unknown return code, can't display the results");
+        return null;
+      }
+      if (stats.getPigContext() == null) {
+        LOGGER.warn("unknown exec type, don't display the results");
+        return null;
+      }
+
       Field userIdField = PigStats.class.getDeclaredField("userId");
       userIdField.setAccessible(true);
       String userId = (String) (userIdField.get(stats));
--- a/pig/pom.xml
+++ b/pig/pom.xml
@@ -177,6 +177,13 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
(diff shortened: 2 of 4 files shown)


zjffdu:
@felixcheung Please help review. CI failure is not relevant.
```
AngularElem
  The code passed to eventually never returned normally. Attempted 1 times over 185.23286 milliseconds. Last failure message: 0 was not equal to 1. (AbstractAngularElemTest.scala:64)
AngularElem
```

pig/src/main/java/org/apache/zeppelin/pig/PigUtils.java
```diff
@@ -171,6 +171,15 @@ public static String extractFromSimplePigStats(SimplePigStats stats) {
   private static String extractFromTezPigStats(TezPigScriptStats stats) {
 
     try {
+      if (stats.getReturnCode() == PigRunner.ReturnCode.UNKNOWN) {
+        LOGGER.warn("unknown return code, can't display the results");
```
felixcheung: should we log that return code then?

zjffdu: It helps for diagnosing.

felixcheung: I reread this, I guess if `getReturnCode` is returning UNKNOWN, there isn't any other meaningful code to put into log. So I'm good with the LOGGER.warn line.

felixcheung:
the same test profile failed in the last 2 CI runs, is that something to be concerned about?

zjffdu:
I think these are flaky test. Because all the failed test are not pig related. Changes in this PR is only on pig interpreter.

```
Failed tests: 
 
Tests run: 148, Failures: 1, Errors: 0, Skipped: 0
```

```
Results :
 
Tests in error: 
  ZeppelinIT.testAngularRunParagraph:315->AbstractZeppelinIT.waitForText:84 Â» StaleElementReference
 
```

felixcheung:
Cool. It also looks like CI is passing.
Merging if no more comment.

felixcheung:
Cool. ~~It also looks like CI is passing.~~ Hmm, I guess not..
Merging if no more comment.

PR #1813, opened 2016-12-29
[ZEPPELIN-1868]fix to not shows up the login button on Anonymous mode.

astroshim:
### What is this PR for?
This PR fixs to not shows up the login button on Anonymous mode.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1868

1. Give permission on note.
2. Turn off the shiro.
3. Delete note that you made on 1.
4. Then you can see the `login button` like as screen shot.

image

### Questions:

### DIFF
--- a/zeppelin-web/src/components/websocketEvents/websocketEvents.factory.js
+++ b/zeppelin-web/src/components/websocketEvents/websocketEvents.factory.js
@@ -70,13 +70,16 @@
       } else if (op === 'LIST_UPDATE_NOTE_JOBS') {
         $rootScope.$broadcast('setUpdateNoteJobs', data.noteRunningJobs);
       } else if (op === 'AUTH_INFO') {
-        BootstrapDialog.show({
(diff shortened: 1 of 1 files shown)


AhyoungRyu:
Yeah it makes sense. Tested and it's working well 👍

soralee:
@astroshim It's great work! I will test it!
But, I would like to suggest one thing. According to what I understand, your screenshot and `Then you can see the login button like as screen shot.` is the content before fixing bug in this PR description, right?
I think in this part, It makes confusing for your description of PR. 
So, it would be clear if there is content after fixing the bug.

soralee:
Tested and It works as expected like below!!

[Before]
z_1813_b

[After]
z_1813_a

AhyoungRyu:
LGTM. Merge if there are no more comments

PR #1815, opened 2016-12-29
[ZEPPELIN-1722] Modernize visualization/transformation using ES6 class and module syntax

Leemoonsoo:
### What is this PR for?
Modernize visualization/transformation using ES6 class and module syntax.
And remove global variable 'zeppelin'

### What type of PR is it?
Refactoring

### Todos
* [x] - Use ES6 class and module syntax

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1722

### Questions:


Leemoonsoo:
\cc @1ambda @felizbear

zeppelin-web/src/app/visualization/visualization.js
```diff
@@ -12,161 +12,161 @@
...
+      var self = this;
+      this._templateRequest(template).then(function(t) {
+        self._renderSetting(targetEl, t, scope);
+      });
+    } else {
+      this._renderSetting(targetEl, template, scope);
+    }
+  };
+
+  _renderSetting(targetEl, template, scope) {
```
felizbear: can we make it a function instead of method? there is no reason for this to be method to my understanding

zeppelin-web/src/app/visualization/visualization.js
```diff
@@ -12,161 +12,161 @@
...
+    var scope = this._createNewScope();
+    for (var k in setting.scope) {
+      scope[k] = setting.scope[k];
+    }
+    var template = setting.template;
+
+    if (template.split('\n').length === 1 &&
+        template.endsWith('.html')) { // template is url
+      var self = this;
+      this._templateRequest(template).then(function(t) {
```
felizbear: instead of confusing `self = this` it is better to use `=>`, e.g. `templateRequest().then(t => {/* use this here safely */})`; it will bind the context of the outer scope

felizbear:
I generally wish that api would be designed more like d3 or jquery, e.g. something along the lines of `visualization(el).render()`; without the need of calling it with `new`; it is more aligned with common practices in js libraries

zeppelin-web/src/app/visualization/visualization.js
Leemoonsoo: Can class have function? or you mean function inside of method?

felizbear: no, i mean outside of the class definition. don't forget, it's not java. we're talking about a module called visualization here. this module exports a class, but it can have private code as well, which could be, for example, a function `renderSetting`. does it make sense?

Leemoonsoo: ah i see make sense. let me update it.

zeppelin-web/src/app/visualization/visualization.js
Leemoonsoo: Removing `self` and use `=>` is not part of this PR's scope. but good to know. let me update this, too.

Leemoonsoo:
Addressed comments. 

I don't think style `visualization(el).transform(...).render();` is important topic here.
Because this `Visualization` class is intended to be used by developer who want to create new visualization. And the developer doesn't need to call either `visualization(el).transform(...).render(); ` or `new ;` because instantiation of visualization is managed by `result.controller.js`.

Leemoonsoo:
Merge to master if there're no more discussions

PR #1832, opened 2017-01-02
[ZEPPELIN-1884] Prevent NullPointerException when JDBC query.

astroshim:
### What is this PR for?
Sometimes users don't want to set the properties which they don't want to care about and `common.max_count` of JDBC interpreter is one of that property.
If it does not set the `common.max_count`, NullPointerException will occur.

### What type of PR is it?
Bug Fix 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1884

 - Set properties like as following image and run select query.
image

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -600,8 +600,11 @@ public class JDBCInterpreter extends Interpreter {
   }
 
   public int getMaxResult() {
-    return Integer.valueOf(
+    if (basePropretiesMap.containsKey(COMMON_KEY)) {
+      return Integer.valueOf(
         basePropretiesMap.get(COMMON_KEY).getProperty(MAX_LINE_KEY, MAX_LINE_DEFAULT));
+    }
+    return Integer.valueOf(MAX_LINE_DEFAULT);
   }
 
   boolean isConcurrentExecution() {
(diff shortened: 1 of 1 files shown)


felixcheung:
LGTM.
shouldn't `basePropretiesMap.containsKey(COMMON_KEY)` be checked in other places then?

astroshim:
@felixcheung Thank you for review!
Currently only `COMMON_KEY` uses for max line count in JDBC interpreter but as like you mentioned it seems more make sense to move the code to `open()` method.
Let me fix then.

astroshim:
@felixcheung I updated code, please review. Thanks in advance.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -175,6 +177,14 @@ public void open() {
     for (String propertyKey : basePropretiesMap.keySet()) {
       propertyKeySqlCompleterMap.put(propertyKey, createSqlCompleter(null));
     }
+    setMaxLineResults();
+  }
+
+  private void setMaxLineResults() {
+    if (basePropretiesMap.containsKey(COMMON_KEY) &&
+        basePropretiesMap.get(COMMON_KEY).containsKey(MAX_LINE_KEY)) {
+      maxLineResults = Integer.valueOf(basePropretiesMap.get(COMMON_KEY).getProperty(MAX_LINE_KEY));
```
felixcheung: I don't want to make this too complication, just to check, do we need to handle when someone set non-integer value in this config property?

that might be a separate, bigger issue though

felixcheung:
LGTM

PR #1087, opened 2016-06-25
Shell interpreter test and doc

fvaleri:
### What is this PR for?
Small refactoring to Shell interpreter.
Add missing unit test and initial doc.

### What type of PR is it?
Improvement, Documentation, Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/
* Put link here, and add [ZEPPELIN-*Jira number*] in PR title, eg. [ZEPPELIN-533]

### How should this be tested?
Outline the steps to test the PR here.

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #1833, opened 2017-01-02
[ZEPPELIN-1880] Fix shell interpreter output streaming result

AhyoungRyu:
### What is this PR for?
Shell interpreter streaming output had been available by #683, but currently it's broken after #1087 merged. This patch is for putting it back.

### What type of PR is it?
Bug Fix

### TODO
- [x] Fix test

### What is the Jira issue?
[ZEPPELIN-1880](https://issues.apache.org/jira/browse/ZEPPELIN-1880)

```
%sh

date && sleep 3 && date
```

the each timestamp must be printed as streaming output

 - before 
shellintpresultbefore

 - after 
shellintpresult

### Questions:

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -20,9 +20,7 @@ package org.apache.zeppelin.shell;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -86,7 +84,8 @@ public class ShellInterpreter extends Interpreter {
 
     try {
       DefaultExecutor executor = new DefaultExecutor();
-      executor.setStreamHandler(new PumpStreamHandler(outStream, outStream));
+      executor.setStreamHandler(new PumpStreamHandler(
+        contextInterpreter.out, contextInterpreter.out));
       executor.setWatchdog(new ExecuteWatchdog(Long.valueOf(getProperty(TIMEOUT_PROPERTY))));
       executors.put(contextInterpreter.getParagraphId(), executor);
       int exitVal = executor.execute(cmdLine);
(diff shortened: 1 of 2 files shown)


astroshim:
Hi.
In my case it seems doesn`t work properly.. What am i missing?
test1

Leemoonsoo:
Tested but i also see the same result @astroshim has.

AhyoungRyu:
Seems the streaming output works only once right after I start Zeppelin (or restart shell intp). Let me dig into more. Thanks for the review @astroshim @Leemoonsoo!

AhyoungRyu:
@astroshim @Leemoonsoo I guess the issue that I told in the above is not a matter of Shell intp itself. I found wired behaviour of current streaming output feature. This occurs in `master` branch. 

If I use only Spark, it's streaming output works properly like below.

 - Run only Spark (tried 3 times in a row)
```
(1 to 5).foreach{ i=>
    Thread.sleep(500)
    println("Hello " + i)
}
```
runonlyspark

But if I use Spark -> **Shell** -> Spark intp, then Spark's streaming output also doesn't work. 
 - Run Spark -> Shell -> Spark 
runsparkandsh


@Leemoonsoo Are there any suspicious points in this problem? The above issue is not in `0.6.2`.

soralee:
I tested and I also faced the problem in this branch. (It is same problem with above @AhyoungRyu's comment.)

At first time, It was worked very well. 
z1833_c

But when I rerun same paragraph, it was not worked. 
z1833_c_1

AhyoungRyu:
@soralee Thanks for testing it out. Right I think the rerunning issue is due to that I mentioned in [this comment](https://github.com/apache/zeppelin/pull/1833#issuecomment-270078954). 

@astroshim @Leemoonsoo BTW just to be making sure, did you build this branch including shell interpreter like `mvn clean package -DskipTests -pl 'shell, zeppelin-interpreter, zeppelin-server, zeppelin-zengine'` to test this patch? After build and restart Zeppelin, the result should be same with @soralee and me.

AhyoungRyu:
I rebased from master after #1922 and #1927 merged. Both are fixing the problem that I mentioned in [this comment](https://github.com/apache/zeppelin/pull/1833#issuecomment-270078954). Here is the screenshot. 
record

@astroshim Thanks again for the fix!!

AhyoungRyu:
@astroshim @Leemoonsoo @soralee If you don't mind, can someone help review this again?  :)

astroshim:
LGTM it's working well.

soralee:
Tested and it works very well!

Leemoonsoo:
Tested. LGTM

AhyoungRyu:
Thanks all you guys for testing it out! 
Will merge if there are no more comments.

PR #1834, opened 2017-01-03
[ZEPPELIN-1864] Improvement to show folder and note after searching note

soralee:
### What is this PR for?
This PR is for improvement to show folder and note after using the filter. And I found some bugs and fixed like this.

the following list is improvement and bug fixed.
1. After using the filter in Zeppelin Home, every note is shown by the form [FolderName/NoteName] like below. It would be nice to show folder icon and note icon as previous status.
z1864_f_b
 
2. When using the filter, some functions and icons next to note and folder are disappeared.
z1864_f_2_b

3. When using the filter, the `Rename Note` function doesn't work. ( When I was handling this PR, this bug was discovered.)
z1864_f_4_b

### What type of PR is it?
[ Bug Fix | Improvement ]

### Todos
* [x] - improve to show folder and note when finishing to use filter.
* [x] - some functions and icons next to note and folder are appeared when using filter.
* [x] - fix `Rename Note` function when using the filter.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1864

 - **[Test 1]** Before/After using the filter, check the list structure of your notes. It must be same appearance.

 - **[Test 2]** When using the filter, check the notes. Some icons and functions next to note and folder must be appeared.

 - **[Test 3]** When using the filter, click `Rename Note`. It must be appeared the path of Note.

**Test 1**.
z1864_f_1_a

**Test 2**.
z1864_f_2_a

**Test 3**.
z1864_f_4_a

### Questions:

### DIFF
--- a/zeppelin-web/src/app/home/home.controller.js
+++ b/zeppelin-web/src/app/home/home.controller.js
@@ -42,6 +42,7 @@ function HomeCtrl($scope, noteListDataFactory, websocketMsgSrv, $rootScope, arra
 
   $scope.isReloading = false;
   $scope.TRASH_FOLDER_ID = TRASH_FOLDER_ID;
+  $scope.query = {q: ''};
 
   var initHome = function() {
     websocketMsgSrv.getHomeNote();
@@ -124,6 +125,7 @@ function HomeCtrl($scope, noteListDataFactory, websocketMsgSrv, $rootScope, arra
     noteActionSrv.removeFolder(folderId);
   };
 
+<<<<<<< c41d4682afdce3bb5fb491229ec442e2617b6ca4
   $scope.emptyTrash = function() {
     noteActionSrv.emptyTrash();
   };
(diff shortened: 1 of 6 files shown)


AhyoungRyu:
@soralee Thanks for the improvement. I've also seen those issues when I was using filter. Let me test this :)

AhyoungRyu:
@soralee Tested and works well for all 3 cases you mentioned. But for the consistency, don't we need to apply this to the Navbar one? I mean currently filter doesn't work for folder structured notes.  
navbar

soralee:
Thanks for detailed test @AhyoungRyu! :heart_eyes: 
Sure, actually I missed the filter in the Navbar! 
Let me check and update PR!!

soralee:
@AhyoungRyu I updated the filter of Navbar. 
But, the folder of notes in navbar don't have previous status. Because If using the hidden value of the note, it works together both the filter in the home and navbar.
z1834_u
I think it will take more time to add this feature so it would be nice to improve this part by separating PR.

AhyoungRyu:
@soralee Thanks for the update and kind reminder! Let me test again :)

AhyoungRyu:
You mean if we 

open the folder in the **_navbar_** (e.g. Zeppelin Tutorial)-> type some text in the search bar -> remove all text -> the opened folder is closed 

But this isn't happened in **_home_**. Am I right? I tested again and checked what you mentioned. I think current status is good for me and we can keep improving it if you need more time to handle this part.

soralee:
@AhyoungRyu Yes, you're right. 
I mean, If I change **navbar** folder option, it works together both the filter in the **home** and **navbar**. So, I did not include this part because I think this improvement should be opened other PR.

Thanks for review again!

AhyoungRyu:
Yup thanks for the explanation. Merge if there are no more discussions.

soralee:
Thank you @AhyoungRyu :+1:

PR #1830, opened 2017-01-01
ZEPPELIN-1867. Update document for pig interpreter and add one sample note

zjffdu:
### What is this PR for?
* Minor update for pig interpreter
* Add one sample pig tutorial note which do the same thing as the spark tutorial note. 

### What type of PR is it?
[Improvement | Documentation ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1867

### How should this be tested?
Tested manually

### Screenshots (if appropriate)
image

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #1837, opened 2017-01-04
ZEPPELIN-1901. Output schema might be null for PigQueryInterpreter

zjffdu:
### What is this PR for?
The output schema might be null if user doesn't specify it explicitly. In this PR, I will use 'col_{pos}' to replace the column name if it is null. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1901

Unit test is added.

### Questions:


zjffdu:
@felixcheung Please help review, I will add document in #1830 to encourage user to specify schema explictly.

zjffdu:
Ping @felixcheung , CI failure is not relevant.

```
AngularElem
  The code passed to eventually never returned normally. Attempted 1 times over 192.484891 milliseconds. Last failure message: 0 was not equal to 1. (AbstractAngularElemTest.scala:64)
```

felixcheung:
LGTM
merging if no more comment

PR #1844, opened 2017-01-05
[ZEPPELIN-1825] Use versioned notebook storage by default 

khalidhuseynov:
### What is this PR for?
Due to recent work on versioned notes under ZEPPELIN-1044, we can use by default Git storage which extends local storage. so the behaviour will be same as before with additional ability to use versioning out of box without configuring it.

### What type of PR is it?
Improvement |  Refactoring

### Todos
* [x] - change VFS to Git

### What is the Jira issue?
[ZEPPELIN-1825](https://issues.apache.org/jira/browse/ZEPPELIN-1825)

start zeppelin and by default be able to use version control (if git client installed)

### Questions:


felixcheung:
LGTM
any documentation to update?

khalidhuseynov:
@felixcheung i didn't see much places to update docs, but maybe @AhyoungRyu knows if there's something to be updated?

Leemoonsoo:
`VFSNotebookRepo` can be configured to use [other filesystems](https://commons.apache.org/proper/commons-vfs/filesystems.html) instead of local filesystem.
Those user might be confused when (default) notebook repo is changed to `GitNotebookRepo`.

How about add an item in [upgrade.md](https://github.com/apache/zeppelin/blob/master/docs/install/upgrade.md#upgrading-from-zeppelin-06-to-07) for this change?

khalidhuseynov:
@Leemoonsoo makes sense, added in 7cb133c8dc59357b0572c4f462ae04d5bbb55cce

Leemoonsoo:
Thanks @khalidhuseynov for the improvement.

LGTM and merge if no more comments!

AhyoungRyu:
@khalidhuseynov Seems a default value of `ZEPPELIN_NOTEBOOK_STORAGE ` under [Apache Zeppelin configuration](https://github.com/apache/zeppelin/blob/master/docs/install/install.md#apache-zeppelin-configuration)  table also needs to be updated :)

khalidhuseynov:
thanks @AhyoungRyu, updated!

AhyoungRyu:
LGTM 👍

PR #1846, opened 2017-01-05
ZEPPELIN-1770. Restart only the client user's interpreter when restarting interpreter setting

zjffdu:
### What is this PR for?
This PR would only restart the trigger user's interpreter rather than all the interpreter. So that restarting won't affect other users. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1770

Tested manually.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -144,6 +144,26 @@ public class InterpreterSetting {
     return key;
   }
 
+  private String getInterpreterSessionKey(String user, String noteId) {
+    InterpreterOption option = getOption();
+    String key;
+    if (option.isExistingProcess()) {
+      key = Constants.EXISTING_PROCESS;
+    } else if (option.perNoteScoped() && option.perUserScoped()) {
+      key = user + ":" + noteId;
+    } else if (option.perUserScoped()) {
+      key = user;
+    } else if (option.perNoteScoped()) {
+      key = noteId;
+    } else {
+      key = "shared_session";
+    }
+
+    logger.debug("Interpreter session key: {}, for note: {}, user: {}, InterpreterSetting Name: " +
+        "{}", key, noteId, user, getName());
+    return key;
+  }
+
   public InterpreterGroup getInterpreterGroup(String user, String noteId) {
     String key = getInterpreterProcessKey(user, noteId);
     if (!interpreterGroupRef.containsKey(key)) {
@@ -194,11 +214,11 @@ public class InterpreterSetting {
     if (user.equals("anonymous")) {
       user = "";
     }
-    String key = getInterpreterProcessKey(user, "");
-
+    String processKey = getInterpreterProcessKey(user, "");
+    String sessionKey = getInterpreterSessionKey(user, "");
     InterpreterGroup groupToRemove = null;
     for (String intpKey : new HashSet<>(interpreterGroupRef.keySet())) {
-      if (intpKey.contains(key)) {
+      if (intpKey.contains(processKey)) {
         interpreterGroupWriteLock.lock();
         groupToRemove = interpreterGroupRef.remove(intpKey);
         interpreterGroupWriteLock.unlock();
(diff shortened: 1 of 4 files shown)


felixcheung:
oh gosh :)
is it possible to have test for this?

jongyoul:
Good catch. LGTM @felixcheung We had too big InterpreterFactory, InterpreterSetting and InterpreterGroup. I will refactor all of this class by next major release - 0.8.0 - and will make them testable. I think we have to do it now. :-)

felixcheung:
Sounds good!!

zjffdu:
@jongyoul agreen on refactoring Interpreter component, I also mention that in other PRs. I'd love to help on that :)

Leemoonsoo:
@zjffdu Thanks for the patch!

I also think this PR deserves a unittest to make sure correct restart behavior, like @felixcheung mentioned, if it is not too much difficult.

zjffdu:
@Leemoonsoo Could you point me the right place to put this unit test? I think this may need to enable shiro, but I didn't find any test with shiro enabled.

Leemoonsoo:
[AuthenticationIT.java](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/integration/AuthenticationIT.java) is test that uses zeppelin with shiro enabled. However it's selenium test and bit difficult to make.

I think it's possible to make unittest by calling methods in InterpreterFactory class, directly.
Like [this test](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/InterpreterFactoryTest.java#L148).

Let me know if you need any help.

zjffdu:
Thanks @Leemoonsoo  Unit test is added. Please help review \cc @felixcheung @jongyoul

zjffdu:
CI failure is not relevant.

```
Failed tests: 
AngularElem
  The code passed to eventually never returned normally. Attempted 1 times over 2.966494567 seconds. Last failure message: 0 was not equal to 1. (AbstractAngularElemTest.scala:64)
AngularElem
```

jongyoul:
Tests LGTM. Thanks!

zjffdu:
ping @jongyoul @Leemoonsoo @felixcheung

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -190,10 +210,30 @@ void closeAndRemoveInterpreterGroup(String noteId) {
     }
   }
 
+  void closeAndRemoveInterpreterGroupByUser(String user) {
+    if (user.equals("anonymous")) {
```
felixcheung: it'd probably be good to have a const for "anonymous" and "" - easier to avoid typos etc.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -190,10 +210,30 @@ void closeAndRemoveInterpreterGroup(String noteId) {
...
+      user = "";
+    }
+    String processKey = getInterpreterProcessKey(user, "");
+    String sessionKey = getInterpreterSessionKey(user, "");
+    InterpreterGroup groupToRemove = null;
+    for (String intpKey : new HashSet<>(interpreterGroupRef.keySet())) {
```
felixcheung: could you elaborate - why not do containsKey?

zjffdu: You might misunderstand the logic here. I iterate through the keyset and then judge whether the key string contains the processKey. It can not be done by Map.containsKey

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
jongyoul: I agree with @felixcheung and we also have to handle all of "anonymous" hardcoded. Actually, zeppelin-server module has a constant of "anonymous" thus we have to move it into another module like zeppelin-interpreter. I, however, it's not the scope of this PR.

jongyoul:
LGTM, merging if there's no more discussion

PR #1520, opened 2016-10-14
[ZEPPELIN-1061] Select default interpreter while creating note

kavinkumarks:
### What is this PR for?

This handles setting the default interpreter on creating a note through the zeppelin UI.
### What type of PR is it?

Feature
### Todos

NA
### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1061
### How should this be tested?
- Go to the 'Create Note' dialog and provide the name with the default interpreter selected.After clicking on the Create button, the selected interpreter should be shown properly in the interpreter binding section.
- If there is no interpreter selected, then the system default will be used.
### Screenshots (if appropriate)

zeppelin-1061
### Questions:
- Does the licenses files need update? No
- Is there breaking changes for older versions? No
- Does this needs documentation? No

PR #1849, opened 2017-01-05
[ZEPPELIN-1902] Fix blank default intp select box after Zeppelin started when Shiro on

AhyoungRyu:
### What is this PR for?
After #1520 we can set default interpreter when we create new note. Currently this feature is implemented to get `interpreterSetting` list when Zeppelin is initiated (can be checked in dev console). And it needs user's info. So there can be two cases like below.

 - Shiro off 
It tries to get interpreter setting lists as `anonymous` user
shirooff
In this case, the problem is not happened. If we don't turn Shiro on, it doesn't need user's info and Zeppelin automatically creates fake user(?) info as `anonymous`. So there is no problem to get the interpreter setting lists. 

 - Shiro on
tries to get interpreter setting lists as '  ' user since we're not loggined yet. 
shiroon 
So it's failed to get interpreter setting lists since it doesn't have any user info. It needs reloading to get it. 
whenshiroon

Actually we don't need to send `GET_INTERPRETER_SETTINGS` msg when Zeppelin started. It would be fine when we click "Create new note" button. So I removed `init()` (it's used only for getting interpreter settings) and made `getInterpreterSettings` called only when user clicks "Create new note" in both Zeppelin home and Navbar. 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1902](https://issues.apache.org/jira/browse/ZEPPELIN-1902)

In master (not this branch)
1. [Turn Shiro on](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/security/shiroauthentication.html#1-enable-shiro) and start(restart) Zeppelin 
2. Click "Create new note" in home (or Navbar)
3. the default interpreter select box will be like 
screen shot 2017-01-05 at 7 16 08 pm

but it'll be okay after reloading. It shouldn't be like that with this patch. 

after fixing this problem
whenshiroonafter
 

### Questions:

### DIFF
--- a/zeppelin-web/src/app/home/home.html
+++ b/zeppelin-web/src/app/home/home.html
@@ -133,7 +133,7 @@ limitations under the License.
           <div>
             <h5><a href="" data-toggle="modal" data-target="#noteImportModal" style="text-decoration: none;">
               <i style="font-size: 15px;" class="fa fa-upload"></i> Import note</a></h5>
-            <h5><a href="" data-toggle="modal" data-target="#noteNameModal" style="text-decoration: none;">
+            <h5 ng-controller="NotenameCtrl as notenamectrl"><a href="" data-toggle="modal" data-target="#noteNameModal" style="text-decoration: none;" ng-click="notenamectrl.getInterpreterSettings()">
               <i style="font-size: 15px;" class="icon-notebook"></i> Create new note</a></h5>
             <ul id="notebook-names">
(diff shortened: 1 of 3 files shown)


AhyoungRyu:
CI is all green now. Ready for review.

minahlee:
Tested and LGTM. Merge if there's no discussion.

PR #1850, opened 2017-01-05
[ZEPPELIN-1731] Fix unexpected EditOnDoubleClick

minahlee:
### What is this PR for?
This PR fixes wrong `EditOnDoubleClick` value in editor setting when default bound interpreter has changed.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1731](https://issues.apache.org/jira/browse/ZEPPELIN-1731)

1. Set default interpreter to `md`.
2. Write below code in first paragraph and execute the paragraph. The editor should be hidden when you run.
```
## abc
pwd
```
3. Double click the result to make editor to be shown.
4. Change default interpreter to `sh`. At this point syntax highlight on `## abc` should be changed from blue to green.
5. Execute paragraph and see if editor stays displayed.

**Before**
jan-05-2017 21-39-46

**After**
jan-05-2017 21-38-57

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -584,7 +584,8 @@
       console.log('Interpreter bindings %o saved', selectedSettingIds);
 
       _.forEach($scope.note.paragraphs, function(n, key) {
-        if (n.text && !n.text.startsWith('%')) {
+        var regExp = /^\s*%/g;
+        if (n.text && !regExp.exec(n.text)) {
           $scope.$broadcast('saveInterpreterBindings', n.id);
         }
       });
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -582,6 +582,13 @@
       }
       websocketMsgSrv.saveInterpreterBindings($scope.note.id, selectedSettingIds);
       console.log('Interpreter bindings %o saved', selectedSettingIds);
+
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -582,6 +582,13 @@
       }
       websocketMsgSrv.saveInterpreterBindings($scope.note.id, selectedSettingIds);
       console.log('Interpreter bindings %o saved', selectedSettingIds);
+
+      _.forEach($scope.note.paragraphs, function(n, key) {
+        if (n.text && !n.text.startsWith('%')) {
```
felixcheung: should we handle the case where the paragraph text starts with `   %foo` or
```

%foo
```

minahlee: Thanks for reviewing!
If user defined `%foo` explicitly in paragraph, we don't need to change `editorSetting` cause `editorSetting` such as `language`, `editOnDblClick` of this paragraph will be same even if the default interpreter changed. (This field is saved in `note.json` file when user type `%foo`.)
The changes will be only needed to those without `%foo` specified.

felixcheung: but what I mean is how it should behavior when the `text` doesn't `startsWith` `%` because of the empty space, but still has a `%`

minahlee: I misunderstood it before. I handled it in 5553f0c, it should be ok now.

PR #1852, opened 2017-01-06
[ZEPPELIN-1910] DON'T show the same dialogs multiple times when don't have permission for run all paragraphs (BUG)

1ambda:
### What is this PR for?

DON't show the multiple same dialog when user doesn't have permission for **Run all paragraphs** inside a note

#### Implementation details

- Introduce new websocket message `RUN_ALL_PARAGRAPHS` since we need to broadcast *bringing dialog* message only once 
(We did same thing about `CLEAR_ALL_PARAGRAPHS`)
- Refactor `NotebookServer.runParagraph` to avoid duplication
- Add necessary functions to backend and frontend

### What type of PR is it?
[Bug Fix]

### Todos

Fixed at once

### What is the Jira issue?
[ZEPPELIN-1910](https://issues.apache.org/jira/browse/ZEPPELIN-1910)

1. Set permission to a note
2. Enable shiro and login as an user **who doens't have write permission**
3. Click *Run all paragraphs* button

#### 1. Before

bug-multiple-dialog

#### 2. After

run-all-after-fixed

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -36,6 +36,7 @@ import javax.servlet.http.HttpServletRequest;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.Sets;
+import com.google.gson.*;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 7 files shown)


Leemoonsoo:
Thanks @1ambda for the fix.
LGTM and merge if no more comments!

PR #1853, opened 2017-01-06
[ZEPPELIN-1911] Adding how to install evaluate package to build R.

astroshim:
### What is this PR for?
This PR inform how to avoid build failing with R.

### What type of PR is it?
Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1911

### Questions:


felixcheung:
`knitr` should install `evaluate` as a dependency actually?
do you know if that is not the case?

astroshim:
@felixcheung You're right! `evaluate` is dependency package of `knitr`.
Thanks for pointing out. It's my misunderstood. 
Let me close this PR.

PR #1856, opened 2017-01-06
remove org.apache.spark.sql.hive.HiveSharedState class check

lichenglin:
### What is this PR for?
remove org.apache.spark.sql.hive.HiveSharedState class check
because it has been removed since spark 2.1.x

### What type of PR is it?
[Bug Fix ]

### What is the Jira issue?
[ZEPPELIN-1909](https://issues.apache.org/jira/browse/ZEPPELIN-1909)

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -205,7 +205,6 @@ public class SparkInterpreter extends Interpreter {
   private boolean hiveClassesArePresent() {
     try {
       this.getClass().forName("org.apache.spark.sql.hive.HiveSessionState");
-      this.getClass().forName("org.apache.spark.sql.hive.HiveSharedState");
       this.getClass().forName("org.apache.hadoop.hive.conf.HiveConf");
       return true;
     } catch (ClassNotFoundException | NoClassDefFoundError e) {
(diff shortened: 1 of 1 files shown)


felixcheung:
looks good, why doesn't CI tests for Spark 2.1 fails for this?

felixcheung:
would be great to understand why it worked before, but might merge this fix first.
merging if no more comment

lichenglin:
I don't know either，It fails with spark 1.6.1.
May be the class check is still needed by 1.6,1?I'm not sure.
How can I trigger the build once more on travis-cli?

felixcheung:
I don't think we need to check for all 3 classes, so I believe this fix is good.
what I don't understand is why it is failing for you but running just fine in travis CI where we have tests for Spark 2.1 - which is why I'm asking.

to trigger test, close this PR and reopen it.

lichenglin:
e...
It works on my local mac with spark 2.1.0.
And pass the build with 2.1.0.
I seems that the build fails with 
{code}
  "env": "PYTHON=\"2\" SCALA_VER=\"2.10\" SPARK_VER=\"1.6.1\" HADOOP_VER=\"2.6\" PROFILE=\"-Pspark-1.6 -Phadoop-2.6 -Ppyspark\" BUILD_FLAG=\"package -pl spark,python -am -DskipTests -DskipRat\" TEST_FLAG=\"verify -DskipRat\" TEST_PROJECTS=\"-pl zeppelin-interpreter,zeppelin-display,spark-dependencies,spark,python -Dtest=org.apache.zeppelin.spark.PySpark*Test,org.apache.zeppelin.python.* -Dpyspark.test.exclude='' -DfailIfNoTests=false\"",
{code}

I think we'd better reopen the PR and build once more.
Maybe the class check is still necessary for spark 1.6.1

felixcheung:
it's only the python2 + spark 1.6 profile. It ran fine with the other spark 1.6 profile so I think we are good.

PR #1858, opened 2017-01-06
[ZEPPELIN-1229] Clear browser cache with version hashtag.

astroshim:
### What is this PR for?
This PR avoids browser cache.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1229

1. Run old version of zeppelin(I was using 0.6.1) and show main page up in your browser.
2. Stop zeppelin.
3. Run this PR of Zeppelin and show page up on the same browser of 1.
4. check main page changed.

- before (Job menu doesn't show)

- after (Job menu show)

### Questions:


corneadoug:
@astroshim Could you provide a way to reproduce the problem, or screenshot of the issue?
If there is a cache issue with html files, I'm not sure this PR would solve it. Considering the solution you provide, I would guess it only solves a few cases and would need to be applied to all ng-include and templateUrl.

astroshim:
@corneadoug I updated screenshot and codes.
Could you help to review?

corneadoug:
@astroshim still some missing in the $routeProvider.
Otherwise I just want to point out that the original goal of the cache buster is to add a hash in the file name so that it is different when the code is built with changes, however by using the method of this PR the cache for the html will be broken every time we refresh the page.

Was there any attempt to do it differently?

astroshim:
@corneadoug Let me find different of more efficient way.

astroshim:
@corneadoug please help to review.

minahlee:
Tested LGTM. @felizbear do you mind to review this?

felizbear:
LGTM

minahlee:

PR #1861, opened 2017-01-06
ZEPPELIN-1293. Livy Interpreter: Automatically attach or create to a new session

zjffdu:
### What is this PR for?
By default, livy session will expire after one hour. This PR would create session automatically for user if session is expired, and would also display the session expire information in frontend. The expire message would only display at the first time of session recreation, after that the message won't be displayed. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1293

Tested manually. 
image

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
@@ -259,7 +259,8 @@ public abstract class BaseLivyInterprereter extends Interpreter {
     if (sessionExpired) {
       InterpreterResult result2 = new InterpreterResult(result.code());
       result2.add(InterpreterResult.Type.HTML,
-          "<font color=\"red\">Previous session is expired, new session is created.</font>");
+          "<font color=\"red\">Previous livy session is expired, new livy session is created. " +
+              "Paragraphs that depend on this paragraph need to be re-executed!" + "</font>");
       for (InterpreterResultMessage message : result.message()) {
         result2.add(message.getType(), message.getData());
       }
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
@@ -46,12 +46,11 @@ public abstract class BaseLivyInterprereter extends Interpreter {
   protected static final Logger LOGGER = LoggerFactory.getLogger(BaseLivyInterprereter.class);
   private static Gson gson = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create();
 
-  protected SessionInfo sessionInfo;
+  protected volatile SessionInfo sessionInfo;
   private String livyURL;
   private long sessionCreationTimeout;
   protected boolean displayAppInfo;
-  private boolean sessionExpired;
-  private AtomicBoolean isInitingSession = new AtomicBoolean(false);
+  private AtomicBoolean sessionExpired = new AtomicBoolean(false);
 
   public BaseLivyInterprereter(Properties property) {
     super(property);
@@ -90,16 +89,17 @@ public abstract class BaseLivyInterprereter extends Interpreter {
         // livy 0.2 don't return appId and sparkUiUrl in response so that we need to get it
         // explicitly by ourselves.
         sessionInfo.appId = extractStatementResult(
-            interpret("sc.applicationId", false).message()
+            interpret("sc.applicationId", false, false).message()
                 .get(0).getData());
       }
 
       interpret(
-          "val webui=sc.getClass.getMethod(\"ui\").invoke(sc).asInstanceOf[Some[_]].get", false);
+          "val webui=sc.getClass.getMethod(\"ui\").invoke(sc).asInstanceOf[Some[_]].get",
+          false, false);
       if (StringUtils.isEmpty(sessionInfo.appInfo.get("sparkUiUrl"))) {
         sessionInfo.webUIAddress = extractStatementResult(
             interpret(
-                "webui.getClass.getMethod(\"appUIAddress\").invoke(webui)", false)
+                "webui.getClass.getMethod(\"appUIAddress\").invoke(webui)", false, false)
                 .message().get(0).getData());
       } else {
         sessionInfo.webUIAddress = sessionInfo.appInfo.get("sparkUiUrl");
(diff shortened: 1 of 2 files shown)


zjffdu:
@prabhjyotsingh @felixcheung Please help review.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -21,10 +21,7 @@
 import com.google.gson.GsonBuilder;
 import com.google.gson.annotations.SerializedName;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.zeppelin.interpreter.Interpreter;
-import org.apache.zeppelin.interpreter.InterpreterContext;
-import org.apache.zeppelin.interpreter.InterpreterResult;
-import org.apache.zeppelin.interpreter.InterpreterUtils;
+import org.apache.zeppelin.interpreter.*;
```
felixcheung: what's our coding style guide on multiple imports? is it clearer to list them out

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -208,7 +208,22 @@ private SessionInfo getSessionInfo(int sessionId) throws LivyException {
 
   public InterpreterResult interpret(String code, boolean displayAppInfo)
       throws LivyException {
-    StatementInfo stmtInfo = executeStatement(new ExecuteRequest(code));
+    StatementInfo stmtInfo = null;
+    try {
+      stmtInfo = executeStatement(new ExecuteRequest(code));
+    } catch (SessionNotFoundException e) {
+      LOGGER.warn("Livy session {} is expired, new session will be created.", sessionInfo.id);
+      this.sessionExpired = true;
```
felixcheung: instead of saving this as a state, would be more self-contain to pass this as a flag to `appendSessionExpire` below?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -219,7 +234,23 @@ public InterpreterResult interpret(String code, boolean displayAppInfo)
...
+  }
+
+  private InterpreterResult appendSessionExpire(InterpreterResult result) {
+    if (sessionExpired) {
+      InterpreterResult result2 = new InterpreterResult(result.code());
+      for (InterpreterResultMessage message : result.message()) {
+        result2.add(message.getType(), message.getData());
+      }
+      result2.add(InterpreterResult.Type.HTML,
+          "<hr/><font color=\"red\">Previous session is expired, new session is created.</font>");
```
felixcheung: perhaps it is better for this message to be at the top of the result output, rather than at the bottom of it? doesn't it require the user to scroll down if he has a long output?

zjffdu: hmm, make sense to put it at the top

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
zjffdu: Good idea !

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
zjffdu: This is done by intellij which would use * when imports more than 5 classes in one package. I think it should be fine to do that.

zjffdu:
@felixcheung I have addressed the comments and also fix one bug in LivySparkSQLInterpreter, and update the screenshot of testing in PR description.

prabhjyotsingh:
Tested locally, looks good.

livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
```diff
@@ -113,37 +113,44 @@ public InterpreterResult interpret(String line, InterpreterContext context) {
...
+            StringBuilder resMsg = new StringBuilder();
+            String[] rows = message.getData().split("\n");
+            String[] headers = rows[1].split("\\|");
+            for (int head = 1; head < headers.length; head++) {
+              resMsg.append(headers[head].trim()).append("\t");
             }
             resMsg.append("\n");
+            if (rows[3].indexOf("+") == 0) {
+
+            } else {
```
felixcheung: this exists before, but this construct is a bit strange?
```
if (rows[3].indexOf("+") == 0) {
 
 } else {
```

felixcheung:
LGTM.
what is the bug in LivySparkSQLInterpreter - should we have a JIRA for it?

livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
zjffdu: I don't know about this logic, @prabhjyotsingh is the original author who might know this.

zjffdu:
The bug in LivySparkSQLInterpreter is caused by this PR. LivySparkSQLInterpreter delegate LivySparkInterpreter to generate InterpreterResult and then convert it to table type. So here since we make it as multiple InterpreterResult, we should only convert the InterpreterMessage with text type that is for sql output.

zjffdu:
ping @felixcheung any more comments ?

felixcheung:
nope, 
merging if no more comment

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -259,7 +259,8 @@ private InterpreterResult appendSessionExpire(InterpreterResult result, boolean
     if (sessionExpired) {
       InterpreterResult result2 = new InterpreterResult(result.code());
       result2.add(InterpreterResult.Type.HTML,
-          "<font color=\"red\">Previous session is expired, new session is created.</font>");
+          "<font color=\"red\">Previous session is expired, new session is created. " +
+              "Paragraphes that depend on this paragraph need to be re-executed !" + "</font>");
```
felixcheung: I think it should be `Paragraphs`. Also, let remove the space before `!`

zjffdu: Thanks, addressed.

felixcheung:
merging if no more comment

zjffdu:
@felixcheung @minahlee Could any of you help to merge it into branch-0.7 as well ? Since this is only livy interpreter specific, so I don't think it would affect the 0.7 release.

zjffdu:
@felixcheung @minahlee , Sorry my mistake, I has been merged into branch-0.7. @minahlee I notice you make the jira fixversion as 0.8, I add 0.7 as well.

PR #1862, opened 2017-01-06
[HOTFIX] Removed non-string properties values since it's not used

jongyoul:
### What is this PR for?
Removing an error when users try to connect running interpreter. Zeppelin Server passes those properties into RemoteInterpreter when it starts with RemoteInterpreter. But it supports only Map<String, String> type for now, thus port information will make an error and also those information were not used in any code

### What type of PR is it?
[Hot Fix]

### Todos
* [ ] - Remove some properties

### What is the Jira issue?
N/A

`Connect to existing process` should work well

N/A

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
@@ -92,7 +92,7 @@ public class RemoteInterpreter extends Interpreter {
    * Connect to existing process
    */
   public RemoteInterpreter(Properties property, String sessionKey, String className, String host,
-      int port, int connectTimeout, int maxPoolSize,
+      int port, String localRepoPath, int connectTimeout, int maxPoolSize,
       RemoteInterpreterProcessListener remoteInterpreterProcessListener,
       ApplicationEventListener appListener, String userName, Boolean isUserImpersonate) {
     super(property);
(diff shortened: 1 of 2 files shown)


jongyoul:
@cloverhearts Can you look into it? I think it can solve the problem.

felixcheung:
LGTM
should there be tests for this?

cloverhearts:
@jongyoul 
Yes, I was take look!
LGTM.

cloverhearts:
Hello @felixcheung 
I have created and tested a remote interpreter.
1. Go to zeppelin/bin.
2. Run the following:
    `bin/bash ./interpreter.sh -d ../interpreter/python -p 20000> / dev / null`
3. Create a python interpreter group on zeppelin web. (For example, "mypython")
4. Click the edit button.
5. Enter `host: localhost`,` port: 20000` as the remote access interpreter address.
6. Save and run the following in the paragraph:
`` `
%mypython
print ("helloworld")
`` `

If everything works properly, everything is working fine.
This did not work properly in the previous Master.

felixcheung:
Great! I guess I was asking if it's possible to have automated tests so it wouldn't break in the future.

jongyoul:
It's good idea to make it but, for now, we cannot do it because we have zombie process problem when we test it. I think we can do it after we solve that former problem.

astroshim:
Tested and LGTM!
Thanks @jongyoul

jongyoul:
@cloverhearts Can you make a new PR of writing test cases?

cloverhearts:
@jongyoul Sure :)

PR #1870, opened 2017-01-08
[ZEPPELIN-1924] Fix "this._isNumeric is not a function" error 

AhyoungRyu:
### What is this PR for?
After #1815 was merged, 
```
uncaught TypeError: this._isNumeric is not a function
    at ColumnSettings._numericValidator [as validator] (handsonHelper.js:172)
    at handsontable.js:5181
```
is shown when click "Numeric" in the result table like below. 

Since ES6 no longer supports autobind for `this`, seems it needs to be bound in the constructor.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1924](https://issues.apache.org/jira/browse/ZEPPELIN-1924)

**To reproduce**
In master, go to Spark tutorial note and click 
screen shot 2017-01-08 at 8 32 05 pm 
`Numeric`. Then the error msg will be shown in browser dev console. 

With this patch, this error msg won't be shown up anymore :)

### Questions:

### DIFF
--- a/zeppelin-web/src/app/handsontable/handsonHelper.js
+++ b/zeppelin-web/src/app/handsontable/handsonHelper.js
@@ -20,6 +20,7 @@ export default class HandsonHelper {
     this.columns = columns || [];
     this.rows = rows || [];
     this.comment = comment || '';
+    this._numericValidator = this._numericValidator.bind(this);
   };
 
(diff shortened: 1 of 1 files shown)


minahlee:
Tested, LGTM

PR #1872, opened 2017-01-08
[ZEPPELIN-1922] Exclude jackson-core and jackson-databind from aws dep

epahomov:
### What is this PR for?
Fix jar hell around jackson. aws-java-sdk-s3 braings 2.5.3

### What type of PR is it?
Bug Fix 

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1922

### How should this be tested?
Just use docker conrainer https://github.com/epahomov/docker-zeppelin/blob/master/Dockerfile  against a yarn cluster. Now it's not working becuase of this issue.

### Questions:

PR #1873, opened 2017-01-08
[ZEPPELIN-1922] Exclude jackson-core and jackson-databind from aws dep

epahomov:
### What is this PR for?
Master version of https://github.com/apache/zeppelin/pull/1872

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1922

### DIFF
--- a/zeppelin-zengine/pom.xml
+++ b/zeppelin-zengine/pom.xml
@@ -115,6 +115,16 @@
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-s3</artifactId>
       <version>${aws.sdk.s3.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-core</artifactId>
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
@epahomov Thanks for the contribution.

I'm trying to understand what cause this problem, because `zeppelin-zengine` shouldn't be a dependency of any interpreter and `/lib` directory supposed to not be included in the classpath of interpreter process. 

So excluding some library from `zeppelin-zengine` or removing jar from `lib/` directory shouldn't affect anything about classloading in interpreter.

@epahomov, could you elaborate how interpreter classpath includes dependencies of `zeppelin-zengine` module?

cc @jongyoul

epahomov:
@Leemoonsoo , honestly speaking, I don't know. But I've tested this fix in very isolated env(docker) and it's fixes it. I'm not saying that it's strong argument to fix issue this way, but may be it can give you some clue

felixcheung:
I guess the question is by excluding this would it break something down the line?

PR #1881, opened 2017-01-10
ZEPPELIN-1935 Add jceks stored password support for jdbc interpreter

r-kamath:
### What is this PR for?
Add support for jceks stored password instead of using password in clear text. This security enhancement prevents any user from reading clear passwords from interpreter json using shell/py/spark etc. #1315 is the parent PR which fixed similar a issue in Active Directory (shiro.ini config).

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-1935

Create a keystore file using the hadoop credential commandline, for this the hadoop commons should be in the classpath.

`hadoop credential create jdbc.password -provider jceks://file/user/zeppelin/conf/zeppelin.jceks`

Use the jceks file and the key to configure jdbc interpreter. Example interpreter setting:

#### Following are the supported settings for backward compatibility

- Settings with username and jceks
```
default.driver	org.postgresql.Driver
default.jceks.credentialKey	jdbc.password
default.jceks.file	jceks://file/tmp/zeppelin.jceks
default.url	jdbc:postgresql://rkamath-local-1:5432/
default.user	rk-user
```

- Settings with user name and clear text password
```
default.driver	org.postgresql.Driver
default.url	jdbc:postgresql://rkamath-local-1:5432/
default.user	rk-user
default.password	password1
```

screen shot 2017-01-10 at 7 02 12 pm

### Questions:


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -399,6 +409,34 @@ public Connection run() throws Exception {
...
+            provider.getCredentialEntry(properties.getProperty(JDBC_JCEKS_CREDENTIAL_KEY));
+        if (credEntry != null) {
+          return new String(credEntry.getCredential());
+        } else {
+          throw new InterpreterException("Failed to retrieve password from JCEKS from key: "
+              + properties.getProperty(JDBC_JCEKS_CREDENTIAL_KEY));
+        }
+      } catch (Exception e) {
+        logger.error("Failed to retrieve password from JCEKS", e);
+        logger.error("For file: " + properties.getProperty(JDBC_JCEKS_FILE));
```
felixcheung: Let's do one logger.error?

felixcheung:
this is important - is there any backward compatibility concerns with changing the password storage?

r-kamath:
@felixcheung JCEKS is used only when the password field is empty. When password and JCEKS properties are empty the connection attempt is made without password.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
r-kamath: 👍

prabhjyotsingh:
@r-kamath IMO this shouldn't be called "backward compatibility" as these are two ways in which password can provided to interpreter.

Tested locally, works as expected, LGTM!

felixcheung:
Got it.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
r-kamath: e62088a

prabhjyotsingh:

PR #1885, opened 2017-01-11
ZEPPELIN-1933. Set pig job name and allow to set pig property in pig interpreter setting

zjffdu:
### What is this PR for?
Two improvements for pig interpreter.
* Set job name via paragraph title if it exists, otherwise use the last line of pig script
* Allow to set any pig property in interpreter setting

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1933

Unit tested and manually tested. 

image

### Questions:

### DIFF
--- a/pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
+++ b/pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.pig;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.pig.PigServer;
 import org.apache.pig.backend.BackendException;
@@ -97,4 +98,21 @@ public abstract class BasePigInterpreter extends Interpreter {
   }
 
   public abstract PigServer getPigServer();
+
+  /**
+   * Use paragraph title if it exists, else use the last line of pig script.
+   * @param cmd
+   * @param context
+   * @return
+   */
+  protected String createJobName(String cmd, InterpreterContext context) {
+    String pTitle = context.getParagraphTitle();
+    if (!StringUtils.isEmpty(pTitle)) {
+      return pTitle;
+    } else {
+      // use the last line of pig script as the job name.
+      String[] lines = cmd.split("\n");
+      return lines[lines.length - 1];
+    }
+  }
 }
--- a/pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
+++ b/pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
@@ -107,12 +107,18 @@ public abstract class BasePigInterpreter extends Interpreter {
    */
   protected String createJobName(String cmd, InterpreterContext context) {
     String pTitle = context.getParagraphTitle();
-    if (!StringUtils.isEmpty(pTitle)) {
+    if (!StringUtils.isBlank(pTitle)) {
       return pTitle;
     } else {
-      // use the last line of pig script as the job name.
+      // use the last non-empty line of pig script as the job name.
       String[] lines = cmd.split("\n");
-      return lines[lines.length - 1];
+      for (int i = lines.length - 1; i >= 0; --i) {
+        if (!StringUtils.isBlank(lines[i])) {
+          return lines[i];
+        }
+      }
+      // in case all the lines are empty, but usually it is almost impossible
+      return "empty_job";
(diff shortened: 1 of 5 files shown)


zjffdu:
@felixcheung Please help review.

zjffdu:
CI failure is not relevant.

```
Failed tests: 
```

pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
```diff
@@ -97,4 +98,21 @@ public Scheduler getScheduler() {
...
+   * @return
+   */
+  protected String createJobName(String cmd, InterpreterContext context) {
+    String pTitle = context.getParagraphTitle();
+    if (!StringUtils.isEmpty(pTitle)) {
+      return pTitle;
+    } else {
+      // use the last line of pig script as the job name.
+      String[] lines = cmd.split("\n");
+      return lines[lines.length - 1];
```
felixcheung: is it possible to end up with some unexpected characters, eg.
```
STORE same1 INTO '/tmp/tutorial-join-results' USING PigStorage();
```
like `/` or `()` or `;`?

felixcheung: check if `cmd` is empty?

pig/src/main/java/org/apache/zeppelin/pig/PigInterpreter.java
```diff
@@ -58,6 +59,12 @@ public void open() {
     }
     try {
       pigServer = new PigServer(execType);
+      for (Map.Entry entry : getProperty().entrySet()) {
+        if (!entry.getKey().toString().startsWith("zeppelin.")) {
+          pigServer.getPigContext().getProperties().setProperty(entry.getKey().toString(),
+              entry.getValue().toString());
```
felixcheung: As far as I recall, this is actually somewhat opposite to how we pass config to other interpreters.
Generally, we say, if it's spark, all property starting with "spark." we will pass through.
Here we are say if it's _not_ "pig." pass through

felixcheung: is it better to set a variable as `pigServer.getPigContext().getProperties()` to reuse?

pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
```diff
@@ -97,4 +98,21 @@ public Scheduler getScheduler() {
...
+
+  /**
+   * Use paragraph title if it exists, else use the last line of pig script.
+   * @param cmd
+   * @param context
+   * @return
+   */
+  protected String createJobName(String cmd, InterpreterContext context) {
+    String pTitle = context.getParagraphTitle();
+    if (!StringUtils.isEmpty(pTitle)) {
```
felixcheung: !isBlank instead?

pig/src/main/java/org/apache/zeppelin/pig/PigInterpreter.java
zjffdu: There's 2 types of property for zeppelin interpreter. Type-1 is used by interpreter itself, Type-2 is used by the engine underneath (like spark engine). All spark configuration start with `spark.`, this is just spark's convention, not zeppelin's convention. zeppelin's convention is to use `zeppelin` as the prefix for Type-1 property, and keep Type-2 property as what it is in the underlying engine. So here I keep the pig property as what it is in pig, otherwise it would confuse users IMO.

pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
zjffdu: correct, !isBlank is better.

pig/src/main/java/org/apache/zeppelin/pig/BasePigInterpreter.java
zjffdu: These characters should be valid for job name. And it make sense to check whether the line is blank. I will use the last non-empty line instead.

pig/src/main/java/org/apache/zeppelin/pig/PigInterpreter.java
felixcheung: I suppose it's a bit inconsistent all around then?
I'm ok with this

felixcheung:
LGTM
merging if no more comment

PR #1888, opened 2017-01-12
[ZEPPELIN-1941] Fix deadlock in cron job

OopsOutOfMemory:
### What is this PR for?
There is a deadlock in concurrent cron job execution with release resource option.
`Scenario`:
Two notebook run with cron job that release resource after job finished.

In `Notebook.CronJob.execute()` method:

`T1. note.runAll(); // locked paragraphs(lock) and wait to interpreterSettings(lock)`

`T2. notebook.getInterpreterFactory().restart() //locked(interpreterSettings) and wait for paragraphs(lock) during jobAbort.`

This will trigger a deadlock that cause zeppelin hang.

### What type of PR is it?
[Bug Fix ]

### Todos
* [x] - Fix this by remove locking interpreterSettings in restart method

### What is the Jira issue?

### Questions:


OopsOutOfMemory:
will use another pr.

PR #1892, opened 2017-01-12
[MINOR] broken logo url in web dev mode

soralee:
### What is this PR for?
zeppelin logo image (zepLogo.png) is broken in `web dev` mode like below screenshot.
screenshot from 2017-01-12 14-50-35

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1943

1. Run front-end application in [dev mode](https://github.com/apache/zeppelin/tree/master/zeppelin-web)
2. Open console in web developer mode.
3. Connect/refresh localhost:9000 
4. Check to be shown Logo image in Zeppelin home.
image
5. Check errors in console.

screenshot from 2017-01-12 16-58-40

### Questions:

### DIFF
--- a/zeppelin-web/webpack.config.js
+++ b/zeppelin-web/webpack.config.js
@@ -163,8 +163,7 @@ module.exports = function makeWebpackConfig () {
       // Pass along the updated reference to your code
       // You can add here any file extension you want to get copied to your output
       test: /\.(png|jpg|jpeg|gif|svg|woff|woff2|ttf|eot)$/,
-      //loader: 'file'
-      loader: 'file-loader?name=[name].[ext]&outputPath=styles/images/'
+      loader: 'file'
     }, {
       // HTML LOADER
       // Reference: https://github.com/webpack/raw-loader
--- a/zeppelin-web/webpack.config.js
+++ b/zeppelin-web/webpack.config.js
@@ -163,7 +163,8 @@ module.exports = function makeWebpackConfig () {
       // Pass along the updated reference to your code
       // You can add here any file extension you want to get copied to your output
       test: /\.(png|jpg|jpeg|gif|svg|woff|woff2|ttf|eot)$/,
-      loader: 'file'
+      //loader: 'file'
+      loader: 'file-loader?name=[name].[ext]&outputPath=styles/images/'
     }, {
       // HTML LOADER
       // Reference: https://github.com/webpack/raw-loader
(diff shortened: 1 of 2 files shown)


minahlee:
This fix will bring same issue in production mode if user set `ZEPPELIN_SERVER_CONTEXT_PATH`. Do you have any other idea to work around?

soralee:
Thanks for review and considering the side effect @minahlee :+1:  
You're right! I missed this option.

As your mention, I checked `ZEPPELIN_SERVER_CONTEXT_PATH` option.

**Test 1**. When I set `ZEPPELIN_SERVER_CONTEXT_PATH` to `home` and connect `web dev mode (localhost:9000/home)`, it is okay.
z_icon_9000

**Test 2**. But, when I set `ZEPPELIN_SERVER_CONTEXT_PATH` to `home` and connect localhost:8080/home`, there is a problem.   
z_icon_8080

I'll update it and inform to you after resolving this issue.

Leemoonsoo:
cc @1ambda

1ambda:
Sorry for late reply. I will review and comment ASAP @soralee

zeppelin-web/webpack.config.js
```diff
@@ -267,9 +267,13 @@ module.exports = function makeWebpackConfig () {
     progress: true,
     contentBase: './src',
     setup: function(app) {
-      app.use(
-        '/bower_components/',
-        require('express').static(path.join(__dirname, './bower_components/')));
+      app.use('**/bower_components/',
+        require('express').static(path.resolve(__dirname, './bower_components/')));
+      app.use('**/app/', require('express').static(path.resolve(__dirname, './src/app/')));
```
1ambda: I think we don't need `app`, `components`, `fonts` in static. since they are already served using relative paths.

1ambda: But if we use nested context paths, it cause another problems like the screenshot below. So i believe **having correct directory structure for static files** is the solution to solve these kind of errors. However, it can be handed in other PRs. 

image

zeppelin-web/webpack.config.js
```diff
@@ -267,9 +267,13 @@ module.exports = function makeWebpackConfig () {
     progress: true,
     contentBase: './src',
     setup: function(app) {
-      app.use(
-        '/bower_components/',
-        require('express').static(path.join(__dirname, './bower_components/')));
+      app.use('**/bower_components/',
+        require('express').static(path.resolve(__dirname, './bower_components/')));
```
soralee: Thanks @1ambda for review! Yes, it makes sense! Let me updated this part ASAP :)

zeppelin-web/webpack.config.js
soralee: Actually I missed this issue in production mode. 
Yes! I totally agree with @1ambda. I think we need reorganize directory to resolve kind of this issue in correct way.

soralee:
@1ambda I updated to extract `require('express')` module as variable and CI is green :)

1ambda:
LGTM!

PR #1893, opened 2017-01-12
[ZEPPELIN-451] Save codes and messages as multi-line

sixmen:
### What is this PR for?
Save Paragraph texts and result msgs as array of strings instead of one long string.
One long strings are not VCS-friendly.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-451

Modify notebooks and check saved note files

### Questions:
* Is there breaking changes for older versions? Notebooks from new zeppelin can not be imported in old zeppelin.


Leemoonsoo:
@sixmen Thanks for the contribution. 
I tested this branch and it works well.

While this makes incompatible changes (previous version of Zeppelin can't load the notebook), i'd like to ask more people review this PR, will there be any type of side effects.

soralee:
Thank @sixmen for improving this. It's very good feature!

I think sudden changes, part of importing notebook, will make users uncomfortable. 
In my opinion, both old version notebook and now version notebook should be nice if it is compatible.

sixmen:
@soralee You mean, it is good if users of old Zeppelin can import new version notebooks?

I think forward compatibility is unnecessary, but if you tell me a solution that you want, I'll try to implement it.

PR #1894, opened 2017-01-12
[ZEPPELIN-1560] avoid generating minus sign in package name

robbinspg:
### What is this PR for?
using Object.hashCode() as part of the REPL wrapper class name can cause a compilation error as hashCode can validly return a negative integer.

We would like this fix backported to 0.6 and later streams.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1560

regression tests with openJdk + tests with IBM jvm

N/A

### Questions:


Leemoonsoo:
@robbinspg Great fix!
CI failure looks not related.

LGTM and merge to master, branch-0.7, branch-0.6 if no more comments!

felixcheung:
ouch! thanks for fixing this. LGTM.

PR #1897, opened 2017-01-13
Can't remove interpreter repository.

astroshim:
### What is this PR for?
It can't be delete the `interpreter repository`(http://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/manual/dependencymanagement.html) when name has include special characters.

### What type of PR is it?
Bug Fix 

please refer to screenshot.

- before

- after
image

### Questions: 

### DIFF
--- a/zeppelin-web/src/components/repository-create/repository-dialog.html
+++ b/zeppelin-web/src/components/repository-create/repository-dialog.html
@@ -28,7 +28,7 @@ limitations under the License.
             <div class="form-group">
               <label class="control-label col-sm-2" for="repoId">ID</label>
               <div class="col-sm-10">
-                <input type="text" class="form-control" id="repoId" ng-model="newRepoSetting.id"
+                <input type="text" pattern="^[a-zA-Z0-9]*$" class="form-control" id="repoId" ng-model="newRepoSetting.id"
                        placeholder="Repository id" required />
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/components/repository-create/repository-dialog.html
```diff
@@ -28,7 +28,7 @@ <h4 class="modal-title">Add New Repository</h4>
             <div class="form-group">
               <label class="control-label col-sm-2" for="repoId">ID</label>
               <div class="col-sm-10">
-                <input type="text" class="form-control" id="repoId" ng-model="newRepoSetting.id"
+                <input type="text" pattern="^[a-zA-Z0-9]*$" class="form-control" id="repoId" ng-model="newRepoSetting.id"
```
felixcheung: would it be too restrictive though? does people want repo with non English name?

astroshim: @felixcheung Thank you for reviewing! 
`repository id` will be the path of the RestAPI URL so how about adding `=-.-:` characters?

felixcheung: sure? I don't know the constraints we are working with here, so...

Leemoonsoo: If it's because of https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java#L274 `repository id` is path of the restAPI, other approach would be encode / decode `reponsitory id` before and after call REST API. Using `encodeURI(uri)` in js and `URLDecoder` in java

prabhjyotsingh: Yes, agreed with @Leemoonsoo we should not have constraints.

PR #1911, opened 2017-01-18
[ZEPPELIN-1972] Preserve context classloader

DrIgor:
### What is this PR for?
`IMain.interpret` changes current thread's context classloader. It may cause different issues and definitely is the reason of [ZEPPELIN-1738](https://issues.apache.org/jira/browse/ZEPPELIN-1738) test failures.

It's a known scala bug. See https://issues.scala-lang.org/browse/SI-9587

As a workaround we need to save and restore context classloader manually

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-1972](https://issues.apache.org/jira/browse/ZEPPELIN-1972)

Run ignite interpreter test and ignite sql interpreter test in the same thread
```
mvn test -pl ignite -am -Pscala-2.11 -Dtest=org.apache.zeppelin.ignite.IgniteInterpreterTest#testInterpret,org.apache.zeppelin.ignite.IgniteSqlInterpreterTest#testSql -DfailIfNoTests=false
```

### Questions:

### DIFF
--- a/ignite/src/main/java/org/apache/zeppelin/ignite/IgniteInterpreter.java
+++ b/ignite/src/main/java/org/apache/zeppelin/ignite/IgniteInterpreter.java
@@ -200,15 +200,20 @@ public class IgniteInterpreter extends Interpreter {
   }
 
   private void initIgnite() {
-    imain.interpret("@transient var _binder = new java.util.HashMap[String, Object]()");
-    Map<String, Object> binder = (Map<String, Object>) getLastObject();
+    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+    try {
+      imain.interpret("@transient var _binder = new java.util.HashMap[String, Object]()");
+      Map<String, Object> binder = (Map<String, Object>) getLastObject();
 
-    if (getIgnite() != null) {
-      binder.put("ignite", ignite);
+      if (getIgnite() != null) {
+        binder.put("ignite", ignite);
 
-      imain.interpret("@transient val ignite = "
-              + "_binder.get(\"ignite\")"
-              + ".asInstanceOf[org.apache.ignite.Ignite]");
+        imain.interpret("@transient val ignite = "
+            + "_binder.get(\"ignite\")"
+            + ".asInstanceOf[org.apache.ignite.Ignite]");
+      }
+    } finally {
+      Thread.currentThread().setContextClassLoader(contextClassLoader);
     }
   }
 
@@ -279,11 +284,14 @@ public class IgniteInterpreter extends Interpreter {
         }
       }
 
+      ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
       try {
         code = getResultCode(imain.interpret(incomplete + s));
       } catch (Exception e) {
         logger.info("Interpreter exception", e);
         return new InterpreterResult(Code.ERROR, InterpreterUtils.getMostRelevantMessage(e));
+      } finally {
+        Thread.currentThread().setContextClassLoader(contextClassLoader);
       }
 
       if (code == Code.ERROR) {


jongyoul:
LGMT. I have a question. Is there any related issue to SparkInterpreter? Can you take a look into it?

DrIgor:
@jongyoul I checked SparkInterpreter, it has the same issue with context class loader

Leemoonsoo:
Thanks @DrIgor for fix and explanation of the issue!

When RemoteInterpreterServer creates Interpreter, it [wraps interpreter instance with LazyOpenInterpreter](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java#L187). If we wrap interpreter instance with [ClassLoaderInterpreter](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/ClassloaderInterpreter.java) as well, like [zeppelin do when creating interpreter locally](https://github.com/apache/zeppelin/blob/ae1cb0527bc223b25761e1370618929e228183f8/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java#L1123), then every interpreter will benefit by ClassloaderInterpreter. ClassloaderInterpreter restores classloader after methods call.

@DrIgor @jongyoul Shell we try make RemoteInterpreterServer wrap interpreter with ClassloaderInterpreter? what do you think?

jongyoul:
@Leemoonsoo I think to wrap it with `ClassloaderInterpreter` would be a good idea. @DrIgor Can you try it on?

DrIgor:
@Leemoonsoo @jongyoul thank you! I didn't know about ClassLoaderInterpreter

Unfortunately, our tests don't use InterpreterFactory and will use unwrapped Interpreter instance. So It won't fix flaky test.

Besides, code inside interpreter itself will use incorrect classloader. Currently it isn't an issue but someday...

Maybe It's better to wrap IMain or create "scala utils" to share common code between several scala-speaking interpreters?

jongyoul:
It actually wrap and conserve classloader for interpreter. I think you just change this line https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java#L192 to use `ClassLoaderInterpreter`. Can you try it?

jongyoul:
Concerning `ClassloaderInterpreter`, you can find it https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/ClassloaderInterpreter.java

DrIgor:
Yes, I tried ClassloaderInterpreter but it didn't fix original issue.

https://github.com/apache/zeppelin/blob/master/ignite/src/test/java/org/apache/zeppelin/ignite/IgniteInterpreterTest.java#L69
Here test creates unwrapped interpreter instance.

So we need wrap every IgniteInterpreter usage. Besides, wrapper restores classloader _after_ interpreter call and interpreter itself use broken classloader.

jongyoul:
Got it. How about wrapping it like `new ClassloaderInterpreter(new IgniteInterpreter(props))`? It's the same logic that RemoteInterpreter want to work. You should change this line https://github.com/apache/zeppelin/blob/master/ignite/src/test/java/org/apache/zeppelin/ignite/IgniteInterpreterTest.java#L45 to `Interpreter`

jongyoul:
Is there any update on it?

DrIgor:
I tried to wrap with `ClassloaderInterpreter` and it works

But I don't like this solution. We force every user of `IgniteInterpreter` to wrap it with `ClassloaderInterpreter`. It's very easy to broke this test again. For example, to add new test and forget about wrapping.

It looks like dirty fix for me. `IgniteInterpreter` is still broker and we try to reduce negative effects by wrapper. I think it is acceptable for third party libraries when we can't change sources. But if we can fix original code we should do it

If you think that wrapper is more appropriate I will update PR

jongyoul:
@DrIgor I agree with you. It's very easily broken. But on the other hands, I will make a new PR for changing `ClassloaderInterpreter` because another interpreter will fails with same issue if it doesn't implement like you.

LGTM, will merge it.

PR #1914, opened 2017-01-19
[ZEPPELIN-1983] Feedback about success/failure when interpreter is re…

beriaanirudh:
### What is this PR for?
When a interpreter is restarted, there is no feedback to user whether the action is received/honoured by zeppelin. This might also lead them to click restart multiple times, before they realize it has taken affect.
Also fixed a null bug.

### What type of PR is it?
[Improvement + Bug fix]

### Todos
* None

### What is the Jira issue?
ZEPPELIN-1983

Restart a interpreter to test the positive case. The message should disappers on its own after sometime if not closed. A simple way to test the negative case is to stop the server and then try restarting.

#### Positive case
positive case

#### Negative case
negative case

### Questions:
-No
-No
-No


jongyoul:
LGTM

beriaanirudh:
Thanks for the review @jongyoul . I looked at the CI failure and it looks irrelevant to this change.

jongyoul:
Will merge it into master and branch-0.7

PR #1917, opened 2017-01-20
[ZEPPELIN-1988] precode execution in JDBCInterpreter

dwhsys:
### What is this PR for?
PR adds new parameter in JDBC interpreter which allows to execute some SQL while opening connection.

Parameter: default.precode
Value examples:
"select 1;"
"set application_name=zeppelin;"

If parameter is null then precode is ignored.

### What type of PR is it?
Feature

### Todos
* [ ] - ZEPPELIN-1988

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1988

* Use %jdbc interpreter with DB user with create schema grants
* Set %jdbc interpreter property name / value to "default.precode" / "create schema zeppelin1988;"
* Check if schema has been created

### Questions:

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -31,7 +31,7 @@
   <packaging>jar</packaging>
   <version>0.7.0-SNAPSHOT</version>
   <name>Zeppelin: JDBC interpreter</name>
-  
+
   <properties>
     <!--library versions-->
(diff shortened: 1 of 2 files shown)


astroshim:
Thank you for contribution!.
please rebase PR and  
If you let us know about testing way on detail, it's easy to review.

PR #1919, opened 2017-01-20
[ZEPPELIN-1986] Fix flaky test: Increase WelcomePageSuite timeout

Leemoonsoo:
### What is this PR for?
Try fix problem described in https://issues.apache.org/jira/browse/ZEPPELIN-1986.
Almost all recent build has this failure on CI build.

### What type of PR is it?
Hot Fix

### Todos
* [x] - Increase timeout

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1986

### Questions:


Leemoonsoo:
CI failures are not related with this fix.
Merge it as a Hotfix.

jongyoul:
@Leemoonsoo I think to prolong timeout would be realistic solution but same error may occurs when travis is under resources starvation. Can we handle it?

Leemoonsoo:
@jongyoul I don't have any good idea except for change test to use less resources when travis is under resource starvation. Do you have any good idea?

jongyoul:
@Leemoonsoo Actually, in this case, I also don't have any better idea than what you do :-) Just worried about becoming more rare flaky test.

PR #1922, opened 2017-01-20
[ZEPPELIN-1991] Can't get the PARAGRAPH_APPEND_OUTPUT from the Interpreter.

astroshim:
### What is this PR for?
This PR fixes the problem of streaming events(PARAGRAPH_APPEND_OUTPUT).
It's because of the queue was not thread safe.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1991

1. make a spark paragraph

2. run following code on spark paragraph.
```
%spark
(1 to 5).foreach{ i=>
    Thread.sleep(1000)
    println("Hello " + i)
}

```
3. make a python paragraph

4. run following code on python paragraph.
```
%python.python

print("hi1")
```

5. retry run step 2, and check if streaming working.

- before

- after

### Questions:


AhyoungRyu:
Thanks for the patch! Tested and works well as you described.

AhyoungRyu:
This patch fixes the issue that I reported in [this comment in #1833 ](https://github.com/apache/zeppelin/pull/1833#issuecomment-270078954). Looks good. LGTM

AhyoungRyu:
Merging into master and branch-0.7 if there are no comments on this.

PR #1927, opened 2017-01-21
[ZEPPELIN-1994] bugfix of streaming output.

astroshim:
### What is this PR for?
If you run the following code, then streaming output doesn't work properly from the second run.
```
%spark.pyspark
import time
print("1")
time.sleep(2)
print("2")
time.sleep(2)
print("3")
time.sleep(2)
print("4")
```
This problem comes from the order of `paragraph update` event timing and `paragraph update-append` event timing is incorrect.
and This PR will fix also https://github.com/apache/zeppelin/pull/1833 too.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1994

- run several times pyspark interpreter with above code.

- before

- after

### Questions:


AhyoungRyu:
I tested #1833 upon this patch and #1922, it's perfectly working now. Appreciate for fixing this!

Leemoonsoo:
Great fix. LGTM.
Merge to master and branch-0.7

PR #1930, opened 2017-01-21
[ZEPPELIN-1886] implementation z.getZeppelinJobStatus 

cloverhearts:
### What is this PR for?
You can get the state of a paragraph through ZeppelinContext.
This allows you to implement code according to the paragraph condition.

related
https://github.com/apache/zeppelin/pull/1799

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1886

1. run paragraph.
2. Record the noteid and paragraph id of the paragraph you have run.
3. You can use the following code:
for example
```
%spark
z.getJobStatus("noteID", "paragraphId").name()
z.getJobStatus("2C8D89Y9N", "20170121-042805_156018837").isFinished()
z.getJobStatus("2C8D89Y9N", "20170121-042805_156018837").isAbort()
z.getJobStatus("2C8D89Y9N", "20170121-042805_156018837").isError()
z.getJobStatus("2C8D89Y9N", "20170121-042805_156018837").isPending()
z.getJobStatus("2C8D89Y9N", "20170121-042805_156018837").isRunning()
```

zeppelinstatus

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
@@ -395,7 +395,7 @@ public class ZeppelinSparkClusterTest extends AbstractTestRestApi {
         Map config1 = p1.getConfig();
         config1.put("enabled", true);
         p1.setConfig(config1);
-        p1.setText("%spark var status = z.getJobStatus(\"" + note.getId() + "\",\"" + p0.getId() + "\").name()\n println(status)");
+        p1.setText("%spark var status = z.getJobStatus(\"" + note.getId() + "\",\"" + p0.getId() + "\").name())\n println(status)");
         p1.setAuthenticationInfo(anonymous);
 
         note.runAll();


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/RemoteZeppelinJobStatus.java
```diff
@@ -0,0 +1,115 @@
...
+
+  public boolean isReady() {
+    return getJobStatus().isReady();
+  }
+
+  public boolean isRunning() {
+    return getJobStatus().isRunning();
+  }
+
+  public String getStatus() {
```
Leemoonsoo: nit. how about change method name to `name()`? 

Then api will be look like, 

before
```
z. getZeppelinJobStatus().getStatus()
```

after
```
z. getZeppelinJobStatus().name()
```

which avoid repeating `Status` and more clear while it's returning name of status instead of returning Status type.

spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
```diff
@@ -364,6 +376,31 @@ public void runNote(String noteId, InterpreterContext context) {
...
   }
 
+  /**
+   * get job status by zeppelin note id and paragraph id
+   * @param noteId
+   * @param paragraphId
+   * @return
+   */
+  @ZeppelinApi
+  public RemoteZeppelinJobStatus getZeppelinJobStatus(String noteId, String paragraphId) {
```
Leemoonsoo: How about remove 'Zeppelin' from method name? so using this api will looks like

before
```
z.getZeppelinJobStatus()
```
after
```
z.getJobStatus()
```

So we API becomes more consistent along with other existing APIs, they don't include 'Zeppelin' in method name, i.e.

```
z.run()
z.runNote()
z.show()
z.getJobStatus()
```

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/RemoteZeppelinJobStatus.java
cloverhearts: @Leemoonsoo 
:)
Thank you for your good idea.
Yes, I will.

spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
cloverhearts: @Leemoonsoo Okay, I will fix it

Leemoonsoo:
Could you rebase/merge master and see if CI passes?

This branch has a lot of CI build failure which is fixed by https://github.com/apache/zeppelin/pull/1939

PR #1937, opened 2017-01-24
Search for interpreter-settings in several jars in folder

DrIgor:
### What is this PR for?
`InterpreterFactory` searches for single `interpreter-settings.json` in some jar in folder. This PR allows us to have `interpreter-settings.json` in several jar in one interpreter folder.

For example, we can have `zeppelin-spark.jar` and `zeppelin-zrinterpreter.jar` in interpreter/spark folder, both with interpreter-settings.json. Zeppelin will load R settings only when `zeppelin-zrinterpreter.jar` exists.

### What type of PR is it?
[Improvement]

### What is the Jira issue?

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -311,16 +312,19 @@ public class InterpreterFactory implements InterpreterGroupFactory {
     URL[] urls = recursiveBuildLibList(new File(interpreterDir));
     ClassLoader tempClassLoader = new URLClassLoader(urls, cl);
 
-    InputStream inputStream = tempClassLoader.getResourceAsStream(interpreterJson);
-
-    if (null != inputStream) {
-      logger.debug("Reading {} from resources in {}", interpreterJson, interpreterDir);
-      List<RegisteredInterpreter> registeredInterpreterList =
-          getInterpreterListFromJson(inputStream);
-      registerInterpreters(registeredInterpreterList, interpreterDir);
-      return true;
+    Enumeration<URL> interpreterSettings = tempClassLoader.getResources(interpreterJson);
+    if (!interpreterSettings.hasMoreElements()) {
+      return false;
     }
-    return false;
+    for (URL url : Collections.list(interpreterSettings)) {
+      try (InputStream inputStream = url.openStream()) {
+        logger.debug("Reading {} from {}", interpreterJson, url);
+        List<RegisteredInterpreter> registeredInterpreterList =
+            getInterpreterListFromJson(inputStream);
+        registerInterpreters(registeredInterpreterList, interpreterDir);
+      }
+    }
+    return true;
   }
 
   private boolean registerInterpreterFromPath(String interpreterDir, String interpreterJson)


DrIgor:
@jongyoul PR for loading several interpreter-setting

jongyoul:
@DrIgor Did you test this PR with `-Pr` profile? this PR looks good to me but we need to test it. I think you can test it by running Zeppelin with r profile. How about you?

DrIgor:
@jongyoul currently R interpreter uses static initialization and isn't loaded at all =)

I will create pull request with interpreter-settings.json for R interpreter. But it needs this PR to be merged first

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -37,17 +37,7 @@
...
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
jongyoul: One small fix. Zeppelin doesn't recommend this style of importing libs with asterisk. Can you fix it?

jongyoul:
@DrIgor I know, I just ask you for making R-related PR on top of this PR.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
DrIgor: Fixed

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
```diff
@@ -311,16 +312,19 @@ private boolean registerInterpreterFromResource(ClassLoader cl, String interpret
...
+    if (!interpreterSettings.hasMoreElements()) {
+      return false;
     }
-    return false;
+    for (URL url : Collections.list(interpreterSettings)) {
+      try (InputStream inputStream = url.openStream()) {
+        logger.debug("Reading {} from {}", interpreterJson, url);
+        List<RegisteredInterpreter> registeredInterpreterList =
+            getInterpreterListFromJson(inputStream);
+        registerInterpreters(registeredInterpreterList, interpreterDir);
```
zjffdu: I think it is very hard to control when we allow multiple interpreter-setting.json. It would bring several potential problems. like which interpreter-setting.json will be read first ? Will the later one overwrite the previous one ? All these problems depends on the implementation.  BTW, what is the exact problem of this PR try to fix ? It seems need to create a ticket for it.

DrIgor: Currently we have the same problems. If folder contain several settings InterpreterFactory will load random one. And any interpreter from interpreters folder may override previous properties. PR doesn't change loading behaviour, it only expands possible file locations

I try to fix problem with R interpreter. It needs to be in spark interpreter group and has its own interpreter-settings.json.

DrIgor:
Hello!

Are there any other ideas how we can add R interpreter to spark group without touching spark's interpreter-settings.json?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterFactory.java
jongyoul: I also agree about @zjffdu's comment and @DrIgor's answer. Both of you guys are correct. we already had potential bug and need to fix it. We also think of override issue which is actually occurs now when two different interpreter try to create interpreters with same interpreter. I, personally, agree this PR to solve R issue which makes Zeppelin complicated.

jongyoul:
I'll merge this PR and will make a new PR to handle with Jeff's worries which are reasonable and should be maintained.

PR #1938, opened 2017-01-24
ZEPPELIN-2006. Livy interpreter doesn't work in anonymous mode

zjffdu:
### What is this PR for?
We should not pass proxy-user to livy server in anonymous mode, otherwise the livy session will be started as user anonymous. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2006

Tested manually in anonymous mode.

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
@@ -206,7 +206,8 @@ public abstract class BaseLivyInterprereter extends Interpreter {
           conf.put(entry.getKey().toString().substring(5), entry.getValue().toString());
       }
 
-      CreateSessionRequest request = new CreateSessionRequest(kind, user, conf);
+      CreateSessionRequest request = new CreateSessionRequest(kind,
+          user.equals("anonymous") ? null : user, conf);
       SessionInfo sessionInfo = SessionInfo.fromJson(
           callRestAPI("/sessions", "POST", request.toJson()));
       long start = System.currentTimeMillis();


zjffdu:
@prabhjyotsingh Please help review. CI failure is not relevant. I have created https://issues.apache.org/jira/browse/ZEPPELIN-2007 for it

prabhjyotsingh:
This is for kerberos enabled environment, correct ?

zjffdu:
No, impersonation works in both secured and unsecured enviroment. Without this PR, we still pass proxy-user `anonymous` to livy server. And this cause we launch the spark app via user `anonymous`. But it should the user `livy` in the anonymous mode.

prabhjyotsingh:
Got it, good catch. Test on local, works as expected, LGTM.

khalidhuseynov:
LGTM

prabhjyotsingh:

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -206,7 +206,8 @@ private SessionInfo createSession(String user, String kind)
           conf.put(entry.getKey().toString().substring(5), entry.getValue().toString());
       }
 
-      CreateSessionRequest request = new CreateSessionRequest(kind, user, conf);
+      CreateSessionRequest request = new CreateSessionRequest(kind,
+          user.equals("anonymous") ? null : user, conf);
```
bikassaha: Will this disable the proxyUser parameter to Livy? When anonymous mode is enabled in Zeppelin then the correct action seems to be to not pass proxyUser param since there is no proxy user.

zjffdu: no proxyUser will be passed to livy when it is anonymous

PR #1946, opened 2017-01-25
[ZEPPELIN-921] Apply new mechanism to KnitR and RRepl

DrIgor:
### What is this PR for?
Applies new registration mechanism for R interpreter

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-921

Build zeppelin with R profile enabled and run several R snippets

### Questions:

### DIFF
--- a/r/src/main/java/org/apache/zeppelin/rinterpreter/KnitR.java
+++ b/r/src/main/java/org/apache/zeppelin/rinterpreter/KnitR.java
@@ -34,12 +34,6 @@ import java.util.Properties;
 public class KnitR extends Interpreter implements WrappedInterpreter {
   KnitRInterpreter intp;
 
-  static {
-    Interpreter.register("knitr", "spark", KnitR.class.getName(),
-        RInterpreter.getProps()
-    );
-  }
-
   public KnitR(Properties property, Boolean startSpark) {
     super(property);
     intp = new KnitRInterpreter(property, startSpark);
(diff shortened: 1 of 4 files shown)


DrIgor:
Needs this changes: #1937

jongyoul:
I merged #1937. Can you rebase it onto master and make CI green?

DrIgor:
Rebased on master and ran build.

I tried several times:

- https://travis-ci.org/apache/zeppelin/builds/197341657
- https://travis-ci.org/apache/zeppelin/builds/197539698
- https://travis-ci.org/apache/zeppelin/builds/197547341

CI is still red, but every time it had a different failure. I think they are not caused by my changes

jongyoul:
@DrIgor The last failures looks related to this PR. Can you check it?
 - https://api.travis-ci.org/jobs/197657609/log.txt?deansi=true

DrIgor:
@jongyoul sorry for late response!

I checked failure. It's related to PR but I don't know how to fix it

Build profile uses
```
$ export SCALA_VER="2.11"
$ export SPARK_VER="1.6.3"
```

`downloadSpark.sh` downloads spark prebuilt for scala 2.10. So we have incompatible versions of scala.

I tried to build spark 1.6 with scala 2.11 — R interpreter works fine. But spark does not provide such binary distribution.

Does Zeppelin support spark 1.6 with scala 2.11? Then we should provide binary spark distribution for this build profile. I think it affects not only R interpreter

This code https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/rest/AbstractTestRestApi.java#L152 exclude R interpreter during tests. But it doesn't work with new registration mechanist

jongyoul:
Thanks for checking it. As you mentioned, we removed those kind of tests but we should take care of it because we couldn't break to install any of modules. Your comment makes sense to me and thank you for letting me know the scala version issue between them. LGTM and I'll merge it

PR #1951, opened 2017-01-26
[ZEPPELIN-2020] Invoke method from resource

Leemoonsoo:
### What is this PR for?
This PR provides a way to invoke method remotely against the resource in the resource pool.
This will be particularly useful when the object in the resource pool is non-serializable (like jdbc connection, spark dataframe).

This PR extends data access API of Resource from

```
class Resource {
  // get object
  public Object get();
}
```

to 
```
class Resource {
  // get object
  public Object get()

  // call method of object and get return value
  public Object invokeMethod(String methodName, Class [] paramTypes, Object [] params);

  // call method of object and save return value into resourcepool
  public Resource invokeMethod(String methodName, Class [] paramTypes, Object [] params, String returnResourceName);
}
```

### What type of PR is it?
Feature

### Todos
* [x] - API for remote method invocation and get result
* [x] - API for remote method invocation and store result into resource pool
* [x] - Unittest

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2020

Unittest included 

### Questions:


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
```diff
@@ -211,6 +318,42 @@ public void putResponseGetResource(String resourceId, ByteBuffer object) {
...
+   */
+  public void putResponseInvokeMethod(
+      InvokeResourceMethodEventMessage invokeMessage, ByteBuffer object) {
+    Object o = null;
+    try {
+      o = Resource.deserializeObject(object);
+    } catch (IOException e) {
+      logger.error(e.getMessage(), e);
+    } catch (ClassNotFoundException e) {
+      logger.error(e.getMessage(), e);
```
felixcheung: should it fail / exit then without passing `o` along? Since `o` can be null in that case

Leemoonsoo: Still need to put null into `getInvokeResponse`, otherwise, https://github.com/apache/zeppelin/pull/1951/files#diff-5d91426e574ad85664eb7a7857ef6e90R207 may not finish.

Leemoonsoo:
Merge to master if no further discussions

PR #1952, opened 2017-01-27
[ZEPPELIN-753] TableData abstraction

Leemoonsoo:
### What is this PR for?
This PR implements TableData abstraction that every interpreter can use to share data through resource pool.

TableData.java - Definition of TableData abstraction
TableDataProxy.java - Proxy object of TableData. This enables data access through ResourcePool
InterpreterResultTableData - Convert Table type interpret result into TableData. Each interpreter can implement it's own TableData abstraction. For example JDBCTableData, DataFrameTableData, and so on.

### What type of PR is it?
Feature

### Todos
* [x] - TableData
* [x] - TableDataProxy
* [x] - Unittest

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-753

Unittest included

### Questions:


1ambda:
@Leemoonsoo Nice to have abstracted table structure inside zeppelin. It enable us many things even i just glanced at. 

I will take look deeply and comment soon.

Leemoonsoo:
@1ambda do you have any comment?

1ambda:
Sorry for late reply. I was taking a look also recent resource related commit to understand fully. 

1. This PR is well abstracted and shipped with enough unit test.
2. And easy to share table type interpreter results with other interpreters. The only thing interpreter develop should do is creating it's own table data class (e.g `JDBCTableData`, `DataFrameTableData` as Leemoonsoo mentioned.) to be inserted into pool using interpreter specific low-level result (e.g `JDBCResultSet`, ...)
3. This also enable us to create table data view for shared result in pools.

@Leemoonsoo Now i have a question.

- I can't imagine how one interpreter (e.g JDBC) is able to use shared table result from another interpreter (e.g spark.sql). I know we have APIs to get results from pool. But getting doesn't mean using (e.g querying) directly. 

Thanks!

Leemoonsoo:
@1ambda Thanks for taking a look.

Regarding query directly over the TableData in the ResourcePool, it depends on each interpreter.
For example, SparkInterpreter can implement [DataSource API](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala) to load data from ResourcePool. Then SparkInterpreter will able to query data created from other interpreter, when the data is abstracted by TableData and exist in the ResourcePool.

The same approaches can be applied to all other interpreter. So any interpreter that supports plugin data source layer, might implement ResourcePool as a data source and make query.

Leemoonsoo:
@1ambda further discussion about "how interpreter access data from other interpreter" can be done in https://issues.apache.org/jira/browse/ZEPPELIN-2019.

Merge this PR to master if no further discussions.

PR #1879, opened 2017-01-10
[HOTFIX : ZEPPELIN-1932] paragraph blur error

cloverhearts:
### What is this PR for?
When one or more hidden editors are present, clicking on the editor will cause a blur error.
This means that when a paragraph is hidden through ng-if,
Caused by calling the blur method in the absence of an editor object.

### What type of PR is it?
Hot Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1932

### How should this be tested?
1. create paragraph and open debug console.
2. enable hide paragraph.
3. page refresh
4. click to anywhere paragraph.

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #1963, opened 2017-01-31
[ZEPPELIN-2033] Handle focus/blur of paragraph with hidden editor

minahlee:
### What is this PR for?
#1879 check if `$scope.editor` is null on `focusParagraph` message, and if it is, just return without handling focus/blur.
Instead of doing null check in the beginning of `$scope.on(focusParagraph)`, I made null check to be scoped only to `$scope.editor`'s method invocation.
FYI, when I say focus/blur, it means paragraph focus. Focused paragraph has different css style from blurred paragraph.

### What type of PR is it?
Bug Fix | Hot Fix

### What is the Jira issue?
[ZEPPELIN-2033](https://issues.apache.org/jira/browse/ZEPPELIN-2033)

Go to `Zeppelin Tutorial/Matplotlib (Python • PySpark)` notebook and see:
 - if first paragraph is blurred, when you click second paragraph.
 - if first paragraph is not run when you run second paragraph with shift + enter. In current master, first editor is not blurred even if you click second paragraph, which makes both first and second paragraph to be focused. This will make both paragraphs to be run.
 - if it is focused when you click third paragraph whose editor is hidden. In current master, it won't work.

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -532,23 +532,12 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
         enableLiveAutocompletion: false
       });
 
-      $scope.handleFocus = function(value, isDigestPass) {
-        $scope.paragraphFocused = value;
-        if (isDigestPass === false || isDigestPass === undefined) {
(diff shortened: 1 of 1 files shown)


minahlee:
@cloverhearts please help review

Leemoonsoo:
Tested and LGTM

minahlee:
Merge into branch-0.7 and master

PR #1968, opened 2017-02-02
[ZEPPELIN-1551] Remove the "default" text against interpreters on interpreters screen as it is providing incorrect semantics

namanmishra91:
### What is this PR for?
This PR addresses the incorrect semantics and confusion caused by the "default" text against an interpreter group on interpreters page. 
The intent of of this tag seems to be informing the user which of the interpreters among a particular group will be used by default if the user doesn't specify any %<interpreter> tag. However, the "default" tag is always tied to the first interpreter of the first interpreter group on the page and isn't displayed for any other interpreter groups; therefore not serving any purpose. It also leads to confusion with the default interpreter bound to a notebook on bindings page.

It would be better to remove this default tag from interpreters in order to avoid the confusion.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1551

Run Zeppelin server. Open Interpreters menu. "default" shouldn't be visible against any interpreter group.

### Questions:

zeppelin-1551

### DIFF
--- a/zeppelin-web/src/app/interpreter/interpreter.html
+++ b/zeppelin-web/src/app/interpreter/interpreter.html
@@ -103,7 +103,6 @@ limitations under the License.
               %<span ng-show="!$parent.$first || $first">{{setting.name}}</span
               ><span ng-show="(!$parent.$first || $first) && !$first">.</span
               ><span ng-show="!$first">{{interpreter.name}}</span>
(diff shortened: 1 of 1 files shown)


jongyoul:
LGTM. merging if there's no more discussion

namanmishra91:
The test failures don't seem to be related to my change.

PR #1969, opened 2017-02-02
[ZEPPELIN-1465] Add an option to allow S3 server-side encryption

jeff-cyft:
### What is this PR for?
Provide a configuration option that will cause the S3 Notebook repo to request server-side encryption of saved notebooks.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1465

Enable the configuration option, save a notebook in zeppelin, and confirm in the AWS S3 Console that the related file was saved with AES-256 encryption on the server-side.  (Properties tab, Detail section)

### Questions:
No

No.

I added mentions of the new option in existing documentation.

Thank you!

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
@@ -55,6 +55,7 @@ import com.amazonaws.services.s3.model.GetObjectRequest;
 import com.amazonaws.services.s3.model.KMSEncryptionMaterialsProvider;
 import com.amazonaws.services.s3.model.ListObjectsRequest;
 import com.amazonaws.services.s3.model.ObjectListing;
+import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.PutObjectRequest;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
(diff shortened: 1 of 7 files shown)


Leemoonsoo:
Thanks @jeff-cyft for the contribution!
LGTM and merge to master if no further discussions

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
```diff
@@ -234,7 +237,17 @@ public void save(Note note, AuthenticationInfo subject) throws IOException {
...
       writer.write(json);
       writer.close();
-      s3client.putObject(new PutObjectRequest(bucketName, key, file));
+
+      PutObjectRequest putRequest = new PutObjectRequest(bucketName, key, file);
+
+      if (useServerSideEncryption) {
+        // Request server-side encryption.
+        ObjectMetadata objectMetadata = new ObjectMetadata();
+        objectMetadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
```
felixcheung: very minor q: should SSEAlogrithm be configurable too?

felixcheung:
LGTM

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
jeff-cyft: it's a good question.
1. There is currently only the one documented valid option. http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ObjectMetadata.html
2. There's no enumeration in the API that can be searched by String so in the future this code would need amendment anyway to support potential new options.  I thought about letting the zeppelin configurer specify any string they'd like but it seemed error-prone to expect someone configuring zeppelin to be aware of constant string literals in the S3 Java API.

felixcheung: great! thanks for the details!

PR #1972, opened 2017-02-04
[WIP][DOC] 0.7.0 release note

minahlee:
### What is this PR for?
Add 0.7.0 release note

### What type of PR is it?
Documentation

### Todos
* [x] Confirm release date and update list of people who verified release once vote pass.
* [x] Update doap.rdf

### How should this be tested?
Open [zeppelin-release-0.7.0.md](https://github.com/minahlee/zeppelin/blob/28dfd26fa6f6fe87cd629770674dcbba10b50e04/releases/zeppelin-release-0.7.0.md) and see if there is missing feature that needs to be mentioned in release note.

### Screenshots (if appropriate)
Download page
screen shot 2017-02-04 at 10 48 07 pm

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? -

PR #1932, opened 2017-01-23
[ZEPPELIN-2000] Run paragraph on enter when select dynamic form value changed

minahlee:
### What is this PR for?
Run paragraph on enter when select dynamic form value changed to make paragraph runnable in report mode.

### What type of PR is it?
Bug Fix | Hot Fix

### What is the Jira issue?
[ZEPPELIN-2000](https://issues.apache.org/jira/browse/ZEPPELIN-2000)

### How should this be tested?
1. Go to `Zeppelin Tutorial/Basic Features (Spark)` notebook
2. Change view mode to `report`.
3. Change selected value from `married` to `single` in 5th paragraph and hit enter.
4. See if paragraph runs

### Screenshots (if appropriate)
jan-23-2017 14-14-05

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? yes. docs update included.

PR #1647, opened 2016-11-16
[ZEPPELIN-1013] Don't run paragraph on selectbox change

Mleekko:
### What is this PR for?
Enabling to change values of multiple selectboxes witout running the paragraph

### What type of PR is it?
[Bug Fix]

### Todos
No

### What is the Jira issue?
[ZEPPELIN-1013]

### How should this be tested?
1. Create a paragraph with  a selectbox via z.select(...)
2. Select any value other than currently selected
Actual result:   value changes and paragraph gets executed
Expected result: value changes

### Questions:
* Does the licenses files need update?
No
* Is there breaking changes for older versions?
No
* Does this needs documentation?
No

PR #1974, opened 2017-02-05
[ZEPPELIN-2000][HOTFIX] Run paragraph on ng-change when select value changed.

prabhjyotsingh:
### What is this PR for?
Run paragraph on ng-change when select value changed.

### What type of PR is it?
[Bug Fix | Hot Fix]

### What is the Jira issue?
* [ZEPPELIN-2000](https://issues.apache.org/jira/browse/ZEPPELIN-2000)

 - Go to Zeppelin Tutorial/Basic Features (Spark) notebook
 - Change selected value from married to single in 5th paragraph and hit enter.
 - Change view mode to report and check step 2.
 - See if paragraph runs

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
@@ -30,6 +30,7 @@ limitations under the License.
       <select class="form-control input-sm"
              ng-if="paragraph.settings.forms[formulaire.name].options && paragraph.settings.forms[formulaire.name].type != 'checkbox'"
              ng-enter="runParagraphFromButton(getEditorValue())"
+             ng-change="runParagraphFromButton(getEditorValue())"
              ng-model="paragraph.settings.params[formulaire.name]"
              ng-class="{'disable': paragraph.status == 'RUNNING' || paragraph.status == 'PENDING' }"
              name="{{formulaire.name}}"


prabhjyotsingh:
@minahlee, @jongyoul, @felixcheung Please help review this one.

felixcheung:
looks good. I guess the difference between this and PR #1932 is

> Change selected value from married to single in 5th paragraph ~~and hit enter.~~

?

Leemoonsoo:
fyi, this behavior change is introduced in #1647.

felixcheung:
hmm.. interesting, so it's by design then?

Leemoonsoo:
I think Zeppelin has been back and forth couple of times with this behavior. 
I agree current behavior (require user run paragraph after dynamic form change) can confuse user because

* Previous version does not require run paragraph after dynamic form change
* After dynamic form change, zeppelin just does nothing. i.e. zeppelin does not kindly guide user run paragraph manually.

If we want to resolve issue https://issues.apache.org/jira/browse/ZEPPELIN-1013 in 0.7.0, then we can keep current behavior and add some visual notification after dynamic form change. (e.g. such as  blink or bounce 'run' button when user change dynamic form)

Or we can revert change and try resolve ZEPPELIN-1013 in 0.8.0

felixcheung:
Sorry I didn't find this earlier or reviewed the fixes more closely.
I know everyone has been putting a lot of effort in rolling out this release, so I don't want to block it.

I do think it could be very confusing for long time user with this behavior change.
So perhaps we should do either what I proposed https://github.com/apache/zeppelin/pull/1974#issuecomment-277502448 or you suggested https://github.com/apache/zeppelin/pull/1974#issuecomment-277502684 with visual clue.

But I'm ok with just documenting this in release note and move on to 0.7+

minahlee:
I agree that this can be confusing for the long time user, but I think this is not a blocking issue since user can still run the paragraph. I believe we can handle this issue in next maintenance release.
I just created ticket [ZEPPELIN-2060](https://issues.apache.org/jira/browse/ZEPPELIN-2060) and set the fixed version to 0.7.1.
Release note mentioned this change [here](https://github.com/apache/zeppelin/pull/1972/files#diff-7c8381692ac3b26e7bc352880cc1126eR116).

@prabhjyotsingh Shall we close this?

prabhjyotsingh:
Yes, sure I'll close this.

PR #1975, opened 2017-02-05
[ZEPPELIN-2046] The final output is a little inconsistent with the streaming output

Leemoonsoo:
### What is this PR for?
he final output is not consistent with the streaming output. The final output has one extra blank lines between outputs.

### What type of PR is it?
Bug Fix

### Todos
* [x] - make streaming output and static output the same

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2046

run 

```
%sh echo hello && echo "" && sleep 1 && echo world && sleep 1
```
and see if streaming output is the same to final output

before
output_append_fix_before

after
output_append_fix

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.css
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.css
@@ -76,6 +76,10 @@ table.dataTable.table-condensed .sorting_desc:after {
   white-space:pre-wrap; /** to preserve white-space and newlines of result */
 }
 
+.plainTextContent div {
+  min-height:17px;
+}
+
 .graphContainer {
   position: relative;
   margin-bottom: 5px;
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -488,12 +488,8 @@ function ResultCtrl($scope, $rootScope, $route, $window, $routeParams, $location
 
     // pop all stacked data and append to the DOM
     while (textResultQueueForAppend.length > 0) {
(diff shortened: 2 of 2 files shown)


zjffdu:
@Leemoonsoo  I still on extra blank between outputs which is inconsistent with zeppelin 0.6

1ambda:
I think we can use only 1 `div` element instead of appending `div` for streaming output by replacing it's contents (`html()`).

Leemoonsoo:
@zjffdu If blank lines of output is output from "import ..." in Scala, that's slightly different issue.
Otherwise, please give me an example

zjffdu:
@Leemoonsoo I don't think the blank line is caused by scala. Here's what I see in zeppelin 0.6.2 and zeppelin 0.7, both use spark 1.6.2

Zeppelin 0.7
image

Zeppelin 0.6.2
image

Leemoonsoo:
@zjffdu If you take a look https://github.com/apache/zeppelin/blob/master/notebook/2A94M5J1Z/note.json#L89, you'll see there're two new lines between outputs. Spark interpreter (scala) seems printing one more new line between outputs.

Current master displays those new lines correctly. 0.6.x didn't displayed those new line well.
So, i think one more new line between output in scala is problem need be addressed in SparkInterpreter in separate issue, while this PR addresses difference between streaming output and static output.

zjffdu:
Thanks @Leemoonsoo , it make sense to fix it in another PR.

Leemoonsoo:
@zjffdu Created an issue to address unnecessary newline from SparkInterpreter https://issues.apache.org/jira/browse/ZEPPELIN-2067

Merge this PR to master if no further discussions.

PR #1978, opened 2017-02-06
Hive proxy user doesn't work if interpreter name is not JDBC

r-kamath:
### What is this PR for?
Hive proxy user doesn't work if interpreter name is not JDBC

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
[ZEPPELIN-2062](https://issues.apache.org/jira/browse/ZEPPELIN-2062)

Create a new interpreter named "hive" and run a hive query in a Kerberos environment.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -375,7 +375,7 @@ public class JDBCInterpreter extends Interpreter {
             if (user == null) {
               connection = getConnectionFromPool(url, user, propertyKey, properties);
             } else {
-              if (url.trim().indexOf("jdbc:hive") == 0) {
+              if (url.trim().startsWith("jdbc:hive")) {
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
Good catch! LGTM!

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -375,7 +375,7 @@ public Connection getConnection(String propertyKey, InterpreterContext interpret
             if (user == null) {
               connection = getConnectionFromPool(url, user, propertyKey, properties);
             } else {
-              if ("hive".equalsIgnoreCase(propertyKey)) {
+              if (url.trim().indexOf("jdbc:hive") == 0) {
```
felixcheung: should it do startsWith instead?

r-kamath:
@felixcheung updated in 6d55d48

prabhjyotsingh:

felixcheung:
LGTM

PR #1979, opened 2017-02-06
[ZEPPELIN-2063] Hive Jdbc interpreter does not relogin if kerberos ticket expired when hive.server2.transport.mode is http

prabhjyotsingh:
### What is this PR for?
Hadoop Client will re-login once the ticket expired in case of RPC and so when hive.server2.transport.mode is binary, Hive Jdbc interpreter does a relogin and works fine. But when Rest API is used i.e when hive.server2.transport.mode is http, it is not doing a re-login and so fails with GSS exception.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2063](https://issues.apache.org/jira/browse/ZEPPELIN-2063)

Run hive in http mode i.e. hive.server2.transport.mode is http. Run any query say `show tables` now wait for key to expire (usually its 24hrs), now try to run the same paragraph again without restarting zeppelin-server or jdbc interpreter. It should not fail with `GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)
` exception

N/A

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
+import com.google.common.base.Throwables;
 import org.apache.commons.dbcp2.ConnectionFactory;
 import org.apache.commons.dbcp2.DriverManagerConnectionFactory;
 import org.apache.commons.dbcp2.PoolableConnectionFactory;
@@ -577,6 +578,7 @@ public class JDBCInterpreter extends Interpreter {
           getJDBCConfiguration(user).saveStatement(paragraphId, statement);
 
           boolean isResultSetAvailable = statement.execute(sqlToExecute);
+          getJDBCConfiguration(user).setConnectionInDBDriverPoolSuccessful(propertyKey);
           if (isResultSetAvailable) {
             resultSet = statement.getResultSet();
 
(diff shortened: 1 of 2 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -618,21 +620,54 @@ private InterpreterResult executeSql(String propertyKey, String sql,
...
+        } else if (UserGroupInformation.isLoginTicketBased()) {
+          UserGroupInformation.getLoginUser().reloginFromTicketCache();
+        }
+      } catch (IOException e) {
+        logger.error("Cannot reloginFromKeytab " + sql, e);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        PrintStream ps = new PrintStream(baos);
+        e.printStackTrace(ps);
+        String errorMsg = new String(baos.toByteArray(), StandardCharsets.UTF_8);
+        interpreterResult.add(errorMsg);
```
felixcheung: do we want to full exception + call stack to show up in the interpret result - that seems a bit much?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -618,21 +620,54 @@ private InterpreterResult executeSql(String propertyKey, String sql,
...
+    }
+    return interpreterResult;
+  }
 
+  private InterpreterResult reLoginFromKeytab(String propertyKey, String sql,
+     InterpreterContext interpreterContext, InterpreterResult interpreterResult) {
+    String user = interpreterContext.getAuthenticationInfo().getUser();
+    try {
+      closeDBPool(user, propertyKey);
+    } catch (SQLException e) {/*ignored*/}
```
felixcheung: should it log the exception here?

prabhjyotsingh: I thought since, this in under reLogin and we don't care about if we were able to successfully  close the connection or not, hence, ignored. But, yes, I'll add some exception here, user may want to know about this.

I'll add exception here.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: Yes, make sense, I'll reduce that.

r-kamath:
Tested. LGTM.

prabhjyotsingh:

PR #1981, opened 2017-02-07
[ZEPPELIN-2067] SparkInterpreter prints unnecessary newline

Leemoonsoo:
### What is this PR for?
Spark interpreter prints unnecessary new line before the evaluation output is printed.
See https://github.com/apache/zeppelin/pull/1975#issuecomment-277581660.

This PR make SparkInterpreter ignores unnecessary preceding newline from ReplReporter.error()

### What type of PR is it?
Bug Fix

### Todos
* [x] - Ignore unnecessary preceding newline
* [x] - unittest

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2067

run
```
%spark
import java.util.Date
import java.net.URL
```

and see if result looks like (no new line in front of each lines)

```
import java.util.Date
import java.net.URL
```

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -993,6 +993,7 @@ public class SparkInterpreter extends Interpreter {
   }
 
   private Results.Result interpret(String line) {
+    out.ignoreLeadingNewLinesFromScalaReporter();
     return (Results.Result) Utils.invokeMethod(
         intp,
         "interpret",
@@ -1261,7 +1262,6 @@ public class SparkInterpreter extends Interpreter {
     if (varName == null || varName.isEmpty()) {
       return;
     }
-
     Object lastObj = null;
     try {
       if (Utils.isScala2_10()) {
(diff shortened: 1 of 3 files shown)


spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -993,6 +993,7 @@ public String getSparkUIUrl() {
   }
 
   private Results.Result interpret(String line) {
+    out.ignoreLeadingNewLinesFromScalaReporter();
```
felixcheung: is there a reason we have to set this every time interpret is called, and not have it sticky?

Leemoonsoo: It's because of Scala Reporter may print some new lines in the message (not preceding new lines). https://github.com/apache/zeppelin/pull/1981/files#diff-b2cec9d09177210f5524d66dbf3fe857R59 unset this flag to print new lines from Scala Reporter. So we need to let InterpreterOutputStream know when new interpreter call starts, so it can ignore only preceding newlines from Scala Reporter, not newlines from Scala Reporter in the middle of the messages.

prabhjyotsingh:
Tried on local, it works. LGTM.

zjffdu:
LGTM

Leemoonsoo:
Thanks @felixcheung @prabhjyotsingh @zjffdu taking a look.
Merge to master if no more comment.

PR #1935, opened 2017-01-24
[gh-pages][ZEPPELIN-1973] List all available Helium packages under zeppelin.apache.org

AhyoungRyu:
### What is this PR for?
After #1842 merged, anyone can create his/her own `VISUALIZATION` pkg as described in [Writing a new Visualization](https://zeppelin.apache.org/docs/0.7.0-SNAPSHOT/development/writingzeppelinvisualization.html) and can publish it to npm registry. The below img describes "the current Helium package mechanism(black outline) + proposing from me to list all Helium pkgs published in npm registry(red one) in Zeppelin official website".  So what I did to do this was the red-outlined factors. 

screen shot 2017-01-24 at 4 44 42 pm

> `helium.js` is located in here: [https://s3.amazonaws.com/helium-package/helium.js](https://s3.amazonaws.com/helium-package/helium.js) and it's updated automatically by CloudWatch every 1 hour.

**Possible future work**
Currently to enable/disable the helium VIZ pkg via Helium GUI menu in Zeppelin, user needs to place Helium info json files under `ZEPPELIN_HOME/helium/` manually. But from now, we can substitute this not user friendly way by using `helium.js`. (already opened #1936 by @Leemoonsoo  )

**Any issues?**
But as I mentioned in [ZEPPELIN-1973](https://issues.apache.org/jira/browse/ZEPPELIN-1973), since I used AWS services (Lambda, S3 and CloudWatch), I needed an account. So for continous management, I'm willing to share the information that need to run/packaging/deploy Lambda function with Zeppelin PMCs if you Zeppelin community thinks I need to do it. 

### What type of PR is it?
Improvement | Feature

### Todos
* [x] - make result filtered by type (`VISUALIZATION` & `SPELL` introduced by #1940 )
* [x] - Adjust navbar when screen width is small 

### What is the Jira issue?
[ZEPPELIN-1973](https://issues.apache.org/jira/browse/ZEPPELIN-1973)

### How should this be tested?
Build `gh-pages` branch as described in [here](https://github.com/apache/zeppelin/tree/gh-pages#build-website). Or just see the below gif sreenshot img :)

### Screenshots (if appropriate)
record

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #1982, opened 2017-02-07
[ZEPPELIN-2069] Helium Package Configuration

1ambda:
### What is this PR for?

Supporting helium package configurations. I attached screenshots.

#### Implementation details.

In case of spell, spell developer can create config spec in their `package.json` and it will be part of `helium.json` which is consumed by Zeppelin.

```
  "config": {
    "repeat": {
      "type": "number",
      "description": "How many times to repeat",
      "defaultValue": 1
    }
  },
```

1. Persists conf per `package name@package version` since each version can require different configs even if they are the same package.
2. Saves key-value config only. Since config spec (e.g `type`, `desc`, `defaultValue`) can be provided. So it's not efficient save both of them. 
3. Extracts config related functions to `helium.service.js` since it can be used not only in `helium.controller.js` for view but also should be used in `paragraph.controller.js`, `result.controller.js` for executing spell. 

### What type of PR is it?
[Feature]

### Todos
* [x] - create config view in `/helium`
* [x] - persist config per `package@version`
* [x] - pass config to spell

### What is the Jira issue?

[ZEPPELIN-2069](https://issues.apache.org/jira/browse/ZEPPELIN-2069)

- Build with examples `mvn clean package -Phelium-dev -Pexamples -DskipTests;`
- Open `/helium` page
- Update the `echo-spell` config
- Execute the spell like the screenshot below. (you don't need to refresh the page, since executing spell will fetch config from server)

config

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/HeliumPackage.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/HeliumPackage.java
@@ -18,6 +18,8 @@ package org.apache.zeppelin.helium;
 
 import org.apache.zeppelin.annotation.Experimental;
 
+import java.util.Map;
+
 /**
  * Helium package definition
  */
@@ -25,6 +27,7 @@ import org.apache.zeppelin.annotation.Experimental;
 public class HeliumPackage {
   private HeliumType type;
   private String name;           // user friendly name of this application
+  private String version;
   private String description;    // description
   private String artifact;       // artifact name e.g) groupId:artifactId:versionId
   private String className;      // entry point
@@ -33,7 +36,8 @@ public class HeliumPackage {
   private String license;
   private String icon;
 
-  public SpellPackageInfo spell;
+  private SpellPackageInfo spell;
+  private Map<String, Object> config;
 
   public HeliumPackage(HeliumType type,
                        String name,
@@ -81,6 +85,10 @@ public class HeliumPackage {
     return name;
   }
 
+  public String getVersion() {
+    return version;
+  }
+
   public String getDescription() {
     return description;
   }
@@ -100,6 +108,7 @@ public class HeliumPackage {
   public String getLicense() {
     return license;
   }
+
   public String getIcon() {
     return icon;
   }
@@ -107,4 +116,6 @@ public class HeliumPackage {
   public SpellPackageInfo getSpellInfo() {
     return spell;
   }
+
+  public Map<String, Object> getConfig() { return config; }
 }
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/HeliumPackage.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/HeliumPackage.java
@@ -27,7 +27,6 @@ import java.util.Map;
 public class HeliumPackage {
   private HeliumType type;
   private String name;           // user friendly name of this application
-  private String version;
   private String description;    // description
   private String artifact;       // artifact name e.g) groupId:artifactId:versionId
   private String className;      // entry point
@@ -83,10 +82,6 @@ public class HeliumPackage {
 
   public String getName() {
     return name;
-  }
-
-  public String getVersion() {
-    return version;
   }
 
   public String getDescription() {
--- a/zeppelin-web/src/app/spell/spell-base.js
+++ b/zeppelin-web/src/app/spell/spell-base.js
@@ -31,9 +31,10 @@ export class SpellBase {
    * Consumes text and return `SpellResult`.
    *
    * @param paragraphText {string} which doesn't include magic
+   * @param config {Object}
    * @return {SpellResult}
    */
-  interpret(paragraphText) {
+  interpret(paragraphText, config) {
     throw new Error('SpellBase.interpret() should be overrided');
   }
 
--- a/zeppelin-examples/zeppelin-example-spell-echo/index.js
+++ b/zeppelin-examples/zeppelin-example-spell-echo/index.js
@@ -26,7 +26,28 @@ export default class EchoSpell extends SpellBase {
         super("%echo");
     }
 
-    interpret(paragraphText) {
-        return new SpellResult(paragraphText);
+    /**
+     * Consumes text and return `SpellResult`.
+     *
+     * @param paragraphText {string} which doesn't include magic
+     * @param config {Object}
+     * @return {SpellResult}
+     */
+    interpret(paragraphText, config) {
+        let repeat = 1;
+
+        try {
+            repeat = parseFloat(config.repeat);
+        } catch (error) {
+            /** ignore, use default value */
+        }
+
+        let repeated = "";
+
+        for (let i = 0; i < repeat; i++) {
+            repeated += `${paragraphText}\n`;
+        }
+
+        return new SpellResult(repeated);
     }
 }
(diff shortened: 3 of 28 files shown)


AhyoungRyu:
Great!
How about adding some guide docs? Even if SPELL can be improved in the future, I believe the docs can help not only users but also the other PR reviewers to understand what SPELL is, how we can use SPELL type of pkgs.

zeppelin-web/src/app/spell/spell-base.js
```diff
@@ -31,9 +31,10 @@ export class SpellBase {
    * Consumes text and return `SpellResult`.
    *
    * @param paragraphText {string} which doesn't include magic
+   * @param config {Object}
    * @return {SpellResult}
    */
-  interpret(paragraphText) {
+  interpret(paragraphText, config) {
```
Leemoonsoo: Curious idea behind passing 'config' to interpret(), instead of passing it to constructor. Could you elaborate?

1ambda:
@AhyoungRyu I totally agree! Now we have basic things, thus we need docs for it. I will work it on https://issues.apache.org/jira/browse/ZEPPELIN-2070

zeppelin-web/src/app/spell/spell-base.js
1ambda: @Leemoonsoo Thanks for review! 

- Config can change easily at anytime. 
- If we pass config to ctor, we need to create class everytime

1ambda: Backend interpreter requires restarting to apply new conf. But we don't need to do that for spell.

1ambda: Another idea came up. 

- Setting config before calling `interpret` like `spell.setConfig(conf)`
- Then user can access config like this 

```javascript
interpret(paragraphText) {
  const conf = this.config; // or getConfig()? 
}
```

Leemoonsoo: either current way or setConfig() looks fine to me

zeppelin-interpreter/src/main/java/org/apache/zeppelin/helium/HeliumPackage.java
```diff
@@ -18,13 +18,16 @@
...
+import java.util.Map;
+
 /**
  * Helium package definition
  */
 @Experimental
 public class HeliumPackage {
   private HeliumType type;
   private String name;           // user friendly name of this application
+  private String version;
```
Leemoonsoo: 'artifact' field usually includes version information of the artifact.
I think 'version' field duplicates information included in 'artifact' field.

What do you think? can you elaborate reason for adding version field?

1ambda: Thanks for detailed review. 

The reason I added `version` field is

- it's hard to get version field when package is local. In this case artifact doens't include version value at all. 

I thought having the duplicated field is more cheap and simple way to implement instead of split logic to support both cases (local, remote)

1ambda: In similar context, 

> It would be great if helium package has version field considering local package and latest
> (from https://github.com/apache/zeppelin/pull/1935#issuecomment-276862781)

Leemoonsoo: > it's hard to get version field when package is local. In this case artifact doens't include version value at all.

When package is local, isn't `artifact` (path of the package, in this case) an enough information that Zeppelin needs? I think having `version` when package is local may introduce some confusions & errors.

 - `version` of local package and online package can be the same. It happens when developing helium package and publishing npm. i.e. after publishing package to npm, local package in development mode and remote package version name will be the same.
 - There're possibility of human error that multiple local package is deployed with the name `version` field.

both cases, result is undefined.
However, if we just treat `artifact` as a version, it's not only a one less field but also no possible confusion or human error. 

what do you think?

1ambda: I see what you mean. 

- using `artifact` as key of config is the simplest way to support local, remote helium packages.
- version may be undefined or the same value which published versions have.

considering both cases, it would be good to use `artifact` as key of config. I will modify this PR.

zeppelin-web/src/app/spell/spell-base.js
1ambda: I don't have strong preference. But having config as a param is more explicit way to tell to user **"here is config you can use"**

1ambda:
@Leemoonsoo 

I'v just updated this PR. 

- Using `artifact` as config - https://github.com/apache/zeppelin/pull/1982/commits/6e91b78880777ace0e780c746e376a8e0c88b520
- Disabling `config` button while fetching (see the screenshot below)

disable

1ambda:
But current implementation is really inefficient in terms of API calls. I will refactor it.

1ambda:
@Leemoonsoo Ready to review again. Thanks!

Leemoonsoo:
I tried last commit and when i click 'save' button in Helium package configuration, save button saves configuration but there're no visual indication of it. This confuses little bit.

spell_config_save

And i can still see 'version' field. Is 'version' field in use somewhere?

1ambda:
@Leemoonsoo Thanks for review.

- Then we need to close the configuration panel after saving. I will update it.
- You have to use echo spell provided by zeppelin (`examples/`). Published echo spell doesn't have repeat implementation since spell config feature is not merged. (https://github.com/1ambda/zeppelin-echo-spell/blob/master/index.js)
- Regarding to version, I will check again.

Leemoonsoo:
> The published echo spell doesn't have repeat implementation since spell config feature is not merged. 

I'm little bit confused. Then how does spell page displays configuration if published echo spell does not have configuration?

1ambda:
@Leemoonsoo cause I added config before to test :(

https://github.com/1ambda/zeppelin-echo-spell/blob/master/package.json#L13-#L18

Leemoonsoo:
@1ambda oh i see :-)

1ambda:
just rebased to apply new CI system. 

### Newly added commits

- Close config panel after saving: https://github.com/apache/zeppelin/pull/1982/commits/008ae4a0e81dc5f3d8f9a8a8d8e5b53edb02c9f4
- Remove version field from helium package while cleaning related code: https://github.com/apache/zeppelin/pull/1982/commits/c3100a38b879d8d6d936f528f8f8078bff2623d4

Leemoonsoo:
Shell we also remove `"version": "local"` from 
zeppelin-examples/zeppelin-example-spell-echo/zeppelin-example-spell-echo.json
zeppelin-examples/zeppelin-example-spell-flowchart/zeppelin-example-spell-flowchart.json
zeppelin-examples/zeppelin-example-spell-markdown/zeppelin-example-spell-markdown.json
zeppelin-examples/zeppelin-example-spell-translator/zeppelin-example-spell-translator.json
?

Otherthan that, LGTM.

1ambda:
I'v just rebased to resolve conflict

## Newly added commits

- Remove `local` values in helium package jsons: https://github.com/apache/zeppelin/pull/1982/commits/ce5f8c0c0f88de79bd9c4172a5d74066c9345a50
- DON't refresh `helium.json` when fetching spell config to reduce execution time: https://github.com/apache/zeppelin/pull/1982/commits/dbc4f10fd3ee556d5e38cb4f6e3966661eaf69a9

@Leemoonsoo Could you help review? Thanks again!

Leemoonsoo:

PR #1985, opened 2017-02-07
[ZEPPELIN-2075] Can't stop infinite `while` statement in pyspark Interpreter.

astroshim:
### What is this PR for?
If following code runs with Pyspark Interpreter, there is no way to cancel except Zeppelin Server restart.
```
%spark.pyspark
import time

while True:
    time.sleep(1)
    print("running..")
```

### What type of PR is it?
Bug Fix | Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2075

Run above code with Pyspark Interpreter and try to cancel.

- before
pyspark before

- after
pyspark after

### Questions:


spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -226,7 +226,11 @@ def getCompletion(self, text_value):
       result = json.dumps(list(filter(lambda x : not re.match("^__.*", x), list(completionList))))
       self.interpreterObject.setStatementsFinished(result, False)
 
+def handler_stop_signals(signum, frame):
+  sys.exit("Got signal. " + str(signum))
 
+
+signal.signal(signal.SIGINT, handler_stop_signals)
```
Leemoonsoo: I with it works without this line and handler_stop_signals() function

astroshim: You're right. It seems that spark has already signal handler.
Thanks.

Leemoonsoo:
@astroshim can you take a look CI error ?

karuppayya:
@astroshim This would be a problem in case of scala snippet  also(a loop with a print in other words , that which runs only within driver). Is there a possible  fix such that all interpreters(scala, python) can  benefit.

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -411,10 +414,25 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
     }
   }
 
+  public void interrupt() throws IOException {
+    if (pythonPid > -1) {
+      logger.info("Sending SIGINT signal to PID : " + pythonPid);
+      Runtime.getRuntime().exec("kill -SIGINT " + pythonPid);
+    } else {
+      logger.warn("Non UNIX/Linux system, close the interpreter");
+      close();
```
felixcheung: doesn't this leave the interpreter in a bad state?

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -411,10 +414,25 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
     }
   }
 
+  public void interrupt() throws IOException {
+    if (pythonPid > -1) {
+      logger.info("Sending SIGINT signal to PID : " + pythonPid);
+      Runtime.getRuntime().exec("kill -SIGINT " + pythonPid);
+    } else {
+      logger.warn("Non UNIX/Linux system, close the interpreter");
```
felixcheung: I'm not sure - why is it going to be `non Unix` if pythonPid == -1?

astroshim:
Hello. Sorry for late response. 
@Leemoonsoo Let me fix CI error. 
@karuppayya Let me fix `scala` too in another PR.

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
astroshim: That's because creating process in windows creates a `pseudo process` (thread), and the `pid` is a negative number. ``

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
astroshim: Yes you're right, but `ERROR` status would be a problem?

spark/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterTest.java
```diff
@@ -120,4 +122,33 @@ public void testCompletion() {
...
+      assertEquals(InterpreterResult.Code.SUCCESS,
+        pySparkInterpreter.interpret("a = 1\n", context).code());
+
+      new Thread(new infinityPythonJob()).start();
+      try {
+        Thread.sleep(5000);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
+      }
+      pySparkInterpreter.cancel(context);
```
Leemoonsoo: Don't we need to something that checks if thread is actually finished or not?
I think `thread.join()` is required here. otherwise test will not fail even if job is not cancelled.

astroshim: Thank you for pointing properly out. let me fix it.

Leemoonsoo:
CI test failure is not related (fixed by #2033).

zjffdu:
@astroshim It looks like the python process will be killed, does it mean the interpreter is closed and the state of all the previous paragraphs is lost ? I try the cancel function in jupyter, it seems jupyter just cancel the current cell but keep the python process alive, the previous cell's state is still in the python process. Maybe there's a better approach for this.

zjffdu:
Sorry, it's my misunderstanding. It just interrupt the process rather than kill it. The behavior is consistent with jupyter. It's awesome.

astroshim:
@zjffdu yes, this PR is not kill the python process. 
Thank you for reviewing!

FireArrow:
Sorry for the late comment.
This solution wouldn't work with impersonation, as the zeppelin user wouldn't have permission to interrupt another users process (unless zeppelin run as root, which comes with a whole set of other problems). I suggest you read in the impersonation command, if there is one, and use that to send of the `kill -SIGINT <pid>`.

astroshim:
@FireArrow Note has permission in the zeppelin so if user doesn't have `write` permission on that note, user can't cancel job.

FireArrow:
@astroshim That's not what I'm talking about. Assume the following:
Zeppelin run as user "zeppelin"
In `zeppelin-env.sh`: `export ZEPPELIN_IMPERSONATE_CMD='sudo -H -i -u ${ZEPPELIN_IMPERSONATE_USER} bash -c '`
User `zeppelin` have permission to execute paswordless sudo to local users allowed to log into Zeppelin
Interpreters will with this run as the user starting the process, and the zeppelin user will not have permission to send any signal to them.
My suggestion is to execute the `kill` with `$ZEPPELIN_IMPERSONATE_CMD` to make sure the `kill` is executed in the same way as the interpreter was started in the first place.

astroshim:
@FireArrow Yea, right. I understood. Let me fix it.
Thanks.

astroshim:
@FireArrow I think it seems not can be possible to know if impersonation mode or not in Interpreter side(I couldn't find it) and checking `cancel` paragraph for impersonation is also not good idea (Interpreter sever can be launched itself not the same system with ZeppelinServer)
So I think that the canceling job can run only by user who launched interpreter process can be one option. and In this way, the other user can't stop the job in shared mode.
What do you think?

FireArrow:
I have basically only set up zeppelin in isolated mode with impersonation, so I'm not sure what is the "expected" behavior is outside of that. Someone else will have to pitch in here.

astroshim:
@Leemoonsoo @zjffdu @felixcheung @karuppayya  Could you guys help to review about the impersonation mode?

zjffdu:
Even in the impersonation mode, it is the process owner to interrupt the process. Here it is the interpreter process owner interrupt the process instead of the zeppelin server owner.  But I can not verify it in impersonation mode. It seems the impersonation mode is broken. I got the following error, can anyone else verify the impersonation mode ?

```
ERROR [2017-02-28 09:20:18,882] ({pool-2-thread-2} Job.java[run]:194) - Job failed
org.apache.zeppelin.interpreter.InterpreterException: org.apache.hadoop.security.authorize.AuthorizationException: User: user1 is not allowed to impersonate user1
```

astroshim:
@zjffdu I tested impersonation mode followed http://zeppelin.apache.org/docs/0.7.0/manual/userimpersonation.html and worked well.
BTW for now anybody can interrupt to interpreter process. and that is @FireArrow's worrying. Am i right?

zjffdu:
> User zeppelin have permission to execute paswordless sudo to local users allowed to log into Zeppelin
> Interpreters will with this run as the user starting the process, and the zeppelin user will not have permission to send any signal to them.

I think @FireArrow 's concern is that zeppelin user can kill the process. Actually it is not the user zeppelin send the signal, it is the interpreter process owner. So I think don't think there's any permission issue here. @FireArrow Please help confirm.

astroshim:
@FireArrow ping.

FireArrow:
Terribly sorry for late answer! Not sure why I didn't get any notifications about your answers :confused: 
@zjffdu I haven't looked into who actually sends the signal, but if it is as you say, with the user owning the interpreter process sending the signal, it should be fine. 
My worry was that it was sent by the user running the server, and that if that user isn't root (which would be a terrible idea for other reasons) the OS wouldn't allow sending signals to another users process.
And it is tricky to set up impersonation, but as far as I know it is not broken. I only have my production environment with this set up right now. I suggest not using spark to test with. The `sudo` approach clashes with the spark impersonation, so if you really want to try that you have to set `export ZEPPELIN_IMPERSONATE_SPARK_PROXY_USER=false`, and find a way to make sure the user starting the interpreter has a kerberos ticket (if you need that)

astroshim:
@FireArrow Thank you for confirm your idea and thank @zjffdu help to review!

jongyoul:
Will merge it

PR #1988, opened 2017-02-08
[ZEPPELIN-2080] Fix null column parse issue in KylinInterpreter

yiming187:
### What is this PR for?
KylinInterpreter use regex pattern "\"results\":\\[\\[\"(.*?)\"]]" to extract the result, but if the last column is null, then no quotes in that field. The regex pattern will fail to match.
The KylinInterpreter should consider to support the null field case.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2080

Has prepared the testcase in UnitTest

### Questions:

### DIFF
--- a/kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
+++ b/kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
@@ -25,7 +25,6 @@ import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.zeppelin.interpreter.Interpreter;
 import org.apache.zeppelin.interpreter.InterpreterContext;
-import org.apache.zeppelin.interpreter.InterpreterPropertyBuilder;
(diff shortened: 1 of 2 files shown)


kylin/src/main/java/org/apache/zeppelin/kylin/KylinInterpreter.java
```diff
@@ -172,7 +172,7 @@ private InterpreterResult executeQuery(String sql) throws IOException {
     return rett;
   }
 
-  private String formatResult(String msg) {
+  public String formatResult(String msg) {
```
jongyoul: LGTM except it. How about changing the package of test class as same as this class and make this method as package-local?

yiming187: Thank, jongyoul. Has updated the code.

jongyoul:
Thanks. LGTM. If CI finished and become green, will merge it.

felixcheung:
LGTM

PR #1989, opened 2017-02-08
[ZEPPELIN-2073]Replace AngularObjectRegistry by ResourcePool

qinchaofeng:
### What is this PR for?
To  hide the menu of  interpreter setting, credentials and configurations based on permission.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2073

Has prepared the testcase in UnitTest

### Questions:


1ambda:
Hi @qinchaofeng. I think you pushed different branch. Could you check again?

PR #1994, opened 2017-02-09
[MINOR] Replace angular inject with ngInject

1ambda:
### What is this PR for?

Replaced angular inject with ngInject to reduce duplication and keep simple. There is no reason to use angular inject since

- it can make human-error (e.g. fixing only inject while doesn't modify function args.)
- ngInject is a way to enforce single source of truth (function args)
- we are already applying [ng-annotate](https://github.com/olov/ng-annotate) in webpack. (https://github.com/apache/zeppelin/blob/0589e27e7bb84ec81e1438bcbf3f2fd80ee5a963/zeppelin-web/webpack.config.js#L142)

```javascript
ParagraphCtrl.$inject = [		
  '$scope',		
  '$rootScope',		
  '$route',		
  '$window',		
  '$routeParams',		
  '$location',		
  '$timeout',		
  '$compile',		
  '$http',		
  '$q',		
  'websocketMsgSrv',		
  'baseUrlSrv',		
  'ngToast',		
  'saveAsService',		
  'noteVarShareService',		
  'heliumService'		
 ];		

// can be replaced with one line
'ngInject';
```

### What type of PR is it?
[Improvement]

### What is the Jira issue?

MINOR

will be tested by CI

NONE

### Questions:

### DIFF
--- a/zeppelin-web/src/app/app.controller.js
+++ b/zeppelin-web/src/app/app.controller.js
@@ -14,9 +14,9 @@
 
 angular.module('zeppelinWebApp').controller('MainCtrl', MainCtrl);
 
-MainCtrl.$inject = ['$scope', '$rootScope', '$window', 'arrayOrderingSrv'];
-
 function MainCtrl($scope, $rootScope, $window, arrayOrderingSrv) {
+  'ngInject';
+
   $scope.looknfeel = 'default';
 
   var init = function() {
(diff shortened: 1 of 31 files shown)


Leemoonsoo:
@1ambda Thanks for the improvement.

LGTM and merge to master if no further comments.

PR #2001, opened 2017-02-09
ZEPPELIN-2057 Extract InterpreterSetting functions from InterpreterFactory

jongyoul:
### What is this PR for?
Reducing size of InterpreterFactory and divide some functions from InterpreterFactory. Currently, InterpreterFactory has a lot of functions including management on InterpreterSetting and interpreter processes. This PR extracts InterpreterSetting-related parts into `InterpreterSettingManager`. It also has two unrelated functions: InterpreterSetting and InterpreterGroup. I'll treat it with another PR. This PR has some mechanical changes because it can help understand which parts has been changed. I'll do refactoring them later. 

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Extract all methods and variables related InterpreterSetting from InterpreterFactory

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2057

Works exactly as same before

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -1,24 +1,1120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.zeppelin.interpreter;
 
+import static java.nio.file.attribute.PosixFilePermission.OWNER_READ;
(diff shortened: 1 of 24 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -0,0 +1,1115 @@
...
+                      .equals(Interpreter.registeredInterpreters.get(interpreterKey)
+                          .getClassName())) {
+                    Interpreter.registeredInterpreters.get(interpreterKey)
+                        .setPath(interpreterDirString);
+                    logger.info("Interpreter " + interpreterKey + " found. class=" + className);
+                    cleanCl.put(interpreterDirString, ccl);
+                  }
+                }
+              } catch (Throwable t) {
+                // nothing to do
```
felixcheung: log this error?

felixcheung:
generally looks good

zjffdu:
@jongyoul Do you have a design to share for the whole refactoring ? I think you would have a lot following up PR for the interpreter component refactoring.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
jongyoul: We don't have to do it because this is a trial to check if className exists or not. From old way, Interpreter class had own static initialization code so we should load that class to initialize it. This `catch` occurs while failing `Class.forName` and it's intended.

jongyoul:
@zjffdu I won't change any interpreter protocol in these PRs. This refactoring PRs rearrange code and divide methods into several classes. The next step is to change interpreter management parts which are related to InterpreterGroup actually. Cluster management is not scope of these PRs. I'll share cluster management things with email or wiki docs

jongyoul:
CI passed with https://travis-ci.org/jongyoul/zeppelin/builds/201829204 Will merge it if there's no further discussion

PR #2011, opened 2017-02-12
ZEPPELIN-2102: Restart interpreter automatically

karuppayya:
### What is this PR for?
Sparkcontext is shutdown in following cases
1. calling `sc.stop` explicitly
In this case, the user has to manually restart the interpreter to submit his next spark job
2. OOM error from spark
Restarting of interpreter will not help. The only recovery is restarting zeppelin server
I have not enumerated all the cases in which sc can go down. There might be other cases.
In this PR, restarting interpreter automatically

Edit: Restart of interpreter works fine. Subsequent para run goes through with interpreter restart.

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-2102

1. Run `sc.stop`. Subsequently run another paragraph
2.  Cause spark driver to crash due to OOM, run subsequent spark paragraph

### Questions:


felixcheung:
could you elaborate, why restarting interpreter doesn't help for case # 2?

zjffdu:
For case 1, why not creating a new SparkContext ? `sc.stop` only cause the spark app shutdown, but the remote interpreter process should still be alive. 
Overall, I don't think restarting `SparkContext` implicitly for user is a proper solution. This might cause confusion for users, as creating new `SparkContext` means all the historical state is lost, user have to rerun all the paragraphs. One proper solution I can think of is that send a warning message to front end to tell user the SparkContext is dead for some unknown reason, you need to either create a new one or restarting the interpreter.

karuppayya:
@zjffdu Yes the remote process will still be up which will consume as much memory as configured for driver. In a multi user environment, we might want to release the resources as soon as the user is done with his application. If a user runs `sc.stop` once he is done, we can kill the remote process to allow for other processes. 
+1, for showing appropriate message on UI.
Will adding a zeppelin specific interpreter setting  to gate this be helpful,.?

zjffdu:
I don't think zeppelin should do extra things for spark interpreter.  This would cause confusion for user. Interpreter should do general thing for user, any specific thing should be handled by users.  You might want to kill the remote process after `sc.stop`, but this doesn't mean the same thing for other users.  They may want to keep the process alive to do scala or python things.

karuppayya:
@felixcheung I am not able to repro this scenario now. Restart works fine(Will update the description)
When spark goes OOM, the subsequent para runs throw connection refused exception. This can be avoided by restart of interpreter.

karuppayya:
@zjffdu Thanks for your feedback.
The change is not specific to spark interpreter.  
It is generic so that any other interpreter also can initiate a restart .
I was targeting free-ing up of resources once the spark application completes
I was not aware of the use cases for using spark interpreter as scala, python REPl after sparkcontext is stopped.Then this change is not helpful.

hero0926:
We're meeting same problem as karuppayya. when we trying to unpersist in pyspark+zeppelin, memory dosen't released - so sc.stop for second hand, zeppelin interpreter dies... I concern OOM error couldn't be escaped but is there any possible way to use pyspark without this problem.

@karuppayya , how'bout uncache and make another sc or make sc to null several times.. I think make sc newer in code is alter way to escape this problem.

PR #2014, opened 2017-02-13
[ZEPPELIN-2065] Flaky test: NotebookTest.testSchedulePoolUsage

AlexanderShoshin:
### What is this PR for?

`testSchedulePoolUsage()` test checks that jobs can be executed several times in a row using a cron expression.

It failed sometimes when it checked job completion by monitoring `job.dateFinished` and then tried to assert that `job.status` is `FINISHED`. It was not always true because `job.status` is changed after `job.dateFinished`.

### What type of PR is it?
Bug Fix

### Todos
* [ ] We need to catch job completion by checking `job.status` instead of `job.dateFinished`.
* [ ] It is better to catch job completion by adding job status change listener instead of using `Thread.sleep(..)` in a test method.
* [ ] We can make this test simpler by removing a check that a job will not run after a cron expressin is removed. We already have such check in a previous unit test named `testSchedule()`.

### What is the Jira issue?
[ZEPPELIN-2065](https://issues.apache.org/jira/browse/ZEPPELIN-2065)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
@@ -25,10 +25,8 @@ import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


AlexanderShoshin:
I changed AtomicInteger to CountDownLatch to simplify the code.

AlexanderShoshin:
One travis job failed - looks like known bag https://issues.apache.org/jira/browse/ZEPPELIN-1761

Leemoonsoo:
Thanks @AlexanderShoshin for the fix!
LGTM and merge to master if no further comments

PR #2017, opened 2017-02-13
[ZEPPELIN-1746] Flaky test: PySparkInterpreterTest

Leemoonsoo:
### What is this PR for?
PySparkInterpreter (and PySparkInterpreterMatplotlibTest, SparkInterpreterTest, SparkSqlInterpreterTest) test does not terminate SparkInterpreter after test end.

In the beginning, keeping SparkInterpreter alive was designed to prevent create/destroy SparkInterpreter multiple times in various tests, because it takes some time.

However, that is somehow misbehaving and getting problems like

1. starting another SparkInterpreter before one SparkInterpreter terminates
2. Does not finish PySparkInterpreter process explicitly cause py4j client somehow creates a lot of connection attempt, that consumes all capacity of systems port open, and make next SparkInterpreter start fail.

This PR terminates SparkInterpreter after test class finishes, to prevent above behavior.

### What type of PR is it?
Hot Fix

### Todos
* [x] - Terminate SparkInterpreter after each test class

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1746

CI becomes green

### Questions:

### DIFF
--- a/spark/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterMatplotlibTest.java
+++ b/spark/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterMatplotlibTest.java
@@ -39,14 +39,14 @@ import static org.junit.Assert.*;
 @FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class PySparkInterpreterMatplotlibTest {
 
-  @Rule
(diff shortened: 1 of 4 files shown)


Leemoonsoo:
Since many CI fails with this error, i'm merging it to master as a hotfix.

PR #2028, opened 2017-02-17
[MINOR] Supported old note format in case of not having name of note

jongyoul:
### What is this PR for?
Supporting old format of note which doesn't have name inside. Long time ago, when creating note without specific name, name field become empty string. but it become error while remove note for current master. It will fix this error

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Enable getName returns getId if it has empty value

### What is the Jira issue?
N/A

See the test added

N/A

### Questions:


felixcheung:
maybe minor, could this automatically fix up the old note so that it will be saved with a name?

felixcheung:
LGTM

jongyoul:
test this please

jongyoul:
Will merge it

PR #2036, opened 2017-02-19
[ZEPPELIN-2135] Don't re-license jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java

Leemoonsoo:
### What is this PR for?
jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java is taken from SQLLine 1.0.2 (BSD license) and we can't relicense it.

Remove Apache License header and exclude from RAT check plugin

### What type of PR is it?
Bug Fix

### Todos
* [x] - remove license header
* [x] - exclude from rat check plugin

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2135

### Questions:
* Does the licenses files need update? related

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -898,6 +898,8 @@
               <exclude>**/interpreter-setting.json</exclude>
               <exclude>**/constants.json</exclude>
               <exclude>scripts/**</exclude>
+
+              <!-- from SQLLine 1.0.2, see ZEPPELIN-2135 -->
               <exclude>**/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java</exclude>
 
               <!-- bundled from bootstrap -->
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -898,6 +898,7 @@
               <exclude>**/interpreter-setting.json</exclude>
               <exclude>**/constants.json</exclude>
               <exclude>scripts/**</exclude>
+              <exclude>**/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java</exclude>
```
felixcheung: shall we add a comment here why this is excluded? or is there another way to track the source of given file?

Leemoonsoo:
@felixcheung added a comment

Leemoonsoo:
merge to master and branch-0.7 if no more discussions.

PR #2039, opened 2017-02-20
[MINOR] Use standard java API to interrupt thread

DrIgor:
### What is this PR for?
Use java `Thread.interrupt` method to stop job progress polling thread.

Standard API is:
* proper synchronized 
* able to interrupt `Thread.sleep` with any interval

### What type of PR is it?
[Refactoring]

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
@@ -182,22 +182,16 @@ public abstract class Job {
       this.exception = null;
       errorMessage = null;
       dateFinished = new Date();
-      progressUpdator.terminate();
-    } catch (NullPointerException e) {
-      LOGGER.error("Job failed", e);
-      progressUpdator.terminate();
-      this.exception = e;
-      setResult(e.getMessage());
-      errorMessage = getStack(e);
-      dateFinished = new Date();
     } catch (Throwable e) {
       LOGGER.error("Job failed", e);
-      progressUpdator.terminate();
       this.exception = e;
       setResult(e.getMessage());
       errorMessage = getStack(e);
       dateFinished = new Date();
     } finally {
+      if (progressUpdator != null) {
+        progressUpdator.interrupt();
+      }
       //aborted = false;
     }
   }
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/JobProgressPoller.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/JobProgressPoller.java
@@ -21,48 +21,45 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
+ * Polls job progress with given interval
+ *
+ * @see Job#progress()
+ * @see JobListener#onProgressUpdate(org.apache.zeppelin.scheduler.Job, int)
+ *
  * TODO(moon) : add description.
  */
 public class JobProgressPoller extends Thread {
   public static final long DEFAULT_INTERVAL_MSEC = 500;
-  Logger logger = LoggerFactory.getLogger(JobProgressPoller.class);
+  private static final Logger logger = LoggerFactory.getLogger(JobProgressPoller.class);
+
   private Job job;
   private long intervalMs;
-  boolean terminate = false;
 
   public JobProgressPoller(Job job, long intervalMs) {
+    super("JobProgressPoller, jobId=" + job.getId());
     this.job = job;
-    this.intervalMs = intervalMs;
+    if (intervalMs < 0) {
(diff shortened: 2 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
```diff
@@ -182,22 +182,16 @@ public void run() {
...
     } catch (Throwable e) {
       LOGGER.error("Job failed", e);
-      progressUpdator.terminate();
       this.exception = e;
       setResult(e.getMessage());
       errorMessage = getStack(e);
       dateFinished = new Date();
     } finally {
+      if (progressUpdator != null) {
+        progressUpdator.interrupt();
```
felixcheung: isn't this changing it to call interrupt every time, even when it has no error?

felixcheung:
for a change like this, it would be useful to open a JIRA and to discuss why we are making this change

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
DrIgor: `terminate` was called in both cases: on line 185 in case of success and on line 195 in catch block. So it is better to place it in `finally` to handle both cases.

DrIgor:
@felixcheung `terminate` flag is shared across several threads and is accessed without any synchronization.
And we don't need to implement custom interruption flag because java already has it — `Thread.interrupt`

Standard flag is even better, because it can interrupt `Thread.sleep`.

This PR doesn't change any logic, it just replaces `terminate` with `Thread.interrupted()` and provides proper handling of `InterruptedException`

DrIgor:
Ping-ping)

Do you have other questions? Do you think it's better to proper synchronize `terminate` flag instead?

I reviewed code to understand how zeppelin core works and found several multi threading issues. I would like to help to fix them so I created this PR

felixcheung:
From my understanding, setting a flag to let the thread gracefully runs to end, seems safer than calling `interrupt` on it

DrIgor:
`interrupt` just sets `Thread.isInterrupted` flag and does not immediately stop the thread. Then thread can check this flag and decide if it needs to stop. So in fact, termination logic remains the same.

Maybe you mixed up `Thread.interrupt` and `Thread.stop` methods? `Thread.stop` terminates a thread without graceful shutdown, `Thread.interrupt` doesn't.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
```diff
@@ -182,22 +182,16 @@ public void run() {
       this.exception = null;
       errorMessage = null;
       dateFinished = new Date();
-      progressUpdator.terminate();
-    } catch (NullPointerException e) {
```
felixcheung: why is this removed?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/JobProgressPoller.java
```diff
@@ -21,48 +21,45 @@
...
         Thread.sleep(intervalMs);
-      } catch (InterruptedException e) {
-        logger.error("Exception in JobProgressPoller while run Thread.sleep", e);
       }
-    }
-  }
-
-  public void terminate() {
-    terminate = true;
+    } catch (InterruptedException ignored) {}
```
felixcheung: should this be logged or it will "never" happen since we are not calling thread.stop()?

felixcheung:
Ok I wasn't aware of the thread.interrupt semantic in JVM.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
DrIgor: Both catch block were identical and `catch (Throwable e)` can handle `NullPointerException` too

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/JobProgressPoller.java
DrIgor: `InterruptedException` will be thrown if we call `interrupt` during `Thread.sleep`. I don't log it because it's expected

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
felixcheung: ok, I guess we didn't do anything different there

felixcheung:
ok LGTM.
any comment from other reviewers?

Leemoonsoo:
Code change Looks good to me.

felixcheung:
merging if no more comment

PR #2043, opened 2017-02-20
[ZEPPELIN-2114] adding reload endpoint

Remilito:
### What is this PR for?
Adding endpoint in NotebookRepoRestApi to trigger reload and broadcast of the note list.
Sending a GET request to /api/notebook-repositories/reload will trigger a reload of the note list, just like the reload button does on the homepage.
Q: If you think this endpoint belongs to another API (NotebookRestApi?), let me know.

### What type of PR is it?
Improvement

### What is the Jira issue?
<https://issues.apache.org/jira/browse/ZEPPELIN-2114>

0. add (or remove) a note in the notebook repository
1. curl -X GET http://HOST:PORT/api/notebook-repository/reload
2. verify that you the note list in Zeppelin's home page now contains (or does not contain anymore) the note.

### Questions:
* Does this needs documentation? NotebookRepoRestApi does not have documentation yet. Could add it in another PR if needed.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
@@ -76,6 +76,18 @@ public class NotebookRepoRestApi {
   }
 
   /**
+   * Reload notebook repository
+   */
+  @GET
+  @Path("reload")
+  @ZeppelinApi
+  public Response refreshRepo(){
+    AuthenticationInfo subject = new AuthenticationInfo(SecurityUtils.getPrincipal());
+    notebookWsServer.broadcastReloadedNoteList(subject, null);
+    return new JsonResponse<>(Status.OK, "", null).build();
+  }
+
+  /**
    * Update a specific note repo.
    *
    * @param message
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
@@ -83,6 +83,7 @@ public class NotebookRepoRestApi {
   @ZeppelinApi
   public Response refreshRepo(){
     AuthenticationInfo subject = new AuthenticationInfo(SecurityUtils.getPrincipal());
+    LOG.info("Reloading notebook repository for user {}", subject.getUser());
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
Thanks for the contribution.
Do you mind add a documentation [here](https://github.com/apache/zeppelin/blob/master/docs/rest-api/rest-notebook.md) ?

AhyoungRyu:
>If you think this endpoint belongs to another API (NotebookRestApi?), let me know.

I think `notebook-repositories` is right place for this as you did. 

@Leemoonsoo  It would be better to have separated docs page for "NotebookRepo". Since it's start \w `api/notebook-repositories` whereas "NotebookRestApi" start \w `api/notebook`. I created issue for this [ZEPPELIN-2140](https://issues.apache.org/jira/browse/ZEPPELIN-2140). I'll handle it in another PR :)

Remilito:
@AhyoungRyu: let me know if need me to enrich the new notebook-repositories doc.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
```diff
@@ -76,6 +76,18 @@ public Response listRepoSettings() {
...
 
   /**
+   * Reload notebook repository
+   */
+  @GET
+  @Path("reload")
+  @ZeppelinApi
+  public Response refreshRepo(){
+    AuthenticationInfo subject = new AuthenticationInfo(SecurityUtils.getPrincipal());
+    notebookWsServer.broadcastReloadedNoteList(subject, null);
```
AhyoungRyu: Maybe it's nit; 
How about adding a log msg in here as other API does? e.g. 
```
LOG.info("Reloading notebook repository for user {}", subject.getUser());
```

AhyoungRyu:
@Remilito Sorry for my late response. Tested and it works as expected. 
Regarding the docs that I mentioned was added in #2067. Will be appreciated if you can take a look :)

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRepoRestApi.java
Remilito: You are right, that was missing. Added in latest commit.

AhyoungRyu:
Merge into master and branch-0.7 if there are no more discussions.

PR #2051, opened 2017-02-21
[ZEPPELIN-1968] Added property to disable hive user impersonation

gfalcone:
### What is this PR for?

Added new property "hive.proxy.user"  to disable hive impersonation (on some clusters, this option is disabled) in order to make Hive Interpreter even without this

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1968

Set "hive.proxy.user" to true in the jdbc interpreter setttings, and you should see "Using hive proxy user" in the jdbc logs.

If "hive.proxy.user" has another value, this is not mentionned in the logs

You can also test with the appropriate hive configuration, but this could take longer :) 

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -374,7 +374,8 @@ public class JDBCInterpreter extends Interpreter {
                 if (lastIndexOfUrl == -1) {
                   lastIndexOfUrl = connectionUrl.length();
                 }
-                if (!property.getProperty("hive.proxy.user").equals("false")){
+                boolean hasProxyUser = property.containsKey("hive.proxy.user");
+                if (!hasProxyUser || !property.getProperty("hive.proxy.user").equals("false")){
                   logger.debug("Using hive proxy user");
                   connectionUrl.insert(lastIndexOfUrl, ";hive.server2.proxy.user=" + user + ";");
                 }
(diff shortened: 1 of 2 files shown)


gfalcone:
CI passed all the tests : https://travis-ci.org/gfalcone/zeppelin

prabhjyotsingh:
This looks OK to me, can you add a line note on https://github.com/apache/zeppelin/blob/master/docs/interpreter/jdbc.md#apache-hive to update the documentation.

gfalcone:
@prabhjyotsingh thank you for reviewing, just pushed the doc :)

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -374,16 +374,22 @@ public Connection getConnection(String propertyKey, InterpreterContext interpret
                 if (lastIndexOfUrl == -1) {
                   lastIndexOfUrl = connectionUrl.length();
                 }
-                connectionUrl.insert(lastIndexOfUrl, ";hive.server2.proxy.user=" + user + ";");
+                if (!property.getProperty("hive.proxy.user").equals("false")){
```
prabhjyotsingh: Just realised, can you do a null check for this, if the property key doesn't exists, this will throw NPE.

gfalcone:
@prabhjyotsingh right, didn't see that

gfalcone:
@prabhjyotsingh is everything good for you :) ?

prabhjyotsingh:
Yes, LGTM!

prabhjyotsingh:

gfalcone:
OK, so I deleted that line because I realized that we were in the condition where it's not the hive jdbc, so we don't have to worry about that.

prabhjyotsingh:
Sure thank you @gfalcone, will merge this soon.

gfalcone:
@prabhjyotsingh  the CI passed !

PR #2059, opened 2017-02-23
[ZEPPELIN-2139] Interpreters based on scala_2.11 aren't installed correctly

minahlee:
### What is this PR for?
pom variables such as `${scala.version}` are not replaced to value when you do `mvn install`.
This makes leaf poms to look for this value in parent pom, which is always `2.10.5` for `${scala.version}` and it causes scala library dependency conflict. For example, zeppelin-flink_2.11 will have scala 2.10.5 as dependency.
This PR fixes this problem by using maven flatten plugin.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-2139](https://issues.apache.org/jira/browse/ZEPPELIN-2139)

```
$ ./dev/change_scala_version.sh 2.11
$ mvn clean install -pl 'zeppelin-server,zeppelin-zengine,zeppelin-interpreter,flink' -am -DskipRat -DskipTests -Pscala-2.11
```
Open `~/.m2/repository/org/apache/zeppelin/zeppelin-flink_2.11/0.8.0-SNAPSHOT/zeppelin-flink_2.11-0.8.0-SNAPSHOT.pom` file and see if scala related libraries dependency version is set to 2.11.7

### Questions:


pom.xml
```diff
@@ -308,6 +308,32 @@
         </configuration>
       </plugin>
 
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+        <version>1.0.0</version>
+        <configuration>
+          <flattenMode>ossrh</flattenMode>
```
felixcheung: I assume by flattening we would still able to build different interpreter on different scala version in one `mvn package` or `mvn install` command though

minahlee: Yes, scala version will be different depend on mvn profile that we use.

minahlee:

PR #2072, opened 2017-02-27
[ZEPPELIN-2060] Show tooltip msg after dynamic form value is changed

AhyoungRyu:
### What is this PR for? 
As described in [ZEPPELIN-2060](https://issues.apache.org/jira/browse/ZEPPELIN-2060), the dynamic form's behaviour has been changed back and forth.
In current status(nothing happened when user change the dynamic form value), I made this tooltip msg automatically shows up. This will be helpful to let users know they need to click run button or `Shift+Enter`. 

screen shot 2017-02-27 at 2 58 16 pm

If there can be better way for this, please let me know :)
### What type of PR is it?
Improvement

### Todos
* [ ] - set timeout 

### What is the Jira issue?
[ZEPPELIN-2060](https://issues.apache.org/jira/browse/ZEPPELIN-2060)

1. Apply this patch and run Zeppelin web as dev mode
```bash
#under zeppelin-web
$yarn run dev
```

2. Go to Spark tutorial note and try to change value in dynamic forms as below screenshot imgs 

After this change
record
record2

### Questions:
* Is there breaking changes for older versions? yes?

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
@@ -22,6 +22,7 @@ limitations under the License.
 
       <input class="form-control input-sm"
              ng-if="!paragraph.settings.forms[formulaire.name].options"
+             ng-change="showTooltip()"
              ng-enter="runParagraphFromButton(getEditorValue())"
              ng-model="paragraph.settings.params[formulaire.name]"
              ng-class="{'disable': paragraph.status == 'RUNNING' || paragraph.status == 'PENDING' }"
@@ -30,6 +31,7 @@ limitations under the License.
       <select class="form-control input-sm"
              ng-if="paragraph.settings.forms[formulaire.name].options && paragraph.settings.forms[formulaire.name].type != 'checkbox'"
              ng-enter="runParagraphFromButton(getEditorValue())"
+             ng-change="showTooltip()"
              ng-model="paragraph.settings.params[formulaire.name]"
              ng-class="{'disable': paragraph.status == 'RUNNING' || paragraph.status == 'PENDING' }"
              name="{{formulaire.name}}"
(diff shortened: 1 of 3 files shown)


zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
```diff
@@ -22,6 +22,7 @@
 
       <input class="form-control input-sm"
              ng-if="!paragraph.settings.forms[formulaire.name].options"
+             ng-change="showTooltip()"
```
felixcheung: would it pop the tooltip for every single character I type into the text box?

Leemoonsoo:
Great. How does it work in report mode?

zeppelin-web/src/app/notebook/paragraph/paragraph-parameterizedQueryForm.html
AhyoungRyu: Something is changed, then the tooltip will be shown.

AhyoungRyu:
@Leemoonsoo I checked now, the behaviour is same (the tooltip is shown right after sth changed in dynamic form) like default mode. I think this should be fixed. Maybe I need to hide this tooltip in report mode. Or maybe another tooltip msg can be shown sth like "You can't run the paragraph in report mode". Not sure which one is better. 

BTW don't we need to block any changes in dynamic form then? Currently we can't run the paragraph since the run button is not shown and `Shift+Enter` doesn't work.

@Leemoonsoo @felixcheung Any suggestion will be appreciated :)

felixcheung:
dynamic form not working in report mode is a problem, right?
perhaps we get shift-enter to work and then anchor the tooltip to the textbox/dropdown etc?

AhyoungRyu:
@felixcheung Yeah that's possible way. Thanks! 
Currently dynamic form itself is working in report mode by pressing `Enter` after the selected value changed. But the problem is it's not intuitive. Ppl don't know what they should do after changing value since it's automatically run before, but now it's not.
I created an another PR to improve the dynamic form behaviour in #2100. Can you take a look? I'm not sure which one is better. Maybe I can integrate two.. But it'll be better to get more feedbacks!

AhyoungRyu:
I'm closing this PR since https://issues.apache.org/jira/browse/ZEPPELIN-2060 will be handled in #2100.

PR #2075, opened 2017-02-27
[ZEPPELIN-791] Build infra: move all RAT to root pom.xml

DmytroShkvyra:
What is this PR for?

It is better have a single place where we manage project-wise RAT exclusions for a contributions under licenses different from Apache, then let maven sub-modules have them as we do now (makes things harder to track)
What type of PR is it?

[Improvement]
Todos

N/A
What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-791
How should this be tested?

mvn verify -DskipTests
Screenshots (if appropriate)
Questions:

    Does the licenses files need update? - No
    Is there breaking changes for older versions? - No
    Does this needs documentation? - No


DmytroShkvyra:
add wildcard

PR #2079, opened 2017-02-28
ZEPPELIN-2195. Use PYSPARK_PYTHON and PYSPARK_DRIVER_PYTHON over zeppelin.pyspark.python

zjffdu:
### What is this PR for?
`zeppelin.pyspark.python` is zeppelin configuration for the python exec on driver side, it won't affect executor side. It would be better to use `PYSPARK_PYTHON` and `PYSPARK_DRIVER_PYTHON`  which is what spark use officially.  So that user can define their own python exec in interpreter setting for different version of python rather than defining them `zeppelin-env.sh` which is shared globally. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2195

Tested it manually. 

### Questions:

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -106,7 +106,7 @@ You can also set other Spark properties which are not listed in the table. For a
   <tr>
     <td>PYSPARK_PYTHON</td>
     <td>python</td>
-    <td>Python binary executable to use for PySpark in both driver and workers (default is <code>python</code> if available, otherwise <code>python</code>).
+    <td>Python binary executable to use for PySpark in both driver and workers (default is <code>python</code>).
             Property <code>spark.pyspark.python</code> take precedence if it is set</td>
   </tr>
   <tr>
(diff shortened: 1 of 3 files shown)


Leemoonsoo:
Thanks for the improvement.

Do you think this change worth to be mentioned in the [upgrade.md](
https://github.com/apache/zeppelin/blob/master/docs/install/upgrade.md#upgrading-from-zeppelin-06-to-07) with new section "Upgrading from Zeppelin 0.7 to 0.8" ?

docs/interpreter/spark.md
```diff
@@ -104,9 +104,16 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>Local repository for dependency loader</td>
   </tr>
   <tr>
-    <td>zeppelin.pyspark.python</td>
+    <td>PYSPARK_PYTHON</td>
     <td>python</td>
-    <td>Python command to run pyspark with</td>
+    <td>Python binary executable to use for PySpark in both driver and workers (default is <code>python</code> if available, otherwise <code>python</code>).
```
felixcheung: this is saying "(default is <code>python</code> if available, otherwise <code>python</code>)."
?

zjffdu:
Thanks @Leemoonsoo @felixcheung for review. Comment is addressed.

felixcheung:
merging if no more comment

PR #2085, opened 2017-03-01
[ZEPPELIN-1999] get interpreter property with replaced context parameters

tinkoff-dwh:
### What is this PR for?
Adds posibility to use context parameters (types: String.class, Double.class, Float.class, Short.class,
Byte.class, Character.class, Boolean.class, Integer.class, Long.class, ) into property value of interpreter.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1999

1. Add text with markers #{contextFieldNAme} (ex. #{noteId} or #{replName}) to interpreter property value (or add new property of interpreter).
2. Get this property (getProperty(key)), markers should be replaced by context values

### Questions:


tinkoff-dwh:
1. uniqueness of the marker to replace
2. similar format of the parameters in the note (${}, not the same, not to be confused)

andrykrp:
Looks good

mgarmash:
Nice feature! +1

tinkoff-dwh:
@felixcheung
 Are there problems for merge?

.gitignore
```diff
@@ -95,6 +95,8 @@ Thumbs.db
 # intelliJ IDEA project files
 .idea/
 *.iml
+*.iws
+*.ipr
```
felixcheung: what are these for?

docs/manual/interpreters.md
```diff
@@ -41,7 +41,13 @@ Zeppelin interpreter setting is the configuration of a given interpreter on Zepp
 
 <img src="../assets/themes/zeppelin/img/screenshots/interpreter_setting.png" width="500px">
 
-Properties are exported as environment variable when property name is consisted of upper characters, numbers and underscore ([A-Z_0-9]). Otherwise set properties as JVM property.
+Properties are exported as environment variable when property name is consisted of upper characters, numbers and underscore ([A-Z_0-9]). Otherwise set properties as JVM property. 
+You may use parameters from the context of interpreter by add #{contextParameterName} (except fields of paragraph) in value, parameter can be of the following types: string, number, boolean. The list of available context parameters can be viewed in the class [InterpreterContext.java](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterContext.java) + parameter #{user}.
```
felixcheung: since this is talking about a new concept, let's put this into a new paragraph (ie. double new line in markdown,before this line)

felixcheung: i think it'd be better to get casual users to look at Java code - could you add a table here of the types/context parameters it could support?

felixcheung: .. and with some code/paragraph example in text that the user can run and test with?

felixcheung:
I think this is a good feature to have. Let's hold on to get a bit more feedback on this - @Leemoonsoo what do you think about something like this?

.gitignore
tinkoff-dwh: it is extensions of project (jetbrains IDEA)

dwhsys:
I think that this PR is a great step forward to enterprise version of Zeppelin. Using this feature there will be an ability to implement native integration between different interpreters based on single schema per user / note in core analytical DBMS. Another feature is the possibility of data access control without using credentials.

@felixcheung, @Leemoonsoo, what do you think?

Leemoonsoo:
Thanks @tinkoff-dwh for useful new feature.
LGTM!

docs/manual/interpreters.md
```diff
@@ -41,8 +41,45 @@ Zeppelin interpreter setting is the configuration of a given interpreter on Zepp
...
+    <td>string</td>
+  </tr>
+</table>
+
+If context parameter is null then replaced by empty string.
+
+<img src="../assets/themes/zeppelin/img/screenshots/interpreter_setting_with_context_parameters.png" width="500px">
+
+###### Usage
+For example in database exist user `user1` with some password and exists user with same name into Zeppelin. Configure the jdbc interpreter (postgres), set `default.user = #{user}`.
```
AhyoungRyu: >For example in database exist user `user1` with some password and exists user with same name into Zeppelin.

As the above sentence, we assume that the readers already know about how they configure user setting and enable Shiro. But as you know, this `manual/interpreters.md` mainly focuses on Zeppelin beginners(e.g. don't know about what interpreter is in Zeppelin/ how they can bind interpreter to the note ..). Not sure what can be the better description for this, i just thought this sentence need more information. What do you think? :)

docs/manual/interpreters.md
```diff
@@ -41,8 +41,45 @@ Zeppelin interpreter setting is the configuration of a given interpreter on Zepp
 
 <img src="../assets/themes/zeppelin/img/screenshots/interpreter_setting.png" width="500px">
 
-Properties are exported as environment variable when property name is consisted of upper characters, numbers and underscore ([A-Z_0-9]). Otherwise set properties as JVM property.
-
+Properties are exported as environment variable when property name is consisted of upper characters, numbers and underscore ([A-Z_0-9]). Otherwise set properties as JVM property. 
+
+You may use parameters from the context of interpreter by add #{contextParameterName} (except fields of paragraph) in value, parameter can be of the following types: string, number, boolean.
```
masyan: list of available parameters described below

docs/manual/interpreters.md
```diff
@@ -41,8 +41,45 @@ Zeppelin interpreter setting is the configuration of a given interpreter on Zepp
...
+  </tr>
+  <tr>
+    <td>className</td>
+    <td>string</td>
+  </tr>
+</table>
+
+If context parameter is null then replaced by empty string.
+
+<img src="../assets/themes/zeppelin/img/screenshots/interpreter_setting_with_context_parameters.png" width="500px">
```
masyan: no, image width more than 500px

docs/manual/interpreters.md
masyan: maybe, but it is common doc to all interpreters and the best place which I found

docs/manual/interpreters.md
AhyoungRyu: This is what I can see for this image. The font in this image is even smaller than the other text around it.
screen shot 2017-03-19 at 10 47 49 pm

docs/manual/interpreters.md
AhyoungRyu: I assumed that the other readers may same with me. Couldn't get "except fields of paragraph"  has same meaning with "list of available parameters described below".

docs/manual/interpreters.md
tinkoff-dwh: The font on the following image less too

tinkoff-dwh: I agree, width should increment  to 800

AhyoungRyu:
@tinkoff-dwh Saw your latest update. As you did, 
```
###### Usage		
 For example in database exist user `user1` with some password and exists user with same name into Zeppelin. Configure the jdbc interpreter (postgres), set `default.user = #{user}`.
		
   1. Sign in (as `user1`)		
   2. Open interpreter		
   3. Execute `select current_user`
```

The above usage example can be added [JDBC interpreter docs](https://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/interpreter/jdbc.html). But it's okay we can update it in later time not in this PR. LGTM

.gitignore
jongyoul: If it's not related to this PR, can you please create new minor PR for handling it?

tinkoff-dwh: @jongyoul 
done

jongyoul: Thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
```diff
@@ -296,6 +302,40 @@ public Interpreter getInterpreterInTheSameSessionByClassName(String className) {
...
+          for (Field field : InterpreterContext.class.getDeclaredFields()) {
+            Class clazz = field.getType();
+            if (!skipFields.contains(field.getName()) && (typesToProcess.contains(clazz)
+                || clazz.isPrimitive())) {
+              try {
+                Object value = FieldUtils.readField(field, interpreterContext, true);
+                p = p.replaceAll(String.format(markerTemplate, field.getName()),
+                    value != null ? value.toString() : StringUtils.EMPTY);
+              } catch (Exception e) {
+                logger.error("Cannot replace context parameter", e);
```
felixcheung: should it stop or continue with `#{name}` in it if it couldn't find a match?

tinkoff-dwh: if field with name `name` exists in context then replaced in any case

tinkoff-dwh: if field with name `name` exists in context then replaced in any case

felixcheung: could you add a test for that new case then?

masyan: test testPropertyWithReplacedContextFields  covers this case, `replName` in context is NULL and replaced with an empty string

tinkoff-dwh: test testPropertyWithReplacedContextFields covers this case, replName in context is NULL and replaced with an empty string

felixcheung:
Ok.

Any other comment? Merging if no more discussion.

PR #2088, opened 2017-03-02
[ZEPPELIN-2203][WIP] In case of 'spark2' default interpreter, both '%spark' and '%spark2' point to Spark2.

prabhjyotsingh:
### What is this PR for?
In case of 'spark2' default interpreter, both '%spark' and '%spark2' point to Spark2.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2203](https://issues.apache.org/jira/browse/ZEPPELIN-2203)

Create two interpreters with name spark(with spark 1.6) and spark2 (with spark 2.0). Now set default interpreter as spark2 for a notebook. Now execute "%spark sc.version" notice it prints spark2 version

Before
zeppelin-2203_before

After
zeppelin-2203_after

### Questions:


zjffdu:
@prabhjyotsingh Can you explain the logic of how we parse `%spark` ?

PR #2100, opened 2017-03-06
[ZEPPELIN-2060] Make dynamic select form turn on or off using checkbox

AhyoungRyu:
### What is this PR for?
I added "Auto Run" checkbox for select dynamic form to make user turn on / off automatic running after the form value changed.

### What type of PR is it?
Improvement

### Todos
* [x] - update docs after getting feedback

### What is the Jira issue?
[ZEPPELIN-2060](https://issues.apache.org/jira/browse/ZEPPELIN-2060)

1. Apply this patch and run Zeppelin web as dev mode
```bash
# under zeppelin-web
$ yarn run dev
```

2. Go to Spark tutorial note and try to change value in select form as below screenshot imgs 

 - "Auto Run" checkbox button will be shown only in select dynamic form's dropdown menu
record1

 - _**turn on**_ "Auto Run"-> auto run right after the value changed / _**turn off**_ -> need to press `Enter`
record2

### Questions:
* Does this needs documentation? yes, maybe [this part](https://github.com/apache/zeppelin/blob/master/docs/manual/dynamicform.md#select-form)


AhyoungRyu:
@Leemoonsoo @felixcheung I think this way will be better than #2072. Can I ask review this PR?

Leemoonsoo:
@AhyoungRyu Tested and it works well.

To me, label "Auto Run" is bit hard to guess that it is related with dynamic form. Can you suggest another label which helps user better understanding?

And do you think you can add test for it?

AhyoungRyu:
@Leemoonsoo Thanks for review! 
Regarding the test, sure. i'll add the test to [ParagraphActionsIT.java](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java).
And let me think about more about intuitive label for this. I think we also need to let them know even if they uncheck this option, still can run the dynamic form by pressing `Enter`. Anyway I will update again :)

felixcheung:
I like this idea - could be hard to explain in a few words though :)

AhyoungRyu:
@Leemoonsoo @felixcheung Sorry for my late update. To provide more intuitive label, I changed it from `Auto run` -> `Run on selection change` and added tooltip msg to let user know they still can run the paragraph by pressing enter even if they uncheck the checkbox. 
screen shot 2017-03-14 at 10 52 22 pm

If it's okay, then I'll add the test and update the docs.

Leemoonsoo:
@AhyoungRyu Sounds good!

AhyoungRyu:
@Leemoonsoo I added two test cases as you said in this commit: 66c6c21d1bb435f5ade327a8f68cf3e8897e1fb3
 - Check "Run on selection change" menu appears or not, when select form is created 
 - Check `isAutoRunTrue` value has been changed to `false` or not, if "Run on selection change" is unchecked

And also added below content to [this part](https://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/manual/dynamicform.html#select-form) in `docs/manual/dynamicform.md`.
screen shot 2017-03-15 at 4 05 27 pm

AhyoungRyu:
CI is also green now :D

Leemoonsoo:
LGTM.

@AhyoungRyu Do you mind improve unittest to not only verify button is checked, but also verify actual dynamic form behavior, in this PR or later in the other PR.

Leemoonsoo:
Ah i realized that "Run on selection change" is not persisted, so refreshing browser resets the option. @AhyoungRyu Could you take care?

AhyoungRyu:
@Leemoonsoo Didn't realize that. Thanks for letting me know.  Let me update again.

AhyoungRyu:
@Leemoonsoo I made `runOnSelectionChange` can be persisted under `note.json`'s `config` field like below.
screen shot 2017-03-16 at 12 56 14 pm

 If user tries to `check`/`uncheck` the option in dropdown menu, this value will be updated. So reloading / Zeppelin restart won't be affect to this value change. 
What do you think? 

And regarding 
>Do you mind improve unittest to not only verify button is checked, but also verify actual dynamic form behavior, in this PR or later in the other PR?

If it's okay, I'll add the test case for this changed behaviour in another PR!

AhyoungRyu:
CI failed when it runs `yarn run test` and it seems relevant with this change. Let me update again.

AhyoungRyu:
CI is green now. Ready to review again.

Leemoonsoo:
Tested and LGTM

AhyoungRyu:
@Leemoonsoo Thanks for reveiw.
Merge into branch-0.7 and master if there are no more comments

PR #2107, opened 2017-03-07
[ZEPPELIN-2149] Each interpreter should have a seperate log file

prabhjyotsingh:
### What is this PR for?
In the current implementation, both spark and spark2 uses spark interpreter, both phoenix and hive use jdbc interpreter log file.
Each interpreter should have separate log file. The same log file for multiple interpreters confuses the user and complicated debugging.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-2149](https://issues.apache.org/jira/browse/ZEPPELIN-2149)

Create multiple interpreters (say shell interpreter) with different names (say sh1, sh2, etc.).
Now on running these different interpreters should log to different files.

### Questions:


bin/interpreter.sh
```diff
@@ -86,6 +89,9 @@ ZEPPELIN_SERVER=org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer
 INTERPRETER_ID=$(basename "${INTERPRETER_DIR}")
 ZEPPELIN_PID="${ZEPPELIN_PID_DIR}/zeppelin-interpreter-${INTERPRETER_ID}-${ZEPPELIN_IDENT_STRING}-${HOSTNAME}.pid"
 ZEPPELIN_LOGFILE="${ZEPPELIN_LOG_DIR}/zeppelin-interpreter-"
+if [[ ! -z "$INTERPRETER_NAME" ]]; then
+    ZEPPELIN_LOGFILE+="${INTERPRETER_NAME}-"
+fi
```
zjffdu: `INTERPRETER_NAME` and `INTERPRETER_ID` are used together ? I assume the spark1 log file should be something like `zeppelin-interpreter-spark1-jzhang-jzhangMBPr.local.log` and spark2 log is  `zeppelin-interpreter-spark2-jzhang-jzhangMBPr.local.log`

zjffdu: I am thinking that we could pass interpreter name as environment variable `INTERPRETER_DIR` to the interpreter process, so that we don't need to change `interpreter.sh`

prabhjyotsingh: Yes, even I initially thought about removing `INTERPRETER_ID` but then no harm in it, and I thought about the case where the user would want to `tail -f *jdbc*` and all rest of the interpreters as HIVE, PHOENIX or POSTGRES, etc.

r-kamath:
Tested. LGTM.

bin/interpreter.sh
prabhjyotsingh: @zjffdu  Any thoughts on this ?

zjffdu: @prabhjyotsingh I verified the PR, seems the log file name format is `zeppelin-interpreter-<interpeter_name>-<interpreter_id>-<host>.log`. But I still think we don't need to differentiate `interpreter_name` and `interpreter_id` as `interperter_name` should already been unique, `zeppelin-interpreter-<interpeter_name>-<host>.log` should be enough. The scenario you mentioned about jdbc,  I don't think user would use this command `tail -f *jdbc*` to read all the jdbc logs.   
Let's hear more comments from others @felixcheung @Leemoonsoo @jongyoul 
BTW, we need to document for this PR.

zjffdu: Also we may also need to consider to include username into log file name if isolation is enabled.

prabhjyotsingh: Yes, username does get included if impersonation in is enabled https://github.com/apache/zeppelin/blob/master/bin/interpreter.sh#L90.

felixcheung:
Is `name` the right name for this? I seem to recall we call it something else in the code for the `%thing`

prabhjyotsingh:
I picked up "Interpreter Name" from UI while creating new interpreter, let me rename this to InterpreterGroup.

prabhjyotsingh:
Any thoughts on this ?

prabhjyotsingh:

PR #2092, opened 2017-03-03
[ZEPPELIN-2148] On creation of Bar graph zeppelin UI shows it as minigraph - 2

prabhjyotsingh:
### What is this PR for?
This is based on https://github.com/apache/zeppelin/pull/2074#issuecomment-283843330. 

In this I've reverted #47a106a and applied, that.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/
* Put link here, and add [ZEPPELIN-*Jira number*] in PR title, eg. [ZEPPELIN-533]

### What is the Jira issue?
* [ZEPPELIN-2148](https://issues.apache.org/jira/browse/ZEPPELIN-2148)

### How should this be tested?
Check screen shot.

### Screenshots (if appropriate)
Before:
zeppelin-2148-before

After:
zeppelin-2148-after

### Questions:
* Does the licenses files need update?
* Is there breaking changes for older versions?
* Does this needs documentation?

PR #2110, opened 2017-03-08
[ZEPPELIN-2234] Can't display the same chart again (master)

1ambda:
### What is this PR for?

Can't display the same chart again. I attached a screenshot.

- This should be backported into 0.7.0 as well.

#### Implementation Details

After https://github.com/apache/zeppelin/pull/2092, 

- result.html will draw chart every time since we use `ng-if` instead of `ng-show`
- that means DOM is deleted, and created too
- so we have to create visualization instance every time which requires a newly created DOM.

```js
builtInViz.instance = new Visualization(loadedElem, config); // `loadedElem` is the newly created DOM.
```

### What type of PR is it?
[Bug Fix]

### Todos

NONE

### What is the Jira issue?

I attached a screenshot

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -266,19 +266,23 @@ function ResultCtrl($scope, $rootScope, $route, $window, $routeParams, $location
   };
 
   $scope.renderDefaultDisplay = function(targetElemId, type, data, refresh) {
-    if (type === DefaultDisplayType.TABLE) {
-      $scope.renderGraph(targetElemId, $scope.graphMode, refresh);
-    } else if (type === DefaultDisplayType.HTML) {
-      renderHtml(targetElemId, data);
-    } else if (type === DefaultDisplayType.ANGULAR) {
-      renderAngular(targetElemId, data);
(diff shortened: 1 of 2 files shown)


soralee:
Let me test master and branch-0.7 out!

Leemoonsoo:
Tested and it fixes the problem.

However, since the change #2092, visualization is re-created everytime when switching. I'd suggest find a way restore `ng-show` and address problem ZEPPELIN-2148 in different way. So we can leverage activate() deactivate() function. Otherwise it is bit annoying especially with visualization that takes long time to create/initialize.

1ambda:
@Leemoonsoo Thanks for review.

Then I will revert 1bdab41 and #2092 while trying to use `ng-show` instead of.

1ambda:
I'v just updated code. @Leemoonsoo Could you review again? Thanks!

Leemoonsoo:
Tested and LGTM!

prabhjyotsingh:
@1ambda Thank you for fixing this.
Tested on local, works well. LGTM.

Leemoonsoo:

1ambda:
Thanks @Leemoonsoo @prabhjyotsingh!

PR #2120, opened 2017-03-10
[ZEPPELIN-1492] fixing the issue where updating a paragraph was not propagated correctly

FRosner:
### What is this PR for?

This pull request fixes two issues regarding paragraphs not being updated and therefore overwritten unintentionally. The first issue yields to local changes being overwritten when remote changes are made. The second issue yields to changes being overwritten when, e.g., the notebook is renamed.

The first change happens in the `updateParagraph` broadcast event handler function. This function has the purpose to update the local state of the paragraph in the paragraph controller scope when there is an update from the web socket.

However, it did not update the state if the only thing that has changed was the text. Now it will, which fixes the original issue in the issue description. This was one of the issues identified by https://issues.apache.org/jira/browse/ZEPPELIN-1492?focusedCommentId=15744928&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15744928

The second issue is fixed by saving the paragraph when the editor loses focus.

### What type of PR is it?

Bug Fix

### What is the Jira issue?

- https://issues.apache.org/jira/browse/ZEPPELIN-1492

The first issue can be reproduced with the following two browser-windows and a single notebook.

Browser A:

```
%pyspark
print "Original Zeppelin Notebook"
```

In Browser B, edit the notebook and the above command to:

```
%pyspark
print "Notebook is updated...."
```

If I run the notebook via browser A followed by Browser B, everything is updated nicely. Now also if I you add the following line to the Notebook though Browser B:

```
print "....once again"
```

and run the notebook through Browser B again, the content in Browser A will be updated.

The second issue can be reproduced by editing a cell without executing it and renaming the notebook right afterwards. The rename will reset your cell to the previous state. With the fix, your state is saved.

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1120,7 +1120,7 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
   function isUpdateRequired(oldPara, newPara) {
     return (newPara.id === oldPara.id &&
       (newPara.dateCreated !== oldPara.dateCreated ||
-      newPara.paragraph.text !== oldPara.paragraph.text ||
+      newPara.text !== oldPara.text ||
       newPara.dateFinished !== oldPara.dateFinished ||
       newPara.dateStarted !== oldPara.dateStarted ||
       newPara.dateUpdated !== oldPara.dateUpdated ||
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -681,6 +681,7 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
 
       $scope.editor.on('blur', function() {
         handleFocus(false);
+        $scope.saveParagraph($scope.paragraph);
       });
 
       $scope.editor.on('paste', function(e) {
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -1120,6 +1120,7 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
   function isUpdateRequired(oldPara, newPara) {
     return (newPara.id === oldPara.id &&
       (newPara.dateCreated !== oldPara.dateCreated ||
+      newPara.paragraph.text !== oldPara.paragraph.text ||
       newPara.dateFinished !== oldPara.dateFinished ||
       newPara.dateStarted !== oldPara.dateStarted ||
       newPara.dateUpdated !== oldPara.dateUpdated ||


1ambda:
Hi @FRosner. Thanks for contribution.

- Zeppelin build system relies on travis. So please setup your travis account and travis project for Zeppelin. 
- https://zeppelin.apache.org/contribution/contributions.html#continuous-integration

FRosner:
Will do 👍

FRosner:
@1ambda it's building fine at travis: https://travis-ci.org/FRosner/incubator-zeppelin

Anything else I need to do? Why is Jenkins complaining?

Leemoonsoo:
@FRosner Jenkins checks `https://travis-ci.org/FRosner/zeppelin` instead of `https://travis-ci.org/FRosner/incubator-zeppelin` fyi.

FRosner:
Do you want me to recreate the fork? I forked it long time ago :D
-- 
Sent from my phone. Please excuse my brevity.

On 03/13/2017, 16:05 Lee moon soo <notifications@github.com> wrote:

@FRosner Jenkins checks https://travis-ci.org/FRosner/zeppelin instead of https://travis-ci.org/FRosner/incubator-zeppelin fyi.

Leemoonsoo:
@FRosner Yes, that would be appreciated.

1ambda:
Thanks for fixing bug. I will review and give you feedback soon.

FRosner:
Anything I can do to support, @1ambda?

We deployed this patch to ~ 5 users to try it out and it is working fine. Are there any problems you can think of?

Thanks for your time and effort :)

1ambda:
@FRosner Sorry for late reply. 😭... 

### I can't reproduce the original issue reported in 1492 in current master

Usually, paragraph will be updated If interpreter updates paragraph object properly for example `dataFinished`. 
I am not sure there is an case when **other paragraph fields are exactly same and only `paragraph.text` is different**. Only in this case, paragraph will be not updated in current implementation. 

```js
// https://github.com/apache/zeppelin/blob/f43d27f0bd9f841d71450ae303dc6f9f671bae0a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js#L1123

  newPara.dateFinished !== oldPara.dateFinished || 
```

But `newPara.text !== oldPara.text ||` is ok it's more accurate condition on if stmt. 

1492-cant-reproduce

### I can't reproduce focus issue too

in 1492, @FRosner said that 

> 1. Enter code in a cell
> 2. Remove the focus to rename the notebook
> 3. The code gets reset

But I can't reproduce in current master.

1492-focus

### Save paragraphs automatically when focus is lost.

So this is more about **save paragraph when focus is lost** instead of resolving the problem submitted in 1492.

```js
$scope.editor.on('blur', function() {
          handleFocus(false);		          handleFocus(false);
 +        $scope.saveParagraph($scope.paragraph);
        });		        });
```

In current implementation Zeppelin will save when user refresh browsers. But this PR will cause more frequent `paragraph.text` overwriting.  
I don't have strong opinion about the behavior, so I would like to other people's opinions. 

@Leemoonsoo @AhyoungRyu What do you think of?

sergeymazin:
Issue we are facing is when somebody else is adding/renaming/deleting any other notebook, unsaved text in current notebook is lost.

Renaming any other notebook resets current notebook:
notebook_rename
Deleting any other notebook resets current notebook:
notebook_delete

Same comes with adding/importing new notebooks. This makes Zeppelin unusable when working in multitenant mode once somebody starts importing/renaming notebooks.

1ambda:
@sergeymazin Thanks for quick reply! 

I see, 

- that's because deleting, renaming or doing other notebook scale actions cause broad casting all notebooks. 
- then, all notebooks will get websocket messages that contains persisted contents (including `paragraph.text`)

I will test and comment again in few hours.

1ambda:
@FRosner Could you attach a GIF for

```
Enter code in a cell
Remove the focus to rename the notebook
The code gets reset
```

FRosner:
Here you go @1ambda 

zeppelin-1492-1

1ambda:
@FRosner 

I'v fixed 2 issues without changing the default behavior of focus.
Please help review #2176, #2177.

FRosner:
Looking at the discussion and the different issues I see what you mean @1ambda. We are actually having two issues:

1. Local changes are overwritten when the notebook gets updates from the server (e.g. someone renames a note)
2. Local changes are not propagated on a regular basis (and you risk bigger conflicts when editing simultaneously)

The first point is addressed by #2176 #2177 more than this PR, because even when saving on losing focus you might risk losing local changes if you didn't lose the focus and someone renames, correct?

The second point is addressed in a way here but as mentioned in one of the other discussions we should really move towards real time updates where every keystroke gets transmitted (or at least while typing maybe after a few seconds) and you can see the cursor of the other users in your paragraph (similar to Google Docs).

Does it make sense?

1ambda:
@FRosner I see and agree with that we need more realtime notification on paragraph editing. 

LGTM

sergeymazin:
@1ambda thanks for working on this.
I guess you figured everything out with @FRosner in #2176 and #2177 already. Please let me know if I can help somehow.

FRosner:
Is there something I can do to accelerate the merge process here? Any open questions?

Leemoonsoo:
@FRosner Thanks for the fix. LGTM and merge to master and branch-0.7

PR #2128, opened 2017-03-13
[WIP] [Discuss] Make use of all grouped data to draw pie chart

sravan-s:
### What is this PR for?
Now, grouped pie charts only uses the data from the first group
With this fix- 
* Add data from all groups to variable d3g, so all groups could be rendered
* Rewrite for loop with map and concat
* Refactor some variables to const and let

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
*  [ZEPPELIN-2237](https://issues.apache.org/jira/browse/ZEPPELIN-2237)

* Create a built in pie chart visualization
* Select a column to group the data
* Should display the visualization based on all the available grouped data

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-piechart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-piechart.js
@@ -33,7 +33,7 @@ export default class PiechartVisualization extends Nvd3ChartVisualization {
   };
 
   render(pivot) {
-    var d3Data = this.d3DataFromPivot(
+    const d3Data = this.d3DataFromPivot(
       pivot.schema,
       pivot.rows,
       pivot.keys,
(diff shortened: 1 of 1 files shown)


sravan-s:
As per [ZEPPELIN-2237](https://issues.apache.org/jira/browse/ZEPPELIN-2237), in situations in which data is grouped, the pie chart only displays the data from first group. With this fix, data from all the groups would be displayed in the pie chart

## Discuss
I would prefer to see pie chart with broken slices: https://www.amcharts.com/demos/pie-chart-broken-slices/. Not sure if nvd3 supports this feature. I won't suggest bringing in a new charting library to implement such an idea. Thoughts...

sravan-s:
@Leemoonsoo @AhyoungRyu  Any comments on this?

Leemoonsoo:
LGTM and merge to master and branch-0.7 if no more comment
Thanks @sravan-s for the fix!

sravan-s:
@Leemoonsoo  Please review https://github.com/apache/zeppelin/pull/2132 . I came across it while working on this. I believe it should be hadled first

PR #2133, opened 2017-03-14
[ZEPPELIN-1357,1892,1370][Umbrella] Text overlap in the MultiBarChart

soralee:
### What is this PR for?
If using the `multiChartBar`, it often occurs overlap in text. So, this PR is for preventing to overlap in text. 

>~~The below is the case what I updated.~~
~~1. The xLabel generally show all of the text without rotation (less than 30).~~
~~2. If the xLabel size is over than 30 and less than 80, the xLabel text are rotated as 90 degree and displayed.~~
~~3. If the xLabel size is over than 80, the xLabel text are disappeared, but the tooltip is displayed.~~
>~~I have made improvements based on my thinking, so I would appreciate to give me your feedback.~~
~~And if merged this PR, I'll update the feature so that user can use `min` and `max` variables by user on web.~~

**[Update]**
To prevent overlap in xLabel text, in one way which is talked the below mentions, user who is using zeppelin could select to visualization type of xLabel such as `rotate 45 degree` or `hide`.

### What type of PR is it?
[Bug Fix | Improvement ]

### What is the Jira issue?
* [ZEPPELIN-1357; UI - label wrapping not done properly in charts](https://issues.apache.org/jira/browse/ZEPPELIN-1357)
* [ZEPPELIN-1892; Display label vertically or horizontally smartly](https://issues.apache.org/jira/browse/ZEPPELIN-1892)
* [ZEPPELIN-1370; Label overlaps - in default visualization example barchart](https://issues.apache.org/jira/browse/ZEPPELIN-1370)

1. Run the paragraph for bank data in  the `Basic Features (Spark)` notebook.
2. Execute the following query.
``` 
%sql
select * from bank
```
3. Use the pivot in the `Setting` toggle in the `MultiBarChart` such as the screenshots.

**[Before]**
z_zeppelin-1357_b

z_1357_b3_

**[After]**
peek 2017-04-17 14-58

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
@@ -45,6 +45,7 @@ export default class BarchartVisualization extends Nvd3ChartVisualization {
       true);
 
     super.render(d3Data);
+    this.config.changeXLabel(this.config.xLabelStatus);
   };
 
   /**
@@ -57,19 +58,99 @@ export default class BarchartVisualization extends Nvd3ChartVisualization {
 
   configureChart(chart) {
     var self = this;
+    var configObj = self.config;
+
     chart.yAxis.axisLabelDistance(50);
     chart.yAxis.tickFormat(function(d) {return self.yAxisTickFormat(d);});
 
-    this.chart.stacked(this.config.stacked);
+    self.chart.stacked(this.config.stacked);
 
-    var self = this;
+    self.config.changeXLabel = function(type) {
+      switch (type) {
+        case 'default':
+          self.chart._options['showXAxis'] = true;
+          self.chart._options['margin'] = {bottom: 50};
+          self.chart.xAxis.rotateLabels(0);
+          configObj.xLabelStatus = 'default';
+          break;
+        case 'rotate':
+          self.chart._options['showXAxis'] = true;
+          self.chart._options['margin'] = {bottom: 140};
+          self.chart.xAxis.rotateLabels(-45);
+          configObj.xLabelStatus = 'rotate';
+          break;
+        case 'hide':
+          self.chart._options['showXAxis'] = false;
+          self.chart._options['margin'] = {bottom: 50};
+          d3.select('#' + self.targetEl[0].id + '> svg').select('g.nv-axis.nv-x').selectAll('*').remove();
+          configObj.xLabelStatus = 'hide';
+          break;
+      }
+    };
+
+    self.config.isXLabelStatus = function(type) {
+      if (configObj.xLabelStatus === type) {
+        return true;
+      } else {
+        return false;
+      }
+    };
+
     this.chart.dispatch.on('stateChange', function(s) {
-      self.config.stacked = s.stacked;
+      configObj.stacked = s.stacked;
 
       // give some time to animation finish
       setTimeout(function() {
-        self.emitConfig(self.config);
+        self.emitConfig(configObj);
       }, 500);
     });
+  };
+
+
+
+  getSetting(chart) {
+    var self = this;
+    var configObj = self.config;
+
+    // default to visualize xLabel
+    if (typeof(configObj.xLabelStatus) === 'undefined') {
+      configObj.changeXLabel('default');
+    }
+
+    return {
+      template: `<div>
+          xAxis :
+      </div>
+
+      <div class='btn-group'>
+        <button type="button"
+              class="xLabel btn btn-default btn-sm"
+              ng-class="{'active' : this.config.isXLabelStatus('default')}"
+              ng-click="save('default')" >
+            Default
+        </button>
+
+        <button type="button"
+              class="btn btn-default btn-sm"
+              ng-class="{'active' : this.config.isXLabelStatus('rotate')}"
+              ng-click="save('rotate')" >
+            Rotate
+        </button>
+
+        <button type="button"
+              class="btn btn-default btn-sm"
+              ng-class="{'active' : this.config.isXLabelStatus('hide')}"
+              ng-click="save('hide')" >
+            Hide
+        </button>
+      </div>`,
+      scope: {
+        config: configObj,
+        save: function(type) {
+          configObj.changeXLabel(type);
+          self.emitConfig(configObj);
+        }
+      }
+    };
   };
 }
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/app/visualization/builtins/visualization-nvd3chart.js
```diff
@@ -37,14 +37,39 @@ export default class Nvd3ChartVisualization extends Visualization {
       this.chart = nv.models[type]();
     }
 
+    // prevent to overlap in text if using multiBarChart
+    // if size of xLabel are over maxRotate, the MutiBarChart don't show text
+    if (type === 'multiBarChart' && d3g[0]) {
```
Leemoonsoo: Could you try place this routine inside of https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/visualization/builtins/visualization-barchart.js#L58 ?

soralee: OK, I'll try do it! Thanks :slightly_smiling_face:

soralee:
@Leemoonsoo I really thanks for your feedback and your think! :blush: 
You mean, do you think it would be better to show `min` and `max` values ​​without `rotate` and `hide`?
Sounds good! Let's look at the opinions of others and change `min` and `max` instead of `rotate` and `hide`, if they want.

And here is my comment about your sayings and why I used the `rotate` and `hide`. 
>IMO, if bar chart automatically calculate and show label without overlap without user parameter that would be optimal.

Right! At first, I had also wanted that bar chart can be automatically calculate and show label without overlap and user parameter. However, because the nvd3 don't have These functions (or I might miss it), in the easy way, many users are using rotation for avoid overlap.
But, if there is a large amount of data, this rotation way also has a limit to showing label without overlap, so I handled by hiding the label.

Leemoonsoo:
> You mean, do you think it would be better to show min and max values ​​without rotate and hide?
Sounds good! Let's look at the opinions of others and change min and max instead of rotate and hide, if they want.

Sorry for confuse you. what i meant was more like select either `rotate` and `hide` is more easy to understand than set number for `min`, `max`.

soralee:
Thanks for quick response! 

> Sorry for confuse you. what i meant was more like select either rotate and hide is more easy to understand than set number for min, max.

Okay! I got it! 
Let's see others opinion :blush: 
I think `hide` is better! because this way can completely prevent overlap.

soralee:
I think the following JIRA issues are similar with this PR
* [ZEPPELIN-1892; Display label vertically or horizontally smartly](https://issues.apache.org/jira/browse/ZEPPELIN/ZEPPELIN-1892)
* [ZEPPELIN-1370; Label overlaps - in default visualization example barchart](https://issues.apache.org/jira/browse/ZEPPELIN-1370)

Leemoonsoo:
@soralee can you rebase or merge master and see if ci becomes green?

zeppelin-web/src/app/visualization/builtins/visualization-nvd3chart.js
soralee: I updated it!

soralee:
Thanks @Leemoonsoo! I did it and CI is green now

zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
```diff
@@ -57,6 +59,29 @@ export default class BarchartVisualization extends Nvd3ChartVisualization {
 
   configureChart(chart) {
     var self = this;
+    // prevent to overlap in text if using multiBarChart
+    // if size of xLabel are over maxRotate, the MutiBarChart don't show text
+    var minRotate = 30;
+    var maxRotate = 80;
+    var xLabelsLength = Object.keys(self.xLabels).length;
```
1ambda: MINOR: `self.xLabels` looks like an array, do we need `Object.keys` here?

Leemoonsoo:
I tested a bit more and found some cases.

1. Label rotated when unnecessary
image

2. Label still overlapping when paragraph width is small
image

To address these problem, I think rotation/hide should be calculated based on actual chart width and combined width of all labels.
Or just providing completely manual (provide option to select normal/rotation/hide) could be alternative.

@soralee what do you think? Do you have any other idea?

soralee:
@Leemoonsoo Thanks for review and suggestions!
I think manual options depending on users seems great so far, because I didn't find automatic calculation about width of `xLabel`. Let me dig this part more and leave a comment after updating.

zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
soralee: You're right, I missed when moving the code!
Let me check and update again, Thanks @1ambda

soralee: Hi, @1ambda 
I check data type of xLabels was object because `d3Data.xLabels` is object. 
so, I removed Line 26 that is not necessary code.

1ambda: great.

soralee:
@Leemoonsoo I updated description and code as that we talked about above comment. I didn't found how to be calculated based on actual width changed so, I updated manual button it can be used by user wants. 
Could you review again, please?

Leemoonsoo:
Tried last commit and it looks great.

However, in the beginning "Default" radio button is not selected automatically.
image

Also it'll be helpful to have a label for these radio buttons. Such as 'xaxis' or something.

soralee:
@Leemoonsoo Thanks for review!
Right! Label would be better :+1: 
Let me update again.

soralee:
@Leemoonsoo I added label and default variable furthermore, I changed from radio to button like this.
peek 2017-04-17 14-58

Leemoonsoo:
Great improvement @soralee. Tested and LGTM.

Leemoonsoo:
Ah, just one more thing. @soralee could you make CI green?

soralee:
@Leemoonsoo Yeah, this CI problem is related in #2255 
if merging #2255, I'll rebase this PR!

soralee:
@Leemoonsoo Ready to merge!

Leemoonsoo:
Thanks @soralee. merge to master.

PR #2134, opened 2017-03-14
[ZEPPELIN-2257] notification about incompleteness of data

tinkoff-dwh:
### What is this PR for?
Added notification about exceeding the limit in the result.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2257

1. Create table *test* with more than 2 records (if not exists)
2. Set parameter `common.max_count ` = 2
3. Execute 
```
%jdbc
select *from test
```
You should see message about exceeds limit

### Questions:


Leemoonsoo:
LGTM

Leemoonsoo:
Merge to master and branch-0.7 if no further discussions.

1ambda:
Sorry for late reply. 

- regarding to design, it would be great if we use bootstrap-alert. (`warning` level would be ok)

For example,

image

tinkoff-dwh:
ok.

1ambda:
LGTM

tinkoff-dwh:
CI green https://travis-ci.org/tinkoff-dwh/zeppelin/builds/211619211

AhyoungRyu:
@tinkoff-dwh Then you can make this CI status green by reopening-closing this PR :)

prabhjyotsingh:
Just saw it, this definitely looks better. I don't have a strong opinion, I just saw there is a small diff, hence, highlighting it.

This is how it gets displayed in spark/livy, with some extra information.
screen shot 2017-03-16 at 10 21 36 pm

And this is how it will look like if there were multiple queries.
screen shot 2017-03-16 at 10 24 54 pm

Leemoonsoo:
In addition to @prabhjyotsingh mentions, there are another type of message controlled by `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT` env variable.

image

All three messages Color, location, Decorations are different.
Basically these three messages are saying the same thing "Not a complete data".

So it make sense to display all three message in similar way, i think.

tinkoff-dwh:
@Leemoonsoo 
what final version, bootstrap?

Leemoonsoo:
#### Message location

Because of existing message `Results are limited by X.`  and `Output exceeds 10240. Truncated.` are displayed below the table. 
I think showing `Attention! Result is incomplete ...` below the table helps keep user experience more consistent.

#### Message style

Bootstrap decoration is pretty. If we apply bootstrap alert to all 3 messages, then that means existing 2 message ux is changed, too.

From taking less height of the screen, receiving less attention,
To taking more height of the screen, receiving more attention.

So which is the way we should go do you guys think? @tinkoff-dwh @1ambda 

#### Message text

All three message are different. 

`Results are limited by X.`
`Output exceeds 10240. Truncated.`
`Attention! Result is incomplete ...`

It'll be better use similar form of message. But how they're limited are all different. And it'll be great if we user give some link how they're limited.
So for example,

`Output is truncated to X rows. Learn more about common.max_count`
`Output is truncated to 10240 bytes. learn more about  ZEPPELIN_INTERPRETER_OUTPUT_LIMIT`


What do you think?

tinkoff-dwh:
@Leemoonsoo 
*location*
I think it's better to display above table because  the output can be large and  user will not see message.
*style*
Bootstreap attracts attention and this is important
*text*
Output is truncated to {count} rows. Learn more about {variable_name}



sorry for offtopic. @Leemoonsoo can you watch https://github.com/apache/zeppelin/pull/2085

tinkoff-dwh:
_jdbc
_jdbc_size
_livy
_livy_size
_pig
_pig_size
_spark
_spark_size

tinkoff-dwh:
CI green https://travis-ci.org/tinkoff-dwh/zeppelin/builds/213063178

tinkoff-dwh:
@Leemoonsoo 
I think now is fine (data table + message separated from footer), but I can increase the top margin

Leemoonsoo:
@tinkoff-dwh thanks! 

It's nit, but bottom margin is bit much compare to other margins in paragraph. i tried mark other margins.

image

Of course there's no rule that all the margin should be the same, but the margin on bottom looks bit inconsistent. Message is already decorated by bootstrap box so it already gives good visual separation.

What do you think?

tinkoff-dwh:
@Leemoonsoo 
possible. This margin within HTML Message maybe there are problems in other messages (without bootstrap). I'll try.

Leemoonsoo:
@tinkoff-dwh Ah, i see. i didn't see where this margin comes from.

AhyoungRyu:
Tested and basically agree with @Leemoonsoo. But one thing I just noticed is this new message can be seen in report mode. 

screen shot 2017-03-24 at 12 20 56 pm
I think it would be better we can hide this msg in this case.

masyan:
@AhyoungRyu 
if you configure the note and working only in report mode then you will not see when the report is incomplete.I can add the close button

tinkoff-dwh:
@AhyoungRyu
if you configure the note and working only in report mode then you will not see when the report is incomplete.I can add the close button

AhyoungRyu:
> I can add the close button

@tinkoff-dwh  Yeah that's better. Having close button will be better for other cases as well.

AhyoungRyu:
@tinkoff-dwh Thanks! Looks good :)

tinkoff-dwh:
@Leemoonsoo 
Yes you were right, this margin from alert (bootstreap class). fix it

tinkoff-dwh:
travis does not work (build was not created for last commit)

tinkoff-dwh:
ready to review

sotnich:
LGTM

Leemoonsoo:

Thanks @tinkoff-dwh for the improvement!

PR #2137, opened 2017-03-15
[ZEPPELIN-2260] Skip node,npm install and bundle when no helium package is selected

Leemoonsoo:
### What is this PR for?
Zeppelin 0.7.0 installs node and npm when it first starts for Helium package.
To be installed, or to failed to be installed due to network timeout, it takes some times.
We can just create empty file when no Helium package is enabled, instead of install npm and build bundle.
See discussion https://github.com/apache/zeppelin/pull/2095#issuecomment-285447619

### What type of PR is it?
Improvement

### Todos
* [x] - skip node,npm install and bundle package when no package is selected

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2260

### How should this be tested?
When no package is selected (e.g. right after clean install Zeppelin), npm and node is no longer installed on startup.

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #2095, opened 2017-03-05
[ZEPPELIN-2094] Decrease npm install retry time (for branch-0.7)

NohSeho:
### What is this PR for?
**This pr is for branch-0.7**
It’s too delayed for npm install when computer do not connected any networks.
Beacause when npm install, it has too long retry timeout.
This PR is to decrease retry timeout when npm install.
[pr for mater](https://github.com/apache/zeppelin/pull/2060) https://github.com/apache/zeppelin/pull/2095

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2094

### How should this be tested?
you must enable any one helium before test

Line 221 In zeppelin-zengine org.apache.zeppelin.helium.HeliumVisualizationFactory.java

First set with 
`String commandForNpmInstall = "install —loglevel=error”;`
and You don’t connect any ethernet or wireless internet.
build & run

and set with
```
String commandForNpmInstall =
     String.format("install --fetch-retries=%d --fetch-retry-factor=%d " +
                   "--fetch-retry-mintimeout=%d",
     FETCH_RETRY_COUNT, FETCH_RETRY_FACTOR_COUNT, FETCH_RETRY_MIN_TIMEOUT);
```
also don’t connect any networks, build & run.

and also line 345. do it `npm install some artifact` again like above.

WHY
retries = 2
factor = 1
mintimeout = 5(sec)?

npm use [retry](https://github.com/tim-kos/node-retry) module to retry.
It refers [this article](http://dthain.blogspot.kr/2009/02/exponential-backoff-in-distributed.html) for retry algorithms.
It is a math which structured _Math.min(Math.round(random * minTimeout * Math.pow(factor, attempt)), maxTimeout)_.
In retry source code, between two retries. First retry doesn't care _Math.min()_, just _Math.round(random * minTimeout * Math.pow(factor, attempt))_)

Description | Before | After
------- | ------- | -------
Condition | npm's default setting<br>random = False = 1<br>retry = 2<br>minTimeout = 10 (sec)<br>maxTimeout = 60 (sec)<br>factor = 10 | custom setting<br>random = False = 1<br>retry = 2<br>minTimeout = 5 (sec)<br>maxTimeout = 60 (sec)<br>factor = 1<br>
First retry | Math.round(1 * 10 (sec) * 10^1)) | Math.round(1 * 5 (sec) * 1^1))
First retry result (Approximately) | 100 (sec) | 5 (sec)
[...]

PR #2138, opened 2017-03-15
[ZEPPELIN-2260] [branch-0.7] Skip node,npm install and bundle when no helium package is selected 

Leemoonsoo:
### What is this PR for?
Apply #2137 to `branch-0.7`

Zeppelin 0.7.0 installs node and npm when it first starts for Helium package.
To be installed, or to failed to be installed due to network timeout, it takes some times.
We can just create empty file when no Helium package is enabled, instead of install npm and build bundle.
See discussion https://github.com/apache/zeppelin/pull/2095#issuecomment-285447619

### What type of PR is it?
Improvement

### Todos
* [x] - skip node,npm install and bundle package when no package is selected

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2260

When no package is selected (e.g. right after clean install Zeppelin), npm and node is no longer installed on startup.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumVisualizationFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumVisualizationFactory.java
@@ -51,6 +51,7 @@ public class HeliumVisualizationFactory {
   private File tabledataModulePath;
   private File visualizationModulePath;
   private Gson gson;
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
Merge to branch-0.7 if no further comments.

prabhjyotsingh:
LGTM! tested on local, now zeppelin (without local-repo) take little less time to restart.

PR #2139, opened 2017-03-15
ZEPPELIN-2261. Support to connect with livy through https

zjffdu:
### What is this PR for?
Livy server support https, but the currently livy interpreter doesn't support it. This PR is for for the supporting to connect with livy through https

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2261

Tested manually on livy server with ssl enabled. 

### Questions:

### DIFF
--- a/docs/interpreter/livy.md
+++ b/docs/interpreter/livy.md
@@ -130,6 +130,16 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
       <td></td>
       <td>Adding extra libraries to livy interpreter</td>
     </tr>
+  <tr>
+    <td>zeppelin.livy.ssl.trustStore</td>
+    <td></td>
+    <td>client trustStore file. Used when livy ssl is enabled</td>
+  </tr>
+  <tr>
+    <td>zeppelin.livy.ssl.trustStorePassword</td>
+    <td></td>
+    <td>password for trustStore file. Used when livy ssl is enabled</td>
+  </tr>  
 </table>
 
 **We remove livy.spark.master in zeppelin-0.7. Because we sugguest user to use livy 0.3 in zeppelin-0.7. And livy 0.3 don't allow to specify livy.spark.master, it enfornce yarn-cluster mode.**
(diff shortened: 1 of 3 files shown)


docs/interpreter/livy.md
```diff
@@ -130,6 +130,16 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
       <td></td>
       <td>Adding extra libraries to livy interpreter</td>
     </tr>
+  <tr>
+    <td>zeppelin.livy.ssl.trustStore</td>
+    <td></td>
```
felixcheung: could you add example of what they look like?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -378,13 +387,49 @@ private void cancelStatement(int statementId) throws LivyException {
     callRestAPI("/sessions/" + sessionInfo.id + "/statements/" + statementId + "/cancel", "POST");
   }
 
-  private RestTemplate getRestTemplate() {
+
+  private RestTemplate getRestTemplate() throws LivyException {
+    HttpClient httpClient = null;
+    if (livyURL.startsWith("https:")) {
+      String keystoreFile = property.getProperty("zeppelin.livy.ssl.trustStore");
+      String password = property.getProperty("zeppelin.livy.ssl.trustStorePassword");
```
felixcheung: should check they are not emply/whitespace before proceeding?

docs/interpreter/livy.md
zjffdu: It is just the path of truststore file, I explain it in Description column

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
zjffdu: Fixed, and also make another improvement to create `RestTemplate` in constructor. Otherwise each livy rest api would construct a new `RestTemplate`.

felixcheung:
is it possible to add tests for https?

zjffdu:
I am afraid not, ssh need keystore file which is host dependent.

felixcheung:
merging if no more comment

PR #2142, opened 2017-03-16
[ZEPPELIN-2124] Missing dependencies array in interpreter.json after upgrade from 0.6.2 to 0.7.0

minahlee:
### What is this PR for?
If there is no `dependencies` field specified in `interpreter.json`, front-end throws error because it tries to push new element to undefined variable. This PR fixes this issue by setting initial value of `dependencies` to empty array.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-2124](https://issues.apache.org/jira/browse/ZEPPELIN-2124)

Remove `dependencies` field from `conf/interpreter.json` and try to add new dependencies in http://localhost:8080/#/interpreter page.

### Questions:


jongyoul:
That's very simple :-)

PR #2145, opened 2017-03-16
[ZEPPELIN-2271] encoding password of credentials

tinkoff-dwh:
### What is this PR for?
Added a simple encoding of the password in the file (the credentials.json). Application administrator (Zeppelin) which has access to the file should not see user passwords at least as plain text.

### What type of PR is it?
Improvement 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2271

Remove credentials.json (if exist), before star Zeppelin.
Add credentials:
  Entity: test.test
  Username: user
  Password: passwd

Check content of file **credentials.conf**, password not equals "passwd"

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/Credentials.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/Credentials.java
@@ -18,22 +18,30 @@
 package org.apache.zeppelin.user;
 
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.io.*;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
 import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


felixcheung:
I'm concerned that this would give people a false sense of security...

zjffdu:
I think zeppelin should provide a pluggable interface for password storing and encrypting. Zeppelin could provide its default implementation for that but allow for extension. Storing password in file is not a mature and long term solution to me.

jongyoul:
I agree with @felixcheung.I'm not sure what the benefit is.

dwhsys:
@felixcheung @jongyoul ,
I want to discuss the problem of "false sense of security". I hope that anyone who want to use "credentials feature" in enterprise solution will be aware of low security while storing passwords in file (even encrypted). And I agree with @zjffdu about temporality of this mechanism. But here we have one obvious benefit: we do not allow someone with good intentions to see credentials accidentally. So it may be enough reason to implement this PR. I'm not really sure in my rationales. Waiting for your reply.

herval:
I've implemented something slightly similar here, except I'm actually encrypting the entire file using AES, using a passphrase you provide (which can live in memory)

it's not ideal, but as others stated, it's better than no protection at all.

thoughts?

tinkoff-dwh:
@herval 
link?

herval:
https://github.com/herval/zeppelin/tree/encrypt-credentials

PR #2151, opened 2017-03-17
[ZEPPELIN-465] Be able to run paragraph and the following ones

Remilito:
### What is this PR for?
Allow to run a paragraph **and** the ones following. This is useful for instance if you are generating a dataset in a paragraph and then have others subsequent paragraphs relying on this dataset.
In a nutshell this PR contains:
* UI: added a command in paragraph dropdown menu
* REST: added an boolean option to runparagraph (sync & async)
* updated rest doc
* z-engine: added relevant function to Note
* also refacto some duplicated code here and there.

### What type of PR is it?
Feature

### Todos
* [ ] - Once committers agree on the visual impact, update relevant documentation.

### What is the Jira issue?
[ZEPPELIN-465](https://issues.apache.org/jira/browse/ZEPPELIN-465)

* UI: click on the "run and next ones" in the dropdown menu and verify subsequent paragraphs are also run, and not precedent ones.
* REST: post request to /api/notebook/[job|run]/$noteId/$paragraphId?subsequent=true

runsubsequent
confirmationdialog
paragraphrunningpending

### Questions:

### DIFF
--- a/docs/rest-api/rest-notebook.md
+++ b/docs/rest-api/rest-notebook.md
@@ -894,6 +894,10 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td> 500 </td>
     </tr>
     <tr>
+      <td> subsequent parameter (optionnal) </td>
+      <td> If set to true, all the paragraphs following the paragraph will be run as well.```http://[zeppelin-server]:[zeppelin-port]/api/notebook/job/[noteId]/[paragraphId]?subsequent=true```</td>
+    </tr>
+    <tr>
       <td> sample JSON input (optional, only needed when if you want to update dynamic form's value) </td>
       <td><pre>
 {
@@ -930,6 +934,10 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
     <tr>
       <td> Fail code</td>
       <td> 500 </td>
+    </tr>
+    <tr>
+      <td> subsequent parameter (optionnal) </td>
+      <td> If set to true, all the paragraphs following the paragraph will be run as well. Warning: if one of the paragraphs fails the API will return ERROR ```http://[zeppelin-server]:[zeppelin-port]/api/notebook/job/[noteId]/[paragraphId]?subsequent=true```</td>
     </tr>
     <tr>
       <td> sample JSON input (optional, only needed when if you want to update dynamic form's value) </td>
--- a/docs/rest-api/rest-notebook.md
+++ b/docs/rest-api/rest-notebook.md
@@ -894,11 +894,7 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td> 500 </td>
     </tr>
     <tr>
-      <td> subsequent parameter (optionnal) </td>
-      <td> If set to true, all the paragraphs following the paragraph will be run as well.```http://[zeppelin-server]:[zeppelin-port]/api/notebook/job/[noteId]/[paragraphId]?subsequent=true```</td>
-    </tr>
-    <tr>
-      <td> sample JSON input (optional, only needed when if you want to update dynamic form's value) </td>
+      <td> sample JSON input (optional, only needed to update dynamic form's value) </td>
       <td><pre>
 {
   "name": "name of new note",
@@ -913,7 +909,6 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td><pre>{"status": "OK"}</pre></td>
     </tr>
   </table>
-
 <br/>
 ### Run a paragraph synchronously
   <table class="table-configuration">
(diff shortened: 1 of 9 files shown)


jongyoul:
I think JobSever is more proper location to handle the job status and relationship. How do you think of it?

AhyoungRyu:
@Remilito Thanks for the useful feature. 
I think it'll be better to display only essential buttons in the paragraph. How about placing it under paragraph control dropdown menu instead? I mean in here 
screen shot 2017-03-20 at 1 59 43 pm

Remilito:
@AhyoungRyu : updated (and now building and passing tests :sweat_smile: ).
@jongyoul : I am not sure to fully understand what you mean. If you are talking "UI" IMO the need to run or not a sequence of paragraphs depends mostly on the paragraphs' contents (i.e. generation of a dataset used by following paragraphs), so I find it more convenient to trigger this directly in the paragraph view rather than putting it in the job view. WDYT?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -750,20 +762,31 @@ public Response runParagraphSynchronously(@PathParam("noteId") String noteId,
...
     }
+    InterpreterResult resultFollowers;
+    if ( subsequent ) {
+      List<Paragraph> paragraphList = note.getParagraphs();
+      for (Paragraph p : paragraphList.subList(paragraphList.indexOf(paragraph) + 1,
+                                               paragraphList.size())) {
+        if (p.getListener() == null) {
+          note.initializeJobListenerForParagraph(p);
+        }
+        p.run();
```
Leemoonsoo: calling `run()` from Paragraph directly should be avoided. Paragraph should always run through scheduler provided by interpreter. i.e `note.run(p.getId())`. 

But currently, there's no method that runs paragraph synchronously in class Note. This PR doesn't need to address, but later it's better implement method such as `note.runSynchronous(p.getId())` which submit job into scheduler one by one, instead of directly run Paragraph.run().

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -319,6 +319,30 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
       paragraphText, $scope.paragraph.config, $scope.paragraph.settings.params);
   };
 
+  $scope.runSubsequentParagraphs = function(note,paragraph) {
+    if (!(paragraph.status in ['RUNNING','PENDING']) && paragraph.config.enabled){
+      BootstrapDialog.confirm({
```
Leemoonsoo: I personally prefer skip confirmation dialog especially when runSubsequentParagraphs() is called from keyboard shortcut. What do you think? 

And after runSubsequentParagraphs() ran, paragraph lost focus, so need to use mouse again and to focus paragraph.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
Remilito: let's keep this for another PR.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
Remilito: agreed, I'll remove the dialog when it is triggered from the keyboard shortcut.

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -133,6 +133,12 @@
             <span class="shortcut-keys">Ctrl+{{ isMac ? 'Option' : 'Alt'}}+m</span></a>
         </li>
         <li>
+          <a  ng-click="runSubsequentParagraphs(parentNote,paragraph)"
+              ng-show="paragraph.status!='RUNNING' && paragraph.status!='PENDING' && paragraph.config.enabled">
+            <span class="icon-control-end shortcut-icon" tooltip="Run this paragraph and the next ones"></span>Run with next ones
```
AhyoungRyu: Not sure other users can notice "Run with next ones" means "Run with following all paragraphs". Can we use more intuitive name for this?

docs/rest-api/rest-notebook.md
```diff
@@ -894,6 +894,10 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td> 500 </td>
     </tr>
     <tr>
+      <td> subsequent parameter (optionnal) </td>
```
AhyoungRyu: optional? :)

AhyoungRyu:
@Remilito Tested and it works as expected. 
Regarding docs, how about adding new section for this feature instead of `subsequent parameter (optional)` as a subsection under "Run a paragraph synchronously" & "Run a paragraph asynchronously" ? I saw many ppl have wanted this feature, so it'll be better to create new section for it. Below each "Run a paragraph synchronously" & "Run a paragraph asynchronously" section would be good place i think. What do you think?

Remilito:
@AhyoungRyu sorry for the late answer, I have updated the doc accordingly to your advice, and changed the wording of the UI command. It is now "Run with successors", but if a native English speaker has a better wording, any help is appreciated.

Leemoonsoo:
LGTM

zjffdu:
Here's one related PR #2195 which do it in `ZeppelinContext`, I hope we can merge these 2 tickets to use the same approach.

soralee:
Hi,@Remilito. I think now you needed a doing rebase.
Could you rebase it, please?

PR #2158, opened 2017-03-19
[ZEPPELIN-2279] excluded comments from SQL

tinkoff-dwh:
### What is this PR for?
Exclusion comments (single-, multiline) from queries before execution. Comments don't need to execute  query and sometimes there are errors. 

### What type of PR is it?
Bug Fix | Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2279

```
/* ; */
select 1;
-- text select 1
/* bla 
bla
bla*/
select 1; -- text
```

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -506,7 +506,7 @@ public class JDBCInterpreter extends Interpreter {
   protected ArrayList<String> splitSqlQueries(String sql) {
     ArrayList<String> queries = new ArrayList<>();
     StringBuilder query = new StringBuilder();
-    Character character;
+    char character;
 
     Boolean antiSlash = false;
     Boolean multiLineComment = false;
@@ -517,8 +517,8 @@ public class JDBCInterpreter extends Interpreter {
     for (int item = 0; item < sql.length(); item++) {
       character = sql.charAt(item);
 
-      if ((singleLineComment && (character.equals('\n') || item == sql.length() - 1))
-          || (multiLineComment && character.equals('/') && sql.charAt(item - 1) == '*')) {
+      if ((singleLineComment && (character == '\n' || item == sql.length() - 1))
+          || (multiLineComment && character == '/' && sql.charAt(item - 1) == '*')) {
         singleLineComment = false;
         multiLineComment = false;
         if (item == sql.length() - 1 && query.length() > 0) {
@@ -531,11 +531,11 @@ public class JDBCInterpreter extends Interpreter {
         continue;
       }
 
-      if (character.equals('\\')) {
+      if (character == '\\') {
         antiSlash = true;
       }
 
-      if (character.equals('\'')) {
+      if (character == '\'') {
         if (antiSlash) {
           antiSlash = false;
         } else if (quoteString) {
@@ -545,7 +545,7 @@ public class JDBCInterpreter extends Interpreter {
         }
       }
 
-      if (character.equals('"')) {
+      if (character == '"') {
         if (antiSlash) {
           antiSlash = false;
         } else if (doubleQuoteString) {
@@ -557,18 +557,18 @@ public class JDBCInterpreter extends Interpreter {
 
       if (!quoteString && !doubleQuoteString && !multiLineComment && !singleLineComment
           && sql.length() > item + 1) {
-        if (character.equals('-') && sql.charAt(item + 1) == '-') {
+        if (character == '-' && sql.charAt(item + 1) == '-') {
           singleLineComment = true;
           continue;
         }
 
-        if (character.equals('/') && sql.charAt(item + 1) == '*') {
+        if (character == '/' && sql.charAt(item + 1) == '*') {
           multiLineComment = true;
           continue;
         }
       }
 
-      if (character.equals(';') && !antiSlash && !quoteString && !doubleQuoteString) {
+      if (character == ';' && !antiSlash && !quoteString && !doubleQuoteString) {
         queries.add(StringUtils.trim(query.toString()));
         query = new StringBuilder();
       } else if (item == sql.length() - 1) {
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -585,12 +585,10 @@ public class JDBCInterpreter extends Interpreter {
   private void executePrecode(Connection connection, String propertyKey) throws SQLException {
     String precode = getProperty(String.format(PRECODE_KEY_TEMPLATE, propertyKey));
     if (StringUtils.isNotBlank(precode)) {
-      List<String> precodeQueries = splitSqlQueries(precode);
+      precode = StringUtils.trim(precode);
+      logger.info("Run SQL precode '{}'", precode);
       try (Statement statement = connection.createStatement()) {
-        for (String precodeQuery : precodeQueries) {
-          logger.info("Run SQL precode '{}'", precodeQuery);
-          statement.execute(precodeQuery);
-        }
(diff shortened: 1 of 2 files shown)


tinkoff-dwh:
longest job is interrupted because of the limit in travis-ci (50 minutes) -> CI red

Leemoonsoo:
@tinkoff-dwh Thanks for contribution.

CI failure on exceeding timelimit has fixed on master branch. Can you try rebase this branch and see if CI becomes green?

tinkoff-dwh:
@Leemoonsoo 
done

Leemoonsoo:
Looks like somehow this branch includes commits not part of this contribution.
image

If you rebase / merge correctly, they'll not be seen as commit of this PR. Could you reset to 6db3c46
 and try again?

tinkoff-dwh:
@Leemoonsoo 
i rebase from master. because if merge then CI red (jenking ignore merge commits).

tinkoff-dwh:
@Leemoonsoo 
fix it.
CI green, https://travis-ci.org/tinkoff-dwh/zeppelin/builds/213118918

Leemoonsoo:
I fixed Jenkins build command the problem taking wrong commit hash when it's got merge commit.
Now Jenkins will take correct commits. @tinkoff-dwh Could you try close / re-open this PR and see if Jenkins check the travis correctly?

felixcheung:
could you elaborate on what errors would come from comment in SQL?
generally we should avoid trying to pre-parse the paragraph text since we don't necessarily know the SQL dialect the interpreter can handle, or not

tinkoff-dwh:
@felixcheung 
example in jira

tinkoff-dwh:
@Leemoonsoo 
it works, thx!

Leemoonsoo:
LGTM
Merge to master if no further comments.

felixcheung:
Actually I want to clarify...
Do you know why the specific example fails? The Postgres spec is pretty clear that it is an accept form of comment.


https://www.postgresql.org/docs/8.0/static/sql-syntax.html#SQL-SYNTAX-COMMENTS



Again my point is we can't possibly know what is and is not supported by all the possible JDBC sources - we should avoid hardcoding the logic here.

https://github.com/apache/zeppelin/pull/2158#issuecomment-287968469

masyan:
@felixcheung 
falls because ";" is delimiter query and after split generated two queries ("/* ", "*/ select ...")

I relied on the standards for SQL89, 92 ... if you're talking about nosql databases, the jdbc driver are implemented in the form of wrappers (SQL queries, mongodb, aerospike).

tinkoff-dwh:
@felixcheung
falls because ";" is delimiter query and after split generated two queries ("/* ", "*/ select ...")

I relied on the standards for SQL89, 92 ... if you're talking about nosql databases, the jdbc driver are implemented in the form of wrappers (SQL queries, mongodb, aerospike).

felixcheung:
that's the thing, each of these implementation handles comments differently.
for instance, solr JDBC shares the same spec as [presto](https://github.com/prestodb/presto/blob/master/presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4) (which is very common) - but it doesn't support empty comment `/**/` or multiline comment `/*\n*/`

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
```diff
@@ -448,4 +448,30 @@ public void testPrecodeWithAnotherPrefix() throws SQLException, IOException {
...
+    properties.setProperty("common.max_retry", "3");
+    properties.setProperty("default.driver", "org.h2.Driver");
+    properties.setProperty("default.url", getJdbcConnection());
+    properties.setProperty("default.user", "");
+    properties.setProperty("default.password", "");
+    JDBCInterpreter t = new JDBCInterpreter(properties);
+    t.open();
+
+    String sqlQuery = "/* ; */\n" +
+        "--select * from test_table\n" +
```
felixcheung: add a test for `-- a ; b`?

tinkoff-dwh:
@felixcheung 
I don't see a problem. 
1. /**/ will not be processed 
2. /*\n*/ will not be processed

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -537,26 +555,42 @@ private boolean isDDLCommand(int updatedCount, int columnCount) throws SQLExcept
...
+
     return queries;
   }
 
   private void executePrecode(Connection connection, String propertyKey) throws SQLException {
     String precode = getProperty(String.format(PRECODE_KEY_TEMPLATE, propertyKey));
     if (StringUtils.isNotBlank(precode)) {
-      precode = StringUtils.trim(precode);
-      logger.info("Run SQL precode '{}'", precode);
+      List<String> precodeQueries = splitSqlQueries(precode);
```
tinkoff-dwh: Yes, there is no need to

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
tinkoff-dwh: added

tinkoff-dwh:
Ready to review

Leemoonsoo:
LGTM \cc @felixcheung

Leemoonsoo:

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -509,15 +509,32 @@ private boolean isDDLCommand(int updatedCount, int columnCount) throws SQLExcept
...
+    Boolean multiLineComment = false;
+    Boolean singleLineComment = false;
     Boolean quoteString = false;
     Boolean doubleQuoteString = false;
 
     for (int item = 0; item < sql.length(); item++) {
       character = sql.charAt(item);
 
+      if ((singleLineComment && (character.equals('\n') || item == sql.length() - 1))
+          || (multiLineComment && character.equals('/') && sql.charAt(item - 1) == '*')) {
```
felixcheung: why sometimes `character.equals()` and sometimes `charAt() ==`?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -537,16 +555,30 @@ private boolean isDDLCommand(int updatedCount, int columnCount) throws SQLExcept
...
+          && sql.length() > item + 1) {
+        if (character.equals('-') && sql.charAt(item + 1) == '-') {
+          singleLineComment = true;
+          continue;
+        }
+
+        if (character.equals('/') && sql.charAt(item + 1) == '*') {
+          multiLineComment = true;
+          continue;
+        }
```
felixcheung: what about
```
-- /* comment
select ...
```

is that a single line or multiline?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
tinkoff-dwh: item > 0 because first condition should  be `multiLineComment == true `

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
tinkoff-dwh: single because `singleLineComment == true` will not yet a symbol of the new string

` if (singleLineComment || multiLineComment) {
        continue;
      }`

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
tinkoff-dwh: conditions fixed

PR #2159, opened 2017-03-19
[HOTFIX] Handle removing interpreters while removing note

jongyoul:
### What is this PR for?
Removing some interpreters while removing note. In case of "Per note", "Isolated" interpreter process should be removed if a note be removed. In case of "Scoped" of "Per note", that session should be removed and if there's no more session in that process, the process should be removed, too.

### What type of PR is it?
[Hot Fix]

### Todos
* [x] - Handle removing interpreters when removing note

### What is the Jira issue?
N/A

1. set python interpreter as "per note/scoped"
1. run ```%python print(1)``` in a note
1. remove that note
1. check if python process exists with grep

### Questions:


jongyoul:
@Leemoonsoo Please review and merge it if it's ok.

Leemoonsoo:
Tested and it works perfectly.

jongyoul:
Will merge it

PR #2168, opened 2017-03-20
[ZEPPELIN-1889] [FIX] Multiple groups in chart doesn't work

sravan-s:
### What is this PR for?
Fix for multiple group in multi-bar chart

If any value is missing under any domain key in stacked multi-bar chart,
it won't get rendered. Fix this bug by adding `fillMissingValues` flag in
`d3DataFromPivot()`

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1889](https://issues.apache.org/jira/browse/ZEPPELIN-1889)

* Create a paragraph with tabular output
* Add multiple groups to input
* Render the stacked multi bar chart
* Chart should be rendered properly

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-barchart.js
@@ -41,7 +41,7 @@ export default class BarchartVisualization extends Nvd3ChartVisualization {
       pivot.groups,
       pivot.values,
       true,
-      false,
+      true,
       true);
 
     super.render(d3Data);


Leemoonsoo:
Thanks for working on the issue.

I have tested problem [ZEPPELIN-1889](https://issues.apache.org/jira/browse/ZEPPELIN-1889) in current master. However i couldn't reproduce the problem.

zeppelin-1889

@zjffdu @sravan-s Is problem described in ZEPPELIN-1889 is reproducible?

sravan-s:
@Leemoonsoo , the bug is limited only for 'stacked' version of bar chart. Notice the radio button inside the chart

* This is  the grouped version
grouped

* This is the buggy stacked version, in some cases, it won't even get rendered
stacked

* This is the fixed stacked version
stacked-fix

p.s. I tried recording gifs, but my laptop hangs(ubuntu 16.04, but I think it's the hardware that's the culprit)

Leemoonsoo:
Thanks @sravan-s for clarify.
Tested and it works well.

Merge to master and branch-0.7 if no further discussions

PR #2126, opened 2017-03-12
[gh-pages][ZEPPELIN-2209, 2110] Show 'INTERPRETER' type & Add pagination

AhyoungRyu:
### What is this PR for?
After #2037 merged, we can see `INTERPRETER` type in Helium menu. So for the consistency, it'll be better we can show`INTERPRETER` type in https://zeppelin.apache.org/helium_packages.html. So what I did in this PR is 
 - enabled to see `INTERPRETER` type 
 - added a pagination

### What type of PR is it?
Improvement

### Todos
* [x] - update overview sentences
(current overview sentences are only for npm package related description. But`INTERPRETER` type pkgs come from mvn central not npm registry. So need to update accordingly)

### What is the Jira issue?
* [ZEPPELIN-2110](https://issues.apache.org/jira/browse/ZEPPELIN-2110) : List community & 3rd party interpreter registered at Maven central repo in Helium menu 
* [ZEPPELIN-2209](https://issues.apache.org/jira/browse/ZEPPELIN-2209) : Add pagination

### How should this be tested?
1. To see `INTERPRETER` type pkg, replace [helium online registry uri](https://github.com/apache/zeppelin/blob/gh-pages/_includes/themes/zeppelin/default.html#L34) with `<script src="https://rawgit.com/AhyoungRyu/spark-notebook-example/master/helium-fake.js"></script>`
2. Run the website(`gh-pages`) locally as described in [here](https://github.com/apache/zeppelin/tree/gh-pages#run-website)

### Screenshots (if appropriate)
record

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? yes need to update overview sentences as I said in TODO

PR #2174, opened 2017-03-22
[ZEPELIN-2209] Add pagination to Helium menu

AhyoungRyu:
### What is this PR for?
Like I did in #2126, added pagination to Helium menu in Zeppelin. 
10 pkgs will be shown per 1 page. 

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-2209](https://issues.apache.org/jira/browse/ZEPPELIN-2209)

1. Replace temporarily [ZeppelinConfiguration.java#L46](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L46) with [this url](https://raw.githubusercontent.com/AhyoungRyu/spark-notebook-example/master/helium.json)

2. Build \w below command and restart Zeppelin server
```
 $ mvn clean package -DskipTests -pl 'zeppelin-zengine, zeppelin-server, zeppelin-interpreter'
```

3. Start web dev server under `ZEPPELIN_HOME/zeppelin-web` and browse `localhost:9000`
```
$ yarn run dev
```

4. Go to Helium menu -> Click `INTERPRETER` btn 
(Currently the other types of pkg are registered in online registry less than 10. That's why I request to change online registry URI in step 1)
5. Check the pagination works well or not

record

### Questions:

### DIFF
--- a/zeppelin-web/src/app/helium/helium.controller.js
+++ b/zeppelin-web/src/app/helium/helium.controller.js
@@ -37,9 +37,15 @@ export default function HeliumCtrl($scope, $rootScope, $sce,
     // get all package info and set config
     heliumService.getAllPackageInfoAndDefaultPackages()
       .then(({ pkgSearchResults, defaultPackages }) => {
+        // pagination
+        $scope.itemsPerPage = 10;
+        $scope.currentPage = 1;
+        $scope.maxSize = 5;
+        
         $scope.pkgSearchResults = pkgSearchResults;
         $scope.defaultPackages = defaultPackages;
-        classifyPkgType($scope.defaultPackages)
+        classifyPkgType($scope.defaultPackages);
+        
         return heliumService.getAllPackageConfigs()
       })
       .then(defaultPackageConfigs => {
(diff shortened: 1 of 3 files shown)


AhyoungRyu:
@1ambda Will be really appreciated if you can help review this!

1ambda:
@AhyoungRyu sure!

AhyoungRyu:
@1ambda Just noticed that `orderBy: publishedDate` doesn't work well properly now. Let me updated again and will let you know when it's resolved. (Actually same problem is in https://zeppelin.apache.org/helium_packages.html. So need to fix it as well..😅)

AhyoungRyu:
@1ambda I fixed it in the latest commit  bc7643c46dc6690b5b1f1a91249bcf4ee0e93a43. You can simply check whether it was fixed or not by comparing `INTERPRETER` type of packages are well located as version `0.7.0` -> `0.6.0` order. The `0.6.0` packages should be in 3rd page of `INTERPRETER` tab. (Before this commit, I could see some `0.6.0` of package in the middle of `INTERPRETER` package list.)

FYI 
>Actually same problem is in https://zeppelin.apache.org/helium_packages.html. So need to fix it as well 

I made a PR for this in #2180 :)

felixcheung:
could this auto scale by screen resolutions? I have seen a lot of cases (a lot of times in presentations/presenting in meeting where projector resolution is low)

1ambda:
@AhyoungRyu pagination DOM should be located outside of `ng-repeat` for helium containers.

screen shot 2017-03-27 at 3 41 25 pm

AhyoungRyu:
@1ambda It's wired. This is what I can see when I build my branch. 
record

There are two `ng-repeat` like below. And [the pagination DOM](https://github.com/AhyoungRyu/zeppelin/blob/7150e7cb596c2bbe9148ea2a09009c6fb273a20e/zeppelin-web/src/app/helium/helium.html#L195) is placed inside of the first `ng-repeat`(for pkg type checking loop) and outside of the second one(each type of package listing). 
screen shot 2017-03-28 at 12 10 18 am

FYI, the data is like this.
screen shot 2017-03-28 at 12 12 37 am

AhyoungRyu:
@felixcheung Let me check again for that part.

1ambda:
@AhyoungRyu Sorry for confusing. It works as described. maybe i mixed commits 😞 

- It would be nice if we hide `<<` and `>>` when users don't need it. (e.g less than 5 pages)

image

AhyoungRyu:
@1ambda No worries :) 
Then how about 1 page ? I mean if there is only 1 page, we can hide `<<` & `>>`.

1ambda:
@AhyoungRyu more precisely,

1. If we are in page 1, we don't need `<<`

image

2. If we are in some pages which can go to page 1, we need both `<<` && `>>`

image

3. the last case

image

AhyoungRyu:
@1ambda Thanks for the detailed explanation. 
record

1ambda:
@AhyoungRyu Yes, it would be great to hide unnecessary things

AhyoungRyu:
@1ambda I updated the pagination as you said and this is the current status. Will be appreciated if you can take a look again.
hide_first_last_arrows

@felixcheung I set the maximum pagination size as "5" and tested in the smallest mobile device size that Chrome dev tool provides (`320 x 568`) like below.
small_device
The pagination is not broken when the screen is small. But `enable` btn and package type selection btn are broken instead 😅 Let me handle them in another PR. What do you think?

felixcheung:
ok to me!

AhyoungRyu:
@felixcheung Thanks :)

1ambda:
@AhyoungRyu Tested and works as described. LGTM!

AhyoungRyu:
@1ambda Thanks! 
Will merge into master if there are no further discussions.

PR #2182, opened 2017-03-23
[HOLD][ZEPPELIN-2248] Fix javax.ws.rs and cxf version mismatch

jongyoul:
### What is this PR for?
Fixing version mismatch among jetty, cxf, jersey

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Match all versions of javax.ws related libraries

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2248

No error while starting and accessing the web at the first time

N/A

### Questions:

### DIFF
--- a/zeppelin-server/pom.xml
+++ b/zeppelin-server/pom.xml
@@ -34,11 +34,11 @@
 
   <properties>
     <!--library versions-->
-    <cxf.version>2.7.8</cxf.version>
+    <cxf.version>3.1.6</cxf.version>
     <hadoop-common.version>2.6.0</hadoop-common.version>
     <quartz.scheduler.version>2.2.1</quartz.scheduler.version>
-    <jersey.servlet.version>1.13</jersey.servlet.version>
-    <javax.ws.rsapi.version>2.0-m10</javax.ws.rsapi.version>
+    <jersey.servlet.version>2.25.1</jersey.servlet.version>
+    <javax.ws.rsapi.version>2.0.1</javax.ws.rsapi.version>
     <libpam4j.version>1.8</libpam4j.version>
     <jna.version>4.1.0</jna.version>
 
@@ -170,18 +170,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.cxf</groupId>
-      <artifactId>cxf-api</artifactId>
-      <version>${cxf.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.eclipse.jetty</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-
-    <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-webapp</artifactId>
       <version>${jetty.version}</version>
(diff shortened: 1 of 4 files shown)


zeppelin-server/pom.xml
```diff
@@ -292,10 +280,15 @@
       <version>${quartz.scheduler.version}</version>
     </dependency>
 
-    <dependency>
+    <!--<dependency>
       <groupId>com.sun.jersey</groupId>
       <artifactId>jersey-servlet</artifactId>
       <version>${jersey.servlet.version}</version>
+    </dependency>-->
```
boky01: Could you delete the unnecessary dependency instead of commenting out?

jongyoul: Thanks. I'll remove it.

jongyoul:
I got one error from travis. It seems to be related issue

echarles:
How does this PR relates to #932 ?
#932 simply replaces cxf with jersey2. Here I see we still have some cxf deps.

jongyoul:
This PR doesn't remove cxf, but just fixs the version mismatch issue. I'm not sure why we have to use cxf but I also don't know what the benefit is to remove cxf. This is WIP and I will focus only the minimum fix not to raise error when starting ZeppelinServer.

echarles:
This issue is recurring since a long time and depends IMHO on each environment (maven version...) - The main parameter to fix this is the `javax.ws.rsapi.version`.

This PR upgrades cxf from v2 to v3 and jersey from v1 to v2 which confuses me. I would rather see using only jersey2 as in #932.

Also Moving from jersey-v1 to jersey-v2 needs an update of the licence file as explained in #932.

Maybe the minimal fix would be leaving cxf as such and simply enforcing `javax.ws.rsapi.version` (I did that some time ago as temp fix).

jongyoul:
Hold until concluding #932

sohaibiftikhar:
Hi, what is the status on this issue. Is there some temporary fix that can be used. Please tell me if you need help fixing/testing this issue. This is important because I cannot get LDAP to work if this keeps triggering on authorisation failure.

jongyoul:
@sohaibiftikhar Hi, I'm holding it because of the former working #932 which looks similar to this. Can you share more about your problem? I'm actually blocked by disabling shiro. Without shiro, I don't know how to handle it. Can you test this PR? I decide to hold because it's one of minor issue but you suffer from the critical problem by this issue, I can handle it sooner with your help. :-)

sohaibiftikhar:
For some reason the jar conflict(with the no such method exception) keeps showing up for me when using shiro with `org.apache.zeppelin.realm.LdapRealm`. When I had earlier used `GroupLdapRealm` there was no issue. I am not really sure if this is the reason or it is plain chance. I can certainly help with testing it but do you know why the Jenkins test are failing with your merge request?

jongyoul:
@sohaibiftikhar #932 didn't adopt CI then you'd better test this build and fix some minor issues from it. Thanks!

jongyoul:
@sohaibiftikhar Sorry for confusing you. I didn't know #932 set the CI. Please test it with that PR

jongyoul:
Will close it as resolved by #932

PR #2185, opened 2017-03-23
[MINOR][ZEPPELIN-2308] Fix vis icon button group

AhyoungRyu:
### What is this PR for?
Since "Save" button which shows up when there is change in the package bundle ordering is included in same button group with other Helium viz icon buttons, the rightmost viz icon is not regarded as last child. For this reason, its border radius is a bit wired.

So I separated "Save" button from the icon btn group. 

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-2308](https://issues.apache.org/jira/browse/ZEPPELIN-2308)

 - Before
before
 
 - After 
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/helium/helium.css
+++ b/zeppelin-web/src/app/helium/helium.css
@@ -113,6 +113,10 @@
   float: left;
 }
 
+.heliumBundleOrder p {
+  margin:0 0 5px 15px
+}
+
 .sortable-row .as-sortable-item-handle {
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
LGTM

AhyoungRyu:
@Leemoonsoo Thanks! 
Merge into master and branch-0.7 if there are no more comments.

PR #2042, opened 2017-02-20
[ZEPPELIN-1588]: bumping nvd3 to 1.8.5

Remilito:
### What is this PR for?
* bump nvd3 to 1.8.5 (and remove depecrated functions)
* display percentage in pie chart [solve ZEPPELIN-1891]

NB: visualization-scatterchart.js's tooltip content generator has been updated to stop using depecrated tooltip property and use tooltip.contentGenerator instead. However I have commented the code as I think nvd3 scatterchart's default tooltip is far more elegant, open to discussion.

### What type of PR is it?
Improvement

### Todos
* [ ] - 

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-1588>
* <https://issues.apache.org/jira/browse/ZEPPELIN-1891>

### How should this be tested?
Visual testing of the builtin visualization.

### Questions:
* Does the licenses files need update? NO
* Is there breaking changes for older versions? NO
* Does this needs documentation? NO

PR #2129, opened 2017-03-13
[ZEPPELIN-2252] discreteness of line charts

tinkoff-dwh:
### What is this PR for?
Added a discreteness in the line chart. Correcting the display of the chart, when the values does not exist at the point (by default Y is set to 0, at the moment). A value of 0 is unclear (it is real value from data or it is default value).

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2252

### How should this be tested?
1. Spark interpreter
```
var text = """g1,1
g1,1
g1,2
g1,3
g1,5
g1,5
g2,1
g2,1
g2,2
g2,2
g2,6
g2,6
g2,7
g2,7
g2,7"""

val array = sc.parallelize(text.split('\n'))

case class Line(group: String, value: Integer)

val line = array.map(s => s.split(",")).map(
    s => Line(s(0), 
            s(1).toInt
        )
).toDF()
line.registerTempTable("lines")
```
2.
```
%sql
select *from lines
```
Settings of chart on screenshot

### Screenshots (if appropriate)
before

after

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #2186, opened 2017-03-24
[ZEPPELIN-2229] revert ndv3 version from 1.8.5 to 1.7.1

soralee:
### What is this PR for?
After [[ZEPPELIN-1588]: bumping nvd3 to 1.8.5](#2042) merged, Zeppelin was occurring front-end side error such as `Uncaught TypeError: Cannot read property 'x' of null` because of upgrading for `1.7.1` to `1.8.5`. 
AFAIK, this is a bug in `ndv3`, and ndv3 is still fixing the problem discussing [here](https://github.com/novus/nvd3/issues/330).
Also The latest release(1.8.5 version) scatter file and the master scatter file have a lot of changes, so we should wait for the next release.
Thus, I think we need to revert version from 1.8.5 to 1.7.1 for stable Zeppelin and our next 0.7.1 release.

_Plus, when I check on current master of nvd3, scatter chart works well like this._
z_2229_master_check

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2229](https://issues.apache.org/jira/browse/ZEPPELIN-2229)

* Switch some of chart such like screenshots.

[Before]
x_of_null_error

[After]
z_2229_revert_chart2

### Questions:

### DIFF
--- a/zeppelin-web/bower.json
+++ b/zeppelin-web/bower.json
@@ -17,7 +17,7 @@
     "ace-builds": "1.2.6",
     "angular-ui-ace": "0.1.3",
     "jquery.scrollTo": "~1.4.13",
-    "nvd3": "~1.8.5",
+    "nvd3": "~1.7.1",
     "angular-dragdrop": "~1.0.8",
     "perfect-scrollbar": "~0.5.4",
     "ng-sortable": "~1.3.3",
--- a/zeppelin-web/src/app/visualization/builtins/visualization-piechart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-piechart.js
@@ -76,8 +76,9 @@ export default class PiechartVisualization extends Nvd3ChartVisualization {
 
   configureChart(chart) {
     chart.x(function(d) { return d.label;})
-	 .y(function(d) { return d.value;})
-	 .showLabels(false)
-	 .showTooltipPercent(true);
+	 .y(function(d) { return d.value;});
+	 // if upgrade ndv3 to 1.8.5 or over, the following code is available
+	 /*.showLabels(false)
+	 .showTooltipPercent(true);*/
   };
 }
--- a/zeppelin-web/src/app/visualization/builtins/visualization-scatterchart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-scatterchart.js
@@ -69,6 +69,16 @@ export default class ScatterchartVisualization extends Nvd3ChartVisualization {
     chart.xAxis.tickFormat(function(d) {return self.xAxisTickFormat(d, self.xLabels);});
     chart.yAxis.tickFormat(function(d) {return self.yAxisTickFormat(d, self.yLabels);});
 
+    // configure how the tooltip looks.
+    chart.tooltipContent(function(key, x, y, graph, data) {
+      var tooltipContent = '<h3>' + key + '</h3>';
+      if (self.config.size &&
+        self.isValidSizeOption(self.config, self.tableData.rows)) {
+        tooltipContent += '<p>' + data.point.size + '</p>';
+      }
+      return tooltipContent;
+    });
+
     chart.showDistX(true).showDistY(true);
     //handle the problem of tooltip not showing when muliple points have same value.
   };


AhyoungRyu:
@soralee Thanks for digging it up! 
Seems https://github.com/novus/nvd3/pull/1934 fixed this issue and merged in Jan. Hopefully they have a plan to release another minor version soon including this fix.

>Thus, I think we need to revert version from 1.8.5 to 1.7.1 for stable Zeppelin and our next 0.7.1 

Then we can just revert #2042 commit from `branch-0.7`(and `master`?) instead of opening another PR and merging it. @Leemoonsoo What do you think?

AhyoungRyu:
@Remilito Will be appreciated if you can take a look this :)

soralee:
@AhyoungRyu This PR reverted some of #2042 and I didn't include to change [this line]( https://github.com/apache/zeppelin/pull/2042/files#diff-7a20bcbe25109dfb2ad811a62a366cbeL63).
If we decide to revert # 2042, I can give you help to check the reverted code.

AhyoungRyu:
@soralee Yeah I can see the error when we have large data. 
again

I tested it in release `0.7.1-rc2` bin package. Can you mention this PR with short description in `0.7.1-rc2` release vote? :)

soralee:
@AhyoungRyu I did!  Thanks for reproduce!
Let me check this part I didn't include. If it is possible, we had better revert it.

soralee:
@AhyoungRyu @Leemoonsoo I just checked it!
We can either revert or merge it. the part I didn't include don't matter.

AhyoungRyu:
@soralee Thanks for the explanation.

I tested this branch to make sure it works well or not, but when I select "Line chart" below error msg shows up. 
```
vendor.bcb986b75e1664c0.js:64 Error: <g> attribute transform: Trailing garbage, "translate(813,NaN)".
```
record

I checked multiple times, but I could see this every time. Could you check this please?

AhyoungRyu:
@soralee And as you can see in the above gif image that i took, the tooltip in scatter chart is shown like "(age, value)". But before #2042 it looks like below.
image

soralee:
@AhyoungRyu Thanks for detail test! Let me check this and leave comment!

Remilito:
My 2 cts here:
Yep the scatter chart bug in 1.8.5 is annoying, and unfortunately nvd3 does not seem to release very frequently ( 2-3 times a year or so). However I am not sure reverting the lib from 1.8.5 to 1.7.1 is the right move, keep in mind that  **a lot of bugs** were fixed between those two releases.
Reverting to 1.7.1 is very likely to reintroduce (or maintain) a lot of others bugs. 

I wonder if we could do something directly in Zeppelin's visualization code to workaround the voronoi issue you spotted until 1.8.6 is released. Your thoughts?

anthonycorbacho:
I will agree with @Remilito on this one.
I think we can come up with a workaround and keep 1.8.5.

soralee:
@AhyoungRyu I checked [your comment about translate error](https://github.com/apache/zeppelin/pull/2186#issuecomment-289178564).
The error seems to be caused by incompatibility with the 1.7.1 version and [this line](https://github.com/apache/zeppelin/pull/2129/files#diff-28a8fa162a27698e421ee716f1822a33L73) that is after merged this #2129 in master branch.
If we decide merge, I would fix issue.

soralee:
@Remilito @anthonycorbacho Thanks for opinions!
Yeah, I think whatever we use 1.7.1 version or 1.8.5 versions, we have a lot of bugs. `nvd3` release process is too slow and, I'm not sure 1.8.5 version has critical bugs than 1.7.1 version :disappointed:
If we don't mind those issues, we would be nice to keep the latest version!

Remilito:
@soralee , @AhyoungRyu , @anthonycorbacho : proposal of workaround here, https://github.com/apache/zeppelin/pull/2194 . Feedbacks welcome!

soralee:
I closed this PR. Because In the better way which keeps on the latest version, @Remilito suggested #2194, I think we don't need to talk about this PR any more.

soralee:
Let's move on #2194 PR!

PR #2191, opened 2017-03-27
[HOTFIX] Keep interpreter group when save and restart

jongyoul:
### What is this PR for?
Not removing interpreter group reference while editing and saving interpreter setting.

### What type of PR is it?
[Hot Fix]

### Todos
* [x] - Move `loadFromFile` into proper location

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2037
* #1976 

1. Enable shiro for login
1. Execute sample code by %python and %pyspark
1. Switch spark interpreter from global mode to Per user mode and save it.
1. Switch python interpreter from Per user mode to global mode save it.
1. Shutdown zeppelin
1. Verify if there's no zombie processes

### Questions:


soralee:
Tested and It works as your described, LGTM!

del680202:
I check this PR, it works well, thanks.

PR #2195, opened 2017-03-27
[ZEPPELIN-2319] new methods for ZeppelinContext

meniluca:
### What is this PR for?
Adds new methods to ZeppelinContext to run paragraphs in a more user-friendly mode.

```java
z.runAllBefore()
z.runAllAfter()
z.run(int... ids)
z.run(String... ids)
z.runNext(int i)
```
more details in the Jira tickets.

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-2319](https://issues.apache.org/jira/browse/ZEPPELIN-2319)

They all wrap the ``run(List<Object> ids)`` method, hence can be tested in the same test suite.

Coming soon

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -413,6 +413,15 @@ public class ZeppelinContext {
     String noteId = interpreterContext.getNoteId();
     run(noteId, idx, interpreterContext);
   }
+  
+  @ZeppelinApi
+  public void run(int... paragraphIndexes) {
+    ArrayList<Object> listIDs = new ArrayList<Object>();
+    for (int i = 0; i < paragraphIndexes.length; i++) {
+      listIDs.add(paragraphIndexes[i]);
+    }
+    run(listIDs);
+  }
 
   /**
    * Run paragraph at index
@@ -457,7 +466,16 @@ public class ZeppelinContext {
       }
     }
   }
-
+  
+  @ZeppelinApi
+  public void run(String... paragraphIDs) {
+    ArrayList<Object> listIDs = new ArrayList<Object>();
+    for (int i = 0; i < paragraphIDs.length; i++) {
+      listIDs.add(paragraphIDs[i]);
+    }
+    run(listIDs);
+  }
+  
   @ZeppelinApi
   public void runAll() {
     runAll(interpreterContext);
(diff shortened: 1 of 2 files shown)


zjffdu:
@meniluca I am a little conservative for adding new public api in ZeppelinContext. As new api means more overhead to maintain it and keep backward compatibility. Could you describe the scenario of why you need these api ? Maybe we could build more generate api or solution for it.

meniluca:
@zjffdu I agree in principle, but if you look at the code they are few rather trivial new methods, all of them wrapping up the already existing ``run()``. I have these function running in my notebooks since 1.5 versions, I don't think you have to worry about backwards compatibility and maintenance too much (they are really simple!). I'll add screenshots later, but it's easy to imagine what they do.

Say that I am in the middle of the notebook and I want to run all paragraphs before a certain point, I will run ``z.runAllBefore`` and that's it. In the same way if I need to run the rest of the notebook I will use ``z.runAllAfter``. Those functionalities are really useful if you are building a notebook with complex dynamic forms usage and they will avoid the user to click repeatedly the run button.

In Jupyter buttons and shortcuts that are doing this are already present, hence I think that sooner or later you will need those functions.

To describe quickly the other 3 functions... Say that I have one block where I setup variables and I want other paragraphs to be updated, instead of issuing:
```scala
val maxAge = z.input("maxAge", 30) // select age and run paragraphs
z.run("parID1")
z.run("parID2")
z.run("parD3")
```
I will have
```scala
z.run("parID1","parID2","parID3")
```
or, even more useful, to avoid repetitions
```scala
val arrayIDs = Array("parID1","parID2","parID3")
z.run(arrayIDs:_*)
// in another paragraph
z.run(arrayIDs:_*)
```

Similarly the function ``z.run(int... idx)`` is doing the same thing, meanwhile ``z.runNext(int N)``, runs the following N paragraphs. Imagine I have the same groups of three blocks in sequence and every time I need to update a variable:

```scala
var maxAge = z.input("maxAge", 30)
z.runNext(3)
// following 3 paragraphs are executed
// ... later, in another paragraph below, $maxAge is updated
maxAge = df.where("balance > 2000").max("age").collect()(0)
z.runNext(3) 
// and so on...
```

Anyway, most of them are meant to be more user-friendly wrappers of the ``run()`` functions, syntactic sugar. The most important are the runAllBefore/After in my opinion. Let me know what you think and how I can be of any help.

Cheers,
Luca

zjffdu:
Sorry for late response. @meniluca I am very interested in your scenario. It make sense to me for adding these api, another concern is whether it is more proper to expose it via ui or expose this api directly ? Any advantage for exposing it via api directly for your scenario ? And it would be better to add test case for it as well.

AhyoungRyu:
@meniluca Great initiative! Thanks for detailed explanation and it makes sense to me as well. That would be helpful to any other reviewers. 
But before you go further, need to setup travis account for Zeppelin first.

 - Please take a look https://zeppelin.apache.org/contribution/contributions.html#continuous-integration
 - The repository name must be Zeppelin (not incubator-zeppelin, ...)
 - It's not mandatory to pass all tests to be reviewed. But some reviewers might select passed PRs first

Trigger travis CI by pushing new commits.

zjffdu:
@meniluca Does it only work with one type of interpreter ? Say does it work with multiple different interpreters, such as spark + sh ? And please also check #2151

meniluca:
@AhyoungRyu now I think it's done :)
@zjffdu to answer your questions: it works with all interpreters and the kind of use case is exactly like PR #2151. In principle he should be able to use my functions to do that, am I right?

meniluca:
Guys, do you need something else from my side to merge?

jongyoul:
Can you make tests for that new method? there's no tests here.

meniluca:
@jongyoul, everybody, would you suggest where to put the test? Consider that all these methods are wrapping the ``ZeppelinContext.run("parID")`` method that is NOT tested in this module (simply because interpreters don't know anything about paragraph IDs). I need to have a list of paragraphs and their IDs, but I cannot ``import org.apache.zeppelin.notebook.*`` in a test of this module.

Can I create a test elsewhere? Maybe in ``zengine``? Or can you suggest a solution?

I want to stress again that this PR is simply adding more user-friendly wrappers of already exposed public methods in the ZeppelinContext. They are necessary to give user more flexibility in building more dynamic/interactive notebooks, not only to add new buttons and whatnot.

Cheers,
Luca

Leemoonsoo:
There can be simpler way to make a test for wrapper function using mockito.

Other way is adding like [zRunTest](https://github.com/apache/zeppelin/blob/v0.7.1/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java#L319)

PR #2197, opened 2017-03-28
ZEPPELIN-2324. Add property zeppelin.spark.enableSupportedVersionCheck for trying new spark version

zjffdu:
### What is this PR for?
For now, every time when I want to try new spark version, I have to change file `SparkVersion.java` and rebuild it. It is not so convenient, so I'd like to add property `zeppelin.spark. enableSupportedVersionCheck` for spark interpreter. So that I can try new spark version by setting this property as false, of course it is only for zeppelin developer.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2324

Verify it in spark master

### Questions:

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -140,6 +140,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>true</td>
     <td>Import implicits, UDF collection, and sql if set true.</td>
   </tr>
+  <tr>
+    <td>zeppelin.spark.unSupportedVersionCheck</td>
+    <td>true</td>
+    <td>Don't change it, It is only for zeppelin developer use, not safe or use in production</td>
+  </tr>
 </table>
 
 Without any configuration, Spark interpreter works out of box in local mode. But if you want to connect to your Spark cluster, you'll need to follow below two simple steps.
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -141,9 +141,9 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>Import implicits, UDF collection, and sql if set true.</td>
   </tr>
   <tr>
-    <td>zeppelin.spark.unSupportedVersionCheck</td>
+    <td>zeppelin.spark.enableSupportedVersionCheck</td>
     <td>true</td>
-    <td>Don't change it, It is only for zeppelin developer use, not safe or use in production</td>
+    <td>Do not change - developer only setting, not for production use</td>
   </tr>
 </table>
 
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -128,7 +128,7 @@ public class SparkInterpreter extends Interpreter {
   private static File outputDir;          // class outputdir for scala 2.11
   private Object classServer;      // classserver for scala 2.11
   private JavaSparkContext jsc;
-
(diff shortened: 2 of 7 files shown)


zjffdu:
@Leemoonsoo @jongyoul  @felixcheung Please help review.

docs/interpreter/spark.md
```diff
@@ -140,6 +140,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>true</td>
     <td>Import implicits, UDF collection, and sql if set true.</td>
   </tr>
+  <tr>
+    <td>zeppelin.spark.unSupportedVersionCheck</td>
+    <td>true</td>
+    <td>Don't change it, It is only for zeppelin developer use, not safe or use in production</td>
```
felixcheung: could you fix the capitalization:
`unSupportedVersionCheck`
`Don't change it, It is only`

felixcheung: I think I'd suggest calling it
`zeppelin.spark.disableSupportedVersionCheck`
or
`zeppelin.spark.debugNoSupportedVersionCheck`

felixcheung: And then in the description as `Do not change - developer only setting, not for production use` or similar

zjffdu:
> perhaps it should be changed in sparkInterpreter.getSparkVersion().isSupportedVersion() instead

I thought about it, but feel it is not proper to create a new member field in class SparkVersion as this is a interpreter level property, even it is not proper to me to create a static field in SparkVersion. So in the new PR, I create a new method `isUnSupportedSparkVersion` in `SparkInterpreter` to wrap all the logic. 

> Can you add testcase for this conf?
I didn't find an easy way for this test. So just manually verified it.

docs/interpreter/spark.md
zjffdu: change it to `zeppelin.spark.enableSupportedVersionCheck` and description to `Do not change - developer only setting, not for production use`

docs/interpreter/spark.md
```diff
@@ -140,6 +140,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>true</td>
     <td>Import implicits, UDF collection, and sql if set true.</td>
   </tr>
+  <tr>
+    <td>zeppelin.spark.enableSupportedVersionCheck</td>
+    <td>true</td>
```
felixcheung: I don't feel strongly about it, but since this is a debug config and it's turning off something important, I feel it is better to default to "off" that to disable it needs to be explicitly done?

I'll let other comment..

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1153,12 +1155,16 @@ public Object getLastObject() {
     return obj;
   }
 
+  boolean isUnSupportedSparkVersion() {
```
felixcheung: I'd totally change S -> s in `isUnsupportedSparkVersion` :)

docs/interpreter/spark.md
zjffdu: I change the configure name to `enable*`, so it is true by default.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: fixed

zjffdu:

PR #2201, opened 2017-03-29
[ZEPPELIN-1965] Livy SQL Interpreter: Should use df.show(1000, false)…

benoyantony:
… to display results

### What is this PR for?
Livy SQL interpreter truncate result strings of size greater than 20. In some cases, we like to see the full string. We are adding a interpreter property **zeppelin.livy.spark.sql.field.truncate** to control whether to truncate strings or not. By default, **zeppelin.livy.spark.sql.field.truncate** is set to **true**.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1965

Set zeppelin.livy.spark.sql.field.truncate to true or false 
Run a SQL query which produces string values of length greater than 20.
Depending on the value of zeppelin.livy.spark.sql.field.truncate, the strings will either get truncated or not.

### Questions:

### DIFF
--- a/docs/interpreter/livy.md
+++ b/docs/interpreter/livy.md
@@ -61,6 +61,11 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>Max number of Spark SQL result to display.</td>
   </tr>
   <tr>
+    <td>zeppelin.livy.spark.sql.truncate</td>
+    <td>true</td>
+    <td>Whether to truncate strings or not</td>
+  </tr>
+  <tr>
     <td>zeppelin.livy.session.create_timeout</td>
     <td>120</td>
     <td>Timeout in seconds for session creation</td>
--- a/docs/interpreter/livy.md
+++ b/docs/interpreter/livy.md
@@ -56,14 +56,14 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>URL where livy server is running</td>
   </tr>
   <tr>
-    <td>zeppelin.livy.spark.maxResult</td>
+    <td>zeppelin.livy.spark.sql.maxResult</td>
     <td>1000</td>
     <td>Max number of Spark SQL result to display.</td>
   </tr>
   <tr>
-    <td>zeppelin.livy.spark.sql.truncate</td>
+    <td>zeppelin.livy.spark.sql.field.truncate</td>
     <td>true</td>
-    <td>Whether to truncate strings or not</td>
+    <td>Whether to truncate field values longer than 20 characters or not</td>
   </tr>
   <tr>
     <td>zeppelin.livy.session.create_timeout</td>
--- a/livy/src/main/resources/interpreter-setting.json
+++ b/livy/src/main/resources/interpreter-setting.json
@@ -118,6 +118,11 @@
         "defaultValue": "1000",
         "description": "Max number of Spark SQL result to display."
       },
+      "zeppelin.livy.spark.sql.truncate": {
+        "propertyName": "zeppelin.livy.spark.sql.truncate",
+        "defaultValue": "true",
+        "description": "If true, truncate strings greater than 20 characters."
+      },
       "zeppelin.livy.concurrentSQL": {
         "propertyName": "zeppelin.livy.concurrentSQL",
         "defaultValue": "false",
--- a/livy/src/main/resources/interpreter-setting.json
+++ b/livy/src/main/resources/interpreter-setting.json
@@ -118,10 +118,10 @@
         "defaultValue": "1000",
         "description": "Max number of Spark SQL result to display."
       },
-      "zeppelin.livy.spark.sql.truncate": {
-        "propertyName": "zeppelin.livy.spark.sql.truncate",
+      "zeppelin.livy.spark.sql.field.truncate": {
+        "propertyName": "zeppelin.livy.spark.sql.field.truncate",
         "defaultValue": "true",
-        "description": "If true, truncate strings greater than 20 characters."
+        "description": "If true, truncate field values longer than 20 characters."
       },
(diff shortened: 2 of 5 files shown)


zjffdu:
Thanks @benoyantony for contribution, would you mind to add this configuration into livy.md as well ?

zjffdu:
One minor thing left, could you add test case for it ?

benoyantony:
There seems to be a problem with Jenkins as the error is " java.io.IOException: Remote call on ubuntu-2 failed". Any idea on howto overcome this ?

The travis build has all checks passed.

zjffdu:
It is caused by `Caused by: java.lang.OutOfMemoryError: Java heap space`, could you retrigger the CI ?

benoyantony:
I had retriggered it once. Doing it again.

benoyantony:
That seems to have worked. thanks @zjffdu .

benoyantony:
@felixcheung , Could you please help commit this patch ?

zjffdu:
Thanks @benoyantony , LGTM

docs/interpreter/livy.md
```diff
@@ -61,6 +61,11 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>Max number of Spark SQL result to display.</td>
   </tr>
   <tr>
+    <td>zeppelin.livy.spark.sql.truncate</td>
+    <td>true</td>
+    <td>Whether to truncate strings or not</td>
```
felixcheung: instead of `strings` should we say "output" or "results"?

felixcheung: and is this only for livy.sql interpreter? should it apply to other livy interpreter?

benoyantony:
Thanks for the prompt review, @zjffdu .

livy/src/main/resources/interpreter-setting.json
```diff
@@ -118,6 +118,11 @@
         "defaultValue": "1000",
         "description": "Max number of Spark SQL result to display."
       },
+      "zeppelin.livy.spark.sql.truncate": {
+        "propertyName": "zeppelin.livy.spark.sql.truncate",
+        "defaultValue": "true",
+        "description": "If true, truncate strings greater than 20 characters."
```
felixcheung: btw, I think perhaps this could be more useful if this is the truncate line numbers (eg. 20), instead of true/false.
I think I recall the Spark interpreter has it that way too.

docs/interpreter/livy.md
benoyantony: I wrote the description to match the documentation of the corresponding truncate argument in the spark dataset method
[http://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.sql.Dataset@show(numRows:Int,truncate:Boolean):Unit](http://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.sql.Dataset@show(numRows:Int,truncate:Boolean):Unit)

> truncate -
> Whether truncate long strings. If true, strings more than 20 characters will be truncated and all cells will be aligned right

This should be applicable only to the LivySparkSQLInterpreter. Only in this interpreter,  we invoke the above method as part of the code. For other interpreters, the user supplied code can call the appropriate show() method.

livy/src/main/resources/interpreter-setting.json
benoyantony: I tired this, but I get the following error:  " found   : Int(20)\n"," required: Boolean\n".

docs/interpreter/livy.md
benoyantony: Should I say 

> Whether to truncate strings in results or not

livy/src/main/resources/interpreter-setting.json
zjffdu: `zeppelin.livy.spark.sql.maxResult` is for truncating line numbers

docs/interpreter/livy.md
felixcheung: right, to consider:
https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java#L56

livy/src/main/resources/interpreter-setting.json
felixcheung: oh got it, sorry, so there's one for max line number.
perhaps to rename this to clarify? `zeppelin.livy.spark.sql.truncate.string.value` or something?

zjffdu: How about `zeppelin.livy.spark.sql.field.truncate` ?  This is for truncating each field in table.

benoyantony: Shall I make the following change ?

> zeppelin.livy.spark.sql.field.truncate
> If true, truncate strings greater than 20 characters. (interpreter-setting.json)

> Whether to truncate fields longer than 20 characters or not  (in livy.md)

docs/interpreter/livy.md
```diff
@@ -56,14 +56,14 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>URL where livy server is running</td>
   </tr>
   <tr>
-    <td>zeppelin.livy.spark.maxResult</td>
+    <td>zeppelin.livy.spark.sql.maxResult</td>
```
felixcheung: I love this, but it might break existing users. Perhaps we don't change this for now?

benoyantony:
It should not break anything as the property name remains same as before.
The documentation had a typo in it. It was missing 'sql' in it.

If it's confusing, I can do it as a separate fix.

benoyantony:
Thank you @zjffdu  and @felixcheung

PR #2206, opened 2017-03-30
[ZEPPELIN-1856] Flaky Test: InterpreterRestApiTest.testRestartInterpreterPerNote

AlexanderShoshin:
### What is this PR for?
A _NullPointerException_ occurs occasionally in `InterpreterRestApiTest.testRestartInterpreterPerNote()` test when `p.getResult()` returns _null_ in the following line of test method:
```
assertEquals(p.getResult().message().get(0).getData(), getSimulatedMarkdownResult("markdown"));
```
It happens when we recieve an _OUTPUT_UPDATE_ALL_ event in _RemoteInterpreterEventPoller_ after `RemoteInterpreter.interpret(..)` is done. The `Paragraph.result` which already contains its final value is setted to _null_ by `NotebookServer.onOutputClear(..)` method. As a result we recieve a _NPE_ in assertion.

### What type of PR is it?
Bug Fix

### Todos
* [x] - To fix this issue we need to ignore an _OUTPUT_UPDATE_ALL_ event after job is finished.
* [x] - To write tests to check that paragraph result can be changed during running and can't after completion.

### What is the Jira issue?
[ZEPPELIN-1856](https://issues.apache.org/jira/browse/ZEPPELIN-1856)

### Questions:


jongyoul:
Can you make a test in case OUTPUT_UPDATE_ALL is passed after job is done?

AlexanderShoshin:
Sure, I'll do it.
But first I need to make some refactoring of `Note.java` class in separate PR. We need to be able to add a custom `Paragraph` to `Note`. This will allow to use a `Paragraph` which we can control while performing. So we will be able to catch `Paragraph` _RUNNING_ and _FINISHED_ phases in a test.

AlexanderShoshin:
Hi @jongyoul,
I've added a test for my changes but it won't run until we restore second Travis job which should start all core tests.
I will write when the test will be able to run on Travis.

jongyoul:
@AlexanderShoshin Can you rebase it from current master?

AlexanderShoshin:
@jongyoul it looks like I need to correct my unit tests due last changes in master

jongyoul:
Can you fix it to fit with current master as well?

PR #2207, opened 2017-03-30
[Zeppelin-802] Support for Zeppelin Context redefinition on Python and Pyspark 

cloverhearts:
### What is this PR for?
If you override the reserved word ZeppelinContext such as `z` in the python language, the whole paragraph output problem occurred.
I have taken care to avoid this issue.

`z` == `_zc` == `zeppelin context`

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-802

The error should not occur in the following situations:
```
%python
z = 1
print("Hello Zeppelin")
```

```
%pyspark
z = 1
print("Hello Zeppelin")
```

#### before
replace zeppelin context-err

#### after
replace zeppelin context

### Questions:

### DIFF
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -271,19 +271,34 @@ else:
 
 java_import(gateway.jvm, "scala.Tuple2")
 
+_zcUserQueryNameSpace = {}
+
 jconf = intp.getSparkConf()
 conf = SparkConf(_jvm = gateway.jvm, _jconf = jconf)
-sc = SparkContext(jsc=jsc, gateway=gateway, conf=conf)
+sc = _zsc_ = SparkContext(jsc=jsc, gateway=gateway, conf=conf)
+_zcUserQueryNameSpace["_zsc_"] = _zsc_
+_zcUserQueryNameSpace["sc"] = sc
+
 if sparkVersion.isSpark2():
-  spark = SparkSession(sc, intp.getSparkSession())
-  sqlc = spark._wrapped
+  spark = __zSpark__ = SparkSession(sc, intp.getSparkSession())
+  sqlc = __zSqlc__ = __zSpark__._wrapped
+  _zcUserQueryNameSpace["sqlc"] = sqlc
+  _zcUserQueryNameSpace["__zSqlc__"] = __zSqlc__
+  _zcUserQueryNameSpace["spark"] = spark
+  _zcUserQueryNameSpace["__zSpark__"] = __zSpark__
 else:
-  sqlc = SQLContext(sparkContext=sc, sqlContext=intp.getSQLContext())
-sqlContext = sqlc
+  sqlc = __zSqlc__ = SQLContext(sparkContext=sc, sqlContext=intp.getSQLContext())
+  _zcUserQueryNameSpace["sqlc"] = sqlc
+  _zcUserQueryNameSpace["__zSqlc__"] = sqlc
+
+sqlContext = __zSqlc__
+_zcUserQueryNameSpace["sqlContext"] = sqlContext
 
 completion = PySparkCompletion(intp)
-z = _zc = __PyZeppelinContext__(intp.getZeppelinContext())
-_zc._setup_matplotlib()
+z = __zeppelin__ = __PyZeppelinContext__(intp.getZeppelinContext())
+__zeppelin__._setup_matplotlib()
+_zcUserQueryNameSpace["z"] = z
+_zcUserQueryNameSpace["__zeppelin__"] = __zeppelin__
 
 while True :
   req = intp.getStatements()
@@ -299,7 +314,7 @@ while True :
       global_hook = None
       
     try:
-      user_hook = _zc.getHook('post_exec')
+      user_hook = __zeppelin__.getHook('post_exec')
     except:
       user_hook = None
       
@@ -334,17 +349,17 @@ while True :
         for node in to_run_exec:
           mod = ast.Module([node])
           code = compile(mod, '<stdin>', 'exec')
-          exec(code)
+          exec(code, _zcUserQueryNameSpace)
 
         for node in to_run_single:
           mod = ast.Interactive([node])
           code = compile(mod, '<stdin>', 'single')
-          exec(code)
+          exec(code, _zcUserQueryNameSpace)
           
         for node in to_run_hooks:
           mod = ast.Module([node])
           code = compile(mod, '<stdin>', 'exec')
-          exec(code)
+          exec(code, _zcUserQueryNameSpace)
       except:
         raise Exception(traceback.format_exc())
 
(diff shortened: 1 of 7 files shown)


spark/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterTest.java
```diff
@@ -123,6 +123,21 @@ public void testCompletion() {
     }
   }
 
+  @Test
+  public void testRedefinitionZeppelinContext() {
+    if (getSparkVersionNumber() > 11) {
+      String redefinitionCode = "z = 1\n";
```
zjffdu: verify `ZeppelinContext` still work after reassign of `z`

cloverhearts: Hello.
If you do not mind, can you tell me more details?
I think I did the verification procedure properly.
Do you have any problems?

```
//  "z = 1\n";
assertEquals(InterpreterResult.Code.SUCCESS, pySparkInterpreter.interpret(redefinitionCode, context).code()); 
// "z.input(\"test\")\n";  <-- If z is reassign, an error occurs.
assertEquals(InterpreterResult.Code.ERROR, pySparkInterpreter.interpret(validCode, context).code());
```

zjffdu: Sorry, misread the code, LGTM 😄

cloverhearts: @zjffdu Thank you!! :-)

astroshim:
Looks great. 
one thing i worrying about is that the `_zc` can be possibly used by user easily?
how about assigning more unique key word like `__zc__` ?

cloverhearts:
@astroshim 
So what about `__zeppelin__`?
Actually, I feel similar like a python system environment word
It is less likely to be redefined by other libraries.

astroshim:
@cloverhearts It looks also great!  👍

AhyoungRyu:
@cloverhearts Great! Tested this patch and it works well as you described. 
 - Before
screen shot 2017-03-31 at 11 58 32 am

 - After 
screen shot 2017-03-31 at 12 04 38 pm

If @astroshim feedback can also be addressed, looks GOOD to me.

felixcheung:
Nice!!

cloverhearts:
@felixcheung Thank you for your good advice.
I will check your opinion.
and fix :)

felixcheung:
right, I think it's everything that is called after calling `exec(code)`, might have been overwritten by the user code

```
or node in to_run_exec:
          mod = ast.Module([node])
          code = compile(mod, '<stdin>', 'exec')
          exec(code)

        for node in to_run_single:
          mod = ast.Interactive([node])
          code = compile(mod, '<stdin>', 'single')
          exec(code)

        for node in to_run_hooks:
          mod = ast.Module([node])
          code = compile(mod, '<stdin>', 'exec')
          exec(code)
```

so I think: `intp`, `output`
and perhaps `to_run_exec`, `to_run_single`, `to_run_hooks`?

zjffdu:
I think a more thorough solution is to separate the namespace between python process and python repl. Only `ZeppelinContext` should be `available` in python repl namespace.  It would be better to specify a space in `exec`.  
e.g. 
```
exec(code, global=repl_namespace)
```

felixcheung:
we need to check if namespace work properly on py2 and py3...

zjffdu:
@cloverhearts You may miss my last comment. We should separate the namespace of python process and repl, otherwise there's maintenance overhead when we update the script to add some other variables in future.

cloverhearts:
@zjffdu 
Yes, I will also reflect the namespace.
Thank you for your advice :)
.

cloverhearts:
apply to namespace on python

and zeppelin context name change

before
```
z = _zc = zeppelin context
```
to
```
z = __zeppelin__ = zeppelin context
```

same
```
z.input("title", "my title")
__zeppelin__.input("title", "my title")
```

cloverhearts:
and tested environment.


| | pyspark intp | python intp |
|-|--------|------|
|python3 with matplotlib | O| O |
|python2 with matplotlib | O | O |

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -230,9 +230,9 @@ def getCompletion(self, text_value):
       self.interpreterObject.setStatementsFinished(result, False)
 
 
-output = Logger()
-sys.stdout = output
-sys.stderr = output
+__zcStdOutput__ = __ZeppelinLogger__()
+sys.stdout = __zcStdOutput__
+sys.stderr = __zcStdOutput__
```
zjffdu: I think you don't need to rename output, as it is not in the namespace of `repl`

cloverhearts: Okay, Thank you.
I will check and fix

cloverhearts:
Okay, Ci pass :)
@felixcheung @zjffdu 
Could you please check on this pr?
Thank you :)

cloverhearts:
I would like to merge this.
This solves the problem of usability.
I will merge this if I no longer have any opinion.
Thank you :)

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -271,19 +271,37 @@ def getCompletion(self, text_value):
...
+  sqlc = __zSqlc__ = SQLContext(sparkContext=sc, sqlContext=intp.getSQLContext())
+  _zcUserQueryNameSpace["sqlc"] = sqlc
+  _zcUserQueryNameSpace["__zSqlc__"] = sqlc
+
+sqlContext = __zSqlc__
+_zcUserQueryNameSpace["sqlContext"] = sqlContext
+
+completion = __zeppelin_completion__ = PySparkCompletion(intp)
+_zcUserQueryNameSpace["completion"] = completion
+_zcUserQueryNameSpace["__zeppelin_completion__"] = __zeppelin_completion__
```
zjffdu: I think user don't use `completion`, so we don't need to expose it into user repl namespace

zjffdu:
@cloverhearts  I left one minor comment. Besides,  we need to add document for this.

spark/src/main/resources/python/zeppelin_pyspark.py
cloverhearts: @zjffdu 
Yes, good point! :)
Clearly, You are right.
but as a result of checking, pyspark is the type that currently calls autocompletion completion from the user query.
I also think this is wrong.
(but, i was had checked and tested then got a error message `not defined completion`)
in my opinion, this is irrelevant to the PR, so it should be corrected in the next PR.

cloverhearts:
@zjffdu 
How to use is the same as the existing one.
If I add a document, should I add something like __zeppelin__ or __spark__?
What do you think? :)

spark/src/main/resources/python/zeppelin_pyspark.py
cloverhearts: ref https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java#L465

zjffdu:
hmm, right. Seems document is not needed 😄

cloverhearts:
will be merging if no further comments

felixcheung:
LGTM

PR #2220, opened 2017-04-04
[Minor] Fix wrong Python lib path in Matplotlib tutorial note

AhyoungRyu:
### What is this PR for?
There is a wrong information about Zeppelin’s matplotlib backend files. It actually located under `ZEPPELIN_HOME/interpreter/lib/python/*` but tutorial note says it as `ZEPPELIN_HOME/lib/python/*`. So I corrected it.

### What type of PR is it?
kind a typo fix 

### What is the Jira issue?
N/A

 - Before 
screen shot 2017-04-04 at 3 50 06 pm

 - After
screen shot 2017-04-04 at 3 50 56 pm

### Questions:


AhyoungRyu:
@felixcheung Thanks! 
Reading `ZEPPELIN_HOME` and providing full path would be better as you said, but it says "usually found in..". So I think it's not that much hard for users to understand or find the matplotlib related files. 
screen shot 2017-04-05 at 1 27 19 pm

AhyoungRyu:
Merge into master and branch-0.7 if there are no more comments.

PR #2221, opened 2017-04-04
[ZEPPELIN-2216] general solution to precode. refactoring jdbc precode

tinkoff-dwh:
### What is this PR for?
General solution to execute precode. Refactoring jdbc precode using general solution. Task contains to subtasks: executeAfterOpen, executeBeforeClose. executeBeforeClose not done because we need the context so there is a solution only for executeAfterOpen. 

### What type of PR is it?
Feature | Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2216

1. Add parameter zeppelin.PySparkInterpreter.precode `someVar='text'`
2. Execute 
```
%pyspark
print(someVar)
```

### Questions:


tinkoff-dwh:
ready to review

tinkoff-dwh:
prevous discussion #2096

tinkoff-dwh:
@Leemoonsoo 
prevous discussion #2096

Leemoonsoo:
Let me take a look in this weekend!

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -481,19 +481,24 @@ protected Object jobRun() throws Throwable {
...
 
+        InterpreterResult result = null;
+
         // Open the interpreter instance prior to calling interpret().
         // This is necessary because the earliest we can register a hook
         // is from within the open() method.
         LazyOpenInterpreter lazy = (LazyOpenInterpreter) interpreter;
         if (!lazy.isOpen()) {
           lazy.open();
+          result = lazy.executePrecode(context);
```
Leemoonsoo: Shouldn't `RemoteInterpreterServer.open()` also call `executePrecode()`?

tinkoff-dwh: it must be called once per instance of the interprete

tinkoff-dwh: @Leemoonsoo 
works same as the previous jdbc precode.
case:
   user: some_user
   JDBC interpreter (per user)
   1. there is precode `CREATE table IF NOT EXISTS tbl_#{user}`
   2. user executes `insert into tbm_some_user ...`
      - execute `precode`
      - execute paragraph text
  2. user executes `select *from tbm_some_user`
     - execute paragraph text

Spark   
1. there is precode `var someConst = 12 ...`
2. User executes `var result = 1 + someConst`
   - execute `precode`
   - execute paragraph text
3. User executes var 
```
result2 = 2 + someConts
print(result2)
```
- execute only paragraph text

Leemoonsoo: Thanks for explain.

Currently, unittest https://github.com/apache/zeppelin/pull/2221/files#diff-9c5e6fbaa5b23bc0192b4fedc12b2680R445 executes `jdbcInterpreter.executePrecode(interpreterContext);` manually after `open()`. 

Shell we have other unittest that does not call `executePrecode()` manually? so `lazy.executePrecode` in InterpreterServer can be verified.

`zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterTest.java` is one place the unittest actually creates RemoteInterpreter process and test against it. This place might be the place you can locate test for this feature.

tinkoff-dwh: @Leemoonsoo 
fix it, if I understand correctly your comment

Leemoonsoo:

PR #2222, opened 2017-04-04
add a configuration zeppelin.jdbc.auth.kerberos.proxy for kerberos proxy behavior

wary:
…e to disable proxy behavior

### What is this PR for?

### What type of PR is it?
[Improvement]
* add a configuration item  zeppelin.jdbc.auth.kerberos.proxy.enable to disable kerberos behaviour 
as we know， in current version of zeppelin， if we have kerberos auth configured，and using zeppelin with a user login，the jdbc interpreter will do a proxy behavior with the login user automatically，but in many cases， we do not want do this，and we do not want bind the zeppelin user system with kerberos auth system. I think it's make senses to add a configuration item to disable this behavior.

### Todos
* [ ] - Task

### What is the Jira issue?

[ZEPPELIN-2353] (https://issues.apache.org/jira/browse/ZEPPELIN-2353)

when we have kerberos auth configured, and using zeppelin with a login user 
add configuration:
"zeppelin.jdbc.auth.kerberos.proxy.enable=false"
for jdbc iterpreter configuration, the jdbc iterpreter will not do the proxy behavior with the login user

### Questions:

### DIFF
--- a/docs/interpreter/jdbc.md
+++ b/docs/interpreter/jdbc.md
@@ -166,7 +166,7 @@ There are more JDBC interpreter properties you can specify like below.
   </tr>
   <tr>
       <td>zeppelin.jdbc.auth.kerberos.proxy.enable</td>
-      <td>when auth type is kerberos auth type,whether do the kerberos proxy behavior with the login user (if not anonymous) to get the connection, default value is true</td>
+      <td>when auth type is kerberos,whether do the kerberos proxy behavior with the login user (if not anonymous) to get the connection, default value is true</td>
   </tr>
   <tr>
     <td>default.jceks.file</td>
--- a/docs/interpreter/jdbc.md
+++ b/docs/interpreter/jdbc.md
@@ -166,7 +166,7 @@ There are more JDBC interpreter properties you can specify like below.
   </tr>
   <tr>
       <td>zeppelin.jdbc.auth.kerberos.proxy.enable</td>
-      <td>true or false. when auth type is kerberos, whether do the kerberos proxy behavior with the login user (if not anonymous) to get the connection, default value is true</td>
+      <td>When auth type is Kerberos, enable/disable Kerberos proxy with the login user to get the connection. Default value is true.</td>
   </tr>
   <tr>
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
@wary Thanks for your contribution. Could you add `zeppelin.jdbc.auth.kerberos.proxy.enable` info to [this table](https://github.com/apache/zeppelin/blob/master/docs/interpreter/jdbc.md#more-properties) as well?

wary:
@AhyoungRyu thank for reminding，the `zeppelin.jdbc.auth.kerberos.proxy.enable` info has been added to the table in jdbc.md

AhyoungRyu:
Thanks! Looks good to me.

prabhjyotsingh:
Tested locally, LGTM!

docs/interpreter/jdbc.md
```diff
@@ -165,6 +165,10 @@ There are more JDBC interpreter properties you can specify like below.
     <td>The path to the keytab file</td>
   </tr>
   <tr>
+      <td>zeppelin.jdbc.auth.kerberos.proxy.enable</td>
+      <td>true or false. when auth type is kerberos, whether do the kerberos proxy behavior with the login user (if not anonymous) to get the connection, default value is true</td>
```
felixcheung: should we document the default value if omitted? (seems to be true)

AhyoungRyu: @wary Sorry for my late blocking but when I saw this sentence, I couldn't understand at a glance. Could you make this description:
"when auth type is kerberos, whether do the kerberos proxy behavior with the login user (if not anonymous) to get the connection, default value is true" more understandable? Something like

"When auth type is Kerberos, enable/disable Kerberos proxy with a logged user to get the connection. Default value is true." 

What do you think?

wary: yeah, I think you are right,  your description is more understandable, thx for your advice，It has been updated

AhyoungRyu:
@wary Thanks! 
Merge into master and branch-0.7 if there are no more comments

PR #2223, opened 2017-04-05
[ZEPPELIN-2355] Livy cancel enhancements

benoyantony:
### What is this PR for?
The Cancel functionality for the Livy interpreter has a couple of issues. One issue is because a variable is not published correctly. Second issue is observed when  there is a delay in launching the application. Any cancel before application launch is ignored.  The third issue is that Cancel is not correctly implemented for SparkSQLInterpreter.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2355

The test cases are modified to test the changes.

### Questions:


benoyantony:
@zjffdu , Could you please review  ?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -162,21 +165,8 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
-        }
-      } catch (LivyException e) {
-        LOGGER.error("Fail to cancel statement " + stmtId + " for paragraph " + paraId, e);
-      } finally {
-        paragraphId2StmtIdMap.remove(paraId);
-      }
-    } else {
-      LOGGER.warn("cancel is not supported for this version of livy: " + livyVersion);
-    }
+    paragraphsToCancel.add(context.getParagraphId());
```
zjffdu: put it in if block to make the adding operation consistent with remove

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
```diff
@@ -284,12 +275,25 @@ public InterpreterResult interpret(String code,
...
       }
     }
   }
 
+  private void cancel(int id, String paragraphId) {
+    if (livyVersion.isCancelSupported()) {
+      try {
+        LOGGER.info("Cancelling statement " + id);
+        cancelStatement(id);
+        paragraphsToCancel.remove(paragraphId);
```
zjffdu: It it better to put it in finally block to ensure it is removed.

zjffdu:
@benoyantony Thanks for the contribution, just left some comments.

benoyantony:
@zjffdu , Will try to add the unit test for cancellation in LivySparkSQLInterpreter. Any suggestion on how to simulate a sleep in SQL ? 
Also, where are the other comments ? Sorry, I am not very familiar with the github code reviews. It may be very obvious.

zjffdu:
@benoyantony You can call cancel in a separate thread, the same as testSparkInterpreterRDD, let me know if you have any other questions.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
benoyantony: I have put the removal in finally block so that paragraph is removed from the set even if cancellation is not supported in the livy version.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
benoyantony: Since livyVersion is available only after the session is initialized, putting it inside the if block will work based on whether livyVersion is available or not. In some cases, this , might even throw NullPointerException.  

I think , it's better to simply add it and then remove it later. 
If the livyVersion supports cancellation, we will cancel and remove.
If not supported, we will warn and remove.

benoyantony:
A testcase errored out and it seems to be unrelated.
ParagraphActionsIT.testRunOnSelectionChange

livy/src/test/java/org/apache/zeppelin/livy/LivyInterpreterIT.java
```diff
@@ -306,6 +306,70 @@ public void testSparkSQLInterpreter() {
...
+              Thread.sleep(1);
+            } catch (InterruptedException e) {
+              e.printStackTrace();
+            }
+            sqlInterpreter.cancel(context);
+          }
+        };
+        cancelThread.start();
+        boolean bCancelled = false;
+        for (int i=0; i <1000; i++) {
```
zjffdu: Why to iterate 1000 times ?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
zjffdu: Good catch, we can clear paragraphsToCancel if it is not supported.

zjffdu: Good catch, we can clear paragraphsToCancel if it is not supported.

livy/src/test/java/org/apache/zeppelin/livy/LivyInterpreterIT.java
benoyantony: @zjffdu , I  do not a way to sleep inside a sql statement. So the approach was to run up to 1000 sql statements while CancelThread will call cancel on the paragraph. If the cancel works, then at least one of the statements will be cancelled and the test will be successful.

Please let me know if there is a better approach.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterprereter.java
benoyantony: Yes, I will do that.

livy/src/test/java/org/apache/zeppelin/livy/LivyInterpreterIT.java
zjffdu: hmm, I think you don't need to run cancel in a separated thread. You can just call it as following to test.  And don't use `select count(1) from table_name` instead, as `show tables` might not invoke spark jobs.

```
sqlInterpreter.cancel(context);
result = sqlInterpreter.interpret("select count(1) from table_name", context);
```

benoyantony: This makes sense. I have modified the test case to make above call. Still calling the cancel on a separate thread to simulate a real scenario and to make it similar to other test cases.

zjffdu:
Thanks @benoyantony will merge it if no more comments.

benoyantony:
Thanks @zjffdu for reviewing and committing.

PR #1940, opened 2017-01-25
[ZEPPELIN-2008] Introduce Spell

1ambda:
### What is this PR for?

Implemented **Spell** as one of Helium categories. *Technically, it's the frontend interpreter* runs on browser not backend.

Spell can provide many benefits.

1. Anyone can install, remove easily using helium package registry by #1936 
2. Implementing spell is extremely easier rather than adding backend interpreter
3. Can use existing javsacript libraries. (e.g [flowchart.js](http://flowchart.js.org/), [sequence diagram js](https://github.com/bramp/js-sequence-diagrams), ...). This enable us to add many visualization tools. Imagine that you can implement some custom interpreters with few lines of code like [flowchart-spell-example](https://github.com/apache/zeppelin/compare/master...1ambda:ZEPPELIN-2008/introduce-spell?expand=1#diff-364845b20d68e4d94688e44fef03da98)
4. The most important thing is, spell is not only interpreter but also display system. Because it runs on browser. So we can use spell display system with another spell **Display System with Spell** (see the screenshot section below)

 **In future**, we will be able to combine existing backend interpreters with spell like (**not supported in this PR cause we need to modify backend code a lot**)

```
// if we have markdown spell, we can use `%markdown` display in the spark interpreter

%spark

val calculated = doSomething()
println(s"%markdown _${calculated})
```

I added some examples. Checkout `echo`, `markdown`, `translator`, `flowchart` spells.

### What type of PR is it?
[Feature]

### Todos
* [x] - Add `SPELL` as one of Helium categories.
* [x] - Implement framework code (`zeppelin-spell`)
* [x] - Make some examples (flowchart, google translator, markdown, echo)
* [x] - Support custom display system
* [x] - Fix some bugs in `HeliumBundleFactory`
* [x] - Save spell rendering result into `note.json` while broadcasting to other websocket clients
* [x] - Fix `renderText` for stream output

### What is the Jira issue?

[ZEPPELIN-2008](https://issues.apache.org/jira/browse/ZEPPELIN-2008)

### How should this be tested?

- Build `mvn clean package -Phelium-dev -Pexamples -DskipTests;`
- Go to helium page `http://localhost:8080/#/helium`
- Enable all spells
- Go to a notebook and refresh
- Follow actions in the screenshots below.

#### Flowchart Spell (Sample)

flowchart-spell

#### Google Translator Spell (Sample)

translator-spell

#### Display System with Spell

display-spell

### Questions:
* Does the licenses files need update - NO
[...]

PR #2227, opened 2017-04-05
[ZEPPELIN-2359] Support Spell as Display

echarles:
### What is this PR for?

SPELL should be considered as another display type, just like TEXT, HTML, TABLE...

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-2359

Run in a paragraph

```
val json = """{"array":[1,2,3],"boolean":true,"null":null,"number":123,"object":{"a":"b","c":"d","e":"f"},"string":"Hello World"}"""
println(s"""%spell %json
$json""")
```

or

```
val hello = "hello all"
println(s"""%spell %echo
$hello""")
```

### Screenshots

NA

### Questions:


Leemoonsoo:
Usage of existing display systems is `%DISPLAY_SYSTEM_NAME` (e.g. `%table`, `%html`). Will where be a way to keep usage the same whether display system is installed by Spell or not?

Leemoonsoo:
\cc @1ambda

echarles:
@Leemoonsoo I wanted to introduce minimum changes in javascript and stay aligned with the idea of interpreters returning different display types. Introducing a new `spell` display has sounded logic and easy to me. But I agree that in this case, user has still to add a second, potentially confusing, magic keyword to define the targeted spell.

The other way would be to consider that if interpreter returns a value that begins with `%`,  we consider it as a spell... @1ambda WDYT?

1ambda:
Thanks for contribution. @echarles 

It would be nice to keep same usage for spell even with backend display system. That's the original intention (https://github.com/apache/zeppelin/pull/1940). Here are few reasons.

- Users might not want to learn new usage for just using display system. 
- Even some users don't care it's spell or not. 
- Sometimes, we need complex implementation to provide simple usage for users. I believe this is the case.

echarles:
@Leemoonsoo @1ambda I see your point and agree with it.

However, we need to decide which InterpreterResult to return. For now it can be { TABLE, IMG, SVG, NULL }. The Java code needs to decide which value to return to the Javascript, and the Javascript based on that value needs to take decisions (display as a known display-type, or render code via Spell).

For now, I don't see how to implement / recognize on Java side the InterpreterResult to return.

Can you describe any option you have in mind?

1ambda:
Hi @echarles. Sorry for late comment. 

I agree it's quite hard to implement. Since `InterpreterOutput` class is widely used.

1. We need to add a variable to `RemoteInterpreterContext` which is a thrift generated class to pass   currently available spell magics so that interpreter can use it for their `InterpreterOutput`

2. Since InterpreterOutput usage depends on interpreter type, we might use static, synchronized variable in `InterpreterOutput` both sides in `ZeppelinServer` and `Interpreter`.

This is what i am thinking about.

echarles:
@1ambda thx a lot for the follow-up and your great work.
This branch now supports what we want:

```
println("""%json {"array":[1,2,3],"boolean":true,"null":null,"number":123,"object":{"a":"b","c":"d","e":"f"},"string":"Hello World"}""")
```
I would be great if you could test it and give me any feedback.

json-spell from scala

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOutput.java
```diff
@@ -229,8 +229,11 @@ public void write(int b) throws IOException {
               return;
             }
           }
-          // not a defined display system
+          // not a defined display system, probably a spell type
+          setType(InterpreterResult.Type.SPELL);
```
1ambda: We can't assume that It's always the spell type.  Since log can include random text. 
What do you think of?

1ambda: IMO, we can pass available spell magics (e.g `%markdown`, `%sigma`) whenever executing paragraph. And it can be passed into `InterpreterOutput`. Then, we don't need to worry about.

echarles:
@1ambda I have added the log. Regarding you second comment, I am not sure what you mean. Is there something I should change? Is it not working as expected?

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
```diff
@@ -326,41 +326,50 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
...
    */
-  $scope.renderCustomDisplay = function (type, data) {
-    // get result from intp
+  $scope.renderCustomDisplay = function(type, data) {
+
+    var splits = data.split(' ');
+    type = data.substring(0, splits[0].length)
+    data = data.substring(splits[0].length, data.length)
+
     if (!heliumService.getSpellByMagic(type)) {
```
1ambda: What i mean is, if text (not paragraph) includes `%fancytext`, it will throw an error like 

> `%fancytext` display doesn't exist. 

In this case, user doesn't want to use `%fancytext` display and it doesn't exist. Just text includes that kind of text. 

LINE 335.

PR #2229, opened 2017-04-06
[Zeppelin 2367] Hive JDBC proxy user option should be available even without kerberos

prabhjyotsingh:
### What is this PR for?
Hive JDBC proxy user option should be available generically.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [Zeppelin 2367](https://issues.apache.org/jira/browse/ZEPPELIN-2367)

Enable Shiro authentication and set `zeppelin.jdbc.auth.type` as `SIMPLE` in the interpreter setting, and observe the connection string for the Hive.

### Questions:


r-kamath:
Tested. LGTM.

docs/interpreter/hive.md
```diff
@@ -161,3 +171,9 @@ GROUP BY ${group_by=product_id,product_id|product_name|customer_id|store_id}
 ORDER BY count ${order=DESC,DESC|ASC}
 LIMIT ${limit=10};
 ```
+
+### Impersonation
+When Zeppelin server is running with authentication enabled, then this interpreter utilizes Hive's user proxy feature i.e. sends extra parameter for creating and running a session ("hive.server2.proxy.user=": "${loggedInUser}"). This is particularly useful when multi users are sharing a Notebook server.
```
felixcheung: `multi users` -> `multiple users`?

felixcheung: does this affect resource allocation, such as each user could be on different YARN queue and so on?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -421,13 +411,31 @@ public Connection run() throws Exception {
...
+            connection = getConnectionFromPool(connectionUrl.toString(),
+                user, propertyKey, properties);
       }
     }
     propertyKeySqlCompleterMap.put(propertyKey, createSqlCompleter(connection));
     return connection;
   }
 
+  private void checkAndAppendHiveProxyUser(StringBuilder connectionUrl, String user) {
+    if (connectionUrl.toString().trim().startsWith("jdbc:hive")) {
```
felixcheung: hmm.. is this the right approach and abstraction, to have specialize type in a general JDBC interpreter?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -421,13 +411,31 @@ public Connection run() throws Exception {
...
+      Integer lastIndexOfUrl = connectionUrl.indexOf("?");
+      if (lastIndexOfUrl == -1) {
+        lastIndexOfUrl = connectionUrl.length();
+      }
+
+      if (user != null && !user.equals("anonymous") &&
+          !"false".equalsIgnoreCase(property.getProperty("hive.proxy.user"))) {
+        logger.debug("Using hive proxy user");
```
felixcheung: logger.info?
also, log the `user` value?

docs/interpreter/hive.md
```diff
@@ -124,6 +124,16 @@ Hive Interpreter will be deprecated and merged into JDBC Interpreter. You can us
     <td></td>
     <td><b>( Optional ) </b>Other properties used by the driver of <code>%hive(${prefix})</code> </td>
   </tr>
+  <tr>
```
AhyoungRyu: @prabhjyotsingh Hive docs is deprecated: https://github.com/apache/zeppelin/blob/master/docs/interpreter/hive.md#important-notice

So it'll be better to add this info to https://github.com/apache/zeppelin/blob/master/docs/interpreter/jdbc.md#apache-hive instead.

docs/interpreter/hive.md
```diff
@@ -124,6 +124,16 @@ Hive Interpreter will be deprecated and merged into JDBC Interpreter. You can us
...
+    <td>zeppelin.jdbc.auth.type</td>
+    <td></td>
+    <td><b>( Optional ) </b>If auth type is needed, Example: KERBEROS/SIMPLE</td>
+  </tr>
+  <tr>
+    <td>hive.proxy.user</td>
+    <td>true</td>
+    <td><b>( Optional ) </b>If want to use `hive.server2.proxy.user`</td>
```
AhyoungRyu: These back quoters won't work properly in `<table>` tag. So please use `<code>` tag like this:
`<code>hive.server2.proxy.user</code>`

docs/interpreter/hive.md
prabhjyotsingh: Sure, then I'll revert my changes from this file, and apply it on jdbc.md

docs/interpreter/hive.md
prabhjyotsingh: Yes, that could happen, but it was never documented, hence thought to document it here.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: I thought about other approaches, like using `propertyKey`, but even in that case name could be anything including default.

docs/interpreter/hive.md
prabhjyotsingh: Thank you @AhyoungRyu , have taken care of this in jdbc.md.

prabhjyotsingh:
@AhyoungRyu @felixcheung thank you for the review,  have made the suggested changes.

prabhjyotsingh:
@AhyoungRyu, @felixcheung ping

prabhjyotsingh:

felixcheung:
From my earlier comment, I'm interested in others thought on having very Hive specific logic in the generic JDBC interpreter?

https://github.com/apache/zeppelin/pull/2229/files#r110512765

prabhjyotsingh:
Yes, I thought of few options but this was bit con

prabhjyotsingh:
@felixcheung any specific change that you are looking for?

felixcheung:
Perhaps generalize this to something non-Hive specific? Like a property for userStringToAppend? Add user to JDBC URL should be common place - seems like in this case we just like a way to change the name and value of the property to add on the JDBC URL.

prabhjyotsingh:
Sure, got it, have renamed it to "appendProxyUserToURL".

felixcheung:
actually, I wasn't referring to the function name at all. instead of having this fairly hive specific, how about a function like this to take the user name property name and value 

```
// Untested!
 private void appendProxyUserToURL(StringBuilder connectionUrl, String user) {
     if ("true".equalsIgnoreCase(property.getProperty("jdbc.url.set.user") {
       Integer lastIndexOfUrl = connectionUrl.indexOf("?");
       if (lastIndexOfUrl == -1) {
         lastIndexOfUrl = connectionUrl.length();
       }
 
       if (user != null && !user.equals("anonymous") &&
           property.getProperty("jdbc.url.user.property").length() > 1)  { // check not empty or whitespace instead
         logger.info("Using jdbc user as :" + user);
         connectionUrl.insert(lastIndexOfUrl, ";" + property.getProperty("jdbc.url.user.property") + "=" + user + ";");
       }
     }
   }
```

prabhjyotsingh:
@felixcheung Sure what you have recommended does make a lot of sense, have implemented what you have suggested.

prabhjyotsingh:
CI is green; 
https://travis-ci.org/prabhjyotsingh/zeppelin/builds/225173700 
https://travis-ci.org/prabhjyotsingh/zeppelin/builds/225214327

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -353,74 +353,68 @@ public Connection getConnection(String propertyKey, InterpreterContext interpret
...
+                  public Connection run() throws Exception {
+                    return getConnectionFromPool(connectionUrl, user, poolKey, properties);
+                  }
+                });
+              } catch (Exception e) {
+                logger.error("Error in doAs", e);
+                StringBuilder stringBuilder = new StringBuilder();
+                stringBuilder.append(e.getMessage()).append("\n");
+                stringBuilder.append(e.getCause());
+                throw new InterpreterException(stringBuilder.toString());
```
felixcheung: why not just include e as inner exception like `InterpreterException(message, e)`?

prabhjyotsingh: Sure just did it, thank you for the hint.

prabhjyotsingh:
> why is zeppelin.jdbc.auth.kerberos.proxy.enable removed?

My bad, while fixing the above issues, I thought this is redundant, but I forgot about the case where KERBEROS is enabled and the user does not want to use proxy behavior. Have reverted that change.

> also is there a way not to hardcode hive jdbc [here](https://github.com/apache/zeppelin/pull/2229/files#diff-ecdae8ee9594a5c4b21a3c217a3f130cR365)

Hive doesn't work with [this](https://github.com/apache/zeppelin/pull/2229/files#diff-ecdae8ee9594a5c4b21a3c217a3f130cR370) (UserGroupInformation.createProxyUser), and since we allow user to configure multiple JDBC sources using the same interpreter setting; there can be a case where user has configured say Phoenix (requires UserGroupInformation.createProxyUser to proxy user)  and Hive (fails if UserGroupInformation.createProxyUser is used), we could introduce new parameter like `{profile}.kerberos.proxy.enable` but this will increase the documentation part.

prabhjyotsingh:
@felixcheung  have removed hard coded string "hive" as well.


CI fails for; not relavent for this change.
```
Tests in error: 
  ParagraphActionsIT.testClearOutputButton:335->AbstractZeppelinIT.clickAndWait:126 Â» ElementNotVisible
```

prabhjyotsingh:
@felixcheung  ping

docs/interpreter/jdbc.md
```diff
@@ -507,8 +507,9 @@ Here are some examples you can refer to. Including the below connectors, you can
...
   </tr>
   <tr>
-    <td>hive.proxy.user</td>
-    <td>true or false</td>
+    <td>default.proxy.user.property</td>
+    <td>hive.server2.proxy.user</td>
+  </tr>
 </table>
 
 Connection to Hive JDBC with a proxy user can be disabled with `hive.proxy.user` property (set to true by default)
```
felixcheung: is `hive.proxy.user` removed then? do we need to doc to explain the migration?
also you have this line here still with `hive.proxy.user property (set to true by default)`

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -179,10 +178,6 @@ public void open() {
     }
     logger.debug("JDBC PropretiesMap: {}", basePropretiesMap);
 
-    if (!isEmpty(property.getProperty("zeppelin.jdbc.auth.type"))) {
```
felixcheung: this is also removed? `zeppelin.jdbc.auth.type`?

docs/interpreter/jdbc.md
felixcheung: I guess the preference would be to have a way to migrate existing users and config by having it automatically converted, add a log warning etc.

but it's your call in the specific case

docs/interpreter/jdbc.md
```diff
@@ -534,6 +535,14 @@ Connection to Hive JDBC with a proxy user can be disabled with `hive.proxy.user`
 
 [Maven Repository : org.apache.hive:hive-jdbc](https://mvnrepository.com/artifact/org.apache.hive/hive-jdbc)
 
+##### Impersonation
+When Zeppelin server is running with authentication enabled, then the interpreter can utilize Hive's user proxy feature i.e. send extra parameter for creating and running a session ("hive.server2.proxy.user=": "${loggedInUser}"). This is particularly useful when multiple users are sharing a notebooks.
+
+To enable this set following:
+  - `zeppelin.jdbc.auth.type` as `SIMPLE` or `KERBEROS` (if required) in the interpreter setting.
+  - `default.proxy.user.property` as `hive.server2.proxy.user`
```
felixcheung: should we clarify, `default` here could also be the name of the hive interpreter instance like `hive.proxy.user.property`?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: `JDBCSecurityImpl.createSecureConfiguration` call is moved to line#358 (https://github.com/apache/zeppelin/pull/2229/files/9fee9d2a9e476769b5d28a50d17f3a7ad38fc453#diff-ecdae8ee9594a5c4b21a3c217a3f130cR358)

felixcheung: ah ok, doc change?

docs/interpreter/jdbc.md
prabhjyotsingh: What I meant in above line was the key is `default.proxy.user.property` with an example value `hive.server2.proxy.user`; fair point, let me tweak this documentation. 

Yes, adding a log (logger.warn) does make sense, let me do it right away.

docs/interpreter/jdbc.md
prabhjyotsingh: Sure, let me change this to `{propertyKey}.proxy.user.property` and post a sample configuration.

docs/interpreter/jdbc.md
prabhjyotsingh: I was looking at http://zeppelin.apache.org/docs/0.7.1/interpreter/jdbc.html#apache-hive, `hive.proxy.user` hasn't made it to the doc section, so I think in this case no need to add migration steps. However, I have added `logger.warn` as well.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: Since it's just a removal of extra call, so no doc change required.

docs/interpreter/jdbc.md
prabhjyotsingh: Thank you for reviewing this. Have handled your review comments.

prabhjyotsingh:
@felixcheung Have handled your review comments. Let me know if I missed any.

prabhjyotsingh:
Thank you @felixcheung. Will merge this to master.

PR #2231, opened 2017-04-07
ZEPPELIN-2150. NoSuchMethodError: org.apache.spark.ui.SparkUI.appUIAddress() for the spark master

zjffdu:
### What is this PR for?
`SparkUI.addUIAddress` is removed in spark master which cause this error. Actually spark 2.0 introduce new api `SparkContext.uiWebUrl` (SPARK-14576) which is public in contrast `SparkUI.addUIAddress` is private. This PR would use `SparkUI.addUIAddress` for spark 1 and `SparkContext.uiWebUrl`  for spark 2.

### What type of PR is it?
[Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2150

Manually verify it on spark 1.6.2, spark 2.1.0 and spark master

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -999,10 +999,19 @@ public class SparkInterpreter extends Interpreter {
     if (sparkUrl != null) {
       return sparkUrl;
     }
-    Option<SparkUI> sparkUiOption = (Option<SparkUI>) Utils.invokeMethod(sc, "ui");
-    SparkUI sparkUi = sparkUiOption.get();
-    String sparkWebUrl = sparkUi.appUIAddress();
-    return sparkWebUrl;
+
+    if (sparkVersion.newerThanEquals(SparkVersion.SPARK_2_0_0)) {
+      Option<String> uiWebUrlOption = (Option<String>) Utils.invokeMethod(sc, "uiWebUrl");
+      if (uiWebUrlOption.isDefined()) {
+        return uiWebUrlOption.get();
+      }
+    } else {
+      Option<SparkUI> sparkUIOption = (Option<SparkUI>) Utils.invokeMethod(sc, "ui");
+      if (sparkUIOption.isDefined()) {
+        return (String) Utils.invokeMethod(sparkUIOption.get(), "appUIAddress");
+      }
+    }
+    return "";
   }
 
   private Results.Result interpret(String line) {
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -1011,7 +1011,7 @@ public class SparkInterpreter extends Interpreter {
         return (String) Utils.invokeMethod(sparkUIOption.get(), "appUIAddress");
       }
     }
-    return "";
+    return null;
   }
 
   private Results.Result interpret(String line) {


zjffdu:
@felixcheung @karuppayya please help review.

prabhjyotsingh:
LGTM, manually verify it on spark 1.6.1, spark 2.0.0 and spark master.

zjffdu:
Thanks for review. @prabhjyotsingh

Leemoonsoo:
LGTM

felixcheung:
LGTM
Is "" handled properly when returned?

zjffdu:
Good catch! @felixcheung It would be better to return null, fixed that.

qwemicheal:
This problem still exist in LivyPySparkBaseInterpreter.java. This file still use sc._jsc.sc().ui().get().appUIAddress()  and %livy.pyspark doesn't work when using spark 2.2.

zjffdu:
Thanks for reporting, created ZEPPELIN-3269

liusztc09:
@zjffdu 

I run into the same issue with Livy 0.4 and Spark 2.2.1.  There is a comment by you saying it's only affecting livy 0.2 in https://issues.apache.org/jira/browse/ZEPPELIN-3269. Not sure whether it's fixed yet.

Seems the old method was still being called. 
https://github.com/apache/zeppelin/search?utf8=%E2%9C%93&q=appUIAddress%28%29+&type=

Any workarounds or ETA on the fix?

Thanks!

rouckas:
I was also running into this issue with Zeppelin 0.7.3, Livy 0.5.0 (0.4.0), and Spark 2.3.0 (or 2.2.1). The only solution I found was to downgrade to Zeppelin 0.6.2.

zjffdu:
@rouckas I try zeppelin master & livy 0.5, it works for me. Do you mind to try it in zeppelin master ?

thesuperzapper:
@zjffdu what version of spark were you using, as this only occurs >=2.2

It looks like we still try to call (the now removed as of Spark 2.2) "appUIAddress":
https://github.com/apache/zeppelin/blob/master/livy/src/main/java/org/apache/zeppelin/livy/LivySparkInterpreter.java#L51

Meaning its pretty likely that this issue still occurs in master for >=2.2. (I cant check in my environment)

ererermao:
I also encountered this problem on aws (emr 5.13.0).
zeppelin(0.7.3)、livy(0.4.0)、saprk(2.3.0)

thesuperzapper:
@zjffdu @felixcheung @ererermao there seems to be no traction in this, it hasn't worked for a long time, https://issues.apache.org/jira/browse/ZEPPELIN-3455.

zjffdu:
@thesuperzapper @ererermao It works for me on zeppelin master + livy 0.5.0 + spark 2.3.0

screen shot 2018-05-23 at 9 08 02 am

DerekKaknes:
I'm having this same issue with zeppelin 0.7.3 + livy 0.5.0 + spark 2.2.1.  I have not yet verified that zeppelin master resolves this issue, but from inspection I would think that @thesuperzapper is right that the issue would persist.  I would think that any invocation of `SparkUI.appUIAddress()` would throw a `NoSuchMethodException`, and zeppelin master still seems to be making those calls (specifically from `LivySparkInterpreter`).  @zjffdu Do you have any notion as to why your interpreter would not be throwing an exception?

zjffdu:
That method would only be called in livy 0.2. In livy 0.2 the spark web url will be returned from livy rest api, so that piece of code won't be invoked. And zeppelin has system test for that, not sure why you would hit this kind of issue. 

Are you running livy in local mode ? It is suggested to run livy in yarn-cluster mode.

DerekKaknes:
Thanks for following-up.  I'm running Livy connected to a standalone cluster, but in digging deeper it's possible that there is an issue with Livy always configuring itself to local mode when building spark sessions.  I'll have to look at that first to see if it resolves this issue as well.

SivaKaviyappa:
This is works for me on aws emr. I am on emr 5.13.0, livy 0.40 , spark 2.3.0.  The important change one has to make is by default aws emr brings up livy in yarn mode. So if you change that yarn-cluster mode and stop livy server, zeppelin (emr automatically restarts the services once stopped) then this above issue does not happen. Thanks to @zjffdu for pointing that out.

# What spark master Livy sessions should use.
livy.spark.master                yarn-cluster

zjffdu:
Thanks @SivaKaviyappa for confirming

yywwd:
@zjffdu  I change the yarn mode into "yarn-cluster" as SivaKaviyappa suggested, Zeppelin works well. However, when I use programmatic API, it will throw such exception: ImportError: No module named cloudpickle.cloudpickle. If I change the yarn mode into default mode, that is "yarn", programmatic API will works well, but Zeppelin still cannot work. My cluster is EMR-5.14.0, Application: Ganglia 3.7.2, Spark 2.3.0, Zeppelin 0.7.3, Livy 0.4.0

zjffdu:
Could you share your code that use programmatic API ?

yywwd:
@zjffdu I'm sorry to response slowly, because I tried to reproduce the bugs. I thought it may caused by my codes, so I tried to use official PySpark examples, the bugs still occurred. This is the code I used. https://github.com/apache/incubator-livy/blob/master/examples/src/main/python/pi_app.py
**Note:** I comment the last line # client.stop(True), cause I won't want to close the session after submitting just one job. The details about this bug are as follows:

1. When I use default yarn mode, that is "yarn", official PySpark examples and programmatic API work well. Using Livy interpreter in Zeppelin will throw  exception:` NoSuchMethodError: org.apache.spark.ui.SparkUI.appUIAddress()` for the spark master.

2. ThenI change the yarn mode into "yarn-cluster" as SivaKaviyappa suggested, Zeppelin works well. But the logs of this statement will hav a warning: `"Warning: Master yarn-cluster is deprecated since 2.0. Please use master \"yarn\" with specified deploy mode instead.`
However, Using programmatic API will have such bug:  

  2.1  I delete all Livy sessions, and run pi_app.py. It throw such exception:  
```
ReadTimeout: HTTPConnectionPool(host='172.31.5.251', port=8998): Read timed out. (read timeout=10)
  File "/home/ec2-user/wandongwu/livy_test_9/pi_app.py", line 35, in <module>
    pi = client.submit(pi_job).result()
 File "/usr/local/lib/python2.7/site-packages/concurrent/futures/_base.py", line 462, in result
    return self.__get_result()
  File "/usr/local/lib/python2.7/site-packages/concurrent/futures/_base.py", line 414, in __get_result
    raise exception_type, self._exception, self._traceback
TypeError: raise: arg 3 must be a traceback or None
```
But I find It has started a new Livy session. So I edit configuration parameter into `'http://<livy-server>:8998/sessions/0' 2`. It will throw another exception:
```
  File "/home/ec2-user/wandongwu/livy_test_9/pi_app.py", line 35, in <module>
    pi = client.submit(pi_job).result()
  File "/usr/local/lib/python2.7/site-packages/concurrent/futures/_base.py", line 462, in result
    return self.__get_result()
  File "/usr/local/lib/python2.7/site-packages/concurrent/futures/_base.py", line 414, in __get_result
    raise exception_type, self._exception, self._traceback
Exception: org.apache.livy.repl.PythonJobException: Client job error:Traceback (most recent call last):
  File "/mnt/yarn/usercache/livy/appcache/application_1528945006613_0302/container_1528945006613_0302_01_000001/tmp/4991895008696585180", line 159, in processBypassJob
    deserialized_job = pickle.loads(serialized_job)
  File "/usr/lib64/python2.7/pickle.py", line 1388, in loads
    return Unpickler(file).load()
  File "/usr/lib64/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/lib64/python2.7/pickle.py", line 1096, in load_global
    klass = self.find_class(module, name)
  File "/usr/lib64/python2.7/pickle.py", line 1130, in find_class
    __import__(module)
ImportError: No module named cloudpickle.cloudpickle
```

3. Then I change yarn mode into default mode, that is "yarn", programmatic API can work well, but Zeppelin still cannot work.

zjffdu:
Sorry, I still not sure what code you run in zeppelin. If it is livy related issue, you need to ask question in livy mail list.

yywwd:
@zjffdu This issue is not related to the codes run in zeppelin. I mean, I cannot guarantee “Using Livy interpreter in Zeppelin”  and "Using Programmatic API" work well at the same time.

Using Zeppelin, I just write codes like this:
```
%livy.pyspark
sc.version
```
just test whether it works.


Using Programmatic API, I just use official examples: https://github.com/apache/incubator-livy/blob/master/examples/src/main/python/pi_app.py

**BUT!!!** 
In **livy.spark.master yarn-cluster** mode zeppelin works, programmatic API cannot work
In **livy.spark.master yarn** mode (the default mode), programmatic API works, zeppelin cannot work

zjffdu:
What kind of error you see in zeppelin when using livy.spark.master yarn mode

yywwd:
@zjffdu NoSuchMethodError: org.apache.spark.ui.SparkUI.appUIAddress()
My AWS cluster is EMR-5.14.0, Application: Ganglia 3.7.2, Spark 2.3.0, Zeppelin 0.7.3, Livy 0.4.0

SivaKaviyappa:
@yywwd - Few things to check
What are your zeppelin-env.sh and livy intrepreter settings
1.have you set the PYTHONPATH in the zeppelin-env.sh
2. zeppelin.livy.url=http://localhost:8998 - have set this value in your livy intrepreter?
3.  you have installed python libraries in all the core nodes?
I am using emr 5.11.0 and everything works fine.

yywwd:
I have solved this problem by editing some configurations in Livy. SivaKaviyappa's solution: setting "livy.spark.master yarn-cluster" does not work for me. I do not edit "livy.spark.master" in livy.conf. Instead,  I uncomment "livy.spark.deploy-mode" and set the value to cluster. The edited configuration is:
# What spark deploy mode Livy sessions should use.
livy.spark.deploy-mode           cluster

PR #2235, opened 2017-04-08
[ZEPPELIN-2375]: Avoid modification of CLASSPATH variable.

cfries:
The script previously modified the CLASSPATH variable. This may lead to
an undesired side effect, where the zeppelin server classpath is
exported to the zeppelin interpreter classpath, see JIRA issue
ZEPPELIN-2375 for details. Instead of modifying classpath we now work
the other way around and set ZEPPELIN_CLASSPATH or
ZEPPELIN_INT_CLASSPATH to include CLASSPATH and then use
ZEPPELIN_CLASSPATH or ZEPPELIN_INT_CLASSPATH in the corresponding exec
(runner), respectively.

See also https://issues.apache.org/jira/browse/ZEPPELIN-2375

### What is this PR for?

Fixing issue https://issues.apache.org/jira/browse/ZEPPELIN-2375

### What type of PR is it?

Bug Fix

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-2375

Without this fix, the following fails on macOS Sierra (OS X):

1.  Open bash, set CLASSPATH to empty string and export it using `export CLASSPATH=`, then start zeppelin via `./bin/zepplin.sh`

2. Create and run a notebook with the md interpreter.

3. Results in an exception.

4. Stop zeppelin. Repeat the test with this fix.

See https://issues.apache.org/jira/browse/ZEPPELIN-2375

### Questions:
No

No

No

### DIFF
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -64,7 +64,7 @@ fi
 
 . "${bin}/common.sh"
 
-ZEPPELIN_INTP_CLASSPATH=""
+ZEPPELIN_INTP_CLASSPATH="${CLASSPATH}"
 
 # construct classpath
 if [[ -d "${ZEPPELIN_HOME}/zeppelin-interpreter/target/classes" ]]; then
@@ -190,8 +190,6 @@ fi
 
 addJarInDirForIntp "${LOCAL_INTERPRETER_REPO}"
 
-CLASSPATH+=":${ZEPPELIN_INTP_CLASSPATH}"
-
 if [[ ! -z "$ZEPPELIN_IMPERSONATE_USER" ]]; then
     suid="$(id -u ${ZEPPELIN_IMPERSONATE_USER})"
     if [[ -n  "${suid}" || -z "${SPARK_SUBMIT}" ]]; then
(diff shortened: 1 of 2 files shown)


felixcheung:
hmm, I can see your approach though I can also see changing CLASS_PATH makes sense in some cases (for example, a custom library to be loaded).

Do you think there is way to accommodate that? also, is there a way this could be configurable to be backward compatible?

cfries:
Hi Felix.

I am not sure if I got your point. The user can change CLASSPATH and zeppelin.sh will use that. Also, the user can change ${ZEPPELIN_CLASSPATH_OVERRIDES} and ${ZEPPELIN_INTP_CLASSPATH_OVERRIDES} to add additional files to the class path without changing CLASSPATH. I assume my change is backward compatible. It is just that the script does not change the CLASSPATH variable. But since the script is not supposed to call another script, this should be OK.
In my option the side effect that the zeppelin.sh is polluting CLASSPATH with the value of ZEPPELIN_CLASSPATH, which is then exported to interpreter.sh is a bug, because it does not allow to use a JAR in zeppelin server which should NOT be used in zeppelin interpreter. With my modification this is possible.
Again: w.r.t. backward compatibility: I would assume that no one should rely on the bug and if he relied on it, he can fix it by setting ZEPPELIN_INTP_CLASSPATH_OVERRIDES or ZEPPELIN_INTP_CLASSPATH.

cfries:
Also: on OS X the zeppelin does not run if you export CLASSPATH. So I assume no body could rely on this issue.

Leemoonsoo:
LGTM

felixcheung:
great, I review again after rereading what you are saying and I think they make sense and starting ZEPPELIN_CLASSPATH with CLASSPATH should make in compatible

felixcheung:
merging if no more comment

felixcheung:
@cfries looks like this PR is now including other changes? Could you rebase?

cfries:
@Leemoonsoo Ups. Sorry for that. Will fix asap...

Leemoonsoo:
CI build passed.
https://travis-ci.org/cfries/zeppelin/builds/224499519

PR #2237, opened 2017-04-10
ZEPPELIN-2377. Hive Support can not be enabled in spark master

zjffdu:
### What is this PR for?
The root cause is that `org.apache.spark.sql.hive.HiveSessionState` is removed in spark master. I change it to `org.apache.spark.sql.hive.execution.InsertIntoHiveTable` which is existed early in spark 1.0. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2377

Verify it manually in spark master, spark 2.1.0 and spark 1.6.2 

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -246,7 +246,7 @@ public class SparkInterpreter extends Interpreter {
    */
   private boolean hiveClassesArePresent() {
     try {
-      this.getClass().forName("org.apache.spark.sql.hive.HiveSessionState");
+      this.getClass().forName("org.apache.spark.sql.hive.execution.InsertIntoHiveTable");
       this.getClass().forName("org.apache.hadoop.hive.conf.HiveConf");
       return true;
     } catch (ClassNotFoundException | NoClassDefFoundError e) {


zjffdu:
@Leemoonsoo @felixcheung Please help review

felixcheung:
I guess it's fine - but we are looking fairly deep into internal stuff. Why don't we just look for HiveConf?

Or SparkSession.hiveClassesArePresent if Spark 2.0+

zjffdu:
`HiveConf` only means the hive jar is loaded, but doesn't mean spark is built with hive profile enabled. 

`org.apache.spark.sql.hive.HiveSessionState` is private in spark and `org.apache.spark.sql.hive.execution.InsertIntoHiveTable` is public which mean it should be stable and won't be removed in spark 2.x I believe. `SparkSession.hiveClassesArePresent` is private, so I don't think it is stable.

felixcheung:
For
hiveClassesArePresent

I mean either call hiveClassesArePresent (it's private yes) or do what it does?

zjffdu:
I don't prefer to call private method `hiveClassesArePresent` which is not guaranteed to be stable. The cause of hive support broken in spark master is due to we copied the implementation of `hiveClassesArePresent` of spark 2.0 in zeppelin, but `hiveClassesArePresent` is changed in spark master. So instead of checking difference classes for different versions, here I propose to checking class `InsertIntoHiveTable` which is public.

felixcheung:
I don't disagree - we have other dependencies on private methods though?

Ok I'm fine with this approach. Thanks for the explanation.

zjffdu:
Thanks @felixcheung for review, will merge it into master if no more discussion.

PR #849, opened 2016-04-21
ZEPPELIN-825: Set appropriate default permissions and allow user to c…

rohannr:
### What is this PR for?
This PR makes users who set only r/w permissions the effective owners of the note.
1. If you set readers, and writers and owners is empty, it is automatically set to the user requesting the change.
2. If you set writers, and owners is empty, it is set it to the user requesting the change.
It also fixes a bug that did not allow one to clear the permissions for a note. 

### What type of PR is it?
Bug Fix, Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-825

### How should this be tested?
Outline the steps to test the PR here.

### Questions:
* Does this needs documentation? YES

…lear permissions

This change makes users who set permissions the effective owners of the note.
1. If you set readers, and writers and owners is empty, it is automatically set to the user requesting the change.
2. If you set writers, and owners is empty, it is set it to the user requesting the change.

PR #2239, opened 2017-04-10
Show dialog if owners field set to be empty on shiro

soralee:
### What is this PR for?
This PR is for a showing dialog when note's permission set `Readers/Writers` but `Owners` is empty. 
Currently, after merged #849, if user set `Owners` is empty, it's automatically set as the current user.
So, It would be better to show a dialog before `Owners` is fill with the current user because user might think it is not intended. 

Plus, if a user only set `Readers` role, all field(`Writers` and `Owners`) role is filled as the current user like below. It's not necessary to fill because the `Owner` role can read and write note.
not_need_wirters

### What type of PR is it?
[ Improvement ]

### What is the Jira issue?
* [ZEPPELIN-2206; Be added a writer account when adding a reader account on Shiro](https://issues.apache.org/jira/browse/ZEPPELIN-2206)
* [ZEPPELIN-2418; show dialog if user set to the empty owner permission](https://issues.apache.org/jira/browse/ZEPPELIN-2418)

1. Turn on shiro.
2. Set permission `Writers` and `Readers` except `Owner` role.
3. Check the dialog appears

[Before]
* Set `Readers` and `Writers` except `Owners`
zeppelin-2206_auto_shiro_b_1

* Set only 'Writers`
zeppelin-2206_auto_shiro_b_2

* set only `Readers`; this screen is attached above.

[After]
* Set `Readers` and `Writers` except `Owners` ( Set only `Writers`; this result is same)
zeppelin-2206_auto_shiro_a_1

* Set only 'Readers`
zeppelin-2206_auto_shiro_a_2

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -208,9 +208,6 @@ public class NotebookRestApi {
     HashSet<String> writers = permMap.get("writers");
     // Set readers, if writers and owners is empty -> set to user requesting the change
     if (readers != null && !readers.isEmpty()) {
-      if (writers.isEmpty()) {
-        writers = Sets.newHashSet(SecurityUtils.getPrincipal());
-      }
       if (owners.isEmpty()) {
         owners = Sets.newHashSet(SecurityUtils.getPrincipal());
       }
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -761,7 +761,40 @@ function NotebookCtrl ($scope, $route, $routeParams, $location, $rootScope,
   }
 
   $scope.savePermissions = function () {
+    if ($scope.isAnonymous || $rootScope.ticket.principal.trim().length === 0) {
+      $scope.blockAnonUsers()
+    }
     convertPermissionsToArray()
+    if ($scope.isOwnerEmpty()) {
+      BootstrapDialog.show({
+        closable: false,
+        title: 'Setting Owners Permissions',
+        message: 'Please fill the [Owners] field. If not, it will set as current user.\n\n' +
+          'Current user : [ ' + $rootScope.ticket.principal + ']',
+        buttons: [
+          {
+            label: 'Set',
+            action: function(dialog) {
+              dialog.close()
+              $scope.permissions.owners = [$rootScope.ticket.principal]
+              $scope.setPermissions()
+            }
+          },
+          {
+            label: 'Cancel',
+            action: function(dialog) {
+              dialog.close()
+              $scope.openPermissions()
+            }
+          }
+        ]
+      })
+    } else {
+      $scope.setPermissions()
+    }
+  }
+
+  $scope.setPermissions = function() {
     $http.put(baseUrlSrv.getRestApiBase() + '/notebook/' + $scope.note.id + '/permissions',
       $scope.permissions, {withCredentials: true})
     .success(function (data, status, headers, config) {
@@ -769,7 +802,7 @@ function NotebookCtrl ($scope, $route, $routeParams, $location, $rootScope,
         console.log('Note permissions %o saved', $scope.permissions)
         BootstrapDialog.alert({
           closable: true,
-          title: 'Permissions Saved Successfully!!!',
+          title: 'Permissions Saved Successfully',
           message: 'Owners : ' + $scope.permissions.owners + '\n\n' + 'Readers : ' +
           $scope.permissions.readers + '\n\n' + 'Writers  : ' + $scope.permissions.writers
         })
@@ -876,6 +909,20 @@ function NotebookCtrl ($scope, $route, $routeParams, $location, $rootScope,
   angular.element(document).click(function () {
     angular.element('.ace_autocomplete').hide()
   })
+
+  $scope.isOwnerEmpty = function() {
+    if ($scope.permissions.owners.length > 0) {
+      for (let i = 0; i < $scope.permissions.owners.length; i++) {
+        if ($scope.permissions.owners[i].trim().length > 0) {
+          return false
+        } else if (i === $scope.permissions.owners.length - 1) {
+          return true
+        }
+      }
+    } else {
+      return true
+    }
+  }
 
   /*
    ** $scope.$on functions below


soralee:
@felixcheung Thanks for review! 
Yeah, you're right. It would be better if we also check on the REST API side stably.
So, let me create a JIRA ticket and revert code in `REST API`.

soralee:
Hi, @felixcheung :slightly_smiling_face: 
I added JIRA url about this improvement in this PR description and I reverted some code in `notebook Rest API` side except that, if only setting `Readers` role, all field(Writers and Owners) role is filled as the current user. Because the `Owners` role can already read and write note, we don't need to fill in `Writers` role.

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -750,6 +750,35 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
 
   $scope.savePermissions = function() {
     convertPermissionsToArray();
+    if ($scope.isOwnerEmpty()) {
+      BootstrapDialog.show({
+        closable: false,
+        title: 'warning Permissions',
```
felixcheung: it's a bit odd if this is the display title of the dialog `warning Permissions`?
perhaps just `Setting Owners Permissions`?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -750,6 +750,35 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
...
+        message: 'Please fill the [Owners] field. If not, it will set as current user.\n\n' +
+          'Current user : [ ' + $rootScope.ticket.principal + ']',
+        buttons: [
+          {
+            label: 'OK',
```
felixcheung: instead of having an `OK` button to set/change permission, how about having it as `Set` to be more clear?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -861,6 +890,14 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
     }
   };
 
+  $scope.isOwnerEmpty = function() {
+    if ($scope.permissions.owners.length > 0){
```
felixcheung: should we check if it's just whitespace? `       `

1ambda:
Can we remove `!!!` from dialog title? it's meaningless and noisy. although it's not @soralee's change, i think we can handle it in this PR.

soralee:
@felixcheung Thanks for detail suggestion! That's sounds good. Let me update as each comment.
@1ambda Sure, I agree! I'll remove `!!!`

1ambda:
@soralee Since conflicts are about lint errors, you can simply rebase by

1. keeping your change and run `rebase --continue`
2. then, execute  `eslint src --fix` in `zeppelin-web` dir.

soralee:
@1ambda Thanks for information always :+1: 
I already worked to rebase and run build successfully, but now, I'm checking the `white space` problem.
Let me updated ASAP!

soralee:
I modified dialog title and `OK` button.
I've checked the `whitespace` and the `$scope.permissions.owners` array don't insert `whitespace` originally. But for stability, I added `if statement`. 
@felixcheung, @1ambda  could you guys review this again, please?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -867,6 +896,20 @@ function NotebookCtrl ($scope, $route, $routeParams, $location, $rootScope,
     angular.element('.ace_autocomplete').hide()
   })
 
+  $scope.isOwnerEmpty = function() {
+    if ($scope.permissions.owners.length > 0) {
+      for (let i = 0; i < $scope.permissions.owners.length; i++) {
+        $scope.permissions.owners[i] = $scope.permissions.owners[i].trim()
```
felixcheung: check `$scope.permissions.owners[i].trim()` doesn't return empty string before setting it?

felixcheung: and what if we have to skip some owners before it's empty/whitespace - might wan to copy to a new list / keep the left hand side index separately?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -752,14 +752,43 @@ function NotebookCtrl ($scope, $route, $routeParams, $location, $rootScope,
 
   $scope.savePermissions = function () {
     convertPermissionsToArray()
+    if ($scope.isOwnerEmpty()) {
+      BootstrapDialog.show({
+        closable: false,
+        title: 'Setting Owners Permissions',
+        message: 'Please fill the [Owners] field. If not, it will set as current user.\n\n' +
+          'Current user : [ ' + $rootScope.ticket.principal + ']',
```
felixcheung: not sure if we need to check $rootScope.ticket.principal  is empty or whitespace

soralee:
@felixcheung Thanks for review and useful comments!
I initially think something wrong. Let me modify this part again!
Thanks a lots!

zeppelin-web/src/app/notebook/notebook.controller.js
soralee: @felixcheung thanks for carefully checking this.
I didn't care this `$rootScope.ticket.principal` variable, because this `$scope.savepermissions` function is not run when user doesn't use Siro and user didn't login. 
But, for prevent, I would be better to add `if statement`for checking about this case that `$rootScope.ticket.principal` is ayonymous or empty.

zeppelin-web/src/app/notebook/notebook.controller.js
soralee: @felixcheung Thanks for review and useful comments!
You're right! I initially think something wrong. Let me modify this part again!
Thanks a lots!

soralee:
@khalidhuseynov Thanks for comments!
You mean, this can be occurred another problem like the following steps. 
1. A user who has admin permissions sets `readers` and `writers` and empty `owners` roles in `Note permissions'.
2. The user clicks 'Save' button in `Note permissions`
3. The user clicks `Cancel button` in the new dialog(`Setting Owners Permissions`) which I made.
3. then, the default `owners` value is disappeared.
4. and, the user add `user1` to `owners` role and click 'Save' button.

As my understanding your comment, this 3 and 4 steps can be a problem. right?
@khalidhuseynov, Could you check what I understood correctly, please?

khalidhuseynov:
@soralee yeah the steps are correct. so if you follow those steps and click `Save`, say for `Note A`, then `Note A` will disappear from the list of your notes, and then you will not have permissions to view `Note A`

soralee:
@khalidhuseynov thanks for checking it really!
You're right! I think so too! this should be considered.
Furthermore, I checked above steps which I addressed, I saw that zeppelin run strange operation.  
Let me improve it so that it does not disappear.

khalidhuseynov:
@soralee it's disappearance is based on multi-user logic and it's correct since if note have `owners`, `readers`, `writers` as `user1` it doesn't have anything to do in `admin`'s list. That's why I was telling that not sure that this PR makes sense, because of intervening into that flow. From another point when creating note, it's `owner` is automatically assigned as current user without asking as well, so I think having already existing logic and assigning current user as owner (unless having admin feature) is fine, and we don't need much changes on it for now

soralee:
@khalidhuseynov Thanks for opinion! Yeah, I see. This PR is for showing a dialog when (already existing backend logic) that note's permission set Readers/Writers but Owners is empty. I think I would better to focus the purpose! 
So, next step is that I'd like to update this PR as @felixcheung addressed.

soralee:
@felixcheung I updated as your advice!
@khalidhuseynov I added a logic which is reverted with old value if user cancel button like above `4 step`. How about this?

> 4. then, the default owners value is disappeared.

revert_shiro

soralee:
Ping!
And this CI fail is irrelevant this PR.
```
Tests in error: 
  SecurityRestApiTest.testGetUserList:69 » JsonSyntax java.lang.IllegalStateExce...
  SecurityRestApiTest.testTicket:55 » JsonSyntax java.lang.IllegalStateException...
Tests run: 58, Failures: 5, Errors: 2, Skipped: 1
```

soralee:
Hi, there.
CI became green now.

soralee:
Ping!
@felixcheung @khalidhuseynov @1ambda, could you guys confirm again, please?

soralee:
Ready to review!

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -208,9 +208,6 @@ public Response putNotePermissions(@PathParam("noteId") String noteId, String re
     HashSet<String> writers = permMap.get("writers");
     // Set readers, if writers and owners is empty -> set to user requesting the change
     if (readers != null && !readers.isEmpty()) {
-      if (writers.isEmpty()) {
-        writers = Sets.newHashSet(SecurityUtils.getPrincipal());
```
felixcheung: I might have missed this, why are we making this change?

soralee: Thanks for good point!
This line `if` statement mean that when filling `readers` and not filling `writers`, it's filled with `writers` as current user. In this case, I think, it's enough for filling with `owners` as current user. 
And, the following screenshot and explain is in this PR description.
> Plus, if a user only set Readers role, all field(Writers and Owners) role is filled as the current user like below. It's not necessary to fill because the Owner role can read and write note.
https://cloud.githubusercontent.com/assets/8110458/24869746/2a77d2da-1e4f-11e7-86a5-794e1aac02d0.gif

soralee:
@felixcheung, regarding your comment 
> hey sorry, so in your screen shot here #2239 (comment)
if cancel when owner is empty it resets also writer/reader, is that a bit odd?

that screenshot means it is reverted with old value if user click `cancel` button.
So, it's describing that 
 1. original data is only `owners: admin`, 
 2. and change data of roles
 3. and then click `cancel` button.
 4. make sure data is reverted by original data (`owners: admin`)

soralee:
reopen for CI

khalidhuseynov:
Merging into master if no more discussion

PR #2240, opened 2017-04-11
ZEPPELIN-2386. Add parameter for check running current paragraph in ZeppelinContext

zjffdu:
### What is this PR for?
I can reproduce it via the test code in ZeppelinIT.testAngularDisplay. First run the 4 paragraphs to initiate the state. Then when I click paragraph 0 the first time, paragraph 2 will run correctly. But if I click paragraph 0 again, interpreter would raise the following exception. The cause is that ZeppelinContext's interpreterContext now point to paragraph 2, because paragraph 2 is the last paragraph that runs in this interpreter. So I propose to add parameter for check running current paragraph in ZeppelinContext.

* Add 2 methods in ZeppelinContext.   `run(idx, checkCurrentParagraph) `, 'run(noteId, idx, context, checkCurrentParagraph)'
* Also make some changes for debug logging. 

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2386

Tested manually and also add integration test. 

Before
zeppelin_before

After
zeppelin_after

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -311,7 +311,7 @@ public class ZeppelinContext {
    */
   @ZeppelinApi
   public void run(String noteId, String paragraphId) {
-    run(noteId, paragraphId, interpreterContext);
+    run(noteId, paragraphId, interpreterContext, true);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


zjffdu:
@Leemoonsoo @jongyoul  Please help review.

Leemoonsoo:
@zjffdu 

In ZeppelinIT.testAngularDisplay, 4th paragraph creates watcher and watcher call `z.run()` without passing interpreter context. like

```scala
z.angularWatch("myVar", (before:Object, after:Object, 
  context:org.apache.zeppelin.interpreter.InterpreterContext)=>{ 
    z.run(2)
  }
)
```

If we pass interpreter context is passed to `z.run()`, the error does not occur. like

```scala
z.angularWatch("myVar", (before:Object, after:Object, 
  context:org.apache.zeppelin.interpreter.InterpreterContext)=>{ 
    z.run(context.getNoteId, 2, context)
  }
)
```

@zjffdu Do you think will there a way `z.run()` takes `InterpreterContext` automatically? If that's possible, then user does not need to care about `checkCurrentParagraph`.

zjffdu:
@Leemoonsoo Thanks for suggestion, the second approach works, but I didn't find a way to take `InterpreterContext` automatically. Another scenario is that user may do need to run the current paragraph. In the following code, I want to rerun the current paragraph if user click the link. 

```
println("%angular <div id=\'angularTestButton\' ng-click=\'myVar=myVar+1\'>BindingTest_{{myVar}}_</div>")

z.angularBind("myVar", 0)
z.angularWatch("myVar", "watcher_1", (before:Object, after:Object, context:org.apache.zeppelin.interpreter.InterpreterContext)=>{
    z.run("20170410-172601_2116783047")
})
```

Leemoonsoo:
Thanks for explain.
LGTM

zjffdu:
Thanks @Leemoonsoo for review, will merge to master if no more comments

PR #2241, opened 2017-04-11
ZEPPELIN-2390. Improve returnType for z.checkbox

zjffdu:
### What is this PR for?
Currently it is not convenient to access the individual item of the return value of z.checkbox, I would propose to return `Seq` for `SparkInterpreter` and `list` for `PySparkInterpreter`.  This might cause some incompatibility, but I think it is acceptable considering the benefits.  Besides that, before this PR, all the items of checkbox would be checked by default in `PySparkInterpreter` which is inconsistent with `SparkInterpreter`, so I change it to nothing is selected by default in this PR. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2390

Unit test is added

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -136,7 +136,7 @@ public class ZeppelinContext {
   }
 
   @ZeppelinApi
-  public scala.collection.Iterable<Object> checkbox(String name,
+  public scala.collection.Seq<Object> checkbox(String name,
       scala.collection.Iterable<Tuple2<Object, String>> options) {
     List<Object> allChecked = new LinkedList<>();
     for (Tuple2<Object, String> option : asJavaIterable(options)) {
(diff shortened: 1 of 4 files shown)


zjffdu:

Leemoonsoo:
LGTM

zjffdu:
Will merge to master if no more discussion.

felixcheung:
A bit concerned with breaking API changes like this but I guess we have never defined the support or versioning plan on the API?

Aren't we breaking the selected default item for checkbox by initializing
defaultChecked to empty?

It was default to first item in checkbox
Now is default to nothing

zjffdu:
Before this PR It was not default to first item in checkbox, it is default to all items. The reason I change it to nothing is to make it consistent with scala implementation. 

Right, this would break the existing API, but I think it is worthwhile. I have several following PRs to improve `ZeppelinContext` and hope to stabilize it. IMO `ZeppelinContext` would play an important role for extending the rich functionality of zeppelin.

felixcheung:
Ok sounds ok to me

zjffdu:
Thanks for review, will merge to master if no more discussion.

PR #2249, opened 2017-04-14
ZEPPELIN-2356. Improvement for z.angularBind and z.angularWatch

zjffdu:
### What is this PR for?
There're 2 issues in the current angular object usage in `ZeppelinContext` when user run the paragraph with code `z.angularBind` and `z.angularWatch` multiple times （It is not idempotent for the paragraph)
* `z.angularBind` would set a new value if the angular object is existed (the second run of paragraph), so the registered watcher will be called which might not be what user expect
* `z.angularWatch` would register the same watcher multiple times if the angular object is existed (the second run), this also might not be what user expect.

I made 2 changes for this issue.
* Remove the existing angular object and create a new one in `angularBind`, and create new method `angularSet` for setting new value for angular object.
* Add watcherId in `AngularObjectWatcher` so that we can uniquely identify watcher, and won't add duplicated watcher.

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2356

Test is added.

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/ZeppelinContext.java
@@ -578,6 +578,26 @@ public class ZeppelinContext {
     angularBind(name, o, null, watcher);
   }
 
+  @ZeppelinApi
+  public void angularSet(String name, Object o) {
+    angularSet(name, o, interpreterContext.getNoteId());
+  }
+
+  @ZeppelinApi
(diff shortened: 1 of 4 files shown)


Leemoonsoo:
@zjffdu do you mind update corresponding docs, too?

`docs/displaysystem/back-end-angular.md`

And front-end expose angular display system apis. Do you think this guy need to support new api, too?
`docs/displaysystem/front-end-angular.md`

PR #2250, opened 2017-04-14
[ZEPPELIN-2085] Interpret scala code in paste mode

DrIgor:
### What is this PR for?
Don't split scala code to lines.

Interpreter code becomes more and more complicated because some lines have to be executed
together (companion objects, etc.). And still it doesn't handle all possibilities.

I propose not to split interpreter input at all.
I checked scala repl sources: repl paste mode interpret code without splitting.

### What type of PR is it?
[Bug Fix | Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2085

Try to run several code snippets attached to jira issue.
See also https://issues.apache.org/jira/browse/ZEPPELIN-658

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -905,14 +905,17 @@ public class SparkInterpreter extends Interpreter {
 
       interpret("import org.apache.spark.SparkContext._");
 
+      if (Utils.isScala2_10()) {
+        // see https://issues.apache.org/jira/browse/SPARK-14146
+        System.setProperty("spark.repl.fallback", "true");
+      }
+
       if (importImplicit()) {
         if (Utils.isSpark2()) {
           interpret("import spark.implicits._");
           interpret("import spark.sql");
           interpret("import org.apache.spark.sql.functions._");
         } else {
-          // see https://issues.apache.org/jira/browse/SPARK-14146
-          System.setProperty("spark.repl.fallback", "true");
           if (sparkVersion.oldSqlContextImplicits()) {
             interpret("import sqlContext._");
           } else {
@@ -1198,23 +1201,14 @@ public class SparkInterpreter extends Interpreter {
     out.setInterpreterOutput(replOutput);
     context.out.clear();
 
-    scala.tools.nsc.interpreter.Results.Result res;
     try {
-      res = interpret(code);
+      scala.tools.nsc.interpreter.Results.Result res = interpret(code);
 
       Code r = getResultCode(res);
 
       if (r == Code.ERROR) {
         return new InterpreterResult(r, "");
-      }
-
-      // make sure code does not finish with comment
-      if (r == Code.INCOMPLETE) {
-        res = interpret(code + "\nprint(\"\")");
-        r = getResultCode(res);
-      }
-
-      if (r == Code.INCOMPLETE) {
+      } else if (r == Code.INCOMPLETE) {
         return new InterpreterResult(r, "Incomplete expression");
       } else {
         putLatestVarInResourcePool(context);
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.spark;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -32,6 +33,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import com.google.common.base.Joiner;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.spark.SparkConf;
 import org.apache.spark.SparkContext;
@@ -909,6 +911,8 @@ public class SparkInterpreter extends Interpreter {
           interpret("import spark.sql");
           interpret("import org.apache.spark.sql.functions._");
         } else {
+          // see https://issues.apache.org/jira/browse/SPARK-14146
+          System.setProperty("spark.repl.fallback", "true");
           if (sparkVersion.oldSqlContextImplicits()) {
             interpret("import sqlContext._");
           } else {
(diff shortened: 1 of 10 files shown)


1ambda:
Using paste mode would be better. We have so many if-else code to manipulate code which is already covered by paste mode. 

If we evaluate using paste mode, evaluation orders will be affected. If it doesn't matter or not that important, i think it's ok to use paste mode.

zjffdu:
+1 for not splitting scala code in zeppelin side, that should be handled by scala repl.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1189,91 +1191,48 @@ public InterpreterResult interpret(String[] lines, InterpreterContext context) {
...
-    for (int l = 0; l < linesToRun.length; l++) {
-      String s = linesToRun[l];
-      // check if next line starts with "." (but not ".." or "./") it is treated as an invocation
-      if (l + 1 < linesToRun.length) {
-        String nextLine = linesToRun[l + 1].trim();
-        boolean continuation = false;
-        if (nextLine.isEmpty()
-           || nextLine.startsWith("//")         // skip empty line or comment
-           || nextLine.startsWith("}")
-           || nextLine.startsWith("object")) {  // include "} object" for Scala companion object
```
zjffdu: Do you have test for this scenario (Scala companion object) ? If it is broken, then we should either fix it or document it.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1189,91 +1191,48 @@ public InterpreterResult interpret(String[] lines, InterpreterContext context) {
...
+      } else {
+        putLatestVarInResourcePool(context);
+        return new InterpreterResult(Code.SUCCESS);
+      }
+    } catch (Exception e) {
+      logger.info("Interpreter exception", e);
+      return new InterpreterResult(Code.ERROR, InterpreterUtils.getMostRelevantMessage(e));
+    } finally {
+      try {
+        context.out.setType(InterpreterResult.Type.TEXT);
```
zjffdu: Why would it must be TEXT ? Isn't other type possible ?

DrIgor: This writes REPL response to output. RELP always respond with plain text

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
DrIgor: Added test for https://issues.apache.org/jira/browse/ZEPPELIN-658

DrIgor:
Finally done PR and travis is green

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: Could you check the case for `println("%html <h1>hello</h1>")` ?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -909,6 +911,8 @@ public void open() {
           interpret("import spark.sql");
           interpret("import org.apache.spark.sql.functions._");
         } else {
+          // see https://issues.apache.org/jira/browse/SPARK-14146
+          System.setProperty("spark.repl.fallback", "true");
```
zjffdu: Isn't needed for spark2 as well ?

zjffdu: BTW, do we have any test for the scenario described in SPARK-14146 ?

zjffdu: We may need to add TODO when SPARK-14146 is fixed

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1162,118 +1166,72 @@ boolean isUnsupportedSparkVersion() {
...
 
-    // make sure code does not finish with comment
-    if (r == Code.INCOMPLETE) {
-      scala.tools.nsc.interpreter.Results.Result res = null;
-      res = interpret(incomplete + "\nprint(\"\")");
-      r = getResultCode(res);
-    }
+      // make sure code does not finish with comment
+      if (r == Code.INCOMPLETE) {
+        res = interpret(code + "\nprint(\"\")");
```
zjffdu: Could it possible be "}" ?

zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
```diff
@@ -136,23 +145,25 @@ public void sparkSQLTest() throws IOException {
...
-            p.setText("%spark val df=sqlContext.createDataFrame(Seq((\"hello\",20)))\n" +
-                    "z.show(df)");
+            p.setText("%spark val df=sqlContext.createDataFrame(Seq((\"hello\",20)))\nz.show(df)");
             p.setAuthenticationInfo(anonymous);
             note.run(p.getId());
             waitForFinish(p);
-            assertEquals(Status.FINISHED, p.getStatus());
-            assertEquals(InterpreterResult.Type.TABLE, p.getResult().message().get(1).getType());
-            assertEquals("_1\t_2\nhello\t20\n", p.getResult().message().get(1).getData());
+            assertThat(p.getStatus(), is(Status.FINISHED));
```
zjffdu: This seems some kind of code style change, it may not be necessary, because I think we should reach consensus among community to make our code style consistent. Open source project also require code style as consistent as possible

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1162,118 +1166,72 @@ boolean isUnsupportedSparkVersion() {
...
-      sc.clearJobGroup();
-      putLatestVarInResourcePool(context);
+      if (r == Code.INCOMPLETE) {
+        return new InterpreterResult(r, "Incomplete expression");
+      } else {
+        putLatestVarInResourcePool(context);
+        return new InterpreterResult(Code.SUCCESS);
+      }
+    } catch (Exception e) {
+      logger.info("Interpreter exception", e);
```
felixcheung: logger.warn?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
felixcheung: this might not be a problem now?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
felixcheung: right, I think we might have different type for handling DataFrame print out too

DrIgor: I checked it:
image

"hello" is printed to `context.out` and REPL output is printed to temporary buffer. Then it is printed to `context.out` as a TEXT

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
DrIgor: I can't understand why we need to check if code is finished with comment.

Unfortunately git history doesn't have any links to jira task. It would be great if you can explain me the case!

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
DrIgor: Implicit imports are covered by tests.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: I don't' have context for that, but IMO we can remove this.

Leemoonsoo: I remember this code was just a trick to handle the case that paragraph finish with comment, such as

```
val a = 1
// comment
```

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
DrIgor: You are right, I added scala version check

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
DrIgor: I removed this code. 
```
val a = 1
// comment
```
works as expected

DrIgor:
Please explain what issues I need to fix)

zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
jongyoul: Agreed

jongyoul:
I think we'd better take care of the comments left.

jongyoul:
I'm a bit confused because there's some unrelated changes regarding this issue. I don't know why you change zeppelin-interpreter package and why you make hash code and equals method. Can you make your changes more smaller? I think you add some minor changes regardless of this issue. Please divide those into another minor issues

PR #2251, opened 2017-04-14
[ZEPPELIN-2403] added interpreter property types

tinkoff-dwh:
### What is this PR for?
Added types to properties of interpreters. Those are types for properties customization. Properties must have the ability to customize the display (for example password).

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2403

- remove conf/interpreter.json
- Try new form (create, edit) of interpreter settings

add
add

edit
edit

view
view

### Questions:
* Is there breaking changes for older versions?  no


Leemoonsoo:
Thanks @tinkoff-dwh for useful feature! 

Tested this PR. In interpreter setting page, I can change type of property.

interpreter_property_type

But shouldn't type supposed to be immutable?

And i think we can add more types like 'number', 'boolean', 'url'. Do you have any plan adding more types?

tinkoff-dwh:
Yes, probably better to make immutable. In the first stage there were no plans to add more although i can add boolean (checkbox)

felixcheung:
probably wouldn't be a good idea to change from password to plaintext?

tinkoff-dwh:
add
n_add

edit
n_edit

view
n_view

1ambda:
One suggestion. What about calling them as 'widget' or something instead of 'type'? cause we might add actual type like 'number', 'string',... as @leemoonsu mentioned.

tinkoff-dwh:
@1ambda 
widget is ok. About another widgets: angular-xeditable has no implementations for that (there are html5 inputs but old browsers not support html5)

zjffdu:
Is it compatible with previous interpreter setting without type ? Maybe use text by default if no type is specified.

tinkoff-dwh:
@zjffdu 
not compatible (conf/notebook.json), another format of 'properties' (object instead of string) https://github.com/tinkoff-dwh/zeppelin/blob/dd5d6c809aeb2bdb72357a072954cf4a9c882487/docs/install/upgrade.md

tinkoff-dwh:
edit
edit

view
view

1ambda:
@tinkoff-dwh Since conflicts are about lint errors, you can simply rebase by

1. keeping your change and run `rebase --continue`
2. then, execute  `eslint src --fix` in `zeppelin-web` dir.

PR #2254, opened 2017-04-16
ZEPPELIN-2407. Livy Interpreter always return plain text result

zjffdu:
### What is this PR for?
It happens when zeppelin.livy.displayAppInfo is true. Straightforward fix. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2407

Test is added

### Questions:


zjffdu:
Thanks @felixcheung , will merge it if no more comments.

PR #2263, opened 2017-04-19
ZEPPELIN-2421. AngularObject miss AngularObjectListener when it is pushed from frontend

zjffdu:
### What is this PR for?
This bug happens when you save your note, and restart zeppelin and run it again. The root cause is that the angular object miss  AngularObjectListener which means you can not sync up the changes on angular object to frontend. This PR would restore the listener correctly when we restore the angular objects saved before.  Also make some changes for logging in this PR. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2421

Tested manually

Before
zeppelin_before

After
zeppelin_after

### Questions:


zjffdu:
@Leemoonsoo Please help review.

Leemoonsoo:
Thanks @zjffdu for the fix.
LGTM!

zjffdu:
Thanks @Leemoonsoo , will merge it if no more comments

PR #2269, opened 2017-04-21
[ZEPPELIN-2434] Credential feature does work in JDBC interpreter

Leemoonsoo:
### What is this PR for?
Credential feature does not work in JDBC interpreter.

This PR fixes unittest that does not detect this bug and fix jdbc interpreter to correctly read credential.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix unittest
* [x] - Fix condition when use credential, when use property.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2434

Set empty `default.user` property and set id/pw in 'credential' menu. And try use jdbc interpreter.

### Questions:


Leemoonsoo:
Merge to master and branch-0.7 if no further discussions.

PR #2276, opened 2017-04-21
[ZEPPELIN-2410]: Using UUID from filename in registry cache dir inste…

cfries:
…ad of uri.

### What is this PR for?

Fixing issue ZEPPELIN-2410:

The previous version of this file used the full uri as file name in the
cache. This included special characters like the : in http: which
caused problems.

Since this is just a cache file, a better approach is to use an UUID.

The UUID is constructed from the URI to avoid the cache dir to grow
over time (in case there is no clean up of the cache dir).

Output of `ls -la` on macOS before the fix:
```
drwxr-xr-x  4 fries  staff  136 17 Apr 22:28 .
drwxr-xr-x  8 fries  staff  272 21 Apr 12:43 ..
drwxr-xr-x  3 fries  staff  102 17 Apr 10:39 https:
```

Output of `ls -la` on macOS after the fix:
```
drwxr-xr-x  4 fries  staff    136 21 Apr 17:32 .
drwxr-xr-x  8 fries  staff    272 21 Apr 17:32 ..
-rw-r--r--  1 fries  staff  89666 21 Apr 17:32 a6618b3e-540a-340e-b624-07bf7f2b5e7d
```

Note: Windows 7 just fails to create the cache before the fix.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-2410?filter=-2

Non-Regression Test.
Run on Windows, check cache dir.

N/A

### Questions:
No
No
No

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumOnlineRegistry.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumOnlineRegistry.java
@@ -32,6 +32,7 @@ import java.net.URL;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 /**
  * This registry reads helium package json data
@@ -55,7 +56,10 @@ public class HeliumOnlineRegistry extends HeliumRegistry {
   public HeliumOnlineRegistry(String name, String uri, File registryCacheDir) {
     super(name, uri);
     registryCacheDir.mkdirs();
-    this.registryCacheFile = new File(registryCacheDir, name);
+
+    UUID registryCacheFileUuid = UUID.nameUUIDFromBytes(uri.getBytes());
+    this.registryCacheFile = new File(registryCacheDir, registryCacheFileUuid.toString());
+
     gson = new Gson();
   }
 


zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumOnlineRegistry.java
```diff
@@ -55,7 +56,10 @@
   public HeliumOnlineRegistry(String name, String uri, File registryCacheDir) {
     super(name, uri);
     registryCacheDir.mkdirs();
-    this.registryCacheFile = new File(registryCacheDir, name);
+
+    UUID registryCacheFileUuid = UUID.nameUUIDFromBytes(uri.getBytes());
```
felixcheung: sounds like `UUID.nameUUIDFromBytes` hashes consistently - ie. given the same uri it will return the same uuid?

cfries: Yes, that was the intention to use `namedUUIDFromBytes` instead of `randomUUID`. But `randomUUID` would also work if the cache dir is purged (but I was not sure if that is guaranteed).

To be precise: nameUUIDFromBytes implement variant 3 from http://www.ietf.org/rfc/rfc4122.txt:

> 4.3.  Algorithm for Creating a Name-Based UUID
>
>    The version 3 or 5 UUID is meant for generating UUIDs from "names"
>    that are drawn from, and unique within, some "name space".  The
>    concept of name and name space should be broadly construed, and not
>    limited to textual names.  For example, some name spaces are the
>    domain name system, URLs, ISO Object IDs (OIDs), X.500 Distinguished
>    Names (DNs), and reserved words in a programming language.  The
>    mechanisms or conventions used for allocating names and ensuring
>    their uniqueness within their name spaces are beyond the scope of
>    this specification.
> 
>    The requirements for these types of UUIDs are as follows:
> 
>    o  The UUIDs generated at different times from the same name in the
>       same namespace MUST be equal.
> 
>    o  The UUIDs generated from two different names in the same namespace
>       should be different (with very high probability).
> 
>    o  The UUIDs generated from the same name in two different namespaces
>       should be different with (very high probability).
> 
>    o  If two UUIDs that were generated from names are equal, then they
>       were generated from the same name in the same namespace (with very
>       high probability).

If we would like to ensure different names, we should use randomUUID, but then the cache dir will pile up new cache files unless it is purged on a regular basis. Since I was not sure if that is the case, I proposed the "hash-like" nameUUIDFromBytes.

Leemoonsoo:
LGTM

Leemoonsoo:

PR #2278, opened 2017-04-24
[ZEPPELIN-2431] Corrected deletion of notes by incorrect interpreter.json

cloverhearts:
## What is this PR for?

We sometimes can not delete a note, or we will be accompanied by an NPE for deleting a note.

This problem occurs when:
When interpreter.json 's note binding is wrong, or there is a problem.
If you are configuring an interpreter that is not through zeppelin's user interface.
As a result, it happens when synchronization of notes deletion and setting retention is not normal.
Therefore, we should add handling for note deletion and exception handling for nonexistent interpreter bindings.
It reduces the synchronization problem of interpreter.json.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2431

1. zeppelin stop
2. edit con/interpreter.json and `interpreterBindings`
  fix any notes or incorrect information.
  for example 
  ```
    },  
  "interpreterBindings": {
    "2CFS9YSM5": [
      "2CFRR1D3TINVALIDINVALIDINVALID", <-- edit
      "2CFZ1JKUR",
      "2CEAJK1VW",
      "2CGSESWBH",
      "2CERNPGW5",
   }
  ```
3. zeppelin start
4. You can try remove invalid interpreter bind note on web. (on example = `2CFS9YSM5`)

result : 
If the modifications to this PR are not reflected,
It will not be deleted or an error will appear on the server.
Also, the interpreterBindings information in interpreter.json does not respond to delete events.

problem animation
cantremove

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -887,14 +887,18 @@ public class InterpreterSettingManager {
     saveToFile();
   }
 
-  public void removeNoteInterpreterSettingBinding(String user, String noteId) {
+  public void removeNoteInterpreterSettingBinding(String user, String noteId) throws IOException {
     synchronized (interpreterSettings) {
       List<String> settingIds = (interpreterBindings.containsKey(noteId) ?
           interpreterBindings.remove(noteId) :
           Collections.<String>emptyList());
       for (String settingId : settingIds) {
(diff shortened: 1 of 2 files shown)


cloverhearts:
added screenshot

tinkoff-dwh:
LGTM

Leemoonsoo:
LGTM

khalidhuseynov:
Thanks for looking into this, LGTM!

cloverhearts:
if no more opinions, i will merge.

PR #2280, opened 2017-04-24
ZEPPELIN-2445. Display which line of python code raise exception

zjffdu:
### What is this PR for?
When python code raise exception, user can only see the stacktrace, but don't know which line cause it. It is not so convenient especially when you write many lines of code. This PR would display which line raise exception. 

Besides that, I also fix another issue that we would have duplicated error message in python3. (See screenshot)

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2445

Unit test is added

Python 2 Before
image

Python2 After
image

Python 3 Before
image

Python3 After
image

### Questions:


spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -365,10 +345,20 @@ def getCompletion(self, text_value):
...
+
+        intp.setStatementsFinished("", False)
       except:
-        raise Exception(traceback.format_exc())
+        exception = traceback.format_exc()
+        m = re.search("File \"<stdin>\", line (\d+).*", exception)
+        if m:
+          line_no = int(m.group(1))
+          intp.setStatementsFinished(
+            "Fail to execute line {}: {}\n".format(line_no, stmts[line_no - 1]) + exception, True)
```
felixcheung: is `line_no` 1 base or 0 base?

felixcheung: why don't we just give the number number - seems like we are trying to include the line text as well?

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -365,10 +345,20 @@ def getCompletion(self, text_value):
...
+        else:
+          intp.setStatementsFinished(exception, True)
+      else:
+        intp.setStatementsFinished("", False)
 
-    intp.setStatementsFinished("", False)
   except Py4JJavaError:
```
felixcheung: does this still trap Py4jJavaError correctly? looks like this will never be hit?

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -257,6 +242,9 @@ def getCompletion(self, text_value):
 java_import(gateway.jvm, "org.apache.spark.mllib.api.python.*")
 
 intp = gateway.entry_point
+output = Logger()
```
zjffdu: move the creation of `output` after `intp`, because `Logger` depends on `intp`

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -328,22 +315,11 @@ def getCompletion(self, text_value):
...
-        continue
-
-      # skip comment
-      s_stripped = s.strip()
-      if len(s_stripped) == 0 or s_stripped.startswith("#"):
-        continue
-
-      final_code.append(s)
-
-    if final_code:
```
zjffdu: Keep the comment and empty line, so that we could get the correct line number.

spark/src/main/resources/python/zeppelin_pyspark.py
zjffdu: line_no is 1 base, you can check the test code where we get the correct line number.
Actually the line number in frontend is not correct, the first line may be the interpreter indicator.
See
image

spark/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,47 @@
...
+log4j.logger.org.quartz=WARN
+log4j.logger.DataNucleus=WARN
+log4j.logger.DataNucleus.MetaData=ERROR
+log4j.logger.DataNucleus.Datastore=ERROR
+
+# Log all JDBC parameters
+log4j.logger.org.hibernate.type=ALL
+
+log4j.logger.org.apache.zeppelin.interpreter=DEBUG
+log4j.logger.org.apache.zeppelin.spark=DEBUG
```
zjffdu: for debugging purpose of test code

zjffdu:
@felixcheung @Leemoonsoo Please help review.

spark/src/main/resources/python/zeppelin_pyspark.py
zjffdu: There's still java object (`intp`) in the while Loop, so I will keep it.

felixcheung: actually, the reason we have this is to extract out the relevant part of the exception text, as there is too much noise in the Py4J error.

from inspecting the code it looks like user code can be run with a inner try except, so I think we need the same for L368

zjffdu: I update the PR to throw Py4J error to outer try except.

zjffdu:
Thanks @felixcheung , will merge it if no more comments.

PR #2282, opened 2017-04-25
[ZEPPELIN-2447] Fix python interpreter as to use max result setting

paularmand:
### What is this PR for?
zeppelin_python.py used a hard-coded 1000 rows limit when returning results. This PR makes sure that the interpreter retrieves the max_results from PythonInterpreter.java and that the latter retrieves this configuration from the configuration via the properties object.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2447

+ use %python interpreter
+ make a dataframe with more than a 1000 rows
+ z.show the dataframe
+ export it in csv 
+ the numbers of rows should exceed 1000 rows

### Questions:
no
no
no

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -96,7 +96,7 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
 
   public PythonInterpreter(Properties property) {
     super(property);
-    this.maxResult = Integer.parseInt(property.getProperty(MAX_RESULT));
+    this.maxResult = Integer.parseInt(property.getProperty(MAX_RESULT, "1000"));
     try {
       File scriptFile = File.createTempFile("zeppelin_python-", ".py", new File("/tmp"));
       scriptPath = scriptFile.getAbsolutePath();
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -96,6 +96,7 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
 
   public PythonInterpreter(Properties property) {
     super(property);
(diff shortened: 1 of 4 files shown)


paularmand:
Ready for review

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -96,6 +96,7 @@
 
   public PythonInterpreter(Properties property) {
     super(property);
+    this.maxResult = Integer.parseInt(property.getProperty(MAX_RESULT));
```
zjffdu: Use `property.getProperty(MAX_RESULT, default_value)`

zjffdu:
ping @paularmand Can you fix the comment ?

paularmand:
@zjffdu Thanks for review. Added 1000 hard coded as default for now, although I believe a better solution would be to add fallbacks to a property class upfront (see https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#defaults).

zjffdu:
Thanks @paularmand LGTM

zjffdu:
@paularmand Could you add jira to the title of your PR ? name it as `[ZEPPELIN-2447] Fix python interpreter as to use max result setting`

paularmand:
Is the Jenkins failure an unrelated selenium test failure, or do I have to look into it?

paularmand:
ping @zjffdu Do I have to look into it the Selenium test failure?

zjffdu:
@paularmand Selenium test is flaky, one thing left is unit test. Could you add unit test in `PythonInterpreterTest` ？

paularmand:
@zjffdu I'll have a look. Can expect something later on this week.

jongyoul:
ping.

paularmand:
Pong @jongyoul . I wrote the unit test, but it is not passing on Jenkins. I have trouble setting up my local IDE to run the python tests in IntelliJ (all other tests, except the python ones, are running fine), so I'll have to dig deeper in the IDE settings. Pointers/help appreciated.

jongyoul:
Let me check

PR #2284, opened 2017-04-26
[ZEPPELIN-2451]: Add JDBC config option for calling connection.commit after paragraph execution

randerzander:
### What is this PR for?
Adding config for automatically calling commit after JDBC paragraph execution

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2451

Create property zeppelin.jdbc.autocommit with value "true" under the JDBC interpreter, run an INSERT statement against a database. The insert should be committed and should be accessible in the database.

Create property zeppelin.jdbc.autocommit with value "false" under the JDBC interpreter, run an INSERT statement against a database. The insert should not be committed and will not be in the database.

### Questions:
No
No
Yes

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -127,6 +127,8 @@ public class JDBCInterpreter extends Interpreter {
   private final String CONCURRENT_EXECUTION_COUNT = "zeppelin.jdbc.concurrent.max_connection";
   private final String DBCP_STRING = "jdbc:apache:commons:dbcp:";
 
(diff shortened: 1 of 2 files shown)


randerzander:
Looks like recent commits conflict with this fix. I'll re-synch with master, but @Leemoonsoo any comments on the approach before I do that?

Leemoonsoo:
Looks like this PR changes default behavior.

Previously, `connection.commit()` is executed if `connection.getAutoCommit()` is false.
But after this patch, `connection.commit()` will be executed if `connection.getAutoCommit()` is false and `zeppelin.jdbc.autocommit` property is true.

And `zeppelin.jdbc.autocommit` property is false by default.

Shell we keep default behavior?

randerzander:
Closed in favor of https://github.com/apache/zeppelin/pull/2396

PR #2285, opened 2017-04-26
ZEPPELIN-2450. Move spark profile to pom file of parent project

zjffdu:
### What is this PR for?
2 issues of spark profile:
* Duplicated profiles in spark and spark-dependencies
* The default profile can not be activated because it is in sub module. This cause spark module always use spark 2.0.2 in intellij even when I activate profile spark-2.1, this cause I fail to run `PySparkInterpreterTest` in intellij.

This PR move spark profiles to pom file of parent project. By default, profile spark-2.1 and scala-2.10 will be activated.

### What type of PR is it?
[ Improvement |  Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2450

CI pass

### Questions:


zjffdu:
@Leemoonsoo @felixcheung @jongyoul Please help review.

jongyoul:
@zjffdu About this issue, moon and I discussed when we split `spark-dependencies` from `spark`. We don't want to make duplication you mentioned, on the contrary, we don't want interpreter's profile to locate in a root `pom.xml` because that means all other interpreters profiles can be located in a root. I know what you intend and how good it is, but can you find another solution for it? I really don't want to see all profiles of modules in a root.

felixcheung:
+1 on what @jongyoul says

zjffdu:
> we don't want interpreter's profile to locate in a root pom.xml because that means all other interpreters profiles can be located in a root

Can you explain more ? Shouldn't all modules share the same profile ? e.g. We have a new module A which also want to use spark-2.1 profile. Then for the current pom structure he has to copy spark-2.1 profile to his pom file. If he want to use a different spark 2.1 profile compared to spark module for his module, then he has to give it a different profile name. Otherwise it would conflict with spark 2.1 profile in spark module.

felixcheung:
spark might be unique in this - almost all other profile are used only within one module

zjffdu:
I mean I don't see harm to move spark profile in root pom file, not to mention the benefit we could remove the duplication. I see spark project also put hadoop profile in root pom file although not all modules using hadoop profile.

felixcheung:
I'd argue possibly both approaches are useful: in some cases you might want all interpreters against the same version of spark, in some other cases you might want spark interpreter to spark 2.2 and livy interpreter to spark 1.6

zjffdu:
> in some other cases you might want spark interpreter to spark 2.2 and livy interpreter to spark 1.6

In that case, in both approaches we all have to create a new profile in livy module. And we can't name as spark-1.6, we need to give it a different name (like livy-spark-1.6).  I mean when user specify profiles in maven build command, all the profiles in root module and sub moduels are in the same scope and they should have unique name.

jongyoul:
@zjffdu All you mentioned are about advantage if we would have done, but I also think we have to consider negative effect for it. For instance, if someone want to move some of ignite profiles, which is less popular than spark, we don't have any reason not to allow it because spark already did. For the worst case, all of profiles can be located into root pom.xml. I totally agree that some popular - I don't know what we measure though - profiles locate in the root pom.xml is a bit helpful to extend new interpreter or other profiles but we should consider the negative effect.

PR #2294, opened 2017-04-27
[ZEPPELIN-2438]: Add local winutils.exe in build step and use if required …

cfries:
### What is this PR for?

Add local winutils.exe in build and use as temporary HADOOP_HOME if required on run (windows only).

The build of the spark dependencies now downloads winutils.exe and installs it in interpreter/spark/winutils.
The zeppelin-env.cmd (only used if we run on windows) then checks if HADOOP_HOME exists, if not, it will set it to point to winutils. This is required to have the spark interpreter work in Windows.

In order to have this active, the zeppelin-env.cmd had to be removed from the .gitignore.

### What type of PR is it?
[Bug Fix | Improvement]

Without this fix, the binary distribution or a local build of Zeppelin will have the spark interpreter (default interpreter) fail on a Windows platform (unless the user follows additional / manual installation steps or already has a local HADOOP). This will frustrate users trying to checkout Zeppelin.

For users having HADOOP or macOS/Linux users, the chance will have no effect.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2438

- Non-Regression Test
- Standard Unit Test
- Run Zeppelin without additional installations (except Java) on Windows

N/A

### Questions:


conf/zeppelin-env.cmd
```diff
@@ -76,3 +76,9 @@ REM
 REM set ZEPPELINHUB_API_ADDRESS	       REM Refers to the address of the ZeppelinHub service in use
 REM set ZEPPELINHUB_API_TOKEN          REM Refers to the Zeppelin instance token of the user
 REM set ZEPPELINHUB_USER_KEY           REM Optional, when using Zeppelin with authentication.
+
+REM Hadoop for Windows - This is a fix for spark/hadoop requiring winutils.exe -  see https://issues.apache.org/jira/browse/ZEPPELIN-2438
+REM
+if not exist "%HADOOP_HOME%" (
+    set HADOOP_HOME=%~dp0\..\interpreter\spark\winutils
+)
```
zjffdu: Could put this in `common.cmd`

felixcheung:
I think it's intentional that we ship a template of conf/zeppelin-env.cmd.template?

felixcheung:
Again, I'm not sure about releasing with a binary embedded in resource and how that correspond to the Apache release process.

But, I understand where you are coming up, and if you have a way to bootstrap the download only on the Windows platform, we could take a look at that?

conf/zeppelin-env.cmd
cfries: Good point. I will check.

cfries:
@felixcheung I don't know if it would be an issue embedding and .exe binary in the binary distribution. If that *is* the case, the either
a) default-dependency on hadoop should be removed
or
b) documentation should be added and - if possible - a dialog prompting the used for the download (s.th. "This Spark interpreter is pre-configures for Hadoop and requires winutils.exe. ....")

cfries:
(close and reopen to kick Jenkins to recognise Travis)

PR #2296, opened 2017-04-28
[ZEPPELIN-2463] Avoid Locking interpreterSettings during Notebook deletion

benoyantony:
### What is this PR for?
Deletion of a notebook requires locking interpreterSettings. If the deletion is delayed , then lock is not released. At that point, we cannot run any notebook because everything is waiting to lock interpreterSettings.
Looking at the code, there is no reason to lock the InterpreterSettings object in InterpreterSettingManager.removeNoteInterpreterSettingBinding.
Similarly in InterpreterSettingManager.getNoteInterpreterSettingBinding only interpreterSettingBinding is accessed and its already a thread safe object. So we can remove synchronization on InterpreterSettings

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2463

Being a concurrency issue, it is difficult to test. 
Please see the comments in the jira to see the issue experienced on a production zeppelin server.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -455,11 +455,9 @@ public class InterpreterSettingManager {
 
   private List<String> getNoteInterpreterSettingBinding(String noteId) {
     LinkedList<String> bindings = new LinkedList<>();
-    synchronized (interpreterSettings) {
-      List<String> settingIds = interpreterBindings.get(noteId);
-      if (settingIds != null) {
-        bindings.addAll(settingIds);
-      }
+    List<String> settingIds = interpreterBindings.get(noteId);
+    if (settingIds != null) {
+      bindings.addAll(settingIds);
     }
     return bindings;
   }
@@ -888,7 +886,6 @@ public class InterpreterSettingManager {
   }
 
   public void removeNoteInterpreterSettingBinding(String user, String noteId) throws IOException {
-    synchronized (interpreterSettings) {
       List<String> settingIds = (interpreterBindings.containsKey(noteId) ?
           interpreterBindings.remove(noteId) :
           Collections.<String>emptyList());
@@ -899,7 +896,6 @@ public class InterpreterSettingManager {
         }
       }
       saveToFile();
-    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


benoyantony:
@prabhjyotsingh , Could you please review this change?

benoyantony:
@Leemoonsoo , Could you please take a look at this change ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -888,18 +886,16 @@ public void removeRepository(String id) throws IOException {
   }
 
   public void removeNoteInterpreterSettingBinding(String user, String noteId) throws IOException {
-    synchronized (interpreterSettings) {
```
Leemoonsoo: This synchronize block is for protecting `interpreterBindings`, and i think we still need for synchronize block for `interpreterBindings`.

However, `this.removeInterpreterForNote()` doesn't have to be protected by synchronize block.
So i think code can be look like

```
List<String> settingIds;
synchronize (interpreterSettings) {
   settingIds = interpreterBindings.containsKey(...) ? interpreterBindings.remove(...) : ...
}

for (String settingId : settingIds) {
  ...
  this.removeInterpreterForNote(...);
  ...
}
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -455,11 +455,9 @@ public InterpreterSetting getDefaultInterpreterSetting(String noteId) {
 
   private List<String> getNoteInterpreterSettingBinding(String noteId) {
     LinkedList<String> bindings = new LinkedList<>();
-    synchronized (interpreterSettings) {
```
Leemoonsoo: I think this synchronized block shouldn't be removed to protect `interpreterBindings`

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
benoyantony: @Leemoonsoo , Thanks for so much reviewing. _interpreterBindings_ is a concurrentMap and hence need not require additional locking.  I believe , its true for both the changes.
The only problem that I see n the remove function is the "check and remove" .  I changed this to a "remove" operation.

Leemoonsoo: Ah that's good point. You're right. if interpreterBindings is concurrentMap, it looks like `getNoteInterpreterSettingBinding` and `removeNoteInterpreterSettingBinding` doesn't need to have synchronized block.

benoyantony: @Leemoonsoo , Thanks for the review. Is there anything that I need to do  to resolve this ?

Leemoonsoo:
@benoyantony Thanks for the fix!
LGTM and merge to master and branch-0.7 if no further discussions.

PR #2299, opened 2017-04-28
[ZEPPELIN-2468] Enable websocket without Origin if allowed.origins is *

elek:
Change-Id: Iaad10a69983036e84b766a22fbc32113b926b60d

### What is this PR for?
With ZEPPELIN-2288 we restored the check of the Origin field for websocket requests.

Unfortunately the current implementation will deny the request if the Origin HTTP header is empty, even if the zeppelin.server.allowed.origins is *

This patch enables websocket requests without Origin in the HTTP header if the zeppelin.server.allowed.origins=*. This fixes the work behind a restrictive reverse proxy (or behind Apache Knox)

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2468

It could be tested with curl as described in ZEPPELIN-2288, but I added additional unit test, so the change has been covered on unit test level.

N/A

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
@@ -60,11 +60,13 @@ public class SecurityUtils {
 
   public static Boolean isValidOrigin(String sourceHost, ZeppelinConfiguration conf)
       throws UnknownHostException, URISyntaxException {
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
LGTM

prabhjyotsingh:
Thank you @elek for the fix. Tested on CentOS works well. LGTM.

Leemoonsoo:

PR #2300, opened 2017-04-28
[ZEPPELIN-2467] Restoring second travis job functionality

AlexanderShoshin:
### What is this PR for?
Second travis job didn't work correctly. It suppose to test all core unit tests plus integration test (except for spark related test) but it was doing nothing.

It was because of the exclamation mark in **-Dtest** property:
```
-Dtest='!ZeppelinSparkClusterTest,!org.apache.zeppelin.spark.*'
```
which is not supported by maven-surefire-plugin of version 2.17 (which is used in Zeppelin). Exclamation mark is supported started from 2.19 but still does not work properly.

I've added plugin **exclude** configuratin instead of **-Dtest** property.

After travis job was restored I had found that not all core tests were working properly. I have excluded them from the travis job for now and created jira issues:
https://issues.apache.org/jira/browse/ZEPPELIN-2469
https://issues.apache.org/jira/browse/ZEPPELIN-2470
https://issues.apache.org/jira/browse/ZEPPELIN-2471
https://issues.apache.org/jira/browse/ZEPPELIN-2473

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
[ZEPPELIN-2467](https://issues.apache.org/jira/browse/ZEPPELIN-2467)

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
 
     # Test core modules
     - jdk: "oraclejdk7"
-      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtest='!ZeppelinSparkClusterTest,!org.apache.zeppelin.spark.*' -DfailIfNoTests=false"
+      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumBundleFactoryTest.java,**/HeliumApplicationFactoryTest.java,**/NotebookTest.java,**/ZeppelinRestApiTest.java -DfailIfNoTests=false"
 
     # Test selenium with spark module for 1.6.3
     - jdk: "oraclejdk7"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -47,7 +47,7 @@ matrix:
 
     # Test core modules
     - jdk: "oraclejdk7"
-      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtest='!ZeppelinSparkClusterTest,!org.apache.zeppelin.spark.*' -DfailIfNoTests=false"
+      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumBundleFactoryTest.java,**/HeliumApplicationFactoryTest.java,**/NotebookTest.java,**/ZeppelinRestApiTest.java -DfailIfNoTests=false"
```
felixcheung: excluding the exact filename is a bit hard to maintain, I think? is there a way to exclude by subfolder/path?

AlexanderShoshin: I agree that excluding tests by direct links is not the best way to solve the problem. But I can't think up anything more appropriate.

Excluding tests by packages will cause 15 more test files won't run because our 4 tests are located in 3 different packages in 2 different modules.

Maybe the better way will be to add an `@Ignore` annotation into test files instead of excluding them using travis configuration?

Leemoonsoo: Or how about add some verbose comments a line above to explain each file in `tests.to.exclude` flag?
The comments can mention JIRA issue. Then i think everyone can easily understand reason why the file is excluded and helps maintaining the file.

AlexanderShoshin: I've added a comment about excluded tests

AlexanderShoshin:
According to @Leemoonsoo suggestion I have added a comment for the travis job configuration about excluded tests.

Leemoonsoo:
Travis build green. https://travis-ci.org/AlexanderShoshin/zeppelin/builds/228261473

1ambda:
even though CI is green in this PR, it seems that wasn't rebased recently. As a result CI is failing only in the second travis JOB. @AlexanderShoshin Could you check that?

jongyoul:
@AlexanderShoshin After merging this PR, we continued to got failed in CI. Can you handle it urgently? If it took some times, it would be better to revert this commit and merge it after fixing it. How do you think of it?

AlexanderShoshin:
I think we should revert this commit then. I won't be able to fix it fast.

jongyoul:
Thanks for your opinion. Then I'll revert it. Please reopen this PR and rebase it from current master.

PR #2301, opened 2017-04-28
[ZEPPELIN-1625] Override Interpreter settings at User level

benoyantony:
### What is this PR for?
When multiple users share a zeppelin instance, it is sometimes desirable to use the same interpreter with different settings. 

It should be possible an administrator to setup some of the Interpreter settings as overridable by setting a place holder and a default value
It should be possible for a user to specify the actual values via properties.

### What type of PR is it?
Feature

### Todos
Override interpreter setting at notebook level

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1625

Unit tests are added. 
Please see screen shots for manual testing

### Screenshots

override-interpreter-setting

set-property

### Questions:
* Does this needs documentation? Yes , added


benoyantony:
@Leemoonsoo , Could you please review this feature ?

zjffdu:
@benoyantony Did you check ZEPPELIN-1338 ? Your approach seems some kind of workaround to me.

benoyantony:
@zjffdu , Thanks for pointing me to ZEPPELIN-1338. I did not know about it. I believe, the intent is same. I agree that the approach is different. We wanted to shield the user from interpreter settings.  

The approach in this change is to specify a property name and default value in the interpreter setting.  Users  can set a value  for the property. When the interpreter is created, then the user's property value is used as the interpreter setting. 
User's properties are persisted. 
One advantage of this approach is to use the same property across multiple interpreter settings.
We have been using this feature at our company for last 3 months and its been effective. 

I'll review ZEPPELIN-1338 carefully to understand the approach mentioned in that. Thanks again for letting me know of the related effort. Its good to know that its a feature needed by many.

esmevane:
Hello there, @benoyantony - could you update me on where you and your team wound up with this?  I'm working on something which sounds pretty similar and I'd love to know what you all decided.

PR #2310, opened 2017-05-02
[ZEPPELIN-2484] Fix NullPointerException in check for empty last paragraph

benoyantony:
### What is this PR for?
Prevent NullPointerException during check to determine whether a new paragraph needs to added.
The fix is to switch order of null check and trim operation so that null check is performed before trim()

### What type of PR is it?
Bug Fix 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2484

This can be tested with a hive interpreter 
Create a note and add a paragraph with some query in it. 
Run all paragraphs.
A new paragraph is automatically added.
Run all paragraphs again. A NullPointerException is logged in the logs.

### Questions:


Leemoonsoo:
LGTM and merge to master if no further comments.

PR #2312, opened 2017-05-03
ZEPPELIN-2486. AngularElem's onChange is only invoked once

zjffdu:
### What is this PR for?
`ng-change` can not capture event, so that means `$event.timeStamp` is undefined. https://github.com/angular/angular.js/issues/6370
This cause AngularElem's onChange is only invoked once. This PR use another approach to update the angularObject. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2486

Tested manually

Before
zeppelin_before
After
zeppelin_after
### Questions:

### DIFF
--- a/zeppelin-display/src/main/scala/org/apache/zeppelin/display/angular/AbstractAngularElem.scala
+++ b/zeppelin-display/src/main/scala/org/apache/zeppelin/display/angular/AbstractAngularElem.scala
@@ -131,10 +131,10 @@ abstract class AbstractAngularElem(val interpreterContext: InterpreterContext,
     // create AngularFunction in current paragraph
     val functionName = eventName.replaceAll("-", "_") + "_" + uniqueId
     val elem = this % Attribute(None, eventName,
-      Text(s"${functionName}=$$event.timeStamp"),
+      Text(s"${functionName}=${functionName} + 1"),
       Null)
 
-    val angularObject = addAngularObject(functionName, "")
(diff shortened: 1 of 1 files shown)


zjffdu:
@Leemoonsoo Please help review.

zeppelin-display/src/main/scala/org/apache/zeppelin/display/angular/AbstractAngularElem.scala
```diff
@@ -131,8 +131,8 @@ abstract class AbstractAngularElem(val interpreterContext: InterpreterContext,
     // create AngularFunction in current paragraph
     val functionName = eventName.replaceAll("-", "_") + "_" + uniqueId
     val elem = this % Attribute(None, eventName,
-      Text(s"${functionName}=$$event.timeStamp"),
-      Null)
+      Text(s"${functionName}=${functionName} + 1"),
+      Null) % Attribute(None, "ng-init", Text(s"${functionName} = 0"), Null)
```
Leemoonsoo: Looks like initializing value using ng-init calls `callback` function once, before the actual event.
I think the `${functionName}` need to be initialized in the following line

```
val angularObject = addAngularObject(functionName, 1)
```

Leemoonsoo:
LGTM

zjffdu:
Thanks, will merge if no more comments.

PR #2317, opened 2017-05-04
[ZEPPELIN-2493] Visualization class should throw an error if an abstract method is not implemented

tae-jun:
### What is this PR for?
`Visualization` class should throw an error if an abstract method is not implemented just like [`SpellBase`](https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/spell/spell-base.js#L25) does.

And I added some comments. In my opinion, though `Visualization` has [a good example](http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/development/writingzeppelinvisualization.html), it needs an API documentation page (`Spell` as well). `Visualization` class has methods which can be overridden (e.g. `refresh`, `destroy`, etc.), but the example doesn't mention them. Therefore, users don't know what and how to implement easily!

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-2493](https://issues.apache.org/jira/browse/ZEPPELIN-2493)

* Inherit `Visualization`
* Don't implement any abstract method
* Run `getTransformation()` and `render()`
* Check they print appropriate errors

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/visualization.js
+++ b/zeppelin-web/src/app/visualization/visualization.js
@@ -13,7 +13,7 @@
  */
 
 /**
- * Base class for visualization
+ * Base class for visualization.
  */
 export default class Visualization {
   constructor (targetEl, config) {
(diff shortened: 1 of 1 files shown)


1ambda:
LGTM!

1ambda:
Mostly about documentation. Not necessary to write a new test.

LGTM.

tae-jun:
@1ambda Thanks for the review! 😄 

I didn't know that you asked me to add a test 😢
Was it added after the editing?

But, since you said it is not necessary, I will just leave this PR 😃 
Thanks!

Leemoonsoo:

PR #2320, opened 2017-05-04
[MINOR] Move remoteinterpreter into zengine

jongyoul:
### What is this PR for?
RemoteInterpreter is only used in the server side then zeppelin-interpreter doesn't have to include this class. Moving this class helps to reduce interpreter binary size and change RemoteInterpreter without adding more dependencies if we want

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Move RemoteInterpreter and related files out of zeppelin-interpreter module

### What is the Jira issue?
N/A

N/A

### Questions:


zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/InterpreterTest.java
```diff
@@ -31,7 +30,37 @@
   public void testDefaultProperty() {
     Properties p = new Properties();
     p.put("p1", "v1");
-    MockInterpreterA intp = new MockInterpreterA(p);
+    Interpreter intp = new Interpreter(p) {
```
felixcheung: instead of duplicating this would it be less error prone to have a common interpreter test java file to define this?

jongyoul: yes, I agree. I'll change it

jongyoul:
Will merge it sooner

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/DummyInterpreter.java
```diff
@@ -0,0 +1,43 @@
+package org.apache.zeppelin.interpreter;
+
+import java.util.Properties;
+
+/**
+ * Created by jl on 06/05/2017.
```
Leemoonsoo: Author tag is not too much encouraged.

jongyoul: That's my mistake. I'll remove it.

PR #2322, opened 2017-05-04
[ZEPPELIN-2502] RemoteInterpreterServer hang forever during shutdown

andreaTP:
### What is this PR for?
There is the chance to have a RemoteServerInterpreter hang forever during shutdown

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
[ZEPPELIN-2502] 

Unit test provided for the fix.

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
@@ -477,15 +477,18 @@ public class RemoteInterpreterEventClient implements ResourcePoolConnector {
   /**
    * Wait for eventQueue becomes empty
    */
-  public void waitForEventQueueBecomesEmpty() {
+  public void waitForEventQueueBecomesEmpty(long atMost) {
+    long startTime = System.currentTimeMillis();
     synchronized (eventQueue) {
-      while (!eventQueue.isEmpty()) {
+      while (!eventQueue.isEmpty() && (System.currentTimeMillis() - startTime) < atMost) {
         try {
           eventQueue.wait(100);
         } catch (InterruptedException e) {
           // ignore exception
         }
       }
+      if (!eventQueue.isEmpty())
+        eventQueue.clear();
     }
   }
 }
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
@@ -80,6 +80,8 @@ public class RemoteInterpreterServer
   private Map<String, Object> remoteWorksResponsePool;
   private ZeppelinRemoteWorksController remoteWorksController;
 
+  private final long DEFAULT_SHUTDOWN_TIMEOUT = 2000;
+
   public RemoteInterpreterServer(int port) throws TTransportException {
     this.port = port;
 
@@ -99,7 +101,7 @@ public class RemoteInterpreterServer
 
   @Override
   public void shutdown() throws TException {
-    eventClient.waitForEventQueueBecomesEmpty();
+    eventClient.waitForEventQueueBecomesEmpty(DEFAULT_SHUTDOWN_TIMEOUT);
     if (interpreterGroup != null) {
       interpreterGroup.close();
     }
@@ -111,7 +113,8 @@ public class RemoteInterpreterServer
     // this case, need to force kill the process
 
     long startTime = System.currentTimeMillis();
-    while (System.currentTimeMillis() - startTime < 2000 && server.isServing()) {
+    while (System.currentTimeMillis() - startTime < DEFAULT_SHUTDOWN_TIMEOUT &&
+        server.isServing()) {
       try {
         Thread.sleep(300);
       } catch (InterruptedException e) {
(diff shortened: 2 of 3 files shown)


andreaTP:
@felixcheung good catch, after a second look I can see that while it is waiting on `pollEvent` it will also delay the `sendEvent` since it is `synchronized` too.

I guess all these problems are already solved by Java standard lib, if I have positive feedback I will go substituting `eventQueue` with a [LinkedBlockingQueue](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/LinkedBlockingQueue.html) and removing all `synchronized` statements since they become unuseful.

felixcheung:
I think you are getting a 👍

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
```diff
@@ -477,15 +477,18 @@ public void onParaInfosReceived(Map<String, String> infos) {
   /**
    * Wait for eventQueue becomes empty
    */
-  public void waitForEventQueueBecomesEmpty() {
+  public void waitForEventQueueBecomesEmpty(long atMost) {
```
FireArrow: I think it would be nice if `atMost` indicated what time unit it wants. Rename it to `atMostMilliseconds`?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -111,7 +113,8 @@ public void shutdown() throws TException {
     // this case, need to force kill the process
 
     long startTime = System.currentTimeMillis();
-    while (System.currentTimeMillis() - startTime < 2000 && server.isServing()) {
+    while (System.currentTimeMillis() - startTime < DEFAULT_SHUTDOWN_TIMEOUT &&
```
FireArrow: While not really part of the scope of this PR, it is a chance to rework the logic to remove some unnecessary arithmetic and improve readability a bit.
Change `long startTime = System.currentTimeMillis();` to `long deadline = System.currentTimeMillis() + DEFAULT_SHUTDOWN_TIMEOUT;`
Change `while (System.currentTimeMillis() < deadline) &&`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
andreaTP: ok, and I harmonize this while with your comments below 
( `deadline` etc.)

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
andreaTP: ok

andreaTP:
Looking at the code it looks quite harmful also the pattern used [here](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java#L138), [here](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java#L191) and [here](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java#L245)

if anything goes wrong the Thread can stay in `wait` status potentially forever and operations performed there are all but not transactional.

Could I ask if somebody can at least provide some kind of deadline also for those?
Better will be a little refactoring using something like [ConcurrentHashMap](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html) for operations.

I can even do the refactoring if I have positive feedbacks, anyhow I believe this PR is ok as is and I open a new issue and new PR in case.

FireArrow:
I'm all for refactoring the use of synchronized into using more proper data types instead. It's to easy to get stuck in a [live|dead]lock when doing the synchronized yourself

andreaTP:
I have had to fix some escaping problems to make travis happy, and it also required some retries ....
anyhow travis is now happy: https://travis-ci.org/nokia/zeppelin/builds/230331180

I hope this PR is addressed and we can move on.
I still wait hint from @felixcheung and @Leemoonsoo on how to proceed on the rest.

thanks in advance

Leemoonsoo:
LGTM

> Looking at the code it looks quite harmful also the pattern used here, here and here

Further refactoring on codes pointed out sounds good as well.

andreaTP:
am I supposed to do anything else?

Leemoonsoo:
@andreaTP Changes on file ZeppelinIT.java are required? And It'll be great if this branch is rebased (or merge) to master and see CI becomes green.

andreaTP:
@Leemoonsoo I can reproduce the error locally running on Chrome (i.e. the current encoding of the test doesn't work for me out of the box with chrome), should I separate these modifications in a separate PR?

Leemoonsoo:
@andreaTP Yes. separate PR would be more helpful.

andreaTP:
@Leemoonsoo sounds good :-) I'll do that tomorrow!
and I will follow up with this PR. Thanks!

Leemoonsoo:
Thanks @andreaTP !

andreaTP:
I reverted back modifications to concurrency management since it looks like they result in Ui response instability, I'm not sure why this happens, I believe that we can conclude with a clean PR like this.
In case I manage to refactor and use proper `java.util.concurrent` into `remote` I will send another PR.
WDYT? @Leemoonsoo @felixcheung

andreaTP:
ping

Leemoonsoo:
Merge to master and branch-0.7 if no further comment

PR #2323, opened 2017-05-08
[ZEPPELIN-2411] Improve Table

1ambda:
### What is this PR for?

**Improve Table**

- column filter
- persist column state: type, order, hide/show, sorting, pinning
- pagination
- `setting` menu to configure table UI
- group by + aggregation

And **all these things are persisted and synchronized among web socket clients**

See the screenshot section for more detail.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Remove handsontable dependencies
* [x] - Use npm packaged moment* packages.
* [x] - Apply ui-grid
* [x] - Add setting menu
* [x] - Fix some issues
* [x] - Persist column type

### What is the Jira issue?

[ZEPPELIN-2411](https://issues.apache.org/jira/browse/ZEPPELIN-2411)

1. Build: `mvn clean package -DskipTests; ./bin/zeppelin-daemon.sh restart`
2. Open a note and create tables. If you don't have proper paragraphs, use this snippet.

```scala
%spark

import org.apache.commons.io.IOUtils
import java.net.URL
import java.nio.charset.Charset

val bankText = sc.parallelize(
    IOUtils.toString(
        new URL("https://s3.amazonaws.com/apache-zeppelin/tutorial/bank/bank.csv"),
        Charset.forName("utf8")).split("\n"))

case class Bank(
    age: Integer, 
    job: String, 
    marital: String, 
    education: String, 
    balance: Integer,
    housing: Boolean,
    loan: Boolean,
    contact: String,
    day: Int,
    month: String,
    duration: Int,
    y: Boolean
)

val bank = bankText.map(s => s.split(";")).filter(s => s(0) != "\"age\"").map(
    s => Bank(s(0).toInt, 
            s(1).replaceAll("\"", ""),
            s(2).replaceAll("\"", ""),
            s(3).replaceAll("\"", ""),
            s(5).replaceAll("\"", "").toInt,
            if (s(6).replaceAll("\"", "") == "yes") true else false,
            if (s(7).replaceAll("\"", "") == "yes") true else false,
            s(8).replaceAll("\"", ""),
            s(9).replaceAll("\"", "").toInt,
            s(10).replaceAll("\"", ""),
            s(11).replaceAll("\"", "").toInt,
            if (s(16).replaceAll("\"", "") == "yes") true else false
        )
).toDF()
bank.registerTempTable("bank")
```

```sql
select age, education, job, balance from bank limit 1000
```

#### Before

image

#### After: Filter

2411_new_filter

#### After: Column related features

2411_new_column

#### After: Pagination

2411_new_pagination

#### After: Group, Aggregation

2411_new_group_aggr

#### After: Synchronized

2411_new_sync

### Questions:
* Does the licenses files need update? - YES, updated


Leemoonsoo:
Tested and overall it's great improvement!

Here's few feedbacks.

1. Table does not refreshed after height / width of paragraph change.
2. How about remove 'type' in column table option?
image
I guess user will be okay without knowing it.
3. showColumnFilter
Following is screenshot with 'showColumnFilter' on. I don't know what this option supposed to do except for adding empty sticky row in the bottom.
image

4. pagenation
`defaultPaginationSize ` size can be only changed after the Table is displayed with `defaultPaginationSize ` applied. So changing default does not mean too much in this case. Isn't it?

`availablePaginationSizes` - It's minor opinion, I think it is okay with hardcoded value and not make user configurable.

5. Header bar in the option
image

Two buttons - 'Save', 'Delete' icon - on right little bit confuses me. It was difficult to guess what those button supposed to do. And figured out

'save' - apply option to the table.
'delete' - restore to default options

While option change applies immediately, how about hide 'Save' button? And change 'delete' icon to something more appropriate? For example "Restore the default values" text would be more intuitive.

1ambda:
Here's few feedbacks.

1. Table does not refreshed after height / width of paragraph change.

TBD

2. How about remove 'type' in column table option?

- Removed
- While adding tooltip for each option

image

3. showColumnFilter

It's used to display aggregated values. I changed the name to `showAggregationFooter` which easier to understand than before.

image

4. pagination

- Removed pagination related settings.

image

5. Header bar in the option

- Removed the `save` button.
- Updated icon for reset and added tooltip for it.

image

guptarajat:
"all these things are persisted and synchronized among web socket clients"
What does this mean?

1ambda:
@guptarajat Hi, refer the last GIF. "persist" means that all state of table is saved in the note.josn. For example, sorting state will be saved after this PR but it is not in current master.

Leemoonsoo:
Each cell in table supposed to print html. For example

```
%sh echo -e "%table key\tvalue\nsun\t%html <img width=50px src=http://www.freeiconspng.com/uploads/sun-icon-4.jpg></img>"
```

This PR
image

In master branch
image

Leemoonsoo:
And this PR adds background in rows. 

Master branch
image

This PR
image

What do you think? Keep appearance the same as before, so we can keep user experience change minimum. Or is adding background color in rows brings more benefit?

1ambda:
@Leemoonsoo 

Colorizing (even rows) is cognitively better because we are not confused when investigating closer rows.

Here are more bright versions

2411_alter

2411_white

Leemoonsoo:
Bright version looks better for me

1ambda:
@Leemoonsoo 

Oops. I missed that feature.

1. Now table renders HTML display in the cells

image
 
2. Fixed to have more bright even rows.

image

1ambda:
@Leemoonsoo just rebased and

updated to use `white` color in each cell background in c56edca
now it looks like 

image

Leemoonsoo:
GTM. Thanks @1ambda for great improvement.

zeppelin-web/src/app/visualization/builtins/visualization-table.js
```diff
@@ -14,50 +14,391 @@
...
+    let gridElem = document.getElementById(gridElemId)
+
+    const config = this.config
+    const self = this // for closure
+
+    if (!gridElem) {
+      // create, compile and append grid elem
+      gridElem = angular.element(
+        `<div id="${gridElemId}" ui-grid="${gridElemId}"
+              ui-grid-edit ui-grid-row-edit
```
vglushak: This made all values in the table editable. Is that expected?

vglushak:
> Now table renders HTML display in the cells and adjusts row height based on the content.

There are some limitations for this feature. img, links works. But when I want to place button it just ignore html and display text only.

Tagar:
Great improvement. 
Submitted https://issues.apache.org/jira/browse/ZEPPELIN-3251 to consider showing rows/columns lazily on scrolling events as otherwise ui-grid has a significant overhead for wide datasets so it randers much slower too. Other ideas?

PR #2329, opened 2017-05-10
[WIP][PoC] ZEPPELIN-2040 ClusterManager to support launching interpreter in a cluster

jongyoul:
### What is this PR for?
Launching interpreters into yarn cluster. This is PoC level now and has many more steps described below. The main classes are `Client`, `ClusterManager`, `RemoteInterpreterYarnProcess`

### What type of PR is it?

### Todos
* [ ] - Divide yarn dependencies with other module
* [ ] - Support yarn-cluster without setting SPARK_HOME
* [ ] - Remove unused files
* [ ] - TBD

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELI-2040

1. Install hadoop in your local https://dtflaneur.wordpress.com/2015/10/02/installing-hadoop-on-mac-osx-el-capitan/
1. add `zeppelin.cluster_manager=yarn` in your spark interpreter setting
1. test spark script

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/cluster/yarn/ApplicationMaster.java
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    @Override
+    public void run() {
+      logger.info("Setting up container launch container for containerid={}", container.getId());
+
+      Map<String, LocalResource> localResources = new HashMap<>();
+
+      Vector<CharSequence> vargs = new Vector<>(30);
+      vargs.add("echo \"madeng\"");
```
zjffdu: I believe this line could be removed

zjffdu:
@jongyoul Thanks for the POC. One question about the cluster mode for spark interpreter. It looks like you are creating a general yarn app support for all the interpreters. But for spark, it has already supported yarn-cluster, we could deploy spark interpreter as yarn-cluster mode via spark-submit. Although your approach may work, but I am afraid it may lose some features compared to the native yarn support of spark and add extra overhead of maintenance of feature parity between native yarn-cluster of spark and zeppelin yarn-cluster.

zeppelin-zengine/src/main/java/org/apache/zeppelin/cluster/yarn/ApplicationMaster.java
jongyoul: Yes, I'll remove that file fully.

jongyoul:
@zjffdu Thanks for quick review. I know spark already have yarn-cluster by itself. In case of Zeppelin, SparkInterpreter will run yarn-client mode which means driver is still in a process zeppelin launches. at the same time, SparkInterpreter will run in a yarn cluster as one of application masters and will launch Spark's application master as a new application master in yarn cluster. It's fully same as spark-summit does except one thing to spark driver is also in yarn cluster. If you test it in a cluster, you will see like this:

pasted image at 2017_05_04 01_07

It's launched as yarn-client mode. The application type of `ZEPPELIN INTERPRETER` is a driver and one of `SPARK` is a application master spark launches. 

BTW, I searched how livy has been supporting yarn-cluster mode and it's adoptable for Zeppelin, too. I'll add that feature sooner.

zjffdu:
Do you mean there would be 2 yarn apps for launching one spark interpreter ? That looks a little weird to me. In this approach, we still launch remote interpreter process in the zeppelin host, that doesn't solve much memory issue of zeppelin host. And besides it would waste yarn resources as it require to launch 2 yarn apps. Why not leverage spark-submit to support yarn-cluster mode ? And I look at all the interpreters of zeppelin. Most of them are just client role. That means the computation happens in the backend not in the interpreter side. so I think it would be fine to launch these interpreters in shared/scoped mode for all users. Spark Interpreter would be a special case, we could use its native yarn-cluster support. That would be the simplest way I think.

jongyoul:
In case of Spark, it has two yarn apps. I agree that it's a bit weird for users. Otherwise, all other interpreters like python get better to launch it in yarn cluster. I think it would be better to support yarn-cluster mode in case of Spark.

jongyoul:
It makes our logic much difficult, but it's improve users' experiences.

zjffdu:
> it's improve users' experiences
Sorry, I don't understand how it improve users' experience compared the native spark yarn-cluster support. And does this approach mean all the interpreter must use the same ClusterManager ? If so, this also doesn't make sense to me. Because we may want to run md in local mode and spark in yarn-cluster mode.

jongyoul:
I mean 'not in case of spark'. And this implementation satisfies your needs. You can launches some interpreters like md in local mode and some interpreters in yarn.

jongyoul:
And what I meant ux is about supporting spark's native yarn-cluster mode in Zeppelin :-)

zeppelin-zengine/src/main/java/org/apache/zeppelin/cluster/yarn/SimpleYarnApp.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.hadoop.yarn.client.api.YarnClientApplication;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import org.apache.hadoop.yarn.util.ConverterUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ */
+public class SimpleYarnApp {
```
zjffdu: This class is never used, just for testing ?

jongyoul: Sorry for confusion. Yes, will be removed. `Client`, `ClusterManager`, `RemoteInterpreterYarnProcess` are used classes. Rest of them will be removed

jongyoul: I'm editing some code. This is just for testing it in local yarn cluster.

zjffdu:
@jongyoul I could not run your PR, seems they only work in your enviroment. There are several issue, one of them is  
* HADOOP_CONF_DIR is not in zeppelin server classpath, so it could not connect with my yarn cluster. It works in your enviroment because you use default yarn configuration I believe.
 
Besides that I am still have concern about this approach:
1. You launch Remote Interpreter Process as yarn AM, and then create Spark Interpreter through thrift in yarn-cluster mode. Seems you don't use spark-submit, that means SparkSubmit.scala is never called, and this would cause potential problems.
2. The other thing I worry about is that now you launch spark app in a remote machine (Yarn AM). But that remote machine may not have the spark configuration (like spark-defualts.conf, hive-site.xml). And even it seems not possible to run multiple versions of spark in this approach.

jongyoul:
@zjffdu I'm adding to read HADOOP_CONF_DIR. Concerning spark, I recommend, for now, to set local[*], and yarn-client. In case of yarn-cluster, I think native approach would be better. I'll push my new changes soon.

zjffdu:
@jongyoul Thanks for quick response. I am thinking that we could do the yarn cluster mode support in 2 steps:
1. Support yarn-cluster mode for spark interpreter via native spark-submit. 
2. Support yarn-cluster mode for other interpreters via a general framework.

Since spark interpreter is the most important interpreter of zeppelin. We can do step 1 first, what do you think ?

zjffdu:
@jongyoul If you consider to use the native yarn cluster support of spark, then you may need to check ZEPPELIN-1263 (#1446), as for yarn cluster mode, we have to set configuration before launching the spark app otherwise some configuration would not take effect.

Tagar:
Great job. Exciting to see this new feature in Zeppelin. 
My two cents:

> through thrift in yarn-cluster mode

Cloudera's Spark don't have thrift service and Cloudera doesn't recommend to use that (they quote security is one of the concerns there).
So I also think spark-submit would be a better option (either it's on a remote or local machine, yarn-cluster or yarn-client). 

> But that remote machine may not have the spark configuration (like spark-defualts.conf, hive-site.xml). 

Not sure how big this of a concern. We normally have all hadoop servers and gateway servers identical all hadoop-related configuration files (it's managed by Cloudera Manager for us).

jongyoul:
@zjffdu Personally, I agree on your opinion. Before I start to do it, I discussed a lot how to support yarn-cluster mode including spark. I concluded it's not much weird to launch spark with yarn-client mode in yarn-cluster. Actually, I expected this kind of reaction while I'm doing this job. Then I'll finish it asap and do that with next step.

jongyoul:
@Tagar Thanks for reaching it out. In case of Spark, I feel like we need to focus on external spark mode more. In case of `spark-defaults.conf`, AFAIK, because we don't need to deploy spark in all cluster to use it.

zjffdu:
> Cloudera's Spark don't have thrift service and Cloudera doesn't recommend to use that (they quote security is one of the concerns there).

@Tagar Sorry for making you confused. The thrift service is not spark thrift server, it is the thrift protocol between zeppelin server and zeppelin interpreter process.

Tagar:
@zjffdu @jongyoul got it - thanks for prompt response.

zjffdu:
@jongyoul What is your current design and plan ? Since this is a big change, it might be better to reach consensus on the design, otherwise it would waste your time if we found issue after you complete the implementation. Overall, I have 2 concerns and suggestions:
1. use spark-submit to launch spark's InterpreterProcess instead of using yarn api to do that. Because using yarn api means to reinvent the wheel of spark's yarn module, and it would take lots of time to keep the behavior consistent with spark yarn module as there's many tricky things and configurations
2. How to specify the port when launching the InterpreterProcess in yarn cluster mode. I mentioned it in ZEPPELIN-2035, as I think we need to finish ZEPPELIN-2035 first.

jongyoul:
@zjffdu In case of Spark, you're so true. But as I mentioned it before, it's for all interpreters including python and so on. I think if we finish ZEPPELIN-2035, we improve spark's way of running interpreter. It will be a definitely better idea to use Spark's native way.

jongyoul:
And it's big change of running interpreters but it rarely affects current code. I divided all cluster manager code into another module to reduce dependencies.

zjffdu:
@jongyoul I understand this is for all interpreters. What I concern is that this approach would not work for spark. e.g. many of spark configuration would not work (e.g. spark.files, spark.jars, keytab, principal and etc).  And since spark interpreter is the most important interpreter of zeppelin, so I would suggest to implement cluster mode for spark first via spark-submit.

jongyoul:
@zjffdu I agree with you. It's very basic and minimum feature to run interpreters in Zeppelin. I'll take that issue ASAP.

jongyoul:
Ready to review.

- Introduced `zeppelin-cluster/common` for cluster manager and its factory class
- Added hadoop related setting in `dependecyManagement` of root `pom.xml`

jongyoul:
CI already has been passed: https://travis-ci.org/jongyoul/zeppelin/builds/249265535

zjffdu:
@jongyoul Could you add more javadoc on these new classes and write a simple summary about the design so that it is easier for other people to review it ?

And I would suggest you to add unit test for yarn-cluster mode, you can use MiniYarnCluster for that. Otherwise it is very easy to introduce regression. 

https://wiki.apache.org/hadoop/HowToDevelopUnitTests

jongyoul:
@zjffdu Thanks for the quick reply. I'll add javadoc for describing it simply. but concerning the unit test, this feature need integration test so I'm not sure if unit test is useful or not. Do you have any suggestion on it?

jongyoul:
Added simple descriptions :-)

zjffdu:
Not sure what integration test you mean ? Do you mean selenium test ? Actually `MiniYarnCluster` will launch a mini yarn cluster locally, and you can submit jobs to this mini cluster. You can refer this.

https://github.com/apache/spark/blob/master/resource-managers/yarn/src/test/scala/org/apache/spark/deploy/yarn/BaseYarnClusterSuite.scala

jongyoul:
What I meant is a kind of selenium test. And I saw it before and it will be useful for testing yarn cluster manager itself. I'll try to add it.

pom.xml
```diff
@@ -263,6 +267,33 @@
         <version>${shiro.version}</version>
       </dependency>
 
+      <!-- Apache Hadoop -->
+      <dependency>
```
zjffdu: You can use `hadoop-client` to replace all of these

python/pom.xml
```diff
@@ -125,6 +125,8 @@
               <target>
                 <unzip src="${project.build.directory}/../../interpreter/python/py4j-${python.py4j.version}.zip"
                        dest="${project.build.directory}/../../interpreter/python"/>
+                <zip destfile="${project.build.directory}/../../interpreter/python/py4j-${python.py4j.version}-src.zip"
```
zjffdu: What is this change ?

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -175,7 +175,7 @@ private Map setupPySparkEnv() throws IOException{
     if (!env.containsKey("PYTHONPATH")) {
       SparkConf conf = getSparkConf();
       env.put("PYTHONPATH", conf.get("spark.submit.pyFiles").replaceAll(",", ":") +
-              ":../interpreter/lib/python");
+              ":../interpreter/lib/python:python");
```
zjffdu: why this change ?

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManager.java
```diff
@@ -0,0 +1,50 @@
...
+  public abstract void start();
+
+  /**
+   * Can throw `RuntimeException`
+   */
+  public abstract void stop();
+
+  public abstract RemoteInterpreterProcess createInterpreter(String id, String name,
+      String groupName, Map<String, String> env, Properties properties, int connectTimeout,
+      RemoteInterpreterProcessListener listener, ApplicationEventListener appListener,
```
zjffdu: Why pass `RemoteInterpreterProcessListener` and `RemoteInterpreterProcessListener ` ? And I don't think they can be passed because they are not serializable.

zeppelin-cluster/yarn/pom.xml
```diff
@@ -0,0 +1,232 @@
...
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-yarn-api</artifactId>
```
zjffdu: use hadoop-client

zeppelin-cluster/yarn/src/main/java/org/apache/zeppelin/cluster/yarn/RemoteInterpreterYarnProcess.java
```diff
@@ -0,0 +1,512 @@
...
+      }
+
+      Map<String, LocalResource> localResources = new HashMap<>();
+
+      FileSystem fileSystem = FileSystem.get(configuration);
+
+      Path interpreterDir = getInterpreterRelativePath();
+      List<Path> interpreterPaths = getPathsFromDirPath(interpreterDir);
+
+      if (isSparkInterpreter()) {
```
zjffdu: @jongyoul I think we would use spark-submit to launch spark interpreter for its yarn-cluster mode. And this PR is for non-spark interpreter,  Do I miss anything ?

pom.xml
jongyoul: Sounds good. I'll try it

python/pom.xml
jongyoul: It was existing bug in python Interpreter. That codes didn't works as expected. So I changed it to normal

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
jongyoul: I have changed the location of Python libraries when launching python process to simplify the structure of libraries used in yarn process.

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManager.java
jongyoul: This is a legacy things to support original behaviours of `RemoteInterpreterProcess`. It doesn't have to be serialized because this object runs under server side.

zeppelin-cluster/yarn/src/main/java/org/apache/zeppelin/cluster/yarn/RemoteInterpreterYarnProcess.java
jongyoul: @zjffdu Yes, we will support yarn cluster mode using spark-submit. But I thought we need to support running spark like other interpreters. After we finish to implement yarn-cluster mode feature for Spark Interpreter, it will be useless and then we will remove this kind of tricky code, too.

jongyoul:
@zjffdu Thanks, Jeff for this kind review. I've addressed your comments and will handle some suggestion, too.

zeppelin-cluster/yarn/src/main/java/org/apache/zeppelin/cluster/yarn/RemoteInterpreterYarnProcess.java
zjffdu: My concern is that this approach for spark interpreter will have lots of problems, many features of spark won't work here. And if we expose this feature to users, I believe it would bring bad experience to users. Because users don't care about the underlying implementation, they would think that any features works in spark itself should also work in zeppelin.

jongyoul: I understand it. So I plan to develop that feature before releasing 0.8.0, and will redirect to use native yarn-cluster mode when users set `zeppelin.cluster_manager=yarn`. Does it make sense?

zjffdu: make sense

echarles:
Late question: This implementation uses the raw Yarn APIs. Did you consider using higher-level APIs such as Twill (http://twill.apache.org) or Slider (https://slider.incubator.apache.org) ?

zjffdu:
Thanks @echarles I also thought about it before, but not sure whether these 2 project still active. Maybe we need do some investigation.

jongyoul:
@echarles Hi, I saw Twill and met one of committers at Apache BigData NA. I agree with @zjffdu's opinion. After merging this PR, I'll take a look into that project.

pom.xml
jongyoul: I have changed some dependencies to use `hadoop-client` but it didn't include `hadoop-yarn-client`. So I left `hadoop-yarn-client`, too.

jongyoul:
Fixed all comments.

zeppelin-cluster/yarn/src/test/java/org/apache/zeppelin/cluster/yarn/YarnClientTest.java
```diff
@@ -0,0 +1,108 @@
...
+    String commonClasspath = parentClasspath + "/common/target";
+    String yarnClasspath = parentClasspath + "/yarn/target";
+    System.out.println("parentClasspath: " + parentClasspath);
+    System.out.println("commonClasspath: " + commonClasspath);
+    System.out.println("yarnClasspath: " + yarnClasspath);
+    String interpreterClasspath = Paths.get(parentClasspath).getParent().toAbsolutePath().toString() + "/zeppelin-interpreter/target";
+    ((RemoteInterpreterYarnProcess)remoteInterpreterProcess).setExtraClasspath(commonClasspath + "/classes" + ApplicationConstants.CLASS_PATH_SEPARATOR + commonClasspath + "/lib/*" + ApplicationConstants.CLASS_PATH_SEPARATOR + yarnClasspath + "/classes" + ApplicationConstants.CLASS_PATH_SEPARATOR + yarnClasspath + "/lib/*" + ApplicationConstants.CLASS_PATH_SEPARATOR + interpreterClasspath+ "/classes" + ApplicationConstants.CLASS_PATH_SEPARATOR + interpreterClasspath + "/lib/*");
+    remoteInterpreterProcess.start("user", false);
+    assertNotNull("Host should not be null", remoteInterpreterProcess.getHost());
+    assertFalse("Port should not be -1", -1 == remoteInterpreterProcess.getPort());
```
zjffdu: Could we do more testing in `YarnClientTest` ? Like start a `ShellInterpreter` in this MiniYarnCluster, and run interpret method for some shell command. The test now doesn't give me sufficient confidence that the YarnClient mode would work.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Constants.java
```diff
@@ -23,10 +23,9 @@
 
 /**
  * Interpreter related constants
- * 
- *
  */
 public class Constants {
```
zjffdu: should we move this zeppelin-cluster module ?

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
```diff
@@ -0,0 +1,168 @@
...
+  private boolean initialized;
+
+  public ClusterManagerFactory(String zeppelinHome, String defaultClusterKey) {
+    this.zeppelinHome = zeppelinHome;
+    this.defaultClusterKey = defaultClusterKey;
+    this.clusterManagerMap = Maps.newHashMap();
+    this.initialized = false;
+  }
+
+  public ClusterManager getClusterManager() {
```
zjffdu: rename it to getDefaultClusterManager()

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
```diff
@@ -0,0 +1,168 @@
...
+    return getClusterManager(defaultClusterKey);
+  }
+
+  public ClusterManager getClusterManager(String key) {
```
zjffdu: key -> name

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
```diff
@@ -0,0 +1,168 @@
...
+  private static final String CLUSTER_COMMON_DIR_NAME = "common";
+  private static final String CLUSTER_CLASS_NAME_FILE = "clustermanager-class";
+
+  private final String zeppelinHome;
+  private final String defaultClusterKey;
+  private final Map<String, ClusterManager> clusterManagerMap;
+
+  private boolean initialized;
+
+  public ClusterManagerFactory(String zeppelinHome, String defaultClusterKey) {
```
zjffdu: defaultClusterKey -> defaultClusterManagerName

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
```diff
@@ -0,0 +1,168 @@
...
+
+    try (InputStream classNameInputstream =
+             classLoader.getResourceAsStream(CLUSTER_CLASS_NAME_FILE);
+         StringWriter classNameWriter = new StringWriter()) {
+      if (null != classNameInputstream) {
+        IOUtils.copy(classNameInputstream, classNameWriter);
+        className = classNameWriter.toString();
+        Class clazz = classLoader.loadClass(className);
+        Object cm = clazz.getConstructor().newInstance();
+        logger.info("Class loaded. {}", clazz.getName());
```
zjffdu: Change logging to `ClusterManager {} is loaded with class {}`

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
```diff
@@ -0,0 +1,168 @@
...
+    ClassLoader classLoader = new URLClassLoader(jars.toArray(new URL[0]), parentClassLoader);
+    String className = null;
+
+    try (InputStream classNameInputstream =
+             classLoader.getResourceAsStream(CLUSTER_CLASS_NAME_FILE);
+         StringWriter classNameWriter = new StringWriter()) {
+      if (null != classNameInputstream) {
+        IOUtils.copy(classNameInputstream, classNameWriter);
+        className = classNameWriter.toString();
+        Class clazz = classLoader.loadClass(className);
```
zjffdu: This classloader only load ClusterManager class, but what about its dependencies ? How can they be loaded at runtime ?

zeppelin-cluster/yarn/src/test/java/org/apache/zeppelin/cluster/yarn/YarnClientTest.java
jongyoul: @zjffdu The main feature of this cluster manager is to start remote process in another machine and then return those two information. I think it's enough and if we start some interpreter while testing it. it would be guaranteed some interpreters already is built before the test and it's too hard to set specific order for it.

zeppelin-cluster/yarn/src/main/java/org/apache/zeppelin/cluster/yarn/Client.java
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.zeppelin.cluster.ClusterManager;
+import org.apache.zeppelin.helium.ApplicationEventListener;
+import org.apache.zeppelin.interpreter.InterpreterException;
+import org.apache.zeppelin.interpreter.remote.RemoteInterpreterProcess;
+import org.apache.zeppelin.interpreter.remote.RemoteInterpreterProcessListener;
+
+/**
+ * This class is in charge of making yarn client and implementing cluster manager.
+ */
+public class Client extends ClusterManager {
```
zjffdu: Could we rename it to YarnClusterManager ? `Client` is a little misleading

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
jongyoul: All dependencies were loaded in this classloader before loading class

zjffdu:
@jongyoul  I try to run shell interpreter in yarn mode.

```
 INFO [2017-07-06 10:32:28,538] ({qtp633070006-49} AbstractService.java[noteFailure]:272) - Service org.apache.hadoop.yarn.client.api.impl.YarnClientImpl failed in state STARTED; cause: org.apache.hadoop.yarn.exceptions.YarnRuntimeException: Failed to load class: [org.apache.hadoop.yarn.api.impl.pb.client.ApplicationClientProtocolPBClientImpl]
org.apache.hadoop.yarn.exceptions.YarnRuntimeException: Failed to load class: [org.apache.hadoop.yarn.api.impl.pb.client.ApplicationClientProtocolPBClientImpl]
    at org.apache.hadoop.yarn.factories.impl.pb.RpcClientFactoryPBImpl.getClient(RpcClientFactoryPBImpl.java:67)
    at org.apache.hadoop.yarn.ipc.HadoopYarnProtoRPC.getProxy(HadoopYarnProtoRPC.java:48)
    at org.apache.hadoop.yarn.client.RMProxy$1.run(RMProxy.java:140)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:360)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1608)
    at org.apache.hadoop.yarn.client.RMProxy.getProxy(RMProxy.java:136)
    at org.apache.hadoop.yarn.client.RMProxy.createRMProxy(RMProxy.java:99)
    at org.apache.hadoop.yarn.client.ClientRMProxy.createRMProxy(ClientRMProxy.java:72)
    at org.apache.hadoop.yarn.client.api.impl.YarnClientImpl.serviceStart(YarnClientImpl.java:174)
    at org.apache.hadoop.service.AbstractService.start(AbstractService.java:193)
    at org.apache.zeppelin.cluster.yarn.Client.start(Client.java:76)
    at org.apache.zeppelin.cluster.yarn.Client.createInterpreter(Client.java:106)
    at org.apache.zeppelin.interpreter.remote.RemoteInterpreter.getInterpreterProcess(RemoteInterpreter.java:196)
    at org.apache.zeppelin.interpreter.remote.RemoteInterpreter.getScheduler(RemoteInterpreter.java:527)
    at org.apache.zeppelin.interpreter.LazyOpenInterpreter.getScheduler(LazyOpenInterpreter.java:129)
    at org.apache.zeppelin.notebook.Note.run(Note.java:656)
    at org.apache.zeppelin.socket.NotebookServer.persistAndExecuteSingleParagraph(NotebookServer.java:1800)
    at org.apache.zeppelin.socket.NotebookServer.runParagraph(NotebookServer.java:1760)
    at org.apache.zeppelin.socket.NotebookServer.onMessage(NotebookServer.java:283)
    at org.apache.zeppelin.socket.NotebookSocket.onWebSocketText(NotebookSocket.java:59)
    at org.eclipse.jetty.websocket.common.events.JettyListenerEventDriver.onTextMessage(JettyListenerEventDriver.java:128)
    at org.eclipse.jetty.websocket.common.message.SimpleTextMessage.messageComplete(SimpleTextMessage.java:69)
    at org.eclipse.jetty.websocket.common.events.AbstractEventDriver.appendMessage(AbstractEventDriver.java:65)
    at org.eclipse.jetty.websocket.common.events.JettyListenerEventDriver.onTextFrame(JettyListenerEventDriver.java:122)
    at org.eclipse.jetty.websocket.common.events.AbstractEventDriver.incomingFrame(AbstractEventDriver.java:161)
    at org.eclipse.jetty.websocket.common.WebSocketSession.incomingFrame(WebSocketSession.java:309)
    at org.eclipse.jetty.websocket.common.extensions.ExtensionStack.incomingFrame(ExtensionStack.java:214)
    at org.eclipse.jetty.websocket.common.Parser.notifyFrame(Parser.java:220)
    at org.eclipse.jetty.websocket.common.Parser.parse(Parser.java:258)
    at org.eclipse.jetty.websocket.common.io.AbstractWebSocketConnection.readParse(AbstractWebSocketConnection.java:632)
    at org.eclipse.jetty.websocket.common.io.AbstractWebSocketConnection.onFillable(AbstractWebSocketConnection.java:480)
    at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:544)
    at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635)
    at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555)
Caused by: java.lang.ClassNotFoundException: Class org.apache.hadoop.yarn.api.impl.pb.client.ApplicationClientProtocolPBClientImpl not found
    at org.apache.hadoop.conf.Configuration.getClassByName(Configuration.java:1980)
    at org.apache.hadoop.yarn.factories.impl.pb.RpcClientFactoryPBImpl.getClient(RpcClientFactoryPBImpl.java:65)
    ... 35 more
ERROR [2017-07-06 10:32:28,539] ({qtp633070006-49} NotebookServer.java[persistAndExecuteSingleParagraph]:1802) - Exception from run
org.apache.hadoop.yarn.exceptions.YarnRuntimeException: Failed to load class: [org.apache.hadoop.yarn.api.impl.pb.client.ApplicationClientProtocolPBClientImpl]
    at org.apache.hadoop.yarn.factories.impl.pb.RpcClientFactoryPBImpl.getClient(RpcClientFactoryPBImpl.java:67)
    at org.apache.hadoop.yarn.ipc.HadoopYarnProtoRPC.getProxy(HadoopYarnProtoRPC.java:48)
    at org.apache.hadoop.yarn.client.RMProxy$1.run(RMProxy.java:140)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:360)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1608)
    at org.apache.hadoop.yarn.client.RMProxy.getProxy(RMProxy.java:136)
    at org.apache.hadoop.yarn.client.RMProxy.createRMProxy(RMProxy.java:99)
    at org.apache.hadoop.yarn.client.ClientRMProxy.createRMProxy(ClientRMProxy.java:72)
```

jongyoul:
Hm... let me check

jongyoul:
@zjffdu Can you check it again? I've reproduced and fixed it.

zjffdu:
@jongyoul unfortunately, I still have this issue. And the only place you use custom classloader is in `ClusterManagerFactory.java`. But I only see you use it to load the ClusterManager class (like org.apache.zeppelin.cluster.yarn.Client), I don't find you use it to load other hadoop classes.  Let me know if I miss anything.

jongyoul:
I changed my code of passing custom classloader to clustermanager and it would be passed into Configuration of hadoop

zjffdu:
This is where you use custom ClassLoader.  Seems YarnClient didn't use the custom class loader. 
```
this.configuration.setClassLoader(classLoader);
this.yarnClient = YarnClient.createYarnClient();
```

BTW, does it work for your enviroment ?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Constants.java
jongyoul: Fixed

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
jongyoul: Fixed

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
jongyoul: Fixed

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
jongyoul: Fixed

zeppelin-cluster/common/src/main/java/org/apache/zeppelin/cluster/ClusterManagerFactory.java
jongyoul: Fixed

zeppelin-cluster/yarn/src/main/java/org/apache/zeppelin/cluster/yarn/Client.java
jongyoul: Fixed

jongyoul:
CI passed: https://travis-ci.org/jongyoul/zeppelin/builds/251742660

zjffdu:
@jongyoul I still could not get shell interpreter work in yarn mode. But this time it is due to AM fail to launch. Here's the log in AM container.
```
Error: Could not find or load main class org.apache.zeppelin.cluster.yarn.YarnRemoteInterpreterServer
```

I would suggest you to add unit test to verify it (Launch ShellInterpreter or some dummy interpreter in MiniYarnCluster)

zjffdu:
@jongyoul Any updates ?

jongyoul:
Sorry for the late reply. I've tried to fix it but now I'm doing other things. Please wait for a moment

PR #2330, opened 2017-05-10
[minor] Remove incorrect configuration and fix description of LdapRealm

VipinRathor:
### What is this PR for?
There is wrong description and incorrect configuration defined in comment in LdapRealm.java. Since the documentation at (http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/security/shiroauthentication.html) does not mention LdapRealm, the information should be correct in the code comments at the least.

### What type of PR is it?
Documentation

### What is the Jira issue?
N/A

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/LdapRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/LdapRealm.java
@@ -78,22 +78,24 @@ import javax.naming.ldap.PagedResultsControl;
  * <p>Sample config for <tt>shiro.ini</tt>:
  * 
  * <p>[main] 
- * ldapRealm=org.apache.zeppelin.realm.LdapRealm
- * ldapRealm.contextFactory=$ldapGroupContextFactory
- * ldapRealm.contextFactory.authenticationMechanism=simple
- * ldapRealm.contextFactory.url=ldap://localhost:33389
- * ldapRealm.userDnTemplate=uid={0},ou=people,dc=hadoop,dc=apache,dc=org
+ * ldapRealm = org.apache.zeppelin.realm.LdapRealm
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
@VipinRathor any specific reason for chaning `ldapRealm = org.apache.zeppelin.realm.LdapRealm` to `ldapRealm=org.apache.zeppelin.realm.LdapRealm` in all the lines ?

prabhjyotsingh:
Also, I notice you have removed these two;
 - ldapRealm.contextFactory=$ldapGroupContextFactory
 - ldapRealm.contextFactory.systemAuthenticationMechanism=simple

any specific reason?

VipinRathor:
@prabhjyotsingh 
Thanks for reviewing this.

> any specific reason for chaning ldapRealm = org.apache.zeppelin.realm.LdapRealm to ldapRealm=org.apache.zeppelin.realm.LdapRealm in all the lines ?

Earlier there were lines with mixed spacing, I just wanted to keep it uniform. We can choose any one style that you want.

> I notice you have removed these two;
>
>    ldapRealm.contextFactory=$ldapGroupContextFactory
>    ldapRealm.contextFactory.systemAuthenticationMechanism=simple
>
> any specific reason?

Yes, if you keep the first one, Shiro will throw error that ldapGroupContextFactory is not defined anywhere so can't be used. And in fact, the sample configuration is actually not defining any such ldapGroupContextFactory. Instead `LdapRealm` is using `JndiLdapContextFactory` , so no need to override a contextFactory.

For `systemAuthenticationMechanism`, there is neither such configuration available in `JndiLdapContextFactory` nor `LdapRealm` introduced any. Looks like a left over from earlier configuration set, hence removed.

prabhjyotsingh:
Sure LGTM!

prabhjyotsingh:

PR #2331, opened 2017-05-10
[minor] Replace 'SIMPLE' auth with 'simple' auth everywhere in documents

VipinRathor:
### What is this PR for?
This PR fixes a security hole in documentation when 'SIMPLE' authentication mechanism is defined in Shiro configuration (http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/security/shiroauthentication.html). With that, user can log in without entering his/her password. Zeppelin documentation should recommend the correct value which is 'simple'.

### What type of PR is it?
Documentation

### What is the Jira issue?
N/A

### Questions:

### DIFF
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -43,7 +43,7 @@ user3 = password4, role2
 #ldapRealm.contextFactory.environment[ldap.searchBase] = dc=COMPANY,dc=COM
 #ldapRealm.contextFactory.url = ldap://ldap.test.com:389
 #ldapRealm.userDnTemplate = uid={0},ou=Users,dc=COMPANY,dc=COM
-#ldapRealm.contextFactory.authenticationMechanism = SIMPLE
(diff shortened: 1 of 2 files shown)


VipinRathor:
Thanks @felixcheung for spending time on this and approving.

> but we should fix it to lower case always just to be safe?

As per Shiro [doc](https://shiro.apache.org/static/1.2.4/apidocs/org/apache/shiro/realm/ldap/JndiLdapContextFactory.html) and JNDI LDAP [doc](https://docs.oracle.com/javase/tutorial/jndi/ldap/auth_mechs.html), the  acceptable values are 'none', 'simple', 'SASL' or 'DIGEST-MD5' etc. So converting AuthenticationMechanism value to lower case might not be a good idea.

Leemoonsoo:
LGTM and merge to master and branch-0.7 if no further comment

PR #2337, opened 2017-05-11
ZEPPELIN-2530: Zeppelin user impersonation with domain name suffix is failing

prabhjyotsingh:
### What is this PR for?
Basically what happens is, if a user login using full name with suffix then the user impersonation fails, as the HDFS expects username without the suffix.
This is because the username is passed to underlying components with suffix and got rejected in security layer with IllegalArgumentException

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2530](https://issues.apache.org/jira/browse/ZEPPELIN-2530)

 - Enable AD authentication 
 - set `activeDirectoryRealm.principalSuffix` in shiro.ini
 - now try to login with the full user name (in my example its zepplin@testdomain.com)

Before:
screen shot 2017-05-11 at 7 01 24 pm

After:
screen shot 2017-05-11 at 7 00 47 pm

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
@@ -202,6 +202,9 @@ public class ActiveDirectoryGroupRealm extends AbstractLdapRealm {
   }
 
   protected AuthenticationInfo buildAuthenticationInfo(String username, char[] password) {
+    if (this.principalSuffix != null && username.indexOf('@') > 1) {
+      username = username.split("@")[0];
+    }
     return new SimpleAuthenticationInfo(username, password, getName());
   }
 
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
@@ -186,7 +186,7 @@ public class ActiveDirectoryGroupRealm extends AbstractLdapRealm {
     LdapContext ctx = null;
     try {
       String userPrincipalName = upToken.getUsername();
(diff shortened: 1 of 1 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
```diff
@@ -202,6 +202,9 @@ protected AuthenticationInfo queryForAuthenticationInfo(
   }
 
   protected AuthenticationInfo buildAuthenticationInfo(String username, char[] password) {
+    if (this.principalSuffix != null && username.indexOf('@') > 1) {
+      username = username.split("@")[0];
```
felixcheung: hmm, check username is not empty/whitespace?

prabhjyotsingh: I'm assuming null/empty/whitespace user wouldn't reach here as it would have already failed from AD server, as not a valid user.
Null check is here https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java#L189

felixcheung:
The issue is the value could have been just @ or @foo.com

So after a split on @ the user could become an empty string or just whitespace

prabhjyotsingh:
Sure, I see, have added a check before sending it to the AD server.

VipinRathor:
Looks good to me. +1

Leemoonsoo:
LGTM

prabhjyotsingh:
Thank you @VipinRathor, @Leemoonsoo  for the review.
@felixcheung Does this looks OK?

prabhjyotsingh:
@felixcheung  Thank you for reviewing this.

Will merge this to master and branch-0.7, if no more discussion.

PR #2345, opened 2017-05-15
[MINOR] Hide pagination button if there is only 1 page in Helium menu

AhyoungRyu:
### What is this PR for?
It'll be better to hide the pagination button if there is only one page. 
([`only-of-type` CSS type selector](https://www.w3schools.com/cssref/sel_only-of-type.asp) can be used for this.)
Please see the attached screenshot img. 

### What type of PR is it?
Improvement

### What is the Jira issue?
N/A

### Screenshots (if appropriate)
 - Before 
before

 - After
screen shot 2017-05-15 at 4 34 09 pm

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #2346, opened 2017-05-15
[gh-pages] Hide pagination btn if there is only 1 page

AhyoungRyu:
### What is this PR for?
Same work with #2345 but for `gh-pages`

### What type of PR is it?
 Improvement

### What is the Jira issue?
N/A

 - Before
screen shot 2017-05-15 at 4 45 45 pm

 - After
screen shot 2017-05-15 at 4 45 53 pm

### Questions:

### DIFF
--- a/assets/themes/zeppelin/css/style.css
+++ b/assets/themes/zeppelin/css/style.css
@@ -693,6 +693,10 @@ a.anchorjs-link:hover { text-decoration: none; }
   border-bottom-right-radius: 3px;
 }
 
+.hide-last-boundaries.pagination:only-of-type {
+  display: none;
+}
+
 /* For what's new section */
 .new {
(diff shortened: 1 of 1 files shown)


AhyoungRyu:
`gh-pages` doesn't have CI check. So the red light is irrelevant with this change. 
Ready to review.

Leemoonsoo:
LGTM

soralee:
Nicely Improvement it, Thanks @AhyoungRyu!
But, tested and I found a bug which if I moved a last page, the page buttons is disappeared so, I couldn't move to before pages. I attached the following screenshot.

[Before]
before_page_number

[After]
after_page_number

Could you confirm this problem?

jongyoul:
@AhyoungRyu Can you please follow up this PR? If you don't want to proceed it now, can you close this PR?

PR #2347, opened 2017-05-16
[Bug Fix][ZEPPELIN-2554] sql parser fix (backslash)

tinkoff-dwh:
### What is this PR for?
Fix of parser to correctly parse backslash in quotes.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2554

Execute `select '\n', ';'`

before
before
text is parsed as 2 queries `select '\n', '` and `'`

after
after

### Questions:


zjffdu:
\cc @prabhjyotsingh

zjffdu:
This is for multiple line sql support, seems this is the only way. @prabhjyotsingh has more context.

prabhjyotsingh:
Thanks, @zjffdu. @tinkoff-dwh  will try to review this soon.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -523,23 +529,19 @@ private boolean isDDLCommand(int updatedCount, int columnCount) throws SQLExcept
...
 
       if (character == '\\') {
-        antiSlash = true;
+        backslash = true;
       }
 
       if (character == '\'') {
-        if (antiSlash) {
-          antiSlash = false;
-        } else if (quoteString) {
```
prabhjyotsingh: Instead of doing all of this, I think the easier fix will be to replace this `else if(quoteString) {` with `if (quoteString && sql.charAt(item - 1) != '\\') {`

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
```diff
@@ -189,6 +189,20 @@ public void testSplitSqlQuery() throws SQLException, IOException {
   }
 
   @Test
+  public void testSqlQueryWithBackslash() throws SQLException, IOException {
+    String sqlQuery = "select '\\n', ';';" +
+        "select replace('A\\;B', '\\', 'text')";
```
prabhjyotsingh: This `select replace('A\\;B', '\\', 'text')` looks incomplete, IMO it should be this `select replace('A\;B', '\'', 'text');`

Can you add one more line to test as well i.e. `select '\'', ';'`

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
tinkoff-dwh: if previous character \ then process will not come to this place

'if (backslash) {
       query.append(character);
       backslash = false;
       continue;
  }'

tinkoff-dwh: so, i think current code is correct

prabhjyotsingh: What I was trying to refer is, the above works, but another way to do the same is just by replacing this line with the above-mentioned changes.

Since I got the inspiration from https://github.com/postgres/pgadmin3/blob/794527d97e2e3b01399954f3b79c8e2585b908dd/pgadmin/dlg/dlgProperty.cpp#L999-L1045 and hence I'm of the opinion to make changes to a minimum.

tinkoff-dwh: found another problem   `select '\'`

prabhjyotsingh: this is supposed to fail, right? correct should be `select  '\''`, any thoughts?

tinkoff-dwh: nope, if we want select backslash, query is correct (or use in `relace('text\text', '\', '')`)

tinkoff-dwh: we have 2 queries (both are correct), but parser works incorrect
select '\''
select '\'

tinkoff-dwh: althought select '\'' it is incorrect

prabhjyotsingh: I see, I just checked on my local, reverted all of this, just removing the else part i.e. just had 

```
if (character == '\'') {
        if (antiSlash) {
          antiSlash = false;
        }
        if (quoteString) {
          quoteString = false;
        } else if (!doubleQuoteString) {
          quoteString = true;
        }
      }
```

And it works for all of these string; in the test case. What do you think?

```
String sqlQuery = "insert into test_table(id, name) values ('a', ';\"');" +
            "select * from test_table;" +
            "select * from test_table WHERE ID = \";'\";" +
            "select * from test_table WHERE ID = ';';" +
            "select '\\n', ';';" +
            "select replace('A\\;B', '\\', 'text');" +
            "select '\';"     +
            "select '\\', ';'"
```

tinkoff-dwh: different DB has the different solutions for escape single quote...

tinkoff-dwh: I see only one solution, allow 1 request per paragraph

tinkoff-dwh: or `select '\''`   incorrect...

tinkoff-dwh: searched information, single quote, usually escaped by single quote '', so  `select '\''` incorrect

prabhjyotsingh: Sure, thanks was about to post the same.

How about we keep changes to minimum, and have a flag in interpreter setting `support multiple queries in one paragraph`.

felixcheung:
Perhaps multiple queries was the reason we tried to parse the text (we really need to document these things better) - though I'm not sure how it would work exactly? Do you get multiple table in result if there are multiple queries in one paragraph? Does it even handle the multiple results correctly since that was added after JDBC interpreter is implemented?

In any case I do think it is much cleaner not to mess with the content.

Perhaps default to single query with a switch to fallback to old code and see if we get any feedback on this.

prabhjyotsingh:
@felixcheung  +1 for defaulting to the single query.

And for all of your questions answer is yes.

felixcheung:
hmm, I see, thanks

tinkoff-dwh:
@prabhjyotsingh 
fixed problems

prabhjyotsingh:
Tested this on local, looks OK. 

Can you also implement this comment https://github.com/apache/zeppelin/pull/2347#issuecomment-302747859, i.e. default to single query with a switch to make multiple queries work?

tinkoff-dwh:
@prabhjyotsingh 
checkbox (for paragraph) which indicates to parse the request or not?

prabhjyotsingh:
IMO property in settings of interpreter makes more sense, with a bit of a documentation on how to enable/disable it.
cc: @felixcheung

felixcheung:
yes for interpreter settings

tinkoff-dwh:
@felixcheung   @prabhjyotsingh 
have a problem with single mode, if user execute multiple queries (all text executes as single query) then result set is incorrect. 

I think this setting is redundant. If the user wants to execute single query, he can write a single query

prabhjyotsingh:
Which I think is fine, because in that case, we will not try to use this logic, and at least there will be a workaround if the user is facing any problem because of this issue, and we can have a document for this.

tinkoff-dwh:
single mode is default and the user will always get an incorrect result if he writes multiple queries to in paragraph

prabhjyotsingh:
Which is fine, with the document there will be a workaround.
And the case that you are talking is true for Postgres, but for Hive if you will try to execute multiple SQLs in one statement then it fails.

tinkoff-dwh:
I think is better without this parameter, what to do with the incorrect result. if there are problems, then think how to solve it. Postgres by default, and many use it and all users will have incorrect data

prabhjyotsingh:
IMO there should be a workaround if the logic fails.

And for handling error scenarios, here is what I think will happen.
Before recommendation:
screen shot 2017-05-23 at 12 11 45 pm

After recommendation:
screen shot 2017-05-23 at 12 12 31 pm

tinkoff-dwh:
if all quieries is valid then the result will return from the first. is it correct?

prabhjyotsingh:
Yes, correct.

tinkoff-dwh:
another solution, but I think it's better

prabhjyotsingh:
Sure, I do see this working for Postgres as posted in the screenshot, do you think this will work for rest of the data sources listed here http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/interpreter/jdbc.html#overview? 

Will try to test this on some of these and update this thread.

tinkoff-dwh:
if jdbc driver does not implement this method (getMoreResults), it will work like you said will return the first result

prabhjyotsingh:
Tried the above:

With multiple queries;
screen shot 2017-05-23 at 7 40 16 pm

With single query;
screen shot 2017-05-23 at 7 41 02 pm

And hence I repeat (https://github.com/apache/zeppelin/pull/2347#discussion_r117439158) keep changes to a minimum, the community has tested it, let's not try to break. And add an option in interpreter where it can default to single query mode (not use this logic).

tinkoff-dwh:
Ready to review

prabhjyotsingh:
looks OK, let me test it out on my cluster, I'll update this thread.

prabhjyotsingh:
Tried locally works fine.
@tinkoff-dwh can you resolve these conflicts.

prabhjyotsingh:

PR #2348, opened 2017-05-16
ZEPPELIN-2555: Run all paragraphs API does not run all paragraph as front-end user.

prabhjyotsingh:
### What is this PR for?
Run all paragraphs API (http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/rest-api/rest-notebook.html#run-all-paragraphs) does not run all paragraph as front-end user.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-2555](https://issues.apache.org/jira/browse/ZEPPELIN-2555)

Testing steps in screenshot

Before:
zeppelin-2555-before

After:
zeppelin-2555-after

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -588,17 +588,17 @@ public class NotebookRestApi {
       throws IOException, IllegalArgumentException {
     LOG.info("run note jobs {} ", noteId);
     Note note = notebook.getNote(noteId);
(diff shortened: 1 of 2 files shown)


1ambda:
Tested and works as described. LGTM.

zjffdu:
I was thinking that we should add test for the rest api. And we should create a test framework to make the test easier. Otherwise it is easy to break the api. And this is not the first time I see the rest api broken. But anyway we could do it in a separated ticket.

prabhjyotsingh:
@1ambda  Thank you for reviewing this.
@zjffdu Yes make sense, let's start a different thread, discuss and fix it there.

prabhjyotsingh:

PR #2353, opened 2017-05-19
[branch-0.7] missed try catch.

cloverhearts:
### What is this PR for?
Currently we can not build on branch-0.7
The cause is a missed try catch.
So, I fixed this part.

### What type of PR is it?
Hot Fix

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -338,7 +338,11 @@ public class Notebook implements NoteEventListener {
       note = notes.remove(id);
       folders.removeNote(note);
     }
-    interpreterSettingManager.removeNoteInterpreterSettingBinding(subject.getUser(), id);
+    try {
+      interpreterSettingManager.removeNoteInterpreterSettingBinding(subject.getUser(), id);
+    } catch (IOException e) {
+      logger.error(e.toString(), e);
+    }
     noteSearchService.deleteIndexDocs(note);
     notebookAuthorization.removeNote(id);
 


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -338,7 +338,11 @@ public void removeNote(String id, AuthenticationInfo subject) {
       note = notes.remove(id);
       folders.removeNote(note);
     }
-    interpreterSettingManager.removeNoteInterpreterSettingBinding(subject.getUser(), id);
+    try {
+      interpreterSettingManager.removeNoteInterpreterSettingBinding(subject.getUser(), id);
+    } catch (IOException e) {
+      logger.error(e.toString(), e);
```
1ambda: Is it ok not to return here? If we get errors while removing interpreter setting binding. (LINE 342)

cloverhearts: Your opinion is worth thinking about.
However, the intent of this fix is to make the build successful,
It is to unify the code of the master branch and branch-0.7.
Therefore, you can not modify it in this PR.
Sorry

cloverhearts: ref on master branch - https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java#L342

Leemoonsoo: Let's merge this PR as a hotfix and then address improvement in another PR.

cloverhearts:
merge

PR #2372, opened 2017-05-29
[ZEPPELIN-2581] Shell ansi codes

DrIgor:
### What is this PR for?
Support ANSI escape codes for colors in shell interpreter output

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2581

image

### Questions:

### DIFF
--- a/shell/pom.xml
+++ b/shell/pom.xml
@@ -69,6 +69,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.fusesource.jansi</groupId>
+      <artifactId>jansi</artifactId>
+      <version>1.15</version>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -38,6 +38,7 @@ import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
 import org.apache.zeppelin.scheduler.Scheduler;
 import org.apache.zeppelin.scheduler.SchedulerFactory;
 import org.apache.zeppelin.shell.security.ShellSecurityImpl;
+import org.fusesource.jansi.HtmlAnsiOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -72,7 +73,7 @@ public class ShellInterpreter extends Interpreter {
   public InterpreterResult interpret(String cmd, InterpreterContext contextInterpreter) {
     LOGGER.debug("Run shell command '" + cmd + "'");
     OutputStream outStream = new ByteArrayOutputStream();
-    
+
     CommandLine cmdLine = CommandLine.parse(shell);
     // the Windows CMD shell doesn't handle multiline statements,
     // they need to be delimited by '&&' instead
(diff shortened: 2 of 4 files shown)


1ambda:
Can we apply some theme to the output? For example,

Solarized dark theme + shell syntax highlight. 

image

AFAIK, we are using highlight.js

https://github.com/apache/zeppelin/blob/master/zeppelin-web/bower.json#L27

felixcheung:
Cool!
I'd love the dark theme, but I don't think we should force this onto users?

shell/pom.xml
```diff
@@ -69,6 +69,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.fusesource.jansi</groupId>
+      <artifactId>jansi</artifactId>
+      <version>1.15</version>
+    </dependency>
```
felixcheung: could you add the LICENSE?

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -85,11 +86,13 @@ public InterpreterResult interpret(String cmd, InterpreterContext contextInterpr
     try {
       DefaultExecutor executor = new DefaultExecutor();
       executor.setStreamHandler(new PumpStreamHandler(
-        contextInterpreter.out, contextInterpreter.out));
+        new HtmlAnsiOutputStream(contextInterpreter.out), contextInterpreter.out));
       executor.setWatchdog(new ExecuteWatchdog(Long.valueOf(getProperty(TIMEOUT_PROPERTY))));
       executors.put(contextInterpreter.getParagraphId(), executor);
+      contextInterpreter.out().write("%html <pre>");
```
felixcheung: if we are writing out html, do we need to escape the original output which might have `<` and so on?

DrIgor: `HtmlAnsiOutputStream` escapes output itself

https://github.com/fusesource/jansi/blob/jansi-project-1.16/jansi/src/main/java/org/fusesource/jansi/HtmlAnsiOutputStream.java#L66

shell/pom.xml
DrIgor: Yes, sure. It uses Apache License v2.

Where should I add this information?

DrIgor:
I am not sure If themes should be shell interpreter feature. Maybe we can provide themes for every interpreter's text output?

highlight.js should be used for input, not output. Output already contains information about colors and text styles

shell/pom.xml
felixcheung: https://github.com/apache/zeppelin/blob/master/LICENSE

DrIgor: Done

1ambda:
- the style is also a part of feature.
- if we don't need styling output, then there is no reason to add css code for JDBC output. (table)
- highlight.js is just a library. we can use anywhere we want to deliver the best.
- if you don't want to include styling in this PR, it's ok. we can improve in the following PR. but Personally, I don't see this is good style. 

image

Here are few examples of what shell output should be. And that's the reason why many people customizing their shell. 

image

image

1ambda:
I thought colorizing is included in this PR. But it's not part of this PR. Please forget about the last comment. I will create another JIRA ticket and work on that issue.

DrIgor:
I understand that `highlight.js` is a library and can be used anywhere. But we can't apply it to shell output because output is already highlighted. For example, `ls` inserts color codes to its output so we need to follow it.

We can provide customization for colors (RGB code for blue, RGB for green and so on) and background color.

Should I add this properties to shell interpreter?

1ambda:
@DrIgor FYI, see the custom initialization section

- https://highlightjs.org/usage/

Not sure colorizing should be handled in this PR. can be handled in JIRA 2604.

DrIgor:
Any other questions?

1ambda:
I failed to test this feature since terminal i am using doesnt support that option. others can help review. sorry for the late reply.

PR #2374, opened 2017-05-29
[ZEPPELIN-2593] Add storage settings to persist on run and commit

khalidhuseynov:
### What is this PR for?
adding settings to storage layer to let user choose to persist continuously or only on commits

### What type of PR is it?
Improvement | Feature

### Todos
* [x] - add notebook repo setting
* [x] - add persist on run & commit (checkpoint)
* [x] - add persist on commit (checkpoint) only
* [x] - add tests

### What is the Jira issue?
[ZEPPELIN-2593](https://issues.apache.org/jira/browse/ZEPPELIN-2593)

1. go to notebook repo menu and click on edit `Global settings` or do through env var `ZEPPELIN_NOTEBOOK_PERSISTENCE`
2. change value of `Note persistence` item to `Persist on note run and commits` or change value of `ZEPPELIN_NOTEBOOK_PERSISTENCE` into "run"
3. go change any note, although it will be changed in memory, it won't be persisted
4. to check above can reload notes in note list view -> will see original note
5. can save note either by `Run` or clicking `Commit` button in version control menu of note

screen shot 2017-07-10 at 3 02 02 am

## Questions:


khalidhuseynov:
this is ready for review

1ambda:
Thanks for the contribution. Let me test and comment soon.

1ambda:
@khalidhuseynov I tested and works well in usual cases. However, 

1. Set `Persist on note commits`
2. Execute paragraph
3. Stop Zeppelin without committing. 

**In this case, The paragraph which was executed is not saved.**  But users normally expect all their changes will be saved. What do you think of?

khalidhuseynov:
@1ambda thanks for testing. yes that's the purpose of `save & commit`, the changes are applied in memory only unless it's explicitly saved and commit (persisted into file)

1ambda:
Can we make *stopping zeppelin* also trigger *persist action*? 

- usually, people expect *stopping zeppelin* will save their changes regardless of whatever mode they are using. they might think *Zeppelin lost my changes!*
- I guess, the original intention is that saving external storage everytime is a burden. But **stopping zeppelin** occurs only once. It's not a burden.

khalidhuseynov:
yes that's true, i'll look into saving when stopping zeppelin

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
```diff
@@ -89,12 +92,14 @@
...
   private final boolean useServerSideEncryption;
   private final ZeppelinConfiguration conf;
+  private boolean saveAndCommit;
 
   public S3NotebookRepo(ZeppelinConfiguration conf) throws IOException {
     this.conf = conf;
     bucketName = conf.getBucketName();
     user = conf.getUser();
     useServerSideEncryption = conf.isS3ServerSideEncryption();
+    saveAndCommit = conf.isPersistOnCommit();
```
zjffdu: It doesn't make sense to ask every NotebookRepo implementation do the `saveAndCommit` logic. It is supposed to be in a base class.

zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -641,6 +645,7 @@ public String getJettyName() {
     ZEPPELIN_NOTEBOOK_MONGO_AUTOIMPORT("zeppelin.notebook.mongo.autoimport", false),
     ZEPPELIN_NOTEBOOK_STORAGE("zeppelin.notebook.storage", GitNotebookRepo.class.getName()),
     ZEPPELIN_NOTEBOOK_ONE_WAY_SYNC("zeppelin.notebook.one.way.sync", false),
+    ZEPPELIN_NOTEBOOK_PERSIST_ON_COMMIT("zeppelin.notebook.persist.on.commmit", false),
```
zjffdu: commmit  -> commit

zjffdu:
What is the current persistent policy ? Does it persist on every run ? 

And the property name `zeppelin.notebook.persist.on.commit` is a little confusing to me. 
* If it is set to false, does it mean it would not persist when user click the saveAndCommit button ? 
* From the Database perspective (since commit is a concept from database), commit should mean persisting. If commit doesn't mean persisting, this might confuse users. 
* So I would suggest to use `zeppelin.notebook.persist.on.run` which mean doing persisting on each run.
* BTW, we should also add doc to mention what is the purpose of this property. I guess it is for performance, but with the risk of losing data.

What do you think ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
khalidhuseynov: Good point, actually originally it was implemented in base class as a global parameter, but after the comments and actual goal of this PR of optimizing "saves" for remote storages, it was changed to target specific storages (currently s3 and zhub). Thus if this setting is fine grained per each storage (e.g. can be disabled in s3 and enabled in zhub storages, and absent in all other storages) they need to handle it separately. The original environment variable just gives initial default values that can be changed after start of server (although can't be persistent when stopped). Does it make sense?

khalidhuseynov:
@zjffdu the persistence policy is saving on run and almost on any note action. originally i introduced this feature globally on base class, however after some discussion above, introducing it for certain classes seems to be more reasonable since remote saves are costlier than local ones. regarding the questions: 
 * when it's set to false there's only `Commit` button as before (make revision) instead of `Save & Commit` (save and make revision) as below 
save commit1
* yeah i agree that `commit` itself implies save. currently it's used in Zeppelin to denote creation of revision/checkpoint. So the idea was that in normal mode everything is saved automatically and you can `commit` current version whenever you want and with this feature enabled (only in S3 and ZHubRepo for now) you can `Save & Commit`.
* persist on run is also another idea that could work, and would be nice to know opinion of more people on that. generally in that case save would be dependent on run, and `commit` would stay as before. @1ambda @Leemoonsoo what do you think about that?
* i'll add documentation for it, regarding losing data since i changed local storage not to have this feature, it'll alway be saved locally and only for remote storages it could be mentioned in docs.

let me know what you think on these points, thanks

khalidhuseynov:
this is ready for more review

1ambda:
> persist on run is also another idea that could work, and would be nice to know opinion of more people on that. generally in that case save would be dependent on run, and commit would stay as before. @1ambda @Leemoonsoo what do you think about that?

One suggestion. Can we provide a dropdown for this option? 

- persist on running
- persist on commit

khalidhuseynov:
@1ambda good point, we can make both possible. i'll modify it to support both config then through dropdown

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java
khalidhuseynov: @zjffdu addressed the changes and moved the logic into base and `Notebook` classes

khalidhuseynov:
@1ambda @zjffdu addressed the feedback and added persist on `run` as well. Also moved the logic out of storage classes. Please help further review and test this one to move on.

1ambda:
@khalidhuseynov Thanks for the update.

1. Could you describe the reason why we need to have separated setting tabs? I mean, what note persistence setting should be under `Global Settings` 

image

2. Could you describe what the option `Persist notes on run and checkpoint` means? 

image

- a note is persisted when the note was executed **and** (**or**?) user committed it
- https://github.com/khalidhuseynov/zeppelin/blob/e3636b44a421562bef7d225b8d58f1fee6746658/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java#L31

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -748,6 +749,16 @@ private void removeAllAngularObjectInParagraph(String user, String paragraphId)
   }
 
   public void persist(AuthenticationInfo subject) throws IOException {
```
1ambda: Who will call the `persist` method? and what the meaning of the prefix `force` in front of the `forcePersist`?

khalidhuseynov: `forcePersist` is for `run` related calls that would be persisted always. persist on the other hand have added condition check if `persist on run` enabled or not, and it's called by other non `run` related calls

khalidhuseynov:
@1ambda 
1. separate tab with `Global Settings` contains the settings which are applied to all storages, so that we don't duplicate in each storage setting.
2. note persisted when user executes, also it's persisted when user makes commit

khalidhuseynov:
@felixcheung actually that means to persist in default way, since Zeppelin persists on most of user activities so it seems to be "continuously". another way would be to call it "default" i guess

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
zjffdu: Agree that `forcePersist` is a  little confusing. Without reading the code, it is hard to guess what does it mean. 
How about name it as `persist(PersistenceStragtegy strategy)` ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -55,14 +59,27 @@
...
+  private static final String GLOBAL_SETTINGS_NAME = "Global Settings";
 
   /**
+   * Options for persisting notes on actions.
+   */
+  public enum NotePersist {
+    CONTINUOUS,       // Default value when all edit actions initiate note persist
+    RUN,              // Persist notes on run and checkpoint (commit) actions 
+    CHECKPOINT        // Persist note on checkpoint (commit) actions only
+  }
```
zjffdu: Why putting `NotePersist ` here ? Does it only for `NotebookRepoSync` ?

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -1204,6 +1208,106 @@ public void testPublicPrivateNewNote() throws IOException, SchedulerException {
...
+    assertNotNull(notebook.getNote(note.getId()).getParagraphs());
+    assertThat(notebook.getNote(note.getId()).getParagraphs().size()).isEqualTo(3);
+    // no notes on disk
+    assertThat(repoManager.list(anonymous).size()).isEqualTo(0);
+    
+    // when
+    note.runAll();
+
+    // wait for finish
+    while(p3.isTerminated() == false || p3.getResult() == null) {
```
zjffdu: persist happens after the paragraph execution finish ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
khalidhuseynov: it applies to base class `NotebookRepoSync` and consecutively to all repo instances when saving. instead of duplicating it in every repo, it's better to keep it here as global setting

felixcheung:
hmm I see. I can see "continuously" can make sense. It's a bit vague though. Perhaps "on all changes"? sorry that might not be better :)

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
khalidhuseynov: persist happens before execution and after any paragraph status change in https://github.com/apache/zeppelin/pull/2374/files#diff-bcbdda87aa29314ed459e8212067e313R2223

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
khalidhuseynov: agree, that's more clear. i'll look into that

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
zjffdu: Ideally, persist could happen in a separated thread when paragraph is running, so that persist won't block the execution. Anyway,  this could be done in a separated PR as another performance improvement.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
zjffdu: NotebookRepoSync is a base class ? Isn't it a implementation of `NotebookRepo` and `NotebookRepo` is the base class

felixcheung:
where are we on this?

khalidhuseynov:
thanks for the ping, i'll check the state and update the branch by end of the week

PR #2377, opened 2017-05-30
User does not land on Zeppelin UI Login page after clicking logout button when Zeppelin is running behind proxy server.

prabhjyotsingh:
### What is this PR for?
The user is no more on Zeppelin UI Login page after clicking logout button or links where user doesn't have access to when Zeppelin is running behind a proxy server like Nginx.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2601](https://issues.apache.org/jira/browse/ZEPPELIN-2601)

Set up any reverse proxy server like Nginx, exmaple config below;

```
location /abc/def/zeppelin/ {
        proxy_pass http://localhost:8080/;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
}
```

And now from browser navigate to this URL, and click on any like for which user doesn't have access to. In my example `user1` did not have access to any of `/api/interpreter/**` `/api/configurations/**` and `/api/credential/**`

Before:
before

After:
after

### Questions:

### DIFF
--- a/zeppelin-web/src/components/baseUrl/baseUrl.service.js
+++ b/zeppelin-web/src/components/baseUrl/baseUrl.service.js
@@ -37,9 +37,7 @@ function baseUrlSrv () {
   }
 
   this.getBase = function() {
-    let aa = location.protocol + '//' + location.hostname + ':' +
-    this.getPort() + location.pathname
-    return aa
+    return location.protocol + '//' + location.hostname + ':' + this.getPort() + location.pathname
   }
 
   this.getRestApiBase = function () {
--- a/zeppelin-web/src/components/baseUrl/baseUrl.service.js
+++ b/zeppelin-web/src/components/baseUrl/baseUrl.service.js
@@ -36,10 +36,14 @@ function baseUrlSrv () {
       skipTrailingSlash(location.pathname) + '/ws'
   }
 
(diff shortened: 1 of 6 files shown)


prabhjyotsingh:
CI green, ready for review.

zeppelin-web/src/components/baseUrl/baseUrl.service.js
```diff
@@ -36,10 +36,14 @@ function baseUrlSrv () {
       skipTrailingSlash(location.pathname) + '/ws'
   }
 
+  this.getBase = function() {
+    let aa = location.protocol + '//' + location.hostname + ':' +
```
1ambda: MINOR: Can we change the `aa` to a more meaningful variable name?

1ambda:
Tested locally and works as described. LGTM except for a comment.

zeppelin-web/src/components/baseUrl/baseUrl.service.js
prabhjyotsingh: ooh! good catch, 've removed it.

prabhjyotsingh:

PR #2389, opened 2017-06-02
[ZEPPELIN-2612] Remove duplicates from ZeppelinConfiguration

DrIgor:
### What is this PR for?
Remove duplicated code from `ZeppelinConfiguration` and make proper use of commons-configuraion

And I propose using interpolation instead of `getRelativeDir` (Not in this PR).
For example,
```
ZEPPELIN_INTERPRETER_DIR("zeppelin.interpreter.dir", "${zeppelin.home}/interpreter")
```

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2612

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,7 @@
 
     <!-- test library versions -->
     <junit.version>4.12</junit.version>
+    <hamcrest-all.version>1.3</hamcrest-all.version>
     <mockito.version>1.10.19</mockito.version>
     <assertj.version>1.7.0</assertj.version>
     <powermock.version>1.6.4</powermock.version>
@@ -260,6 +261,13 @@
         <groupId>junit</groupId>
         <artifactId>junit</artifactId>
         <version>${junit.version}</version>
+        <scope>test</scope>
+      </dependency>
+
+      <dependency>
+        <groupId>org.hamcrest</groupId>
+        <artifactId>hamcrest-all</artifactId>
+        <version>${hamcrest-all.version}</version>
         <scope>test</scope>
       </dependency>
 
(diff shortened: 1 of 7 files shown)


zjffdu:
@DrIgor This is not a trivial change, please create a ticket and describe the purpose.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -171,7 +172,6 @@ public ZeppelinServer() throws Exception {
   public static void main(String[] args) throws InterruptedException {
 
     ZeppelinConfiguration conf = ZeppelinConfiguration.create();
-    conf.setProperty("args", args);
```
DrIgor: Usage was removed in https://github.com/apache/zeppelin/commit/0330452231c9dae85bb642fbd7a8b9cc16c6777d#diff-4bd397174ee0f21a8d3b788b97bc8711L487

pom.xml
```diff
@@ -109,6 +109,7 @@
 
     <!-- test library versions -->
     <junit.version>4.12</junit.version>
+    <hamcrest-all.version>1.3</hamcrest-all.version>
```
felixcheung: why is this needed?

felixcheung: if this is testing only, should it go to submodule pom file?
also we need to add this to LICENSE

DrIgor: `jUnit` depends on `hamcrest-core 1.3` and we already used it in the project. I placed it right after jUnit.

PR #2392, opened 2017-06-03
[HFix] : Fix JS exception because of Zeppelin-2519

vipul1409:
### What is this PR for?
Fix JS exception because of Zeppelin-2519. Added the missing dependency for scroll monitor before viewport.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2519

Load a notebook and you should not see any exceptions in the dev console.

### Questions:


Leemoonsoo:
Thanks @vipul1409 for the hot fix. Tested and it fixes JS exception. Merge to master as a hot fix.

1ambda:
@vipul1409 Thanks for the quick fix. 👍

PR #2395, opened 2017-06-05
[ZEPPELIN-2515] After 100 minutes R process quits silently and spark.r interpreter becomes unresponsive.

jongyoul:
### What is this PR for?
Enabling to set timeout value for R backend

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Remove timeout parameter

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2515

### Questions:

### DIFF
--- a/spark/src/main/resources/R/zeppelin_sparkr.R
+++ b/spark/src/main/resources/R/zeppelin_sparkr.R
@@ -31,7 +31,7 @@ print(paste("LibPath ", libPath))
 library(SparkR)
 
 
-SparkR:::connectBackend("localhost", port, 6000)
+SparkR:::connectBackend("localhost", port)
 
 # scStartTime is needed by R/pkg/R/sparkR.R
 assign(".scStartTime", as.integer(Sys.time()), envir = SparkR:::.sparkREnv)


spark/src/main/resources/R/zeppelin_sparkr.R
```diff
@@ -31,7 +31,7 @@ print(paste("LibPath ", libPath))
 library(SparkR)
 
 
-SparkR:::connectBackend("localhost", port, 6000)
+SparkR:::connectBackend("localhost", port)
```
zjffdu: This is for supporting spark 2.1, remove this would cause sparkr break in spark 2.1

zjffdu: And this is timeout for establishing connection, it seems it doesn't related to the issue here.

jongyoul: Can you check the original jira issue? I'm not good at R and sparkR but it looks related.

zjffdu: This is due to sparkR api change

https://github.com/apache/spark/blob/master/R/pkg/R/client.R#L22

jongyoul: Do you mean SparkR occurs error if I change like that? I think so. Let me check it again.

zjffdu: I mean this code change is due to SparkR api change in spark 2.1.  And I suspect whether sparkr shell has the same issue, if yes then I don't think zeppelin need to fix this.

felixcheung: this was made configurable
https://github.com/apache/spark/blob/422aa67d1bb84f913b06e6d94615adb6557e2870/R/pkg/R/sparkR.R#L157

piotrp88: @felixcheung 
`connectBackend <- function(hostname, port, timeout) { ... } ` doesn't specify a default value for parameter `timeout`, so any call made outside Spark need to specify it, overriding the parameters passed by user in `spark-env.sh` or in `spark-defaults.conf`.

felixcheung: to be clear, this was never an actual public API in Spark. This is an internal private method and that is being called here.

but in any case, I'm not sure I follow why the setting can't be set by setting `SPARKR_BACKEND_CONNECTION_TIMEOUT`

`connectionTimeout <- as.numeric(Sys.getenv("SPARKR_BACKEND_CONNECTION_TIMEOUT", "6000"))`

ArunMj: any update regarding this?

PR #2396, opened 2017-06-05
[ZEPPELIN-2451]: Add JDBC config option for calling connection.commit after paragraph execution

randerzander:
Updated version of: https://github.com/apache/zeppelin/pull/2284

### What is this PR for?
Adding config for automatically calling commit after JDBC paragraph execution

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2451

Create property zeppelin.jdbc.autocommit with value "true" under the JDBC interpreter, run an INSERT statement against a database. The insert should be committed and should be accessible in the database.

Create property zeppelin.jdbc.autocommit with value "false" under the JDBC interpreter, run an INSERT statement against a database. The insert should not be committed and will not be in the database.

### Questions:
No
No
Yes

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -132,6 +132,8 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private final String CONCURRENT_EXECUTION_COUNT = "zeppelin.jdbc.concurrent.max_connection";
   private final String DBCP_STRING = "jdbc:apache:commons:dbcp:";
 
+  private final String JDBC_AUTOCOMMIT = "zeppelin.jdbc.autocommit";
+
   private final HashMap<String, Properties> basePropretiesMap;
   private final HashMap<String, JDBCUserConfigurations> jdbcUserConfigurationsMap;
   private final HashMap<String, SqlCompleter> sqlCompletersMap;
@@ -667,6 +669,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
     ResultSet resultSet = null;
     String paragraphId = interpreterContext.getParagraphId();
     String user = interpreterContext.getAuthenticationInfo().getUser();
+    boolean commit = Boolean.valueOf(getProperty(String.format(JDBC_AUTOCOMMIT, propertyKey)));
 
     boolean splitQuery = false;
     String splitQueryProperty = getProperty(String.format("%s.%s", propertyKey, SPLIT_QURIES_KEY));
@@ -763,7 +766,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
       //In case user ran an insert/update/upsert statement
       if (connection != null) {
         try {
-          if (!connection.getAutoCommit()) {
+          if (!connection.getAutoCommit() && commit) {
             connection.commit();
           }
           connection.close();


randerzander:
retest

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -634,6 +636,7 @@ private InterpreterResult executeSql(String propertyKey, String sql,
     ResultSet resultSet = null;
     String paragraphId = interpreterContext.getParagraphId();
     String user = interpreterContext.getAuthenticationInfo().getUser();
+    boolean commit = Boolean.valueOf(getProperty(String.format(JDBC_AUTOCOMMIT, propertyKey)));
```
1ambda: I think If old JDBC interpreter setting doesn't have `zeppelin.jdbc.autocommit` option what you created in this PR, it's going to be `false` 
(https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#valueOf(java.lang.String))

Is it the same with current behavior?

tzolov:
The `JDBC interpreter` uses Connection's auto-commit status to decide whether forceful to commit  the connection. As it as pointed above `commit` is not supported by many `NoSql` Jdbc implementation. The Apache Calcite's Avatica (which Druid uses by the way) is one of them. 

The interpreter doesn't allow setting the desired connection auto-commit status. Since the JDBC API provides a standard [setAutoCommit(boolean)](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#setAutoCommit-boolean-) method to alter connections auto-commit status i believe this should be exposed as Explicit property. 

Shouldn't the `zeppelin.jdbc.autocommit` be implemented as: 
```
JDBCInterpreter#getConnection(..)
....
connection.setAutoCommit(use zeppelin.jdbc.autocommit value)
....
```

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
jongyoul: I've checked it's `false` by default

jongyoul:
LGTM

jongyoul:
@randerzander Can you fix the failure of CI?

jongyoul:
You'd better rebase from current master to make your branch newer.

randerzander:
@jongyoul I'm having trouble with CI.

Jenkins tests pass except 3 that relate to Python interpreter, I believe unrelated to these small JDBC interpreter changes. Are the Python tests broken generally?

AppVeyor seems stuck queued on the older version of this PR. Do you know of a way to refresh it?

tinkoff-dwh:
@randerzander 
autocommit is already in JDBCInterpreter. 
```
 finally {
      //In case user ran an insert/update/upsert statement
      if (connection != null) {
        try {
          if (!connection.getAutoCommit()) {
            connection.commit();
          }
          connection.close();
        } catch (SQLException e) { /*ignored*/ }
      } 
```

You may close this PR.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
tinkoff-dwh: String.format(JDBC_AUTOCOMMIT, propertyKey)

JDBC_AUTOCOMMIT have no markers to replace **propertyKey**

randerzander:
@jongyoul I fixed the Appveyor test, but having problems with the Jenkins Python tests.

I don't believe these are related to my changes.. can you suggest what I can try next?

PR #2397, opened 2017-06-06
[ZEPPELIN-2592] Ensure open stream is closed

necosta:
### What is this PR for?
Fix for missing close statement on input streams found on Helium Bundle factory

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2592

NA

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
@@ -212,29 +212,28 @@ public class HeliumBundleFactory {
     return new File(heliumBundleDirectory, pkgName + "/" + HELIUM_BUNDLE_CACHE);
   }
 
-  public static List<String> unTgz(File tarFile, File directory) throws IOException {
-    List<String> result = new ArrayList<String>();
-    InputStream is = new FileInputStream(tarFile);
-    GzipCompressorInputStream gcis = new GzipCompressorInputStream(is);
-    TarArchiveInputStream in = new TarArchiveInputStream(gcis);
-    TarArchiveEntry entry = in.getNextTarEntry();
-    while (entry != null) {
-      if (entry.isDirectory()) {
+  private static List<String> unTgz(File tarFile, File directory) throws IOException {
+    List<String> result = new ArrayList<>();
+    try (TarArchiveInputStream in = new TarArchiveInputStream(
+            new GzipCompressorInputStream(new FileInputStream(tarFile)))) {
+      TarArchiveEntry entry = in.getNextTarEntry();
+      while (entry != null) {
+        if (entry.isDirectory()) {
+          entry = in.getNextTarEntry();
+          continue;
+        }
+        File curfile = new File(directory, entry.getName());
+        File parent = curfile.getParentFile();
+        if (!parent.exists()) {
+          parent.mkdirs();
+        }
+        OutputStream out = new FileOutputStream(curfile);
+        IOUtils.copy(in, out);
+        out.close();
+        result.add(entry.getName());
         entry = in.getNextTarEntry();
-        continue;
       }
-      File curfile = new File(directory, entry.getName());
-      File parent = curfile.getParentFile();
-      if (!parent.exists()) {
-        parent.mkdirs();
-      }
-      OutputStream out = new FileOutputStream(curfile);
-      IOUtils.copy(in, out);
-      out.close();
-      result.add(entry.getName());
-      entry = in.getNextTarEntry();
     }
-    in.close();
     return result;
   }
 
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
@@ -48,7 +48,7 @@ import org.apache.zeppelin.conf.ZeppelinConfiguration;
  * Load helium visualization & spell
  */
 public class HeliumBundleFactory {
-  Logger logger = LoggerFactory.getLogger(HeliumBundleFactory.class);
+  private Logger logger = LoggerFactory.getLogger(HeliumBundleFactory.class);
   private static final String NODE_VERSION = "v6.9.1";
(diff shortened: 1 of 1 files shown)


necosta:
I was hoping I had more time to refactor some other streams usage but, for now, just created this pull request for the missing close stream in Helium. Any questions, please let me know. Thanks.

jongyoul:
LGTM

andreaTP:
to be 100% sure to close properly the stream the `close` method should be called within a `finally` statement of a `try`

jongyoul:
@andreaTP  try() doesn't call close when it occurs exception?

andreaTP:
`try-with-resources` is used only for `InputStream` here.
there is nothing on outputstream.

jongyoul:
@andreaTP Thanks for fixing it. BTW, I think we'd better use try-with-resources of outputStream

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
```diff
@@ -215,29 +215,28 @@ public File getHeliumPackageBundleCache(String pkgName) {
...
+          continue;
+        }
+        File curfile = new File(directory, entry.getName());
+        File parent = curfile.getParentFile();
+        if (!parent.exists()) {
+          parent.mkdirs();
+        }
+        OutputStream out = new FileOutputStream(curfile);
+        IOUtils.copy(in, out);
+        out.close();
```
jongyoul: @necosta Can you handle it with try-with-resource as well?

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
```diff
@@ -215,29 +215,28 @@ public File getHeliumPackageBundleCache(String pkgName) {
...
 
-  public static List<String> unTgz(File tarFile, File directory) throws IOException {
-    List<String> result = new ArrayList<String>();
-    InputStream is = new FileInputStream(tarFile);
-    GzipCompressorInputStream gcis = new GzipCompressorInputStream(is);
-    TarArchiveInputStream in = new TarArchiveInputStream(gcis);
-    TarArchiveEntry entry = in.getNextTarEntry();
-    while (entry != null) {
-      if (entry.isDirectory()) {
+  private static List<String> unTgz(File tarFile, File directory) throws IOException {
```
felixcheung: this changes it to `private`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
```diff
@@ -215,29 +215,28 @@ public File getHeliumPackageBundleCache(String pkgName) {
...
+    List<String> result = new ArrayList<>();
+    try (TarArchiveInputStream in = new TarArchiveInputStream(
```
felixcheung: by closing TarArchiveInputStream, does it close the inner nested streams too?

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
necosta: Well I did change it to private because this method was only used in this class. 
In this last commit I took a look at other methods are reduced visibility when applicable. Kinda of unrelated with the bug itself but does not hurt in my opinion.

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
necosta: That is my understanding yes.

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
necosta: Good find! Done in latest commit

necosta:
Ready for another review round. Thanks

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
felixcheung: I checked and they do

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
```diff
@@ -200,48 +200,47 @@ public void buildAllPackages(List<HeliumPackage> pkgs) throws IOException {
...
-  public static List<String> unTgz(File tarFile, File directory) throws IOException {
-    List<String> result = new ArrayList<String>();
-    InputStream is = new FileInputStream(tarFile);
-    GzipCompressorInputStream gcis = new GzipCompressorInputStream(is);
-    TarArchiveInputStream in = new TarArchiveInputStream(gcis);
-    TarArchiveEntry entry = in.getNextTarEntry();
-    while (entry != null) {
-      if (entry.isDirectory()) {
+  private static List<String> unTgz(File tarFile, File directory) throws IOException {
+    List<String> result = new ArrayList<>();
```
felixcheung: why change this from `List<String> result = new ArrayList<String>();`?

necosta: It's the same :) I mean, List<String> result = new ArrayList<>() implies the ArrayList is of type String. No need to code it.

necosta:
@felixcheung LG, but this is a fair bit more than ensure open stream is closed, so leave for other people to comment...  
I agree. This whole class is very new and work-in-progress. I'm using it and testing it in detail so trying to get out a more stable version. Just as a note, I pushed and reverted a commit on a bit of code that confused me. Looks that the npm package always needs to have root folder name "package". I'm keeping that logic.

necosta:
@felixcheung , @jongyoul , @Leemoonsoo , can this be merged? Thanks.

Leemoonsoo:
LGTM and merge to master if no further comment

PR #2398, opened 2017-06-06
[ZEPPELIN-2622] Add Zeppelin Git info REST API

necosta:
### What is this PR for?
Added a new REST API that provides more detailed info. on the exact source code Git commit Id that created the build. Useful to track down bugs and replicate issues.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Add more Git properties

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2622

Build Zeppelin. Open http://localhost:8080/api/git

Expected output: {"status":"OK","message":"Zeppelin Git info","body":{"commitId":"4c2d295","timestamp":"2017-06-06 19:00:01"}}

### Questions:
* Does this needs documentation? Y (Done)

### DIFF
--- a/docs/rest-api/rest-zeppelin-server.md
+++ b/docs/rest-api/rest-zeppelin-server.md
@@ -49,12 +49,11 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
       <td>200</td>
     </tr>
     <tr>
-      <td> Fail code</td>
-      <td> 500 </td>
+      <td>Fail code</td>
+      <td>500</td>
     </tr>
     <tr>
-      <td> sample JSON response
-      </td>
+      <td>sample JSON response</td>
       <td>
         <pre>
 {
@@ -62,12 +61,12 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
   "message": "Zeppelin version",
   "body": [
     {
-      "version": "0.8.0-SNAPSHOT",
-      "commitId": "abc0123",
-      "timestamp": "2017-01-02 03:04:05"
+      "version": "0.8.0",
+      "git-commit-id": "abc0123",
+      "git-timestamp": "2017-01-02 03:04:05"
     }
   ]
-}     
+}
         </pre>
       </td>
     </tr>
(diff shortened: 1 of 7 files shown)


Leemoonsoo:
Thanks @necosta for very useful improvement.

It's not a strong opinion, but how about make `api/version` return version and git hash instead of add new api endpoint `api/git`? I think commit hash and timestamp can be part of version information. What do you think?

necosta:
Thanks for the feedback @Leemoonsoo. Yes, I was 50-50 on having a new API or reusing the "version" one. I'll go with your suggestion and refactor this later this week. Thanks!

necosta:
Ready for another look. Thanks

jongyoul:
LGTM

docs/rest-api/rest-zeppelin-server.md
```diff
@@ -33,6 +33,46 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
...
+    <tr>
+      <td>URL</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/version```</td>
+    </tr>
+    <tr>
+      <td>Success code</td>
+      <td>200</td>
+    </tr>
+    <tr>
+      <td> Fail code</td>
```
felixcheung: there are extra space in here, next line and "sample JSON response" (<-- and a newline there)
could you check?

docs/rest-api/rest-zeppelin-server.md
```diff
@@ -33,6 +33,46 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
...
+      <td> sample JSON response
+      </td>
+      <td>
+        <pre>
+{
+  "status": "OK",
+  "message": "Zeppelin version",
+  "body": [
+    {
+      "version": "0.8.0-SNAPSHOT",
```
felixcheung: perhaps don't put `-SNAPSHOT` as an example? in case we forget to change this when we make release

docs/rest-api/rest-zeppelin-server.md
```diff
@@ -33,6 +33,46 @@ If you work with Apache Zeppelin and find a need for an additional REST API, ple
...
+      "commitId": "abc0123",
+      "timestamp": "2017-01-02 03:04:05"
```
felixcheung: should this be `gittimestamp` instead? having it fairly general as `timestamp` people might expect it to be the last modified time etc (which might be good idea to return)

docs/rest-api/rest-zeppelin-server.md
necosta: Good find. Fixed.

docs/rest-api/rest-zeppelin-server.md
necosta: I agree. Renamed commitId to git-commit-id and timestamp to git-timestamp.

pom.xml
```diff
@@ -519,6 +519,27 @@
         <version>${plugin.deploy.version}</version>
       </plugin>
 
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
```
felixcheung: I think this is a build-time only plugin so we don't need to report the license for this?

necosta: No idea. Who can help? I assumed we could use it without concerns

jongyoul: AFAIK, it doesn't need to update LICENSE or LICENSE_BIN if it's not included in binary. @Leemoonsoo Can you verify it?

Leemoonsoo: As long as git-commit-id-plugin binary is not included in Zeppelin binary package release, we don't need (and we shouldn't) add it in LICENSE_BIN. 

Same for source package. as long as it's not included in source package distribution, we don't need (and shouldn't) add it in LICENSE.

jongyoul:

necosta:
Thanks guys!

PR #2400, opened 2017-06-07
[ZEPPELIN-2473] ZeppelinRestApiTest fails on travis

pravin-dsilva:
### What is this PR for?
The test `testNoteCreateWithParagraphs`  fails with a `java.lang.NullPointerException`. The error occurs while getting the value of 'mode' from the Map graph. Added a fix to get the value correctly.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2473

Tests should pass on the CI

### Questions:


Leemoonsoo:
Thanks @pravin-dsilva for the fix!

LGTM and merge to master if no further comment.

PR #2402, opened 2017-06-09
[ZEPPELIN-2636] User role lookup via interfaces

volumeint:
### What is this PR for?
Adds an interface (UserLookup) that a realm can implement to support looking up users and roles.  This makes it easier to implement a Realm and allows Zeppelin to ask the configured realms for user and role information without having to know which implementation of Realm that it is interacting with.  The pre-exising lookup code that was contained in GetUserList.java has been moved into the Realm implementations. 
This PR includes a change to SecurityUtils to allow it to get the Roles of the currently authenticated user regardless of Realm used.  
This PR also includes an endpoint /xlogin to make it easier to configure shiro to work with indirect authorization (eXternal login) like OAuth or OpenID through buji-pac4j.  
This PR expands the shiro security coverage to the entire application (/*).  It previously only covered /api/*. 

### What type of PR is it?
Improvement

### Todos
* [ ] - 

### What is the Jira issue?
[ZEPPELIN-2636] - https://issues.apache.org/jira/browse/ZEPPELIN-2636

Configure shiro.ini to use the various realms (Ini, Jdbc, Pam, etc).  Verify that the user and role lookup functionality used for notebook sharing still works.  I've been checking /api/security/userlist/<searchtext>

N/A

### Questions:
* Is there breaking changes for older versions?  using the stock shiro JdbcRealm, you won't be able to lookup users and roles... Transition to org.apache.zeppelin.realms.JdbcRealm.  


jongyoul:
Thanks for contribution. I suggest you'd better divide this PR into three or more different PRs as you mentioned in description. This PR includes many changes. I hope every PR changes as small as possible. Can you do that? And there's no test. If you add some tests, it makes your implementation concrete and easy to understand what you want to change.

PR #2407, opened 2017-06-13
[ZEPPELIN-1907] Shell Interpreter does not renew ticket on secure cluster

prabhjyotsingh:
### What is this PR for?
Kerberos ticket and renew lifetime are set to 1 hour. On accessing secure Hadoop from shell interpreter, it does kinit and returns result successfully but after 1 hour, the ticket gets expired and Hadoop list fails with below exception.

```
%sh
hadoop fs -ls /

17/01/05 09:29:45 WARN ipc.Client: Exception encountered while connecting to the server : 
javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]
	at com.sun.security.sasl.gsskerb.GssKrb5Client.evaluateChallenge(GssKrb5Client.java:211)
	at org.apache.hadoop.security.SaslRpcClient.saslConnect(SaslRpcClient.java:413)
	at org.apache.hadoop.ipc.Client$Connection.setupSaslConnection(Client.java:595)
	at org.apache.hadoop.ipc.Client$Connection.access$2000(Client.java:397)
	at org.apache.hadoop.ipc.Client$Connection$2.run(Client.java:762)
	at org.apache.hadoop.ipc.Client$Connection$2.run(Client.java:758)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1724)
	at org.apache.hadoop.ipc.Client$Connection.setupIOstreams(Client.java:757)
	at org.apache.hadoop.ipc.Client$Connection.access$3200(Client.java:397)
	at org.apache.hadoop.ipc.Client.getConnection(Client.java:1618)
	at org.apache.hadoop.ipc.Client.call(Client.java:1449)
	at org.apache.hadoop.ipc.Client.call(Client.java:1396)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Invoker.invoke(ProtobufRpcEngine.java:233)
	at com.sun.proxy.$Proxy10.getFileInfo(Unknown Source)
ls: Failed on local exception: java.io.IOException: javax.security.sasl.SaslException: GSS initiate failed [Caused by GSSException: No valid credentials provided (Mechanism level: Failed to find any Kerberos tgt)]; Host Details : local host is: "zeppelin1.hwxblr.com/10.0.1.57"; destination host is: "zeppelin1.hwxblr.com":8020; 
ExitValue: 1
```

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-1907](https://issues.apache.org/jira/browse/ZEPPELIN-1907)

On a Kerberos enabled cluster, run this paragraph
```
%sh
hdfs dfs -ls /user/zeppelin/
```
Wait for key-tab to expire (or run `kdestroy`), and re-run the same paragraph.

Before:
screen shot 2017-06-13 at 3 44 30 pm

After:
screen shot 2017-06-13 at 3 44 04 pm

### Questions:


shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -93,18 +95,31 @@ public InterpreterResult interpret(String cmd, InterpreterContext contextInterpr
...
-      Code code = Code.ERROR;
-      String message = outStream.toString();
-      if (exitValue == 143) {
-        code = Code.INCOMPLETE;
-        message += "Paragraph received a SIGTERM\n";
-        LOGGER.info("The paragraph " + contextInterpreter.getParagraphId() 
-          + " stopped executing: " + message);
+      if (isGSSInit &&
+          contextInterpreter.out != null &&
+          contextInterpreter.out.getCurrentOutput().toString().contains("GSS")) {
```
zjffdu: It is better to use a longer string in error message to replace 'GSS'

zjffdu:
@prabhjyotsingh I left one minor comment. Besides that I found 1 other issues.  

* Code duplication and inconsistency between `ShellInterpreter` and `ShellSecurityImpl`. Both of these 2 classes will create DefaultExecutor and do the real  shell execution. But it seems the command in `ShellSecurityImpl` can not be canceled, because it is not put in `executors`. It would be better to put the common logic together. 

But this issue is not related with this ticket, could be done in another ticket.

prabhjyotsingh:
@zjffdu Yes make sense I have created a JIRA (https://issues.apache.org/jira/browse/ZEPPELIN-2646) for it, I'll handle it in a different PR.

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
prabhjyotsingh: Do you mean to replace `GSS` with `GSSException` or even longer like `Caused by GSSException: No valid credentials provided`?

prabhjyotsingh:
@felixcheung I like the idea of renewing token periodically, do you think we should have both "if fail then renew" and "renew token periodically", or just renewing token periodically is sufficient?

prabhjyotsingh:
@felixcheung have pushed another change that will just renew token periodically. If this looks good, will implement the same for JDBC interpreter as well and fix the CI.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
```diff
@@ -486,4 +494,70 @@ public static RegisteredInterpreter findRegisteredInterpreterByClassName(String
...
+      throw new IllegalArgumentException("Invalid suffix: \"" + suffix + "\"");
+    }
+
+    return TimeUnit.MILLISECONDS.convert(val,
+        suffix != null ? Constants.TIME_SUFFIXES.get(suffix) : TimeUnit.MILLISECONDS);
+  }
+
+  protected abstract boolean runKerberosLogin();
+
+  protected ScheduledExecutorService startKerberosLoginThread() {
```
zjffdu: I don't think we should put the kinit stuff in `Interpeter.java` as it is not general for all interpreters.

prabhjyotsingh: I kind of thought about where to put this, but then finally have `protected abstract boolean runKerberosLogin();`  and depending upon what all interpreters need to override this function will have this.

Like in this particular case (of shell interpreter) in the open method this function (startKerberosLoginThread) gets called, and `runKerberosLogin` function talks about how to kinit.

But sure I'm open for any other thoughts.

felixcheung:
How about the addition of a secured/Kerberosize interpreter as a base class that individual interpreter can inherit from as applicable?

prabhjyotsingh:
@felixcheung  have implemented your feedback, will update the doc in a while.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.zeppelin.annotation.ZeppelinApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Interpreter wrapper for Kerberos initialization
+ *
+ * TODO: coming soon
+ * Please see,
+ * https://zeppelin.apache.org/docs/latest/development/writingzeppelininterpreter.html
```
felixcheung: we dont' need this?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * runKerberosLogin() method you need to implement that determine Zeppelin's behavior.
+ * startKerberosLoginThread() needs to be called inside the open() and
+ * shutdownExecutorService() inside close().
+ */
+public abstract class KerberosInterpreter extends Interpreter {
+
+  Integer kinitFailCount = 0;
```
felixcheung: perhaps a child class wants to check this? we should report this in the error if an operation fails because the token can't be refreshed in the background?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
```diff
@@ -0,0 +1,124 @@
...
+  public KerberosInterpreter(Properties property) {
+    super(property);
+  }
+
+  @ZeppelinApi
+  protected abstract boolean runKerberosLogin();
+
+  public String getKerberosRefreshInterval() {
+    if (System.getenv("KERBEROS_REFRESH_INTERVAL") == null) {
+      return "1d";
```
felixcheung: I think I recall each token has a expiry interval - perhaps we need to check or respect that. If the token expires in 4 hrs then perhaps we should refresh automatically before it expires?

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -143,4 +152,16 @@ public Scheduler getScheduler() {
     return null;
   }
 
+  @Override
+  protected boolean runKerberosLogin() {
+    try {
+      ShellSecurityImpl.createSecureConfiguration(getProperty(), shell);
+    } catch (Exception e) {
+      return false;
```
felixcheung: log error

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -60,12 +60,21 @@ public void open() {
...
 
   @Override
-  public void close() {}
+  public void close() {
+    shutdownExecutorService();
+
+    for (String executorKey : executors.keySet()) {
+      DefaultExecutor executor = executors.remove(executorKey);
+      if (executor != null) {
+        executor.getWatchdog().destroyProcess();
```
felixcheung: btw, do we need try catch around destroyProcess?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
prabhjyotsingh: Yes so usually this is controlled by https://web.mit.edu/kerberos/krb5-1.12/doc/admin/conf_files/krb5_conf.html with the parameter called `renew_lifetime`. But in some cases, it may expire after Kerberos ticket lifetime (1 day always due to a JDK Bug https://bugs.openjdk.java.net/browse/JDK-8044500). Hence, if this "KERBEROS_REFRESH_INTERVAL" is not set by user I'm trying to return "1d".

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
```diff
@@ -101,8 +97,11 @@ public Object call() throws Exception {
           scheduledExecutorService
               .schedule(this, getTimeAsMs(getKerberosRefreshInterval()), TimeUnit.MILLISECONDS);
         } else {
+          kinitFailCount++;
+          logger.info("runKerberosLogin failed for   " + kinitFailCount + "time");
```
felixcheung: nit: space before `"time"`?

prabhjyotsingh: Oh! yes, sure. Thanks.

prabhjyotsingh:
@felixcheung have handled your review comments, except https://github.com/apache/zeppelin/pull/2407#discussion_r122361683 as KERBEROS_REFRESH_INTERVAL is configurable and is documented [here](https://github.com/apache/zeppelin/pull/2407/files#diff-5a017ebfb7e890a2b475f9c8c7844fb0R71) as well.

prabhjyotsingh:
@felixcheung ping

docs/interpreter/shell.md
```diff
@@ -65,4 +65,15 @@ The following example demonstrates the basic usage of Shell in a Zeppelin notebo
 
 <img src="../assets/themes/zeppelin/img/docs-img/shell-example.png" />
 
-If you need further information about **Zeppelin Interpreter Setting** for using Shell interpreter, please read [What is interpreter setting?](../manual/interpreters.html#what-is-interpreter-setting) section first.
\ No newline at end of file
+If you need further information about **Zeppelin Interpreter Setting** for using Shell interpreter, please read [What is interpreter setting?](../manual/interpreters.html#what-is-interpreter-setting) section first.
+
+
+### Kerberos refresh interval
```
felixcheung: other section is using `##` 2 instead

docs/interpreter/shell.md
```diff
@@ -65,4 +65,15 @@ The following example demonstrates the basic usage of Shell in a Zeppelin notebo
...
+For changing the default behavior of when to renew Kerberos ticket following changes can be made in `conf/zeppelin-env.sh`.
+
+```
```
felixcheung: bash

prabhjyotsingh: Sorry, I did not get this, bash as in?

felixcheung:
` follow by bash to indicate the code block type
Sorry github really doesn't like backtick - so it's hard for me to write it out here. See other md for example.

prabhjyotsingh:
Thank you @felixcheung. I see it now, it makes a lot of difference.

prabhjyotsingh:

zjffdu:
@prabhjyotsingh @felixcheung I thought about this again. I don't have concern on this PR, but have concern on how user should interact with shell interpreter interpreter.  I think the shell interpreter should behave as close as bash terminal of linux. That means the user should call `kinit` if he want to run shell command in kerberized cluster just as he did in linux shell terminal. For now, the shell interpreter of zeppelin share the same keytab/principal for all users even when in isolated and impersonated mode which is a big security hole.

prabhjyotsingh:
IMO, this is just one less hassle. A user whoever configures Zeppelin with this configuration i.e. have a keytab/principal configured and enable "user impersonation", is aware that this keytab file is being share with various users, and if (s)he doesn't want this behaviour they can always fall back to the default option (which is not configuring any). 
The reason shell interpreter does this is to make this operation seamless (just like JDBC, Livy, etc.) and when we finally have user level interpreter setting (each user have their own interpreter setting), they can configure their private keytab/principal and it will continue to work.

Tagar:
Great addition. Thank you.
Btw, `renew_lifetime` although can be set in `krb5.conf`, it's maximum value is limited on Active Directory side if AD is used for Kerberos, look for example [Kerberos Policy](https://technet.microsoft.com/en-us/library/cc757692(v=ws.10).aspx#w2k3tr_sepol_accou_set_hpjo) .
Might be worth adding to the documentation?

> A user whoever configures Zeppelin with this configuration i.e. have a keytab/principal configured and enable "user impersonation", is aware that this keytab file is being shared with various users (which is not the default interpreter setting that we ship with Zeppelin), and if (s)he doesn't want this behaviour they can always fall back to the default option (which is not configuring any).

Would be great if Zeppelin would launch user's own Zeppelin interpreter processes under their own uid through setuid() call. So then keytabs could be locked down to be accessible to that one user. For example, after I LDAP-authenticated as "tagar" user, Zeppelin will drop down uid to tagar user and its keytab will have unix access bits set to `0600`. Makes sense? This probably should be part of another jira though.

prabhjyotsingh:
@Tagar sure, would you like to create a JIRA for the same?

Tagar:
@prabhjyotsingh,  created https://issues.apache.org/jira/browse/ZEPPELIN-2703
Let me know what do you gyus think?

PR #2413, opened 2017-06-15
[ZEPPELIN-2654] add roles to permissions of interpreter

tinkoff-dwh:
### What is this PR for?
Added the ability to specify roles (groups) which is available to use an interpreter (тow it is possible to set permissions to users only). The same solution as permissions of notes.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2654

1. Create note with Interpreter (for ex. spark)
2. Set permissions to Interpreter (spark) for `role1`
3. Log in by user with role `role1` and open note (from item 1)
4. Run paragraph
``` println("test")```
**Result == test**
6. Logout
7. Log in by user without role `role1` and open note (form item 1)
8.  Run paragraph
``` println("test")```
**Result == "...no permissions..."**

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
@@ -37,7 +37,7 @@ public class InterpreterOption {
 
   boolean isExistingProcess;
   boolean setPermission;
-  List<String> users;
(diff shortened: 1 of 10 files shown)


tinkoff-dwh:
Ready to review

Leemoonsoo:
LGTM and merge to master if no further discussions!

PR #2414, opened 2017-06-15
[ZEPPELIN-2657] Add group search filter option to LdapRealm

VipinRathor:
### What is this PR for?
Problem:
While performing LDAP authentication, current Shiro module does a group=* search while trying to get group-to-role mapping for any LDAP user. On a large LDAP directory, this is a serious problem which might render RolesByGroup feature not working as expected.

Fix:
Currently while doing LDAP authentication, there is no available option to limit the group search results to the only groups that user is interested in. This bug addresses the same and adds group search filter to Shiro configuration for LdapRealm which will allow user to define a search filter and limit the group search results.

### What type of PR is it?
Improvement

### What is the Jira issue?
* [ZEPPELIN-2657] https://issues.apache.org/jira/browse/ZEPPELIN-2657

1. Use org.apache.zeppelin.realm.LdapRealm as Shiro realm
2. In the shiro_ini configruation, define a group search filter like this:
ldapRealm.groupSearchFilter = (&(objectclass=groupofnames)(member={0}))
or 
ldapRealm.groupSearchFilter = (&(objectclass=groupofnames)(cn=zeppelin-users*))
3. Also define other LdapRealm parameters as necessary like rolesByGroup etc.
4. When an LDAP user, who is part of the group that matches filter above, logs in, then the roles are applied. If the LDAP user is not part these defined groups, then the roles are not applied.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/LdapRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/LdapRealm.java
@@ -101,6 +101,8 @@ import javax.naming.ldap.PagedResultsControl;
  * # ability set searchScopes subtree (default), one, base
  * ldapRealm.userSearchScope = subtree;
  * ldapRealm.groupSearchScope = subtree;
+ * ldapRealm.userSearchFilter = (&(objectclass=person)(sAMAccountName={0}))
+ * ldapRealm.groupSearchFilter = (&(objectclass=groupofnames)(member={0}))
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
@VipinRathor can you setup your Travis for this? https://zeppelin.apache.org/contribution/contributions.html#continuous-integration

VipinRathor:
@gss2002 & @weand,
Can you please help with review? Thanks.

prabhjyotsingh:
Tried on local LGTM.

prabhjyotsingh:

PR #2416, opened 2017-06-16
[ZEPPELIN-2647] Set admin user as owner when user create a notebook

yu74n:
### What is this PR for?
When a user creates a notebook, default users (an administrator can set default users in zeppelin-site.xml) are set as owner.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2647

1. Change Notebook workspace to private through whether ZEPPELIN_NOTEBOOK_PUBLIC = false or zeppelin.notebook.public = false.
2. Set default owners through ZEPPELIN_DEFAULT_OWNER = <comma separated username> or zeppelin.notebook.owner = <comma separated username>.
3. Create notebook.
4. Check if notebook's owner is set to both default users and the user who create the notebook.

### Questions:
* Is there breaking changes for older versions? Y/N

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -21,8 +21,11 @@ import java.io.File;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.XMLConfiguration;
@@ -41,6 +44,7 @@ public class ZeppelinConfiguration extends XMLConfiguration {
   private static final String ZEPPELIN_SITE_XML = "zeppelin-site.xml";
   private static final long serialVersionUID = 4749305895693848035L;
   private static final Logger LOG = LoggerFactory.getLogger(ZeppelinConfiguration.class);
+  private static final Pattern COMMA_DELIMITER = Pattern.compile("\\s*,\\s*");
 
   private static final String HELIUM_PACKAGE_DEFAULT_URL =
       "https://s3.amazonaws.com/helium-package/helium.json";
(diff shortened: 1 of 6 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -506,6 +510,11 @@ public String getWebsocketMaxTextMessageSize() {
     return getString(ConfVars.ZEPPELIN_WEBSOCKET_MAX_TEXT_MESSAGE_SIZE);
   }
 
+  public Set<String> getDefaultOwner() {
+    String values = getString(ConfVars.ZEPPELIN_DEFAULT_OWNER);
+    return new HashSet<>(Arrays.asList(COMMA_DELIMITER.split(values)));
```
felixcheung: owner should be the current user when ZEPPELIN_DEFAULT_OWNER is empty, right?

yu74n: yes, I added new unit test for checking if current user is set when ZEPPELIN_DEFAULT_OWNER is empty.

zeppelin-zengine/src/test/resources/test-zeppelin-site3.xml
```diff
@@ -0,0 +1,167 @@
...
+    <value>http://onehost:8080,http://otherhost.com,</value>
+    <description>Allowed sources for REST and WebSocket requests.</description>
+  </property>
+  <property>
+    <name>zeppelin.notebook.public</name>
+    <value>true</value>
+    <description>Make notebook public by default when created, private otherwise</description>
+  </property>
+  <property>
+    <name>zeppelin.notebook.owner</name>
```
1ambda: One suggestion. Instead of setting admin as one of owners, what about bypassing privilege check in case of admin? 

This approach doesn't modify existing notebooks and backward compatible.

yu74n: You mean check if the login user is included in this property value during creating notebook and reading, etc..?
For existing notebooks' permission, I expected that an administrator needs to directly edit json file which manages who is owner.

bhavintandel:
Have you guys planned to merge this request anytime soon ? If Yes, for which release have you planned

zeppelin-zengine/src/test/resources/test-zeppelin-site3.xml
1ambda: Hi, @yu74n 

Even if admins can modify the `note.json`, it's not easy to do that. So I suggested 

- not to set admin `role` as one of the owners since it's changing the content of `note.json`
- instead of, let's make admin **role** be the owner by passing admin role in auth logic.

1ambda:
@yu74n Could you resolve conflict and check my last comment? Thanks.

1ambda:
@bhavintandel I haven't checked this can be merged into the 0.7 branch. But if there is no problem to merge, and we can include this PR in 0.7.3

- even if there's conflict, we can create a separate PR.

yu74n:
@1ambda I created new PR for this issue. Could you check the new one?
I will close the PR.

yu74n:
Close this PR because Continued at #2585

PR #2417, opened 2017-06-16
Add  credential to spark interpreter

buptljy:
### What is this PR for?
If we want to access databases in a spark interpreter, we need to specify our own private username and password in a spark interpreter paragraph, which is not always safe due to visible authentifier information.

### What type of PR is it?
**[ Improvement ]**

### What is the Jira issue?
* [ZEPPELIN-2660](https://issues.apache.org/jira/browse/ZEPPELIN-2660)

* Create a new spark interpreter paragraph
* use userCredentials.get("your entity").get("username") to get username
* user userCredentials.get("your entity").get("password") to get password

### Questions:


spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1172,6 +1174,14 @@ public InterpreterResult interpret(String line, InterpreterContext context) {
       return new InterpreterResult(Code.ERROR, "Spark " + sparkVersion.toString()
           + " is not supported");
     }
+
+    ObjectMapper om = new ObjectMapper();
+    om.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+    Map<String, Map<String, String>> userCredentials = om.convertValue(
+            context.getAuthenticationInfo().getUserCredentials().getUserCredentials(), Map.class);
+    interpreter.intp().directBind("userCredentials",
```
felixcheung: hmm, this `userCredentials` will be accessible by *any* code running in the notebook... that doesn't feel very safe or secure

buptljy: But userCredentials is different if you are using different accounts, what can be accessible is just your own private username and password. 
[See here how to create credentials](https://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/security/datasource_authorization.html)

felixcheung: not necessary - an administrator could have set this up for a team and with this change everyone who has access to the notebook can now see the credentials/password?

PR #2420, opened 2017-06-19
[ZEPPELIN-2628] Serialization issue with RuntimeTypeAdapterFactory

necosta:
### What is this PR for?
Serialization bugfix in RuntimeTypeAdapterFactory
More info: https://github.com/google/gson/issues/712

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2628

See https://issues.apache.org/jira/browse/ZEPPELIN-2628

### Questions:


1ambda:
I saw this error in https://issues.apache.org/jira/browse/ZEPPELIN-2652 as well.

necosta:
Yes, @1ambda, 
The issue is on notebooks that contain "z.input("
On commit 7b585c7399f29, z.input was renamed to z.textbox, but loading old notebooks into 0.8 was not handled. 
The bug is on Note.java, addCloneParagraph, where we import forms without checking.
We are working on a fix that will be made available in a new pull request

FYI: @andreaTP

1ambda:
It seems that we have a backward compatibility problem.

PR #2428, opened 2017-06-22
[ZEPPELIN-1470] limiting results from jdbc

herval:
### What is this PR for?

One thing we tracked down is that if you issue a large query on a very large table, it will simply try to load all results (and then cap them on Zeppelin's side), which seems suboptimal (and will freeze the server). Setting this on the JDBC level seems to solve the problem.

### What type of PR is it?
Bug Fix

### Todos
* [x] Tests

- Create or use a table with a very large number of rows. In our tests, I simply created a:

```
createdb zeppelin_test

psql zeppelin_test

create table too_many_rows(n int)
```

And added 5m rows to it.

Making a paragraph like this will hang without setting a limit:
```
%zeppelin_test

select * from too_many_rows
```

### Questions:
No
No
No

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -681,7 +681,10 @@ public class JDBCInterpreter extends KerberosInterpreter {
       for (int i = 0; i < sqlArray.size(); i++) {
         String sqlToExecute = sqlArray.get(i);
         statement = connection.createStatement();
+
+        // fetch n+1 rows in order to indicate there's more rows available (for large selects)
         statement.setFetchSize(getMaxResult());
+        statement.setMaxRows(getMaxResult() + 1);
 
         if (statement == null) {
           return new InterpreterResult(Code.ERROR, "Prefix not found.");
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -681,6 +681,8 @@ public class JDBCInterpreter extends KerberosInterpreter {
       for (int i = 0; i < sqlArray.size(); i++) {
         String sqlToExecute = sqlArray.get(i);
         statement = connection.createStatement();
+        statement.setFetchSize(getMaxResult());
+
         if (statement == null) {
           return new InterpreterResult(Code.ERROR, "Prefix not found.");
         }


tinkoff-dwh:
user will not know that the table has more rows and the result is not complete. Maybe set fetchSize?

herval:
@tinkoff-dwh you're right, picked the wrong method - using setFetchSize now

herval:
I'm also not exactly sure how to write a test for this, as the JDBCInterpreterTest doesn't use anything in the line of mocks (and the outside behavior - limiting the number of returned rows - is already guaranteed by some filtering code. Any ideas @tinkoff-dwh ?

tinkoff-dwh:
it would be better getMaxResult() + 1

tinkoff-dwh:
I think it's impossible to test.

herval:
Why +1?

tinkoff-dwh:
yep, getMaxResult is correct. i thought the condition <= 
```while (displayRowCount < getMaxResult() && resultSet.next())```
and last next() will request the next part (fetchSize)

**Comment by herval** [comment]
It seems apache/zeppelin master (which is what this branch is based on) is failing the build - should I rebase it to 0.7 instead, or wait for master to be fixed? Please let me know what should I do next here (this is my first contribution to the project, pardon the ignorance)

**Comment by tinkoff-dwh** [comment]
https://travis-ci.org/herval/zeppelin/builds/245889847
try restart 3 and 5 job

**Comment by herval** [comment]
@tinkoff-dwh build #3 is failing in both my master & zeppelin master: https://travis-ci.org/apache/zeppelin - restarting didn't make it pass: https://travis-ci.org/apache/zeppelin/jobs/246113686

**Comment by tinkoff-dwh** [comment]
failed test wich load dependencies, it is possible too long or problems with the network. try restart again

File: jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java [hunk 1]
```diff
@@ -652,6 +652,8 @@ private InterpreterResult executeSql(String propertyKey, String sql,
       for (int i = 0; i < sqlArray.size(); i++) {
         String sqlToExecute = sqlArray.get(i);
         statement = connection.createStatement();
+        statement.setFetchSize(getMaxResult());
```
zjffdu: If `getMaxResult()` is used in statement directly, then it might not be necessary in method `getResults`

tinkoff-dwh: @zjffdu 
? it is count for size of fetch, not limit

zjffdu: Sorry, my misunderstanding of this API. Please ignore it.

tinkoff-dwh:
LGTM

herval:
Build's passing: https://travis-ci.org/herval/zeppelin/builds/247193502

Let me know if this is mergeable :-) @tinkoff-dwh

tinkoff-dwh:
@herval 
reopen pull request. and wait another reviewers or committers)

herval:
hmm.. It actually seems that setFetchSize does not limit the query execution (so it takes forever & hangs Zeppelin)

Figuring out if there's a way to show results are truncated + use setMaxRows

herval:
screenshot 2017-06-29 12 37 41

masyan:
we already have a check to limit. setMaxRows duplicate it

tinkoff-dwh:
we already have a check to limit. setMaxRows duplicate it

herval:
@tinkoff-dwh  it's not duplicate. The bug here is that if you don't do `statement.setMaxRows`, some drivers (eg postgres) will try to load everything when you issue a `select *`. Using `setFetchSize` won't affect that (I tested it). So the solution I found is to set the limit to (fetch size + 1). There's a function that will prune the results *after they're back from JDBC* (`getResults`) - I'm guessing we could remove that, if all it did was checking that limit, but don't think it's worth pursuing at this time.

TLDR - the current limit check is only applied _after_ you get results. If you query a big enough table, you'll never get the results (and will kill Zeppelin in the process)

felixcheung:
got it. I agree it's generally better to restrict the job at the source
perhaps you can use the existing settings but apply it at the JDBC instead? it would be easier to switch existing user over the new behavior this way - actually, let's see what other thinks about this approach.

herval:
I am using the original setting (getMaxResults). I had to set it in two calls because setFetchSize will determine that there is a “next” page, but if I only do “setMaxResults”, it will not display the truncation warning (as it translates to a “limit”)

herval:
any additional thoughts on this? Is it mergeable? @khalidhuseynov @tinkoff-dwh @felixcheung

herval:
Rebased to latest master - please advise

herval:
ping - please advise

astroshim:
LGTM~
I think this PR can be merged.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -681,6 +681,11 @@ private InterpreterResult executeSql(String propertyKey, String sql,
       for (int i = 0; i < sqlArray.size(); i++) {
         String sqlToExecute = sqlArray.get(i);
         statement = connection.createStatement();
+
+        // fetch n+1 rows in order to indicate there's more rows available (for large selects)
+        statement.setFetchSize(getMaxResult());
+        statement.setMaxRows(getMaxResult() + 1);
```
felixcheung: so hopefully getMaxResult() won't return Integer.MAX_VALUE? because +1 will overflow?

herval: Unless someone specifies a fetch size of two billion rows, this should be fine (the UI would break and Zeppelin would run OOM and the world would melt w/ that amount of rows anyway, so I wouldn't particularly worry about that scenario? :-))

felixcheung: haha, yes

felixcheung:
merging if no more comment

PR #2434, opened 2017-06-24
ZEPPELIN-2687: Show code in description section of Spark UI

karuppayya:
### What is this PR for?
Show code in description section of Spark UI

### What type of PR is it?
 Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
ZEPPELIN-2687

Run scala, pyspark and sql snippet(should result in a spark job), and check Spark Web UI

Before:
screen shot 2017-06-24 at 12 26 02

After:
Collapsed(short desc):
screen shot 2017-06-24 at 12 12 38

Expanded(Long desc):
screen shot 2017-06-24 at 12 12 43

### Questions:

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -46,6 +46,7 @@ import org.apache.spark.scheduler.SparkListenerJobStart;
 import org.apache.spark.sql.SQLContext;
 import org.apache.spark.ui.SparkUI;
 import org.apache.spark.ui.jobs.JobProgressListener;
+import org.apache.spark.util.CallSite;
 import org.apache.zeppelin.interpreter.*;
 import org.apache.zeppelin.interpreter.InterpreterResult.Code;
 import org.apache.zeppelin.interpreter.util.InterpreterOutputStream;
@@ -1183,6 +1184,9 @@ public class SparkInterpreter extends Interpreter {
   public InterpreterResult interpret(String[] lines, InterpreterContext context) {
     synchronized (this) {
       z.setGui(context.getGui());
+      String paraText = context.getParagraphText();
+      sc.setCallSite(new CallSite(Utils.getJobShortText(paraText),
+          Utils.getJobLongText(paraText)));
       String jobDesc = "Started by: " + Utils.getUserName(context.getAuthenticationInfo());
       sc.setJobGroup(Utils.buildJobGroupId(context), jobDesc, false);
       InterpreterResult r = interpretInput(lines, context);
--- a/spark/src/main/java/org/apache/zeppelin/spark/Utils.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/Utils.java
@@ -135,4 +135,33 @@ class Utils {
     }
     return uName;
   }
+
+  public static String getJobShortText(String paraText) {
+    paraText = sanitizeText(paraText);
+    String text = paraText;
+    if (paraText != null && !paraText.isEmpty()) {
+      String[] splits = paraText.split(System.lineSeparator(), 2);
+      if (splits.length > 0) {
+        text = splits[0];
+      }
+    }
+    return text;
+  }
+
+  public static String getJobLongText(String paraText) {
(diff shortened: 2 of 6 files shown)


spark/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -135,4 +135,33 @@ public static String getUserName(AuthenticationInfo info) {
     }
     return uName;
   }
+
+  public static String getJobShortText(String paraText) {
```
zjffdu: Could you add unit test for these util methods ?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1183,6 +1184,9 @@ public InterpreterResult interpret(String line, InterpreterContext context) {
   public InterpreterResult interpret(String[] lines, InterpreterContext context) {
     synchronized (this) {
       z.setGui(context.getGui());
+      String paraText = context.getParagraphText();
+      sc.setCallSite(new CallSite(Utils.getJobShortText(paraText),
```
zjffdu: `SparkRInterpreter` is missing here

zjffdu:
@karuppayya Thanks for this PR. It is really helpful. Besides this, with more and more features added we have more and more duplicated code in `PySparkInterpreter`, `SparkInterpreter`, `SparkSQLInterpreter` and `SparkRInterpreter`.  It is better to have a base class to handle all these common things. Anyway, we could do this in another PR. But I just feel we need to to asap. Because zeppelin code base become too large than it could be.

karuppayya:
@zjffdu +1 on the refactoring.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
karuppayya: @zjffdu I didnot find api to `setLocalProperty()` in R.    
 Any idea/pointer on how to implement his with sparkR?

zjffdu: hmm, then it is fine not do it for SparkR for now. Could you create a spark ticket for that ?

karuppayya: filed https://issues.apache.org/jira/browse/SPARK-21208

spark/src/main/java/org/apache/zeppelin/spark/Utils.java
karuppayya: done

karuppayya:
On restrating the failed job(selenium tests), it succeeded
https://travis-ci.org/karuppayya/zeppelin/builds/246794338
Ready for reviewing

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
felixcheung: ah nice - so this was where it came from ;)

karuppayya:
@felixcheung 

1. Regarding the security exploitation
Spark does escapes any HTML, except for a anchor element with relative links
https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/ui/UIUtils.scala#L439
2. There can be multiple jobs per notebook paragraph
When a sceond paragrph with different code is executed, the corresponding paragraph code shows up in the description section

karuppayya:
@felixcheung any suggestion on the length of text(for the short description).   
Also in case of scala, python code, when the first code line is a import , should we skip it and show the first line of actual code as short description, instead of the import statement?.

Currently, we show the entire code in the long description. This can potentially bloat the evnt logs. Should we prune it, to say last n lines?.
@felixcheung @Leemoonsoo @jongyoul

felixcheung:
what is the goal of this? if the goal is identification, then either the last line (or last few lines) and/or with paragraph id should be sufficient.
I believe the pig interpreter has a similar approach, right @zjffdu ?

zjffdu:
Yes, we need to truncate that. @karuppayya Could you add some comment to explain what is the jobShortText and jobLongText.

PR #2439, opened 2017-06-26
[ZEPPELIN-2680] allow opening notebook as a reader

herval:
### What is this PR for?

We have a few notebooks that have a list of writers & owners, but no restricted readers list.
When you try to open these notebooks as a reader AND they have a chart the app sends a COMMIT_PARAGRAPH for each, which causes a write permission popup to show ("Insufficient privileges to write note").
I tracked down the issue to the result.controller.js (commitParagraphResult) - propagating the "viewOnly" state fixes that behavior.

### What type of PR is it?
Bug Fix

- Create a notebook. Add a paragraph with a graph. Set the "owner" of the notebook to something other than yourself. You won't be able to load the notebook without the security popup anymore (unless logging in as the owner)

### Questions:
No
No
No

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -637,7 +637,7 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
     let newParagraphConfig = angular.copy(paragraph.config)
     newParagraphConfig.results = newParagraphConfig.results || []
     newParagraphConfig.results[resultIndex] = config
-    if ($scope.revisionView === true || $scope.viewOnly) {
+    if ($scope.revisionView || $scope.viewOnly || $scope.reportMode) {
       // local update without commit
       updateData({
         type: $scope.type,
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -637,7 +637,7 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
     let newParagraphConfig = angular.copy(paragraph.config)
     newParagraphConfig.results = newParagraphConfig.results || []
     newParagraphConfig.results[resultIndex] = config
-    if ($scope.revisionView === true) {
+    if ($scope.revisionView === true || $scope.viewOnly) {
       // local update without commit
       updateData({
         type: $scope.type,
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/integration/AuthenticationIT.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/integration/AuthenticationIT.java
@@ -249,9 +249,16 @@ public class AuthenticationIT extends AbstractZeppelinIT {
               "//div[contains(.,'Insufficient privileges')]"));
       collector.checkThat("Check is user has permission to view this note", 0,
           CoreMatchers.equalTo(privilegesModal.size()));
-      deleteTestNotebook(driver);
-      authenticationIT.logoutUser("finance2");
 
+      // delete option not available for non-writer/owner
+      collector.checkThat(
+        driver.findElement(
+          By.xpath(".//*[@id='main']//button[@ng-click='moveNoteToTrash(note.id)']")
+        ).isDisplayed(),
+        CoreMatchers.is(false)
+      );
+
+      authenticationIT.logoutUser("finance2");
 
     } catch (Exception e) {
       handleException("Exception in ParagraphActionsIT while testGroupPermission ", e);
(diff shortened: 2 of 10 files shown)


herval:
Build is green: https://travis-ci.org/herval/zeppelin/builds/247345573

herval:
any thoughts on this?

soralee:
Hi, @herval. Thanks for contribution and sorry for late response.
Let me take a look this one.

soralee:
I have a question quickly.
For reproduce, I just make sure test flow clearly. 
1. login as user1
2. create a note
3. add a paragraph
4. set `owners` and `writers` field set user1
5. set `readers` field sets user2
6. logout user1 and login user2.
7. back to note and check whether popup is appeared or not.

Is it correct?

herval:
@soralee thanks for taking the time :-) there's one more step to your list:

1. login as user1
2. create a note
3. add a paragraph
**4. change the visualization to a chart (instead of table)**
5. set owners and writers field set user1
6. set readers field sets user2
7. logout user1 and login user2.
8. back to note and check whether popup is appeared or not.

soralee:
Oh, Thanks for quick answer and check! 
Let me test as your comment and come back here!

soralee:
I could reproduce this problem on master branch but, I could not see to resolve on your branch.
It has same error.
Could you test it out on your branch?

herval:
Hmm I’ll try again.

Btw, did you `mvn clean` after switching branches? I think changes on the js/html files don’t get picked up unless you do that...

soralee:
Yes, I did. I run `mvn clean package -DskipTests`
I'll check again.

soralee:
I just checked again, but there is still a same problem. could you check this?

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
```diff
@@ -631,7 +631,7 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
     let newParagraphConfig = angular.copy(paragraph.config)
     newParagraphConfig.results = newParagraphConfig.results || []
     newParagraphConfig.results[resultIndex] = config
-    if ($scope.revisionView === true) {
+    if ($scope.revisionView === true || $scope.viewOnly) {
```
soralee: I checked the `$scope.viewOnly` value in this line but it had `false` value. And I checked `true` value when note view mode is `report mode`. 

Thus, I'm not sure that I guess if you want to resolve this problem which is for working `readers` role on the every mode, you refer the following line.
https://github.com/apache/zeppelin/blob/a39ce0141006d5762058b6f7d0cf1ba349d869ac/zeppelin-web/src/app/notebook/notebook.controller.js#L440

herval: whoops, you're right :-) I missed that from our internal commit. Fixing this branch, will test locally and get back to you. Thanks for the patience!

herval:
Ok, retested now and it's working as expected.

If you need to login (using anonymous mode support), you may need this PR first: https://github.com/apache/zeppelin/pull/2440

soralee:
Good!
Let me test it out agian!

soralee:
Tested and LGTM! 
It looks cool that when a `writers` logins a note, it shows report mode.

herval:
all tests passing now

khalidhuseynov:
thanks for the improvement, this scenario needs to be fixed indeed. after trying the pr, i noticed couple of things
 * when following scenario and going into note as user2, my browser got stuck with the continuous requests to the helium endpoint, anyone experienced it?
 * the note opens in report mode, although it say in the top-right corner `default` mode. that seems a bit inconsistent

herval:
hmm I'll see if I can somehow split "report mode" and "view only mode".

I didnt experience the Helium issue, personally

herval:
@soralee @khalidhuseynov would you or anyone else like to give this a new try?

herval:
Hi folks,

Please let me know if I should close this. Tks!

weand:
@herval would you mind merging again on latest master?

@soralee @zjffdu as we face the same issue, and we are using helium, I will also test it.

weand:
@herval: I merged latest master: [https://github.com/weand/zeppelin/tree/ZEPPELIN-2680](https://github.com/weand/zeppelin/tree/ZEPPELIN-2680)

But I also face continuous requests against helium endpoint (without having any helium configured at all). Same behavior like @khalidhuseynov. I recognized this only by having a look at the Network Communication in the browsers developer tools.

I'm not into the angular stuff so much. Could you have a look?

bern80:
Dear All,
Is this _zeppelin-web/src/app/notebook/paragraph/result/result.controller.js_ script already fixed on master branch?

Thanks,
R.

redsk:
Dear all,

I'm also interested in this; the problem does not seem to be solved in master.

PR #2442, opened 2017-06-27
[ZEPPELIN-2679] JDBC. precode for session

tinkoff-dwh:
### What is this PR for?
We have a precode for interpreter (all interpreters) which executes once after start interpreter. With it we can create shared code that will be available in the paragraph, create tables in the database and etc. This precode for jdbc only which executes before run paragraph in same session (sql). In this precode we can declare session variables that will be available when you run paragraph (any sql which will be executed before the paragraph). For exabple `set search_path='public,#{user}'`

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2679

(jdbc interpreter, postgres)
1. set property **default.sessionPrecode** `set search_path='test_session_precode'`
2. run paragraph `show search_path`

### Questions:


tinkoff-dwh:
@felixcheung 
session == database session. added documentation

zjffdu:
@tinkoff-dwh What do you mean `database session` ? From the code, `precode` is executed  per statement.  And what is the usage scenario for this ? Could you give an example ? Thanks

tinkoff-dwh:
@zjffdu 
yep, per statement. 
for exp.
There is jdbc interpreter per user  (postgres).
There are many users and each user has scheme to work. 
1. set search_path='#{user}'   (default.sessionPrecode)
2. run paragraph `select ...`
the user will work with the correct scheme

or set another variables and values will be available to code in the paragraph (SET TIME ZONE, custom variables, etc...). User == analitic (usually) and he doesn't have to write this in the code of paragraph

zjffdu:
I am not sure whether this is a good idea to put custom initialization code in interpreter setting. Because I think one goal of notebook is to allow user to reproduce it easily without change note. So any non-trivial thing should be put to note itself. Otherwise user don't know how to reproduce the note result in another zeppelin instance.

tinkoff-dwh:
@zjffdu 
another instance? user == analitic. It is real task from production. The user does't configure interpreter, it is work of administrator.

zjffdu:
I am thinking that every interpreter could have one special interpreter for initializing interpreter setting and custom enviroment requirement for his note. And do the initialization at the first paragraph. e.g.

```
%spark.init

set property_1=value_1
set property_2=value_2
```
So that the note could be shared to other people, and other people can reproduce the note result easily. My point is that note should be self-contained thing that could be reproduced when it is shared to other people


However it is still very immature idea.

tinkoff-dwh:
it is JDBC interpretator, set variables in first paragraph  will not work because we have connection pool and the paragraphs will be executed in different statemnets. What problem with shared notes. Фny user can execute and everyone will have their `precode`.  `set search_path='#{user}`  #{user} will be replaced with the username from the context

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

dwhsys:
Can someone review please?

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

mebelousov:
@zjffdu @felixcheung 

Every day our users run about 1000 queries in the Greenplum (open source Postgres-based DB).
There is great need for binding the query to the user and/or to the note.

The precode `set application_name=#{noteId};` allows quickly find the note.
This precode must be set by Zeppelin administrator, business users will not write any precode in each note for each jdbc-interpreter.

I believe the precode in JDBC interpreter is necessary feature for entreprise users.

zjffdu:
Mostly LGTM, one concern is about the name `session` which might confuse users.  How about changing it to `statement` ? 
Besides, I am curious to know how do you use jdbc interpreter. Which mode do you use ? (Shared/Scoped/Isolated) . Because jdbc interpreter use connection pool, that's why I ask this question, just want to know how do you schedule the connection across users

mebelousov:
@zjffdu, you are right, sessionPrecode is renamed to statementPrecode.

Since 0.8.0 I will recommend to run JDBC interpreter per note in isolated mode due to Interpreter Lifecycle Manager: the table results accumulate in ResourcePool and JVMs grow in size, but ILM will drop non-active JVMs.

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
```diff
@@ -523,6 +525,65 @@ public void testPrecodeWithAnotherPrefix() throws SQLException, IOException {
     assertEquals("ID\n2\n", interpreterResult.message().get(0).getData());
   }
 
+  @Test
+  public void testSessionPrecode() throws SQLException, IOException {
```
zjffdu: session-> statement

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
```diff
@@ -523,6 +525,65 @@ public void testPrecodeWithAnotherPrefix() throws SQLException, IOException {
...
+
+    InterpreterResult interpreterResult = jdbcInterpreter.interpret(sqlQuery, interpreterContext);
+
+    assertEquals(InterpreterResult.Code.SUCCESS, interpreterResult.code());
+    assertEquals(InterpreterResult.Type.TABLE, interpreterResult.message().get(0).getType());
+    assertEquals("@V\nsession\n", interpreterResult.message().get(0).getData());
+  }
+
+  @Test
+  public void testIncorrectSessionPrecode() throws SQLException, IOException {
```
zjffdu: dito

jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
```diff
@@ -523,6 +525,65 @@ public void testPrecodeWithAnotherPrefix() throws SQLException, IOException {
...
+    String sqlQuery = "select 1";
+
+    InterpreterResult interpreterResult = jdbcInterpreter.interpret(sqlQuery, interpreterContext);
+
+    assertEquals(InterpreterResult.Code.ERROR, interpreterResult.code());
+    assertEquals(InterpreterResult.Type.TEXT, interpreterResult.message().get(0).getType());
+  }
+
+  @Test
+  public void testSessionPrecodeWithAnotherPrefix() throws SQLException, IOException {
```
zjffdu: dito

zjffdu:
@tinkoff-dwh I just realised that zeppelin already support hooks for interpreter. I think you don't need to do it specifically for jdbc interpreter, instead we could introduce one general property like `zeppelin.interpreter.pre_code` which could execute pre code before running the paragraph text. What do you think ?

mebelousov:
@zjffdu
The hooks as pre-code were discussed [previously ](https://github.com/apache/zeppelin/pull/2096#issuecomment-284220634)
We want precode execute before pre-hook for easy finding the note. Meanwhile the hooks doesn't realized in JDBC interpreter.

The JDBC interpreter allow to split query and precode must run before each statement in paragraph. Therefore the single way for precode for all interpreters looks unrealizable.

The rename to `zeppelin.interpreter.precode` is good idea. Do this?

zjffdu:
Sorry @mebelousov  I don't get why the general hooks could not satisfy you. The hook can run for each paragraph. Could you explain a little bit more ? Maybe an example would be very helpful.

mebelousov:
@zjffdu
hooks are user defined code.
We want to have opportunity to log the user sessions, to bind queries to notes. It can be possible only by force execution of special query before user queries, that is in precode.

kapustor:
+1
Must-have feature for enterprise usage. Pre-code may be a resource-management code, or a security code, etc. - users must not have possibility to change it, only admins.

anermakov:
+1 Must have

Aspekt112:
+1

Sanikadze:
+1

zjffdu:
@tinkoff-dwh Could you rebase the PR ?

zjffdu:
@tinkoff-dwh @mebelousov Could you retrigger the build ? Travis seems fail

mebelousov:
@zjffdu I have retriggered several times. Travis fails with the same error on all jobs.

zjffdu:
@mebelousov Could you clean the travis cache and for a dummy force push to trigger the build ?

zjffdu:
ping @tinkoff-dwh @mebelousov

mebelousov:
Cache is clean now. I try to retrigger from both accounts. The result is the sameю

zjffdu:
Could you create another PR to trigger the build ?

PR #2443, opened 2017-06-27
[ZEPPELIN-2698] Apply KerberosInterpreter to JDBCInterpreter

prabhjyotsingh:
### What is this PR for?
This is to apply new KerberosInterpreter mechanism to JDBCInterpreter for relogin from keytab, instead of on fail relogin.

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
* [ZEPPELIN-2698](https://issues.apache.org/jira/browse/ZEPPELIN-2698)

- zeppelin.jdbc.auth.type = KERBEROS
- zeppelin.jdbc.principal = principal value
- zeppelin.jdbc.keytab.location = keytab location

Now try and run any of hive's query (say `show tables`) it should return with valid results.
Again, wait for this kerberos ticket to expire (usually its 24hrs) then try the above again, and it should work.

### Questions:


prabhjyotsingh:
@felixcheung @zjffdu can you help review this as well?

zjffdu:
This PR LGTM. Just one comment on the base class `KerberosInterpreter`. I think `startKerberosLoginThread` and `shutdownExecutorService` should be called in `KerberosInterpreter`'s open and close method rather than ask the sub class to call them. The sub class should only implement 2 abstract classes:
* isKerboseEnabled
* runKerberosLogin

Which means

```
public void open() {
   if (isKerboseEnabled()) {
      startKerberosLoginThread()
   }
}

public void close() {
   if (isKerboseEnabled()) {
      shutdownExecutorService()
   }
}
```

prabhjyotsingh:
@zjffdu sure, that does make sense.

prabhjyotsingh:
@zjffdu I've made the suggested changes. Let me know if I'm missing something else as well.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
```diff
@@ -48,23 +48,38 @@ public KerberosInterpreter(Properties property) {
...
+  }
+
+  public void close() {
+    if (isKerboseEnabled()) {
+      shutdownExecutorService();
+    }
+  }
+
+  private String getKerberosRefreshInterval() {
     if (System.getenv("KERBEROS_REFRESH_INTERVAL") == null) {
```
zjffdu: I think we need to add javadoc for `KerberosInterpreter.java` to explain the internal mechnism of it.  So basically I believe user need to do 3 things to extend `KerberosInterpreter`.

* implement `runKerberosLogin `
* implement `isKerboseEnabled `
* define `KERBEROS_REFRESH_INTERVAL` in interpreter setting. Maybe we could also add abstract method like `getKerborseRefreshInterval` to enforce its sub class to implement it.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -147,12 +148,31 @@ public JDBCInterpreter(Properties property) {
     maxLineResults = MAX_LINE_DEFAULT;
   }
 
+  @Override
+  protected boolean runKerberosLogin() {
+    try {
+      UserGroupInformation.AuthenticationMethod authType = JDBCSecurityImpl.getAuthtype(property);
+      if (authType.equals(KERBEROS)) {
```
zjffdu: authType is not needed to check as `runKerberosLogin` would be only called when kerberos is enabled.

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -57,17 +57,14 @@ public ShellInterpreter(Properties property) {
 
   @Override
```
zjffdu: I think we can move `createSecureConfiguration` from `ShellSecurityImpl` to `ShellInterpreter`, then `ShellSecurityImpl` could be deleted

prabhjyotsingh: Sure, done.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: Yes, make sense, removed unnecessary if condition.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java
prabhjyotsingh: Sure, have added isKerboseEnabled in javadoc. 
For `KERBEROS_REFRESH_INTERVAL` I think all interpreters can share a common its value defined in `zeppelin-env.sh` (https://github.com/apache/zeppelin/blob/master/conf/zeppelin-env.sh.template#L58-L59) instead of having it at interpreter level, and increasing the complexity (number of lines) of interpreter config.

zjffdu:
LGTM

prabhjyotsingh:
@zjffdu Thank you for the review. CI is green, will merge if no more discussion.

PR #2448, opened 2017-06-28
[ZEPPELIN-2702] save notes in reader-friendly format with zpln extension for VFS and Git repo

khalidhuseynov:
### What is this PR for?
first task in converting note files into human readable and Zeppelin own format

### What type of PR is it?
Improvement | Feature

### Todos
* [x] - initial flow
* [x] - rename through save
* [x] - tests for vfs

### What is the Jira issue?
[ZEPPELIN-2702](https://issues.apache.org/jira/browse/ZEPPELIN-2702)

1. go create new notes or rename title of old ones
2. see the notebook directory, under [NOTE_ID]/title.zpln

### Questions:
* Does this needs documentation? separately


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
```diff
@@ -0,0 +1,80 @@
...
+    fileName = Maps.newConcurrentMap();
+    forceRename = false;
+  }
+  
+  public static FileInfo createInstance() {
+    return new FileInfo();
+  }
+  
+  public String getFileName(String storageName) {
+    if (!fileName.containsKey(storageName)) {
```
zjffdu: different filename for different storage ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -825,6 +837,26 @@ public void setInfo(Map<String, Object> info) {
...
   }
 
+  public String getVersion() {
+    String version = StringUtils.EMPTY;
+    if (info.get("version") == null) {
+      return version;
+    }
+    try {
+      version = (String) info.get("version");
+    } catch (ClassCastException e) {
```
felixcheung: should it also check the content of the version string?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
khalidhuseynov: let me look into having one file name side of it

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/Util.java
```diff
@@ -29,6 +30,9 @@
   private static final String PROJECT_PROPERTIES_VERSION_KEY = "version";
   private static final String GIT_PROPERTIES_COMMIT_ID_KEY = "git.commit.id.abbrev";
   private static final String GIT_PROPERTIES_COMMIT_TS_KEY = "git.commit.time";
+  private static final String ZEPPELIN_NOTE_FILE_EXTENSION = "zpln";
+  private static final int MAX_ZEPPELIN_NOTE_FILE_LENGTH = 127;
```
felixcheung: why is this 127?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
khalidhuseynov: the problem with having single file name is that during rename of file (title changes), it's possible that file was renamed in one storage, but update haven't been applied in another one. so it would show inconsistent state of filename in that case

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
khalidhuseynov: actually I may remove `version` info in this PR since haven't been used actively here. instead i'll add it as separate pr with more details

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/Util.java
khalidhuseynov: after short review of max filename lengths on various filesystems 255 seemed to be correct number of chars. However some mobile platforms (e.g. android) have problems with having more than 127 chars, so i thought it would be safer. any other opinion on this point is welcome

felixcheung: I see. this 127 does it generally include extension though? I ask since the calculate below is adding extension length on top of this 127

my guess is we could probably limit to lower than 127 (say 110 or 100) - most people probably don't care.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
zjffdu: Do you mean `NotebookRepoSync` ? I think file renaming is a very light operation which doesn't need to be async.

zeppelin-zengine/src/main/java/org/apache/zeppelin/util/Util.java
khalidhuseynov: Yeah i also was thinking to lower it, will modify to 100  then

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
khalidhuseynov: no, actually `NotebookReopSync` is unaware of it and each storage layer needs to handle saving into file/renaming. 

So here's how it's handled in this PR, let's say `noteA` is read from local `GitNotebookRepo` with name `noteA.zpln` and it's also saved in `S3NotebookRepo` under same name via sync. when renaming it, those two storages will refer to `note` object to locate the filename and rename it. in case we have one filename, first storage (git) will rename file locally and update filename in `note` object, then when second storage (s3) updating it will refer to `note` object for old filename (in order to rename into new one), however it will contain new filename overwritten by first storage. that's why i added filename per storage so that they don't overwrite each other. In this PR i tried not to change method signatures, and want to handle it storage by storage. once we have this functionality in all storages, we may change some method signatures or put more load into `NotebookRepoSync`, for now i didn't add much functionality in that base class. let me know if it brings some clarity

zjffdu: But isn't the old file name `note_id.json` ? Why we need to remember the old file name in the FileInfo ?

khalidhuseynov: right first time when renaming old file name is note.json, but here's couple of reasons to keep current filename:
 *  if user changes title of notebook further, it needs to be renamed again to reflect uptodate status.
 * having the old filename would help to handle the case of multiple notes under same folder
 * it would further help linking folder structure shown in Zeppelin with physical folder structure under notebooks folder

zjffdu: Sorry, but it is still weird to me that `Note` need to keep the note file name of each storage by itself. It should be the responsibility of `NotebookRepo`. How about add one interface `rename(old_name, new_name)` in `NotebookRepo` ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java
```diff
@@ -238,14 +273,51 @@ public synchronized void save(Note note, AuthenticationInfo subject) throws IOEx
...
-    FileObject noteJson = noteDir.resolveFile(".note.json", NameScope.CHILD);
+    FileObject noteJson = noteDir.resolveFile(".note.zpln", NameScope.CHILD);
+    
     // false means not appending. creates file if not exists
     OutputStream out = noteJson.getContent().getOutputStream(false);
     out.write(json.getBytes(conf.getString(ConfVars.ZEPPELIN_ENCODING)));
     out.close();
-    noteJson.moveTo(noteDir.resolveFile("note.json", NameScope.CHILD));
+    
+    setFileName(note, noteJson, noteDir);
```
zjffdu: The code before this line has already write file `.note.zpln`, but `setFileName` here again do the move operation. The logic is a little confusing to me.

khalidhuseynov: `.note.zpln` (noteJson) is temporary file where it's written initially, and further it's moved either into `note.json` if keeping old format, or `title.zpln` if saving in new format

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileInfo.java
khalidhuseynov: who would keep and pass `old_name` in that case?

zjffdu: Note will trigger that operation when each time note title is changed. IMHO, where to store the note and in what kind of format should be transparent to Note. Note just need to call `NotebookRepo` to save or update via its interface. It is `NotebookRepo`'s responsibility to map note to some object in the storage system (file for `VFSNotebookRepo`). 

Another thing I am thinking is that whether the rename operation should just happen in `getNote` instead of `saveNote`.  So basically there would be 3 apis in `NotebookRepo`.  
 
* getNote
* renameNote
* saveNote

The renaming operation could happen in `getNote` & `renameNote`. The renaming in `getNote` is for old/new version conversion, the `renameNote` is for title update (Actually it depends on the implementation, some `NotebookRepo` may not use title as file name or they may not use file to store, such as hbase)


My point is that we should make the boundary between `Note` and `NotebookRepo` clean. The interaction between these 2 things only through interfaces. Anyway this may involve some refactor, may not be related to this PR. But it would benefit zeppelin for long term. 
What is your opinion?

khalidhuseynov: in that case all notes would converted into new format once loaded on start, which is okay too. yeah agree completely on keeping the boundary clean, and myself was thinking on refactoring `NotebookRepo` from base  class in longer term. for this one i'll look into adding new `rename` method into `NotebookRepo`, since current approach was trying to introduce minimal changes

zjffdu: Thanks @khalidhuseynov Looking forward your refactoring in `NotebookRepo` component in another PR. 😄  BTW, for the file name conversion, What is your plan for all the NotebookRepo implementation. Because it may affect the zeppelin upgrade, so it is better to it all in one major release. Maybe complete it in 0.8

khalidhuseynov: yeah right, it needs to be included in 0.8, and the plan is to implement it for S3 and Azure which require usage of filename/extension. other storages as far as i know do not store in filesystem using names

khalidhuseynov:
@zjffdu addressed the feedback on adding new `rename` method and having only one fileInfo, please help review

khalidhuseynov:
rebased from master, /cc @1ambda @Leemoonsoo please help review and test as well

1ambda:
Tested locally and works 3 in 2 cases (creating, renaming)

I have a question. Subclasses of the`NotebookRepo` have auto-generated method `rename`. That  means, If we want to rename for that repos, should we implement those methods later?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/AzureNotebookRepo.java
```diff
@@ -244,4 +245,11 @@ public Note setNoteRevision(String noteId, String revId, AuthenticationInfo subj
     // Auto-generated method stub
     return null;
   }
+
+  @Override
+  public FileInfo rename(FileInfo oldFile, FileInfo newFile, AuthenticationInfo subject)
```
zjffdu: Sorry, I don't think it is time to add rename method interface. Because rename should be called when note title is changed. But for now I think title renaming should not trigger any operation in NotebookRepo. So I think it is better to do the renaming in save method.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -754,6 +766,20 @@ public void persist(AuthenticationInfo subject) throws IOException {
...
+    stopDelayedPersistTimer();
+    snapshotAngularObjectRegistry(subject.getUser());
+    index.updateIndexDoc(this);
+    
+    FileInfo currentFile = this.getFileInfo().copy();
+    FileInfo newFile = FileInfo.createInstance();
+    // assuming in the same folder
+    newFile.setFolder(currentFile.getFolder());
+    newFile.setFile(Util.convertTitleToFilename(this.getName()));
+    this.setFileInfo(repo.rename(currentFile, newFile, subject));
```
zjffdu: Is the FileInfo used to store the original note title ? If it is, then I think just one filed `oldName` should be sufficient. The class `FileInfo` is not a very clear concept for me what it is used for.

khalidhuseynov: `FileInfo` is used to store current file path under `notebook` dir of the note. Thus, `oldName` represents old (current) path of note before rename, and `newFile` represents where to move the note. This approach is helpful for connecting `folder` structure in GUI with physical folder structure, in that case we could use it if folder was create/changed on front, we would call it with `a/b/noteA.zpln` -> `c/d/noteA.zpln`

khalidhuseynov:
@1ambda on 1. yes, we would need to implement renaming for other repos and i was planning to do it for S3, Azure right after this PR. As an alternative to adding this method to interface, could be having another interface, say `Renamable` and any storage that wants to support extension, naming, folder structure would implement it.

khalidhuseynov:
@Leemoonsoo @1ambda @zjffdu I wanted to have opinion on the note conversion strategy. lazy strategy - converting when note is renamed, vs aggressive - converting all notes when starting zeppelin. currently it's implemented here with lazy strategy, however the more i think, the more aggressive way seems more relevant. One of the problems with lazy strategy is that, if user has many notes, it would be burdensome to convert notes to new format by manually renaming each one.

zjffdu:
How about do the renaming in saving ? We just need to save to the new file and delete the old note file.

khalidhuseynov:
@zjffdu yeah, the renaming would be done probably on saving which would apply if user enters note, and does some action so that save->conversion (rename) is initiated. This would work further everytime to rename from `title1.zpln` -> `title2.zpln`. But I was concerned more about scenario when user has 100 notes, and if he wants to convert them into new format (rename), he would need to enter every note and do something to initiate conversion. That would be little troublesome for user, so i thought maybe to initiate it during initial load for all notes as well. For now i'll change to save, and we can discuss further about loading part.

khalidhuseynov:
@zjffdu @1ambda addressed the feedback to convert in save, w/o additional interface. probably will need to separate base class with each storage after this pr to handle more logic in there. this one is ready to verify

1ambda:
well abstracted and tested PR. LGTM except for

- Currently, it seems that converting the note names only when creating / renaming. Is it right? I so, can we rename them initially as you suggested?

https://github.com/khalidhuseynov/zeppelin/blob/2c36e33deeb2b1802da13ba7fd12214c463c5581/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java#L298

PR #2449, opened 2017-06-29
[Zeppelin-2676] recent notes list

tinkoff-dwh:
### What is this PR for?
Adding recent notes feature: every user now can see his recent notes on homepage. Also you can add recent list on your custom home page.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-2676

Go to homepage. Open some note. Return to homepage. This note has appeared in recent list.
If you open note, which is already in recent list, this note move to the beginning of the recent list.

### Questions:
* Does this needs documentation? Yes (for add recent list to custom homepage) (done)


zeppelin-zengine/src/main/java/org/apache/zeppelin/users/UsersRepo.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public void clearRecent(String user) throws IOException {
+    synchronized (usersInfo) {
+      usersInfo.get(user).clearRecent();
+      saveToFileIfPersistMode();
+    }
+  }
+
+  private void loadFromFile() {
```
zjffdu: I believe `loadFromFile` and `saveToFile` is copied from `InterpreterSetting`. We should avoid copying code as possible as we can. Copying code introduce duplicated code which means we should abstract out some common things.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -85,6 +86,7 @@
   public static Server jettyWebServer;
   public static NotebookServer notebookWsServer;
   public static Helium helium;
+  public static UsersRepo usersRepo;
```
zjffdu: I don't have a more proper name, but `UsersRepo` seems a little confusing to me when it is used to stored the recent opened notes.

zeppelin-zengine/src/main/java/org/apache/zeppelin/users/UsersRepo.java
zjffdu: And I see many places in zeppelin that any little change will cause the whole configuration file to be written, this is not a good practise, and will hurt the performance. IIUC, here each time when user opens a new note, this configuration file be written again. Say we have 100 users, and we keep each user's recent 10 notes, then we need to write 1000 records when user just open a new note. This really doesn't make sense to me. `InterpreterSetting` has the same issue.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
felixcheung: userRecentNote?

docs/usage/other_features/customizing_homepage.md
```diff
@@ -86,4 +86,16 @@ After running the paragraph, you will see output similar to this one:
...
+## Show recent notes in your custom homepage
+To display the recent notes on your custom homepage 
+add the following code to a paragraph in your Apache Zeppelin note and run it.
+
+```javascript
+%spark
 
+println(
+"""%angular
+  <div ng-include="'app/home/recent-notes.html'"></div>
```
felixcheung: I like the idea but this seems somewhat... hacky to use?

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
lunawyll: It is assumed that this class can be used for storing another info about user in future

tinkoff-dwh: It is assumed that this class can be used for storing another info about user in future

zeppelin-zengine/src/main/java/org/apache/zeppelin/users/UsersRepo.java
tinkoff-dwh: `UsersRepo` was written as `Credentials` class. Probably `loadFromFile` and `saveToFile` should be extracted to some `Utils` class?

Problem with written file can be improved by saving `Map<String, UserInfo> usersInfo` to file by timer, but it can cause loss of information

1ambda:
I have a suggestion.

- Place `Recent` before `Notebook` (more upper side)
- Limit the maximum number of recent notebooks. 

That's because users are accessing recent notebook more and 
if there are lots of notebooks, the recent menu will not be seen without scrolling

tinkoff-dwh:
@1ambda 
Number of recent is limited: `MAX_RECENT_SIZE = 10` (https://github.com/apache/zeppelin/pull/2449/files#diff-289945053c8adc894aa4838564528ba5)
Place `Recent` before `Notebook`: good idea, done

zeppelin-zengine/src/main/java/org/apache/zeppelin/users/UsersRepo.java
tinkoff-dwh: @zjffdu
need your reply

docs/usage/other_features/customizing_homepage.md
zjffdu: Agree with @felixcheung This seems not a builtin feature, but some kind of hack.

tinkoff-dwh: @zjffdu 
I've just done this feature as "Show note list in your custom homepage"
https://github.com/apache/zeppelin/blob/master/docs/usage/other_features/customizing_homepage.md

zjffdu: I know, but it is a little weird for me that user need to run a paragraph via spark interpreter to get this feature. Sorry, personally I don't think it worthwhile to add this kind of feature at the cost of adding more complexity to zeppelin codebase. I will wait for others to comment.

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

felixcheung:
I think we have a comment thread on this PR
https://github.com/apache/zeppelin/pull/2449#discussion_r127705438

docs/usage/other_features/customizing_homepage.md
tinkoff-dwh: @felixcheung @zjffdu 
What if I change it on
```
%angular
  <div ng-include="'app/home/recent-notes.html'"></div>
```
?

tinkoff-dwh:
Ready to review

zeppelin-zengine/src/main/java/org/apache/zeppelin/users/UsersRepo.java
Leemoonsoo: As we store more and more configuration, the number of configuration file increases. This introduces some problems

  - A high cost on development.
    Each developer need to implement load/store function when a feature need to store some informations. Also this new configuration file creation needs went through discussion/review process.
  - High I/O overhead.
    Like @zjffdu mentioned, depends on how frequently information should be stored, in some case it may introduce high I/O overhead.
  - Not really following standard convention for *nix application
    Many *nix application stores configuration under /conf or /config dir.
    And it's common system administrator does not give edit (or sometimes both edit/read) permission of those configuration files to the user.
   However, Zeppelin mixes 'human' edit configuration (like conf/zeppelin-site.xml) and machine stored configuration (like interpreter.json, helium.json) in the same directory.
  That makes system administrator difficult to manage permission of two different configuration files.
   https://github.com/apache/zeppelin/pull/2539 is one of an example of the problem.

@zjffdu @tinkoff-dwh How about we create a separate issue for the problem related to configuration files and address separately, while this PR focus on implementing recent list feature?

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
Leemoonsoo: Maybe `usersRepo` does not best describe what it is right now or what it's going to be in the future. I'm not a fan of making a lot of comment in the source code, but in this case, a just simple comment that explains the purpose of this variable may help other developers understand the intention.

zjffdu: I would sugeest not to think too far, just use a proper name to indicate the current purpose. We can refactor that in future if necessary.  It just like `TODO` in codebase that unfortunately many people will never do the `TODO`. @tinkoff-dwh What do you think ?

1ambda: +1 for `TODO` with the variable name for the current state. @tinkoff-dwh and few of community would know the history why this variable name is `userRepo`? But I think the others wouldn't

tinkoff-dwh:
@1ambda @zjffdu 
Thank you for your comments. Classes have been renamed to match its functionality

tinkoff-dwh:
Ready to review (red for other tests )

tinkoff-dwh:
Ready to review

1ambda:
LGTM!

tinkoff-dwh:
Ready to review

1ambda:
@Leemoonsoo, @felixcheung, @zjffdu I think it's good to go. 

Any concerns about backend code like persistency, ...?

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Need rework.

PR #2454, opened 2017-06-29
[ZEPPELIN-2710] Programmatically update progress bar

Leemoonsoo:
### What is this PR for?
This PR adds setProgress(n) method to InterpreterContext.
So user can simply update progress bar manually in the code.

This can be useful when user runs some loops that takes lots of time (e.g. training) and user want to display progress with progress bar.

### What type of PR is it?
Feature

### Todos
* [x] - add setProgress() method
* [x] - Unittest

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2710

run following example and see if progress bar is updating
```
%python
import time

ic = z.getInterpreterContext()

for i in range(0, 100):
    time.sleep(0.05)
    ic.setProgress(i)
```

zeppelin_progress

### Questions:


Leemoonsoo:
Merge to master if no further comment.

PR #2455, opened 2017-06-30
ZEPPELIN-1515. Notebook: HDFS as a backend storage (Use hadoop client jar)

zjffdu:
### What is this PR for?
This PR is trying to add hdfs as another implementation for `NotebookRepo`. There's another PR about using webhdfs to implement that. Actually hdfs client library is compatibility cross major versions. See http://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/Compatibility.html#Wire_compatibility, if using webhdfs, the code become more complicated and may lose some features of hdfs. 

This PR is also required for HA of zeppelin, so that multiple zeppelin instances can share notes via hdfs.  I add hadoop-client in pom file. So zeppelin will package hadoop client jar into its binary distribution. This is because zeppelin may be installed in a gateway machine where no hadoop is installed (only hadoop configuration file is existed in this machine) And since the hadoop client will work with multiple versions of hadoop, so it is fine to package into binary distribution. Spark also package hadoop client jar in its binary distribution. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1515

Unit test is added.  Also manually verify it in a single node cluster. 

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -173,6 +173,26 @@
 </property>
 -->
 
+<!-- Notebook storage layer using hdfs file system
+<property>
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.HdfsNotebookRepo</value>
+  <description>hdfs notebook persistence layer implementation</description>
+</property>
+
+<property>
+  <name>zeppelin.hdfs.keytab</name>
+  <value></value>
+  <description>keytab for accessing kerberized hdfs</description>
+</property>
+
+<property>
+  <name>zeppelin.hdfs.principal</name>
+  <value></value>
+  <description>principal for accessing kerberized hdfs</description>
+</property>
+-->
+
 <!-- For connecting your Zeppelin with ZeppelinHub -->
 <!--
 <property>
--- a/docs/setup/storage/storage.md
+++ b/docs/setup/storage/storage.md
@@ -30,6 +30,7 @@ There are few notebook storage systems available for a use out of the box:
 
   * (default) use local file system and version it using local Git repository - `GitNotebookRepo`
   * all notes are saved in the notebook folder in your local File System - `VFSNotebookRepo`
+  * all notes are saved in the notebook folder in hdfs - `HdfsNotebookRepo`
   * storage using Amazon S3 service - `S3NotebookRepo`
   * storage using Azure service - `AzureNotebookRepo`
   * storage using MongoDB - `MongoNotebookRepo`
@@ -50,6 +51,22 @@ To enable versioning for all your local notebooks though a standard Git reposito
   <description>notebook persistence layer implementation</description>
 </property>
 ```
+
+</br>
+
+## Notebook Storage in Hdfs repository <a name="Hdfs"></a>
+
+Notes may be stored in hdfs, so that multiple Zeppelin instances can share the same notes. It supports all the versions of hadoop 2.x. If you use `HdfsNotebookRepo`, then `zeppelin.notebook.dir` is the path on hdfs. And you need to specify `HADOOP_CONF_DIR` in `zeppelin-env.sh` so that zeppelin can find the right hadoop configuration files.
+If your hadoop cluster is kerberized, then you need to specify `zeppelin.hdfs.keytab` and `zeppelin.hdfs.principal`
+
+```
+<property>
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.HdfsNotebookRepo</value>
+  <description>hdfs notebook persistence layer implementation</description>
+</property>
+```
+
 
 </br>
 
--- a/zeppelin-zengine/pom.xml
+++ b/zeppelin-zengine/pom.xml
@@ -36,6 +36,7 @@
 
   <properties>
     <!--library versions-->
+    <hadoop.version>2.6.0</hadoop.version>
     <commons.lang3.version>3.4</commons.lang3.version>
     <commons.vfs2.version>2.0</commons.vfs2.version>
     <aws.sdk.s3.version>1.10.62</aws.sdk.s3.version>
@@ -301,6 +302,71 @@
       <version>1.5</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-client</artifactId>
+      <version>${hadoop.version}</version>
+
+      <exclusions>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-json</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-server</artifactId>
+        </exclusion>
+
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.jackrabbit</groupId>
+          <artifactId>jackrabbit-webdav</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-httpclient</groupId>
+          <artifactId>commons-httpclient</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.zookeeper</groupId>
+          <artifactId>zookeeper</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.eclipse.jgit</groupId>
+          <artifactId>org.eclipse.jgit</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.jcraft</groupId>
+          <artifactId>jsch</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-compress</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>xml-apis</groupId>
+          <artifactId>xml-apis</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>xerces</groupId>
+          <artifactId>xercesImpl</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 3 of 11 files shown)


zjffdu:
@Leemoonsoo @felixcheung @khalidhuseynov  Could you help review it ? Thanks

jongyoul:
Hi, Recently, I've found some version of hadoop related library make a crash of jetty which is used in Zeppelin Server. I've also changed yarn cluster manager to use another classloader. How do you think you can use another classloader while running hdfs notebook storage? It would make more changes in Zeppelin server side but it will be better for the future.

zjffdu:
@jongyoul What kind of issue do you see in cluster manager PR ? Usually this is due to jar version conflict. You can exclude this transitive jar of hadoop client library.

conf/zeppelin-site.xml.template
```diff
@@ -173,6 +173,14 @@
 </property>
 -->
 
+<!-- Notebook storage layer using hdfs file system
+<property>
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.HdfsNotebookRepo</value>
+  <description>local notebook persistence layer implementation</description>
```
felixcheung: `local notebook persistence layer ` -> `hdfs notebook persistence layer `?

docs/setup/storage/storage.md
```diff
@@ -53,6 +54,20 @@ To enable versioning for all your local notebooks though a standard Git reposito
 
 </br>
 
+## Notebook Storage in Hdfs repository <a name="Hdfs"></a>
+
+Notes may be stored in hdfs, so that multiple zeppelin instance can share the same notes. It supported all the versions of hadoop 2.x. If you use `HdfsNotebookRepo`, then `zeppelin.notebook.dir` is the path on hdfs. And you need to specify `HADOOP_CONF_DIR` in `zeppelin-env.sh` so that zeppelin can find the right hadoop configuration files.
```
felixcheung: capital `Zeppelin`

`Notes may be stored in hdfs, so that multiple zeppelin instance can share the same notes.`
- if this is the intended goal, will there be any race condition on write when several instances are changing it the same time?

zeppelin-zengine/pom.xml
```diff
@@ -301,6 +301,71 @@
       <version>1.5</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-client</artifactId>
+      <version>2.6.0</version>
```
felixcheung: use a property, `hadoop.version` instead of hardcoding to 2.6

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,134 @@
...
+  @Override
+  public void save(Note note, AuthenticationInfo subject) throws IOException {
+    Path tmpNotePath = new Path(notebookDir.toString() + "/" + note.getId() + "/.note.json");
+    if (fs.exists(tmpNotePath)) {
+      fs.delete(tmpNotePath, true);
+    }
+    InputStream in = new ByteArrayInputStream(note.toJson().getBytes("UTF-8"));
+    IOUtils.copyBytes(in, fs.create(tmpNotePath), hadoopConf);
+    Path notePath = new Path(notebookDir.toString() + "/" + note.getId() + "/note.json");
+    fs.delete(notePath);
```
felixcheung: should it overwrite without warning like this?

docs/setup/storage/storage.md
zjffdu: Right, but I think it is supposed to be handled in a base class.

khalidhuseynov: minor: `It supports`?

docs/setup/storage/storage.md
```diff
@@ -53,6 +54,20 @@ To enable versioning for all your local notebooks though a standard Git reposito
...
+ 
+```
+<property>
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.HdfsNotebookRepo</value>
+  <description>notebook persistence layer implementation</description>
```
khalidhuseynov: i guess should be changed to same description as above with hdfs

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
zjffdu: I am afraid this is what we only can do for now. Each time when we save a note, it would overwrite the previous one. And actually you don't know the file is the previous note or other file.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,134 @@
...
+    return Note.fromJson(new String(noteBytes.toString("UTF-8")));
+  }
+
+  @Override
+  public void save(Note note, AuthenticationInfo subject) throws IOException {
+    Path tmpNotePath = new Path(notebookDir.toString() + "/" + note.getId() + "/.note.json");
+    if (fs.exists(tmpNotePath)) {
+      fs.delete(tmpNotePath, true);
+    }
+    InputStream in = new ByteArrayInputStream(note.toJson().getBytes("UTF-8"));
```
khalidhuseynov: i think we would need `zConf.ZEPPELIN_ENCODING` instead of hard-coding `"UTF-8` in all these encoding cases

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,134 @@
...
+          "specify another directory");
+    }
+    LOGGER.info("Use hdfs directory {} to store notebook", notebookDir);
+  }
+
+  @Override
+  public List<NoteInfo> list(AuthenticationInfo subject) throws IOException {
+    List<NoteInfo> noteInfos = new ArrayList<>();
+    for (FileStatus status : fs.globStatus(new Path(notebookDir, "*/note.json"))) {
+      NoteInfo noteInfo = new NoteInfo(status.getPath().getParent().getName(), "", null);
```
khalidhuseynov: shouldn't we create `Note` object first and then `new NoteInfo(note)`? otherwise it would be missing name and config

zjffdu: This is what I am confused. I don't know the purpose of `NoteInfo`, previously I thought `NoteInfo` represent the basic metadata of Note, not include the note content. And from my testing, this PR works well. Besides, I create ZEPPELIN-2712 that we would read Note twice, the root cause is that we construct `NoteInfo` from `Note`.

khalidhuseynov: good point, as i checked only Id field of `NoteInfo` is used to retrieve the whole note which would lead to duplicate get operation. from another point of view if there's no special folder structure as `noteId/note.json` it would make more sense to retrieve `List<Note>` in `list()` method. then the way used here looks good for now

jongyoul:
@zjffdu Almost conflicts will be removed by extracting jars but how we guarantee those libraries are not related my features? I'm just curious.

zjffdu:
@jongyoul Yeah, since your PR is also using the hadoop jar. I think we need to do 2 things:
* Use the same hadoop jar version
* Decide whether include it in distribution or not. For this, I personally prefer to include the hadoop jar in distribution. Several Reasons:
  - It is much easy to implement, we don't need to find the jars at runtime by ourselves.
  - Zeppelin might be installed in a gateway machine where hadoop is not installed, in this case, zeppelin would not work because if could not find the hadoop jars.  
  - Spark put hadoop jar in its distribution and approve it can work, so I think we can trust this approach. 

Let me know your concern, thanks.

hayssams:
@zjffdu I think that the zeppelin.sh file should be updated with the HADOOP_CONF_DIR in the classpath

zjffdu:
Thanks for review. @hayssams I did that in zeppelin-daemon.sh.

hayssams:
Yes but sometimes we need to launch zeppelin through zeppelin.sh equally in Mesos by Marathon

zjffdu:
@hayssams Any reason why `zeppelin-daemon.sh` doesn't fit for you ? Because zeppelin does not guarantee `zeppelin.sh` can launch zeppelin server properly. `zeppelin-daemon.sh` not only add `HADOOP_CONF_DIR` but also other libraries, if you don't use `zeppelin-daemon.sh`, zeppelin-server may fail to launch.

hayssams:
@zjffdu Yes when you want to start zeppelin not as a daemon. 
Please also note the docker image rely on zeppelin.sh

zjffdu:
Also add HADOOP_CONF_DIR in zeppelin.sh, but it looks like there's some code duplication between `zeppelin.sh` and `zeppelin-daemon.sh`, we need a followup PR to remove these duplication.

zjffdu:
I also verified the kerberized hdfs manually in my local box.  @jongyoul @Leemoonsoo @felixcheung @prabhjyotsingh @hayssams Please help review it, Thanks

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,200 @@
...
+    LOGGER.warn("updateSettings is not implemented for HdfsNotebookRepo");
+  }
+
+  private interface HdfsOperation<T> {
+    T call() throws IOException;
+  }
+
+  public <T> T callHdfsOperation(final HdfsOperation<T> func) throws IOException {
+    if (isSecurityEnabled) {
+      UserGroupInformation.getLoginUser().reloginFromKeytab();
```
prabhjyotsingh: re login on every operation?

prabhjyotsingh:
Tested on local, works as expected, have tried on both environments, with and without Kerberos.
LGTM, except a minor comment.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
zjffdu: `reloginFromKeytab()` will check whether the keytab is expired, if not it won't relogin

https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L1288

hayssams:
LGTM 👍

zjffdu:
Thanks for review @prabhjyotsingh @hayssams , will merge it if no more comments

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HdfsNotebookRepo.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @Override
+  public void remove(final String noteId, AuthenticationInfo subject) throws IOException {
+    callHdfsOperation(new HdfsOperation<Void>() {
+      @Override
+      public Void call() throws IOException {
+        Path noteFolder = new Path(notebookDir.toString() + "/" + noteId);
+        fs.delete(noteFolder, true);
+        return null;
```
zjffdu: Right, there's nothing needs to be return for this method.

PR #2456, opened 2017-06-30
[ZEPPELIN-2714] Soft-code Spark UI button visualization

necosta:
### What is this PR for?
When "spark.ui.enabled" property is set to "false" we should not show the Spark UI button.
We keep the same behaviour when this property does not exist or when it exists and it's set to true.

### What type of PR is it?
[ Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2714

1. Open Zeppelin, go to Interpreter > Spark
1. Click spark ui > msg: "No spark application running"
1. Go to a paragraph, run "sc.parallelize(1 to 100).count()", check "SPARK JOB" button
1. Go to Interpreter > Spark > Set "spark.ui.enabled" to "false"
1. Rerun paragraph > "SPARK JOB" button not visible
1. Go to Interpreter > Spark > "spark ui" button not visible
Also, test with "spark.ui.enabled"="true" and other workflow combinations

### Questions:

### DIFF
--- a/docs/usage/rest_api/interpreter.md
+++ b/docs/usage/rest_api/interpreter.md
@@ -615,7 +615,7 @@ The role of registered interpreters, settings and interpreters group are describ
     </tr>
     <tr>
       <td>URL</td>
-      <td>```http://[zeppelin-server]:[zeppelin-port]/api/interpreter/getmetainfos/[setting ID]```</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/interpreter/metadata/[setting ID]```</td>
     </tr>
     <tr>
       <td>Success code</td>
--- a/docs/usage/rest_api/interpreter.md
+++ b/docs/usage/rest_api/interpreter.md
@@ -604,3 +604,26 @@ The role of registered interpreters, settings and interpreters group are describ
     </td>        
   </table>  
   
+<br/>
+### Get interpreter settings metadata info
+
+  <table class="table-configuration">
+    <col width="200">
+    <tr>
+      <td>Description</td>
+      <td>This ```GET``` method returns interpreter settings metadata info. </td>
+    </tr>
+    <tr>
+      <td>URL</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/interpreter/getmetainfos/[setting ID]```</td>
+    </tr>
+    <tr>
+      <td>Success code</td>
+      <td>200</td>
+    </tr>
+    <tr>
+      <td>Fail code</td>
+      <td> 500 </td>
+    </tr>
+  </table>
+  
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -184,11 +184,12 @@ public class SparkInterpreter extends Interpreter {
         super.onJobStart(jobStart);
         int jobId = jobStart.jobId();
         String jobGroupId = jobStart.properties().getProperty("spark.jobGroup.id");
+        String uiEnabled = jobStart.properties().getProperty("spark.ui.enabled");
         String jobUrl = getJobUrl(jobId);
         String noteId = Utils.getNoteId(jobGroupId);
         String paragraphId = Utils.getParagraphId(jobGroupId);
 
-        if (jobUrl != null && noteId != null && paragraphId != null) {
+        if (uiEnabled != "false" && jobUrl != null) {
           RemoteEventClientWrapper eventClient = BaseZeppelinContext.getEventClient();
           Map<String, String> infos = new java.util.HashMap<>();
           infos.put("jobUrl", jobUrl);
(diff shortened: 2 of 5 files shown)


docs/usage/rest_api/interpreter.md
```diff
@@ -541,3 +541,26 @@ The role of registered interpreters, settings and interpreters group are describ
...
+
+  <table class="table-configuration">
+    <col width="200">
+    <tr>
+      <td>Description</td>
+      <td>This ```GET``` method returns interpreter settings metadata info. </td>
+    </tr>
+    <tr>
+      <td>URL</td>
+      <td>```http://[zeppelin-server]:[zeppelin-port]/api/interpreter/getmetainfos/[setting ID]```</td>
```
felixcheung: hmm, why not call it get `matadata`?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -168,11 +167,12 @@ public synchronized void onJobStart(SparkListenerJobStart jobStart) {
         super.onJobStart(jobStart);
         int jobId = jobStart.jobId();
         String jobGroupId = jobStart.properties().getProperty("spark.jobGroup.id");
+        String uiEnabled = jobStart.properties().getProperty("spark.ui.enabled");
         String jobUrl = getJobUrl(jobId);
         String noteId = Utils.getNoteId(jobGroupId);
         String paragraphId = Utils.getParagraphId(jobGroupId);
           
-        if (jobUrl != null && noteId != null && paragraphId != null) {
+        if (uiEnabled != "false" && jobUrl != null) {
```
felixcheung: does this need to check for string content, eg. trim whitespaces, lower case etc? since this is user input?

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -723,16 +723,21 @@ function InterpreterCtrl ($rootScope, $scope, $http, baseUrlSrv, ngToast, $timeo
...
-    $http.get(baseUrlSrv.getRestApiBase() + '/interpreter/getmetainfos/' + settingId + '?propName=url')
+    $http.get(baseUrlSrv.getRestApiBase() + '/interpreter/getmetainfos/' + settingId)
       .success(function (data, status, headers, config) {
-        let url = data.body.url
-        if (!url) {
-          BootstrapDialog.alert({
-            message: 'No spark application running'
-          })
+        if (data.body === undefined || !data.body.url) {
+          if (data.body !== undefined && data.body.sparkUiEnabled === 'false') {
```
Leemoonsoo: Thanks for the contribution.

Do you think we can make this function more generic? so all other interpreter can be benefit by it.

For example, 

* `data.body.url` : URL to open on click
* `data.body.message` : Message to show on click

instead of `data.body.sparkUiEnabled`.

What do you think?

docs/usage/rest_api/interpreter.md
necosta: Renamed to metadata

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
necosta: Refactored code. Now checks for whitespaces and case sensitivity.

zeppelin-web/src/app/interpreter/interpreter.controller.js
necosta: Good idea. Refactored code so we only have url and message. More generic no doubt.

necosta:
@felixcheung , @Leemoonsoo , ready for re-review. Thanks

Leemoonsoo:
LGTM!

Leemoonsoo:
Merge to master if no further comment

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -168,11 +167,14 @@ public synchronized void onJobStart(SparkListenerJobStart jobStart) {
...
         int jobId = jobStart.jobId();
         String jobGroupId = jobStart.properties().getProperty("spark.jobGroup.id");
+        String uiEnabled = jobStart.properties().getProperty("spark.ui.enabled");
         String jobUrl = getJobUrl(jobId);
         String noteId = Utils.getNoteId(jobGroupId);
         String paragraphId = Utils.getParagraphId(jobGroupId);
-          
-        if (jobUrl != null && noteId != null && paragraphId != null) {
+        // Button visible if Spark UI property not set, set as invalid boolean or true
+        java.lang.Boolean showSparkUI =
```
felixcheung: btw, why is the full name `java.lang.Boolean` needed here?

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -723,16 +723,21 @@ function InterpreterCtrl ($rootScope, $scope, $http, baseUrlSrv, ngToast, $timeo
...
             message: 'No spark application running'
           })
           return
         }
-        window.open(url, '_blank')
+        if (data.body.url) {
+          window.open(data.body.url, '_blank')
+        } else {
+          BootstrapDialog.alert({
+            message: data.body.message
```
felixcheung: is popping up a dialog a bit disruptive? it seems data.body.url can be null when the spark job is not running/started yet, for example?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -1018,21 +1020,28 @@ public String getSparkUIUrl() {
...
-        if (ctx != null && ctx.getClient() != null) {
-          logger.info("Sending metainfos to Zeppelin server: {}", infos.toString());
-          getZeppelinContext().setEventClient(ctx.getClient());
-          ctx.getClient().onMetaInfosReceived(infos);
-        }
+    sparkUrl = getSparkUIUrl();
+    Map<String, String> infos = new java.util.HashMap<>();
+    infos.put("url", sparkUrl);
+    java.lang.Boolean uiEnabled = java.lang.Boolean.parseBoolean(
+            property.getProperty("spark.ui.enabled", "true"));
```
felixcheung: so we have trim here https://github.com/apache/zeppelin/pull/2456/files#diff-dbda0c4083ad9c59ff05f0273b5e760fR176
but not this line? I think it would be better to be consistent. I'm ok if we don't trim for other boolean property value

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
necosta: Error: Boolean is ambiguous, both 'scala.Boolean' and 'java.lang.Boolean' match

zeppelin-web/src/app/interpreter/interpreter.controller.js
necosta: I kept the same logic as before in the sense that, if Spark is not running, user gets popup: "No Spark application running". 
Cannot comment about a popup being disruptive or not. 
And on data.body.url being null and Spark running (only case we hit line 737 dialog), user gets either message "Spark UI disabled" which I think is correct or message "No spark url defined", which is a non-replicable scenario in a normal workflow.
I don't see anything we should change

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
necosta: Ok, I can trim here as well as I also need to rebase to fix some merge conflicts. parseBoolean already deals with case sensitive, so only trim() is needed.

necosta:
@Leemoonsoo , @felixcheung , one minor commit added. Fixed merge conflicts. Thanks.

felixcheung:
LGTM.
merging if no more comment

zeppelin-web/src/app/interpreter/interpreter.controller.js
felixcheung: ok then, thanks

PR #2465, opened 2017-07-06
[ZEPPELIN-2724] Move NotebookSocket.send() out of synchronized block

Leemoonsoo:
### What is this PR for?
This PR tries to address https://issues.apache.org/jira/browse/ZEPPELIN-2724.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Move NotebookSocket.send() out of synchronized block
* [x] - Waits for verification this actually fixes the problem

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2724

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -483,40 +483,45 @@ public class NotebookServer extends WebSocketServlet
   }
 
   private void broadcast(String noteId, Message m) {
+    List<NotebookSocket> socketsToBroadcast;
     synchronized (noteSocketMap) {
       broadcastToWatchers(noteId, StringUtils.EMPTY, m);
       List<NotebookSocket> socketLists = noteSocketMap.get(noteId);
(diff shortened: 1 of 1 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -483,40 +483,45 @@ private void broadcastToNoteBindedInterpreter(String interpreterGroupId, Message
   }
 
   private void broadcast(String noteId, Message m) {
+    List<NotebookSocket> socketsToBroadcast;
```
felixcheung: shouldn't this be initialized to an empty list, I think?

sergeymazin:
I can confirm that deadlock issue is fixed by this PR.

Leemoonsoo:
Thanks @sergeymazin for confirming.

I'm merging it to master and branch-0.7

jeffonia:
I'm curios about can we change the noteSocketMap from the HashMap to ConcurrentHashMap, a segment level lock make the throughput about 16 times than this for the most optimistic scenario.

PR #2468, opened 2017-07-06
[ZEPPELIN-2733] Remove System Information Leak in Authentication.java.

yanboliang:
### What is this PR for?
An information leak occurs when system data or debugging information leaves the program through an output stream or logging function.
In the file Authentication.java,
```
Line 137: LOG.debug("Encrypted user key is {}", userKey);
Line 148: LOG.debug("IV is {}, IV length is {}", initVector, initVector.length());
```
These lines may print information which can reveal some important data to user making it vulnerable to attacks, we should not log this sensitive information.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2733

Existing tests.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/zeppelinhub/security/Authentication.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/zeppelinhub/security/Authentication.java
@@ -134,7 +134,6 @@ public class Authentication implements Runnable {
 
   // returns login:password
   private String getAuthKey(String userKey) {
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @yanboliang for the fix.
LGTM and merge to master if no further discussion.

soralee:
@yanboliang Thanks for taking care of this. 
Yeah, It makes sense.
I guess this PR works well, because you just removed log. But it would be great to make sure that CI pass.
Could you confirm it?

yanboliang:
@Leemoonsoo @soralee It passed CI test now. Thanks.

PR #2471, opened 2017-07-07
[ZEPPELIN-2744] [minor] Increase error handling in JDBC interpreter

prabhjyotsingh:
### What is this PR for?
At times running a phoenix sql query throws "ResultSet is closed."

```
java.sql.SQLException: ERROR 1101 (XCL01): ResultSet is closed.
at org.apache.phoenix.exception.SQLExceptionCode$Factory$1.newException(SQLExceptionCode.java:442)
at org.apache.phoenix.exception.SQLExceptionInfo.buildException(SQLExceptionInfo.java:145)
at org.apache.phoenix.jdbc.PhoenixResultSet.checkOpen(PhoenixResultSet.java:215)
at org.apache.phoenix.jdbc.PhoenixResultSet.next(PhoenixResultSet.java:772)
at org.apache.commons.dbcp2.DelegatingResultSet.next(DelegatingResultSet.java:191)
at org.apache.commons.dbcp2.DelegatingResultSet.next(DelegatingResultSet.java:191)
at org.apache.zeppelin.jdbc.JDBCInterpreter.executeSql(JDBCInterpreter.java:705)
at org.apache.zeppelin.jdbc.JDBCInterpreter.interpret(JDBCInterpreter.java:775)
at org.apache.zeppelin.interpreter.LazyOpenInterpreter.interpret(LazyOpenInterpreter.java:101)
at org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer$InterpretJob.jobRun(RemoteInterpreterServer.java:503)
at org.apache.zeppelin.scheduler.Job.run(Job.java:181)
at org.apache.zeppelin.scheduler.ParallelScheduler$JobRunner.run(ParallelScheduler.java:162)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)

```

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-2744](https://issues.apache.org/jira/browse/ZEPPELIN-2744)

Try running a Phoenix query like `select * from SYSTEM.CATALOG` for me it was failing on a Kerberos enabled environment (on CentOS6.6 with Hbase-1.1.2).

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -702,6 +702,13 @@ public class JDBCInterpreter extends KerberosInterpreter {
               String results = getResults(resultSet,
                   !containsIgnoreCase(sqlToExecute, EXPLAIN_PREDICATE));
               interpreterResult.add(results);
+              try {
+                if (resultSet.isClosed()) {
+                  break;
+                }
+              } catch (Exception e) {
+                logger.info("Exception checking isClosed on resultSet, nothing to worry");
+              }
               if (resultSet.next()) {
                 interpreterResult.add(ResultMessages.getExceedsLimitRowsMessage(getMaxResult(),
                     String.format("%s.%s", COMMON_KEY, MAX_LINE_KEY)));


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -702,6 +702,13 @@ private InterpreterResult executeSql(String propertyKey, String sql,
               String results = getResults(resultSet,
                   !containsIgnoreCase(sqlToExecute, EXPLAIN_PREDICATE));
               interpreterResult.add(results);
+              try {
+                if (resultSet.isClosed()) {
+                  break;
+                }
+              } catch (Exception e) {
```
prabhjyotsingh: I had to do this in a try-catch block because this method is not implemented in Hive https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveBaseResultSet.java#L628-L630, which in my opinion is not an error.

felixcheung: causing an exception on every execution sounds expensive though?

prabhjyotsingh: Yes, agreed, but until `java.sql.SQLException: ERROR 1101 (XCL01): ResultSet is closed.
` is fixed for Phoenix or `isClosed()` implemented in Hive, it's a conscious choice that I think we should take. Because for me it happens all the time;

screen shot 2017-07-08 at 9 26 31 am


But, yes it is expensive. Any other thought?

felixcheung: I see.. if the exception "only happens" with Phoenix maybe it's not too bad...

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -702,6 +702,13 @@ private InterpreterResult executeSql(String propertyKey, String sql,
               String results = getResults(resultSet,
                   !containsIgnoreCase(sqlToExecute, EXPLAIN_PREDICATE));
               interpreterResult.add(results);
+              try {
+                if (resultSet.isClosed()) {
+                  break;
```
felixcheung: `if (resultSet.isClosed())` is true, we break out of the loop to run the sql statement.
is that an error condition there - since not everything in the paragraph is executed?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: Yes, agreed, since this is generic I thought this interpreter should try to handle errors/exception gracefully if possible.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
prabhjyotsingh: Yes, kind of thought about it, but since resultSet is closed; which means no other option can be performed on this resultSet, and hence break loop, enter into finally block and over there close both resultSet and statement.

felixcheung:
It should then report an error in the InterpreterResult?

prabhjyotsingh:
Yes, that's what it is doing now (in the case of Phoenix SQL) it gives a valid result with error stack trace, and paragraph results in error. Hence, I thought we could either choose to improve it or leave it.

felixcheung:
ok, I guess to clarify, the code is getting one `sqlToExecute` [at a time](https://github.com/apache/zeppelin/pull/2471/files#diff-ecdae8ee9594a5c4b21a3c217a3f130cR681).
In the case of Phoenix, perhaps somehow `resultSet.isClosed()` happens, and can be ignored if it is already at the end of the `sqlArray`.
But, if it is not, then the sql in the paragraph is partially executed (say, only 3 out of 10), and it will be an error? Would that happen that way and if it does, that would be something to report in InterpreterResults?

prabhjyotsingh:
Sure, I see your point, will close this, and mark ZEPPELIN-2744 works as designed.

PR #2474, opened 2017-07-10
[ZEPPELIN-2753] Basic Implementation of IPython Interpreter

zjffdu:
### What is this PR for?
This is the first step for implement IPython Interpreter in Zeppelin.  I just use the jupyter_client to create and manage the ipython kernel. We don't need to care about python compilation and execution, all the things are delegated to ipython kernel. Ideally all the features of ipython should be available in Zeppelin as well.  

For now, user can use %python.ipython for IPython Interpreter. And if ipython is available, the default python interpreter will use ipython. But user can still set `zeppelin.python.useIPython` as false to enforce to use the old implementation of python interpreter. 

Main features:
* IPython interpreter support
** All the ipython features are available, including visualization, ipython magics. 
* ZeppelinContext support
* Streaming output support
* Support Ipython in PySpark

Regarding the visualization, ideally all the visualization libraries work in jupyter should also work here.
In unit test, I only verify the following 3 popular visualization library. could add more later.
* matplotlib
* bokeh
* ggplot

### What type of PR is it?
[Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2753

Unit test is added. 

Verify bokeh in IPython Interpreter
image

Verify matplotlib
image

Verify ZeppelinContext

image

Verify Streaming
streaming

### Questions:


alluxio/pom.xml
```diff
@@ -47,6 +47,7 @@
         <dependency>
               <groupId>com.google.guava</groupId>
               <artifactId>guava</artifactId>
+              <version>15.0</version>
```
zjffdu: We should not put guava in zeppelin-interpreter as each interpreter may need different versions of guava. E.g. Pig needs 15.0 while Python needs 18.0

zjffdu:
@Leemoonsoo @1ambda @felixcheung @prabhjyotsingh Please help review.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -0,0 +1,276 @@
...
+  private long ipythonLaunchTimeout;
+
+  public IPythonInterpreter(Properties properties) {
+    super(properties);
+  }
+
+  @Override
+  public void open() {
+    try {
+      pythonExecutable = getProperty().getProperty("zeppelin.python", "python");
```
Leemoonsoo: PythonCondaInterpreter allows [change python environment](https://github.com/apache/zeppelin/blob/master/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java#L115) by running `%python.conda activate [env name]`.

Do you think PythonCondaInterpreter can handle for IPythonInterpreter, too?

PythonDockerInterpreter also have ability to run python env using docker container. Can this applied to IPythonInterpreter?
https://github.com/apache/zeppelin/blob/master/python/src/main/java/org/apache/zeppelin/python/PythonDockerInterpreter.java

zjffdu:
Thanks for testing @Leemoonsoo . These special characters are for font color, create ZEPPELIN-2756 for that which might need help from @1ambda

zjffdu:
@Leemoonsoo Regarding the ZeppelinContext, could you try that again ? It is covered in unit test, so should be able to work.  I force committed it yesterday, the latest commit Id is 956befa4151128b9ee76df2e3eada512ab9e1025

docs/interpreter/python.md
```diff
@@ -232,6 +232,67 @@ SELECT * FROM rates WHERE age < 40
...
+IPython is more powerful than the default python interpreter with extra functionality. You can use IPython2 or IPython3 which depends on which python you set `zeppelin.python`.
+
+   **Pre-requests**
+   
+    - Jupyter `pip install jupyter`
+    - IPython `pip install IPython`
+    - ipykernel `pip install ipykernel`
+    - grpcio `pip install grpcio`
+
+Besides the basic function of python interpreter. You can use IPython as you use it in Jupyter Notebook.
```
felixcheung: is the goal everything in jupyter would work here as-is?
like ipywidgets?

python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
```diff
@@ -0,0 +1,95 @@
...
+  public StatusResponse status(StatusRequest request) {
+    return blockingStub.status(request);
+  }
+
+  public void stop(StopRequest request) {
+    asyncStub.stop(request, null);
+  }
+
+  public static void main(String[] args) {
+    IPythonClient client = new IPythonClient("localhost", 50053);
```
felixcheung: should this port by fixed? it won't work if in isolated mode, right?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -0,0 +1,276 @@
...
+      ipythonLaunchTimeout = Long.parseLong(
+          getProperty().getProperty("zeppelin.ipython.timeout", "10000"));
+      this.zeppelinContext = new PythonZeppelinContext(
+          null, Integer.parseInt(getProperty().getProperty("zeppelin.python.maxResult", "1000")));
+      int ipythonPort = RemoteInterpreterUtils.findRandomAvailablePortOnAllLocalInterfaces();
+      int jvmGatewayPort = RemoteInterpreterUtils.findRandomAvailablePortOnAllLocalInterfaces();
+      LOGGER.info("Launching IPython Kernel at port: " + ipythonPort);
+      LOGGER.info("Launching JVM Gateway at port: " + jvmGatewayPort);
+      ipythonClient = new IPythonClient("127.0.0.1", ipythonPort);
```
felixcheung: this uses localhost here https://github.com/apache/zeppelin/pull/2474/files#diff-2b2ea0856ba412da72d4880f52ce0717R82

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -0,0 +1,276 @@
...
+      launchIPythonKernel(ipythonPort);
```
felixcheung: call launchIPythonKernel before `new IPythonClient` above?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -0,0 +1,276 @@
...
+        StatusResponse response = ipythonClient.status(StatusRequest.newBuilder().build());
+        if (response.getStatus() == IPythonStatus.RUNNING) {
+          LOGGER.info("IPython Kernel is Running");
+          break;
+        } else {
+          LOGGER.info("Wait for IPython Kernel to be started");
+        }
+      } catch (Exception e) {
+        // ignore the exception, because is may happen when grpc server has not started yet.
+        LOGGER.info("Wait for IPython Kernel to be started");
```
felixcheung: log the exception too

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -0,0 +1,276 @@
...
+    executor.setWatchdog(watchDog);
+    executor.execute(cmd, EnvironmentUtils.getProcEnvironment(), this);
+
+    // wait until IPython kernel is started or timeout
+    long startTime = System.currentTimeMillis();
+    while (true) {
+      try {
+        Thread.sleep(100);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
```
felixcheung: don't print, use LOGGER?

python/src/main/resources/grpc/python/ipython_pb2.py
```diff
@@ -0,0 +1,718 @@
...
...status\x18\x01 \x01(\x0e\x32\x16.ipython.ExecuteStatus\x12\x13\n\x0btext_output\x18\x02 \x01(\t\x12\x14\n\x0cimage_output\x18\x03 \x01(\t\"\x0f\n\rCancelRequest\"\x10\n\x0e\x43\x61ncelResponse\"!\n\x11\x43ompletionRequest\x12\x0c\n\x04\x63ode\x18\x01 \x01(\t\"%\n\x12\x43ompletionResponse\x12\x0f\n\x07matches\x18\x01 \x03(\t\"\x0f\n\rStatusRequest\"8\n\x0eStatusResponse\x12&\n\x06status\x18\x01 \x01(\x0e\x32\x16.ipython.IPythonStatus\"\r\n\x0bStopRequest\"\x0e\n\x0cStopResponse*\'\n\rExecuteStatus\x12\x0b\n\x07SUCCESS\x10\x00\x12\t\n\x05\x45RROR\x10\x01**\n\rIPythonStatus\x12\x0c\n\x08STARTING\x10\x00\x12\x0b\n\x07RUNNING\x10\x01\x32\xc1\x02\n\x07IPython\x12>\n\x07\x65xecute\x12\x17.ipython.ExecuteRequest\x1a\x18.ipython.ExecuteResponse\"\x00\x12\x45\n\x08\x63omplete\x12\x1a.ipython.CompletionRequest\x1a\x1b.ipython.CompletionResponse\"\x00\x12;\n\x06\x63\x61ncel\x12\x16.ipython.CancelRequest\x1a\x17.ipython.CancelResponse\"\x00\x12;\n\x06status\x12\x16.ipython.StatusRequest\x1a\x17.ipython.StatusResponse\"\x00\x12\x35\n\x04stop\x12\x14.ipython.StopRequest\x1a\x15.ipython.StopResponse\"\x00\x42\x36\n\x1aorg.apache.zeppelin.pythonB\x0cIPythonProtoP\x01\xa2\x02\x07IPythonb\x06proto3')
```
felixcheung: this is hard to read - what is this for?

python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
zjffdu: The main functions is only for my local testing, won't be called in zeppelin.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: `launchIPythonKernel` needs `iPythonClient` to check its status. IPython constructor won't connect to that host:port, it would only connect when grpc method is called.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: As the comment mentions, we can ignore this exception, because it may happen when grpc server has not started yet.

python/src/main/resources/grpc/python/ipython_pb2.py
zjffdu: This is generated by grpc from proto file , just ignore it.

docs/interpreter/python.md
zjffdu: Good points, I havn't tried that. Let me try.

zjffdu: It doesn't work, it needs some javascripts, seems it depends on jupyter notebook. I need to investigate more. 

```
[IPKernelApp] WARNING | Widget Javascript not detected.  It may not be installed properly. Did you enable the widgetsnbextension? If not, then run "jupyter nbextension enable --py --sys-prefix widgetsnbextension"
```

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: It should be able to do that. But ipython require other python packages, so conda may need to do some extra things for that. I can do that in a followup ticket.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: Fixed

Leemoonsoo:
Tested 88a1f2f and ZeppelinContext works well.

Does auto completion supposed to work?

zjffdu:
Yes, that works. It is covered in unit test (https://github.com/apache/zeppelin/pull/2474/files#diff-d9c8252d37a199d13b7157e88041c0e9R124).  Also I manually verified it

Leemoonsoo:
Is it expected completion list?
image

zjffdu:
Thanks @Leemoonsoo for catching that. This is due to I miss cursor for code completion. Just pushed another commit to fix that and add more unit test for it.

```
    // there's no completion for 'a.' because it is not recognized by compiler for now.
    String st = "a='hello'\na.";
    completions = interpreter.completion(st, st.length(), getInterpreterContext());
    assertEquals(0, completions.size());

    // define a first
    st = "a='hello'";
    result = interpreter.interpret(st, getInterpreterContext());
    assertEquals(InterpreterResult.Code.SUCCESS, result.code());

    // now we can get the completion for `a.`
    st = "a.";
    completions = interpreter.completion(st, st.length(), getInterpreterContext());
    // it is differnet for python2 and python3 and may even different for different minor version
    // so only verify it is larger than 20
    assertTrue(completions.size() > 20);

    st = "a.ca";
    completions = interpreter.completion(st, st.length(), getInterpreterContext());
    // a.capitalize
    // a.casefold
    assertEquals(2, completions.size());

    // cursor is in the middle of code
    st = "a.ca\b='hello";
    completions = interpreter.completion(st, 4, getInterpreterContext());
    // a.capitalize
    // a.casefold
    assertEquals(2, completions.size());

```

docs/interpreter/python.md
felixcheung: yes, that's a kind of interactive widget, I wouldn't think that would work

1ambda:
Let create a PR to fix the ASCII issue, we can test this PR based on the ASCII PR.

zjffdu:
@1ambda I did some investtigation on that, and puts some links in https://issues.apache.org/jira/browse/ZEPPELIN-2756, hope it would be helpful for you.

zjffdu:
Just push another commit to use IPythonInterpreter first, if it is not available, I will fallback it to the old PythonInterpreter.

zjffdu:
Push another commit to use IPython in PySparkInterpreter if IPython is available , otherwise fallback to the old implementation.

zjffdu:
Push another commit to support streaming output

zjffdu:
@Leemoonsoo @1ambda @felixcheung @prabhjyotsingh I made several follow up commit. All the features in this stage is done. Please help review that. I believe it would be a big improvement for python users.

python/src/main/resources/interpreter-setting.json
```diff
@@ -17,6 +17,29 @@
...
+    }
+  },
+  {
+    "group": "python",
+    "name": "ipython",
+    "className": "org.apache.zeppelin.python.IPythonInterpreter",
+    "properties": {
+      "zeppelin.ipython.launch.timeout": {
+        "propertyName": "zeppelin.ipython.launch.timeout",
+        "defaultValue": "30000",
```
felixcheung: this seems rather long?

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -111,6 +113,36 @@ private void createPythonScript() {
...
+    iPySparkInterpreter = null;
+
+    if (property.getProperty("zeppelin.spark.useIPython", "true").equals("true")) {
+      // don't print it when it is in testing, just for easy output check in test.
+      try {
+        InterpreterContext.get().out.write(("IPython is not available, " +
+            "use the native PySparkInterpreter\n")
+            .getBytes());
+      } catch (IOException e) {
+        LOGGER.warn(e.getMessage());
```
felixcheung: ditto

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -190,9 +222,24 @@ private Map setupPySparkEnv() throws IOException{
       }
     }
 
+    LOGGER.debug("PYTHONPATH: " + env.get("PYTHONPATH"));
     return env;
   }
 
+  // Run python shell
+  // Choose python in the order of
+  // PYSPARK_DRIVER_PYTHON > PYSPARK_PYTHON > zeppelin.pyspark.python
```
felixcheung: hmm.. should the order be
`zeppelin.pyspark.python > PYSPARK_DRIVER_PYTHON > PYSPARK_PYTHON`

usually I think of zeppelin settings overriding what is set for spark

spark/src/test/java/org/apache/zeppelin/spark/IPySparkInterpreterTest.java
```diff
@@ -0,0 +1,203 @@
...
+        "replName",
+        "paragraphTitle",
+        "paragraphText",
+        new AuthenticationInfo(),
+        new HashMap<String, Object>(),
+        new GUI(),
+        null,
+        null,
+        null,
+        new InterpreterOutput(null));
```
felixcheung: add a test for something specific from IPython? perhaps one of its magic?

testing/install_external_dependencies.sh
```diff
@@ -44,5 +44,6 @@ if [[ -n "$PYTHON" ]] ; then
   conda update -q conda
   conda info -a
   conda config --add channels conda-forge
-  conda install -q matplotlib pandasql
+  conda install -q matplotlib pandasql ipython jupyter_client ipykernel matplotlib bokeh
```
felixcheung: need `bokeh` for test?

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -219,6 +220,36 @@ private void createGatewayServerAndStartScript() throws UnknownHostException {
...
+      try {
+        iPythonInterpreter.open();
+        if (InterpreterContext.get() != null) {
+          InterpreterContext.get().out.write(("IPython is available, " +
+              "use IPython for PythonInterpreter\n")
+              .getBytes());
+        }
+        return;
+      } catch (Exception e) {
+        iPythonInterpreter = null;
```
zjffdu: There's another case that the if block is not executed.

python/src/main/resources/grpc/python/zeppelin_python.py
```diff
@@ -0,0 +1,107 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
zjffdu: Only `PyZeppelinContext` is duplicated. If IPythonInterpreter become mature enough, I think we can remove the old implementation.

python/src/main/resources/interpreter-setting.json
zjffdu: This is millisecond.

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
zjffdu: This is done in ZEPPELIN-2195.  The order should be `PYSPARK_DRIVER_PYTHON > PYSPARK_PYTHON > zeppelin.pyspark.python`. `zeppelin.pyspark.python` only affect the driver side, won't affect the executor side. So we should use the spark standard configuration. IMO, zeppelin should use spark standard configuration as possible as it can, rather than creating new configuration.

spark/src/test/java/org/apache/zeppelin/spark/IPySparkInterpreterTest.java
zjffdu: IPySparkInterpreterTest will run all the test of IPythonInterpreterTest. see L94

testing/install_external_dependencies.sh
zjffdu: yes, I test bokeh in unit test.

Leemoonsoo:
Tried last commit.

I installed jupyter, IPython, ipykernel, grpcio in my conda environment and i configured my python interpreter like
image

And then i'm getting `No module named py4j.java_gateway` error on first `%python.ipython` execution

image

And second execution, `z` is not available.

image

Looks like py4j library is not in python path. @zjffdu Do you have any idea on this?

spark/src/test/java/org/apache/zeppelin/spark/IPySparkInterpreterTest.java
felixcheung: great!

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
felixcheung: I'm not feeling strongly about this, my take is for someone migrating environment from Spark only setup, they might feel the need to override Spark only setting from within Zeppelin (different virtualenv for example) such that both Spark and Spark in Zeppelin will work at the same time with one set of environment.

IMO having PYSPARK_DRIVER_PYTHON overriding zeppelin.pyspark.python is not very useful because if zeppelin.pyspark.python is not set PYSPARK_DRIVER_PYTHON is going to be used by spark-submit anyway.

zjffdu: Zeppelin doesn't use spark-submit to launch python process, instead it launch python via its own logic. That's why I want to use `PYSPARK_DRIVER_PYTHON` first to make it consistent with `spark-submit`. Ideally `zeppelin.pyspark.python` should be removed IMO, but I have to keep it here now for compatibility. IMO zeppelin should not introduce new properties that have  the duplicated property in spark.

zjffdu:
@Leemoonsoo I have fixed the issue, could you try it again ? Thanks

zjffdu:
ping Leemoonsoo @1ambda @felixcheung @prabhjyotsingh for review

docs/interpreter/python.md
```diff
@@ -232,6 +232,70 @@ SELECT * FROM rates WHERE age < 40
 Otherwise it can be referred to as `%python.sql`
 
 
+## IPython Support
+
+IPython is more powerful than the default python interpreter with extra functionality. You can use IPython2 or IPython3 which depends on which python you set `zeppelin.python`.
```
felixcheung: `You can use IPython2 or IPython3 which depends on which python you set zeppelin.python.`
should this say `You can use IPython with Python2 or Python3 which depends on which python you set zeppelin.python.`

IPython has separate versions independent of python2 or python3

docs/interpreter/python.md
```diff
@@ -232,6 +232,70 @@ SELECT * FROM rates WHERE age < 40
...
+
+   **Pre-requests**
+   
+    - Jupyter `pip install jupyter`
+    - grpcio `pip install grpcio`
+
+If you already install anaconda, then you just need to install `grpcio` as Jupyter is already included in anaconda.
+
+Besides the basic function of python interpreter. You can use all the IPython advanced features as you use it in Jupyter Notebook.
```
felixcheung: I think this should be one sentence
`Besides the basic function of python interpreter. You`
=>
`Besides the basic function of python interpreter, you`
or\
 =>
`In addition to all basic functions of the python interpreter, you`

python/README.md
```diff
@@ -50,3 +50,22 @@ mvn -Dpython.test.exclude='' test -pl python -am
...
+ * jupyter 5.x
+ * IPython
+ * ipykernel
+ * grpcio
+ 
+If you have installed anaconda, then you just need to install grpc.
+
+# IPython Architecture
+Current interpreter delegate the whole work to ipython kernel via `jupyter_client`. Zeppelin would launch a python process which host the ipython kernel.
+Zeppelin interpreter process will communicate with the python process via `grpc`. Ideally every feature works in IPython should work in Zeppelin as well.
```
felixcheung: as commented before `Ideally every feature works in IPython should work in Zeppelin as well.` is kinda broad

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -219,6 +220,37 @@ private void createGatewayServerAndStartScript() throws UnknownHostException {
...
+    iPythonInterpreter = null;
+
+    try {
+      if (InterpreterContext.get() != null) {
+        InterpreterContext.get().out.write(("IPython is not available, " +
+            "use the native PythonInterpreter\n")
+            .getBytes());
+      }
+    } catch (IOException e) {
+      LOG.warn("Fail to write InterpreterOutput", e.getMessage());
```
felixcheung: `LOG.warn("Fail to write InterpreterOutput", e);`

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -111,6 +113,37 @@ private void createPythonScript() {
 
   @Override
   public void open() {
+    // try IPySparkInterpreter first
+    iPySparkInterpreter = getIPySparkInterpreter();
```
felixcheung: save the value of `property.getProperty("zeppelin.spark.useIPython", "true").equals("true")` instead of calling it twice?

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -111,6 +113,37 @@ private void createPythonScript() {
...
+    iPySparkInterpreter = null;
+
+    if (property.getProperty("zeppelin.spark.useIPython", "true").equals("true")) {
+      // don't print it when it is in testing, just for easy output check in test.
+      try {
+        InterpreterContext.get().out.write(("IPython is not available, " +
+            "use the native PySparkInterpreter\n")
+            .getBytes());
+      } catch (IOException e) {
+        LOGGER.warn("Fail to write InterpreterOutput", e.getMessage());
```
felixcheung: `LOGGER.warn("Fail to write InterpreterOutput", e);`

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -190,9 +223,24 @@ private Map setupPySparkEnv() throws IOException{
...
 
+    LOGGER.debug("PYTHONPATH: " + env.get("PYTHONPATH"));
     return env;
   }
 
+  // Run python shell
+  // Choose python in the order of
+  // PYSPARK_DRIVER_PYTHON > PYSPARK_PYTHON > zeppelin.pyspark.python
+  public static String getPythonExec(Properties properties) {
+    String pythonExec = properties.getProperty("zeppelin.pyspark.python", "python");
```
felixcheung: what if `zeppelin.python` is set?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/InterpreterOutputStream.java
```diff
@@ -78,7 +78,7 @@ public void write(byte [] b, int off, int len) throws IOException {
 
   @Override
   protected void processLine(String s, int i) {
-    logger.debug("Interpreter output:" + s);
+//    logger.debug("Interpreter output:" + s);
```
felixcheung: this is commented out?

zeppelin-server/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -141,6 +141,7 @@ public Boolean apply(WebDriver d) {
       fail();
     }
 
+    driver.manage().window().maximize();
```
felixcheung: is this related to this PR?

python/README.md
zjffdu: Yeah, I know, but I am afraid it's almost impossible to verify each feature of IPython in  zeppelin. I would wait for more feedback from community

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
zjffdu: opps, this is a bug in the second if block. I should check `InterpreterContext.get() != null`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/InterpreterOutputStream.java
zjffdu: Fixed

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
zjffdu: Fixed

zeppelin-server/src/test/java/org/apache/zeppelin/WebDriverManager.java
zjffdu: Yes, this cause flaky selenium test.

docs/interpreter/python.md
zjffdu: Fixed

docs/interpreter/python.md
zjffdu: Fixed

spark/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
zjffdu: `zeppelin.python` is only for python interpreter.  And `zeppelin.pyspark.python` is for pyspark interpreter.

zjffdu:
I will merge it if no more comments

PR #2479, opened 2017-07-10
[ZEPPELIN-2762] Use regex to retrieve interpreter/repl name, and to retrieve property key; add and fix tests

dabaitu:
### What is this PR for?
Make logic to parse interpreter or repl name more robust by using regex.
Make logic to parse property name in JDBC interpreter more robust by using regex.
Add tests for cases missed before that the regexes would catch now.
Fix failing tests in Paragraph and JDBCInterpreter.

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2762

Appropriate unit tests should be added to fully test regexes. They are added as part of this PR.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
@@ -70,6 +70,18 @@ public class ParagraphTest {
 
     text = "%table 1234567";
     assertEquals("1234567", Paragraph.getScriptBody(text));
+
+    text = "%presto (testPropertyKey) 1234567";
+    assertEquals("(testPropertyKey) 1234567", Paragraph.getScriptBody(text));
+
+    text = "%spark (1234567)";
+    assertEquals("(1234567)", Paragraph.getScriptBody(text));
+
+    text = "%jdbc(presto) (testPropertyKey) 1234567";
+    assertEquals("(testPropertyKey) 1234567", Paragraph.getScriptBody(text));
+
+    text = "%jdbc(presto) (testPropertyKey) (select 1234567)";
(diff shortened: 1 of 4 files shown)


zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
```diff
@@ -70,6 +70,18 @@ public void scriptBodyWithReplName() {
...
+    assertEquals("(testPropertyKey) 1234567", Paragraph.getScriptBody(text));
+
+    text = "%spark (1234567)";
+    assertEquals("(1234567)", Paragraph.getScriptBody(text));
+
+    text = "%jdbc(presto) (testPropertyKey) 1234567";
+    assertEquals("(testPropertyKey) 1234567", Paragraph.getScriptBody(text));
+
+    text = "%jdbc(presto) (testPropertyKey) (select 1234567)";
+    assertEquals("(testPropertyKey) (select 1234567)", Paragraph.getScriptBody(text));
```
zjffdu: IIUC, the scriptBody should be `(presto) (testPropertyKey) (select 1234567)` which would be passed to JdbcInterpreter. Otherwise JdbcInterpreter don't know this is `presto`. \cc @prabhjyotsingh

dabaitu: I think "presto" would be treated as the replName which should be stripped.

There was a syntax change from %jdbc(replName) to %replName. So replName is figured out first then by looking at interpreter configs, it figures out that replName belong to jdbc group.

My change uses regex, and would also make the old syntax work in the new version.

dabaitu:
Ping for review.
This regex will avoid a lot of bugs...

PR #2480, opened 2017-07-11
[ZEPPELIN-1249] build all submodules w/ the same scala version

herval:
### What is this PR for?
This uses the same Scala version throughout modules, allowing us to use only Scala 2.11.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1249
https://issues.apache.org/jira/browse/ZEPPELIN-2681

### Questions:
No

No

No

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -27,7 +27,7 @@
     </parent>
 
     <groupId>org.apache.zeppelin</groupId>
-    <artifactId>zeppelin-cassandra_2.10</artifactId>
+    <artifactId>zeppelin-cassandra_${scala.binary.version}</artifactId>
     <packaging>jar</packaging>
     <version>0.8.0-SNAPSHOT</version>
     <name>Zeppelin: Apache Cassandra interpreter</name>
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -34,6 +34,7 @@
     <description>Zeppelin cassandra support</description>
 
     <properties>
+        <cassandra.scala.version>${scala.version}</cassandra.scala.version>
         <cassandra.driver.version>3.0.1</cassandra.driver.version>
         <snappy.version>1.0.5.4</snappy.version>
         <lz4.version>1.3.0</lz4.version>
@@ -91,19 +92,19 @@
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-compiler</artifactId>
-            <version>${scala.version}</version>
+            <version>${cassandra.scala.version}</version>
         </dependency>
 
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-library</artifactId>
-            <version>${scala.version}</version>
+            <version>${cassandra.scala.version}</version>
         </dependency>
 
         <dependency>
             <groupId>org.scala-lang</groupId>
             <artifactId>scala-reflect</artifactId>
-            <version>${scala.version}</version>
+            <version>${cassandra.scala.version}</version>
             <scope>runtime</scope>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -84,8 +84,8 @@
 
   <properties>
     <!-- language versions -->
-    <scala.version>2.10.5</scala.version>
-    <scala.binary.version>2.10</scala.binary.version>
+    <scala.version>2.11.7</scala.version>
+    <scala.binary.version>2.11</scala.binary.version>
     <scalatest.version>2.2.4</scalatest.version>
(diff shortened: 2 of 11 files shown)


herval:
Regarding verification, how do you suggest I go about that? We've been using this setup internally for some time, but we only use a few of the interpreters (spark, scalding, jdbc, md) - not sure if that could lead to any sort of runtime issue on other interpreters, but I find it highly unlikely..

Can you clarify on the flexibility bit? Not sure I understand what's the benefit of using different versions at the same time (not to mention Scala 2.10 is EOL since 2016).

felixcheung:
I realize 2.10 is EOL but as mentioned, from a while back many platforms are still running only on 2.10. Even Spark still supports 2.10 as of the latest release 2.2.0 from yesterday. The reason is many enterprises have applications that cannot be easily migrated to 2.11.

It is possible that some users are only using a small number of interpreters (as you do) and they need a particular scala version. So instead of forcing all submodule to always use the same scala version, how about set a global scala version as the default but each submodule still has its own spark.scala.version that can be set in the maven command line when building?

herval:
sounds good, I’ll do that then

_____________________________

cassandra/pom.xml
```diff
@@ -27,13 +27,14 @@
     </parent>
 
     <groupId>org.apache.zeppelin</groupId>
-    <artifactId>zeppelin-cassandra_2.10</artifactId>
+    <artifactId>zeppelin-cassandra_${scala.binary.version}</artifactId>
```
felixcheung: here for instance, should artifactId match the actual version of scala used, ie. `cassandra.scala.version`?

pom.xml
```diff
@@ -84,8 +84,8 @@
 
   <properties>
     <!-- language versions -->
-    <scala.version>2.10.5</scala.version>
-    <scala.binary.version>2.10</scala.binary.version>
+    <scala.version>2.11.7</scala.version>
```
felixcheung: 2.11.8? that's the one used by spark, https://github.com/apache/spark/blob/master/pom.xml#L158

cassandra/pom.xml
herval: I'm not sure this should matter:

- If you're downloading the lib from maven, and use Scala 2.11, then all submodules would be 2.11 (as expected)

- If you're building from source, the artifact names would retain your _base_ project scala version number, which wouldn't really affect compatibility with anything.

Other than naming the artifacts, I don't see any other place where binary.version is used/useful.

WDYT? Let me know if you think these should match the internal version for any reason, though, and I can copy+pasta some more

pom.xml
herval: fixed

cassandra/pom.xml
felixcheung: > and use Scala 2.11, then all submodules would be 2.11 (as expected)

I guess this is something we need to enforce when making the release

herval: right - right now, it's being enforced to "part as 2.10, part as 2.11". I see the "binary version" as essentially the "global version", as it's used for naming - doesn't disallow you to build from source, internally, and use "Spark 1.6 w/ Scala 2.10.5"

thoughts? Again, I can make the binary.version per-package as well if you prefer, just don't feel it's really necessary

felixcheung: anyone else has any preference on this?
@Leemoonsoo ?

Leemoonsoo: I don't have preference, but doesn't [dev/change_scala_version.sh](https://github.com/apache/zeppelin/blob/master/dev/change_scala_version.sh) take care of artifact name?

felixcheung:
hmm, I did a quick check, eg, for beam it's still building against 2.10 https://github.com/apache/beam/blob/master/runners/spark/pom.xml#L229
Same for Ignite https://github.com/apache/ignite/blob/master/pom.xml#L521

zjffdu:
IMO, it is not so urgent to do this. Interpreters and zeppelin server run in different process, so theoretically it is not necessary to build them together. After we move interpreters out of zeppelin project, most of problems can be addressed.

herval:
@zjffdu what's the benefit of *not* doing it, given it's a simple change?

zjffdu:
@herval This is an improvement PR, not bug fix PR. And based on the discussion above, it may also introduce some other issues or overhead for users (e.g. some interpreters don't support scala 2.11).
Besides, after we move out interpreters from zeppelin project, scala version should not be an issue anymore. Let me know your concern and if I miss anything. Thanks

PR #2481, opened 2017-07-11
[HOTFIX][ZEPPELIN-2760] fix JDBC regression caused after ZEPPELIN-2698

prabhjyotsingh:
### What is this PR for?
This is WRT to https://github.com/apache/zeppelin/commit/e1f0a3205eb5aede0b2d80c9d3de59b3f47b699c#commitcomment-23016062

> This caused a regression:
https://github.com/apache/zeppelin/blame/master/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java#L212
Leads to:
java.lang.NoClassDefFoundError: org/apache/hadoop/security/UserGroupInformation$AuthenticationMethod
	at org.apache.zeppelin.jdbc.security.JDBCSecurityImpl.getAuthtype(JDBCSecurityImpl.java:64)
	at org.apache.zeppelin.jdbc.JDBCInterpreter.isKerboseEnabled(JDBCInterpreter.java:213)

### What type of PR is it?
[Bug Fix | Hot Fix]

### What is the Jira issue?
* [ZEPPELIN-2760](https://issues.apache.org/jira/browse/ZEPPELIN-2760)

Check JDBC interpreter without any Kerberos setting 

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -210,9 +210,11 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
 
   protected boolean isKerboseEnabled() {
-    UserGroupInformation.AuthenticationMethod authType = JDBCSecurityImpl.getAuthtype(property);
-    if (authType.equals(KERBEROS)) {
-      return true;
+    if (!isEmpty(property.getProperty("zeppelin.jdbc.auth.type"))) {
+      UserGroupInformation.AuthenticationMethod authType = JDBCSecurityImpl.getAuthtype(property);
+      if (authType.equals(KERBEROS)) {
+        return true;
+      }
     }
     return false;
   }


prabhjyotsingh:
@herval Thank you for pointing this, would you like to review/test this PR as well?

herval:
would it be worth adding a test around this? I imagine the tests would fail with a "class not found", so it would be easy to add one that proves that you don't get the error if auth.type is empty

prabhjyotsingh:
@herval The reason this exception didn't get caught with any of the existing test cases is because this particular dependency (for UserGroupInformation) is marked as `provided` (https://github.com/apache/zeppelin/blob/master/jdbc/pom.xml#L106) and while executing any of the test cases it gets resolved but doesn't get bundled.

herval:
I actually don't see any test on `JdbcInterpreterTest` covering that code path at all - I think the test suite is passing simply because that bit is not tested, and feel it should be

prabhjyotsingh:
Well, any test case calling `open()` will hit `isKerboseEnabled()`, and in `JdbcInterpreterTest` when ever `isKerboseEnabled` function is called will returning false all the time.


Ref: https://github.com/apache/zeppelin/blob/master/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java#L173
https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/KerberosInterpreter.java#L64

herval:
oh, Indeed. There's some inheritance there 😶

prabhjyotsingh:
@herval  and @felixcheung Thank you for the review. Will merge this if no more discussion.

PR #2485, opened 2017-07-12
[ZEPPELIN-2758]. NumberFormatException on importing notebook

zjffdu:
### What is this PR for?

This is due to the custom TypeAdapter for Paragraph. I don't know why we introduce it. To me it is not necessary. So I remove this in this PR. And the spark UI still can be accessed in frontend. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2758

Unit test is updated and manually tested

### Questions:


zjffdu:
@karuppayya I remove `NotebookTypeAdapterFactory.java` which I think is not necessary. Let me know your concern.

necosta:
LGTM

zjffdu:
Will merge it if no more comments.

karuppayya:
@zjffdu the `NotebookTypeAdapterFactory ` was added to avoid `runtimeInfos` being persisted in json. (These infos are valid only during the runtime of the Zeppelin server, interpreter)   
How are we handling this now?

zjffdu:
Thanks @karuppayya So do you mean Note will be serialized to different formats for runtime serialize to frontend and serialized to file ? What is the side effect if I store it in file ?

karuppayya:
Currently only the frontend receives this info.
If we remove the AdapterFactory, The `runtimeInfos` will be serialised to json file.
Say runtimeinfos contain spark UI link.  
if it is persisted in json, when Zeppelin server restarts, we will see a invalid spark UI link(we can clean the runtimeinfos during zep server initialization though), which we might not want

zjffdu:
Thanks @karuppayya for explaining. How about reset the `runtimeInfos` to null when deserialized from json (it happens when it is read from file). Serializing to different formats in different scenarios seems a little complicated to me

karuppayya:
Yeah , we can reset the infos when the json is deserialized.

zjffdu:
Thanks @karuppayya push another commit to reset the runtimeInfo when deseriliazed from json.

zjffdu:
Will merge it if no more comments.

prabhjyotsingh:
LGTM!

PR #2486, opened 2017-07-12
[ZEPPELIN-2768]. Bump up Spark version to 2.2.0

zjffdu:
### What is this PR for?
Spark 2.2.0 is just released, this PR is to support spark 2.2.0. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2768

Travis is updated

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,8 @@ matrix:
     # Several tests were excluded from this configuration due to the following issues:
     # HeliumApplicationFactoryTest - https://issues.apache.org/jira/browse/ZEPPELIN-2470
     # After issues are fixed these tests need to be included back by removing them from the "-Dtests.to.exclude" property
-    - jdk: "oraclejdk7"
-      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
+    - jdk: "oraclejdk8"
+      env: SCALA_VER="2.11" SPARK_VER="2.2.0" HADOOP_VER="2.6" PROFILE="-Pspark-2.2 -Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
 
     # Test selenium with spark module for 1.6.3
     - jdk: "oraclejdk7"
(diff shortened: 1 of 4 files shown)


zjffdu:
@Leemoonsoo @jongyoul @felixcheung Please help review it.

.travis.yml
```diff
@@ -50,8 +50,8 @@ matrix:
     # Several tests were excluded from this configuration due to the following issues:
     # HeliumApplicationFactoryTest - https://issues.apache.org/jira/browse/ZEPPELIN-2470
     # After issues are fixed these tests need to be included back by removing them from the "-Dtests.to.exclude" property
-    - jdk: "oraclejdk7"
-      env: SCALA_VER="2.11" SPARK_VER="2.1.0" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
+    - jdk: "oraclejdk8"
```
zjffdu: spark 2.2.0 use jdk8

1ambda: Then, what about using trusty If we are not using jdk7. It's faster and provides more memory. 

- https://github.com/1ambda/zeppelin/blob/ad08a3bc0da420f7a7b4771d1aee8bfdd962efa7/.travis.yml#L53-#L62
- https://github.com/travis-ci/travis-ci/issues/7019

felixcheung:
I think Jdk 7 is supported by Spark, before the 2.2.0 release.
We can choose not to test with it in Travis. Not sure if we need to make sure we test with jdk 7 before making a release.

.travis.yml
Leemoonsoo: Does it mean binary package need to be built with JDK8 and also running Zeppelin requires JDK8 ?

felixcheung:
For Spark 2.2.0, yes...
Zeppelin could run a different jdk version? The interpreter jvm hosting spark would need to be 8.

.travis.yml
1ambda: image

zjffdu:
One workaround for spark 2.2 is that user could specify JAVA_HOME in its interpreter setting. I have verified it. 

image

For JDK 8 support of Zeppelin, I think we could do it in another PR if it is OK to upgrade to JDK8 for all the components of Zeppelin. We may need to discuss it in community first.

Leemoonsoo:
LGTM.
I think we can have separate discussion for JDK version support, thanks for the explain.

zjffdu:

spark/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
```diff
@@ -34,10 +34,10 @@
   public static final SparkVersion SPARK_1_6_0 = SparkVersion.fromVersionString("1.6.0");
 
   public static final SparkVersion SPARK_2_0_0 = SparkVersion.fromVersionString("2.0.0");
-  public static final SparkVersion SPARK_2_2_0 = SparkVersion.fromVersionString("2.2.0");
+  public static final SparkVersion SPARK_2_3_0 = SparkVersion.fromVersionString("2.3.0");
 
   public static final SparkVersion MIN_SUPPORTED_VERSION =  SPARK_1_0_0;
-  public static final SparkVersion UNSUPPORTED_FUTURE_VERSION = SPARK_2_2_0;
+  public static final SparkVersion UNSUPPORTED_FUTURE_VERSION = SPARK_2_3_0;
```
bikassaha: Why do we need something like this. Unless Spark breaks things, Zeppelin should continue to work without needing a new build of Zeppelin. This prevents people from easily using Zeppelin with a potentially compatible version of Spark.

zjffdu: This is due to historical reason that zeppelin will check the supported version. If user want to try some spark version that is not officially supported by zeppelin, he can set `zeppelin.spark.enableSupportedVersionCheck` to `false`.

surlypants: can/how said check be disabled in zeppelin-env.sh ?

zjffdu: NO, you need to set it in interpreter setting.

1ambda:
@zjffdu Can we support 2.2 for the release 0.7.3 (not built-in because beginner might see errors with JDK 7 when running tutorial)

image

- https://issues.apache.org/jira/browse/ZEPPELIN-2864

zjffdu:
Yes, I think so. I always build zeppelin with JDK 8 in my local box.

PR #2496, opened 2017-07-18
[HOTFIX] fix 'checkbox' type for create iterpreter form

tinkoff-dwh:
### What is this PR for?
Not visible checkboxes in form to create new interpreter. 

### What type of PR is it?
Bug Fix

before
before

after
after

### Questions:


1ambda:
@tinkoff-dwh Thanks for the contribution.

Tested locally and works well. Merge if no more discussion.

PR #2502, opened 2017-07-24
[ZEPPELIN-2807] Passing Z variables to SQL Interpreter (One part of ZEPPELIN-1967)

sanjaydasgupta:
### What is this PR for?
This feature allows embedding/interpolating Z variables into SQL statements passed to the Spark-SQL interpreter. It implements one part of the functions desired by the pre-existing issue (ZEPPELIN-1967)

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2807

A new test has been added to the unit tests. The attached screenshot also shows tests of the functionality.

sparksqlchange

### Questions:
* Does this needs documentation? Yes, a small update to any existing documentation is ideal.


.travis.yml
```diff
@@ -17,6 +17,8 @@ language: java
 
 sudo: false
 
+dist: precise
```
felixcheung: is there any reason for having this?

spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java
```diff
@@ -161,6 +163,41 @@ public void test_null_value_in_row() {
...
+    InterpreterContext newContext = new InterpreterContext("NoteId", "ParaId", null,
+            "testVariableInterpolation", "",
+            null, null, null, null,
+            new LocalResourcePool("testVariableInterpolation"),
+            null, null);
+    SparkZeppelinContext zc = getSparkInterpreter().getZeppelinContext();
+    zc.setInterpreterContext(newContext);
+    zc.put("n", "100");
+    zc.put("table", "name");
+    String commandWithVariables = "select * from {table} where count = {n}";
```
felixcheung: we would need to add more test for the regex handling code. for example,
`{}` or `{n`

to start with

sanjaydasgupta:
Thank you for the observations @felixcheung.
Will review these points and make suitable changes.

spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java
zjffdu: how to escape these special character. e.g. 
```
select * from {table} where count = '{n}'
```
Here `{n}` is a string, will you still replace it using ZeppelinContext ?

.travis.yml
sanjaydasgupta: The build was [failing on travis](https://travis-ci.org/sanjaydasgupta/zeppelin/builds/256845905) without it, and there was a comment in the travis mail suggesting this course of action.

sanjaydasgupta:
The choice of {...} for identifying parameter/variable names comes from the parent issue [ZEPPELIN-1967](https://issues.apache.org/jira/browse/ZEPPELIN-1967). Though #{...} seems more widely used in other settings and languages, I have no opinion in the matter, and would like to request the original issue's author @Tagar to offer any comments. Thanks for your ideas and help with this.

spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java
Tagar: Great catch. Escaping should be done similarly like in Jupyter.
Since this will be more similar for users coming to Zeppelin from Jupyter world.
With double curly braces –
  {{ will be escaped to single left curly brace - {
  }} will be escaped to single right curly brace - }
Thanks!

Tagar:
@sanjaydasgupta great job on getting this done.
ZEPPELIN-1967 was inspired by Jupyter so I just we should do the same as it is done there.
So it's less confusing for users who are switching from Jupyter.
I would suggest not to use #{...} as it'll be confusing to use this naming convention in shell interpreter, for example.
Escaping should be done similarly like in Jupyter.
With double curly braces –
{{ will be escaped to single left curly brace - {
}} will be escaped to single right curly brace - }
Thanks!

sanjaydasgupta:
I would like your views @felixcheung @zjffdu @Tagar on the following approach to escaping and regex errors etc:

I wish to suggest that no escaping is needed, as the substitution code kicks in only when 2 conditions are met: (1) the pattern {...} exists in the command, and (2) the variable named in {...} actually has a mapping in Z. If the second condition is not met, because a prior z.put(...) has not been executed for the same variable name, the original command including the {...} is left unchanged. Since the variable name used is expected to be meaningful, and is entirely under the end-user's control, it is always possible to avoid the need for escaping. Leaving the original command unchanged has other utilities as shown in the examples below. 

**Setup**
```
val df = spark.createDataFrame(Seq((1, "one"), (3, "three"), 
    (Int.MaxValue / 2, "##"), (Int.MaxValue, "####"))).toDF("id", "name")
df.createOrReplaceTempView("name")

z.put("n", 3)
z.put("table", "name")
```
**Trivial Case**
The following SQL command works _normally_ -- performing variable substitution as expected.
`%sql select * from {table} where id = {n}`
The output has one row: (3, three)

**Unchanged Pattern-Like Construction**
But the following command also works -- with the pattern-like structure (which is actually a regex passed to _regex_) left unchanged. 
`%sql select * from {table} where name rlike '#{2}'`
The output has 2 rows: (1073741823, ##) and (2147483647, ####)

Note that allowing substitution patterns inside quotes is an advantage, giving us a string-interpolation facility like in Scala (and other languages). Again, escaping is not required here also as long as the content of {...} is judiciously chosen.

**One Last Example**
The following example has two {...} structures, the first is part of a regex, the second is a real pattern.
```
z.put("what", "e")
%sql select * from {table} where name rlike '^.{3}{what}.$'
```
The output has one row: (3, three)

Incomplete patterns are also left unchanged, this could be an advantage for future embedded mini-languages in the same scope.

Would like to hear your views on the this proposal to not implement escapes for embedded variable patterns. I only started looking at Zeppelin internals about 10 days back, so your insight will be very helpful.

felixcheung:
I like what you propose, but think it might be very confusing for user if behavior changes depending on whether the variable is in the z context. for instance, it could be a shared notebook and someone might have set `what` and that would changes the outcome of the sql query example you have

sanjaydasgupta:
Thank you for the insight. Do you think we may retain variable substitution within literal strings as in the following contrived example?

`%sql select * from spares where part_no = 'pump-{manufacturer-code}'`

Thanks again for your help.

spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java
sanjaydasgupta: Can you direct me to some documentation on the Jupyter feature? I'd like to understand how it handles errors too: unknown variable name, non-matching { and }, and possibly others. 

Thanks for your help.

felixcheung:
I guess that's tricky.
like in your example, if it is in `'`, should we leave it alone, or not? `%sql select * from table where name rlike '#{2}'`

I suppose we should first make sure the behavior is consistent in all interpreters and in all/most of token replacement

sanjaydasgupta:
Agreed, thanks.

spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java
Tagar: @sanjaydasgupta in Jupyter that feature is actually more flexible - that gets eval()ed as Python code. So I can have {var1+100} for example there. I can't find documentation on this feature of Jupyter, but I just ran {non_existent_var} and it does not show errors, it actually returns "{non_existent_var}" as is not substituted at all. I think that's in line with what you were suggesting earlier. See screenshot below. But I agree with @felixcheung  it might be less confusing for users just to show error if a variable is not found.

image

sanjaydasgupta: Thanks @Tagar. Interestingly, Jupyter's implementation substitutes within strings too, see the following example.
jupyter

Tagar: Yep. It would be awesome to have that in Zeppelin too.

sanjaydasgupta:
Hello @felixcheung @Tagar @zjffdu, this is for any further comments and suggestions. Here are the features implemented:

1. - The substitution pattern is `{ ... }` as suggested for uniformity with Jupyter
2. - The substitution pattern can be escaped by using `{{ ... }}`, and such patterns are translated into `{ ... }` while retaining the original contents of `{{ ... }}`
3. - Substitution patterns inside strings (like `'part-{number}'`) are active, and will be substituted. This is also as in the Jupyter feature (@felixcheung this is different from our last exchange)

I have expanded the unit-tests to include these cases, and have added error checks for a variety of ill-formed patterns. The following screenshots show some examples of the feature in use.

fig-01
fig-02
fig-03
fig-04

Thanks for all your pointers and ideas.

felixcheung:
I think it's fine to replace token within quoted string - this is probably a better, more consistent behavior anyway.

I am concerned with having different ways to replace token in different interpreters though, as pointed out in https://github.com/apache/zeppelin/pull/2502#pullrequestreview-52283678; as adding to that we also have the `${token}` already https://zeppelin.apache.org/docs/0.7.2/manual/dynamicform.html#text-input-form

Can you point me to how this is used in jupyter?

sanjaydasgupta:
Digging a little deeper to clarify the conflict ... The class `BaseZeppelinContext` uses two different key-value repositories for two different needs:

1. **Storage of user-provided inputs entered into dynamic form fields**. These are held in an `AngularObjectRegistry` associated with the particular interpreter in effect, and are managed either (a) programmatically by the functions `z.angularBind(...)` and `z.angularUnbind(...)`, or (b) by using the `${ ... }` notation described at the link you provided (text-input-form) -- this second option is the one that seems to be (but in fact is not) in conflict with the substitution proposal described here.
2. **Programmatically assigned data intended to be shared across interpreters** (currently Spark, R, and Python). These are held in a `ResourcePool`, and are managed by the functions `z.put(...)`, and `z.get(...)`. This pull request seeks to substitute values from a `ResourcePool` into `{ ... }` patterns in interpreter commands.

So, as described above, these are two different facilities, and use similar but different substitution patterns: the dynamic-forms facility uses `${ ... }` while this pull request (inter-interpreter data sharing facility) proposes to use `{ ... }` and `{{ ... }}` (for escaping the basic substitution pattern).

The Jupyter Python examples quoted are drawn from the context of shell commands spawned by prefixing a command with `!` operator. In this case the { ... } pattern can be used to embed Python snippets (not just a variable name) into the command, and `{{ ... }}` is used for escaping the `{ ... }` substitution pattern. 

I hope this clarifies the two patterns and uses.

sanjaydasgupta:
I can have a go at creating/updating the documentation if someone can help me with any guideance on where/how to begin.

My familiarity with Jupyter is very limited, so while I will dig around, I would also request @Tagar to help if possible.

Tagar:
Info from Jupyter comitter @takluyver 

> It's fairly brief, but here:
> http://ipython.readthedocs.io/en/stable/interactive/reference.html#manual-capture-of-command-output-and-magic-output
>  
> E.g. found that quoting of curly brackets is done using {{..}} etc.
> If this documentation doesn't exist (I can't find where this is documented), please point to code 
> that handles this logic.
> 
> The expansion of variables and expressions is done by the var_expand() method:
> https://github.com/ipython/ipython/blob/8f50a3771614d3e22505eabc9332d1bc44af6f0e/IPython/core/interactiveshell.py#L3037
> 
> Running a shell command and capturing the output is the getoutput() method:
> https://github.com/ipython/ipython/blob/8f50a3771614d3e22505eabc9332d1bc44af6f0e/IPython/core/interactiveshell.py#L2279
> 
> The escaped double curly braces {{...}} is part of Python's standard string formatting machinery, which we use in var_expand().
>

sanjaydasgupta:
@felixcheung I have added a couple of paragraphs of documentation including an example showing interaction between the Scala and SQL interpreters to _spark.md_ (at the end of the text under the title [Object Exchange](https://zeppelin.apache.org/docs/0.7.2/interpreter/spark.html#object-exchange) and just above the title [Form Creation](https://zeppelin.apache.org/docs/0.7.2/interpreter/spark.html#form-creation).

docs/interpreter/spark.md
```diff
@@ -343,6 +343,38 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
...
+
+  </div>
+  <div data-lang="sql" markdown="1">
+
+{% highlight sql %}
+# Inline interpolate into SQL command
+%spark.sql
+select * from {tableName}
+
+  </div>
```
felixcheung: in this case, this might not be the best way to document in multiple language - I think this way will result in scala and sql in its own tab within one code block in the generated doc? whereas in this case we might prefer to show them in sequence in explain how to flow from one to another

sanjaydasgupta: I see your point @felixcheung and have changed the presentation accordingly. There are also a few other small changes of wording that hopefully make things clearer.

Can you take another look please. Thanks for your views on this.

sanjaydasgupta:
Please let me know if there are any further comments or concerns @felixcheung. I have addressed the suggestion on structure of the code in the documentation ([here](https://github.com/apache/zeppelin/pull/2502/files/9b7673b7048992cbd07ccaad300abd16b4021521))

Leemoonsoo:
Testing this branch a bit and found there's an edge case with new line

```
select
*
from {table}
limit 10
````

will throw error

```
cannot recognize input near '{' 'table' '}' in join source; line 3 pos 5
set zeppelin.spark.sql.stacktrace = true to see full stacktrace
```

**Comment by sanjaydasgupta** [comment]
Thanks for the catch # @Leemoonsoo 

I have fixed and committed the issue.

**Comment by Leemoonsoo** [comment]
@sanjaydasgupta Thanks for addressing newline issue.

If we take a look dynamic form, each interpreter can choose whether interpreter wants to use Dynamic form or not by overriding [Interpreter. getFormType()](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java#L102).

Do you think this feature can be generalized and each interpreter will able to choose to enable/disable this feature, in the future?

**Comment by sanjaydasgupta** [comment]
I had not thought about that @Leemoonsoo, thanks for alerting me to it.

As the code now stands (in relation to [Interpreter.java](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java) and it's concrete child classes) object interpolation is not automatically applied to the command string when [interpret(...)](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java#L86) is called. So, it is as if the feature is off by default, and each interpreter can opt in by adding a few lines of code to its overridden version of `interpret(...)` for calling the command interpolating function [SparkSqlInterpreter.interpolateVariable(...)](https://github.com/sanjaydasgupta/zeppelin/blob/906b698187bf5d4f358cb1263bdb15bfb0b286f4/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java#L92). 

In the long run (when more interpreters use this feature), the interpolating function [SparkSqlInterpreter.interpolateVariable(...)](https://github.com/sanjaydasgupta/zeppelin/blob/906b698187bf5d4f358cb1263bdb15bfb0b286f4/spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java#L92) could be moved up into `Interpreter.java`, and the command interpolation logic could be applied automatically -- conditionally upon the return value of a function like [Interpreter. getFormType()](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java#L102).

My feeling is that fairly extensive refactoring affecting all the existing interpreter classes will be needed when the changes in the previous paragraph are made. But I can take that path if you think it is preferable.

**Comment by Leemoonsoo** [comment]
Thanks, @sanjaydasgupta for sharing your thoughts.
I just wanted to think about possibility generalization of this feature for the future.

LGTM

File: spark/src/test/java/org/apache/zeppelin/spark/SparkSqlInterpreterTest.java [hunk 4]
```diff
@@ -161,6 +163,74 @@ public void test_null_value_in_row() {
...
+    resultString = sql.interpolateVariable(commandWithVariables);
+    assertEquals("ERROR:bad pattern '{{2}'", resultString);
+
+    // Error due to bad pattern (3) ...
+    commandWithVariables = "select * from names where name rlike 'a{{{2}}}'";
+    resultString = sql.interpolateVariable(commandWithVariables);
+    assertEquals("ERROR:bad pattern '{{{2}}}'", resultString);
+
+    // Error due to unpaired { ...
+    commandWithVariables = "select * from names where name = {alpha";
```
felixcheung: might be very similar to this, could you add a test for `something}`?

felixcheung: not sure if it matter, how about one `{{}}` ie not inside a `'`?

spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -114,7 +170,11 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
       // to    def sql(sqlText: String): DataFrame (1.3 and later).
       // Therefore need to use reflection to keep binary compatibility for all spark versions.
       Method sqlMethod = sqlc.getClass().getMethod("sql", String.class);
-      rdd = sqlMethod.invoke(sqlc, st);
+      String interpolateResult = interpolateVariable(st);
+      if (interpolateResult.startsWith("ERROR:"))
```
felixcheung: using a String as status return is a bit odd, how about returning a bool and the result separately?

sanjaydasgupta:
Allow me to take your suggestions from last to first @felixcheung 

I have tested a variant of the function [interpolateVariable(...)](https://github.com/apache/zeppelin/pull/2502/files/906b698187bf5d4f358cb1263bdb15bfb0b286f4#diff-851965d16af3e64741e7375e5bf9b6c5R92) to return a `scala.Tuple2[Bool, String]` (with the Bool indicating success/failure, and the String containing either the interpolated command or the error message). The resulting code is elegant and intuitive, but could be a stumbling block for those not familiar with Scala. Please let me know if it is ok to add a bit of Scala in this way.

After thinking hard about the implications (to other interpreters) of assumptions we make now about the legal forms of the use of '{' and '}', I tried to seek inspiration from Jupyter's implementation of a similar feature (when spawning a shell command using !). The following clear cases exist:

1) any `{varname}` (formed with exactly one '{' and '}') is interpolated into the command -- only if `varname` is defined. If `varname` is not defined, no error is reported, and the pattern is left as is -- to be handled by the spawned shell command.

2) any `{{something}}` (formed with exactly two '{' and '}' characters) is interpolated as `{something}`. There is no change in this behavior even when `something` has a defined value.

3) All other occurrences of '{' and '}' (singly or in groups) are treated as ordinary characters, and passed through unchanged to the underlying shell command.

Considering all the different usages of '{' and '}' that there might be in many different interpreters (to which we extend this feature progressively), it may be reasonable to follow something like the above convention from Jupyter. 

Please let me know your views on these. Thanks for your time and help with this.

sanjaydasgupta:
Thanks for the suggestions @felixcheung 

I was not able to find more documentation on Jupyter's feature, but tried it with many different variants. There are some behaviors that had not been found earlier, so I have added it to the implementation and the documentation. I also gave the documentation section a more distinctive title.

One of the things I learned about Jupyter's implementation is that it never intercepts an error case. All command strings are passed through to the underlying command -- with or without interpolation. Any use of '{' and/or '}' that does not appear to be part of a well-formed _pattern_ causes Jupyter to pass on the entire command string unchanged. This happens even when the command string contains other patterns that are well-formed, and could be interpolated.

So, there is now no need to return a status value to indicate the success/failure of the interpolation. The string returned by the interpolation function is generated by emulating Jupyter's behavior in the same situation.

I have also moved the interpolation function into the superclass  [Interpreter.java](https://github.com/sanjaydasgupta/zeppelin/commit/9e75333ff7ef9c27678382c598b524eafc6191f7#diff-f9d22f7302ae356454cdcc637942856fR493) to enable it to be shared and used by other interpreters too. This follows a suggestion from @Leemoonsoo earlier in the PR. My implementation of this feature for the Shell interpreter is also ready, and I will submit a PR once this one is done.

Tagar:
My two cents 

> One of the things I learned about Jupyter's implementation is that it never intercepts an error case

I personally not necessarily like how this piece is implemented in Jupyter.. that an error can be silently ignored, and in case of an error whole {.. } block is returned unchanged. 

Should we have a Zeppelin option that controls this behavior?
1. Option 1 if used - it'll completely follow Jupyter bahavior (no error, {..} block return unchanged).
2. Option 2 - fail explicitly if variable in curly braces can't be found.

That being said, I think most of folks may actually prefer option 1, as introduction of { .. } may be a breaking change for some notebooks' code. So I think Option 1 should be default.

> This follows a suggestion from @Leemoonsoo earlier in the PR. My implementation of this feature for the Shell interpreter is also ready, and I will submit a PR once this one is done.

That's great. Thanks for working on this. I see there are so many good use cases for this feature.

sanjaydasgupta:
Would like to hear about any further improvements required on this feature @felixcheung, @Leemoonsoo 

Thanks for your thoughts @Tagar

docs/interpreter/spark.md
```diff
@@ -343,6 +343,29 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
...
 </div>
 
+#### Alternative Object Retrieval Mechanism
+In addition to the programmatic access using `z.get()` shown above, 
+some interpreters also support inline interpolation of such objects
+into command lines by using substitution patterns of the form `{object-key}`.
+The following example shows how the value of an object `put` into `ZeppelinContext` 
+by the Scala interpreter is spliced into a command of the SQL interpreter.
+
+__In Scala interpreter:__
```
felixcheung: you should still do `{% highlight scala %}` for proper highlighting, just don't use `div`

docs/interpreter/spark.md
```diff
@@ -343,6 +343,29 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
   </div>
 </div>
 
+#### Alternative Object Retrieval Mechanism
+In addition to the programmatic access using `z.get()` shown above, 
+some interpreters also support inline interpolation of such objects
+into command lines by using substitution patterns of the form `{object-key}`.
+The following example shows how the value of an object `put` into `ZeppelinContext` 
+by the Scala interpreter is spliced into a command of the SQL interpreter.
```
felixcheung: `command` might not be the right word, `paragraph`? `query`?

docs/interpreter/spark.md
```diff
@@ -343,6 +343,29 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
...
+__In SQL interpreter:__
+```
+select * from {tableName}
+```
+The escaping pattern `{{anything}}` may be used in situations where `{` and `}` are needed 
+in the command line. The pattern `{{anything}}` is translated into `{anything}` and spliced 
+into the containing command.
+
+The use of the `{object-key}` format for object interpolation is currently only available 
+in the SQL interpreter, but will be provided in other interpreters progressively.
```
felixcheung: I'm a bit confused which commits are going to which. There seems to be more explanation in other commits that is not showing up in this PR.

but as discussed, let's explain more the behavior. It's fine to say this is modeled on Jupyter but different in these ways. eg. only when object-key is found in context, etc. I see some of these in the other commit.

felixcheung: I think we need quite a bit in the documentation to explain this more.

spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -85,6 +89,58 @@ public boolean concurrentSQL() {
...
+    // Iterate while there is any pattern left in the residual-command text
+    while ((matcher = pattern.matcher(residualCommand)).matches()) {
+      // Get any characters until the next pattern, add to new-command buffer
+      newCommandBuffer.append(matcher.group(1));
+      // Get the variable pattern
+      String template = matcher.group(2);
+      if (template.matches("[{][^{}][^}]*[}]")) {
+        // Handle variable substitution pattern ...
+        // Extract variable-name from template
+        String variableName = template.substring(1, template.length() - 1);
```
felixcheung: handle when template is empty or len <= 1?

spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -85,6 +89,58 @@ public boolean concurrentSQL() {
...
+        // Get variable-value from interpreter's context
+        Object variableValue = sparkZeppelinContext.get(variableName);
+        if (variableValue == null) {
+          // No such variable - Error!
+          return String.format("ERROR:unknown variable '%s'", variableName);
```
felixcheung: as commented, let's not use the return string as the success/error status as well.

spark/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -85,6 +89,58 @@ public boolean concurrentSQL() {
...
+        } else {
+          // Append value to new-command buffer
+          newCommandBuffer.append(variableValue.toString());
+        }
+      } else if (template.matches("[{]{2}[^{}][^}]*[}]{2}")) {
+        // Handle '{' escaping pattern ...
+        // Extract escaped sub-string from template
+        String escapedSubString = template.substring(1, template.length() - 1);
```
felixcheung: ditto len of template

felixcheung:
could you review my comments? there are a number of comments from before that I feel we should address first.

sanjaydasgupta:
@felixcheung - We've exchanged many ideas back and forth on this PR to the point where it has become more of a discussion document than a software submission :-)

I'm in favor of closing this PR, and starting over with a new PR that gathers together all of the learnings from the many suggestions and review comments in this one. That should also allow me to fix somethings that seem to have got messed with my repo -- I'm relatively new to git, and am not being able to reverse some of this damage.

Let me know if you agree, or if there is a better way to take this forward.

felixcheung:
Sure - feel free to close this and open a new PR.

PR #2503, opened 2017-07-25
[ZEPPELIN-2808] remember me support

herval:
### What is this PR for?
Add support for "remember me" cookies, so we can activate LDAP without forcing users to re-login every time

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2808

- Activate "rememberMe" by following the steps on the `shiro.ini.template`
- Log in with the "remember me" option checked
- Close the browser
- Reopen the browser
- You should still be logged in

### Questions:
* While this is standard behavior on many web-apps, some folks (as exemplified on the Shiro documentation) find it "not secure". I'd like to know Zeppelin folks' opinion on this.
* The "remember me" checkbox will show, regardless of whether you configure the rememberMeManager or not. Should this be a default? Should we hide it from the UI based on some config?

### DIFF
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -46,6 +46,12 @@ user3 = password4, role2
 #ldapRealm.userDnTemplate = uid={0},ou=Users,dc=COMPANY,dc=COM
 #ldapRealm.contextFactory.authenticationMechanism = simple
 
+### If you want "remember me" to keep used logged in between sessions
+# securityManager = org.apache.zeppelin.security.CookieBasedSecurityManager
+# rememberMeManager = org.apache.shiro.web.mgt.CookieRememberMeManager
+# securityManager.rememberMeManager = $rememberMeManager
+# securityManager.rememberMeManager.cipherKey = <generate a cypher key here>
+
 ### A sample PAM configuration
 #pamRealm=org.apache.zeppelin.realm.PamRealm
 #pamRealm.service=sshd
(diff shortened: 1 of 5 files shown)


1ambda:
I don't have a preference, but in my experience, many web apps rely on the browser provided remember-me feature. Here is an example of Github sign-in page

- https://github.com/login

### Chrome 

image

### Safari

image

herval:
That popup is not the site’s “remember me” - github actually keeps you logged in by default (close the browser and reopen, and you’re still logged, no need to re-authenticate)

You can see some examples of the “remember me”/“keep logged in” I’m talking about on this ticket here: http://www.uxforthemasses.com/login-page/

1ambda:
Thanks. I updated the previous comment.

conf/shiro.ini.template
```diff
@@ -46,6 +46,12 @@ user3 = password4, role2
 #ldapRealm.userDnTemplate = uid={0},ou=Users,dc=COMPANY,dc=COM
 #ldapRealm.contextFactory.authenticationMechanism = simple
 
+### If you want "remember me" to keep used logged in between sessions
```
1ambda: Hi, without enalbing these lines, remember me feature works. Is it intended?

herval:
Hmm It shouldn’t work without the Cookie manager. The checkbox will still show, though, which begs the question if it should be a default? (Otherwise we’ll have to hide the checkbox based on some shiro config)

1ambda:
Then, I think we need to hide when cookie manager is disabled.

herval:
any arguments against following the current way most sites do it (which is having "remember me" on by default, instead of a checkbox option)?

Showing/hiding the checkbox depending on a config will introduce quite a lot of config code, so unless there's a reasoning not to always support remembering credentials, I'd say it's a more sensible way to go

1ambda:
1. Let's expose the config using REST API we already have. 

- http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/usage/rest_api/configuration.html#configuration-rest-api-list

2. Check it in frontend and hide if it's disabled using `ng-if` or something.

PR #2504, opened 2017-07-26
ZEPPELIN-2811. Specify precise explicitly for travis build

zjffdu:
### What is this PR for?
Travis is upgrading ubuntu from precise to trusty, so sometimes your travis will fails if it is routed to trusty.
I confirmed that with travis support

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2811

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ matrix:
   include:
     # Test License compliance using RAT tool
     - jdk: "oraclejdk7"
+      dist: precise
       env: SCALA_VER="2.11" SPARK_VER="2.0.2" HADOOP_VER="2.6" PROFILE="-Prat" BUILD_FLAG="clean" TEST_FLAG="org.apache.rat:apache-rat-plugin:check" TEST_PROJECTS=""
 
     # Run e2e tests (in zeppelin-web)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -66,6 +67,7 @@ matrix:
     # HeliumApplicationFactoryTest - https://issues.apache.org/jira/browse/ZEPPELIN-2470
     # After issues are fixed these tests need to be included back by removing them from the "-Dtests.to.exclude" property
     - jdk: "oraclejdk8"
+      dist: precise
       env: SCALA_VER="2.11" SPARK_VER="2.2.0" HADOOP_VER="2.6" PROFILE="-Pspark-2.2 -Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
 
     # Test selenium with spark module for 1.6.3
```
felixcheung: need `dist` here?

zjffdu:
This is because travis use 'precise' by default before, so at that time it is OK when we don't specify dist. But now travis is upgrading ubuntu from `precise` to `trusty`, `trusty` will become default in the near future. That's why I  specify `precise` explicitly in this PR.

felixcheung:
I get that, I see one test matrix is explicitly setting dist to trusty, and so I assume some tests passes on trusty at least

e2e test https://github.com/apache/zeppelin/pull/2504/files#diff-354f30a63fb0907d4ad57269548329e3R54

zjffdu:
yeah, seems e2e tests needs trusty. @1ambda may have more context. 
A long term solution might be to upgrade all tests to trusty, but may need more changes. This PR is just trying to make the current .travis.yml works

felixcheung:
fair enough

zjffdu:
I will merge it soon, as it affect all the following PRs.

PR #2506, opened 2017-07-27
[ZEPPELIN-2813] revisions comparator

tinkoff-dwh:
### What is this PR for?
Sometimes need to see the difference between versions and to switch to another version and look for changes are not convenient (the page reloaded). This feature allows you to compare any two versions of the notebook.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2813

1 make some commits. Сchange the contents of paragraphs (delete, add, edit)
2 open Revisions comparator
3 compare revisions and check diff

comparator

### Questions:
* Does the licenses files need update? yes (updated)


zjffdu:
@tinkoff-dwh Do you mind to attach a gif for easy review ?

tinkoff-dwh:
@zjffdu done

sravan-s:
Tested this PR, worked fine. Nice addition 👍

tinkoff-dwh:
@felixcheung 
https://github.com/apache/zeppelin/tree/master/docs/quickstart here?

i don't know. I think this is not such a difficult thing to write in documentation...

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1920,6 +1920,25 @@ private void getNoteByRevision(NotebookSocket conn, Notebook notebook, Message f
...
+  private void getNoteByRevisionForCompare(NotebookSocket conn, Notebook notebook,
+      Message fromMessage) throws IOException {
+    String noteId = (String) fromMessage.get("noteId");
+    String revisionId = (String) fromMessage.get("revisionId");
+
+    String position = (String) fromMessage.get("position");
+    AuthenticationInfo subject = new AuthenticationInfo(fromMessage.principal);
+    Note revisionNote;
+    if (revisionId.equals("Head")) {
+      revisionNote = notebook.loadNoteFromRepo(noteId, subject);
```
khalidhuseynov: for this case, i think it's better to use `notebook.getNote(noteId)` since the latest note will be in memory always

felixcheung:
+1 to that https://github.com/apache/zeppelin/pull/2506#pullrequestreview-53385386

felixcheung:
re: doc - I see what you mean. I think it'll be great to have doc on import/export, and this, maybe like https://zeppelin.apache.org/docs/0.7.2/manual/notebookashomepage.html

zjffdu:
@tinkoff-dwh  It looks great. One suggestion is that is it possible to show the revision comparison inline instead of popping up a new window. Because I think we can some followup work for this feature, like revert to some revision, doing it is inline is more convenient for users.

tinkoff-dwh:
@zjffdu 
made a separate window because I think that it's more convenient (when a lot of code) and now rewritten for a long time as it is necessary to rewrite all the layout

tinkoff-dwh:
@felixcheung @khalidhuseynov 
fixed. added documentation

tinkoff-dwh:
Ready to review

zeppelin-web/src/app/notebook/notebook.html
```diff
@@ -13,6 +13,7 @@
 -->
 <!-- Here the controller <NotebookCtrl> is not needed because explicitly set in the app.js (route) -->
 <div id="actionbar" ng-include src="'app/notebook/notebook-actionBar.html'"></div>
+<div id="note-revisions-comparator-modal-container" ng-include src="'app/notebook/revisions-comparator.html'"></div>
```
1ambda: Hi, can we use angular1's component instead of using ng-include? 

Here are few reasons. 

- ng-include can't be handled in webpack. It means we can't take advantage of fast re-build and HMR
- it's blocker when we are moving to other stacks. (whatever it's is)

Here is ES6 based component guide and an example in Zeppelin

- https://github.com/toddmotto/angularjs-styleguide
- https://github.com/apache/zeppelin/tree/master/zeppelin-web/src/app/jobmanager/job

1ambda:
regarding UX, 

- `Choose` box is too big compared to the text inside
- It would be nice to have the list-like view for each item. For example,

image

tinkoff-dwh:
@1ambda 
fixed
comparator

tinkoff-dwh:
Ready to review

1ambda:
@tinkoff-dwh Thanks for the update. 

In feature and code level, it's good. But personally, I hope we can get improved UI.

tinkoff-dwh:
@1ambda 
is it "LGTM" or no?) ' we can get improved UI' sounds unclear)

1ambda:
I didn't mean both. Just gave an opinion. 
Personally, I don't believe that reviewer only can reply `LGTM` or `NO`.

tinkoff-dwh:
@1ambda 
I mean, it is not clear what to do, wait for other reviews or to do something (but then it is not clear what)

1ambda:
For me, it's up to you.

- keep current implementation
- try and propose few different UIs

tinkoff-dwh:
@1ambda 
what about implementation like a Jetbrains Idea comparator?
1) header contains comboboxes for choice commit and combobox with mode of diff
  - side-by-side paragraph
  - side-by-side all paragraphs (join text of all paragraphs and compare)
2) actions block contains buttons (next paragraph with changes, prev paragraph with changes, next changes into current paragraph, prev changes into current paragraph, combobox with all paragraphs )
3) diff

1ambda:
The one of famous product what does the same thing with this PR is Source Tree as u know. And it has this kind of UI. I think @tinkoff-dwh you can get ideas from this UI

image

- you can see the whole commit list and scroll down to navigate more. 
- if you click one commit, automatically display the diff between prev commit
- if you want to compare to separate commits explicit, you can select the commits for comparison.

image

tinkoff-dwh:
@1ambda 
Restyling done.
New UI:
gif

Leemoonsoo:
Tried this branch and it works really well

tinkoff-dwh:
@1ambda 
what about new UI?

Leemoonsoo:
@1ambda what do you think?

Merge to master if no further discussions

PR #2507, opened 2017-07-27
[ZEPPELIN-2815] Improve interpreter dependencies logging

necosta:
### What is this PR for?
As a developer, I want to be able to debug issues on dependency resolution "easier". 
To achieve this, making available (in debug mode), existing listeners. 
Also, exposing exceptions on dependency resolver.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2815

Enable DEBUG logging mode. Import dependencies. Check detailed info. from dependency loading being logged.

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
@@ -177,8 +177,9 @@ public class DependencyResolver extends AbstractDependencyResolver {
             DependencyFilterUtils.andFilter(exclusionFilter, classpathFilter));
     try {
       return system.resolveDependencies(session, dependencyRequest).getArtifactResults();
-    } catch (NullPointerException ex) {
-      throw new RepositoryException(String.format("Cannot fetch dependencies for %s", dependency));
+    } catch (Exception ex) {
+      throw new RepositoryException(
+              String.format("Cannot fetch dependencies for %s", dependency), ex);
     }
   }
 }
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
@@ -19,7 +19,6 @@ package org.apache.zeppelin.dep;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URL;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


necosta:
Hi @Leemoonsoo and others, should be a trivial one. Low priority. Many thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/TransferListener.java
```diff
@@ -122,7 +120,7 @@ public void transferSucceeded(TransferEvent event) {
   @Override
   public void transferFailed(TransferEvent event) {
     transferCompleted(event);
-    event.getException().printStackTrace(out);
+    logger.warn(event.getException().getMessage());
```
necosta: Warn and not error as it might be legit to not find a dependency in one repo (should fallback to the next)

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
```diff
@@ -177,8 +177,9 @@ public synchronized void copyLocalDependency(String srcPath, File destPath)
             DependencyFilterUtils.andFilter(exclusionFilter, classpathFilter));
     try {
       return system.resolveDependencies(session, dependencyRequest).getArtifactResults();
-    } catch (NullPointerException ex) {
-      throw new RepositoryException(String.format("Cannot fetch dependencies for %s", dependency));
+    } catch (Exception ex) {
```
1ambda: should be still NPE? since Exception can cache more exception types.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/TransferListener.java
1ambda: It's a good update

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/TransferListener.java
```diff
@@ -55,13 +53,13 @@ public void transferInitiated(TransferEvent event) {
   @Override
   public void transferProgressed(TransferEvent event) {
     TransferResource resource = event.getResource();
-    downloads.put(resource, Long.valueOf(event.getTransferredBytes()));
```
1ambda: Don't we need converting? (LINE 62 as well)

necosta: No. It's already a long

necosta: Both are already a long I should say

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
necosta: Good question. I  changed it to Exception as I see that org.sonatype.aether.RepositorySystem can throw DependencyResolutionException and I suspect we can have file exceptions and other in the whole org.sonatype.aether library. But because I didn't actually get an error here, I'll leave the final decision to you. Should we revert this to only have NPE or keep it broad with Exception?

felixcheung: I think more specific exception would be preferred...

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/TransferListener.java
```diff
@@ -135,10 +133,10 @@ private void transferCompleted(TransferEvent event) {
 
   @Override
   public void transferCorrupted(TransferEvent event) {
-    event.getException().printStackTrace(out);
+    logger.error(event.getException().getMessage());
```
felixcheung: `logger.error("some message", event.getException());` instead

necosta: Done

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/DependencyResolver.java
necosta: Now 

> catch (NullPointerException | DependencyResolutionException ex)

necosta:
@1ambda , @felixcheung , ready for another round of review. Thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/TransferListener.java
felixcheung: I'd convert to logger.warn(message, event.getException()) here
https://www.slf4j.org/api/org/slf4j/Logger.html#warn(java.lang.String,%20java.lang.Throwable)

necosta:
Makes sense @felixcheung . Changed warn message to 
`logger.warn("Unsuccessful transfer", event.getException());`

Leemoonsoo:

PR #2508, opened 2017-07-28
[ZEPPELIN-2817] Support default interpreter setting in create note re…

reminia:
…st api

### What is this PR for?
Can set default interpreter in create note rest api.

### What type of PR is it?
[Improvement]

### Todos
None

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2817

Post name and interpreter to rest api/notebook, a note with the interpreter should be created.

### Questions:

### DIFF
--- a/docs/usage/rest_api/notebook.md
+++ b/docs/usage/rest_api/notebook.md
@@ -109,7 +109,7 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td><pre>
 {
   "name": "name of new note",
-  "interpreter": "python"
+  "defaultInterpreter": "python"
 }</pre></td>
     </tr>
     <tr>
@@ -117,7 +117,7 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td><pre>
 {
   "name": "name of new note",
-  "interpreter": "python",
+  "defaultInterpreter": "python",
   "paragraphs": [
     {
       "title": "paragraph title1",
--- a/docs/usage/rest_api/notebook.md
+++ b/docs/usage/rest_api/notebook.md
@@ -80,6 +80,7 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
     <tr>
       <td>Description</td>
       <td>This ```POST``` method creates a new note using the given name or default name if none given.
+      Default interpreter and initial paragraphs can also be set.
           The body field of the returned JSON contains the new note id.
       </td>
     </tr>
@@ -96,14 +97,23 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td> 500 </td>
     </tr>
     <tr>
-      <td> sample JSON input (without paragraphs) </td>
+      <td> sample JSON input (without interpreter and paragraphs) </td>
       <td><pre>{"name": "name of new note"}</pre></td>
     </tr>
     <tr>
-      <td> sample JSON input (with initial paragraphs) </td>
+      <td> sample JSON input (with interpreter) </td>
+      <td><pre>
+{
+"name": "name of new note",
+"interpreter": "python"
+}</pre></td>
+    </tr>
+    <tr>
+      <td> sample JSON input (with interpreter, initial paragraphs) </td>
       <td><pre>
 {
   "name": "name of new note",
+  "interpreter": "python",
   "paragraphs": [
     {
       "title": "paragraph title1",
--- a/docs/usage/rest_api/notebook.md
+++ b/docs/usage/rest_api/notebook.md
@@ -90,7 +90,11 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
     </tr>
     <tr>
       <td>Success code</td>
-      <td>200</td>
+      <td>201</td>
+    </tr>
+    <tr>
+      <td>Not Found code</td>
+      <td>404</td>
     </tr>
     <tr>
       <td> Fail code</td>
--- a/docs/usage/rest_api/notebook.md
+++ b/docs/usage/rest_api/notebook.md
@@ -90,7 +90,7 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
     </tr>
     <tr>
       <td>Success code</td>
-      <td>201</td>
+      <td>200</td>
     </tr>
     <tr>
       <td>Not Found code</td>
@@ -108,8 +108,8 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
       <td> sample JSON input (with interpreter) </td>
       <td><pre>
 {
-"name": "name of new note",
-"interpreter": "python"
+  "name": "name of new note",
+  "interpreter": "python"
 }</pre></td>
     </tr>
     <tr>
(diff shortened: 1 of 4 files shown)


reminia:
Add doc and some modify. @felixcheung @sravan-s , plz help review again.
Some details:

1. post interpreter name not id, cause name is user friendly and id is something like CRB87NS
2. if the interpreter name doesnt exist,  return **bad request status** and no note is created.
3. if successfully created the note, return CREATED status  like the old DOC says.
4. name/interpreter/paragraphs no need to be supplied together, empty message is also ok.

reminia:
@felixcheung @Leemoonsoo  can you help review this?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -345,16 +370,20 @@ public Response createNote(String message) throws IOException {
...
+        noteName = request.getName();
+      }
     }
 
     note.setName(noteName);
     note.persist(subject);
     notebookServer.broadcastNote(note);
     notebookServer.broadcastNoteList(subject, SecurityUtils.getRoles());
-    return new JsonResponse<>(Status.OK, "", note.getId()).build();
+    return new JsonResponse<>(Status.CREATED, "", note.getId()).build();
```
felixcheung: if I read this right, the doc says it should be 200 OK
https://github.com/apache/zeppelin/pull/2508/files#diff-ccd7e2a2e181424ea2f4e346189e7688R93
?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -334,7 +335,31 @@ public Response createNote(String message) throws IOException {
...
+          break;
+        }
+      }
+      if (defaultId != null) {
+        interpreterIDs.remove(defaultId);
+        interpreterIDs.add(0, defaultId);
+        note = notebook.createNote(interpreterIDs, subject);
+      } else {
+        String errorMsg = String.format("interpreter %s doesn't exist", interpreter);
+        return new JsonResponse<>(Status.BAD_REQUEST, errorMsg).build();
```
felixcheung: we should document this response code
but this should be 404 Not Found instead?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
reminia: @felixcheung , I mod the success code in doc to 201.  Cause it's a created api. And also the response json in old doc is CREATED status. [see](https://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/usage/rest_api/notebook.html#create-a-new-note).

reminia:
Test added @felixcheung

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
felixcheung: I think we need to be super careful with changing an existing response code for an existing API - there is likely existing usage that could be broken with a simple code change.

I get that 201 might be better and/or more consistent with other APIs though. Do you think it makes a difference if we change it now, or we leave it as before?

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
```diff
@@ -207,4 +210,51 @@ public void testClearAllParagraphOutput() throws IOException {
...
     ZeppelinServer.notebook.removeNote(note.getId(), anonymous);
   }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testCreateNote() throws IOException {
+    // prepare data
+    Map<String, Object> data = Maps.newHashMap();
+    data.put("name", "test");
+    data.put("interpreter", "spark");
```
felixcheung: perhaps we should test with a different interpreter? `spark` is likely the default even when this isn't set in the request

reminia: change spark to python interpreter.

reminia:
I decide to leave the api no change at now.  Maybe another jira should be created to address this. I've found that several create operation api doesnt return 201 but 200， such as repo add, paragraph add. @felixcheung @Leemoonsoo

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -334,7 +335,31 @@ public Response createNote(String message) throws IOException {
...
+      String defaultId = null;
+      for (String id : interpreterIDs) {
+        if (manager.get(id).getName().equals(interpreter)) {
+          defaultId = id;
+          break;
+        }
+      }
+      if (defaultId != null) {
+        interpreterIDs.remove(defaultId);
+        interpreterIDs.add(0, defaultId);
```
zjffdu: Would this cause 2 duplicated `interpreterId` in this `interpreterIDs` ?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -345,9 +370,13 @@ public Response createNote(String message) throws IOException {
       }
     }
     note.addNewParagraph(subject); // add one paragraph to the last
-    String noteName = request.getName();
-    if (noteName.isEmpty()) {
-      noteName = "Note " + note.getId();
+
+    //deal with possible empty note name
+    String noteName = "Note " + note.getId();
```
zjffdu: Another option is to do in frontend that we can enforce user to provide a note name ? Personally, I prefer to do it in frontend to enforce user providing note name.

docs/usage/rest_api/notebook.md
```diff
@@ -92,18 +93,31 @@ Notebooks REST API supports the following operations: List, Create, Get, Delete,
...
+  "name": "name of new note",
+  "interpreter": "python"
+}</pre></td>
+    </tr>
+    <tr>
+      <td> sample JSON input (with interpreter, initial paragraphs) </td>
       <td><pre>
 {
   "name": "name of new note",
+  "interpreter": "python",
```
zjffdu: defaultInterpreter is better for me. `interpreter` is a little misleading.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
reminia: Hey Jeff, this is rest api without frontend check. Now frontend can create note without name supplied. This api just tries to stay the same with the web function.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
reminia: sorry, I didnt get your idea. Do you mean interpreterID duplicate test should be added? Why?Remove then add , no duplicate.

docs/usage/rest_api/notebook.md
reminia: I'll change the name.

reminia:
@zjffdu  @Leemoonsoo can you help review?

PR #2516, opened 2017-08-03
[ZEPPELIN-2827] Spark's pyspark fails if unsupported version of matplotlib is present

prabhjyotsingh:
### What is this PR for?
%spark.pyspark paragraphs keep running for more than > 5 minutes and then at the end, throw TTransportException.

```
org.apache.thrift.transport.TTransportException
	at org.apache.thrift.transport.TIOStreamTransport.read(TIOStreamTransport.java:132)
	at org.apache.thrift.transport.TTransport.readAll(TTransport.java:86)
	at org.apache.thrift.protocol.TBinaryProtocol.readAll(TBinaryProtocol.java:429)
	at org.apache.thrift.protocol.TBinaryProtocol.readI32(TBinaryProtocol.java:318)
	at org.apache.thrift.protocol.TBinaryProtocol.readMessageBegin(TBinaryProtocol.java:219)
	at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:69)
```

This happens because Zeppelin refers to `matplotlib.rcParams['savefig.format']` https://github.com/apache/zeppelin/blob/master/interpreter/lib/python/mpl_config.py#L83 which is only present in matplotlib 1.2.0 (https://github.com/matplotlib/matplotlib/blob/v1.2.0/matplotlibrc.template#L355) and onwards.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2827](https://issues.apache.org/jira/browse/ZEPPELIN-2827)

Try running spark.pyspark on a machine with matplotlib==1.1.1 installed, pyspark should work as expected.

### Questions:

### DIFF
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -120,6 +120,11 @@ class PyZeppelinContext(dict):
     # If we don't have matplotlib installed don't bother continuing
     try:
       import matplotlib
+
+      #matplotlib.rcParams['savefig.format'] is only avail in version 1.2.0 and above
+      from distutils.version import LooseVersion
+      if LooseVersion("1.2.0") >= LooseVersion(matplotlib.__version__):
+        return
     except ImportError:
       return
     
--- a/spark/src/main/resources/python/zeppelin_pyspark.py
+++ b/spark/src/main/resources/python/zeppelin_pyspark.py
@@ -120,11 +120,6 @@ class PyZeppelinContext(dict):
     # If we don't have matplotlib installed don't bother continuing
     try:
       import matplotlib
-
-      #matplotlib.rcParams['savefig.format'] is only avail in version 1.2.0 and above
-      from distutils.version import LooseVersion
-      if LooseVersion("1.2.0") >= LooseVersion(matplotlib.__version__):
-        return
     except ImportError:
       return
     
--- a/interpreter/lib/python/mpl_config.py
+++ b/interpreter/lib/python/mpl_config.py
@@ -71,7 +71,11 @@ def _on_config_change():
     supported_formats = _config['supported_formats']
     if fmt not in supported_formats:
         raise ValueError("Unsupported format %s" %fmt)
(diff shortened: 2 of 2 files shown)


prabhjyotsingh:
@Leemoonsoo @zjffdu @agoodm please help review this.

Leemoonsoo:
Looks good to me!

spark/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -120,6 +120,11 @@ def _setup_matplotlib(self):
     # If we don't have matplotlib installed don't bother continuing
     try:
       import matplotlib
+
+      #matplotlib.rcParams['savefig.format'] is only avail in version 1.2.0 and above
+      from distutils.version import LooseVersion
+      if LooseVersion("1.2.0") >= LooseVersion(matplotlib.__version__):
+        return
```
agoodm: With this approach, the plotting backend setup is completely aborted if the matplotlib version is 1.2.0 or older. However this is bad because this also means matplotlib will not fallback to the Agg backend, which is essential since for most systems the default backend is not notebook compatible.

An alternative is to just catch a `KeyError`, but in general I think this is still not the proper solution to the problem since removing the custom backend over a single missing key is beyond overkill.

prabhjyotsingh: Right, agreed in most of the cases if user in interested in using matplotlib library he would have at-least 1.5.x installed if not 2.0.x.

But the idea of this PR is not to support older version, but to make sure pyspark continues to work even is a non-supported version of matplotlib is installed and it should behave like there was no matplotlib installed.

agoodm: Fair enough. Do you think you can address point (2)?

prabhjyotsingh: Sure, done. Also, have reverted the changes from `zeppelin_pyspark.py`. Let me know your thoughts.

prabhjyotsingh:
Thank you, everyone, for the review. Will merge this if no more discussion.

PR #2519, opened 2017-08-07
ZEPPELIN-2833 JDBC Interpreter doesn't follow JDBC specification when…

ss77892:
… getting the results.

### What is this PR for?
Fix for ZEPPELIN-2833. JDBC Interpreter uses result set next() method after it's already reached the end. According the JDBC documentation that may cause SQL exception. 
 
### What type of PR is it?
Bug Fix

### Todos
None

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2833

Use Phoenix 4.7+ select * query for tables that have less than 1000 records. Without patch an exception about closed Result set will be thrown. 

### Questions:
No
No
No


prabhjyotsingh:
@ss77892  Thanks for your contribution, your changes look good, but can you update the description of this PR a bit, specifically this section "What is this PR for?"

PR #2520, opened 2017-08-07
ZEPPELIN-2833 JDBC Interpreter doesn't follow JDBC specification when…

ss77892:
… getting the results.

### What is this PR for?

Fix for ZEPPELIN-2833. JDBC Interpreter uses result set next() method after it's already reached the end. According the JDBC documentation that may cause SQL exception.

### What type of PR is it?
Bug Fix

### Todos
None

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-2833

Use Phoenix 4.7+ select * query for tables that have less than 1000 records. Without patch an exception about closed Result set will be thrown.

### Questions:
No
No
No

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -38,6 +38,7 @@ import org.apache.commons.dbcp2.DriverManagerConnectionFactory;
 import org.apache.commons.dbcp2.PoolableConnectionFactory;
 import org.apache.commons.dbcp2.PoolingDriver;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
Tested on local (for hive and phoenix) works as expected. LGTM!

ss77892:
@felixcheung As I mentioned in the JIRA the default jdbc driver that is used in tests has different behavior ( allows to make call next() even if the result set is already at the end). Not sure how to develop a test in this case.

ss77892:
@felixcheung all the tests that doesn't reach maxCount may be considered as the test that calls resultSet.next() multiple time (if the default jdbc would be Phoenix they would produce this exception) and there is a test testSelectQueryMaxResult that tests that the limit has been reached.

prabhjyotsingh:

PR #2521, opened 2017-08-08
[ZEPPELIN-2841] fix a problem in shell interpreter . Working directory '.' can not be found in docker environment

vistep:
### What is this PR for?
shell interpreter complained that working directory '.' can not be found in docker environment.
I add a line of code to set current working directory to USER`s home, and it works.

### What type of PR is it?
Bug Fix

### Todos
* tests

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2841

run shell interpreter`s test units

### Questions:
No
No
No


vistep:
Jenkins said that "Can't find build for commit e8cbdd180a22d1d53366c94c5432c217d70a5727 from vistep".
I have no idea about what happened. How to  fix it?

vistep:
I have fix the build problem by setting travis according to what Jenkins said at last part.

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -98,6 +98,7 @@ public InterpreterResult interpret(String cmd, InterpreterContext contextInterpr
         contextInterpreter.out, contextInterpreter.out));
       executor.setWatchdog(new ExecuteWatchdog(Long.valueOf(getProperty(TIMEOUT_PROPERTY))));
       executors.put(contextInterpreter.getParagraphId(), executor);
+      executor.setWorkingDirectory(new File(System.getenv("HOME")));
```
felixcheung: this won't work in windows though (well)

vistep: Thanks so much, I will try to figure it out.

vistep:
I replaced System.getenv("HOME") with  System.getProperty("user.home") .
This time, Jenkins said "[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.3:yarn (yarn test) on project zeppelin-web: Failed to run task: 'yarn run test' failed. (error code 1) -> [Help 1]"
I don`t understand why my code affects zeppelin-web ?

Leemoonsoo:
build sometimes fail because of network connection problem of flaky tests.
Could you try restart failed one?

vistep:
I tried again and it passed.

Leemoonsoo:
LGTM and merge to master and branch-0.7 if no further comment.

sanjaydasgupta:
This implementation has a harmful side effect that can break existing notebooks. 

The problem with this fix is that it changes the meaning of all relative path-names used in any shell commands. For example, after this fix, the command `ls -l data` would operate on the _data_ subdirectory of the user's home directory, whereas before the fix it would operate on the  _data_ subdirectory of zeppelin's working directory.

One immediate change that is suggested is to change the fix to make it apply only to the narrow situation it is needed in (a Docker environment).

vistep:
Can we use a try-catch block to test if we need to set working directory when zeppelin opens a shell interpreter?

sanjaydasgupta:
I have not looked at the code in detail, so will refrain from making suggestions about the code :-)

vistep:
@Leemoonsoo How about the idea above?

prabhjyotsingh:
@sanjaydasgupta @vistep have create JIRA for the same https://issues.apache.org/jira/browse/ZEPPELIN-2903.

PR #2525, opened 2017-08-11
[ZEPPELIN-2647] Make admin role to bypass auth logic

yu74n:
### What is this PR for?
For administrator, make new admin role that assigned user can see all notebooks.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2647

1. Change Notebook workspace to private through whether ZEPPELIN_NOTEBOOK_PUBLIC = false or zeppelin.notebook.public = false.
2. Set role name to use as admin through ZEPPELIN_OWNER_ROLE = <role name> or zeppelin.owner.role = <role name>.
Default role name is admin
3. Login as user who is not assigned as admin and create notebook.
4. Logout the user and login another user who is assigned as admin, open the created notebook.

### Questions:
* Is there breaking changes for older versions? Y/N

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
@@ -24,7 +24,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -297,7 +296,8 @@ public class NotebookAuthorization {
   }
 
   public boolean isWriter(String noteId, Set<String> entities) {
-    return isMember(entities, getWriters(noteId)) || isMember(entities, getOwners(noteId));
+    return isMember(entities, getWriters(noteId)) ||
+           isMember(entities, getOwners(noteId));
   }
 
   public boolean isReader(String noteId, Set<String> entities) {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
@@ -403,6 +403,7 @@ public class NotebookAuthorization {
         setReaders(noteId, entities);
         entities = getRunners(noteId);
         entities.add(subject.getUser());
+        entities.add(admin);
         setRunners(noteId, entities);
         entities = getWriters(noteId);
         entities.add(subject.getUser());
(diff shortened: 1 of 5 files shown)


jongyoul:
LGTM

jongyoul:
I think so. If we add admin when creating, it would be better.

1ambda:
LGTM.

1ambda:
I think this feature is useful.  That's because admin can't see all notes currently 

- since they not owner even if admin has `admin` role. 
- so admin needs to ask all notebook creator to set owner admin as well.

IMO, it's quite weird and i think this PR can solve the problem.

1ambda:
@yu74n Could you solve the conflict and reply @felixcheung's last comment?

>  but seems like this isn't just default owner/admin for private mode only, but in all cases?

yu74n:
@1ambda thx
@felixcheung You are right, this feature affects all modes(private and public). I don't understand public mode in detail, is there any problem if we set admin role in public mode? Should I set admin role at NotebookAuthorization.setNewNotePermissions() when a notebook is created?

yu74n:
One more thing. Should we set admin role to notebook owner in public mode? I'm going to change this feature to set admin role to notebook owner, reader and writer in private mode though.

1ambda:
Let me check public and private modes and reply as well. Seems that CI constantly fails.

yu74n:
I've fixed failed unit tests.

1ambda:
@yu74n Thanks for the update. Could u restart / check the failed job? https://travis-ci.org/yu74n/zeppelin/jobs/273160145

yu74n:
After fixed 'testSyncWithAcl', I manually restart CI. All test passed.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
```diff
@@ -384,24 +385,30 @@ public boolean apply(NoteInfo input) {
   
   public void setNewNotePermissions(String noteId, AuthenticationInfo subject) {
     if (!AuthenticationInfo.isAnonymous(subject)) {
+      String admin = conf.getString(ConfVars.ZEPPELIN_OWNER_ROLE);
       if (isPublic()) {
         // add current user to owners - can be public
         Set<String> owners = getOwners(noteId);
         owners.add(subject.getUser());
+        owners.add(admin);
```
1ambda: I remember when I saw this PR first time, we don't have to add admin as the owner. Isn't it? Or am I missing something?

- https://github.com/apache/zeppelin/pull/2525/commits/1a1374ee4d93602b37577b96546ac80906cb7017

@yu74n

yu74n: @1ambda  Beginning of this PR, I was going to bypass authentication when a user login as admin, then we didn't have to add the admin user as the owner. But @felixcheung and @jongyoul suggested me that it would be better to add admin role when a notebook is created, after that I fixed this PR to do so... then we need to add admin as the owner, I think.

1ambda: IMO, it's not a scalable solution. For instance, if user wants to change the admin role name for some reason, they have to change the all existing owners of all notes.

yu74n: You are right, I didn't notice that. I'm trying to revert the code, but I think it would be easier to create new PR and cherry pick the code we need from this PR.

yu74n: I've created new PR #2585. Could you review the PR again please ?

yu74n:
Close this PR because Continued at #2585

PR #2526, opened 2017-08-11
[ZEPPELIN-2848] Added new type of user to only run notebook

gfalcone:
### What is this PR for?

The idea of this PR is to provide a new kind of user : Runner.

Basically, what it does is that it just removes write authorization and allow user to read and run note.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
[ZEPPELIN-2848] https://issues.apache.org/jira/browse/ZEPPELIN-2848

- Log in as admin
- Create new notebook and create a paragraph with the interpreter you want
- Assign runner right to user1
- Log in as user1
- Try to run the paragraph (should work)
- Try to modify the paragraph (should fail)
- Log in as user2
- Try to run the paragraph (should fail)

### Questions:

### DIFF
--- a/docs/setup/security/notebook_authorization.md
+++ b/docs/setup/security/notebook_authorization.md
@@ -36,13 +36,15 @@ As you can see, each Zeppelin notebooks has 3 entities :
 * Owners ( users or groups )
 * Readers ( users or groups )
 * Writers ( users or groups )
+* Runners ( users or groups )
 
 <center><img src="{{BASE_PATH}}/assets/themes/zeppelin/img/docs-img/permission_setting.png"></center>
 
 Fill out the each forms with comma seperated **users** and **groups** configured in `conf/shiro.ini` file.
 If the form is empty (*), it means that any users can perform that operation.
 
-If someone who doesn't have **read** permission is trying to access the notebook or someone who doesn't have **write** permission is trying to edit the notebook, Zeppelin will ask to login or block the user.
+If someone who doesn't have **read** permission is trying to access the notebook or someone who doesn't have **write** permission is trying to edit the notebook,
+or someone who doesn't have **run** permission is trying to run a paragraph Zeppelin will ask to login or block the user.
 
 <center><img src="{{BASE_PATH}}/assets/themes/zeppelin/img/docs-img/insufficient_privileges.png"></center>
 
@@ -63,13 +65,13 @@ or set `zeppelin.notebook.public` property to `false` in `conf/zeppelin-site.xml
 </property>
 ```
 
-Behind the scenes, when you create a new note only the `owners` field is filled with current user, leaving `readers` and `writers` fields empty. All the notes with at least one empty authorization field are considered to be in `public` workspace. Thus when setting `zeppelin.notebook.public` (or corresponding `ZEPPELIN_NOTEBOOK_PUBLIC`) to false, newly created notes have `readers` and `writers` fields filled with current user, making note appear as in `private` workspace.
+Behind the scenes, when you create a new note only the `owners` field is filled with current user, leaving `readers`, `runners` and `writers` fields empty. All the notes with at least one empty authorization field are considered to be in `public` workspace. Thus when setting `zeppelin.notebook.public` (or corresponding `ZEPPELIN_NOTEBOOK_PUBLIC`) to false, newly created notes have `readers`, `runners`, `writers` fields filled with current user, making note appear as in `private` workspace.
 
 ## How it works
 In this section, we will explain the detail about how the notebook authorization works in backend side.
 
 ### NotebookServer
-The [NotebookServer](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java) classifies every notebook operations into three categories: **Read**, **Write**, **Manage**.
+The [NotebookServer](https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java) classifies every notebook operations into three categories: **Read**, **Run**, **Write**, **Manage**.
 Before executing a notebook operation, it checks if the user and the groups associated with the `NotebookSocket` have permissions.
 For example, before executing a **Read** operation, it checks if the user and the groups have at least one entity that belongs to the **Reader** entities.
 
--- a/docs/setup/security/notebook_authorization.md
+++ b/docs/setup/security/notebook_authorization.md
@@ -46,7 +46,7 @@ If the form is empty (*), it means that any users can perform that operation.
 If someone who doesn't have **read** permission is trying to access the notebook or someone who doesn't have **write** permission is trying to edit the notebook,
 or someone who doesn't have **run** permission is trying to run a paragraph Zeppelin will ask to login or block the user.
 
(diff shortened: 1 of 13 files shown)


gfalcone:
Hi, 

Thanks for reviewing my PR. 

Actually readers cannot run the notebook. 

So what runners have read access + the ability to run the notebook

gfalcone:
@felixcheung runners have execute rights then in the Unix sense

gfalcone:
@felixcheung CI passed https://travis-ci.org/gfalcone/zeppelin

1ambda:
I saw that people were asking there is this kind of permission. 

- https://stackoverflow.com/questions/43971836/how-do-i-set-permissions-to-a-zeppelin-note-to-be-able-to-execute-but-not-change

gfalcone:
@1ambda yes this PR totally address this problem

gfalcone:
@felixcheung CI ok

felixcheung:
Can Readers here Run as well, so it's a bit different from the Linux permission model. Can Runners Read?

I am not sure if we should change the existing Readers role and it's semantics though. Would love to see how other think about this.

gfalcone:
@felixcheung no readers cannot run, that's why I made a new type of user to enable the possibility of running a notebook without modifying its content. 

This feature also avoids changing rights to readers

felixcheung:
I see. In the case someone is upgrading from the current version of Zeppelin with an existing notebook with Readers role set, would they run into trouble that those users can no longer run the notebook?

gfalcone:
@felixcheung nothing would change for them because readers cannot run the notebook. The only rights allowed to run a notebook are writers and owners

felixcheung:
ok, sorry for dragging it on... then I'll flip the question, are Writers also Runners then?
In the case someone is upgrading from the current version of Zeppelin with an existing notebook with ~~Readers~~ Writers role set, would they run into trouble that those users can no longer run the notebook?

gfalcone:
@felixcheung no problem :), so yes writers can run in the current version, but since I didn't want to modify how writers and readers behave, I created an in-between user

felixcheung:
I see, so to summarize, before:
```
Readers
Writers
Owners
```

After
```
Readers
Writers
Runners
Owners
```
where Writers are also Runners

I think it'll be great to get some community feedback on whether this is a good model to go forward.
Open to others to comment?

Leemoonsoo:
Tested this branch and it works well.

Looks good to me.

docs/setup/security/notebook_authorization.md
```diff
@@ -36,13 +36,15 @@ As you can see, each Zeppelin notebooks has 3 entities :
...
+* Runners ( users or groups )
 
 <center><img src="{{BASE_PATH}}/assets/themes/zeppelin/img/docs-img/permission_setting.png"></center>
 
 Fill out the each forms with comma seperated **users** and **groups** configured in `conf/shiro.ini` file.
 If the form is empty (*), it means that any users can perform that operation.
 
-If someone who doesn't have **read** permission is trying to access the notebook or someone who doesn't have **write** permission is trying to edit the notebook, Zeppelin will ask to login or block the user.
+If someone who doesn't have **read** permission is trying to access the notebook or someone who doesn't have **write** permission is trying to edit the notebook,
+or someone who doesn't have **run** permission is trying to run a paragraph Zeppelin will ask to login or block the user.
```
felixcheung: let's add that "writers" can run too

felixcheung: or owner too?

gfalcone: Owners too

gfalcone: @felixcheung I updated the doc to explain the mechanism

gfalcone:
@felixcheung CI ok https://travis-ci.org/gfalcone/zeppelin/builds/266447940?utm_source=email&utm_medium=notification

gfalcone:
@felixcheung do you need me for something?

zjffdu:
Agree with this @felixcheung , linux permission mode : `read + write + execute` seems more suitable and understable. We should be very careful to deliver this kind of feature, because once we deliver it, it is better not to change it to keep compatibility and user experience.

gfalcone:
@zjffdu it would have been a good idea to do that but rights on zeppelin are different than rights on a Unix system. For example, a user that has write permission can write, read and execute the note. So the rights are not properly isolated and I think this should bring another type of discussion to whether isolate the rights or not in order to make it more similar to a Unix system

zjffdu:
I think about the problem again. And wondering would only `read + write` be suffient for zeppelin ?
And `read` means user can open the notebook and run it, but can not change the code. Because there's one scenario that require reader to run the notebook. E.g. We build a dashboard in zeppelin which allow user to interact with the dashboard (with textbox and dropdown list), then as a reader, he may want to interact with the dashboard via changing value of textbox of dropdown list. If we don't allow reader to run the notebook, I think user would be confused and doesn't make sense to me. And introducing Runner role for this scanrio seems a little over design to me. 

@gfalcone @felixcheung @Leemoonsoo what do you think ?

1ambda:
IMO introducing `EXECUTE` permission makes sense. For instance.

1. A shared notebook (dashboard) contains important indices and it requires large query executions. Thus I don't want to allow readers to change dynamic forms (dropdown). **It will be used only view-mode**

2. I have another notebook made for cooperation between UI team and developer team. the developer team wrote queries and UI team will change dynamic forms and execute queries. But the developer also wants to allow a manager of the project sees the notebook but not execute.

for case (1), we need READ-ONLY permissions.
for case (2). we need to support READ-ONLY + EXECUTE permissions.

**introducing `EXECUTE` permission gives Zeppelin users flexibility can be applied to multiple use cases**

zjffdu:
@1ambda Thanks for clarfiying these scenarios.

gfalcone:
@1ambda @zjffdu @felixcheung do you still need clarifications or modifications from me?

docs/setup/security/notebook_authorization.md
```diff
@@ -36,13 +36,17 @@ As you can see, each Zeppelin notebooks has 3 entities :
...
+
+By default, owners and writers have **write** permission, owners, writers and runners have **run** permission, owners, writers, runners and readers have **read** permission
```
felixcheung: I think it is confusing to mix in the roles with the permissions, just "permission" isn't something someone can set or change.
my vote would be to remove the wording of "permission" and just can owners and writers can write, can run and so go
but seeing the wording of "permission" is there before, I am fine leaving it as-is.
I think also a table would be better to explain the roles and permissions they have

felixcheung:
any more comment?

felixcheung:
@gfalcone do you have a JIRA account? we could assign this to you

gfalcone:
@felixcheung thanks for merging. Yes I have a jira account, same nickname as github's

felixcheung:
@gfalcone Could you check if it is an account on https://issues.apache.org/jira/projects/ZEPPELIN?
It couldn't find gfalcone or falcone

gfalcone:
@felixcheung this might help : 
https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gfalcone&selectedTab=jira.user.profile.panels:user-profile-summary-panel

PR #2529, opened 2017-08-14
[ZEPPELIN-1515] Use Hadoop-FileSystem api to implement NotebookRepo

prabhjyotsingh:
### What is this PR for?
With this PR, I attempt to use Hadoop-filesystem API to implement `NotebookRepo`, which in my opinion is very generic in nature, as with this we will be able to support multiple scheme of the URI like local file system `file:///`, Hadoop file system `hdfs://localhost:8020`, S3 etc. So far I have tested this only with local and HDFS.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Periodic refresh of kerberos ticket.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1515

In zeppelin-site.xml enable these two configs; to write into HDFS.

```
<property>
  <name>zeppelin.notebook.storage</name>
  <value>org.apache.zeppelin.notebook.repo.HDFSNotebookRepo</value>
  <description>notebook persistence layer implementation</description>
</property>

<property>
  <name>zeppelin.hadoop.uri</name>
  <value>hdfs://localhost:8020</value>
  <description>The scheme of the URI determines a configuration property name, fs.scheme.class whose value names the FileSystem class.
  e.g. `hdfs://localhost:8020`, or `file:///`
  </description>
</property>
```

### Questions:


prabhjyotsingh:
@zjffdu  @jongyoul @felixcheung @halidhuseynov Please help review this.

prabhjyotsingh:
PING

conf/zeppelin-site.xml.template
```diff
@@ -173,6 +173,47 @@
...
+<property>
+  <name>zeppelin.hadoop.uri</name>
+  <value>hdfs://localhost:8020</value>
+  <description>The scheme of the URI determines a configuration property name, fs.scheme.class whose value names the FileSystem class.
+  e.g. `hdfs://localhost:8020`, or `file:///`
+  </description>
+</property>
+
+<property>
+  <name>kerberos.principal</name>
```
felixcheung: should this be prefix with `zeppelin.hadoop`? there might be other uses of kerberos keytab/principal?

felixcheung: like the key refresh stuff for shell interpreter...

conf/zeppelin-site.xml.template
```diff
@@ -173,6 +173,47 @@
...
+  <name>zeppelin.notebook.storage</name>
+  <value>org.apache.zeppelin.notebook.repo.HDFSNotebookRepo</value>
+  <description>notebook persistence layer implementation</description>
+</property>
+
+<property>
+  <name>zeppelin.hadoop.uri</name>
+  <value>hdfs://localhost:8020</value>
+  <description>The scheme of the URI determines a configuration property name, fs.scheme.class whose value names the FileSystem class.
+  e.g. `hdfs://localhost:8020`, or `file:///`
```
felixcheung: I guess this might be confused with webhdfs url? I'm not sure if there is a better name for it than `hadoop.uri`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HDFSNotebookRepo.java
```diff
@@ -0,0 +1,503 @@
...
+          + " defaulting to 1d ", e);
+      refreshInterval = getTimeAsMs("1d");
+    }
+
+    return refreshInterval;
+  }
+
+  private Long getTimeAsMs(String time) {
+    if (time == null) {
+      LOG.error("Cannot convert to time value.", time);
```
felixcheung: does this print out `null` properly?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HDFSNotebookRepo.java
```diff
@@ -0,0 +1,503 @@
...
+    } catch (IllegalArgumentException e) {
+      LOG.error("Cannot get time in MS for the given string, " + refreshIntervalString
+          + " defaulting to 1d ", e);
+      refreshInterval = getTimeAsMs("1d");
+    }
+
+    return refreshInterval;
+  }
+
+  private Long getTimeAsMs(String time) {
```
felixcheung: would be good to add some tests for this function

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HDFSNotebookRepo.java
```diff
@@ -0,0 +1,503 @@
...
+
+        if (runKerberosLogin()) {
+          LOG.info("Ran runKerberosLogin command successfully.");
+          kinitFailCount = 0;
+          // schedule another kinit run with a fixed delay.
+          scheduledExecutorService
+              .schedule(this, getKerberosRefreshInterval(), TimeUnit.MILLISECONDS);
+        } else {
+          kinitFailCount++;
+          LOG.info("runKerberosLogin failed for " + kinitFailCount + " time(s).");
```
felixcheung: LOG.warn?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HDFSNotebookRepo.java
```diff
@@ -0,0 +1,503 @@
...
+    }
+    return getNotebookDirPath();
+  }
+
+  private Path getPath(String path) {
+    String absolutePath;
+    if (path == null || path.trim().length() == 0) {
+      absolutePath = fileSystem.getHomeDirectory().toString();
+    } else if (path.startsWith("/")) {
+      absolutePath = fileSystem.getHomeDirectory().toString() + path;
```
felixcheung: if a HDFS path starts with `/`, eg. `/root` - shouldn't the absolutePath be `/root`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/HDFSNotebookRepo.java
```diff
@@ -0,0 +1,503 @@
...
+  @Override
+  public List<NoteInfo> list(AuthenticationInfo subject) throws IOException {
+    FileStatus[] children = fileSystem.listStatus(getRootDir());
+
+    List<NoteInfo> infos = new LinkedList<>();
+    for (FileStatus f : children) {
+      String fileName = f.getPath().getName();
+      if (fileName.startsWith(".")
+          || fileName.startsWith("#")
+          || fileName.startsWith("~")) {
```
felixcheung: shouldn't this be `_` and `.` only
https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapred/FileInputFormat.java#L92

Leemoonsoo:
How about the compatibility of Hadoop filesystem API against different hadoop versions?
Does it work with various versions without rebuild Zeppelin?

zjffdu:
@prabhjyotsingh What's the difference of this PR compared with #2455 ?

prabhjyotsingh:
@zjffdu yes you are right, I originally thought I have a different idea, but now I see, this will eventually be that. Will close this then.

PR #2539, opened 2017-08-15
[ZEPPELIN-2863] change notebook-authorization.json and interpreter.json to NotebookDir

sadhen:
### TODO
+ [x] Fix travis (have tries several times and it seems the error is not relevant)
+ [ ] backward compatibility 

### What is this PR for?
When dockerizing Zeppelin, we need to persist the mutable part of Zeppelin on storage out of docker (for example NFS).

For this reason, notebook-authorization.json and interpreter.json should be persisted in Notebook Dir.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2863

Minor change, no need to test it.

### Questions:
No
No
Yes


1ambda:
Hi, Thanks for the contribution. @sadhen 

I think you can manually move `interpreter*.json` inside the docker container. 
Isn't it?

sadhen:
@1ambda  `interpreter.json` and `notebook-authorization.json` will be changed during the use of Zeppelin. If the docker instance is killed and restarted, the updated json files will be lost. 

And logically the two json files are related to the note.json, and should be kept along side with them.

Currently, this files are located under conf/ ,but other files under conf will not be changed thus we do not have to put them into the mounted NFS.

1ambda:
@sadhen 

1. If the docker instance is killed and restarted, the updated json files will be lost.

You can persist files using volumn 

- https://docs.docker.com/engine/admin/volumes/volumes/

2.

sadhen:
We deployed Zeppelin in K8S cluster, and are using volumes for persistence as you stated.

A1. I will provide a solution later. A possible solution: mv the files in `conf/` if exists in `zeppelin.sh`.

A2. Although `notebook` dir is version controlled, it is not relevant to this particular issue. Untracked files also lives well under version(git) controlled dir.

-----

The critical part is that files under `conf/` such as `shiro.ini` is part of the docker image and should be immutable. But `notebook-authorization.json` and `interpreter.json` is mutable and is changed in user activities. They are different in nature.

The differences:
1. `shiro.ini` is provided by admin and these two json is generated by Zeppelin itself
2. `shiro.ini` should be part of the docker image, two docker images with totally different `shiro.ini` behaves in a totally different way. It is part of the deployment but the json files are record of user activities.

Why should we keep these in different dirs?
Since volumes are dirs. And the two json files should be kept in a volume.

zjffdu:
`
The critical part is that files under conf/ such as shiro.ini is part of the docker image and should be immutable.
`
I don't feel it is a good idea to move notebook-authorization.json and interpreter.json to notebook_dir just for docker. One workaround for you is that you can configure ZEPPELIN_CONF_DIR to NOTEBOOK_DIR for your docker image.

sadhen:
@zjffdu 

Your solution didn't solve the problem in the quoted:

`The critical part is that files under conf/ such as shiro.ini is part of the docker image and should be immutable.`

Imaging the following use case, I have a base docker image, and two other end-point docker images. The base docker image consists of jars and a default configuration. And the end-point docker images just config the base image, that means for example image0 use CAS auth in `shiro.ini` and image1 use LDAP in `shiro.ini`. The immutable part of a deployment identifies a docker image. That is the charming part of docker. We should not modify the conf files in a mounted volume. If you keep this files in a mounted volume,  you will never correctly start up zeppelin with ill configurations.

This is my understanding of dockerization. If you don't agree with me, please downvote and I will not waste my time completing this PR.

-----

TODOs:
1. provide an environment variable to tell Zeppelin where is the notebook dir. The variable should be prior to configs in `zeppelin-site.xml` (regardless of this pr, this would be nice feature for dockerization)
2. add some logic in `bin/*sh` for compat.

zjffdu:
`provide an environment variable to tell Zeppelin where is the notebook dir. The variable should be prior to configs in zeppelin-site.xml (regardless of this pr, this would be nice feature for dockerization)
`
You can use set env varaible `ZEPPELIN_CONF_DIR` for that.

zjffdu:
@sadhen would set env variable `ZEPPELIN_CONF_DIR` to a folder out of docker image work for you ?

sadhen:
@zjffdu 
It won't satisfy my need.

You may understand my concerns when you start dockerizing Zeppelin.

And these days, I encounter another dockerization problem. If I put interpreter and server processes all in one container, the interpreter process will run out of memory. My final solution is deploying interpreters elsewhere.

```
ERROR [2017-08-24 00:12:17,391] ({Thread-0} TThreadPoolServer.java[serve]:194) - ExecutorService threw error: java.lang.OutOfMemoryError: unable to create new native thread
java.lang.OutOfMemoryError: unable to create new native thread
        at java.lang.Thread.start0(Native Method)
        at java.lang.Thread.start(Thread.java:714)
        at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:950)
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1368)
        at org.apache.thrift.server.TThreadPoolServer.serve(TThreadPoolServer.java:168)
        at org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer.run(RemoteInterpreterServer.java:104)
 INFO [2017-08-24 00:24:30,979] ({Thread-0} RemoteInterpreterServer.java[run]:103) - Starting remote interpreter server on port 40623
```

zjffdu:
Can you explain a bit more why setting `ZEPPELIN_CONF_DIR` to outside of docker image won't work for you ? I am not so familiar with docker, may need more context to understand that.

There's one in-progress PR to put conf folder in hdfs, would that work for you ?

There's another in-progress PR to support to run interpreter process in a separate machine.

PR #2547, opened 2017-08-22
Fix trivial typo

byung-u:
### What is this PR for?
Fixes a trivial typo in travis_check.py.

The semicolon does nothing in travis_check.py.
Python use semicolon when several statements on the same line, here is [python doc.](https://docs.python.org/3/reference/compound_stmts.html)

### Questions:


tae-jun:
LGTM! 👍

1ambda:
LGTM merge if no more discussion.

PR #2545, opened 2017-08-22
ZEPPELIN-2873 - Add documentation on secure cookie in Shiro

VipinRathor:
### What is this PR for?
Adding a section in Shiro Authentication about how to enable secure cookie via Shiro. Shiro do support configuring 'HttpOnly' flag in response cookie. A Zeppelin user, who is security conscious, should know how to enable this in Zeppelin's Shiro configuration.

### What type of PR is it?
Documentation

### What is the Jira issue?
ZEPPELIN-2873

### How should this be tested?
Doc changes. CI test should pass.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #2550, opened 2017-08-24
[MINOR] Updated shiro.ini.template to include secure cookie option

VipinRathor:
### What is this PR for?
Based on discussion in https://github.com/apache/zeppelin/pull/2545 , I'm updating the shiro.ini.template to include secure cookie option. With this change, Zeppelin Shiro will always set 'HttpOnly' flag in cookie. This will help to prevent majority of cross-site scripting (XSS) attacks.

### What type of PR is it?
Minor Improvement

### What is the Jira issue?
Minor change in shiro.ini

CI tests should pass

### Questions:
* Does this needs documentation? Doc changes already done in https://github.com/apache/zeppelin/pull/2545

### DIFF
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -62,10 +62,12 @@ sessionManager = org.apache.shiro.web.session.mgt.DefaultWebSessionManager
 #cacheManager = org.apache.shiro.cache.MemoryConstrainedCacheManager
 #securityManager.cacheManager = $cacheManager
 
+### Enables 'HttpOnly' flag in Zeppelin cookies
 cookie = org.apache.shiro.web.servlet.SimpleCookie
 cookie.name = JSESSIONID
-cookie.secure = true
 cookie.httpOnly = true
+### Uncomment the below line only when Zeppelin is running over HTTPS
+#cookie.secure = true
 sessionManager.sessionIdCookie = $cookie
 
 securityManager.sessionManager = $sessionManager
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -62,6 +62,12 @@ sessionManager = org.apache.shiro.web.session.mgt.DefaultWebSessionManager
 #cacheManager = org.apache.shiro.cache.MemoryConstrainedCacheManager
 #securityManager.cacheManager = $cacheManager
 
+cookie = org.apache.shiro.web.servlet.SimpleCookie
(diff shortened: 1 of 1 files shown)


krishna-pandey:
@VipinRathor Can you please update the description from "Updated shiro.init.template ..." to "Updated shiro.ini.template ...".

conf/shiro.ini.template
```diff
@@ -62,6 +62,12 @@ sessionManager = org.apache.shiro.web.session.mgt.DefaultWebSessionManager
 #cacheManager = org.apache.shiro.cache.MemoryConstrainedCacheManager
 #securityManager.cacheManager = $cacheManager
 
+cookie = org.apache.shiro.web.servlet.SimpleCookie
+cookie.name = JSESSIONID
+cookie.secure = true
```
krishna-pandey: Also, I am not sure if you want to enable "secure" flag by default. As per OWASP, "Browsers which support the secure flag will not send a cookie with the secure flag set over an unencrypted HTTP request".  Refer (https://www.owasp.org/index.php/SecureFlag). This might cause an issue with installations where Zeppelin is running on plain HTTP.

However, httpOnly flag can be set by default and won't have any implications in both HTTP and HTTPS enabled Zeppelin.

In short comment this line.

krishna-pandey:
@VipinRathor Also, I think it will be a good idea to mention an one-liner in the shiro.ini.template "Enable 'cookie.secure = true' only when Zeppelin is running on HTTPS" or something similar. 

If enabled when Zeppelin is running on HTTP, Browser does not set the cookie and hence user session won't be maintained.

VipinRathor:
@krishna-pandey You are right. Setting "secure=true" in cookie will force cookie to travel only over HTTPS. When Zeppelin is running over HTTP, browsers won't send this 'secure' cookie back. So Zeppelin will generate new session id every time when a request is made, therefore user sessions will not be maintained. I've commented that line and added comment too.

VipinRathor:
@felixcheung Thanks for reviewing. Zeppelin, by default, should set 'HttpOnly' flag for all session cookies as security measure against XSS attacks. That's why we have decided to include that in shiro.ini.template as default setting.
Let me know if you have any concerns over this.

krishna-pandey:
LGTM!

Leemoonsoo:
LGTM

prabhjyotsingh:

PR #2551, opened 2017-08-25
[ZEPPELIN-2880] - Fix username output when OIDC is enabled

byamthev:
### What is this PR for?
When OIDC is enabled, user's roles/permissions/tokens are sent to Zeppelin's client via websocket and appears in the web browser when running a paragraph in a notebook.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2880](https://issues.apache.org/jira/browse/ZEPPELIN-2880)

Enable OIDC , login to Zeppelin and run a paragraph.  Check the text "Last updated by". 

### Questions:
* Does the licenses files need update?  N/A
* Is there breaking changes for older versions?  N/A
* Does this needs documentation?  N/A

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
@@ -20,6 +20,7 @@ import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
+import java.security.Principal;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -88,9 +89,20 @@ public class SecurityUtils {
 
     String principal;
     if (subject.isAuthenticated()) {
-      principal = subject.getPrincipal().toString();
+      principal = extractPrincipal(subject);
     } else {
       principal = ANONYMOUS;
+    }
+    return principal;
+  }
+
+  private static String extractPrincipal(Subject subject) {
+    String principal;
+    Object principalObject = subject.getPrincipal();
+    if (principalObject instanceof Principal) {
+      principal = ((Principal) principalObject).getName();
+    } else {
+      principal = String.valueOf(principalObject);
     }
     return principal;
   }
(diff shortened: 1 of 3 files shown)


andreaTP:
this is possible since this PR from @volumeint to bujiio:
https://github.com/bujiio/buji-pac4j/pull/63

byamthev:
Yes, I have now added a JUnit test.

byamthev:
ping @felixchang

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -88,13 +89,24 @@ public static String getPrincipal() {
...
     return principal;
   }
 
+  private static String extractPrincipal(Subject subject) {
+    String principal;
+    Object principalObject = subject.getPrincipal();
+    if (principalObject instanceof Principal) {
+      principal = ((Principal) principalObject).getName();
+    } else {
+      principal = String.valueOf(principalObject);
```
felixcheung: what's the reason for `String.valueOf` vs `.toString()` from before?

andreaTP:
> should we consider returning/displaying more than just the principal name?
now the name field is configurable via `shiro.ini`, basically you can return different values in that field depending on configuration (e.g. look at tests https://github.com/bujiio/buji-pac4j/pull/63/files)

felixcheung:
I see, it would be great to add some example in /docs/setup/security?

andreaTP:
@felixcheung I think that docs improvements have sense in this case only on top of #2552 , could post-pone them?

felixcheung:
ok, how about this question
https://github.com/apache/zeppelin/pull/2551/files/cb6ec7047a8abc1a79da3c1261178e1a9e0b4f17#diff-0d6ccf388c8aa67062870388536ed274

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
byamthev: That is because if the argument is null, then a string equal to "null" is returned instead of NullPointerException (just in case).

felixcheung:
LGTM
any more comment?

andreaTP:
ping, probably this is good to merge

felixcheung:
merged, do you have a JIRA user on ASF JIRA, @byamthev ?

PR #2555, opened 2017-08-28
[ZEPPELIN-2885] Have Logger subclass StringIO and override write method

sctincman:
### What is this PR for?
H2O python module uses the `isatty` method when outputting to stdout. As Zeppelin replaces stdout with a custom Logger class, the lack of this method causes H2O to fail with most operations.

Subclassing StringIO and overriding the write method fixes this ensures other behavior in the TextIO classes used for stdout is present as well, obviating the need to keep adding method stubs to workaround other modules in the future.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Does `reset` still need to be present?

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2885

* Python/Pyspark interpreter output still works
* Running the following paragraph should return `False`

```
%python
import sys
sys.stdout.isatty()
```

### Questions:
* Does the licenses files need update?			No
* Is there breaking changes for older versions?		No
* Does this needs documentation?      			No

### DIFF
--- a/python/src/main/resources/python/zeppelin_python.py
+++ b/python/src/main/resources/python/zeppelin_python.py
@@ -34,7 +34,11 @@ except ImportError:
 
 # for back compatibility
 
-class Logger(StringIO):
+class Logger(StringIO, object):
+  def __init__(self, *args, **kwargs):
+    if sys.version_info < (3, 0):
+      self.encoding = None
+    super(Logger, self).__init__(*args, **kwargs)
   def write(self, message):
     intp.appendOutput(message)
     StringIO.write(self, message)
--- a/python/src/main/resources/python/zeppelin_python.py
+++ b/python/src/main/resources/python/zeppelin_python.py
@@ -34,19 +34,12 @@ except ImportError:
 
 # for back compatibility
 
-class Logger(object):
-  def __init__(self):
-    pass
-
+class Logger(StringIO):
   def write(self, message):
     intp.appendOutput(message)
-
+    StringIO.write(self, message)
   def reset(self):
     pass
-
-  def flush(self):
-    pass
-
 
 class PyZeppelinContext(object):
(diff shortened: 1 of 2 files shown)


sctincman:
As it turns out, we do need to inherit from `object` in order to keep the class a new-style python class in python 2. I had mistakenly thought `StringIO.StringIO` was a new-style class, but turns out it is not.

There was another python2 problem I came across (`StringIO.StringIO` is missing the `encoding` attribute, which was needed).

I had attempted to have `Logger` extend `TextIOWrapper`, but this caused a lot of problems with python2's byte-based `str` class, and extending `BytesIO` caused problems in python3 due to their unicode-based strings.

Leemoonsoo:
@sctincman could you try restart a failed CI build job? https://travis-ci.org/sctincman/zeppelin/jobs/270477925 I think it's flaky test but just in case.

felixcheung:
thanks we definitely need it to work for python 2 and 3.
could you also look into adding some tests - i think we have interpreter tests for python

sctincman:
@Leemoonsoo Done. Appears a few of the tests were timing out?
@felixcheung Can do :]

Leemoonsoo:
Checked tests are green. Is this one good to go?

felixcheung:
think we should have some tests for this

Tagar:
@sctincman Did you find a workaround for this issue ? Thanks

Tagar:
Found this works around the issue on Python 2.. will test more.

```
import sys
sys.stdout.isatty = lambda : False
sys.stdout.encoding = None
```

Tagar:
FYI 
This Zeppelin issue will be worked around in Sparkling Water in following PR - https://github.com/h2oai/h2o-3/pull/1705

felixcheung:
hi any update on this?

PR #2558, opened 2017-09-01
sc.setLocalProperty(...) should be more deterministic

ruseel:
### What is this PR for?
executing these two pagragraph in spark interperter 
```
sc.setLocalProperty("a", "1")
```

```
sc.getLocalProerty("a") 
```

evaluated to unexpected result "null". fix that confusion.

---

sc.setLocalProperty(...) should be more deterministic.

and honor design of SparkContext. SparkContext.setLocalProperty(...) is using ThreadLocal. So let me say that design of SparkContext has assumtion single-thread user.

But before this commit, SparkInterpreter's ExecutorService is created with Executor.newSchedulerService(100). So User might perceive situation like sc.setLocalProperty(...) is not working.

fix by using singleThread for Schduler.

### What type of PR is it?
Bug Fix

make note with two paragraph 

```
sc.setLocalProperty("a", "1")
```

```
sc.getLocalProerty("a") 
```

run paragraph separatly

### DIFF
--- a/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -137,10 +137,13 @@ public class SparkInterpreter extends Interpreter {
   private Object classServer;      // classserver for scala 2.11
   private JavaSparkContext jsc;
   private boolean enableSupportedVersionCheck;
+  private ExecutorService singleThreadedExecutorSerivce;
 
   public SparkInterpreter(Properties property) {
     super(property);
     out = new InterpreterOutputStream(logger);
+
+    this.singleThreadedExecutorSerivce = createSingleThreadExecutorService();
   }
 
   public SparkInterpreter(Properties property, SparkContext sc) {
@@ -149,6 +152,14 @@ public class SparkInterpreter extends Interpreter {
     this.sc = sc;
     env = SparkEnv.get();
     sparkListener = setupListeners(this.sc);
+
+    this.singleThreadedExecutorSerivce = createSingleThreadExecutorService();
+  }
+
+  private ExecutorService createSingleThreadExecutorService() {
+    return new ThreadPoolExecutor(1, 1, 0, TimeUnit.NANOSECONDS,
+            new ArrayBlockingQueue<Runnable>(100),
+            new ThreadPoolExecutor.AbortPolicy());
   }
 
   public SparkContext getSparkContext() {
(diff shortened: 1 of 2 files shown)


swkimme:
@Leemoonsoo Let me introduce you a new guy on Between, @ruseel (Stephen) 
We found a bug while looking into Spark's scheduler, would u review this PR?

1ambda:
Hi, @ruseel! Thanks for the contribution.

Could u setup travis CI your forked zeppelin repositiory? 

Please setup by swich on 'zeppelin' repository at https://travis-ci.org/profile and travis-ci.
And then make sure 'Build branch updates' option is enabled in the settings https://travis-ci.org/ruseel/zeppelin/settings.

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -149,6 +152,14 @@ public SparkInterpreter(Properties property, SparkContext sc) {
     this.sc = sc;
     env = SparkEnv.get();
     sparkListener = setupListeners(this.sc);
+
+    this.singleThreadedExecutorSerivce = createSingleThreadExecutorService();
+  }
+
+  private ExecutorService createSingleThreadExecutorService() {
+    return new ThreadPoolExecutor(1, 1, 0, TimeUnit.NANOSECONDS,
```
zjffdu: Why not using `Executors.newSingleThreadExecutor()`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/SchedulerFactory.java
```diff
@@ -63,11 +64,15 @@ public void destroy() {
   }
 
   public Scheduler createOrGetFIFOScheduler(String name) {
+    return createOrGetFIFOScheduler(name, this.executor);
```
zjffdu: This would affect all the other interpreters. Since this is only for spark interpreter, we can create a specific scheduler for spark interpreter, and not change the existing `createOrGetFIFOScheduler`, `createOrGetParallelScheduler` and `createOrGetRemoteScheduler`

zjffdu:
What about `PySparkInterpreter` and `SparkRInterpreter` ?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
ruseel: because it creates `unbounded queue` and I though having an unbounded queue is bad thing.

ruseel: If I have an unbounded queue and If I use unbounded queue incorrectly and add huge amount of work to queue, I would notice that symptom only with memory usage.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/SchedulerFactory.java
ruseel: I see. I will make another pull request by without touching existing function.

ruseel:
@1ambda setup for forked repository `ruseel/zeppelin` is completed.

1ambda:
@ruseel Thanks for the update. Could you resolve zjffdu's last comment?

felixcheung:
hi - where are we on this?

PR #2487, opened 2017-07-12
[ZEPPELIN-2769] Prevent SQL injection for GetUserList.getUserList.

yanboliang:
### What is this PR for?
Prevent SQL injection for ```GetUserList.getUserList```.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2769

PR #2559, opened 2017-09-01
[ZEPPELIN-2894] Show users in notebook permission using Shiro JDBC

janusd:
### What is this PR for?
Show user list/suggestions in the notebook permission form when using Shiro and JDBC Realm.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-2894](https://issues.apache.org/jira/browse/ZEPPELIN-2894)

- Shiro with JDBC Realm (e.g. PostgreSQL JDBC Driver)
- Login to any account
- Open Notebook permission form
- Try to get any user suggestion in the dropdown menu by typing an existing name  

### Screenshots 
**After:**
userlist_working

**Before:**
userlist_error

### Questions:
No.
No.
No.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/GetUserList.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/GetUserList.java
@@ -231,7 +231,7 @@ public class GetUserList {
         return userlist;
       }
 
-      userquery = "select ? from ?";
+      userquery = String.format("SELECT %s FROM %s", username, tablename);
 
     } catch (IllegalAccessException e) {
       LOG.error("Error while accessing dataSource for JDBC Realm", e);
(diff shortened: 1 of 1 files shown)


1ambda:
Hi, @janusd 

Could you check https://github.com/apache/zeppelin/pull/2487/files? which was merged to prevents SQL injection.

janusd:
Thanks @1ambda for your reply. Please see the linked JIRA Issue: https://issues.apache.org/jira/browse/ZEPPELIN-2894 where I describe the problem with that pull request you mentioned.

1ambda:
Read the description and LGTM. 

> ZEPPELIN-2769 introduced a mechanism to prevent SQL injection, but unfortunately table names can not be parameterised in PreparedStatements. Also the column variable "username" might be interpreted as a quoted string and the final list would contain x times "username" instead of the real names (see Figure).
> Other solutions preventing SQL injections mostly rely on other libraries (e.g. escaping) or assumptions (e.g. widely database access). 
I would consider to revert the changes. The SQL statement for getting the user list should not be a security threat as the query parameters will be parsed server-sided from the authenticationQuery, no user input will be provided at all.

1ambda:
CI is green, merge if no more discussion.

- https://travis-ci.org/janusd/zeppelin/builds/270815865

PR #2560, opened 2017-09-02
[ZEPPELIN-2876]. Integrate with Livy 0.4.0

zjffdu:
### What is this PR for?

This is to integrate livy 0.4.0 which is livy's first apache release. Since the package name is changed, I can not support both livy 0.3.0 and 0.4.0 in the integration test, but since the rest api is not changed. Ideally livy before 0.4.0 should still work. 

Regarding the integration test, unfortunately I have to comment out some test due to livy's issue, I will enable it after next livy version release. 

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2876

integration test is added

### Questions:


prabhjyotsingh:
feasibility ask: Like we support python2, and python3 do you think it makes sense to have 2 different interpreters (at-least for now) one that supports Livy-0.2.0/0.3.0 and other as Livy-4?

The reason I'm asking this is because what I see in this PR is, this PR is trying to download Livy-0.4.0 for integration/unit tests but most of the integration tests are turned off.

zjffdu:
Actually this livy interprete support both livy 0.2.0 / 0.3.0 /0.4.0, because livy's rest api doesn't change, only its java package name changes. So I can not add integration test for both of these versions.

zjffdu:
Will commit if no more discussion.

prabhjyotsingh:
Sure yes, rest all LGTM

PR #2561, opened 2017-09-03
ZEPPELIN-2893. Pass non spark.* of spark interpreter properties to spark app driver

zjffdu:
### What is this PR for?
Straightforward change for allowing pass non spark.* properties to spark app driver.  See discussion here http://apache-zeppelin-users-incubating-mailing-list.75479.x6.nabble.com/Setting-spark-config-properties-in-Zeppelin-0-7-2-td6143.html

### What type of PR is it?
[Bug Fix ]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2893

Unit test is added

### Questions:

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -196,6 +196,13 @@ Staring from 0.6.1 SparkSession is available as variable `spark` when you are us
 
 <a name="dependencyloading"> </a>
 
+### How to pass property to SparkConf
+
+There're 2 kinds of properties that would be passed to SparkConf
+
+ * Standard spark property (prefix with `spark.`). e.g. `spark.executor.memory` will be passed to `SparkConf`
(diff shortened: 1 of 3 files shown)


zjffdu:
@felixcheung Please help review

felixcheung:
for example, Spark passes on `spark.hadoop.foo` as `foo` to hadoop/hive conf, which I think it's a better approach?
https://github.com/apache/spark/blob/3f958a99921d149fb9fdf7ba7e78957afdad1405/sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveUtilsSuite.scala#L41

we could pass `zeppelin.spark.foo` as `foo` to Spark (and `spark.*`, assuming foo != spark)

zjffdu:
hmm, seems a better approach. Update the PR

zjffdu:
Will merge if no more discussion

PR #2564, opened 2017-09-04
[ZEPPELIN-2896] Replacing addHeader with setHeader method in CorsFilter.java

krishna-pandey:
### What is this PR for?
HTTP Response Headers were being added multiple times. Replacing addHeader method with setHeader overrides the Response Header value with new/existing value instead of adding another duplicate response Header.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2896

Open the Zeppelin URL in Chrome Browser. Select "More Tools" -> "Developer Tools" from the right-side menu. Under Network Section, select the request with name "localhost" and check for "Response Headers". You should see response headers appearing only once.

screen shot 2017-09-04 at 3 21 32 pm

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/CorsFilter.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/CorsFilter.java
@@ -73,19 +73,19 @@ public class CorsFilter implements Filter {
   }
 
   private void addCorsHeaders(HttpServletResponse response, String origin) {
-    response.addHeader("Access-Control-Allow-Origin", origin);
-    response.addHeader("Access-Control-Allow-Credentials", "true");
-    response.addHeader("Access-Control-Allow-Headers", "authorization,Content-Type");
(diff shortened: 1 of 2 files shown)


krishna-pandey:
@1ambda, @Leemoonsoo, @felixcheung, @jongyoul, @prabhjyotsingh Please help review this.

jongyoul:
LGTM

krishna-pandey:
@felixcheung IMHO, it happens coz we first hit the root Context Path and Jetty redirecting it later to the landing/welcome page i.e. index.html. These headers are duplicated only for such requests and not every request.

prabhjyotsingh:

krishna-pandey:
@1ambda @Leemoonsoo Can you please help review this fix? As this is related to [ZEPPELIN-2775] which is getting shipped in 0.7.3, we may want to merge this now and get this as well into RC2.

1ambda:
Tested locally and works well. Thanks for the fix.

1ambda:
@prabhjyotsingh I think we can merge this PR into 0.7.3-RC1 as well. What do u think?

PR #2566, opened 2017-09-05
[ZEPPELIN-2903] Make setting of working directory to user-home optional for shell interpreter

prabhjyotsingh:
### What is this PR for?
With ZEPPELIN-2841, it had changed the default working directory of Shell Interpreter from the relative path where Zeppelin is running to user-home. This is to make the configuration optional.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2903

by default when the user runs shell interpreter and executes `pwd` will the path where Zeppelin server is running, but when `shell.working.directory.user.home` is set to true in Zeppelin's interpreter setting, it will point to the user's home directory by which the interpreter is running.

### DIFF
--- a/docs/interpreter/shell.md
+++ b/docs/interpreter/shell.md
@@ -44,6 +44,11 @@ At the "Interpreters" menu in Zeppelin dropdown menu, you can set the property v
     <td>Shell command time out in millisecs</td>
   </tr>
   <tr>
+    <td>shell.working.directory.user.home</td>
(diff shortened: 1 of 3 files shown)


jongyoul:
I think it's realistic solution we can choose. LGTM

minahlee:
LGTM

prabhjyotsingh:

PR #2574, opened 2017-09-08
[ZEPPELIN-2921 : 0.7x] does not work conda environment in python interpreter

cloverhearts:
### What is this PR for?
It seems that the environment of the python interpreter has changed to the py4j environment, causing problems with the library path and the default environment.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2921

Please run the following command line for each paragraph.
If the module such as scipy is normally imported, it is a success.
`%python.conda create --name Hello6 python=2.7`
`%python.conda activate Hello6`
`%python.conda install seaborn pandas numpy scipy matplotlib`
```
%python
import scipy as sp
import seaborn as sns
```
`%python.conda deactivate`

#### Before
image

#### After
image

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
@@ -135,9 +135,13 @@ public class PythonCondaInterpreter extends Interpreter {
         }
       }
     }
-    python.setCurrentCondaEnvName(envName);
-    python.setPythonCondaLibPath(condaLibraryPath);
-    python.setPythonCommand(binPath);
+
+    File pythonBin = new File(binPath);
+    if (pythonBin.exists() && pythonBin.isFile() && pythonBin.canExecute()) {
+      python.setCurrentCondaEnvName(envName);
+      python.setPythonCondaLibPath(condaLibraryPath);
+      python.setPythonCommand(binPath);
+    }
   }
 
   private void restartPythonProcess() {
--- a/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
@@ -115,21 +115,28 @@ public class PythonCondaInterpreter extends Interpreter {
   private void changePythonEnvironment(String envName)
       throws IOException, InterruptedException {
     PythonInterpreter python = getPythonInterpreter();
+    String condaLibraryPath = null;
     String binPath = null;
     if (envName == null) {
       binPath = getProperty(ZEPPELIN_PYTHON);
       if (binPath == null) {
         binPath = DEFAULT_ZEPPELIN_PYTHON;
       }
+      condaLibraryPath = null;
     } else {
       Map<String, String> envList = getCondaEnvs();
+      for (String key : envList.keySet()) {
+      }
       for (String name : envList.keySet()) {
         if (envName.equals(name)) {
           binPath = envList.get(name) + CONDA_PYTHON_PATH;
+          condaLibraryPath = String.format("%s/lib", envList.get(name));
           break;
         }
       }
     }
+    python.setCurrentCondaEnvName(envName);
+    python.setPythonCondaLibPath(condaLibraryPath);
     python.setPythonCommand(binPath);
   }
 
@@ -200,22 +207,24 @@ public class PythonCondaInterpreter extends Interpreter {
 
   private InterpreterResult runCondaActivate(String envName)
       throws IOException, InterruptedException {
-
+    PythonInterpreter python = getPythonInterpreter();
     if (null == envName || envName.isEmpty()) {
       return new InterpreterResult(Code.ERROR, "Env name should be specified");
     }
 
     changePythonEnvironment(envName);
     restartPythonProcess();
-
+    python.setCurrentCondaEnvName(envName);
     return new InterpreterResult(Code.SUCCESS, "'" + envName + "' is activated");
   }
 
   private InterpreterResult runCondaDeactivate()
       throws IOException, InterruptedException {
 
+    PythonInterpreter python = getPythonInterpreter();
     changePythonEnvironment(null);
     restartPythonProcess();
+    python.setCurrentCondaEnvName(null);
     return new InterpreterResult(Code.SUCCESS, "Deactivated");
   }
 
@@ -255,20 +264,26 @@ public class PythonCondaInterpreter extends Interpreter {
 
   private String runCondaInstall(List<String> restArgs)
       throws IOException, InterruptedException {
-
+    PythonInterpreter python = getPythonInterpreter();
+    String envName = python.getCurrentCondaEnvName();
     restArgs.add(0, "conda");
     restArgs.add(1, "install");
     restArgs.add(2, "--yes");
+    restArgs.add(3, "-n");
+    restArgs.add(4, envName);
 
     return runCondaCommandForTextOutput("Package Installation", restArgs);
   }
 
   private String runCondaUninstall(List<String> restArgs)
       throws IOException, InterruptedException {
-
+    PythonInterpreter python = getPythonInterpreter();
+    String envName = python.getCurrentCondaEnvName();
     restArgs.add(0, "conda");
     restArgs.add(1, "uninstall");
     restArgs.add(2, "--yes");
+    restArgs.add(3, "-n");
+    restArgs.add(4, envName);
 
     return runCondaCommandForTextOutput("Package Uninstallation", restArgs);
   }
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -44,6 +44,7 @@ import org.apache.commons.exec.ExecuteWatchdog;
 import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.exec.environment.EnvironmentUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.display.GUI;
 import org.apache.zeppelin.interpreter.*;
 import org.apache.zeppelin.interpreter.InterpreterResult.Code;
@@ -57,7 +58,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import py4j.GatewayServer;
-import py4j.commands.Command;
 
 /**
  * Python interpreter for Zeppelin.
@@ -76,6 +76,9 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
   private int maxResult;
   private String py4jLibPath;
   private String pythonLibPath;
+
+  private String currentCondaEnvName = StringUtils.EMPTY;
+  private String currentCondaEnvLibPath = StringUtils.EMPTY;
 
   private String pythonCommand;
 
@@ -152,6 +155,10 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
       Path workingPath = Paths.get("..").toAbsolutePath();
       py4jLibPath = workingPath + File.separator + ZEPPELIN_PY4JPATH;
       pythonLibPath = workingPath + File.separator + ZEPPELIN_PYTHON_LIBS;
+    }
+
+    if (currentCondaEnvLibPath != null && !currentCondaEnvLibPath.isEmpty()) {
+      pythonLibPath = String.format("%s:%s", pythonLibPath, currentCondaEnvLibPath);
     }
 
     port = findRandomOpenPortOnAllLocalInterfaces();
@@ -450,6 +457,28 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
     } else {
       return pythonCommand;
     }
+  }
+
+  public void setCurrentCondaEnvName(String envName) {
+    if (envName == null) {
+      envName = StringUtils.EMPTY;
+    }
+    this.currentCondaEnvName = envName;
+  }
+
+  public String getCurrentCondaEnvName() {
+    return currentCondaEnvName;
+  }
+
+  public void setPythonCondaLibPath(String path) {
+    if (path == null) {
+      path = StringUtils.EMPTY;
+    }
+    this.currentCondaEnvLibPath = path;
+  }
+
+  public String getPythonCondaLibPath() {
+    return this.currentCondaEnvLibPath;
   }
 
   public String getPythonBindPath() {


1ambda:
Thanks for the fix. Let me test and feedback soon.

cloverhearts:
@1ambda Thank you, and i will more test :)

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -452,6 +459,28 @@ private String getPythonCommand() {
     }
   }
 
+  public void setCurrentCondaEnvName(String envName) {
```
1ambda: these conda related funcs can be moved into `PythonCondaInterpreter`?

1ambda:
@cloverhearts I am getting this error. 

image

image

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
cloverhearts: I have considered that.
but, I think it is not good for two reasons.

1. `PythonCondaInterpreter` actually executes only` conda` command line.
It has nothing to do with python.
PythonInterpreter.class is actually affected.
In fact, we can posible uninstall for python via Conda.

2. All derived classes, such as other PythonDockerInteprters, including `PythonCondaInterpreter`, do not have state. I am burdened with keeping the conda configuration information in a PythonCondaInterpreter that does not need to maintain it.

cloverhearts:
@1ambda I'll check it, thank you :)

cloverhearts:
```
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 12.456 sec <<< FAILURE! - in org.apache.zeppelin.interpreter.remote.RemoteInterpreterServerTest
testStartStop(org.apache.zeppelin.interpreter.remote.RemoteInterpreterServerTest)  Time elapsed: 10.037 sec  <<< FAILURE!
java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.apache.zeppelin.interpreter.remote.RemoteInterpreterServerTest.testStartStop(RemoteInterpreterServerTest.java:62)
```

This error occurs equally in other PRs and branches.
It has been confirmed that it is not related to my changes.

cloverhearts:
https://travis-ci.org/malayhm/zeppelin/jobs/272788659
https://travis-ci.org/1ambda/zeppelin/jobs/273195461

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
1ambda: @zjffdu Hi, What do you think of?

zjffdu: I agree with @1ambda that we should not couple `PythonCondaInterpreter` & `PythonInterpreter` closely. `PythonInterpter` doesn't need to depend on PythonCondaInterpreter, it just depends on `zeppelin.python` IIUC.

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
```diff
@@ -115,22 +115,33 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
         }
       }
     }
-    python.setPythonCommand(binPath);
+
+    File pythonBin = new File(binPath);
+    if (pythonBin.exists() && pythonBin.isFile() && pythonBin.canExecute()) {
+      python.setCurrentCondaEnvName(envName);
+      python.setPythonCondaLibPath(condaLibraryPath);
+      python.setPythonCommand(binPath);
```
zjffdu: Is `setPythonCondaLibPath` necessary ? Is `setPythonCommand` enough ?

IMO, we could use `python.getProperty().setProperty('zeppelin.python', binPath)` .

felixcheung: ... for binPath
`binPath = python.getProperty().setProperty('zeppelin.python', binPath)`
and then go into this block with `pythonBin = new File(binPath);`

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
```diff
@@ -115,22 +115,33 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
       }
+      condaLibraryPath = null;
     } else {
       Map<String, String> envList = getCondaEnvs();
+      for (String key : envList.keySet()) {
+      }
       for (String name : envList.keySet()) {
         if (envName.equals(name)) {
           binPath = envList.get(name) + CONDA_PYTHON_PATH;
+          condaLibraryPath = String.format("%s/lib", envList.get(name));
```
felixcheung: is this quite standard? is it possible the lib path is not env/lib?

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
cloverhearts: The new conda python environment is not related to `zeppelin.python`.
(If you have a new installation of python using conda)
We can freely change the python environment in a new environment via conda.
Therefore, you must register an additional Python library path for the new environment for zeppelin py4j gateway.
Otherwise you will not be able to use the python library installed via conda.

In version 0.7, there is an add library operation.
but in the current version, it seems that the structure has been changed and it is missing.
I modified it.

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
cloverhearts: Yes, it is guaranteed.
python installed via conda is always guaranteed path.

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
zjffdu: @cloverhearts I don't understand why `you must register an additional Python library path for the new environment for zeppelin py4j gateway`. This is not consistent with the behavior when I use conda in shell. e.g. I create an env using conda in mac terminal, then I just need to set `zeppelin.python` to use this conda env.

cloverhearts: `` `
python.getProperty (). setProperty ('zeppelin.python', binPath)
`` `

This is not the process for setting up Zeppelin Python.
It is the action that changes the environment of Python Interpreter by judging whether the user installs new Python through Conda.

There are two cases.

1. Does not install Python via Conda
User has installed another library via conda.
In this case, the Python environment should not be changed in the default environment of zeppelin.
I checked it out and in the worst case Zeppelin's Python Interpreter is not recovered because of an infinite loop.
It's a defense code for it.

2. Install Python via Conda.
If user has installed a Python environment via Conda,
When conda activate (env name), Zeppelin should change user Python environment to user's new settings.
And if you do a Conda deactivate, Zeppelin should return to the original Zeppelin setting.
Therefore, Zeppelin can not use the binPath setting as follows:
`` `
python.getProperty (). setProperty ('zeppelin.python', binPath)
`` `

cloverhearts: @zjffdu 
Thank you for your opinion,
I present scenarios for environment changes using conda.
Python version and module names are arbitrary, so do not worry.
What is important is change in every environment.

1. Basically Zeppelin's python
python: 2.71
installed library: (none)

use to zeppelin
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 2.7.1
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME
```
```
%python
import myModule
```
result
```
no module myModule
```


2. install conda and new python 
```
%python.conda create --name python3 python=3.4
```
```
%python.conda activate python3
```
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 3.4
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME:%CONDA_ENV%/python3/lib
```
```
%python
import myModule
```
result
```
no module myModule
```
try again
```
%python.conda install myModule
```

```
import myModule
```
result
```
imported.
```

3. restore python environment
```
%python.conda deactivate
```
check for recovered to the original Zeppelin environment.
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 2.7.1
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME
```
```
import myModule
```
result
```
no module myModule
```

All three must be supported by conda.
1. Support to installing and using the python library
2. Support to change the python version
3. Support to configuring non-python environments

old zeppelin does works for this part(0.71).
but, the Python Interpreter has changed significantly since 0.71.
I guess missing the environment part for this.
I modified it.


python env set code
(https://github.com/apache/zeppelin/blob/master/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L151)

cloverhearts:
@zjffdu @felixcheung 
Thank you for your opinions :)
I was left my comment.
Could you check this one?

cloverhearts:
@zjffdu 
Thank you for your opinion,
I present scenarios for environment changes using conda.
Python version and module names are arbitrary, so do not worry.
What is important is change in every environment.

1. Basically Zeppelin's python
python: 2.71
installed library: (none)

use to zeppelin
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 2.7.1
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME
```
```
%python
import myModule
```
result
```
no module myModule
```


2. install conda and new python 
```
%python.conda create --name python3 python=3.4
```
```
%python.conda activate python3
```
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 3.4
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME:%CONDA_ENV%/python3/lib
```
```
%python
import myModule
```
result
```
no module myModule
```
try again
```
%python.conda install myModule
```

```
import myModule
```
result
```
imported.
```

3. restore python environment
```
%python.conda deactivate
```
check for recovered to the original Zeppelin environment.
```
%python
import sys
print(sys.version)
print(sys.path)
```
result
```
python version 2.7.1
%ZEPPELIN_HOME%/interpreter/python/lib:%PYTHON_HOME
```
```
import myModule
```
result
```
no module myModule
```

All three must be supported by conda.
1. Support to installing and using the python library
2. Support to change the python version
3. Support to configuring non-python environments

old zeppelin does works for this part(0.71).
but, the Python Interpreter has changed significantly since 0.71.
I guess missing the environment part for this.
I modified it.


python env set code
(https://github.com/apache/zeppelin/blob/master/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L151)

cloverhearts:
I think we are worried about `conda interpreter` and coupling for` conda env name`.
However, the conda env name is more important to the python interpreter than the conda interpreter.
Also, the python interpreter should not query the conda interpreter for its own state.

I think this misunderstanding is due to the member name `conda env name`.
If I have any additional problems, I will change this to `virtualEnvName`.

1ambda:
@cloverhearts I have few questions. 

What happens when we want to implement `PythonPyenvInterpreter` which is similar to `PythonCondaInterpreter`?

- In that case, we have to add new functions to PythonInterpreter like `setPyEnvLib()...`?

zjffdu:
@cloverhearts Just to clarify my concern. Here's what I see when I use conda outside of zeppelin.

Step 1. Create enviroment `env_1` using conda, and install pandas
Step 2. Set `zeppelin.python` to the binPath of `env_1` and restart python interpreter
Step 3. Now when I use python interpreter, I am using python of `env_1` and can use any libraries installed under `env_1` 

That means I just need to set `zeppelin.python` to switch to another conda env.  That's why I don't understand why you need to set additional Python library path for the new environment

cloverhearts:
@zjffdu 
This change will work exactly as your request.
There is only one python library path to be added.
1. If the user does not install python via conda.
(If the user installs only the library)
Zeppelin use to `zeppelin.python`.
(Used in the original environment.)

2. If you install python via conda.
(separate python from zeppelin.python)
if when the conda environment is active, add to library path and replace to installed python via conda from zeppelin.python.
Zeppelin to work in the new python environment.

3. The user disables the environment. (conda deactiavte)
It is restored to the original zeppelin.python environment.

cloverhearts:
@1ambda 
Currently there is a possibility that it will be structurally.
However, if you store it in PythonCondaInterpreter, the problem is greatly aggravated.

1. PythonInterpreter needs to monitor PythonCondaInterpreter at all times.
     (Python Interpreter should check Conda's environment activation when restarting.)

2. Other configurations other than PythonCondaInterpreter If an interpreter is added (as you might have guessed), the Python Interpreter is in a situation where you need to constantly query for all other Interpreter environments. This will affect the life cycle of the Python Interpreter.

cloverhearts:
I have a solution to remove conda related members from PythonInterpreter. 
also PythonCondaInterpreter will not have any members.
However, it differs in focus from this PR.
(The problem is that there are changes related to the start of the Python Interpreter.)

This PR is intended to be used when the user changes the environment by installing new python while using conda.
This also solves the Zeppelin Python Interpreter's infinite loop problem.

cloverhearts:
In my opinion, allowing for fluctuations in the python environment is itself a problem.
But zeppelin are supporting it.
If it does not, then the semantics of conda activate and conda deactivate disappear.

In fact, I also think that changing the python environment can be a problem.
However, the current Zeppelin supports it, and a bug exists.
This PR is a PR that solves the bug.

cloverhearts:
@zjffdu @1ambda @felixcheung 
Thank you for your kind reply despite my lack of English.
I hope this PR is merged.
So, i will try to reflect your opinions as much as possible.
Thank you. :)

1ambda:
@cloverhearts There is no problem about python lib paths. 

`%python.conda install` command is the problem.

So you don't need to write code for python path.

image

zjffdu:
@cloverhearts 
`There is only one python library path to be added.`
Do you mean the py4j ? IIUC, PythonInterpreter will add py4j to PYTHONPATH, it doesn't relate with `PythonCondaInterpreter`. That's why I mention we just need to set `zeppelin.python`, Let me know if I miss anything.

1ambda:
@cloverhearts There is no problem about python lib paths. 

`%python.conda install` and `%python conda list` commands are the problem. They don't aware activated env by Zeppelin.

So you don't need to write code for python path.

image

cloverhearts:
Thank you @1ambda @zjffdu @felixcheung 
I think my code with various purposes is distracting you.
And, unfortunately, my English ability is difficult to describe in a sophisticated way.
Therefore, I will share PR.
In this PR, we will only deal with the wrong library installation of conda.

Many thanks for your comments.

cloverhearts:
I have separated this PR.
I'm sorry for the confusion.
https://github.com/apache/zeppelin/pull/2578

PR #2575, opened 2017-09-08
[ZEPPELIN-2921] does not work conda environment in python interpreter

cloverhearts:
### What is this PR for?
It seems that the environment of the python interpreter has changed to the py4j environment, causing problems with the library path and the default environment.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2921

Please run the following command line for each paragraph.
If the module such as scipy is normally imported, it is a success.
`%python.conda create --name Hello6 python=2.7`
`%python.conda activate Hello6`
`%python.conda install seaborn pandas numpy scipy matplotlib`
```
%python
import scipy as sp
import seaborn as sns
```
`%python.conda deactivate`

#### Before
image

#### After
image

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
@@ -135,9 +135,13 @@ public class PythonCondaInterpreter extends Interpreter {
         }
       }
     }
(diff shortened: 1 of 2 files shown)


cloverhearts:
https://github.com/apache/zeppelin/pull/2574

same pr and different base branch

PR #2576, opened 2017-09-08
[hotfix] JDBC connection does not release when got exception.

astroshim:
### What is this PR for?
This PR fixes JDBC connection release problem.

for example whenever i run not executable command like following
image

new JDBC connection is made like following.
```
$ netstat -an |grep EST |grep 3306 |wc -l
       1
$ netstat -an |grep EST |grep 3306 |wc -l
       2
$ netstat -an |grep EST |grep 3306 |wc -l
       3
```

### What type of PR is it?
Bug Fix

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -747,11 +747,6 @@ public class JDBCInterpreter extends KerberosInterpreter {
     } catch (SQLException e) {
       logger.error("Cannot run " + sql, e);
       String errorMsg = Throwables.getStackTraceAsString(e);
-      try {
-        closeDBPool(user, propertyKey);
-      } catch (SQLException e1) {
-        logger.error("Cannot close DBPool for user, propertyKey: " + user + propertyKey, e1);
-      }
       interpreterResult.add(errorMsg);
       return new InterpreterResult(Code.ERROR, interpreterResult.message());
     } finally {
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -744,7 +744,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
           }
         }
       }
-    } catch (SQLException e) {
+    } catch (Throwable e) {
       logger.error("Cannot run " + sql, e);
       String errorMsg = Throwables.getStackTraceAsString(e);
       interpreterResult.add(errorMsg);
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Tested jdbc interpreter test manually and CI failure looks irrelevant.

LGTM

cloverhearts:
Tested :)
Thank you for good fix.

cloverhearts:
If there is no new comment, I will merge it.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -744,16 +760,6 @@ private InterpreterResult executeSql(String propertyKey, String sql,
         } catch (SQLException e) { /*ignored*/ }
       }
       getJDBCConfiguration(user).removeStatement(paragraphId);
-    } catch (Throwable e) {
```
felixcheung: the new finally block can throw too right?

astroshim: I thought `SQLException` is enough. 
otherwise let me fix it to `Throwable`.

cloverhearts: Hello, @felixcheung.
Could you check for @astroshim 's comment?

cloverhearts:
This is a hot-fix change.
Therefore, they merge quickly.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
felixcheung: I think there is a misunderstanding.
It was
```
  try {
        List<String> sqlArray;
...
         } finally {
               resultSet.close();
               statement.close();
 ....
         }
        }
             connection.commit();
           connection.close();
        }
        getJDBCConfiguration(user).removeStatement(paragraphId);
 -    } catch (Throwable e) {
,,,
      }
```

Now is
```
  try {
        List<String> sqlArray;
...
        } finally {
               resultSet.close();
               statement.close();
 ...
        }
 +    } catch (Throwable e) {
 +      logger.error("Cannot run " + sql, e);
 +      String errorMsg = Throwables.getStackTraceAsString(e);
 +      interpreterResult.add(errorMsg);
 +      return new InterpreterResult(Code.ERROR, interpreterResult.message());
 +    } finally {
             connection.commit();
           connection.close();
      }
```

so now there is no catch for 
```
             connection.commit();
           connection.close();

```

astroshim: @felixcheung you mean we need to return `InterpreterResult` as error when exception occurs from `connection.commit()` and `connection.close()` ?

cloverhearts: @felixcheung 
I think enough try-catch for `Throwable`.
This is due to the fact that the exception has been moved.
or if this is not your focus, then could you explain more to details?


Because this problem continues to cause new connections,
I hope this will be incorporated soon as possible and contribute to stabilization.

or you can handle your comments with new tickets and PR.

Thank you :)

astroshim: and there is `catch` statement for `connection.commit`.
image

felixcheung:
Ok I think it's fine then

cloverhearts:
@felixcheung 
Thank you :)

cloverhearts:
I will merge to master

PR #2577, opened 2017-09-09
ZEPPELIN-2898. Support Yarn-Cluster for Spark Interpreter

zjffdu:
### What is this PR for?
This is the first version for supporting yarn-cluster of `SparkInterpreter`.   I just delegate all the function to `spark-submit` as yarn-cluster is natively supported by spark, we don't need to reinvent the wheel. But there's still improvement to be done in future, e.g. I put some spark specific logic in `InterpreterSetting` which is not a good practise.  I plan to improve it when I refactor the `Interpreter` class (ZEPPELIN-2685). 

Besides that, I also add `MiniHadoopCluster` & `MiniZeppelin` which help for the integration test of yarn-client & yarn-cluster mode, otherwise I have to manually verify yarn-client & yarn-cluster mode which would easily cause regression issue in future.

To be noticed:

* SPARK_HOME must be specified for yarn-cluster mode
* HADOOP_CONF_DIR must be specified for yarn-cluster mode

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
https://github.com/zjffdu/zeppelin/tree/ZEPPELIN-2898

System test is added in `SparkInterpreterIT`. 

### Questions:

### DIFF
--- a/conf/log4j_yarn_cluster.properties
+++ b/conf/log4j_yarn_cluster.properties
@@ -1,0 +1,23 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+log4j.rootLogger = INFO, stdout
+
+log4j.appender.stdout = org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.layout = org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%5p [%d] ({%t} %F[%M]:%L) - %m%n
+
--- a/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
@@ -78,6 +78,7 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
   private long ipythonLaunchTimeout;
   private String additionalPythonPath;
   private String additionalPythonInitFile;
+  private boolean useBuiltinPy4j = true;
 
   private InterpreterOutputStream interpreterOutput = new InterpreterOutputStream(LOGGER);
 
@@ -92,6 +93,7 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
    * @param additionalPythonPath
    */
   public void setAdditionalPythonPath(String additionalPythonPath) {
+    LOGGER.info("setAdditionalPythonPath: " + additionalPythonPath);
     this.additionalPythonPath = additionalPythonPath;
   }
 
@@ -105,6 +107,10 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
     this.additionalPythonInitFile = additionalPythonInitFile;
   }
 
+  public void setAddBulitinPy4j(boolean add) {
+    this.useBuiltinPy4j = add;
+  }
+
   @Override
   public void open() {
     try {
@@ -113,6 +119,7 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
         return;
       }
       pythonExecutable = getProperty().getProperty("zeppelin.python", "python");
+      LOGGER.info("Python Exec: " + pythonExecutable);
       ipythonLaunchTimeout = Long.parseLong(
           getProperty().getProperty("zeppelin.ipython.launch.timeout", "30000"));
       this.zeppelinContext = new PythonZeppelinContext(
@@ -218,29 +225,34 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
     watchDog = new ExecuteWatchdog(ExecuteWatchdog.INFINITE_TIMEOUT);
     executor.setWatchdog(watchDog);
 
-    String py4jLibPath = null;
-    if (System.getenv("ZEPPELIN_HOME") != null) {
-      py4jLibPath = System.getenv("ZEPPELIN_HOME") + File.separator
-          + PythonInterpreter.ZEPPELIN_PY4JPATH;
-    } else {
-      Path workingPath = Paths.get("..").toAbsolutePath();
-      py4jLibPath = workingPath + File.separator + PythonInterpreter.ZEPPELIN_PY4JPATH;
+    if (useBuiltinPy4j) {
+      String py4jLibPath = null;
+      if (System.getenv("ZEPPELIN_HOME") != null) {
+        py4jLibPath = System.getenv("ZEPPELIN_HOME") + File.separator
+            + PythonInterpreter.ZEPPELIN_PY4JPATH;
+      } else {
+        Path workingPath = Paths.get("..").toAbsolutePath();
+        py4jLibPath = workingPath + File.separator + PythonInterpreter.ZEPPELIN_PY4JPATH;
+      }
+      if (additionalPythonPath != null) {
+        // put the py4j at the end, because additionalPythonPath may already contain py4j.
+        // e.g. PySparkInterpreter
+        additionalPythonPath = additionalPythonPath + ":" + py4jLibPath;
+      } else {
+        additionalPythonPath = py4jLibPath;
+      }
     }
-    if (additionalPythonPath != null) {
-      // put the py4j at the end, because additionalPythonPath may already contain py4j.
-      // e.g. PySparkInterpreter
-      additionalPythonPath = additionalPythonPath + ":" + py4jLibPath;
-    } else {
-      additionalPythonPath = py4jLibPath;
-    }
+
     Map<String, String> envs = EnvironmentUtils.getProcEnvironment();
     if (envs.containsKey("PYTHONPATH")) {
-      envs.put("PYTHONPATH", additionalPythonPath + ":" + envs.get("PYTHONPATH"));
+      if (additionalPythonPath != null) {
+        envs.put("PYTHONPATH", additionalPythonPath + ":" + envs.get("PYTHONPATH"));
+      }
     } else {
       envs.put("PYTHONPATH", additionalPythonPath);
     }
 
-    LOGGER.debug("PYTHONPATH: " + envs.get("PYTHONPATH"));
+    LOGGER.info("PYTHONPATH: " + envs.get("PYTHONPATH"));
     executor.execute(cmd, envs, this);
 
     // wait until IPython kernel is started or timeout
--- a/spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
+++ b/spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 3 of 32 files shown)


zjffdu:
@Leemoonsoo @jongyoul Could you help review ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +717,138 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
+          "interpreter" + File.separator + "spark" + File.separator + "pyspark");
+    } else {
+      pysparkPath = new File(pysparkBasePath,
+          "python" + File.separator + "lib");
+    }
+
+    //Only one of py4j-0.9-src.zip and py4j-0.8.2.1-src.zip should exist
+    //TODO(zjffdu), this is not maintainable when new version is added.
+    String[] pythonLibs = new String[]{"pyspark.zip", "py4j-0.9-src.zip", "py4j-0.8.2.1-src.zip",
+      "py4j-0.10.1-src.zip", "py4j-0.10.3-src.zip", "py4j-0.10.4-src.zip"};
```
felixcheung: we really need to fix this....

jongyoul: Yes, I wouldn't have thought it used for a long time...

zjffdu: Sure, this is on my plan for the next step.

zjffdu:
@Leemoonsoo @jongyoul As I mentioned in the PR description, this is not a perfect PR. If you don't have any more comments, I will merge it and continue the next PR to improve it.  Thanks

conf/log4j_yarn_cluster.properties
```diff
@@ -0,0 +1,23 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
felixcheung: why do we need to release with a new .properties file?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -218,29 +225,32 @@ private void launchIPythonKernel(int ipythonPort)
...
-      Path workingPath = Paths.get("..").toAbsolutePath();
-      py4jLibPath = workingPath + File.separator + PythonInterpreter.ZEPPELIN_PY4JPATH;
-    }
-    if (additionalPythonPath != null) {
-      // put the py4j at the end, because additionalPythonPath may already contain py4j.
-      // e.g. PySparkInterpreter
-      additionalPythonPath = additionalPythonPath + ":" + py4jLibPath;
-    } else {
-      additionalPythonPath = py4jLibPath;
+    if (addBuiltinPy4j) {
```
felixcheung: nit `addBuiltinPy4j` is a bit confusing? `useBuiltinPy4j`? `addBuiltinPy4jPath`?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -218,29 +225,32 @@ private void launchIPythonKernel(int ipythonPort)
...
+        // e.g. PySparkInterpreter
+        additionalPythonPath = additionalPythonPath + ":" + py4jLibPath;
+      } else {
+        additionalPythonPath = py4jLibPath;
+      }
     }
+
     Map<String, String> envs = EnvironmentUtils.getProcEnvironment();
-    if (envs.containsKey("PYTHONPATH")) {
+    if (envs.containsKey("PYTHONPATH") && additionalPythonPath != null) {
```
felixcheung: `||` instead? if PYTHONPATH is not set but additionalPythonPath is, don't we want to set it?

spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
```diff
@@ -0,0 +1,96 @@
...
+      File[] py4j = new File(sparkHome + "/python/lib").listFiles(new FilenameFilter() {
+        @Override
+        public boolean accept(File dir, String name) {
+          return name.startsWith("py4j");
+        }
+      });
+      if (py4j.length == 0) {
+        throw new RuntimeException("No py4j files found under " + sparkHome + "/python/lib");
+      } else if (py4j.length > 1) {
+        throw new RuntimeException("Multiple py4j files found under " + sparkHome + "/python/lib");
```
felixcheung: One wonder if this might be a bit too restrictive ... I for one have multiple py4j versions and Spark always picks the right one

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +719,101 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
+      sparkConfBuilder.append(" --conf " + name + "=" + sparkProperties.getProperty(name));
+    }
+
+    env.put("ZEPPELIN_SPARK_CONF", sparkConfBuilder.toString());
+    LOGGER.debug("getEnvFromInterpreterProperty: " + env);
     return env;
   }
 
+  private void setupPropertiesForPySpark(Properties sparkProperties, String sparkMaster) {
+    if (sparkMaster.startsWith("yarn")) {
```
felixcheung: use `isYarnMode()`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +719,101 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
-        env.put((String) key, property.getProperty((String) key));
+    Map<String, String> env = new HashMap<String, String>();
+    Properties sparkProperties = new Properties();
+    String sparkMaster = "local[*]";
+    for (String key : property.stringPropertyNames()) {
+      if (RemoteInterpreterUtils.isEnvString(key)) {
+        env.put(key, property.getProperty(key));
       }
+      if (key.equals("master")) {
+        sparkMaster = property.getProperty("master");
```
felixcheung: should this do the same as the other code (or perhaps re-use as a method) - fall back to get spark.master?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +719,101 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
+          if (sparkMaster == null) {
+            sparkMaster = property.getProperty(key);
+          }
+        }
+      }
+    }
+
+    setupPropertiesForPySpark(sparkProperties, sparkMaster);
+    setupPropertiesForSparkR(sparkProperties, sparkMaster, property.getProperty("SPARK_HOME"));
+    if (sparkMaster != null && sparkMaster.equals("yarn-cluster")) {
```
felixcheung: shouldn't this match `isYarnMode()`?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +719,101 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
+      env.put("SPARK_YARN_CLUSTER", "true");
     }
+
+    StringBuilder sparkConfBuilder = new StringBuilder();
+    if (sparkMaster != null) {
+      sparkConfBuilder.append(" --master " + sparkMaster);
+    }
+    if (sparkMaster.equals("yarn-cluster")) {
```
felixcheung: ditto

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -713,16 +719,101 @@ RemoteInterpreterProcess createInterpreterProcess() {
...
+          "interpreter" + File.separator + "spark" + File.separator + "R");
+    } else {
+      sparkRBasePath = new File(sparkHome, "R" + File.separator + "lib");
+    }
+
+    File sparkRPath = new File(sparkRBasePath, "sparkr.zip");
+    if (sparkRPath.exists() && sparkRPath.isFile()) {
+      mergeSparkProperty(sparkProperties, "spark.yarn.dist.archives", sparkRPath.getAbsolutePath());
+    } else {
+      LOGGER.warn("sparkr.zip is not found, sparkr may not work.");
```
felixcheung: `sparkr may not work` => `SparkR may not work`

conf/log4j_yarn_cluster.properties
zjffdu: Because the default log4j.properties use `DailyRollingFileAppender`, this is not suitable for yarn container.

spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
zjffdu: Because spark hard code each py4j version, but zeppelin needs to work for multiple versions of spark.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
zjffdu: No, this only works for yarn-cluster mode, not yarn-client mode. yarn-client mode still use the default logj4.properties, because driver runs in the zeppelin host

spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
felixcheung: sure, but it might break people with existing setup...

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
felixcheung: well, my point is we are not handling yarn cluster mode consistently.
yarn cluster mode can set by master=yarn, deployMode=cluster or
master=yarn-cluster (deprecated)
we have both in this PR.. either we should support both, or the supported not deprecated way, I think

spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
zjffdu: Why ? If there's multiple version of py4j under `SPARK_HOME/python/lib` or `ZEPPELIN_HOME/interpreter/spark/pyspark`, it must be something wrong with user's enviroment.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
zjffdu: Right, I was planning to fix it in the next PR. This is a messy. Let me fix it in this PR.

spark/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
felixcheung: maybe, maybe it is an edge case.
do we need to force the logic here? is it possible to call say, `sbin/spark-config.sh`?

zjffdu: I think `sbin/spark-config.sh` only affect pyspark shell. Spark also hard py4j version in scala code when it needs to detect the py4j and distribute to yarn containers. . See https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/PythonUtils.scala#L35

zjffdu:
@Leemoonsoo @jongyoul @felixcheung Any more comments ?

jongyoul:
I don't

zjffdu:
Thanks, I will merge it to continue the next follow up PR.

felixcheung:
hey @zjffdu how are we tracking the followup tasks, if there is any?
for example, it'd good to have documentation on how to run this yarn cluster mode in spark.md

zjffdu:
@felixcheung I will do the follow up in https://issues.apache.org/jira/browse/ZEPPELIN-2685

PR #2579, opened 2017-09-10
Zeppelin-2922 Allow removing last paragrap

malayhm:
### What is this PR for?
Before this change, it was not allowed to remove the last paragraph which would not make the notebook look clean.

After this change, it will be allowed to remove the last paragraph. 
- Whenever there will be only one paragraph in the notebook, remove paragraph option will be hidden so that user will not able to make the notebook empty. 

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2922

1. Build: mvn clean package -Denforcer.skip -DskipTests -Drat.skip
2. Create paragraphs in a notebook and try to remove it
3. You can delete any paragraphs(even last paragraphs), but when there will be only one paragraph, remove button will be hidden

After:

zep-2922 after

### Questions:
* Does the licenses files need an update? No
* Does this needs documentation? May be


1ambda:
@malayhm Hi, Thanks for the contribution.

I just read the JIRA issue and have a question. Could you elaborate the situation when 

> sometimes user want to delete the last empty paragraph to make the notebook interface clean.

Leemoonsoo:
This is correct behavior. Preventing delete 'Last paragraph' was intended to keep at least one paragraph in the notebook. but it somehow implemented in a way to prevent removal of bottom paragraph.

LGTM and merge this to master if no further discussions.

PR #2580, opened 2017-09-11
[ZEPPELIN-2924] Fix broken javascript source map

1ambda:
### What is this PR for?

Fixed broken javascript source map. The coverage loader should be used only for test env.

**that's really annoying because we can't set breakpoint and even can't see the original source code.**

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
[ZEPPELIN-2924](https://issues.apache.org/jira/browse/ZEPPELIN-2924)

1. run the webapp as dev mode: `localhost:9000`
2. open the developer console and click any log message
3. check whether source is minified or not.
4. execute `yarn run test` in the terminal to check that coverage feature works well.

#### Before

image

image

#### After

image

image

### Questions:

### DIFF
--- a/zeppelin-web/webpack.config.js
+++ b/zeppelin-web/webpack.config.js
@@ -120,6 +120,7 @@ module.exports = function makeWebpackConfig () {
    * Reference: http://webpack.github.io/docs/configuration.html#devtool
    * Type of sourcemap to use per build type
    */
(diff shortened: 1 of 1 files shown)


jongyoul:
LGTM

1ambda:
thanks for review. merge if no more discussion

PR #2582, opened 2017-09-12
fixed jdbc connection issue.

astroshim:
### What is this PR for?
This PR fixes JDBC connection release problem.

for example whenever i run not executable command like following
image

new JDBC connection is made like following.
```
$ netstat -an |grep EST |grep 3306 |wc -l
       1
$ netstat -an |grep EST |grep 3306 |wc -l
       2
$ netstat -an |grep EST |grep 3306 |wc -l
       3
```

### What type of PR is it?
Bug Fix

### Questions:


astroshim:
CI fails seems not related to this PR

minahlee:
Merging if there is no more discussion since this is same fix as #2576 for branch-0.7

PR #2587, opened 2017-09-13
[ZEPPELIN-2909]. Support shared SparkContext across language in livy interpreter

zjffdu:
### What is this PR for?
LIVY-194 implement the shared SparkContext across languages, this ticket is trying to integrate this feature.

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2909

Tested is added

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -57,6 +57,8 @@ import org.apache.zeppelin.interpreter.InterpreterException;
 import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.InterpreterResultMessage;
 import org.apache.zeppelin.interpreter.InterpreterUtils;
+import org.apache.zeppelin.interpreter.LazyOpenInterpreter;
+import org.apache.zeppelin.interpreter.WrappedInterpreter;
 import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -94,6 +96,9 @@ public abstract class BaseLivyInterpreter extends Interpreter {
   protected LivyVersion livyVersion;
   private RestTemplate restTemplate;
   private Map<String, String> customHeaders = new HashMap<>();
+
+  // delegate to sharedInterpreter when it is available
+  protected LivySharedInterpreter sharedInterpreter;
 
   Set<Object> paragraphsToCancel = Collections.newSetFromMap(
       new ConcurrentHashMap<Object, Boolean>());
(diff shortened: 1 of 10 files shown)


zjffdu:
@prabhjyotsingh Could you help review it ? Thanks

livy/src/main/resources/interpreter-setting.json
```diff
@@ -227,5 +227,21 @@
       "editOnDblClick": false,
       "completionKey": "TAB"
     }
+  },
+  {
+    "group": "livy",
+    "name": "shared",
+    "className": "org.apache.zeppelin.livy.LivySharedInterpreter",
```
prabhjyotsingh: Will this be automatically added to `interpreter.json`, or the user has to add it manually? And if this class is not present in `interpreter.json` (by default), will the previous interpreter works?

prabhjyotsingh:
This looks good, with one minor comment, this was one long pending feature missing in Livy a big +1 for adding this functionality.

livy/src/main/java/org/apache/zeppelin/livy/LivySharedInterpreter.java
```diff
@@ -0,0 +1,107 @@
...
+  @Override
+  public void open() throws InterpreterException {
+    try {
+      // check livy version
+      try {
+        this.livyVersion = getLivyVersion();
+        LOGGER.info("Use livy " + livyVersion);
+      } catch (APINotFoundException e) {
+        this.livyVersion = new LivyVersion("0.2.0");
+        LOGGER.info("Use livy 0.2.0");
```
felixcheung: we should clarify? "assume livy version 0.2.0"?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -181,32 +216,37 @@ protected void initLivySession() throws LivyException {
...
 
   public SessionInfo getSessionInfo() {
+    if (sharedInterpreter != null && sharedInterpreter.isSupported()) {
+      return sharedInterpreter.getSessionInfo();
+    }
     return sessionInfo;
   }
 
+  public String getCodeType() {
+    if (getSessionKind().equalsIgnoreCase("pyspark3")) {
```
felixcheung: doesn't it have both python2 and python3 variants?

livy/src/main/resources/interpreter-setting.json
zjffdu: It will be added to interpreter.json automatically.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
zjffdu: pyspark3 is removed in livy 0.4, pyspark will support both python2 and python3

livy/src/main/java/org/apache/zeppelin/livy/LivySharedInterpreter.java
zjffdu: fixed

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
felixcheung: ah ok

zjffdu:
@felixcheung User don't need to explicitly enable this feature, it is enable by default if user use livy 0.5. As this feature is already implemented in zeppelin's native spark interpreter, livy just try to make the behavior as consistent as zeppelin spark interpreter.

felixcheung:
still, I think this feature is distinctive enough that we should document and bring some visibility (even if it just works)

zjffdu:
@felixcheung  It is documented here. https://github.com/apache/zeppelin/blame/master/docs/interpreter/livy.md#L219

felixcheung:
ok thanks, didn't realize it's been 7 months... :)

PR #2272, opened 2017-04-21
ZEPPELIN-2397. Add toJson and fromJson method for serializable objects

zjffdu:
### What is this PR for?
Add toJson and fromJson method for all the objects that need ser/der in zeppelin. 

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2397?filter=-2

### How should this be tested?
Outline the steps to test the PR here.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #2591, opened 2017-09-17
[HOTFIX] Forward compatibility of Note class

jongyoul:
### What is this PR for?
Solving inconsistency issue of Note's dateformat between branch-0.7 and master. It's related to #2272 

### What type of PR is it?
[Hot Fix]

### Todos
* [x] - Support several dateformat for Note class

### What is the Jira issue?
* N/A

* Sync with some remote repository

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
@@ -19,6 +19,9 @@ package org.apache.zeppelin.notebook;
 
 
 import com.google.common.collect.Lists;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


jongyoul:
@Leemoonsoo This is hotfix. Please test it and leave comments.

zjffdu:
@jongyoul I hit this issue before, I think just adding `yyyy-MM-dd HH:mm:ss.SSS` into `NotebookImportDeserializer` should be able to resolve this issue.

jongyoul:
Yes, if we use that gson, it can be solved. But there're several gson instances and used different one. I fixed it in order to use the same one to handle that type.

jongyoul:
Some code used only vanila gson for handling json.

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
```diff
@@ -232,4 +226,11 @@ public void returnUnchangedResultsWithDifferentUser() throws Throwable {
 
 
   }
+
+  @Test
+  public void deserializeParagraphFromDifferentDateFormat() {
+    Paragraph paragraph = Note.GSON
+        .fromJson("{\"dateUpdated\":\"2017-09-01 00:01:02.345\"}", Paragraph.class);
+    System.out.println(paragraph.dateUpdated);
```
anthonycorbacho: Can we have a better unit test for that? i dont think `System.out.println` is enough.
can we use something like `assert`?

zjffdu:
@jongyoul Which code path use vanila gson for handling json ?

jongyoul:
@zjffdu I found S3,VFS,Zeppelinhub repo used different gson from Note. I know it's fixed in master but need to fix them in 0.7.x branch.

zjffdu:
Thanks @jongyoul

jongyoul:
Merge if there's no more discussion

PR #2592, opened 2017-09-19
ZEPPELIN-2685. Improvement on Interpreter class

zjffdu:
### What is this PR for?
Follow up of #2577. Main changes on Interpreter
* Add throw `InterpreterException` which is checked exception for the abstract methods of `Interpreter`, this would enforce the interpreter implementation to throw `InterpreterException`.
* field name refactoring.

     * `property` -> `properties`
     * `getProperty()` --> `getProperties()`
* Introduce launcher layer for interpreter launching. Currently we only use shell script to launch interpreter, but it could be any other service or component to launch interpreter, such as livy server , other 3rd party tools or even we may create a separate module for interpreter launcher 

     * abstract cass `InterpreterLauncher`
     * For now, only 2 implementation: `ShellScriptLauncher` & `SparkInterpreterLauncher`. We could add method in class `Interpreter` to allow interpreter to specify its own launcher class, but it could be future work.
  

### What type of PR is it?
[Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2685

Unit test is covered. `ShellScriptLauncherTest` & `SparkInterpreterLauncherTest`

### Questions:


zjffdu:
@Leemoonsoo @jongyoul @felixcheung Could you help review it ? Thanks. It is followup of #2577

jongyoul:
Launcher concept looks good to me.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -508,7 +509,11 @@ protected boolean jobAbort() {
     if (job != null) {
       job.setStatus(Status.ABORT);
     } else {
-      repl.cancel(getInterpreterContextWithoutRunner(null));
+      try {
+        repl.cancel(getInterpreterContextWithoutRunner(null));
+      } catch (InterpreterException e) {
+        throw new RuntimeException(e);
```
felixcheung: why rewrap the exception here?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -892,7 +892,11 @@ public void execute(JobExecutionContext context) throws JobExecutionException {
       if (releaseResource) {
         for (InterpreterSetting setting : notebook.getInterpreterSettingManager()
             .getInterpreterSettings(note.getId())) {
-          notebook.getInterpreterSettingManager().restart(setting.getId());
+          try {
+            notebook.getInterpreterSettingManager().restart(setting.getId());
+          } catch (InterpreterException e) {
+            logger.warn("Fail to resetart interpreter: " + setting.getId(), e);
```
felixcheung: `resetart` -> `restart`

felixcheung: I think this should be an error...

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  private String getDeployMode() {
+    String master = getSparkMaster(properties);
+    if (master.equals("yarn-client")) {
+      return "client";
+    } else if (master.equals("yarn-cluster")) {
+      return "cluster";
+    } else if (master.startsWith("local")) {
+      return "client";
```
felixcheung: why is "local" == "client"?

felixcheung: isn't this never going to hit? isYarnMode is true before calling this method

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -0,0 +1,186 @@
...
+   * 4. use local[*]
+   * @param properties
+   * @return
+   */
+  private String getSparkMaster(Properties properties) {
+    String master = properties.getProperty("master");
+    if (master == null) {
+      master = properties.getProperty("spark.master");
+      if (master == null) {
+        master = System.getenv("SPARK_HOME");
```
felixcheung: wait? master != SPARK_HOME?!?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -0,0 +1,186 @@
...
+      }
+      if (isSparkConf(key, properties.getProperty(key))) {
+        sparkProperties.setProperty(key, toShellFormat(properties.getProperty(key)));
+      }
+    }
+
+    setupPropertiesForPySpark(sparkProperties);
+    setupPropertiesForSparkR(sparkProperties, properties.getProperty("SPARK_HOME"));
+    if (isYarnMode() && getDeployMode().equals("cluster")) {
+      env.put("SPARK_YARN_CLUSTER", "true");
```
felixcheung: thinking more about this, I think we should avoid Zeppelin specific env with a `SPARK_` prefix, to avoid potential conflict

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/ShellScriptLauncher.java
```diff
@@ -0,0 +1,84 @@
...
+    LOGGER.info("Launching Interpreter: " + context.getInterpreterGroupName());
+    this.properties = context.getProperties();
+    InterpreterOption option = context.getOption();
+    InterpreterRunner runner = context.getRunner();
+    String groupName = context.getInterpreterGroupName();
+
+    int connectTimeout =
+        zConf.getInt(ZeppelinConfiguration.ConfVars.ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT);
+    if (option.isExistingProcess()) {
+      // TODO(zjffdu) remove the existing process approach seems no one is using this.
```
felixcheung: why is no one using existing process?

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -353,7 +353,7 @@ private DepInterpreter getDepInterpreter() {
   public boolean isYarnMode() {
     String master = getProperty("master");
     if (master == null) {
-      master = getProperty().getProperty("spark.master", "local[*]");
+      master = getProperty("spark.master", "local[*]");
     }
     return master.startsWith("yarn");
```
felixcheung: will this be refactored to share code with SparkInterpreterLauncher.java?

zjffdu:
It would not break the existing interpreter. I introduce `Launcher` concept to allow customize interpreter to launch. But for now all the interpreter still use `interpreter.sh` to launch it even for spark.  But I have a plan to create a specific script to launch spark, because there's lots of specific logic for spark interpreter.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
zjffdu: Because InterpreterException is checked exception. If I throw it, I have to change signature of the caller method, this would cause lots of code change. In order to keep the PR small, I just throw it as RuntimeException.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
zjffdu: my mistake. will fix it

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
zjffdu: there's only 2 valid value for `spark.submit.deployMode` (client or cluster). It is `client` for local mode IIUC.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/ShellScriptLauncher.java
zjffdu: Remve this todo, as some one is using the existing process for restarting scenario [ZEPPELIN-2879]

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
zjffdu: They are in different module (spark, zeppelin-zengine). so I am afraid there's no chance to reuse code.

zjffdu:
@Leemoonsoo Do you have any concern about the compatibility break in this PR ?

* Add throw InterpreterException which is checked exception for the abstract methods of Interpreter, this would enforce the interpreter implementation to throw InterpreterException.

* field name refactoring.
** property -> properties
** getProperty() --> getProperties()

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -626,6 +640,7 @@ public void clearNoteIdAndParaMap() {
     }
     return interpreters;
   }
+<<<<<<< 0c64d9ca676e48a749db9879fa3cebc06eb78b54
```
Leemoonsoo: I think this is unexpected line

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -266,6 +272,14 @@ public InterpreterSetting(InterpreterSetting o) {
     this.conf = o.getConf();
   }
 
+  private void createLauncher() {
+    if (group.equals("spark")) {
+      this.launcher = new SparkInterpreterLauncher(this.conf);
```
Leemoonsoo: @zjffdu Do you think it's possible to make this part more generic? For example, read class name of launcher from interpreter-setting.json and instantiate it using reflection. So any interpreter can define their own launcher without changing zeppelin core modules?

Leemoonsoo:
@zjffdu I don't have much concern about compatibility break you have mentioned. Left few comments, please check.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
zjffdu: Yeah，I think about it before. But it may involve more work. Because I don't want to make interpreter jar mix with zeppelin server jar. Although I can use URLClassLoader, still feel a little risky to do it right now because Interpreter may not be instantiated correctly in zeppelin server process if it has some other dependencies, it requires interpreter do initialization in method open instead of its constructor.

zjffdu:

PR #2596, opened 2017-09-24
[ZEPPELIN-2949] Allow custom Spark UI URL

necosta:
### What is this PR for?
Allow an override of default Spark logic in building the URL path and redirect the Zeppelin user directly to a configured reverse proxy path

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2949

* Set "zeppelin.spark.uiWebUrl" to http://localhost/spark-reverse-proxy-path
* Run Spark code
* Open Spark UI
* Expected to be redirected to http://localhost/spark-reverse-proxy-path

### Questions:

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -148,7 +148,7 @@ You can also set other Spark properties which are not listed in the table. For a
   <tr>
   <td>zeppelin.spark.uiWebUrl</td>
     <td></td>
-    <td>Override Spark UI default URL</td>
+    <td>Overrides Spark UI default URL (ex: http://{hostName}/{uniquePath)</td>
   </tr>
 </table>
 
@@ -188,7 +188,7 @@ For example,
  * **yarn-client** in Yarn client mode
  * **mesos://host:5050** in Mesos cluster
 
-That's it. Zeppelin will work with any version of Spark and any deployment type without rebuilding Zeppelin in this way. 
+That's it. Zeppelin will work with any version of Spark and any deployment type without rebuilding Zeppelin in this way.
 For the further information about Spark & Zeppelin version compatibility, please refer to "Available Interpreters" section in [Zeppelin download page](https://zeppelin.apache.org/download.html).
 
 > Note that without exporting `SPARK_HOME`, it's running in local mode with included version of Spark. The included version may vary depending on the build profile.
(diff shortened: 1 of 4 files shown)


necosta:
Thanks @felixcheung. Test looks unrelated but I'll double check.
Added one commit for documentation

Not sure how much we want to document on how to work with Spark behind reverse proxy but I can add that by setting properties:
* spark.ui.enabled=true
* spark.ui.reverseProxy=false (disables '/proxy/applicationId' web url path)
* spark.ui.proxyBase=/reverseProxyUniqueKey
* zeppelin.spark.uiWebUrl=http://{hostName}/reverseProxyUniqueKey

https://issues.apache.org/jira/browse/SPARK-20044 is preventing us from using a path prefix and it is assumed we are routing from spark port to reverseProxyUniqueKey

necosta:
Not sure what we can do to have the tests passing:
https://travis-ci.org/nokia/zeppelin/jobs/279484561
https://travis-ci.org/nokia/zeppelin/jobs/279484562
Any suggestions?

docs/interpreter/spark.md
```diff
@@ -145,6 +145,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>true</td>
     <td>Do not change - developer only setting, not for production use</td>
   </tr>
+  <tr>
+  <td>zeppelin.spark.uiWebUrl</td>
+    <td></td>
+    <td>Override Spark UI default URL</td>
```
felixcheung: could you add a bit more on what are the condition to set this, example value etc..?

necosta:
Thanks @felixcheung and apologies for the delay on this.
Rebased on top of latest master and updated documentation.
Can you have another look?
Thanks

docs/interpreter/spark.md
```diff
@@ -145,6 +145,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>true</td>
     <td>Do not change - developer only setting, not for production use</td>
   </tr>
+  <tr>
+  <td>zeppelin.spark.uiWebUrl</td>
+    <td></td>
+    <td>Overrides Spark UI default URL (ex: http://{hostName}/{uniquePath)</td>
```
felixcheung: should this be `{uniquePath)` -> `{uniquePath})`

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -936,6 +936,11 @@ public String getSparkUIUrl() {
       return sparkUrl;
     }
 
+    String sparkUrlProp = property.getProperty("zeppelin.spark.uiWebUrl", "");
+    if (!sparkUrlProp.isEmpty()) {
```
felixcheung: let's check for isBlank https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-

docs/interpreter/spark.md
felixcheung: I think we should clarify that the value should be a *full* URL

necosta: Fixed on latest commit

spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
necosta: I agree. Fixed on latest commit

felixcheung:
merging if no more comment

PR #2597, opened 2017-09-25
ZEPPELIN-2942. Remove non-remote interpreter approach

zjffdu:
### What is this PR for?

This PR is just trying to remove non-remote interpreter approach.  

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2942

Current test is passed

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
@@ -28,7 +28,8 @@ public class InterpreterOption {
   public static final transient String SCOPED = "scoped";
   public static final transient String ISOLATED = "isolated";
 
-  boolean remote;
+  // always set it as true, keep this field just for backward compatibility
+  boolean remote = true;
   String host = null;
   int port = -1;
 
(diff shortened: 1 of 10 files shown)


zjffdu:
@Leemoonsoo Could you help review ? Thanks

zjffdu:
ping @Leemoonsoo

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
```diff
@@ -28,7 +28,8 @@
   public static final transient String SCOPED = "scoped";
   public static final transient String ISOLATED = "isolated";
 
-  boolean remote;
+  // always set it as true, keep this field just for backward compatibility
+  boolean remote = true;
```
felixcheung: why leave this variable? it won't be helpful for backward compatibility if it is removed in the constructor, and setRemote, isRemote anyway?

zjffdu: Because this correspond to one field in `interpreter-setting.json`. And I don't want to break compatibility of `interpreter-setting.json` for now. I have a plan to define the spec of `interpreter-setting.json` & `interpreter.json`, at that time I may break the compatibility of them. But for now, I just want to keep them backward compatible.

felixcheung: ok

zjffdu:
Will merge if no more discussion.

Leemoonsoo:
LGTM

PR #2599, opened 2017-09-25
[ZEPPELIN-2952] encrypt credentials.json with AES

herval:
### What is this PR for?
Support encrypting passwords using a private key

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2952

- Set the env variable `ZEPPELIN_CREDENTIALS_ENCRYPT_KEY=something`
- Save a few credentials
- Check that the `credentials.json` file is storing encrypted passwords
- Restart server using the same env variable for `ZEPPELIN_CREDENTIALS_ENCRYPT_KEY`
- The credentials should still be decryptable

### Questions:
No

No

Yes

### DIFF
--- a/zeppelin-interpreter/pom.xml
+++ b/zeppelin-interpreter/pom.xml
@@ -133,6 +133,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk15on</artifactId>
+      <version>1.52</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-aether-provider</artifactId>
       <version>${maven.aeither.provider.version}</version>
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -443,6 +443,10 @@ public class ZeppelinConfiguration extends XMLConfiguration {
     return getBoolean(ConfVars.ZEPPELIN_CREDENTIALS_PERSIST);
   }
 
+  public String getCredentialsEncryptKey() {
+    return getString(ConfVars.ZEPPELIN_CREDENTIALS_ENCRYPT_KEY);
+  }
+
   public String getCredentialsPath() {
     return getRelativeDir(String.format("%s/credentials.json", getConfDir()));
   }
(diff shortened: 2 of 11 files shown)


Leemoonsoo:
Thanks @herval for the contribution!!

I tried this branch and have few feedbacks.

#### Document
I could able to create encrypt key with following command.
```
openssl enc -aes-128-cbc -k secret -P -md sha1
```
I'm not sure it's preferred way or there's any better way to create key.
Anyway, It'll be easier to use if we can document this new environment variable with how to generate key in `/docs/setup/operation/configuration.md`.

#### Protect AES key

In my understanding, if any user can get both AES key (in `conf/zeppelin-env.sh`) and `conf/credentials.json`, user can decrypt the file.

What do you think? will there be a way to prevent user get both key and encrypted file?

zeppelin-interpreter/pom.xml
```diff
@@ -133,6 +133,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk15on</artifactId>
+      <version>1.52</version>
```
felixcheung: we would need this in the LICENSE file...

zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -680,6 +684,7 @@ public String getStrictTransport() {
     ZEPPELIN_ALLOWED_ORIGINS("zeppelin.server.allowed.origins", "*"),
     ZEPPELIN_ANONYMOUS_ALLOWED("zeppelin.anonymous.allowed", true),
     ZEPPELIN_CREDENTIALS_PERSIST("zeppelin.credentials.persist", true),
+    ZEPPELIN_CREDENTIALS_ENCRYPT_KEY("zeppelin.credentials.encryptKey", null),
```
felixcheung: does this mean we need to save the entire key in config?

herval: you need the private key somewhere accessible by the process, yes. Since these configs also load from the OS env, you can (should?) keep the key there, instead of the .sh file. That's how 12-factor environments deal with these kinds of keys, at least.

Addressing @Leemoonsoo question above - if someone gets hold of your system config, there's basically nothing you can do to prevent them from accessing private keys or any other environment config you might have on your server. You can get better security by using a third-party PCI-compliant "service" to store keys, for instance, but that kind of solution is highly context-dependent and hard to implement generically

zeppelin-interpreter/pom.xml
herval: done

zeppelin-zengine/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
Leemoonsoo: Thanks for the explanation.

Leemoonsoo:
One CI failure https://travis-ci.org/herval/zeppelin/jobs/283036367 is license check and looks like `zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/Encryptor.java` misses apache license header in the beginning of the file.

@herval do you mind add license header to this file?

Other CI failure looks unrelated. so looks good to me except for this license header.

herval:
done :)

Leemoonsoo:
Thanks! LGTM and merge to master if no further comments.

PR #2600, opened 2017-09-26
ZEPPELIN-2953 Allow custom http header for livy interpreter

zjffdu:
### What is this PR for?
This PR is trying to allow user to add custom http headers when calling livy rest api. User just need to specify `zeppelin.livy.http.headers` in livy interpreter setting

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2953

### Questions:

### DIFF
--- a/docs/interpreter/livy.md
+++ b/docs/interpreter/livy.md
@@ -144,7 +144,12 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>zeppelin.livy.ssl.trustStorePassword</td>
     <td></td>
     <td>password for trustStore file. Used when livy ssl is enabled</td>
-  </tr>  
+  </tr>
+  <tr>
+    <td>zeppelin.livy.http.headers</td>
+    <td>key_1: value_1; key_2: value_2</td>
+    <td>custom http headers when calling livy rest api. Each http header is separated by `;`, and each header is one key value pair where key value is separated by `:`</td>
+  </tr>
 </table>
 
 **We remove livy.spark.master in zeppelin-0.7. Because we sugguest user to use livy 0.3 in zeppelin-0.7. And livy 0.3 don't allow to specify livy.spark.master, it enfornce yarn-cluster mode.**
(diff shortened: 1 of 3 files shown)


docs/interpreter/livy.md
```diff
@@ -144,7 +144,12 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>zeppelin.livy.ssl.trustStorePassword</td>
     <td></td>
     <td>password for trustStore file. Used when livy ssl is enabled</td>
-  </tr>  
+  </tr>
+  <tr>
+    <td>zeppelin.livy.http.headers</td>
+    <td>key_1: value_1; key_2: value_2</td>
```
bikassaha: Are ; and : valid separators that would not be present in any header itself?

zjffdu: See http://greenbytes.de/tech/webdav/rfc7230.html#rfc.iref.d.2

zjffdu:
Custom http headers is needed when integrating knox with livy interpreter.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -95,6 +96,13 @@ public BaseLivyInterpreter(Properties property) {
     this.pullStatusInterval = Integer.parseInt(
         property.getProperty("zeppelin.livy.pull_status.interval.millis", 1000 + ""));
     this.restTemplate = createRestTemplate();
+    if (!StringUtils.isBlank(property.getProperty("zeppelin.livy.http.headers"))) {
+      String[] headers = property.getProperty("zeppelin.livy.http.headers").split(";");
+      for (String header : headers) {
+        String[] splits = header.split(":");
+        customHeaders.put(splits[0].trim(), splits[1].trim());
```
felixcheung: handle the case when `splits[1]` is empty?

felixcheung: per rfc, looks like `:` is required but field value is not.
we should probably error out if `:` is missing but handle empty value

zjffdu: Fixed

PR #2601, opened 2017-09-27
[ZEPPELIN-2956] [font-end] Downloaded CSV/TSV data will get unexpected division when the column value contains both delimiter and quotation mark.

BruceXu1991:
### What is this PR for?
Fix for ZEPPELIN-2956. This PR is to deal with the case when column value contains both delimiter and quotation mark. So the downloaded data can be well recogized and seperated by Excel.

### What type of PR is it?
Bug Fix

### Todos
None

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2956

Manual test.
After applying this PR, you can test the patch as following steps:
1、suppose one of the column value contain both delimiter and quotation mark:  
=>  hello  beijing  i  lo"v,"e  you

2、download the data as csv/tsv file
take download as csv format file =>  hello,beijing,"lo""v,""e",you

3、open the data file through Excel to verify the bug fix.

before patch:
csvdata

after patch:
afterpr

### Questions:
no
no
no

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -769,6 +769,9 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
       for (let index in row) {
         let stringValue = (row[index]).toString()
         if (stringValue.indexOf(delimiter) > -1) {
+          if (stringValue.indexOf("\"")) {
+            stringValue = stringValue.replace("\"", "\"\"")
+          }
           dsvRow += '"' + stringValue + '"' + delimiter
         } else {
           dsvRow += row[index] + delimiter
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -769,8 +769,8 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
       for (let index in row) {
         let stringValue = (row[index]).toString()
         if (stringValue.indexOf(delimiter) > -1) {
-          if (stringValue.indexOf("\"")) {
-            stringValue = stringValue.replace("\"", "\"\"")
+          if (stringValue.indexOf('\"')) {
+            stringValue = stringValue.replace('\"', '\"\"')
           }
           dsvRow += '"' + stringValue + '"' + delimiter
         } else {
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -769,8 +769,8 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
       for (let index in row) {
         let stringValue = (row[index]).toString()
         if (stringValue.indexOf(delimiter) > -1) {
-          if (stringValue.indexOf('\"')) {
-            stringValue = stringValue.replace('\"', '\"\"')
+          if (stringValue.indexOf('"') > -1) {
+            stringValue = stringValue.replace(/"/g, '""')
           }
           dsvRow += '"' + stringValue + '"' + delimiter
         } else {
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -62,6 +62,14 @@ sessionManager = org.apache.shiro.web.session.mgt.DefaultWebSessionManager
 #cacheManager = org.apache.shiro.cache.MemoryConstrainedCacheManager
 #securityManager.cacheManager = $cacheManager
 
+### Enables 'HttpOnly' flag in Zeppelin cookies
+cookie = org.apache.shiro.web.servlet.SimpleCookie
+cookie.name = JSESSIONID
+cookie.httpOnly = true
+### Uncomment the below line only when Zeppelin is running over HTTPS
+#cookie.secure = true
+sessionManager.sessionIdCookie = $cookie
+
 securityManager.sessionManager = $sessionManager
 # 86,400,000 milliseconds = 24 hour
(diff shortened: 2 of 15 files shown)


zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
```diff
@@ -769,6 +769,9 @@ function ResultCtrl ($scope, $rootScope, $route, $window, $routeParams, $locatio
       for (let index in row) {
         let stringValue = (row[index]).toString()
         if (stringValue.indexOf(delimiter) > -1) {
+          if (stringValue.indexOf('\"')) {
+            stringValue = stringValue.replace('\"', '\"\"')
```
felixcheung: I think the way to escape `"` is `\"`?
actually according to wikipedia both approaches might work, depending on the preference of the app

BruceXu1991: From above PR test description, the comma separator is inside the quote, this csv case is ok.

When the string value is **fo"o \t b"ar**, then it goes to tsv case. 
The "\t" in **fo"o \t b"ar** can be correctly read by Excel. But if the string value is **fo"o	b"ar**, in which contains "\t" typed through **keyboard**, this string value will be misunderstand by Excel. So we still need to deal with tsv case.

felixcheung:
could you setup CI tests
https://zeppelin.apache.org/contribution/contributions.html#continuous-integration

BruceXu1991:
@felixcheung  I have setup the CI tests, but Jenkins server seems to have no space on device.

FATAL: Unable to produce a script file
java.io.IOException: No space left on device
	at java.io.FileOutputStream.writeBytes(Native Method)

Caused: java.io.IOException: Failed to create a temp file on /home/jenkins/jenkins-slave/workspace/zeppelin-pull-request
	at hudson.FilePath.createTextTempFile(FilePath.java:1386)
	at hudson.tasks.CommandInterpreter.createScriptFile(CommandInterpreter.java:162)
	at hudson.tasks.CommandInterpreter.perform(CommandInterpreter.java:94)

felixcheung:
hmm, that's odd, I don't think I've seen that before.... perhaps.. try again?

BruceXu1991:
I tryed again, but CI test failed job 6.4, and I could not figure out what the error means.
https://travis-ci.org/BruceXu1991/zeppelin/jobs/281177302

Leemoonsoo:
@BruceXu1991 Right, https://github.com/apache/zeppelin/pull/2603 fix CI job 6.4 error.
You can try merge master and see if CI becomes green!

BruceXu1991:
After merging to master, CI tests still have some exception, but the exception seems have no relation with this patch.

> job 7.2 https://travis-ci.org/BruceXu1991/zeppelin/jobs/281329390
> job 7.3 https://travis-ci.org/BruceXu1991/zeppelin/jobs/281329391


```
Failed: No element found using locator: By(xpath, //div[@class="new-paragraph" and @ng-click="insertNew('below');"])

15:11:48,992 ERROR org.apache.spark.scheduler.LiveListenerBus:91 - Listener  threw an exception
java.lang.NullPointerException
	at org.apache.zeppelin.spark.Utils.getNoteId(Utils.java:156)
	at org.apache.zeppelin.spark.SparkInterpreter$1.onJobStart(SparkInterpreter.java:189)
	at org.apache.spark.scheduler.SparkListenerBus$class.doPostEvent(SparkListenerBus.scala:37)
	at org.apache.spark.scheduler.LiveListenerBus.doPostEvent(LiveListenerBus.scala:36)
```

1ambda:
Hi, https://github.com/apache/zeppelin/pull/2323 added a new table library and it supports downloads. (for 0.8.0-SNAPSHOT)

image

_ http://ui-grid.info/docs/#/tutorial/206_exporting_data

Could we remove the custom csv / tsv generation for 0.8.0-SNAPSHOT and above versions? 

What do you think of?

BruceXu1991:
@1ambda  If the table library in #2323 could deal with the problem of downloading column value both contain delimiter and quotation mark, It may be better to use library.

Tagar:
I wonder if it's related to https://issues.apache.org/jira/browse/ZEPPELIN-1803 ?

BruceXu1991:
@Tagar  Yes, ZEPPELIN-2956 and ZEPPELIN-1803 point to the same issue. RFC-4180 explains the issue.

BruceXu1991:
ok, I will do it.

BruceXu1991:
@felixcheung The API(http://ui-grid.info/docs/#/tutorial/206_exporting_data) only support export data as csv format but no tsv format supporting. So if using this API, tsv format related PR should be submited to the API lib first. Maybe we could change to [#2601 (comment)](https://github.com/apache/zeppelin/pull/2601#issuecomment-333452621) after the API support tsv format.

felixcheung:
@BruceXu1991 I think that's fair. @1ambda what do you think?

BruceXu1991:
@1ambda   hi, is there any conclusion?

mebelousov:
https://github.com/apache/zeppelin/pull/2601
Hi, @BruceXu1991!
As @1ambda, I believe one could remove old button "Download Data as CSV".

The new exporter works fine and conform to RFC-4180 in most cases.
Excel 2013 processes well your case with new CSV exporter.
```
%python
import pandas as pd
data = [['hello', 'moscow', 'i', 'lo"v, "e', 'you']]
df = pd.DataFrame(data)
print('%table ' + df.to_csv(sep='\t', index=False))
```
This code gives the next result:

BruceXu1991:
@mebelousov   OK, I will close this PR.

Tagar:
> I believe one could remove old button "Download Data as CSV".

filed https://issues.apache.org/jira/browse/ZEPPELIN-3511 to remove old export csv/tsv 

This idea was also briefly discussed on [PR-2971](https://github.com/apache/zeppelin/pull/2971#issuecomment-391219166)

PR #2605, opened 2017-10-03
[ZEPPELIN-2963] Fix paragraph aborting on next run after cancel

namanmishra91:
### What is this PR for?
As mentioned in ZEPPELIN-2963, if a paragraph is cancelled while it was in scheduler's waiting queue, the next run of the paragraph always ends up in abort status. This change fixes this issue.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2963

The scenario that we want to reproduce is that a paragraph in the notebook should be cancelled when it is still in the remote scheduler's waiting queue. This can be reproduced in various ways. Following is the approach that I used:
1. Start zeppelin server.
2. Create a notebook with around 10-15 paragraphs with a few paragraphs being long running (at least a few seconds)
3. Do a "Run All" on the notebook
4. Cancel the last paragraph (non-empty) by clicking the "Pause" button. It will move to abort status.
5. After all paragraphs finish running, run the cancelled paragraph again. It should start running.

Before

paragraph_abort

After

paragraph_abort_fixed

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -493,6 +493,7 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
 
   @Override
   protected boolean jobAbort() {
+    boolean aborted = true;
     Interpreter repl = getRepl(getRequiredReplName());
     if (repl == null) {
       // when interpreters are already destroyed
@@ -507,10 +508,11 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
     Job job = scheduler.removeFromWaitingQueue(getId());
     if (job != null) {
       job.setStatus(Status.ABORT);
+      aborted = false;
     } else {
       repl.cancel(getInterpreterContextWithoutRunner(null));
     }
-    return true;
+    return aborted;
   }
 
   private InterpreterContext getInterpreterContext() {


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -507,10 +508,11 @@ protected boolean jobAbort() {
     Job job = scheduler.removeFromWaitingQueue(getId());
     if (job != null) {
       job.setStatus(Status.ABORT);
+      aborted = false;
```
felixcheung: this seems a bit counter intuitive - if the job is removed from waiting queue and its status is set to abort, isn't it aborted? perhaps add a comment here to explain why it is needed?

namanmishra91: I agree it seems counter intuitive. Basically, the issue was that in `JobRunner` we check if a job is aborted, then just toggle the flag, set status and return. Below is the code:
```
public void run() {
      if (job.isAborted()) {
        job.setStatus(Status.ABORT);
        job.aborted = false;

        synchronized (queue) {
          running.remove(job);
          queue.notify();
        }
        jobSubmittedRemotely = true;
        return;
      }
```

The best way would have been to be able to remove the above check but I am not sure what side-effects will it have (it could lead to race conditions). The next best way I could think of to avoid this problem was to disable `aborted` so that next run of the paragraph passes.

namanmishra91:
Build failures don't seem to be related to changes in this PR (selenium webdriver tests and flaky interpreter tests). Not sure what needs to be done to make them succeed.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
namanmishra91: @felixcheung Does the comment above help?

guptarajat:
This PR seems to be fixing functionality of Run All. 
I think if PR looks good from correctness pov then maybe we should merge this and file a follow up jira to address consistency issue, with more details about what exactly is needed.

zjffdu:
I think the root cause is that there're 2 fields in Job.java to represent job status. (`aborted` & `status`). Why not just removing `aborted`

namanmishra91:
Agreed that should be the ideal solution but considering how tightly it is integrated with the core execution workflow, this will be a risky change and needs to be done with utmost care to avoid any side-effects.

PR #2614, opened 2017-10-10
Add support for minio storage over s3 api

dmvk:
Adds support for github.com/minio/minio block storage


dmvk:
I'll try to implement new storage using official minio sdk https://github.com/minio/minio-java. It seems this patch is still not enough for s3 client to work properly with minio.

felixcheung:
also probably would be a good idea to open a JIRA

PR #2615, opened 2017-10-10
[ZEPPELIN-1999][IMPROVEMENT] refactoring replacement context parameters into interpreter settings

tinkoff-dwh:
### What is this PR for?
Field `protected Properties property` is protected. sometimes this field used derictly. Sometimes  interpreter is called from another insterpreter (for ex. Spark, SparkInterpreter creates spark context and calls PySpark). This improvement fix it problems with context parameters replacement into properties.

### What type of PR is it?
Improvement

1) Setting Spark interpreter
   set property 'spark.app.name=#{noteId}'
2) Run paragraph
  ```%pyspark 
      print(sc._conf.get("spark.app.name"))
```
3) Check result (note id)
4) Open http://localhost:4040/environment/ (spark ui), check property `spark.app.name` = note id

### Questions:


tinkoff-dwh:
Ready to review

zjffdu:
@tinkoff-dwh Could you explain more what kind of issue you try to fix in this PR ? I don't get it from the PR description.

tinkoff-dwh:
@zjffdu 
refactoring replacement values from the context in the interpreter settings. the previous solution didn't cover all cases:
1) Field protected Properties properties is protected. and interpreters has direct access (without replace vlalues from context). now the value in this field is already changed (and added field source settings)
2) changed the level at which occurred the replacement of the parameters, becauce replacement not works if interpreter call another interpreter. (For ex. %pyspark  -> SparkInterpreter call PySparkInterpreter)
3) refactoring JDBCInterpreter to work with replaces settings

tinkoff-dwh:
Ready to review

tinkoff-dwh:
Ready to review

zjffdu:
@tinkoff-dwh Does the replaceContext parameter happen for each interpret call ?

mebelousov:
@zjffdu replaceContext happens every call of jobRun() in RemoteInterpreterServer.java

PR #2618, opened 2017-10-11
[WIP] ZEPPELIN-2986. Add hadoop-aws and hadoop-azure to zeppelin server

zjffdu:
### What is this PR for?
Trivial fix. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2986

### Questions:

### DIFF
--- a/zeppelin-zengine/pom.xml
+++ b/zeppelin-zengine/pom.xml
@@ -374,6 +374,18 @@
 
     <dependency>
       <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-azure</artifactId>
+      <version>${hadoop.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-aws</artifactId>
+      <version>${hadoop.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
       <version>${hadoop.version}</version>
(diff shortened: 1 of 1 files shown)


zjffdu:
notebook storage is in module zeppelin-zengine. Just found it has jar conflicts with the native azure and s3 storage. Will keep it pending for now.

steveloughran:
# Spark 2.3 has a spark-hadoop-cloud module which adds a set of dependencies on the hadoop cloud modules, throwing out all the transitive conflicting modules. Either pick that up or lift all the exclusion logic there to keep out things like: jackson, guava, joda-time which will only cause pain. Hadoop 2.9+ will move to a shaded aws JAR here.

Do be aware, hadoop-aws is very fussy about aws-SDK versions: that library is a moving target. You can't expect to change versions and not expect to have the code link, let alone run.

felixcheung:
I agree this is going to be a huge moving target to include in the release of the project
Can this be just a list of steps people can follow? Dockerfile or actual Docker images people can use, instead?

steveloughran:
bq. Can this be just a list of steps people can follow?

It's a very brittle list of steps. It's easier at build time as you already have your hadoop version fixed, the hadoop-aws and hadoop-azure poms give you the library versions they need. All you need is to add them *and evict all conflict with the later stuff Spark has chosen*. It's really hard to get this right.

The spark work adds a new optional module and profile to set this up. I'd recommend doing the same thing for now, using the code in spark's POMs to tell you what to exclude

PR #2621, opened 2017-10-11
[Minor] Remove hardcoded key in zeppelinhub.

yanboliang:
### What is this PR for?
Hardcoded seed/key may cause security issue, so remove potential risk.

### What type of PR is it?
[Improvement]

### Todos
No.

### What is the Jira issue?
No jira created.


yanboliang:
@prabhjyotsingh Please help review.

khalidhuseynov:
LGTM

felixcheung:
@Leemoonsoo

krishna-pandey:
We need to consider below while removing the key. Are we relying on any such behaviour by specifying the seed value?
"If two instances of Random are created with the same seed, and the same sequence of method calls is made for each, they will generate and return identical sequences of numbers."

Refer: http://docs.oracle.com/javase/8/docs/api/java/util/Random.html

krishna-pandey:
LGTM.
It seems we are using SecureRandom which does not require seed explicitly to initialise and hence hardcoded key can be removed, unless we intentionally wanted a deterministic output.

prabhjyotsingh:

PR #2624, opened 2017-10-15
[ZEPPELIN-2965] Add code completion for livy interpreter

pellmont:
### What is this PR for?
This PR adds code autocompletion feature to LivyInterpreter.
Livy version 0.5 will have an auto completion API.

### What type of PR is it?
Feature

### Todos
* [ ] - Task

### What is the Jira issue?
[ZEPPELIN-2965] https://issues.apache.org/jira/browse/ZEPPELIN-2965

Pulled out server calls to a separate class to support proper unit-testing with mockito.

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -222,18 +222,7 @@ public abstract class BaseLivyInterpreter extends Interpreter {
       try {
         candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
       } catch (SessionNotFoundException e) {
-        LOGGER.warn("Livy session {} is expired, new session will be created.", sessionInfo.id);
-        // we don't want to create multiple sessions because it is possible to have
-        // multiple thread
-        // to call this method, like LivySparkSQLInterpreter which use
-        // ParallelScheduler. So we need
-        // to check session status again in this sync block
-        synchronized (this) {
-          if (isSessionExpired()) {
-            initLivySession();
-          }
-        }
-        candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
+        LOGGER.warn("Livy session {} is expired. Will return empty list of candidates.", sessionInfo.id);
       }
     } catch (LivyException le) {
       logger.error("Failed to call code completions. Will return empty list of candidates", le);
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -219,11 +219,9 @@ public abstract class BaseLivyInterpreter extends Interpreter {
       InterpreterContext interpreterContext) {
     List<InterpreterCompletion> candidates = Collections.emptyList();
     try {
-      try {
-        candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
-      } catch (SessionNotFoundException e) {
-        LOGGER.warn("Livy session {} is expired. Will return empty list of candidates.", sessionInfo.id);
-      }
+      candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
+    } catch (SessionNotFoundException e) {
+      LOGGER.warn("Livy session {} is expired. Will return empty list of candidates.", sessionInfo.id);
     } catch (LivyException le) {
       logger.error("Failed to call code completions. Will return empty list of candidates", le);
     }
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks for the contribution @pellmont , but I don't think it make sense to make so large change just for unit test. Actually livy interpreter can run integration which is much better than unit test to guarantee the function of livy interpreter.

pellmont:
welcome @zjffdu.
Actually I am not a big fan of integration tests, because they are usually slow and brittle. In my opinion integration tests should be used sparely, the main functionality should be tested by unit tests.
Also the way I put it now separates the code for making the HTTP call of the actual features of the livy API.
But if you prefer, I can put everything into BaseLivyInterpreter...

zjffdu:
IMHO，it is better to have integration test for interpreter. Because interpreter is usually 3rd party library outside zeppelin. Unit test usually has many assumption which may not be true in real scenario. While integration test can simulate the real scenario.

zjffdu:
@pellmont have you verified in from end to end ? It seems doesn't work for me when I type tab in frontend. I think it may need frontend change as well.

pellmont:
Of course I did :-)
But:
- Code-Completion is by ctrl-. (as before ZEPPELIN-277). If it should be tab it has to be configured probably in livy/interpreter-settings.json
- You need a fresh 0.5.0-SNAPSHOT version of livy (since 0.5.0 with LIVY-7 is not yet released). But code completion in LivyInterpreter won't produce any errors if the feature is not available in livy...

zjffdu:
@pellmont Could you update interpreter-setting.json to make tab as the completion key ? Because livy should be consistent with the built-in spark interpreter.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -205,6 +215,47 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
 
   @Override
+  public List<InterpreterCompletion> completion(String buf, int cursor,
+      InterpreterContext interpreterContext) {
+    List<InterpreterCompletion> candidates = Collections.emptyList();
+    try {
+      try {
+        candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
+      } catch (SessionNotFoundException e) {
+        LOGGER.warn("Livy session {} is expired, new session will be created.", sessionInfo.id);
```
zjffdu: Have concern whether we need to recreate session when code completion is called.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -205,6 +215,47 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
+  private List<InterpreterCompletion> callCompletion(CompletionRequest req) throws LivyException {
+    List<InterpreterCompletion> candidates = new ArrayList<>();
+    try {
+      CompletionResponse resp = CompletionResponse.fromJson(
+          callRestAPI("/sessions/" + sessionInfo.id + "/completion", "POST", req.toJson()));
+      for (String candidate : resp.candidates) {
+        candidates.add(new InterpreterCompletion(candidate, candidate, StringUtils.EMPTY));
+      }
+    } catch (APINotFoundException e) {
+      logger.debug("completion api seems not to be availabele. (available from livy 0.5)", e);
```
zjffdu: typo. `availabele`

zjffdu:
ping @pellmont I also left some comments, in case you miss it.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
pellmont: you are right. and the chance for getting a useful result is also probably quite low (since context is missing).
Will remove.
Just return empty result?

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
pellmont: thanks, will fix.

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -205,6 +215,36 @@ public InterpreterResult interpret(String st, InterpreterContext context) {
...
   @Override
+  public List<InterpreterCompletion> completion(String buf, int cursor,
+      InterpreterContext interpreterContext) {
+    List<InterpreterCompletion> candidates = Collections.emptyList();
+    try {
+      try {
+        candidates = callCompletion(new CompletionRequest(buf, getSessionKind(), cursor));
+      } catch (SessionNotFoundException e) {
+        LOGGER.warn("Livy session {} is expired. Will return empty list of candidates.", sessionInfo.id);
+      }
```
zjffdu: Why not reduce 2 try catch to 1 ?

zjffdu:
@pellmont CI is failed, could you rebase and rerun it ?

Tagar:
This commit seems broke Tab to indent text behavior.
https://issues.apache.org/jira/browse/ZEPPELIN-3253

PR #2625, opened 2017-10-16
ZEPPELIN-2988. Zeppelin does not start new Livy session if yarn livy session application is killed

zjffdu:
### What is this PR for?

This PR will display a warning message when the livy session is dead. It won't restart it automatically, because livy session may be dead due to user code (like driver OOM). Restarting it automatically will hide the root cause and confuse users. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2988

* Manually tested

image

### Questions:


zjffdu:
@prabhjyotsingh Could you help review it ? Thanks

prabhjyotsingh:
> Restarting it automatically will hide the root cause and confuse users.

Make sense, LGTM

zjffdu:
Merge if no more comments

PR #2628, opened 2017-10-23
[ZEPPELIN-2999] Cannot create shell interpreter without timeout property

prabhjyotsingh:
### What is this PR for?
A user can not run any shell interpreter if "shell.command.timeout.millisecs" is not present in interpreter setting.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [https://issues.apache.org/jira/browse/ZEPPELIN-2999](https://issues.apache.org/jira/browse/ZEPPELIN-2999)

* Remove "shell.command.timeout.millisecs" from sh interpreter settings and then try to run any sh paragraph, it should run without any error.

### Questions:

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -49,7 +49,7 @@ public class ShellInterpreter extends KerberosInterpreter {
   private static final Logger LOGGER = LoggerFactory.getLogger(ShellInterpreter.class);
 
   private static final String TIMEOUT_PROPERTY = "shell.command.timeout.millisecs";
-  private Long DEFAULT_TIMEOUT_PROPERTY = 60000l;
+  private String DEFAULT_TIMEOUT_PROPERTY = "60000";
 
   private static final String DIRECTORY_USER_HOME = "shell.working.directory.user.home";
   private final boolean isWindows = System.getProperty("os.name").startsWith("Windows");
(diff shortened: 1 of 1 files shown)


toughrogrammer:
I'm reporter of ZEPPELIN-2999. IMO it's invalid solution. With this code, it cannot run without timeout. What I intend with no timeout option is I will approve waiting command execution forever.

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -98,7 +101,15 @@ public InterpreterResult interpret(String cmd, InterpreterContext contextInterpr
       DefaultExecutor executor = new DefaultExecutor();
       executor.setStreamHandler(new PumpStreamHandler(
         contextInterpreter.out, contextInterpreter.out));
-      executor.setWatchdog(new ExecuteWatchdog(Long.valueOf(getProperty(TIMEOUT_PROPERTY))));
+
+      Long timeOutProperty;
+      try {
+        timeOutProperty = Long.valueOf(getProperty(TIMEOUT_PROPERTY));
```
zjffdu: you can use getProperty(key, default)

prabhjyotsingh: Sure, I've made the suggested changes.

prabhjyotsingh:
@toughrogrammer  What I think you are looking for is "INFINITE_TIMEOUT = -1L", in which case I would recommend instead of removing this "shell.command.timeout.millisecs" parameter pass "-1" in the Zeppelin interpreter setting. Any thoughts?

toughrogrammer:
@prabhjyotsingh Exactly.

prabhjyotsingh:
Thanks for the review, will merge this if no more discussion.

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -47,7 +47,10 @@
  */
 public class ShellInterpreter extends KerberosInterpreter {
   private static final Logger LOGGER = LoggerFactory.getLogger(ShellInterpreter.class);
+
   private static final String TIMEOUT_PROPERTY = "shell.command.timeout.millisecs";
+  private String DEFAULT_TIMEOUT_PROPERTY = "60000";
```
toughrogrammer: To wait forever, it should be modified.

prabhjyotsingh: That is what I was trying to convey earlier, the default is 60 second, which means even if user "accidentally" deletes this key, by fail-safe design Zeppelin will make sure it stays 60 seconds. However, if user wants to, can set this to "-1" which will mean infinite timeout (wait forever).

zjffdu: @prabhjyotsingh I think @toughrogrammer means to make -1 by default which seems make sense. e.g. if user invoke wget in shell interpreter, it is very likely to take more than 60 seconds.

prabhjyotsingh: Then wouldn't it conflict with the default recommended value that Zeppelin is shipped with http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/interpreter/shell.html#configuration. 

I mean we can make a change in all of these three places (this, doc and interpreter-setting.json), do we all agree that 60sec is too less, or should we make "-1" as the default value.

felixcheung: I'm concerned about having anything waiting indefinitely...

prabhjyotsingh: Based on above comment, will merge this if no more discussion.

PR #2631, opened 2017-10-25
ZEPPELIN-2197. Interpreter Lifecycle Manager

zjffdu:
### What is this PR for?
This PR implement the lifecycle manager. There're 2 implementions.
* NullLifecycleManager.     Nothing is done as before. User has to start/stop interpreter explicitly in UI. 
* TimeoutLifecycleManager.    Interpreter will be closed after idle for one threshold of time. By default it is 1 hour.

### What type of PR is it?
[ Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2197

Unit test is added.

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -411,6 +411,25 @@
   <description>Enable directory listings on server.</description>
 </property>
 
+<property>
+  <name>zeppelin.interpreter.lifecyclemanager.class</name>
+  <value>org.apache.zeppelin.interpreter.lifecycle.TimeoutLifecycleManager</value>
+  <description>LifecycleManager class for managing the lifecycle of interpreters, by default interpreter will
+  be closed after timeout</description>
+</property>
+
+<property>
+  <name>zeppelin.interpreter.lifecyclemanager.timeout.checkinterval</name>
+  <value>60000</value>
+  <description>milliseconds of the interval to checking whether interpreter is time out</description>
+</property>
+
+<property>
+  <name>zeppelin.interpreter.lifecyclemanager.timeout.threshold</name>
+  <value>3600000</value>
+  <description>milliseconds of the interpreter timeout threshold, by default it is 1 hour</description>
+</property>
+
 <!--
 <property>
     <name>zeppelin.server.jetty.name</name>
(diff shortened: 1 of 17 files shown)


zjffdu:
@Leemoonsoo @jongyoul @Tagar  Please help review. Thanks

jongyoul:
I have a basic question. Does it work in `scoped` and `isolated`?

Tagar:
Thank you @zjffdu.
I just thought about this scenario: a Spark job runs for 1.5 hours, would it be killed by the LifeCycleManager in this case? (assuming here default timeout of 1 hour)
If it is, then might be nice also to have a grace period when an interpreter wouldn't be killed if it has a running job?
In the above example, let's say timeout=1 hour and grace period=1 hour. So an interpreter would be killed if it is completely inactive for 1 hour, or in 2 hours if it had a Spark job that was still spinning.
Thoughts?

zjffdu:
@Tagar It won't be killed, because LifecycleManager will know client is polling job progress via api Interpreter.getProgress.  https://github.com/apache/zeppelin/pull/2631/files#diff-f55969c5887615d2142806214747bba9R330

zjffdu:
@jongyoul For now, only Interpreter Process's lifecycle is controlled in TimeoutLifecycleManager. session level control will be done in future if necessary. https://github.com/apache/zeppelin/pull/2631/files#diff-53b72da47096c21dca5f395fe86761b3R16

Gauravshah:
@zjffdu what if I am not on my desk and not polling the job ?

zjffdu:
@Gauravshah It won't be killed. Because JobRunner in zeppelin server process will poll the job status periodically .  Add one more test to verify it.

Tagar:
@zjffdu got it - thank you.

Tagar:
One last thing - from user experience it would be convenient to know when their interpreters timed out. 
Something like a popup or just some sort of a graphical flag would do, I guess?
Not sure how hard it'll be add this at this point.

zjffdu:
@Tagar Good point, Currently there's no such things in frontend. For now, what user see in frontend is that a new interpreter group is created. (For spark interpreter, a new spark app is started)

jongyoul:
Got it. Thanks. I will be able to help you for the future feature.

zjffdu:

felixcheung:
could we add some documentation on this? this sounds important...

zjffdu:
Make sense, will add document for this feature.

PR #2633, opened 2017-10-26
[Bug Fix] fix exceptions in jdbc interpreter

tinkoff-dwh:
### What is this PR for?
Now used class Throwables to convert stack trace into String. This class from guava which is dependency of hadoop-common. hadoop-common is **provided**. Throwables changed to ExceptionsUtils (from commons-lang), commons-lang there is in **zeppelin-interpreter**.

### What type of PR is it?
[Bug Fix]

Run incorrect query

before

after

### Questions:


jongyoul:
LGTM

tinkoff-dwh:
merge?

PR #2635, opened 2017-10-30
ZEPPELIN-3013. Only parse paragraph text when text is updated

zjffdu:
### What is this PR for?

Before this PR, each time when paragraph use interpreter, it would parse the text to get the interpreter component which is inefficient. This PR would do it when `setText` is called. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3013

* Unit test is added

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -1515,7 +1515,7 @@ public class NotebookServer extends WebSocketServlet
     if (paragraph == null) {
       throw new IllegalArgumentException("Unknown paragraph with id : " + paragraphId);
     }
-    return paragraph.getInterpreter().getInterpreterGroup();
+    return paragraph.getBindedInterpreter().getInterpreterGroup();
   }
 
   private void pushAngularObjectToRemoteRegistry(String noteId, String paragraphId, String varName,
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -1515,7 +1515,7 @@ public class NotebookServer extends WebSocketServlet
     if (paragraph == null) {
       throw new IllegalArgumentException("Unknown paragraph with id : " + paragraphId);
     }
-    return paragraph.getCurrentRepl().getInterpreterGroup();
+    return paragraph.getInterpreter().getInterpreterGroup();
   }
 
   private void pushAngularObjectToRemoteRegistry(String noteId, String paragraphId, String varName,
@@ -1780,7 +1780,7 @@ public class NotebookServer extends WebSocketServlet
     // if it's the last paragraph and not empty, let's add a new one
     boolean isTheLastParagraph = note.isLastParagraph(p.getId());
     if (!(Strings.isNullOrEmpty(p.getText()) ||
-        p.getText().trim().equals(p.getMagic())) &&
+        Strings.isNullOrEmpty(p.getScriptText())) &&
         isTheLastParagraph) {
       Paragraph newPara = note.addNewParagraph(p.getAuthenticationInfo());
       broadcastNewParagraph(note, newPara);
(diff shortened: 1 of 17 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -73,21 +72,24 @@
  */
 public class Paragraph extends Job implements Cloneable, JsonSerializable {
 
-  private static final long serialVersionUID = -6328572073497992016L;
-
   private static Logger logger = LoggerFactory.getLogger(Paragraph.class);
-  private transient InterpreterFactory factory;
-  private transient InterpreterSettingManager interpreterSettingManager;
+  private static Pattern REPL_PATTERN = Pattern.compile("%([\\w\\.]+).*", Pattern.DOTALL);
```
zjffdu: use Reg expression to parse text which is much more concise.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -620,44 +615,7 @@ public boolean run(String paragraphId) {
   public boolean run(String paragraphId, boolean blocking) {
     Paragraph p = getParagraph(paragraphId);
     p.setListener(jobListenerFactory.getParagraphJobListener(this));
-
```
zjffdu: Move this into Paragraph.java. Paragraph has sufficient info to run itself instead of doing it in Note.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
felixcheung: hmm,... I think we break this often when we change it, could we have some more tests for this?
for example, does this work properly if the paragraph text is 
```
foo %3
```

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1515,7 +1515,7 @@ private InterpreterGroup findInterpreterGroupForParagraph(Note note, String para
     if (paragraph == null) {
       throw new IllegalArgumentException("Unknown paragraph with id : " + paragraphId);
     }
-    return paragraph.getCurrentRepl().getInterpreterGroup();
+    return paragraph.getInterpreter().getInterpreterGroup();
```
felixcheung: I think `getInterpreter` is a bit generic - as we already have `getInterpreter` in other classes and their purposes are very different from this?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
zjffdu: Add more test in `ParagraphTest.java`

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
zjffdu: Change it to `getBindedInterpreter`

zjffdu:
@felixcheung @prabhjyotsingh Could you help review it ? Thanks

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java
```diff
@@ -95,36 +95,53 @@ public void replSingleCharName() {
...
+    assertEquals("% foo", paragraph.getScriptText());
+  }
+
+  @Test
   public void replNameEndsWithWhitespace() {
     Note note = createNote();
     Paragraph paragraph = new Paragraph(note, null, interpreterFactory);
     paragraph.setText("%test\r\n###Hello");
-    assertEquals("test", paragraph.getReplText());
+    assertEquals("test", paragraph.getIntpText());
```
felixcheung: not sure if it matters, how about some tests after a newline, if there isn't any?
like `\r\n%test` (which I think is valid) `%\r\ntest`

zjffdu: Good point, more test is added.

zjffdu:
Merge if no more comments

PR #2643, opened 2017-11-02
[ZEPPELIN-3021] fix bugs for paragraph terminated status judgement

wary:
### What is this PR for?
fix bugs: [ZEPPELIN-3021]
Paragraph terminated status judgement error if it's called for a disabled paragraph with status ready. As we know the method isTerminated of  class Paragraph extends from the super class Job, it only consider the status of job

```java
 public boolean isTerminated() {
    return !this.status.isReady() && !this.status.isRunning() && !this.status.isPending();
  }
```
If we call this method in a disabled paragraph with status ready, it return false forever. If we schedule a note which have a paragraph like that, the cron job will be blocked by the code:

```java
while (!note.isTerminated()) {
        try {
          Thread.sleep(1000);
        } catch (InterruptedException e) {
          logger.error(e.toString(), e);
        }
      }
```

if too many cron jobs blocked by that problem, all cron jobs will be blocked because the quartz schedule thread pool is full.

* overwrite metod isTerminated in class Paragraph, add disable/enable status considered, if paragraph disabled return true directly

* add config null value judgement to avoid NullPointException, because when we create a paragraph the method isTerminated will be invoked in ParagraphListenerImpl.afterStatusChange before propery config initialized

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
[ZEPPELIN-3021](https://issues.apache.org/jira/browse/ZEPPELIN-3021)

Schedule a note which have a disabled paragraph with status ready， test weather the cron job will get blocked by the code block mentioned above

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -235,7 +235,7 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
     //avoid NullPointException when method isTerminated
     //invoked in ParagraphListenerImpl.afterStatusChange
     if (config == null) {
-      return true;
+      return false;
     }
     Boolean enabled = (Boolean) config.get("enabled");
     return enabled == null || enabled.booleanValue();
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -232,6 +232,11 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
   }
 
   public boolean isEnabled() {
+    //avoid NullPointException when method isTerminated
+    //invoked in ParagraphListenerImpl.afterStatusChange
+    if (config == null) {
+      return true;
+    }
     Boolean enabled = (Boolean) config.get("enabled");
     return enabled == null || enabled.booleanValue();
   }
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Thanks @wary for the fix. LGTM

felixcheung:
could you update the description?

wary:
@felixcheung  done！please help to check it again, thank you very much

felixcheung:
please also update/remove from the description here https://github.com/apache/zeppelin/pull/2643#issue-270506946
```
How should this be tested?

First time? Setup Travis CI as described on https://zeppelin.apache.org/contribution/contributions.html#continuous-integration
Strongly recommended: add automated unit tests for any new or changed behavior
Outline any manual steps to test the PR here.
Screenshots (if appropriate)

Questions:

Does the licenses files need update?
Is there breaking changes for older versions?
Does this needs documentation?
```

wary:
@felixcheung  done! thx for reminding

wary:
@felixcheung any other problems ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -232,6 +232,11 @@ public Note getNote() {
   }
 
   public boolean isEnabled() {
+    //avoid NullPointException when method isTerminated
+    //invoked in ParagraphListenerImpl.afterStatusChange
+    if (config == null) {
+      return true;
```
felixcheung: if no config, should this return false?

wary: @felixcheung  According  to code：
```java
      Boolean enabled = (Boolean) config.get("enabled");
      return enabled == null || enabled.booleanValue();
 ```
we can know that if no enabled configured, return true, so I  think the default value of enabled is true. So I returned true at this place.

PR #2647, opened 2017-11-02
[ZEPPELIN-3012] Interpreter Permissions not working properly for groups

weand:
### What is this PR for?
This fixes issues when using Zeppelin with permissions for users belonging to more than one roles / groups. Constructor of org.apache.zeppelin.user.AuthenticationInfo didn't consider that the roles String will contain spaces, in case there is more than one role, e.g. 'role1, role2'.

This change fixes the issue, by invoking trim() on each role.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3012

* Enable shiro.ini, including 'admin' user
* Login with 'admin' and enable Interpreter Permissions for any Interpreter, granting access to 'role2'
* Try to use this Interpreter with 'user1', who belongs to role2.

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
@@ -57,8 +57,10 @@ public class AuthenticationInfo implements JsonSerializable {
     this.user = user;
     this.ticket = ticket;
     if (StringUtils.isNotBlank(roles) && roles.length() > 2) {
-      String[] r = roles.substring(1, roles.length() - 1).split(",");
-      this.roles = Arrays.asList(r);
+        this.roles = new ArrayList<>();
+        for (final String role : roles.substring(1, roles.length() - 1).split(",")) {
+            this.roles.add(role.trim());
+        }
     }
   }
 
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
@@ -20,13 +20,14 @@ package org.apache.zeppelin.user;
 
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
-import com.google.gson.Gson;
+
 import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.common.JsonSerializable;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
```diff
@@ -57,8 +58,11 @@ public AuthenticationInfo(String user, String roles, String ticket) {
     this.user = user;
     this.ticket = ticket;
     if (StringUtils.isNotBlank(roles) && roles.length() > 2) {
-      String[] r = roles.substring(1, roles.length() - 1).split(",");
-      this.roles = Arrays.asList(r);
+      this.roles = new ArrayList<>();
+      for (final String role : roles.substring(1, roles.length() - 1)
```
zjffdu: not sure why roles has single-quote around it which seems weird.

zjffdu:
LGTM

zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
weand: It's not single quotes but brackets, e.g. ["role1, role2"].
As far as I see, this comes from the client through the web socket connection from Method org.apache.zeppelin.socket.NotebookServer.onMessage(NotebookSocket, String). But is not related to this Issue here IMHO.

weand:
Tests added added. Can you merge that into master?

zeppelin-interpreter/src/test/java/org/apache/zeppelin/user/AuthenticationInfoTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Test
+  public void testRoles() {
+    final String roles = "[role1, role2, role with space]";
+
+    final AuthenticationInfo authenticationInfo = new AuthenticationInfo("foo",
+        roles, "bar");
+
+    assertEquals(authenticationInfo.getRoles(),
+        new ArrayList<>(Arrays.asList("role1", "role2", "role with space")));
```
zjffdu: Minor. Switch left and right. The left should be expected, and right is actual.

weand: Good catch. Fixed :+1:

zjffdu:
Will merge if no more discussion

PR #2650, opened 2017-11-05
[ZEPPELIN-3035] Ability to display the same result multiple times for each paragraph.

SleepyThread:
### What is this PR for?
Sometimes we need to have multiple visualizations with each result of the query. Currently, we can only do it by copying query which increases the load on the server.
There should be a way to specify the number of result copies we want and the results should be duplicated in the UI.

### What type of PR is it?
[ Feature | Documentation ]

### Todos
* [Done] - Adding the abilty to specify the number of results in UI.
* [Done] - Adding documentations for the same. 

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3035

* Open any new notebook and select the num results options in paragraph settings. 
* Run paragraph and see the results. 

image1

multiple_result_visualization

### Questions:
   No
   No
   Yes.


SleepyThread:
@felixcheung Yeah, I am new to Zeppelin project. I have incorporated your feedback and updated the pull request accordingly.

SleepyThread:
@felixcheung Is this pull request not relevant or team is busy with the backlog of the other task?

felixcheung:
@Leemoonsoo @1ambda what do you think?

Leemoonsoo:
Thanks @SleepyThread for the contribution!
I have tested this branch. I'd like to more people take a look and leave feedback on this feature.

Anyway, this looks like break current behavior. i.e. You can't print multiple results in one paragraph.
Run following paragraph in both master and this branch.
```
%sh echo -e "%table key\tvalue\nsun\t100\nmoon\t200\n\n%table fruit\tcolor\norange\t1\napple\t2"
```

Master branch will shows two table and this branch will display only the first one.

PR #2664, opened 2017-11-10
[ZEPPELIN-771] Escape tab and CR when the result is shown in table

mgaido91:
### What is this PR for?

The presence of the characters `\t` and `'n` in the data to be displayed can cause the failure of successfully displaying a table. Indeed, these two characters are used as separators in a table. Some interpreters, don't escape the data to be shown.
This PR address the issue replacing the characters as previosly done for some of the interpreters like geode.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-771

added UTs

I am not uploading the screenshot since I think it is not necessary to understand the issue and the PR.

### DIFF
--- a/cassandra/src/main/scala/org/apache/zeppelin/cassandra/InterpreterLogic.scala
+++ b/cassandra/src/main/scala/org/apache/zeppelin/cassandra/InterpreterLogic.scala
@@ -207,7 +207,10 @@ class InterpreterLogic(val session: Session)  {
         .append("%table ")
         .append(columnsDefinitions.map { case (columnName, _) => columnName }.mkString("\t")).append("\n")
 
-      val escape: AnyRef => String = _.toString.replaceAllLiterally("\t", " ").replaceAllLiterally("\n", " ")
+      val escape: AnyRef => String = {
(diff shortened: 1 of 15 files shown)


mgaido91:
@felixcheung Since so far `\t` and `\n` are used as delimiters, there is no way to do that only on the front-end side, because currently there is no way for the frontend to know whether it is a `\t` or `\n` in the field or a separator.
If we want to fix this in the front-end, I think this is a much bigger task which would require a whole refactor of the communication protocol between back-end and frontend (so that we don't have anymore such separator or that there is a way to escape them) and the frontend itself in order to be able to manage those things. If this is the way we choose to go I am fine and I can help with them, but this would probable require a design document and some splitting in subtasks since it is going to be a huge refactor.
What do you think?

zjffdu:
@mgaido91 Check the comments on https://issues.apache.org/jira/browse/ZEPPELIN-2992

mgaido91:
Thanks @zjffdu , then I am closing this. Let me know if I can help somehow.

PR #2671, opened 2017-11-15
[ZEPPELIN-3045] Add the option which prevents the cron executing user from being changed to users other than the login user

kjmrknsn:
### What is this PR for?
Add the option to Zeppelin, that prevents the cron executing user from being changed to users other than the login user.

Now, the cron executing user can be set to any users. That could cause a security issue when the Hadoop cluster has data access control (managed by Apache Ranger and so forth) because Zeppelin users can access the data which they ordinary cannot access by setting other users as the cron executing user.

Thus, under the circumstances strict data access control is required, it is necessary to add the option to Zeppelin, that prevents the cron executing user from being changed to users other than the login user.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3045

* Tested manually
    * I confirmed that when `zeppelin.notebook.cronExecutingUser.loginUserOnly` was true and "Cron executing user" was changed to users other than the login user, the "Insufficient privileges" dialog was shown and "Cron executing user" was not updated.
    * screen shot 2017-11-15 at 22 15 29

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -726,8 +726,7 @@ public class NotebookServer extends WebSocketServlet
   }
 
   private void cronExecutingUserUpdatePermissionError(NotebookSocket conn, String op, String user,
-                                                      String cronExecutingUser)
-      throws IOException {
+                                                      String cronExecutingUser) throws IOException {
     LOG.info("Cannot {} due to the cron executing user update permission error. user: {}, " +
                     "submitted cron executing user: {}", op, user, cronExecutingUser);
 
@@ -1059,8 +1058,8 @@ public class NotebookServer extends WebSocketServlet
                                            Map<String, Object> configA,
                                            Map<String, Object> configB) {
     boolean updated = false;
-    if (configA.get(key) != null && configB.get(key) != null && configA.get(key)
-        .equals(configB.get(key))) {
+    if (configA.get(key) != null && configB.get(key) != null &&
+            configA.get(key).equals(configB.get(key))) {
       updated = false;
     } else if (configA.get(key) == null && configB.get(key) == null) {
       updated = false;
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -725,6 +725,18 @@ public class NotebookServer extends WebSocketServlet
             .toString())));
   }
 
+  private void cronExecutingUserUpdatePermissionError(NotebookSocket conn, String op, String user,
+                                                      String cronExecutingUser)
+      throws IOException {
+    LOG.info("Cannot {} due to the cron executing user update permission error. user: {}, " +
+                    "submitted cron executing user: {}", op, user, cronExecutingUser);
+
+    conn.send(serializeMessage(new Message(OP.AUTH_INFO).put("info",
+            "Cron executing user cannot be set to users other than the login user.\n\n" +
+                    "user: " + user + ", submitted cron executing user: " +
+                    cronExecutingUser)));
+  }
+
   /**
    * @return false if user doesn't have reader permission for this paragraph
    */
@@ -799,6 +811,48 @@ public class NotebookServer extends WebSocketServlet
     }
 
     return true;
+  }
+
+  /**
+   *
+   * @param isLoginUserOnly
+   * @param curConf
+   * @param newConf
+   * @return true if the cron executing user update permission check is needed.
+   */
+  private boolean isCronExecutingUserUpdatePermissionCheckNeeded(boolean isLoginUserOnly,
+                                                                 Map<String, Object> curConf,
+                                                                 Map<String, Object> newConf) {
+    // Permission check is not needed if the cron executing user is not
+    // restricted to the login user.
+    if (!isLoginUserOnly) {
+      return false;
+    }
+
+    // Permission check is not needed if the cron executing user is not updated.
+    if (!isNoteConfigValueUpdated("cronExecutingUser", curConf, newConf)) {
+      return false;
+    }
+
+    // Permission check is not needed if the cron configuration is cleared.
+    if (StringUtils.isEmpty((String) newConf.get("cron"))
+            && StringUtils.isEmpty((String) newConf.get("cronExecutingUser"))) {
+      return false;
+    }
+
+    return true;
+  }
+
+  /**
+   * @return false if user doesn't have cron executing user update permission
+   */
+  private boolean hasCronExecutingUserUpdatePermission(String user,
+                                                       String cronExecutingUser) {
+    if (cronExecutingUser == null) {
+      return true;
+    }
+
+    return cronExecutingUser.equals(user);
   }
 
   private void sendNote(NotebookSocket conn, HashSet<String> userAndRoles, Notebook notebook,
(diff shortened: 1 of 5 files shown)


kjmrknsn:
The build result on Travis CI turned to green by rebuilding: https://travis-ci.org/kjmrknsn/zeppelin/builds/302473619

Leemoonsoo:
Thanks @kjmrknsn for the great contribution.

One question, how about make `zeppelin.notebook.cronExecutingUser.loginUserOnly` default value `true`?  Does it make sense to you? or do you see any possible problem on making it true by default?

kjmrknsn:
@Leemoonsoo Thanks for your comment and suggestion.

> One question, how about make zeppelin.notebook.cronExecutingUser.loginUserOnly default value true? Does it make sense to you? or do you see any possible problem on making it true by default?

I'll change the default value to `true`.
I set it to `false` just because I didn't know whether I might change the default behavior of Zeppelin or not.

kjmrknsn:
The default value of `zeppelin.notebook.cronExecutingUser.loginUserOnly` was changed to `true`. The Commits are squashed.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -725,6 +725,18 @@ void permissionError(NotebookSocket conn, String op, String userName, Set<String
             .toString())));
   }
 
+  private void cronExecutingUserUpdatePermissionError(NotebookSocket conn, String op, String user,
+                                                      String cronExecutingUser)
+      throws IOException {
```
felixcheung: this is a bit hard to read - could it fit after the last line?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -989,18 +1055,20 @@ private void renameFolder(NotebookSocket conn, HashSet<String> userAndRoles,
...
-    } else if (configA.get("cron") == null && configB.get("cron") == null) {
-      cronUpdated = false;
-    } else if (configA.get("cron") != null || configB.get("cron") != null) {
-      cronUpdated = true;
+  private boolean isNoteConfigValueUpdated(String key,
+                                           Map<String, Object> configA,
+                                           Map<String, Object> configB) {
+    boolean updated = false;
+    if (configA.get(key) != null && configB.get(key) != null && configA.get(key)
+        .equals(configB.get(key))) {
```
felixcheung: a bit hard to read - put `configA.get(key)` before `.equals(configB.get(key))`

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -879,7 +933,19 @@ private void updateNote(NotebookSocket conn, HashSet<String> userAndRoles, Noteb
 
     Note note = notebook.getNote(noteId);
     if (note != null) {
-      boolean cronUpdated = isCronUpdated(config, note.getConfig());
+      if (isCronExecutingUserUpdatePermissionCheckNeeded(
+              notebook.getConf().isNotebookCronExecutingUserLoginUserOnly(),
+              note.getConfig(),
+              config)) {
+        String cronExecutingUser = (String) config.get("cronExecutingUser");
```
felixcheung: check that `cronExecutingUser` is a valid value?

felixcheung: same in other places

zeppelin-server/src/test/java/org/apache/zeppelin/socket/NotebookServerTest.java
```diff
@@ -412,6 +416,88 @@ public void testCreateNoteWithDefaultInterpreterId() throws IOException {
...
+    assertTrue((boolean) method.invoke(server, "user1", null));
+    assertFalse((boolean) method.invoke(server, "user1", ""));
+    assertTrue((boolean) method.invoke(server, "user1", "user1"));
+    assertFalse((boolean) method.invoke(server, "user1", "user2"));
+  }
+
+  @Test
+  public void testIsCronExecutingUserUpdatePermissionCheckNeeded()
+          throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+     Method method = NotebookServer.class.getDeclaredMethod(
```
zjffdu: It is a too complicated to me to use reflection for the unit test of this trivial method that only compare 2 variables.

zeppelin-server/src/test/java/org/apache/zeppelin/socket/NotebookServerTest.java
```diff
@@ -412,6 +416,88 @@ public void testCreateNoteWithDefaultInterpreterId() throws IOException {
     notebook.removeNote(createdNote.getId(), anonymous);
   }
 
+  @Test
+  public void testHasCronExecutingUserUpdatePermission()
+          throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+    Method method = NotebookServer.class.getDeclaredMethod("hasCronExecutingUserUpdatePermission",
```
zjffdu: It is a too complicated to me to use reflection for the unit test of this trivial method that only compare 2 variables.

zjffdu:
Would it be much easier that we can just do not allow user to set cron executing user in frontend when this `zeppelin.notebook.cronExecutingUser.loginUserOnly` is true  ? So that there would be less change in backend.

kjmrknsn:
> Would it be much easier that we can just do not allow user to set cron executing user in frontend when this zeppelin.notebook.cronExecutingUser.loginUserOnly is true ? So that there would be less change in backend.

It seems nice. I'll try that later.

kjmrknsn:
@felixcheung @zjffdu Thanks for your reviewing. I'll rewrite this PR to modify frontend instead of backend as @zjffdu said at the beginning of next week.

felixcheung:
we still want a check in the backend right, since this is a security/authentication issue.

kjmrknsn:
@felixcheung @zjffdu 

> we still want a check in the backend right, since this is a security/authentication issue.

OK, thanks.  So, I won't rewrite this PR to modify frontend. I'll reply to your review comments.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
kjmrknsn: @felixcheung This was fixed at https://github.com/apache/zeppelin/pull/2671/commits/f4f53c5e9d65a31d57cadc86820a96f385cedee7. Thanks.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
kjmrknsn: @felixcheung This was fixed at https://github.com/apache/zeppelin/pull/2671/commits/f4f53c5e9d65a31d57cadc86820a96f385cedee7. Thanks.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
kjmrknsn: @felixcheung I think it is simple to just pass the content of `config.get("cronExecutingUser")` to the `hasCronExecutingUserUpdatePermission` and `cronExecutingUserUpdatePermissionError` methods without validating it at this point.

The case that `cronExecutingUser` is null is considered at `hasCronExecutingUserUpdatePermission` and `cronExecutingUserUpdatePermissionError` methods.

What do you think?

zeppelin-server/src/test/java/org/apache/zeppelin/socket/NotebookServerTest.java
kjmrknsn: @zjffdu Reflection is necessary to test the `hasCronExecutingUserUpdatePermission` method because it is a private method. Should I omit this `testHasCronExecutingUserUpdatePermission` test code to keep the code clean and simple?

zjffdu: Usually private method is necessary to be tested, instead we should test the public method which use this private method.

kjmrknsn: @zjffdu OK, thanks. In this case, I decided to remove this `testHasCronExecutingUserUpdatePermission` unit test because the test target method was so trivial at https://github.com/apache/zeppelin/pull/2671/commits/718e32955d3e6bca26aff0bd8b0a05f71669ac3a.

prabhjyotsingh:
Ping @kjmrknsn, are you still working on this?

msk84:
For me having the solution with the restriction to set only myself as a cron user would be perfect.

zjffdu:
@msk84 This is what I propsed in [ZEPPELIN-3350](https://issues.apache.org/jira/browse/ZEPPELIN-3350)

PR #2673, opened 2017-11-16
[ZEPPELIN-3054] Update conf/shiro.ini.template so that non-admin users can restart interpreters on a notebook page

kjmrknsn:
### What is this PR for?
On the default configuration of `conf/shiro.ini.template`, non admin users cannot restart interpreters on their notebook pages because the path `/api/interpreter/setting/restart/**` is not allowed to being accessed by non admin users.

screen shot 2017-11-16 at 14 07 05

It is convenient that all authenticated users can restart interpreters on their notebook pages by default, so the following line should be inserted into conf/shiro.ini.template:

```
/api/interpreter/setting/restart/** = authc
```

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3054

* Tested manually. I confirmed that non admin users can restart interpreters on a notebook page but still cannot access to the interpreter settings page by using `conf/shiro.ini.template`.

### Questions:

### DIFF
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -96,6 +96,9 @@ admin = *
 # uncomment the line second last line (/** = anon) and comment the last line (/** = authc)
 #
 /api/version = anon
+# Allow all authenticated users to restart interpreters on a notebook page.
+# Comment out the following line if you would like to authorize only admin users to restart interpreters.
+/api/interpreter/setting/restart/** = authc
 /api/interpreter/** = authc, roles[admin]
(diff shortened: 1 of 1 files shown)


kjmrknsn:
The build result on Travis CI turned to green: https://travis-ci.org/kjmrknsn/zeppelin/builds/302858138

Leemoonsoo:
LGTM!

Leemoonsoo:
Merge to master if no further discussions. Thanks @kjmrknsn for the improvement!

PR #2675, opened 2017-11-16
[ZEPPELIN-3060] Unable to use interpreter names with special characters (e.g. dash or underscore)

weand:
[ZEPPELIN-3060] Unable to use interpreter names with special characters
(e.g. dash or underscore)

### What is this PR for?
This fixes a regression which was introduced with ZEPPELIN-3013. There
is no restriction regarding characters when creating new interpreters.
But when running a note the interpreter name was parsed using the regex
%([\w\.]+).*, which only considered the word character class and dots.
Interpreter names with special characters (e.g. _ or -) were not matched
correctly. This change enables usage of interpreter names that use other
characters by changing the regex group to %([^\s\(]+).*, which means:
parse the interpreter name from % until any whitespace or open
parenthesis.

### What type of PR is it?
Bug Fix

### Todos
-

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3060

Create an interpreter with a name containing one or more special
characters, e.g. dash or underscore. Try running that interpreter.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -18,6 +18,7 @@
 package org.apache.zeppelin.notebook;
 
 import java.io.IOException;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


zjffdu:
@weand In ZEPPELIN-3013, I try to make interpreter name only contains alphabet ,number and under score, just like valid variable name in most of programing languages. Is it necessary and normal for you to have special character in your interpreter ? although it is possible to do that, I would recommend user don't include special character in interpreter name.

felixcheung:
well, IMHO `-` and `_` are very common in programming languages too (eg. python)
I think it's reasonable to accept these two in names.

Not sure about other characters though, `(` can be problematic

zjffdu:
'_' is acceptable in interpreter name and current code support that I believe.
But `-` is not acceptable IMHO, because no programming language allow `-` in variable identifier IIRC.

weand:
Well, the current implementation does not enforce any naming pattern when creating a new interpreter.

Before ZEPPELIN-3013 interpreters with any characters e.g. - could be created and executed.

If that's not supported anymore, there should be a failure when creating such an interpreter.

The current behavior is quite misleading, as the interpreter name is not even display correctly in the error message. See the corresponding Jira ticket of this PR.

zjffdu:
Good point @weand , We should validate the interpreter name when creating new interpreter.

weand:
Closed and created https://issues.apache.org/jira/browse/ZEPPELIN-3083 instead.

PR #2556, opened 2017-08-30
[Zeppelin-2839] permissions for folders

tinkoff-dwh:
### What is this PR for?
Adds permissions for folders.
Folder with permissions is way to unite notes with the same permissions and change the permissions together.

### What type of PR is it?
[Feature]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2839

### How should this be tested?
folderperm2

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? Yes, added

PR #2676, opened 2017-11-17
[ZEPPELIN-3048] Add the option which prevents paragraphs from being edited while their notebook is cron scheduled

kjmrknsn:
### What is this PR for?

Add the option which prevents paragraphs from being edited by users other than the cron executing user while their notebook is cron scheduled.

Now, the cron scheduled notebook's paragraphs can be edited by users who have the "writer" authority. That could cause a security issue when the Hadoop cluster has data access control (managed by Apache Ranger and so forth) because the "writer" users can abuse the access right of the "cron executing user" and can access the data which they ordinary cannot access by updating the paragraphs whose notebook is cron scheduled and its cron executing user is set to other users.

Thus, under the circumstances strict data access control is required, it is necessary to add the option to Zeppelin, that prevents paragraphs from being edited by users other than the cron executing user while their notebook is cron scheduled.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3048

* Tested manually.
    * I confirmed that when `zeppelin.notebook.cron.restrictEdit` was set to `true` and users other than the cron executing user tried to edit the paragraph whose interpreter was not on `zeppelin.notebook.cron.editableInterpreters`,  the "Insufficient privileged" dialog was shown and the paragraph was not updated.
    * screen shot 2017-11-17 at 12 31 41

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -406,6 +406,18 @@
 </property>
 
 <property>
+  <name>zeppelin.notebook.cron.restrictEdit</name>
+  <value>false</value>
+  <description>Restrict editable interpreters of cron scheduled notebooks' paragraphs. The cron executing user still can edit all paragraphs of the notebook even if this property is set to true.</description>
+</property>
+
+<property>
+  <name>zeppelin.notebook.cron.editableInterpreters</name>
+  <value></value>
+  <description>Editable interpreters of cron scheduled notebooks' paragraphs. This property is in effect when zeppelin.notebook.cron.restrictEdit is set to true. This property value should be a fully qualified class name such as "org.apache.zeppelin.angular.AngularInterpreter". Comma separated value such as "org.apache.zeppelin.angular.AngularInterpreter,org.apache.zeppelin.markdown.Markdown" is acceptable.</description>
+</property>
+
+<property>
   <name>zeppelin.websocket.max.text.message.size</name>
   <value>1024000</value>
   <description>Size in characters of the maximum text message to be received by websocket. Defaults to 1024000</description>
--- a/docs/setup/operation/configuration.md
+++ b/docs/setup/operation/configuration.md
@@ -258,6 +258,18 @@ If both are defined, then the **environment variables** will take priority.
     <td>Make notebook public (set only <code>owners</code>) by default when created/imported. If set to <code>false</code> will add <code>user</code> to <code>readers</code> and <code>writers</code> as well, making it private and invisible to other users unless permissions are granted.</td>
   </tr>
   <tr>
+    <td><h6 class="properties">ZEPPELIN_NOTEBOOK_CRON_RESTRICT_EDIT</h6></td>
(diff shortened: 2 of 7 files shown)


felixcheung:
hmm, this is interesting, and likely important.
but given the set of issues related to cron perhaps it makes sense to rethink the entire auth/role-base access control around here? it feels like we are patching this with a bunch of extra config to manage instead of considering how this fits into the existing role system.

kjmrknsn:
@felixcheung I'm managing Zeppelin with a Hadoop cluster which data access control is introduced to via Apache Ranger because my company needs to control each data access to our Hadoop cluster strictly. I found only the following 2 security issues about the Zeppelin cron scheduler: 

1. https://github.com/apache/zeppelin/pull/2671
2. https://github.com/apache/zeppelin/pull/2676 (This PR)

In my opinion, it's enough to fix these 2 issues instead of reconsider the whole access control of Zeppelin.

Leemoonsoo:
Thanks @kjmrknsn for the contribution.

In case of a note is cron scheduled and doesn't want anyone else edit the notebook other than the cron execution user, wouldn't 'owner' of note remove users from 'writers' and add them in 'readers'?

kjmrknsn:
@Leemoonsoo 

> In case of a note is cron scheduled and doesn't want anyone else edit the notebook other than the cron execution user, wouldn't 'owner' of note remove users from 'writers' and add them in 'readers'?

Yes, ideally, every Zeppelin user should do so. However, it's not realistic to force every Zeppelin user to do so because almost all of them do not realize this kind of security issue, it takes a huge effort to do so if there are a lot of users and notebooks, or they might forget to do so. Thus, it is reasonable and secure to do this kind of permission control at the Zeppelin system side not the Zeppelin user side.

felixcheung:
I agree we should make it easier. My concerns are whether we are possibly making it more confusing if we have 2 separate systems? First with the roles, second this.

kjmrknsn:
@felixcheung @Leemoonsoo 

Then, how about the following solution which fixes this and #2671 together?

* Add the `zeppelin.notebook.cron.secure` property to Zeppelin Configuration.
* If `zeppelin.notebook.cron.secure` is set to `true`, 
   * A cron executing user cannot be set to other users than the login user
   * If a cron schedule is set to a notebook, its owners and writers are demoted to runners and the login user becomes its owner and writer.

tinkoff-dwh:
The security is also important for our team.
I agree with @Leemoonsoo and @felixcheung about the sufficiency of existing roles.

We have PR about permissions for folders https://github.com/apache/zeppelin/pull/2556
Our plan is to have restricted subfolder in each user and group folder. The subfolder with name "restricted" will be viewble only for user/group owner of parent folder. If user cares about the security then he will put note in restricted subfolder.

@kjmrknsn, could folders with permissions solve your issue?

kjmrknsn:
@tinkoff-dwh The folder level permission looks nice. However, in my company's use case, we'd like to make a notebook readable for many users and restrict the user who can edit it only to the cron executing user when the cron schedule is set to it.

PR #2678, opened 2017-11-18
[ZEPPELIN-3063] Notebook loses formatting when importing from 0.6.x

prabhjyotsingh:
### What is this PR for?
Notebook loses formatting (shows table instead of the graph) when importing from 0.6.x if the respective paragraph doesn't have result in it.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3063](https://issues.apache.org/jira/browse/ZEPPELIN-3063)

Before:
before

After:
after

* Try importing [this](https://issues.apache.org/jira/secure/attachment/12898326/oldnotebook-0.6-clear.json) notebook, and then on running this notebook, the second paragraph should display BarChart instead of a table.

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -447,6 +447,24 @@ public class Notebook implements NoteEventListener {
             }
             config.put("results", results);
           }
+        } else if (ret == null && p.getConfig() != null) {
+          if (p.getConfig().get("graph") != null && p.getConfig().get("graph") instanceof Map
+            && !((Map) p.getConfig().get("graph")).get("mode").equals("table")) {
+            Map<String, Object> config = p.getConfig();
+            Object graph = config.remove("graph");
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Tested and LGTM

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -445,6 +445,24 @@ public void convertFromSingleResultToMultipleResultsFormat(Note note) {
                 results.add(new HashMap<>());
               }
             }
+            config.put("results", results);
+          }
+        } else if (ret == null && p.getConfig() != null) {
+          if (p.getConfig().get("graph") != null && p.getConfig().get("graph") instanceof Map
```
felixcheung: can you add a comment here what this block of code is for? this way we could consider removing it when we remove support for 0.6 format perhaps?

prabhjyotsingh:
Thanks for the review, have made the suggested changes

prabhjyotsingh:

PR #2680, opened 2017-11-20
[ZEPPELIN-3066] Make the welcome header of the homepage for non login users customizable

kjmrknsn:
### What is this PR for?
Make the welcome header of the homepage for non login users customizable.

In our company's use case, we would like to customize the contents of the welcome header of the homepage for non login users so that we can introduce how to create an LDAP account to login Zeppelin to them.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3066

* Tested manually.
    * I confirmed that the content of the HTML file whose path was set by `zeppelin.homescreen.header` was shown on the top of the home screen page.
        * screen shot 2017-11-20 at 18 13 33
    * I also confirmed that if `zeppelin.homescreen.header` was not set, the default contents (`app/home/header.html`) was shown.
        * screen shot 2017-11-20 at 18 14 30

### Questions:

### DIFF
--- a/conf/shiro.ini.template
+++ b/conf/shiro.ini.template
@@ -95,6 +95,7 @@ admin = *
 # To allow anonymous access to all but the stated urls,
 # uncomment the line second last line (/** = anon) and comment the last line (/** = authc)
 #
+/api/configurations/key/zeppelin.homescreen.header = anon
 /api/version = anon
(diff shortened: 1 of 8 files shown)


kjmrknsn:
This PR is replaced with https://github.com/apache/zeppelin/pull/2682.

PR #2681, opened 2017-11-20
[ZEPPELIN-2995] "auto-restart interpreter on cron execution" should restart interpreter to specific note, not all interpreters

kjmrknsn:
### What is this PR for?
Make "auto-restart interpreter on cron execution" restart the interpreters which are specific to the note, not all interpreters.
This issue was reported by https://github.com/apache/zeppelin/pull/1302#issuecomment-336521420.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2995

* Tested Manually.
   * I confirmed that the "auto-restart interpreter on cron execution" feature restarted only the interpreters which were specific to the notebook.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -882,10 +882,14 @@ public class Notebook implements NoteEventListener {
       }
 
       boolean releaseResource = false;
+      String cronExecutingUser = null;
       try {
         Map<String, Object> config = note.getConfig();
-        if (config != null && config.containsKey("releaseresource")) {
-          releaseResource = (boolean) note.getConfig().get("releaseresource");
+        if (config != null) {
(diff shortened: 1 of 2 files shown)


Leemoonsoo:
Thanks @kjmrknsn for the fix. It looks great to me!

If we can add a automated test, like [testAutoRestartInterpreterAfterSchedule ](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java#L405), it'll help make sure any future change doesn't break this behavior.

Do you think you can add a test? Let me know if you need help on adding test!

kjmrknsn:
@Leemoonsoo I'll try to add a test. Thanks.

kjmrknsn:
@Leemoonsoo The unit test was added to `NotebookTest.java`. Thanks.

Leemoonsoo:
Thanks @kjmrknsn for adding test!
LGTM and merge to master if no further reviews.

PR #2682, opened 2017-11-21
[ZEPPELIN-3066] Make the welcome header of the homepage for non login users customizable

kjmrknsn:
### What is this PR for?
Make the welcome header of the homepage for non login users customizable.

In our company's use case, we would like to customize the contents of the welcome header of the homepage for non login users so that we can introduce how to create an LDAP account to login Zeppelin to them.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3066

* Tested manually.
    * I confirmed that the content of the HTML file whose path was set by `zeppelin.homescreen.header` was shown on the top of the home screen page.
        * screen shot 2017-11-20 at 18 13 33
        * screen shot 2017-11-23 at 16 29 44
    * I also confirmed that if `zeppelin.homescreen.header` was not set, the default contents (`app/home/header.html`) was shown.
        * screen shot 2017-11-20 at 18 14 30
        * screen shot 2017-11-23 at 16 30 45

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
@@ -185,7 +185,13 @@ public class Message implements JsonSerializable {
     RUN_PARAGRAPH_USING_SPELL,    // [s-c] run paragraph using spell
     PARAS_INFO,                   // [s-c] paragraph runtime infos
     SAVE_NOTE_FORMS,              // save note forms
-    REMOVE_NOTE_FORMS             // remove note forms
+    REMOVE_NOTE_FORMS,            // remove note forms
+    GET_HOMESCREEN_HEADER,        // [c-s] get homescreen header
+    HOMESCREEN_HEADER;            // [s-c] homescreen header
+
+    public boolean isAnonymousAllowed() {
+      return this == GET_HOMESCREEN_HEADER;
+    }
   }
 
   private static final Gson gson = new Gson();
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -67,6 +67,11 @@
   <description>hide homescreen notebook from list when this value set to true</description>
 </property>
 
+<property>
+  <name>zeppelin.homescreen.header</name>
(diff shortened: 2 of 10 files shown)


kjmrknsn:
Travis CI status turned to green: https://travis-ci.org/kjmrknsn/zeppelin/builds/305087756

Leemoonsoo:
Thanks @kjmrknsn for the feature. However, this feature looks like subset of [customizing homepage](http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/usage/other_features/customizing_homepage.html).

Do you think [customizing homepage](http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/usage/other_features/customizing_homepage.html) provides what you need? if not, what makes it difficult to use in your use case?

kjmrknsn:
@Leemoonsoo We are also using the customizing homepage feature. However, in our company's use case, anonymous access is not allowed and the custom homepage notebook is shown to users only after they log in to Zeppelin. We would like to show a custom homepage to users who have not yet had their Zeppelin account (LDAP account) so that we can introduce them how to create their own Zeppelin account (LDAP account).

kjmrknsn:
@felixcheung 

> can you do a screen shot of anonymous access where only the header is shown?

I added the screen shots to the "How should this be tested?" section of the top comment.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
```diff
@@ -21,7 +21,9 @@
 import org.apache.zeppelin.common.JsonSerializable;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
```
kjmrknsn: The unnecessary imports of `java.util.HashSet` and `java.util.Set` were added accidentally. I'll remove them by the next commit.

mebelousov:
@kjmrknsn, if user knows zeppelin url, why he doesn't know about the need of LDAP account?

I believe that the custom welcome header is not nice way for announcing.
One may support guest account with open password.
Also the homescreen is excellent new feature, if it will be.

PR #2688, opened 2017-11-27
[Zeppelin-2571] & [Zeppelin-465] Run paragraphs: from first/current to current/last

tinkoff-dwh:
### What is this PR for?
This pr add the ability to run all paragraphs from the first to the current and from the current to the last. This makes it easier to update the data if changes are made in one of the related paragraphs.

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-2571](https://issues.apache.org/jira/browse/ZEPPELIN-2571) - (Add a "Run to here" option on paragraphs)
[ZEPPELIN-465](https://issues.apache.org/jira/browse/ZEPPELIN-465)     - (Capability to run all cells below the current cell)

1. Click on the "Run: from first to this" or "Run: from this to last" in the dropdown menu. Paragraphs from the first/current to the current/last will be started in order.
2. Press the key combination: CTRL + SHIFT + ENTER.
A window will appear with the choice of the desired action. 
capture2
Choose one of two actions. The selected action will be performed.

capture1_edit

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -155,7 +155,7 @@ limitations under the License.
                   style="position: relative; transform: rotate(-90deg); left: -4px;">
             </span>
             <span class="shortcut-keys">Ctrl+{{ isMac ? 'Option' : 'Shift'}}+Enter</span>
-             Run current and all below
+             Run all below
           </a>
         </li>
         <li>
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
@@ -145,7 +145,7 @@ limitations under the License.
             <span class="icon-action-redo shortcut-icon"
                   style="position: relative; transform: rotate(-90deg); left: -4px;">
             </span>
-            <span class="shortcut-keys">Ctrl+{{ isMac ? 'Option' : 'Shift'}}+Enter</span>
+            <span class="shortcut-keys">Ctrl+Shift+Enter</span>
             Run all above
           </a>
         </li>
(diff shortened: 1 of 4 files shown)


Leemoonsoo:
Thanks @tinkoff-dwh for the contribution. The feature works as expected.

Jupyer has "Run All Above" and "Run All Below".
image

One interesting thing is, "Run All Above" exclude current cell (paragraph), "Run All Below" include the current cell (paragraph).

I think there's a reason why Jupyer designed two very related button behavior in slightly different way. @tinkoff-dwh What do you think? Do you think Jupyter's "Run All Above" behavior is useful?

felixcheung:
I guess the Jupyter behavior sort of make senses to me...
It sounds like "Run All Above" means "runs everything from the top to right before this current paragraph", which can be useful for reruns.

tinkoff-dwh:
@Leemoonsoo 
Yes, I think it will be more convenient.
I made some changes.

Leemoonsoo:
Tested last commit and it works well. Looks good to me.

It's minor, if paragraph keeps focus after 'run all above' and 'run all below', that would be more convenient to use.

tinkoff-dwh:
@Leemoonsoo
Now paragraph's focus and cursor position stay in the same position after launch.

zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
```diff
@@ -140,6 +140,24 @@
             Insert new
           </a>
         </li>
+        <li>
+          <a ng-click="runAllToThis(paragraph)" ng-hide="$first">
+            <span class="icon-action-redo shortcut-icon"
+                  style="position: relative; transform: rotate(-90deg); left: -4px;">
+            </span>
+            <span class="shortcut-keys">Ctrl+{{ isMac ? 'Option' : 'Shift'}}+Enter</span>
```
Leemoonsoo: In Mac, Shift + Enter works but Option + Enter doesn't.

tinkoff-dwh:
@Leemoonsoo 
You're right. The description was not true. 
The description has been changed.

Leemoonsoo:
Thanks @tinkoff-dwh. LGTM and merge to master if no further discussions.

PR #2691, opened 2017-11-28
[FIX] fix autocomplete

tinkoff-dwh:
### What is this PR for?
After refactoring of Interpreter autocomplete (from server side) not works without first Run of interpreter.
This PR fix it.

### What type of PR is it?
[Fix]

* Create new Note (JDBC interpreter), try to use autocomplete (schema, tables)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -191,7 +191,7 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
         this.scriptText = this.text.substring(headingSpace.length() + intpText.length() + 1).trim();
       } else {
         this.intpText = "";
-        this.scriptText = this.text;
+        this.scriptText = this.text.trim();
       }
     }
   }
@@ -252,10 +252,9 @@ public class Paragraph extends Job implements Cloneable, JsonSerializable {
     }
     this.interpreter = getBindedInterpreter();
 
-    String trimmedBuffer = buffer != null ? buffer.trim() : null;
-    cursor = calculateCursorPosition(buffer, trimmedBuffer, cursor);
+    setText(buffer);
 
-    setText(trimmedBuffer);
+    cursor = calculateCursorPosition(buffer, cursor);
 
     InterpreterContext interpreterContext = getInterpreterContextWithoutRunner(null);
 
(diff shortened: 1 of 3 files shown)


tinkoff-dwh:
@zjffdu 
pls help with review

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -250,14 +250,18 @@ public void setInterpreter(Interpreter interpreter) {
         return note.getInterpreterCompletion();
       }
     }
+    this.interpreter = getBindedInterpreter();
+
     String trimmedBuffer = buffer != null ? buffer.trim() : null;
     cursor = calculateCursorPosition(buffer, trimmedBuffer, cursor);
 
+    setText(trimmedBuffer);
```
felixcheung: should we change the text here?

tinkoff-dwh: Yes, otherwise the cursor position may be incorrect in line 264 
```return this.interpreter.completion(this.scriptText, cursor, interpreterContext);```

Comment (zjffdu) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: But why we trim the text, it is possible to have blank text in the front of paragraph. Changing text in the server side may not be a good user experience. Maybe there's better solution for cursor position.

Comment (tinkoff-dwh) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: trim was added to fix autocomplete. For example autocomplete works incorrect (JDBC may contains name od datasource %jdbc (mysql)   ...)

Comment (zjffdu) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: It seems jdbc specific issue, is it possible to fix it in jdbc interpreter instead of `Paragraph.java`.

Comment (tinkoff-dwh) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: @zjffdu 
but you trim text in method setText()

```
this.scriptText = this.text.substring(headingSpace.length() + intpText.length() + 1).trim();
```

Comment (zjffdu) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: @tinkoff-dwh Could you give one example that wrong cursor position is detected. I suspect it can be fixed in method `calculateCursorPosition`

Comment (tinkoff-dwh) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: @zjffdu 
https://github.com/apache/zeppelin/blob/afd2bca46c778e9e84deb7ba5e4f5b258735022b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTest.java#L272

but what is the problem if the text in any case is trimmed? (in method setText)

Comment (zjffdu) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java]: I suspect it would change the code of user write in frontend.

e.g.

```
jdbc

select * from table
```

may become to 
```
jdbc
select * from table
```

Changing user code is not user friendly experience IMO

tinkoff-dwh:
@zjffdu @felixcheung 
fixed

zjffdu:
Sorry for late response @tinkoff-dwh , LGTM, will merge it if no more comments

PR #2692, opened 2017-11-30
ZEPPELIN-3085 Introduce generic ConfInterpreter for more fine-grained control of interpreter setting

zjffdu:
### What is this PR for?
Zeppelin's interpreter setting is shared by all the users and notes, if you want to have different setting you have to create new interpreter, e.g. you can create `spark_jar1` for running spark with dependency jar1 and `spark_jar2` for running spark with dependency jar2.
This approach works, but not so convenient. `ConfInterpreter` can provide more fine-grained control on interpreter setting and more flexibility. 

`ConfInterpreter` is a generic interpreter that could be used by any interpreters. The input format should be property file format.

In the first paragraph, we use ConfInterpreter to make custom configuration of spark interpreter (set app name, yarn-client mode & add spark-csv dependencies). Then you can run the second paragraph which use spark-csv.

conf_interpreter

It can be used to make custom setting for any interpreter. `ConfInterpreter` would run in the zeppelin server side, it would update the interpreter properties before you launch the interpreter process, so it needs to run before interpreter process launched. And when interpreter process is launched is determined by interpreter mode setting. So users needs to understand the interpreter mode setting of zeppelin and be aware when interpreter process is launched. E.g. If we set spark interpreter setting as isolated per note. Under this setting, each note will launch one interpreter process. In this scenario, user need to put `ConfInterpreter` as the first paragraph as the above example. Otherwise the customized setting can not be applied (The paragraph using ConfInterpreter will report ERROR).

### What type of PR is it?
[Feature | Documentation]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3085

* Unit test, System test is added, also manually verified it. 

conf_interpreter

### Questions:


zjffdu:
@Leemoonsoo @jongyoul @felixcheung @prabhjyotsingh Could you help review it ? Thanks

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -716,6 +719,11 @@ private String getInterpreterClassFromInterpreterSetting(String replName) {
         return info.getClassName();
       }
     }
+    //TODO(zjffdu) It requires user can not create interpreter with name `conf`,
+    // conf is a reserved word of interpreter name
+    if (replName.equals("conf")) {
```
felixcheung: why do we have to hardcode this?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -728,6 +736,29 @@ private ManagedInterpreterGroup createInterpreterGroup(String groupId) {
...
+   */
+  public void setInterpreterGroupProperties(String interpreterGroupId, Properties properties)
+      throws IOException {
+    ManagedInterpreterGroup interpreterGroup = this.interpreterGroups.get(interpreterGroupId);
+    for (List<Interpreter> session : interpreterGroup.sessions.values()) {
+      for (Interpreter intp : session) {
+        if (!intp.getProperties().equals(properties) &&
+            interpreterGroup.getRemoteInterpreterProcess() != null &&
+            interpreterGroup.getRemoteInterpreterProcess().isRunning()) {
+          throw new IOException("Can not change interpreter properties when interpreter process " +
```
felixcheung: I'm a bit worry about this? I think this is the way why `%dep` is so hard to use - the implicit order of interpreter is not very visible to an user, especially with different interpreter isolation mode etc.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
zjffdu: I make `conf` as a reserved interpreter name for this generic `ConfInterpreter` so that all the interpreters can use it. e.g. `%md.conf`, `%sh.conf`, `%spark.conf`

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
zjffdu: Right, this is what I mention in the documentation. This generic `ConfInterpreter` is for advanced users that they needs to understand the interpreter mode setting of zeppelin and be aware when interpreter process is launched.

zjffdu:
> can I have multiple ConfInterpreter at the same time?

I don't quite get it. For now each interpreter process have one `ConfInterpreter` which is used for customize this interpreter process setting.

felixcheung:
right, that's what I'm wondering

zjffdu:
@Leemoonsoo @jongyoul @felixcheung @prabhjyotsingh any more comments, I will merge it if no more comments

PR #2708, opened 2017-12-18
[Zeppelin-3095] Fix UI when all paragraphs executing sequentially

tinkoff-dwh:
### What is this PR for?
This PR fixes the incorrect behavior of UI during the execution of the RunAllParagraphs. 
Including: allows you to copy the note at runtime (before that, the new note was created only after the end of runAllParagraph), blocks some interface elements.
The following items are blocked:

**Paragraphs Control :**
* Move up
* Move down
* Insert new
* Clone paragraph
* Run all above
* Run all below
* Disable run
* Remove

**Note :**
* Insert new paragraph

### What type of PR is it?
Improvement

### What is the Jira issue?
[Zeppelin-3095](https://issues.apache.org/jira/browse/ZEPPELIN-3095)

Run the paragraphs by clicking the "run all paragraphs" button
Ensures that the required items are locked

### Questions:


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1666,30 +1688,57 @@ private void runAllParagraphs(NotebookSocket conn, HashSet<String> userAndRoles,
...
+              paragraphId, text, title, params, config);
+
+          try {
+            if (!persistAndExecuteSingleParagraph(conn, note, p, true)) {
+              // stop execution when one paragraph fails.
+              runAllStatusBroadcast(note, false);
+              break;
+            }
+          } catch (Exception e) {
+            LOG.error("Error in runAllParagraphs(). Error: " + e.toString());
```
felixcheung: `LOG.error("Error in runAllParagraphs()", e);`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
```diff
@@ -185,7 +185,8 @@
     RUN_PARAGRAPH_USING_SPELL,    // [s-c] run paragraph using spell
     PARAS_INFO,                   // [s-c] paragraph runtime infos
     SAVE_NOTE_FORMS,              // save note forms
-    REMOVE_NOTE_FORMS             // remove note forms
+    REMOVE_NOTE_FORMS,            // remove note forms
+    SEQEUNTIAL_RUN_STATUS         // [s-c] seqeuntial run status will be change
```
felixcheung: `seqeuntial`-> sequential?

tinkoff-dwh:
@felixcheung 
I fixed it

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -836,13 +837,35 @@ private void sendNote(NotebookSocket conn, HashSet<String> userAndRoles, Noteboo
...
       }
+      checkSequentialRunStatus(note);
       conn.send(serializeMessage(new Message(OP.NOTE).put("note", note)));
       sendAllAngularObjects(note, user, conn);
     } else {
       conn.send(serializeMessage(new Message(OP.NOTE).put("note", null)));
     }
   }
 
+  //check sequential running status (if NoteServer crash last time, it recover really status)
```
felixcheung: `really` - do you mean `ready`?

zeppelin-web/src/app/notebook/notebook.html
```diff
@@ -161,7 +162,9 @@ <h4 class="plus-sign">&#43; Add Paragraph</h4>
          ng-hide="currentParagraph.config.tableHide && viewOnly"
          ng-dblclick="paragraphOnDoubleClick(currentParagraph.id);">
     </div>
-    <div class="new-paragraph last-paragraph" ng-click="insertNew('below');" ng-hide="!$last || viewOnly || asIframe || revisionView">
+    <div class="new-paragraph last-paragraph" ng-click="insertNew('below');"
+         ng-if="!isSequentialRun()"
```
felixcheung: shouldn't this be using ng-class?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
tinkoff-dwh: If the server crashes during RunAllParagraphs execution, then after switching on, the note will be in the same status. Here this case is just being processed.

zeppelin-web/src/app/notebook/notebook.html
tinkoff-dwh: Having looked again at the code, I decided that it would be better to transfer the verification to the ng-hide. Do you think this is the best solution?

felixcheung:
yes, I think

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -836,13 +837,35 @@ private void sendNote(NotebookSocket conn, HashSet<String> userAndRoles, Noteboo
...
+  private void checkSequentialRunStatus(Note note) {
```
zjffdu: This logics seems wrong. It is possible to be in sequential run mode but all paragraphs are in PENDING state.  There's time gap between PENDING and RUNNING.

zjffdu:
@tinkoff-dwh I am not frontend expert. Just going through the PR and have a few thoughts:

1. The changes in frontend seems a little large. What if we add new UI control ? Do we also need to check the sequential running state ?
2. I have another idea for this problem. How about doing the most of the work in backend ? Backend know whether it is in sequential running mode. So that backend can send back error message to frontend if user do a ui operation on a note which is in sequential running state. e.g. when user click the clone button, zeppelin can pop up a error message to show that this operation is not allowed as it is in sequential running state. This might be much easier to do. But I might be biased. Need more thoughts from some frontend guys.

tinkoff-dwh:
@zjffdu 
i think it is not user friendly interface (popups from backend)

zjffdu:
@tinkoff-dwh I am not saying popup from backend. I mean backend send error message to frontend, and frontend pop up this error message.

tinkoff-dwh:
@zjffdu 
I meant that the popup is not user friendly. The user sees a clickable elements, clicks and the popup will be displayed

tinkoff-dwh:
@prabhjyotsingh @r-kamath 
could you help with the review?

prabhjyotsingh:
Tested it locally, and front-end changes looks OK.
In this case, I agree popup is not user-friendly. Also since we are blocking (or temporarily disabling) Paragraphs Control, should we block editing of paragraphs as well, since from backend it is also blocked?

tinkoff-dwh:
@prabhjyotsingh 
Add block editing of paragraphs.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
Savalek: @felixcheung, I mean this function is to check the true running status of the paragraph.

jongyoul:
Can someone share the status of this issue?

mebelousov:
@jongyoul status is ready to merge to master =)

I hope I will merge this PR to our fork of branch-0.8 and will confirm that works fine.
This needs 1-2 months.

jongyoul:
@mebelousov Got it. Thanks

tinkoff-dwh:
Current version : [pull#3226](https://github.com/apache/zeppelin/pull/3226)

PR #2712, opened 2017-12-22
ZEPPELIN-2742. Unify Zeppelin's Config Storage

zjffdu:
### What is this PR for?

This PR is for unifying the config storage of zeppelin. Currently we store it as local file, this PR introduce an interface `ConfigStorage` and one implementation based on hadoop filesystem api. User could add others in future.  

### What type of PR is it?
[ Improvement  | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2742

Unit test is added and manually tested

### Questions:


felixcheung:
shouldn't this PR say "config storage" instead of "storage layer"?

zjffdu:
make sense, PR and jira title changed

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -518,7 +522,17 @@ public boolean isNotebookPublic() {
   }
 
   public String getConfDir() {
-    return getRelativeDir(ConfVars.ZEPPELIN_CONF_DIR);
+    return getRelativeDir(getString(ConfVars.ZEPPELIN_CONF_DIR));
```
felixcheung: why this change? other places are calling `getRelativeDir(ConfVars.something)`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -518,7 +522,17 @@ public boolean isNotebookPublic() {
...
+  }
+
+  public String getConfigFSDir() {
+    String fsConfigDir = getString(ConfVars.ZEPPELIN_CONFIG_FS_DIR);
+    if (StringUtils.isBlank(fsConfigDir)) {
+      LOG.warn(ConfVars.ZEPPELIN_CONFIG_FS_DIR.varName + " is not specified, fall back to local " +
+          "conf directory " + ConfVars.ZEPPELIN_CONF_DIR.varName);
+      return "file://" + getConfDir();
```
felixcheung: shouldn't this fallback to `getString(ConfVars.ZEPPELIN_HOME) + "/" + path`? This was the behavior before this change

zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
```diff
@@ -19,15 +19,14 @@
...
-
+import com.google.gson.Gson;
 import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.common.JsonSerializable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.gson.Gson;
+import java.util.ArrayList;
+import java.util.List;
```
felixcheung: let's revert this file?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -149,7 +161,6 @@ public Void call() throws IOException {
 
   public synchronized <T> T callHdfsOperation(final HdfsOperation<T> func) throws IOException {
     if (isSecurityEnabled) {
-      UserGroupInformation.getLoginUser().reloginFromKeytab();
```
felixcheung: why this change?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
```diff
@@ -133,27 +115,14 @@ public void setRoles(String user, Set<String> roles) {
...
-    }
-
-    try {
-      File settingFile = new File(filePath);
-      if (!settingFile.exists()) {
-        settingFile.createNewFile();
+      try {
+        configStorage.save(info);
+      } catch (IOException e) {
+        LOG.error("Error saving notebook authorization file: " + e.getMessage());
```
felixcheung: LOG.error("Error saving notebook authorization file", e);

zeppelin-zengine/src/test/java/org/apache/zeppelin/user/EncryptorTest.java
```diff
@@ -17,10 +17,10 @@
 
 package org.apache.zeppelin.user;
 
-import java.io.IOException;
-
 import org.junit.Test;
 
+import java.io.IOException;
+
```
felixcheung: hmm, lets revert?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
zjffdu: Hadoop will treat it as hdfs file if no prefix with `file://`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
zjffdu: It is not necessary to do relogin because hadoop rpc will do it.

zjffdu:
Merge if no more comments

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NotebookAuthorization.java
```diff
@@ -133,27 +115,14 @@ public void setRoles(String user, Set<String> roles) {
...
-    }
-
-    try {
-      File settingFile = new File(filePath);
-      if (!settingFile.exists()) {
-        settingFile.createNewFile();
+      try {
+        configStorage.save(info);
+      } catch (IOException e) {
+        LOG.error("Error saving notebook authorization file", e);
```
steveloughran: recommend adding `filePath` to that error string

zjffdu: The storage file path will be logged in `FileSystemConfigStorage`. And it doesn't expose api for getting the config file path. Actual `ConfigStorage` is a generic interface, it doesn't need to store into a file, the storage could be other storage like DB and etc.

steveloughran: OK. I'm just worrying about fielding the support calls where all you get is that stack trace.

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/FileSystemConfigStorage.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private FileSystemStorage fs;
+  private Path interpreterSettingPath;
+  private Path authorizationPath;
+  private Path credentialPath;
+
+  public FileSystemConfigStorage(ZeppelinConfiguration zConf) throws IOException {
+    super(zConf);
+    this.fs = FileSystemStorage.get(zConf);
+    this.fs.tryMkDir(new Path(zConf.getConfigFSDir()));
```
karth295: This is incorrect. `FileSystemStorage.get(zconf)` gives you an FS for zconf.getNotebookDir(). The notebook directory is *not* guaranteed to be on the same FS. Context: I tried to use the local `conf/` directory for zeppelin configuration, but put the notebook directory on a `gs://` path so that I could persistently save the notebook.

The same applies to interpreterSettingPath, authorizationPath, credentialPath.

FWIW, I think it's a mistake for FileSystemStorage/FileSystemNotebookRepo to use zconf.getNotebookDir(). Instead, there should be a separate config like zconf.getHcfsNotebookDir(). That way, each storage implementation has its own independent configuration. Right now, you cannot use FileSystemNotebookRepo with GitNotebookRepo, because they rely on the same configuration.

I'm happy to file JIRAs for these if you would like.

PR #2714, opened 2017-12-29
[ZEPPELIN-3121] Fix to create log per each interpreter

woowahan-jaehoon:
### What is this PR for?
When interpreter policy is not set globally, zeppelin run many interpreters but log file is still single.
This PR is for fix this.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix

### What is the Jira issue?
[ZEPPELIN-3121]

After running many interpreters, many log files could be created per each interpreter with same condition.

### Questions:
    * No
    * No
    * No

https://issues.apache.org/jira/browse/ZEPPELIN-3121


bin/interpreter.cmd
```diff
@@ -57,7 +57,7 @@ call "%bin%\functions.cmd" ADDJARINDIR "%INTERPRETER_DIR%"
 set HOSTNAME=%COMPUTERNAME%
 set ZEPPELIN_SERVER=org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer
 
-set ZEPPELIN_LOGFILE=%ZEPPELIN_LOG_DIR%\zeppelin-interpreter-%INTERPRETER_ID%-%ZEPPELIN_IDENT_STRING%-%HOSTNAME%.log
+set ZEPPELIN_LOGFILE=%ZEPPELIN_LOG_DIR%\zeppelin-interpreter-%INTERPRETER_ID%-%ZEPPELIN_IDENT_STRING%-%HOSTNAME%-%PORT%.log
```
zjffdu: It doesn't make sense to append port to log file name. Because port is unknown to users. Facing so many log files, it is hard for user to figure out which log file is for his interpreter log.

woowahan-jaehoon: Yes. It is hard to figure out but I think that it is better than lose many logs.

Let's think interpreter policy is set 'per user' and 'isolated' in interpreter option.
When 50 users use zeppelin, zeppelin will run 50 interpreters. 
But `interpreter.sh` will run with same log arguments and same log files.
And first-ran interpreter open and lock log file.
So, another 49 interpreters' log will be lose.

I found this situation when I needed to get statistic user's usage.
I got just one interpreter's log.

zjffdu: There's 2 solutions in my mind:
1. Append username+noteId to log file. 
2. Still writing to a single log file, but prepend username + noteId for each line of log.

woowahan-jaehoon: `interpreter.*` cannot get noteid and `interpreter.cmd` also cannot get user, now.

And I saw that `-u` option is not used default zeppelin setting.

In my case, zeppelin run interpreter with like this command
```sh
/bin/bash /zeppelin/path/bin/interpreter.sh -d /zeppelin/path/interpreter/spark -p 53959 -l /zeppelin/path/local-repo/2C4QDS3Q7

/bin/bash /zeppelin/path/bin/interpreter.sh -d /zeppelin/path/interpreter/md -p 44895 -l /zeppelin/path/local-repo/2C5M6QK8U
```
(and `/zeppelin/path/local-repo/2C4QDS3Q7` is not note path)

So, do you mean calling java logic also have to be changed?
But in this case, I think that interpreter policy have to be considered because 'per note' and 'per user' can be different and 'isolated' and 'scoped' also can be different.

woowahan-jaehoon:
I seperated issue of pid and log because log has some issue.

PR #2717, opened 2018-01-05
[ZEPPELIN-3127] Upgrade selenium version

prabhjyotsingh:
### What is this PR for?
We are currently using a very old selenium version (i.e. 2.48.2) this is to upgrade it to 3.8.1

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3127](https://issues.apache.org/jira/browse/ZEPPELIN-3127)

* CI should be green

### Questions:
* Does the licenses files need update? yes


.travis.yml
```diff
@@ -69,9 +69,9 @@ matrix:
       env: PYTHON="3" SCALA_VER="2.11" SPARK_VER="2.2.0" HADOOP_VER="2.6" PROFILE="-Pspark-2.2 -Pweb-ci -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="package -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/ZeppelinSparkClusterTest.java,**/org.apache.zeppelin.spark.*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
 
     # Test selenium with spark module for 1.6.3
-    - jdk: "oraclejdk7"
+    - jdk: "oraclejdk8"
```
felixcheung: my concern is spark 1.6.3 doesn't officially support jdk 8
https://spark.apache.org/docs/1.6.3/#downloading

zeppelin-web/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -146,37 +148,46 @@ public Boolean apply(WebDriver d) {
...
-    LOG.info("firebug version: " + firefoxVersion + ", will be downloaded to " + tempPath);
+    LOG.info("Geeko version: " + firefoxVersion + ", will be downloaded to " + tempPath);
     try {
-      File firebugFile = new File(tempPath + "firebug.xpi");
-      URL firebugUrl = new URL(firebugUrlString);
-      if (!firebugFile.exists()) {
-        FileUtils.copyURLToFile(firebugUrl, firebugFile);
+      if (SystemUtils.IS_OS_WINDOWS) {
+        if (System.getProperty("sun.arch.data.model").equals("64")) {
+          geekoDriverUrlString = "https://github.com/mozilla/geckodriver/releases/download/v0.19.1/geckodriver-v0.19.1-win64.zip";
```
felixcheung: is there a way to not hardcode version v0.19.1?

zeppelin-web/src/test/java/org/apache/zeppelin/ProcessData.java
```diff
@@ -0,0 +1,249 @@
...
+              (System.currentTimeMillis() > unconditionalExitTime));
+          this.checked_process.destroy();
+          try {
+            if ((System.currentTimeMillis() > unconditionalExitTime))
+              LOG.error("!@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@Unconditional exit occured@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@!\nsome process hag up for more than " + unconditionalExitDelayMinutes + " minutes.");
+            LOG.error("!##################################!");
+            StringWriter sw = new StringWriter();
+            new Exception("Exited from buildOutputAndErrorStreamData by timeout").printStackTrace(new PrintWriter(sw)); //Get stack trace
+            String exceptionAsString = sw.toString();
+            LOG.error(exceptionAsString);
```
felixcheung:
```
Exception e = new Exception("Exited from buildOutputAndErrorStreamData by timeout").printStackTrace(new PrintWriter(sw)); //Get stack trace
LOG.error(e);
```
?

zeppelin-web/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -146,37 +151,48 @@ public Boolean apply(WebDriver d) {
...
-    if (firefoxVersion < 23)
-      firebugUrlString = "http://getfirebug.com/releases/firebug/1.11/firebug-1.11.4.xpi";
-    else if (firefoxVersion >= 23 && firefoxVersion < 30)
-      firebugUrlString = "http://getfirebug.com/releases/firebug/1.12/firebug-1.12.8.xpi";
-    else if (firefoxVersion >= 30 && firefoxVersion < 33)
-      firebugUrlString = "http://getfirebug.com/releases/firebug/2.0/firebug-2.0.7.xpi";
-    else if (firefoxVersion >= 33)
-      firebugUrlString = "http://getfirebug.com/releases/firebug/2.0/firebug-2.0.17.xpi";
+  public static void downloadGeekoDriver(int firefoxVersion, String tempPath) {
+    String geekoDriverUrlString =
```
prabhjyotsingh: @felixcheung  is this OK? I've seen something like https://github.com/Ardesco/selenium-standalone-server-plugin/blob/master/src/main/resources/RepositoryMap.xml but even that looks bit ugly.

prabhjyotsingh:
Selenium 3.x requires JDK8, guava-23, google.gson-2.8.2, which was conflicting with existing zeppelin-server package, hence have moved if to selenium testing to a new package called "zeppelin-integration".
Also in the Travis matrix have change dist to trusty (which might be required for ZEPPELIN-3118) and set JDK as oraclejdk8.

cc: @felixcheung , @jongyoul , @Leemoonsoo , @zjffdu

jongyoul:
I agree with moving integration tests to new module. I have a question. AFAIK, modern web browsers already had web driver for selenium tests. Can you tell me why you add gecko driver and Firefox 31? You can upgrade the version of browsers up-to-date. How do you think of it?

prabhjyotsingh:
I had to use gecko because "Firefox is only fully supported at version 47.0.1 or earlier. Support
  for later versions of firefox is provided by geckodriver" 
Ref: https://github.com/SeleniumHQ/selenium/blob/master/java/CHANGELOG

Yes, I plan to upgrade version of Firefox as well, I was just waiting for CI to pass as it used to pass on dist=precise (which uses FireFox 31)

zeppelin-integration/src/test/resources/zeppelin-site-star.xml
```diff
@@ -0,0 +1,157 @@
+<?xml version="1.0"?>
```
zjffdu: where is this file used ?

zeppelin-integration/src/test/scala/org/apache/zeppelin/AbstractFunctionalSuite.scala
```diff
@@ -17,14 +17,16 @@
 
 package org.apache.zeppelin
 
+import org.apache.commons.io.FileUtils
```
zjffdu: Not sure why we use scala for this integration test, because I see other selenium test use java.  Using scala make zeppelin-integration need to depend on scala which seems a little unnecessary. Not a big deal, but it would better to have all selenium test written in java code.

zeppelin-integration/src/test/resources/zeppelin-site-star.xml
prabhjyotsingh: Copied from `zeppelin-server`, removed both unused "zeppelin-site-star.xml" and "zeppelin-site.xml".

zeppelin-integration/src/test/scala/org/apache/zeppelin/AbstractFunctionalSuite.scala
prabhjyotsingh: Got this from `zeppelin-server`, but yes agreed this looks redundant, removed it.

prabhjyotsingh:
Tried FireFox-56, it works, is failing with some of the flaky tests (https://api.travis-ci.org/v3/job/326228047/log.txt, https://api.travis-ci.org/v3/job/326219155/log.txt)

Along with that build (https://travis-ci.org/prabhjyotsingh/zeppelin/builds/3262280430) is also failing for profile 6-10 with:
```
  ZeppelinSparkClusterTest.sparkRTest:208 expected:<FINISHED> but was:<ERROR>
```
which is not related to this.

.travis.yml
```diff
@@ -69,9 +69,11 @@ matrix:
...
 
     # Test selenium with spark module for 1.6.3
-    - jdk: "oraclejdk7"
-      dist: precise
-      env: PYTHON="2" TEST_SELENIUM="true" SCALA_VER="2.10" SPARK_VER="1.6.3" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pspark-1.6 -Phadoop-2.6 -Phelium-dev -Pexamples" BUILD_FLAG="package -DskipTests -DskipRat" TEST_FLAG="verify -DskipRat" TEST_PROJECTS="-pl .,zeppelin-interpreter,zeppelin-zengine,zeppelin-server,zeppelin-display,spark-dependencies,spark,python -Dtest=org.apache.zeppelin.AbstractFunctionalSuite -DfailIfNoTests=false"
+    - jdk: "oraclejdk8"
+      dist: trusty
+      addons:
+        firefox: "31.0"
+      env: PYTHON="2" SCALA_VER="2.10" SPARK_VER="1.6.3" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pspark-1.6 -Phadoop-2.6 -Phelium-dev -Pexamples -Pintegration" BUILD_FLAG="package -DskipTests -DskipRat" TEST_FLAG="verify -DskipRat" TEST_PROJECTS="-pl .,zeppelin-integration -DfailIfNoTests=false"
```
felixcheung: why is `TEST_SELENIUM="true" ` removed?

prabhjyotsingh: Yes, have moved all selenium tests to a new module "zeppelin-integration" which will now run only when maven is run with "-Pintegration". Hence, with this, I thought TEST_SELENIUM is redundant.

.travis.yml
prabhjyotsingh: Should I change both Scala and Spark to `SCALA_VER="2.11" SPARK_VER="2.2.0"`? As selenium-3.x requires JDK 8. However, I see travis CI (https://travis-ci.org/prabhjyotsingh/zeppelin/jobs/326501440/config) is passing.

zjffdu: You can keep JDK7 here. But only use JDK8 when running selenium test.

prabhjyotsingh:
Travis job (https://travis-ci.org/prabhjyotsingh/zeppelin/jobs/326663402) for selenium passed, it fails for 608.[6-10] which is not related to this.

Will add support for newer Firefox versions and fix flaky test failures in a different PR.

prabhjyotsingh:

PR #2721, opened 2018-01-09
[ZEPPELIN-3129] Zepplin UI Doesnt logs out in IE

prabhjyotsingh:
### What is this PR for?
Zepplin UI Doesn't logs out in IE/Edge, its Shows Logout Success and automatically reload the page 
But the User is still logged in. 

This happens when shiro is configured with "authc" however it works for "authcBasic".

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3129](https://issues.apache.org/jira/browse/ZEPPELIN-3129)

* Logout should work for all browsers and for both shiro configuration i.e. "authc" and "authcBasic"

### Questions:

### DIFF
--- a/zeppelin-web/src/components/navbar/navbar.controller.js
+++ b/zeppelin-web/src/components/navbar/navbar.controller.js
@@ -89,9 +89,6 @@ function NavCtrl ($scope, $rootScope, $http, $routeParams, $location,
   function logout() {
     let logoutURL = baseUrlSrv.getRestApiBase() + '/login/logout'
 
-    // for firefox and safari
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
Have tested this for FF31, FF57, Chrome-63, Safari-11, Windows7's IE-12 and Windows10's Edge-25

Akhilsnaik:
Have Tested in Windows 7 IE 11 , 
The Fix works fine in IE 11.

prabhjyotsingh:

PR #2729, opened 2018-01-12
[ZEPPELIN-2826] support for carriage return '\r', on result window

prabhjyotsingh:
### What is this PR for?
This PR implements carriage return '\r' on result window, currently since HTML treats "\n", "\r" and "\r\n"  equally.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-2826](https://issues.apache.org/jira/browse/ZEPPELIN-2826)

Here is a sample code to test it
```
%python
import time,sys
end_val = 10
bar_length = 20
for i in xrange(0, end_val + 1):
    time.sleep(0.5)
    percent = float(i) / end_val
    hashes = '#' * int(round(percent * bar_length))
    spaces = ' ' * (bar_length - len(hashes))
    sys.stdout.write("\rPercent: [{0}] {1}%".format(hashes + spaces, int(round(percent * 100))))
    #print "Percent: [{0}] {1}%".format(hashes + spaces, int(round(percent * 100)))
```

after2

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOutput.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOutput.java
@@ -20,14 +20,15 @@ package org.apache.zeppelin.interpreter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.*;
-import java.net.URISyntaxException;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * InterpreterOutput is OutputStream that supposed to print content on notebook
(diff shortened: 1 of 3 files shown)


zjffdu:
Thanks @prabhjyotsingh, it is really useful. I tried it, it works very well. Just one minor issue, the output size might needs to be adjusted, other it is very easy to get the following issue.
screen shot 2018-01-13 at 9 19 19 am

prabhjyotsingh:
I think that can be adjusted via `zeppelin.interpreter.output.limit`? Or help me understand if you meant something else.

zjffdu:
What I mean is that we need the ajust how to calculate the `size` in `InterpreterOutput`. Because as you can see in the above screenshot, the output size in front end is not so large, but still get the problem of exceeding `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT`, I believe this is because in `InterpreterOutput` we would always increase the `size` no matter whether it contains `\r`.

prabhjyotsingh:
I now see what you were saying earlier, let me check if there is a quick fix for this.

prabhjyotsingh:
@zjffdu  That "calculation" looks like a bigger change and might change the content of notebook as well.

zjffdu:
@prabhjyotsingh Could you check this commit ? https://github.com/zjffdu/zeppelin/commit/00bce395f3ed557eb822095f490dae5f8bfd5a23

I tested it locally, and works well.

prabhjyotsingh:
That works well, have merged that commit to this PR.

zjffdu:
Thanks @prabhjyotsingh  The PR LGTM overall

prabhjyotsingh:

PR #2730, opened 2018-01-15
ZEPPELIN-3166. R plotting resolution and image width is not proper

zjffdu:
### What is this PR for?
Change the resolution and image width, because the previous default value is not proper.

By adding `fig.retina = 2` for high resolution. And use `40%` as default image width. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3166

* Manually verfifed

Before

screen shot 2018-01-15 at 2 27 01 pm

After 

screen shot 2018-01-15 at 2 27 11 pm

### Questions:

### DIFF
--- a/spark/src/main/resources/interpreter-setting.json
+++ b/spark/src/main/resources/interpreter-setting.json
@@ -213,7 +213,7 @@
       "zeppelin.R.render.options": {
         "envName": "ZEPPELIN_R_RENDER_OPTIONS",
         "propertyName": "zeppelin.R.render.options",
-        "defaultValue": "out.format = 'html', comment = NA, echo = FALSE, results = 'asis', message = F, warning = F",
+        "defaultValue": "out.format = 'html', comment = NA, echo = FALSE, results = 'asis', message = F, warning = F, fig.retina = 2",
         "description": "",
         "type": "textarea"
       }


zjffdu:
@felixcheung @echarles Could you help review it ? Thanks

echarles:
`fig.retina` sounds like a good property to play with. However, fixing `out.width` to an arbitrary value may give good result for the example you give, but did you try with other plots?

Showing nice-looking plots generated by knitr depends on your screen resolution and browser actual window width.

We had before (I hope it is still there) the ability to force the image width passing a json after the magic keyword

```
%spark.r {"imageWidth": "400px"}
```

If you use that `imageWidth` property and push a bit more `fig.retina` to 3 or 4, could it be that most cases will be covered?

zjffdu:
Oops, never realize there's such magic, seems no document mention that. But we should not rely on these kinds of magic for most of users. We should provide a default value for most of users that no need of any extra overhead no matter it is configuration or such kind of magic.  Thoughts ?

echarles:
Agree, but the default will always find case where it will not work. For this image size / resolution, there may be more failure than success I am afraid. To be tested with various plots, screen resolutions and paragraphs width to measure this.

zjffdu:
@echarles I am fine to keep image width as 100%. But setting `fig.retina=2` is necessary, otherwise the image displayed is so unclear.  See below comparison (use both ggplot2 and R base plotting)

Before

screen shot 2018-01-16 at 11 28 33 am

After

screen shot 2018-01-16 at 11 28 39 am

echarles:
GTM

zjffdu:
Thanks @echarles  will merge if no more comments

PR #2733, opened 2018-01-17
Set ExecutorFactory 65536

mebelousov:
### What is this PR for?
The number at ExecutorFactory limits the maximum number of interpreters. When the number of interpreters becomes equal to the schedulers number, zeppelin will be fully stuck: no paragraph can run now even if the paragraph run a few minutes ago.
This HF makes the number of schedulers very big.

### What type of PR is it?
[Hot Fix]

### Todos
* [ ] - Task

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/SchedulerFactory.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/SchedulerFactory.java
@@ -52,7 +52,7 @@ public class SchedulerFactory implements SchedulerListener {
   }
 
   SchedulerFactory() throws Exception {
-    executor = ExecutorFactory.singleton().createOrGet("SchedulerFactory", 100);
+    executor = ExecutorFactory.singleton().createOrGet("SchedulerFactory", 65536);
   }
 
   public void destroy() {


zjffdu:
I think it is not `limits the maximum number of interpreters`, it is `limits the maximum number of concurrently running paragraphs`. I have concern that 65536 would be too large for single process to schedule and coordinate so many threads. @mebelousov Do you see this issue in production ? Because concurrently running 100 paragraphs would be a very busy zeppelin server with many users.

mebelousov:
@zjffdu, yes, I've got this error in production.
Schedulers aren't shared, there is one scheduler per each interpreter processes. If user starts new interpreter processes, Zeppelin creates new scheduler object.
ExecutorFactory limits the number of schedulers and therefore the number of java processes.

In description I mean that interpreter is java process.
How correctly to say about started interpreter process/JVM?

I wrote about this bug in mailing lists: [1](https://lists.apache.org/thread.html/30966875b50b5ac8b4326c23a012d702ea7cac24d75f540ae58f31b3@%3Cusers.zeppelin.apache.org%3E), [2](https://lists.apache.org/thread.html/a5e4de950d74a858e249af9e07fb17f63c94a136661ca78fc0c26c90@%3Cdev.zeppelin.apache.org%3E)

@felixcheung
Our team uses interpreters in isolated mode per user/per note.
Every night we restart Zeppelin.
The analysts run notes and zeppelin starts more than a 100 JVMs in the evening.

zjffdu:
@mebelousov I manually change it to 2, but I could still create 3 interpreter processes. So I suspect the issue you hit is due to other bugs.

zjffdu:
Thanks @mebelousov for verifying

PR #2734, opened 2018-01-19
ZEPPELIN-3174. ZeppelinSparkClusterTest fails when using hadoop3 and jdk7

zjffdu:
### What is this PR for?
Trivial change to make travis pass

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3174

* Travis pass

### Questions:


.travis.yml
```diff
@@ -103,15 +103,15 @@ matrix:
...
     # Test spark module for 1.6.3 with scala 2.11
-    - jdk: "openjdk7"
+    - jdk: "oraclejdk8"
       dist: trusty
       env: PYTHON="2" SCALA_VER="2.11" SPARK_VER="1.6.3" HADOOP_VER="2.6" PROFILE="-Pweb-ci -Pspark-1.6 -Phadoop3 -Phadoop-2.6 -Pscala-2.11" SPARKR="true" BUILD_FLAG="install -DskipTests -DskipRat" TEST_FLAG="test -DskipRat" MODULES="-pl .,zeppelin-interpreter,zeppelin-zengine,zeppelin-server,zeppelin-display,spark-dependencies,spark,python" TEST_PROJECTS="-Dtest=ZeppelinSparkClusterTest,org.apache.zeppelin.spark.* -DfailIfNoTests=false"
 
     # Test python/pyspark with python 2, livy 0.2
     - sudo: required
       dist: trusty
       jdk: "openjdk7"
```
masyan: it may be better to use the oraclejdk everywhere...

zjffdu: sounds fair, but better to do in another PR. I don't want to do other things beyond the issue this ticket try to address.

felixcheung: Actually I'd prefer we test with both

PR #2736, opened 2018-01-21
ZEPPELIN-3179. Improve error message when IPython is not available

zjffdu:
### What is this PR for?
Minor change to improve the error message when IPython is not available. See the following screenshot. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3179

* Manually
screen shot 2018-01-20 at 5 00 25 pm
screen shot 2018-01-20 at 5 00 19 pm

 verified

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
@@ -25,9 +25,11 @@ import org.apache.commons.exec.ExecuteWatchdog;
 import org.apache.commons.exec.LogOutputStream;
 import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.exec.environment.EnvironmentUtils;
+import org.apache.commons.httpclient.util.ExceptionUtil;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.zeppelin.interpreter.Interpreter;
 import org.apache.zeppelin.interpreter.InterpreterContext;
 import org.apache.zeppelin.interpreter.InterpreterException;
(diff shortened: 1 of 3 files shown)


zjffdu:
@felixcheung @prabhjyotsingh Could you help review it ? Thanks

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -150,33 +154,27 @@ public boolean checkIPythonPrerequisite() {
...
       if (!freezeOutput.contains("grpcio=")) {
-        InterpreterContext.get().out.write("grpcio is not installed\n".getBytes());
-        return false;
+        return "grpcio is not installed";
       }
       LOGGER.info("IPython prerequisite is meet");
-      return true;
     } catch (Exception e) {
       LOGGER.warn("Fail to checkIPythonPrerequisite", e);
-      return false;
```
felixcheung: if there's an exception we should return a generic error string here?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -150,33 +154,27 @@ public boolean checkIPythonPrerequisite() {
...
       }
       if (!freezeOutput.contains("ipython=")) {
-        InterpreterContext.get().out.write("ipython is not installed\n".getBytes());
-        return false;
+        return "ipython is not installed";
       }
       if (!freezeOutput.contains("grpcio=")) {
-        InterpreterContext.get().out.write("grpcio is not installed\n".getBytes());
-        return false;
+        return "grpcio is not installed";
```
felixcheung: are these return strings going to trace log?

zjffdu: Yes, it would be logged as well.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: Fixed

zjffdu:
More if no more comments

PR #2719, opened 2018-01-06
Spark impersonation

hayssams:
### What is this PR for?
Makes Spark Interperter impersonate front-end user using spark.yarn.principal who becomes the only user to be declared in hadoop core-site.xml (hadoop.proxyuser.USER.hosts)
Added Kerberos auto-renewal flag in zeppelin-site.xml (zeppelin.kerberos.renew.period) / ZEPPELIN_KERBEROS_RENEW_PERIOD env var

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-2066]

### How should this be tested?
* Manually against kerberized HDFS cluster.

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? Yes. To document the kerberos ticket renewal period flag

PR #2737, opened 2018-01-22
ZEPPELIN-3181. Enable impersonation support for kerberized cluster

zjffdu:
### What is this PR for?
This is to enable kerberos support for spark yarn mode in impersonation. Spark has one limitation that you can not specify keytab & proxyuser together. So this PR would run kinit before launching spark interpreter. so that user can enable impersonation for secured cluster.

### What type of PR is it?
[Improvement | Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3181

* Verified manually,  see the following screenshot.

screen shot 2018-01-22 at 9 45 26 am

### Questions:

### DIFF
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -114,6 +114,11 @@ fi
 
 # set spark related env variables
 if [[ "${INTERPRETER_ID}" == "spark" ]]; then
+
+  // # run kinit
+  if [[ -n "${ZEPPELIN_SERVER_KERBEROS_KEYTAB}" ]] && [[ -n "${ZEPPELIN_SERVER_KERBEROS_PRINCIPAL}" ]]; then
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
Does it not requires renew of keytab? Or this takes care of it ?
```
env.put("ZEPPELIN_SERVER_KERBEROS_KEYTAB", keytab);
env.put("ZEPPELIN_SERVER_KERBEROS_PRINCIPAL", principal);
```

zjffdu:
Merge if no more comments

hayssams:
I suspect a problem with this PR if the Kerberos ticket duration is shorter than the interpreter lifetime.
That is why I previously opened a PR that explicitly do a ticket renewal at an interval defined by the user. Please check the following PR https://github.com/apache/zeppelin/pull/2719

PR #2739, opened 2018-01-22
[ZEPPELIN-3183] Add date format support for line chart as x-axis

prabhjyotsingh:
### What is this PR for?
Add date format support for line chart as x-axis

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3183](https://issues.apache.org/jira/browse/ZEPPELIN-3183)

Before
before

After
after

### Questions:


zjffdu:
LGTM，one minor concern: is the millisecond displayed by  default ? IMO, second should be sufficient for most of cases.

prabhjyotsingh:
I think the number what you see after +{number} is the time zone. Eg:
```
Default which is coming from query -> 2018-01-22T14:57:43+0000
When enabled Date -> Mon Jan 22 2018 20:27:43 GMT+0530 (IST)
```

prabhjyotsingh:

PR #2740, opened 2018-01-23
[ZEPPELIN-3185] Add profiles for adding dependencies for Hive and Phoenix

prabhjyotsingh:
### What is this PR for?
This PR adds profiles for adding dependencies for Hive and Phoenix

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3185](https://issues.apache.org/jira/browse/ZEPPELIN-3185)

### Questions:


prabhjyotsingh:
@zjffdu  Please help review this.

jdbc/pom.xml
```diff
@@ -32,6 +32,201 @@
   <packaging>jar</packaging>
   <version>0.8.0-SNAPSHOT</version>
   <name>Zeppelin: JDBC interpreter</name>
+
+  <profiles>
+    <profile>
+      <id>hive</id>
+      <properties>
```
zjffdu: hive jdbc interpreter can support hive1 and hive2 in one interpreter instance ?

jdbc/pom.xml
```diff
@@ -32,6 +32,201 @@
   <packaging>jar</packaging>
   <version>0.8.0-SNAPSHOT</version>
   <name>Zeppelin: JDBC interpreter</name>
+
+  <profiles>
+    <profile>
+      <id>hive</id>
```
zjffdu: better to rename it as jdbc-hive

jdbc/pom.xml
```diff
@@ -32,6 +32,201 @@
...
+        <dependency>
+          <groupId>org.apache.hive.shims</groupId>
+          <artifactId>hive-shims-0.23</artifactId>
+          <version>${hive2.version}</version>
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>phoenix</id>
```
zjffdu: dito. rename it as jdbc-phoenix

jdbc/pom.xml
```diff
@@ -32,6 +32,201 @@
...
+        <dependency>
+          <groupId>xerces</groupId>
+          <artifactId>xercesImpl</artifactId>
+          <version>2.11.0</version>
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>hadoop2</id>
```
zjffdu: there's profile hadoop2 & hadoop3 in zeppelin-zengine as well, so better to rename it as jdbc-hadoop2 and jdbc-hadoop3 here.

jdbc/pom.xml
prabhjyotsingh: Haven't seen any error while using jdbc-hive-1.2.1, guess they are backward compatible.

zjffdu:
LGTM

prabhjyotsingh:

jdbc/pom.xml
```diff
@@ -32,6 +32,201 @@
...
+            </exclusion>
+          </exclusions>
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>jdbc-hadoop3</id>
+      <properties>
+        <hadoop-common.version>3.0.0</hadoop-common.version>
```
felixcheung: shouldn't hadoop2 vs hadoop3 be using the hadoop profile

prabhjyotsingh: I initially thought about it, but then changed, as there can be a case where only Postgres or Hive without Hadoop is required and because hadoop2/3 is specified for zeppelin-zengine it will download Hadoop dependencies for JDBC as well resulting in increased size of JDBC. or do you mean instead of using jdbc-hadoop2/jdbc-hadoop3 there should only be jdbc-hadoop?

felixcheung:
I guess it could make sense if people want to use jdbc alone (and that is Hadoop 2 set of jdbc)

Would it confict if zengine or something is hadoop2 and then the interpreter is hadoop3?

prabhjyotsingh:
I believe `zeppelin-jdbc` depends only on `zeppelin-interpreter`, and not on `zeppelin-zengine`, so it should be safe.

prabhjyotsingh:
No I think it should be disabled by default, it may increase the size of binary.

felixcheung:
great

prabhjyotsingh:

PR #2742, opened 2018-01-23
[ZEPPELIN-3168] Interpreter Settings Authorization

necosta:
### What is this PR for?
Attempts to introduce some control/authorization over interpreter settings.
Added 2 levels of permissions: Owner and Reader
Also added a new attribute to InterpreterProperty to prevent editing a value (UI only)

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3168

* Test interpreter setting creation/update/deletion
* Test interpreter setting property creation/update/deletion
* Test interpreter settings API

int_sett1

### Questions:
* Is there breaking changes for older versions? N?


necosta:
Hi @zjffdu , I didn't create a design doc but I hope the interpreter/overview.md file helps. 
We can always improve the .md file of course.

In a nutshell, I tried to use as much functionality as it exists, leveraging `set permissions` and re-using interpreter.json for persistence.

As you can see in the screenshot above, I slightly changed the position and title of "New Interpreter" button so it's more recognizable for the end-user.

I tested locally with basic Shiro setup and disabling `zeppelin.anonymous.allowed`.

I don't think there are any breaking changes but more tests are needed to be sure.

zeppelin-web/src/app/interpreter/interpreter.controller.js
```diff
@@ -737,6 +741,24 @@ function InterpreterCtrl($rootScope, $scope, $http, baseUrlSrv, ngToast, $timeou
     }
   }
 
+  $scope.shouldEdit = function (form, settingId) {
```
necosta: I would like to highlight that the permissions check is only done when clicking "Save" after "Edit". 
Although not perfect, this allows the user to check what permissions are set. Suggestions are welcome.

docs/usage/interpreter/overview.md
```diff
@@ -152,3 +152,19 @@ In such cases, interpreter process recovery is necessary. Starting from 0.8.0, u
...
+
+These settings can be set directly on `interpreter.json` or in the Interpreter Settings page, `Set permissions` option.
+
+## Interpreter properties
+
+Zeppelin's interpreter properties allow the setting of the following attributes:
+* Name - Name of the property
+* Value - Value of the property
+* Type - Value type of the property (number, string, url, ...)
+* Readonly - Flag indicating if property is editable or not in UI
```
zjffdu: Even admin can not edit it on UI if Readonly is true ?

zjffdu:
If user create new interpreter property, is it readonly by default or not ?

docs/usage/interpreter/overview.md
necosta: Good point. This should be exposed to the UI. I'll update the ToDo list

necosta:
Q: If user create new interpreter property, is it readonly by default or not ?
A: Editable by default

docs/usage/interpreter/overview.md
necosta: Done

necosta:
A few more changes:
- Exposed interpreter property `Description`
- Exposed interpreter property `Readonly`
- Added interpreter option `disallowCustomInterpreter` (Prevents users from using interpreter group to create a new interpreter)
- Validate permissions on selecting interpreters
- Disallow deletion of property when readonly
- Check for duplicate properties

necosta:
ping @zjffdu , @Leemoonsoo , @mebelousov ,
Feedback appreciated :) Thanks!

necosta:
ping @zjffdu , @Leemoonsoo , @mebelousov , @felixcheung 
Many thanks

zjffdu:
sorry @necosta I will review it after 0.8.0 release

necosta:
No problem @zjffdu ! Didn't know we were releasing 0.8.0. Happy we hear that :)

necosta:
FYI, got this one rebased on top of latest master. Happy to wait for 0.8.0 release before this MR gets reviewed

Tagar:
@necosta thanks for this great feature. It makes a lot of sense for Zeppelin instance shared with different set of users. Would it be possible to do this interpreter settings authorization for LDAP groups too, not just for users?

jongyoul:
I have a very basic question. Does this feature prevent from changing configurations by `ConfInterpreter` dynamically?

zjffdu:
This PR seems didn't touch ConfInterpreter, but it should do I believe.

jongyoul:
I agree. It should do

PR #2744, opened 2018-01-24
ZEPPELIN-3184. Use hadoop-azure to replace azure-storage 

zjffdu:
### What is this PR for?
So that user can use azure storage via hadoop filesystem. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3184

Travis CI Pass

### Questions:

### DIFF
--- a/zeppelin-zengine/pom.xml
+++ b/zeppelin-zengine/pom.xml
@@ -40,7 +40,7 @@
     <commons.lang3.version>3.4</commons.lang3.version>
     <commons.vfs2.version>2.0</commons.vfs2.version>
     <aws.sdk.s3.version>1.10.62</aws.sdk.s3.version>
-    <azure.storage.version>4.0.0</azure.storage.version>
+    <adl.sdk.version>2.1.4</adl.sdk.version>
     <jackrabbit.webdav.version>1.5.2</jackrabbit.webdav.version>
     <quartz.scheduler.version>2.2.1</quartz.scheduler.version>
     <lucene.version>5.3.1</lucene.version>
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/pom.xml
```diff
@@ -657,6 +625,69 @@
...
+          <exclusions>
+            <exclusion>
+              <groupId>com.fasterxml.jackson.core</groupId>
+              <artifactId>jackson-core</artifactId>
+            </exclusion>
+          </exclusions>
+        </dependency>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-aws</artifactId>
```
felixcheung: shouldn't this be under some vendor profile instead of building with all of azure + aws all the time?

zjffdu: hmm, make sense

zjffdu: Think about it twice. It might not be proper to put in vender profile. Because zeppelin's builtin AzureNotebookRepo and S3NotebookRepo has some dependencies which is dependency of hadoop-azure and hadoop-aws.

felixcheung: true... I think they should have a plugin model then, like interpreter, can be installed as needed.

zjffdu: That's correct. I have a plan to introduce plugin framework to zeppelin as zeppelin provide more extension point now like conf storage, notebook storage, lifecycle manager, interpreter launcher and etc. Plugin framework can help manage these plugins and create isolated classpath loader for them so that we don't need to worry about the jar conflict between them. But it would be in the next major release at least.

Tagar:
This commit broke build for CDH (maven convergence error on dependencies).

Should there be a automated test to check if a commit breaks compatibility with Cloudera distro of Hadoop?

zjffdu:
@Tagar I recently realize that zeppelin should work with CDH even built with apache hadoop. Could you try that ?

Tagar:
Do you mean compile without option like 

> -Dhadoop.version=2.6.0-cdh5.12.1

 ?

It compiles fine but I was getting runtime errors (like some non-existent method etc). Can't find details now. But I will give it a try again. 

Thanks.

PR #2746, opened 2018-01-25
ZEPPELIN-3190. Should not use singleton for FileSystemStorage

zjffdu:
### What is this PR for?
For now, `FileSystemNotebookRepo`, `FileSystemConfigStorage`, `FileSystemRecoveryStorage` use `FileSystemStorage`, but the singleton pattern means that all the notebook, config and recovery need to be stored in the same storage which might not be proper for some users. So this PR is trying to use separate `FileSystemStorage` instance for `FileSystemNotebookRepo`, `FileSystemConfigStorage`, `FileSystemRecoveryStorage` 

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3190

* Travis pass

### Questions:


zjffdu:
@prabhjyotsingh Please help review it, Thanks

PR #2750, opened 2018-01-30
ZEPPELIN-3108. Support Spark 2.3

zjffdu:
### What is this PR for?
Spark 2.3 remove `JobProgressListener` which cause zeppelin unable to run spark 2.3.  
This PR try to make Zeppelin support spark 2.3 via using `sc.statusTracker`, see `JobProgressUtil.scala`

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3108

* Verified manually. 

screen shot 2018-01-30 at 9 45 01 pm

### Questions:

### DIFF
--- a/spark/interpreter/pom.xml
+++ b/spark/interpreter/pom.xml
@@ -53,15 +53,7 @@
     <pyspark.test.exclude>**/PySparkInterpreterMatplotlibTest.java</pyspark.test.exclude>
     <pyspark.test.include>**/*Test.*</pyspark.test.include>
 
-
-    <spark.archive>spark-${spark.version}</spark.archive>
-    <spark.src.download.url>
-      http://d3kbcqa49mib13.cloudfront.net/${spark.archive}.tgz
-    </spark.src.download.url>
-    <spark.bin.download.url>
-      http://d3kbcqa49mib13.cloudfront.net/spark-${spark.version}-bin-without-hadoop.tgz
-    </spark.bin.download.url>
-
+    
   </properties>
 
   <dependencies>
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
@@ -35,6 +35,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.spark.JobProgressUtil;
 import org.apache.spark.SecurityManager;
 import org.apache.spark.SparkConf;
 import org.apache.spark.SparkContext;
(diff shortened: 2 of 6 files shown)


spark-dependencies/pom.xml
```diff
@@ -565,6 +565,15 @@
       </properties>
     </profile>
 
+    <profile>
```
felixcheung: can you update to 2.1.2 and 2.2.1 above?

zjffdu: Fixed

zjffdu: @felixcheung , looks like spark-2.2.1 is not uploaded in http://d3kbcqa49mib13.cloudfront.net/spark-2.2.1.tgz which is used for zeppelin to download it for integration test. Can you help check what's missing in spark side ?

zjffdu: Same issue for spark-2.1.2.  http://d3kbcqa49mib13.cloudfront.net/spark-2.1.2.tgz

felixcheung:
The s3 is no longer being used. We need to move to apache mirrors as per ASF policy..

zjffdu:
@felixcheung , download url is fixed, please help review again.

felixcheung:
Hmm we cant fix to one or two mirrors either, i think they tend to disappear - please see how other projects handles this

zjffdu:
Fixed. Now using apache dist url

felixcheung:
Ah sorry.
Please see
http://www.apache.org/dev/release-publishing.html#distribution

And specifically
http://www.apache.org/dev/release-download-pages.html#links

zjffdu:
@felixcheung IIUC, these are for release download page, for external project, we could use the dist link.

felixcheung:
well... that's not my understand and other implementations are pulling from the download script and then to the referred mirror, eg.

https://github.com/apache/spark/blob/master/build/mvn#L81
https://github.com/apache/spark/blob/master/sql/hive/src/test/scala/org/apache/spark/sql/hive/HiveExternalCatalogVersionsSuite.scala#L64
https://github.com/apache/spark/blob/master/R/pkg/R/install.R#L213
https://github.com/apache/spark/blob/master/external/docker/spark-mesos/Dockerfile#L27
https://github.com/graphframes/graphframes/commit/a1427217a2458ae89dde038f0ce79eec7913a7d6
https://github.com/Homebrew/homebrew-core/blob/master/Formula/apache-spark.rb

zjffdu:
Thanks @felixcheung , These examples seems to fetch the download url dynamically, I don't think maven download plugin support this. If you don't mind, I will leave it as follow up PR to improve. As I want to include this PR in 0.8 which I plan to release soon.

felixcheung:
I don't feel strongly about it, but fetching the url from the download script (.cgi url) is actually fairly simple to do, but I admit may not be what the maven download plugin support.

I don't remember exactly what these downloads are for ... can some of these be artifacts on maven central repo? that might be easier.

in any case, I'm only worry about push back from the ASF release process of downloading directly from dist.apache.org

zjffdu:
These download are for spark interpreter integration test. So spark distribution is needed, only spark artifact is not sufficient.

zjffdu:
ping @felixcheung

felixcheung:
Well, my question on the download link
https://github.com/apache/zeppelin/pull/2750#issuecomment-363663111

PR #2752, opened 2018-01-30
[ZEPPELIN-3195] Remove the limit on the number of run paragraphs at cron

mebelousov:
### What is this PR for?
Increase interpreter pool size for unlimit the number of paragraph that run at cron schedule.

### What type of PR is it?
[Hot Fix]

### What is the Jira issue?
*https://issues.apache.org/jira/browse/ZEPPELIN-3195

* Create note with 15 paragraphs
* Check that at cron will be executed only 10 paragraphs
* Checkout this branch and rebuild app
* Check that all paragraphs are executed at cron

### Questions:

### DIFF
--- a/docs/setup/operation/configuration.md
+++ b/docs/setup/operation/configuration.md
@@ -258,18 +258,6 @@ If both are defined, then the **environment variables** will take priority.
     <td>Make notebook public (set only <code>owners</code>) by default when created/imported. If set to <code>false</code> will add <code>user</code> to <code>readers</code> and <code>writers</code> as well, making it private and invisible to other users unless permissions are granted.</td>
   </tr>
   <tr>
-    <td><h6 class="properties">ZEPPELIN_INTERPRETERS</h6></td>
-    <td><h6 class="properties">zeppelin.interpreters</h6></td>
-  <description></description>
-    <td>org.apache.zeppelin.spark.SparkInterpreter,<br />org.apache.zeppelin.spark.PySparkInterpreter,<br />org.apache.zeppelin.spark.SparkSqlInterpreter,<br />org.apache.zeppelin.spark.DepInterpreter,<br />org.apache.zeppelin.markdown.Markdown,<br />org.apache.zeppelin.shell.ShellInterpreter,<br />
-    ...
-    </td>
-    <td>
-      Comma separated interpreter configurations [Class] <br/><br />
-      <span style="font-style:italic; color: gray">NOTE: This property is deprecated since Zeppelin-0.6.0 and will not be supported from Zeppelin-0.7.0.</span>
-    </td>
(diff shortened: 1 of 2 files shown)


zjffdu:
Actually, you can change the configuration in `zeppelin-site.xml`,  and change it to 65535 seems too large for me.

mebelousov:
It's not easy to understand that "ZEPPELIN_INTERPRETER_MAX_POOL_SIZE" affects the number of paragraphs executed at cron. I would be glad if nobody will meet this bug and no need to change zeppelin-site.xml to fix note scheduling.

@zjffdu which number will be more appropriate? =)

mebelousov:
@felixcheung 
Where I could document about ZEPPELIN_INTERPRETER_MAX_POOL_SIZE?
There is no such cases in documentation.

Also I cannot imagine than Zeppelin administrator will go to documentation when users will complain that manual note run is OK but only 10 first paragraphs are done at cron.

mebelousov:
I renewed docs: remove old "ZEPPELIN_INTERPRETERS" and added ZEPPELIN_INTERPRETER_MAX_POOL_SIZE.
The default value for ZEPPELIN_INTERPRETER_MAX_POOL_SIZE remains 10.

docs/setup/operation/configuration.md
```diff
@@ -281,6 +269,12 @@ If both are defined, then the **environment variables** will take priority.
     <td>http://repo1.maven.org/maven2/</td>
     <td>Remote principal repository for interpreter's additional dependency loading</td>
   </tr>
+  <tr>
+    <td><h6 class="properties">ZEPPELIN_INTERPRETER_MAX_POOL_SIZE</h6></td>
+    <td><h6 class="properties">zeppelin.interpreter.max.poolsize</h6></td>
+    <td>10</td>
+    <td>The maximum number of paragraphs that will be run at cron.</td>
```
zjffdu: This doesn't sound like a correct description. `zeppelin.interpreter.max.poolsize` is for the number threads of per session configuration. That means if you are using 20 different interpreters, they can run simultaneously without affecting each other.

mebelousov: I took note with 4 jdbc paragraph and a lof of python paragraphs.
You can see that only first 10 paragraphs were done.
first10

mebelousov:
@zjffdu @felixcheung 
What will we do with this bug?

docs/setup/operation/configuration.md
zjffdu: @mebelousov This might be fixed in ZEPPELIN-3171, could you try that ?

mebelousov: Year! This is fixed.
@zjffdu thank you!

PR #2757, opened 2018-02-01
[ZEPPELIN-3198] UI should not show Version/GIT Control if the same if not supported

prabhjyotsingh:
### What is this PR for?
Currently, UI shows an option for version/GIT Control even when it is not supported by the underlying implementing storage configuration.

It is only after users try to save a commit and get an error "Couldn't checkpoint note revision: possibly storage doesn't support versioning. Please check the logs for more details.".

So, if implementing storage configuration doesn't support git storage, UI should not show those options.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3198](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3198)

On using "org.apache.zeppelin.notebook.repo.GitNotebookRepo"  for `zeppelin.notebook.storage`  user should see revision/version control option, for rest of the others e.g. "FileSystemNotebookRepo" user should not see that option 

### Questions:


jhonderson:
The methods for the versioning of the notes are defined at interface level in NotebookRepo.java:

 - Revision checkpoint(String noteId, String checkpointMsg, AuthenticationInfo)
 - Note get(String noteId, String revId, AuthenticationInfo)
 - List<Revision> revisionHistory(String noteId, AuthenticationInfo)
 - Note setNoteRevision(String noteId, String revId, AuthenticationInfo)

They could be implemented by a non-git repository, for example a repository that saves the notes in S3 and its versioning information in a database or something like that. So i guess a better solution would be a method in NotebookRepo.java that indicates if the implementation supports versioning (maybe doesSupportVersioning()); and use that method to show or hide the versioning feature.

What do you think?.

prabhjyotsingh:
Yes, I believe that is exactly what I'm trying to do over here, for example in S3NotebookRepo, since none of checkpoint, get, revisionHistory, or setNoteRevision is implemented (https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/S3NotebookRepo.java#L292) hence for the same the method isRevisionSupported returns false. Let me know if I'm missing out on a case.

prabhjyotsingh:
@felixcheung @zjffdu  can you please help review this?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/GitNotebookRepo.java
```diff
@@ -39,8 +36,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Joiner;
-import com.google.common.collect.Lists;
+import java.io.File;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
```
felixcheung: any idea why changes are happening?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -535,4 +528,15 @@ public Note setNoteRevision(String noteId, String revId, AuthenticationInfo subj
     }
     return revisionNote;
   }
+
+  @Override
+  public Boolean isRevisionSupported() {
+    try {
+      return getRepo(0).isRevisionSupported();
+    } catch (IOException e) {
```
felixcheung: is IOException sufficient here? we have many different repo implementations..

prabhjyotsingh: I used this based on one of the previous method ` List<Revision> revisionHistory` implementation
https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java#L481

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/GitNotebookRepo.java
prabhjyotsingh: Will revert this, mostly done by my IDE.

zjffdu:
It sounds a little redundant to ask user to both implement these version related method and `isRevisionSupported`. How about creating a sub-interface of `NotebookRepo` which support version control ? And if NotebookRepo using is the implementation of this sub-interface, we show version control on UI

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -431,6 +432,28 @@ public void close() {
     }
   }
 
+  public Boolean isDefaultRepoGit() {
```
jhonderson: Could be {return isRepoGit(0)}. And in isRepoGit maybe a better error message for catching of IOException.

prabhjyotsingh: Yes, that sounds simpler, thanks.

prabhjyotsingh:
@zjffdu have made the suggested changes, waiting for CI, does this looks good?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookGitRepo.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.zeppelin.user.AuthenticationInfo;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Notebook repository (persistence layer) abstraction
+ */
+public interface NotebookGitRepo extends NotebookRepo {
```
zjffdu: Git is just one implementation of version control, How about naming it as `NotebookRepoWithVersionControl` ? It might be a little long, maybe you or others can think of a better name.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoCommon.java
```diff
@@ -0,0 +1,261 @@
...
+import org.apache.zeppelin.notebook.ApplicationState;
+import org.apache.zeppelin.notebook.Note;
+import org.apache.zeppelin.notebook.NoteInfo;
+import org.apache.zeppelin.notebook.Paragraph;
+import org.apache.zeppelin.scheduler.Job;
+import org.apache.zeppelin.user.AuthenticationInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NotebookRepoCommon {
```
zjffdu: Any reason to create a new class instead of putting them in VFSNotebookRepo ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookGitRepo.java
prabhjyotsingh: Sure, sounds right.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoCommon.java
prabhjyotsingh: There's a lot of common functionality between `GitNotebookRepo` and `VFSNotebookRepo` one way was to refactor these to `NotebookRepoCommon` or `class GitNotebookRepo extends VFSNotebookRepo implements NotebookRepoWithVersionControl`, shall I do the later?

prabhjyotsingh:
Shall I merge this, if no more discussion?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoCommon.java
zjffdu: The later make more sense to me.

prabhjyotsingh: Sure, have reverted `NotebookRepoCommon` related changes.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -477,7 +488,9 @@ public Revision checkpoint(String noteId, String checkpointMsg, AuthenticationIn
   public Note get(String noteId, String revId, AuthenticationInfo subject) {
     Note revisionNote = null;
     try {
-      revisionNote = getRepo(0).get(noteId, revId, subject);
+      if (isDefaultRepoGit()) {
```
zjffdu: should be `getRepo(0) instanceof NotebookRepoWithVersionControl`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -301,6 +302,14 @@ void setNotebookRepo(NotebookRepo repo) {
     this.repo = repo;
   }
 
+  void setRevisionSupported(NotebookRepo repo) {
+    if (repo instanceof NotebookRepoWithVersionControl) {
```
zjffdu: IIUC, repo must be `NotebookRepoSync`,  so it would always be NotebookRepoWithVersionControl

prabhjyotsingh: Have made the relevant changes.

zjffdu:
LGTM

prabhjyotsingh:

PR #2759, opened 2018-02-01
ZEPPELIN-3202. Added missing test dependencies in the scio module

HorizonNet:
### What is this PR for?

Added missing test dependencies for the **scio** module to prevent test failures when running `mvn clean install`.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3202

* CI pass

### Questions:

### DIFF
--- a/scio/pom.xml
+++ b/scio/pom.xml
@@ -43,6 +43,9 @@
     <plugin.shade.version>2.3</plugin.shade.version>
     <plugin.scala.version>2.15.2</plugin.scala.version>
     <plugin.avro.version>1.7.7</plugin.avro.version>
+
+    <!--test library versions-->
+    <hamcrest.all.version>1.3</hamcrest.all.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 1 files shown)


scio/pom.xml
```diff
@@ -43,6 +43,9 @@
     <plugin.shade.version>2.3</plugin.shade.version>
     <plugin.scala.version>2.15.2</plugin.scala.version>
     <plugin.avro.version>1.7.7</plugin.avro.version>
+
+    <!--test library versions-->
+    <hamcrest.all.version>1.3</hamcrest.all.version>
```
felixcheung: see https://github.com/apache/zeppelin/pull/2759/files#diff-608f20f45ac75cfdb867d8d83eb0216bR98?

felixcheung: this should be added as test scope

HorizonNet: **hamcrest-all** is already set as test scope (see line 107ff).

HorizonNet:
I kicked off another Travis build and it seems that these failing tests are flakey.

HorizonNet:
I just reran the failing builds, as suggested by @zjffdu in ZEPPELIN-3182, and all are green now.

felixcheung:
merging if no more comment

PR #2769, opened 2018-02-07
ZEPPELIN-3171. Restart of interpreter in note also aborts running interpreter in another note

zjffdu:
### What is this PR for?

The root cause is that in isolated mode interpreters will share the same scheduler. That means when one interpreter is terminated, all the running jobs under the scheduler of this interpreter will be aborted too. This PR will create one scheduler for each session. So when one session is closed, only the running jobs under this session's scheduler is aborted. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3171

* Unit test is added.
* Also verify it manually. 

### Questions:


zjffdu:
@prabhjyotsingh Mind to help review it ? Thanks

prabhjyotsingh:
LGTM!

PR #2777, opened 2018-02-08
ZEPPELIN-3206. Move spark integration test to zeppelin-integration

zjffdu:
### What is this PR for?
For now, zeppelin-zengine depends on spark module because there's some spark integration test under zeppelin-zengine. This PR is trying to move that into zeppelin-integration so that zeppelin-zengine doesn't need depend on spark.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3206

* CI pass

### Questions:


jongyoul:
Is this still effective?

PR #2782, opened 2018-02-09
[ZEPPELIN-2729] Paragraph numbering

timovwb:
### What is this PR for?
Added a button on the top toolbar of the notebook to toggle paragraph numbering. It adds a "Paragraph N" on the top-right of each paragraph, mainly useful when collaborating and guiding others to the right place.

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-2729](https://issues.apache.org/jira/browse/ZEPPELIN-2729)

### Screenshots
[gfycat mp4 video](https://gfycat.com/LeafyAntiqueCrab)

### Questions:


masyan:
there is Title of paragraph for that

mebelousov:
@timovwb I also thought about good way to point on the right paragraph.
How about to add new item in drop-down paragraph menu (like "Link this paragraph") that will form url for this paragraph in note?

timovwb:
@masyan Using paragraph titles is still not a great way in large notebooks. When having a descriptive title, you still have to search the notebook for that (although the browser search function might help here, but still not perfect). And manually setting a paragraph number gives problems when moving/adding/removing paragraphs.

@mebelousov A similar feature already exists, but links to a single paragraph. Linking inside a notebook itself could be a nice feature. The nice thing with actual numbering is when collaborating through Skype for example it's very quick to give directions.

Tagar:
Since we migrated from Jupyter to Zeppelin, I miss a lot floating ToC menu Jupyter extension.
Super useful for larger notebooks - navigation becomes very transparent.

http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/toc2/README.html

demo3

demo21

demo4

Hope to see something like this in Zeppelin one day.

mebelousov:
@Tagar there is "zeppelin-toc-spell" in the Helium packages.
Now it supports only headers inside markdown interpreter. It will be good to bring new functions to "zeppelin-toc-spell".

image

Tagar:
I think a low-hanging fruit here might be if Zeppelin would treat paragraph titles as mark down code?
So if you added leading `#` it would mean a high-level section of the notebook, `##` would mean
a sub-section etc. So then it should be more straight-forward to implement something like above 
mentioned floating ToC with automatic section numbering from Jupyter natively in Zeppelin. 
I've seen zeppelin-toc-spell but haven't tried to use it. What do you guys think?

mebelousov:
I would prefer that the paragraph title is always added to the ToC as high-level section.

Tagar:
I think it's a good improvement. As an example, Databricks spark notebooks also have paragraph numbering across whole notebook.
Thanks.

r-kamath:
@timovwb please resolve the conflicts

jongyoul:
ping @timovwb

jongyoul:
Thank you for the contribution. BTW, this looks like showing only in frontside and didn't store/pass any information to the backend. It would be better to be passed and saved to the backend. WDYT?

timovwb:
I'm not familiar with the Zeppelin codebase, so I'm not sure what the best practice is on how to handle this feature. It's even been a while since working on this, it was mostly looking at how other similar features were implemented. If it's a better idea in general to pass this information to the backend and nobody is willing to do it, I'll give it a go if you can point me into the right direction or documentation on how this is done.

jongyoul:
@timovwb Basically, all contributions are welcome :-) I just wonder how we can improve this feature, not just showing paragraph's number because I don't have any idea to improve this feature. Your contribution is enough and someone will add passing them to the backend if they need it.

jongyoul:
BTW, could you please check CI and make green?

https://travis-ci.org/timovwb/zeppelin/builds/401408671

ParkGyeongTae (approved):
👍 LGTM

jongyoul:
@ParkGyeongTae Thank you for checking it up. We can merge it but can we get this feature in the new UI?

PR #2793, opened 2018-02-13
ZEPPELIN-3226. Fail to launch IPySparkInterpreter in embedded mode

zjffdu:
### What is this PR for?
Trivial PR for fixing this issue. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3226

* CI Pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
@@ -51,9 +51,10 @@ public class IPySparkInterpreter extends IPythonInterpreter {
         PySparkInterpreter.getPythonExec(getProperties()));
     sparkInterpreter = getSparkInterpreter();
     SparkConf conf = sparkInterpreter.getSparkContext().getConf();
-    // only set PYTHONPATH in local or yarn-client mode.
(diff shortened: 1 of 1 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
```diff
@@ -51,9 +51,10 @@ public void open() throws InterpreterException {
         PySparkInterpreter.getPythonExec(getProperties()));
     sparkInterpreter = getSparkInterpreter();
     SparkConf conf = sparkInterpreter.getSparkContext().getConf();
-    // only set PYTHONPATH in local or yarn-client mode.
+    // only set PYTHONPATH in embedded, local or yarn-client mode.
     // yarn-cluster will setup PYTHONPATH automatically.
-    if (!conf.get("spark.submit.deployMode").equals("cluster")) {
+    if (!conf.contains("spark.submit.deployMode") ||
+        !conf.get("spark.submit.deployMode").equals("cluster")) {
```
felixcheung: I think it's better to whitelist instead of a blacklist. also we should check the `master` - other places people are using with mesos, kubernetes etc.

felixcheung:
@Leemoonsoo

PR #2802, opened 2018-02-15
ZEPPELIN-3236. Make grpc framesize configurable

zjffdu:
### What is this PR for?

Add one new property `zeppelin.ipython.grpc.framesize` which is an advanced configuration. By default it is 32M which should be sufficient for most of scenarios. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3236

* Unit test is added.

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
+++ b/python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
@@ -20,6 +20,7 @@ package org.apache.zeppelin.python;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.StreamObserver;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.zeppelin.interpreter.util.InterpreterOutputStream;
 import org.apache.zeppelin.python.proto.CancelRequest;
 import org.apache.zeppelin.python.proto.CancelResponse;
@@ -131,11 +132,18 @@ public class IPythonClient {
       @Override
       public void onError(Throwable throwable) {
         try {
+          interpreterOutput.getInterpreterOutput().write(ExceptionUtils.getStackTrace(throwable));
           interpreterOutput.getInterpreterOutput().flush();
         } catch (IOException e) {
           LOGGER.error("Unexpected IOException", e);
         }
         LOGGER.error("Fail to call IPython grpc", throwable);
+        finalResponseBuilder.setStatus(ExecuteStatus.ERROR);
+
+        completedFlag.set(true);
+        synchronized (completedFlag) {
+          completedFlag.notify();
+        }
       }
 
       @Override
--- a/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.python;
 
+import io.grpc.ManagedChannelBuilder;
 import org.apache.commons.exec.CommandLine;
 import org.apache.commons.exec.DefaultExecutor;
 import org.apache.commons.exec.ExecuteException;
@@ -137,7 +138,10 @@ public class IPythonInterpreter extends Interpreter implements ExecuteResultHand
       int jvmGatewayPort = RemoteInterpreterUtils.findRandomAvailablePortOnAllLocalInterfaces();
       LOGGER.info("Launching IPython Kernel at port: " + ipythonPort);
       LOGGER.info("Launching JVM Gateway at port: " + jvmGatewayPort);
-      ipythonClient = new IPythonClient("127.0.0.1", ipythonPort);
+      int framesize = Integer.parseInt(getProperty("zeppelin.ipython.grpc.framesize",
+          32 * 1024 * 1024 + ""));
+      ipythonClient = new IPythonClient(ManagedChannelBuilder.forAddress("127.0.0.1", ipythonPort)
+          .usePlaintext(true).maxInboundMessageSize(framesize));
       launchIPythonKernel(ipythonPort);
       setupJVMGateway(jvmGatewayPort);
     } catch (Exception e) {
--- a/python/src/main/resources/interpreter-setting.json
+++ b/python/src/main/resources/interpreter-setting.json
@@ -40,6 +40,12 @@
         "defaultValue": "30000",
         "description": "time out for ipython launch",
         "type": "number"
+      },
+      "zeppelin.ipython.grpc.framesize": {
+        "propertyName": "zeppelin.ipython.grpc.framesize",
+        "defaultValue": "33554432",
+        "description": "grpc framesize, default is 32M",
+        "type": "number"
       }
     },
     "editor": {
--- a/python/src/test/java/org/apache/zeppelin/python/IPythonInterpreterTest.java
+++ b/python/src/test/java/org/apache/zeppelin/python/IPythonInterpreterTest.java
@@ -56,9 +56,7 @@ public class IPythonInterpreterTest {
   private static final Logger LOGGER = LoggerFactory.getLogger(IPythonInterpreterTest.class);
   private IPythonInterpreter interpreter;
 
-  @Before
-  public void setUp() throws InterpreterException {
-    Properties properties = new Properties();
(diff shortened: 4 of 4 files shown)


Tagar:
Thanks for the heads up, yep I figured out that I have to tune up zeppelin.ipython.grpc.framesize to a large number. 

I looked over the PR. Two quick suggestions
1) Would it be possible to make spark interpreter keep and not close the stream if such an exception happens? We can see a higher limit, but I am sure users will have cases when they will try to go higher. The Spark interpreter is then in a bad state and only way to fix this is to try increase a limit again.. Not sure if this problem belongs to Zeppelin or to grpc, so provisionally opened an issue in grpc too - https://github.com/grpc/grpc-java/issues/4086
2) Should we increase the default? .. 4Mb isn't that hard to hit when ipython returns a mid-size dataset / table.

zjffdu:
@Tagar , 4 Mb is the default value of gprc, I am fine to use another large. But from your error logs, your code will produce 200M data, I am curious to know how do you produce that, do you mind to paste your code to reproduce it ?

```
io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: io.grpc.netty.NettyClientTransport$3: Frame size 216695976 exceeds maximum: 4194304.
```

Tagar:
1) My main point was that this exception should be thrown to the user, so he or she has a chance to increase this limit. Currently if it breaks, only way to find out about this limitation is to enable debugging and not a lot of users can do that. 

2) You're right .. it's 200M not sure how that user got that much data. That wasn't from my code, but from a colleague of mine. I guess it was a larger table of data. Would you mine making default somewhere in the range 16-32M? I think a lot of folks would run into the 4M limit.

3) Also, it would be great if IPythonInterpreter would catch exceptions better. Found another problem - https://issues.apache.org/jira/browse/ZEPPELIN-3239 - unrelated to this one, but it also shows the same symptoms to the user - Spark interpreter just becomes irresponsive.

zjffdu:
@Tagar Do you have any sample code to reproduce this issue ? So that I can verify it and also may include it in unit test.

Tagar:
@zjffdu I don't have a reproducible case yet. This error hasn't been very consistent for me so it might be also a side effect of another problem. That's why I asked above if we could capture exceptions in grpc better. Also check what you think of the suggestion made in grpc/grpc-java#4086 to make a new grpc stream for each call?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -137,7 +138,10 @@ public void open() throws InterpreterException {
       int jvmGatewayPort = RemoteInterpreterUtils.findRandomAvailablePortOnAllLocalInterfaces();
       LOGGER.info("Launching IPython Kernel at port: " + ipythonPort);
       LOGGER.info("Launching JVM Gateway at port: " + jvmGatewayPort);
-      ipythonClient = new IPythonClient("127.0.0.1", ipythonPort);
+      int framesize = Integer.parseInt(getProperty("zeppelin.ipython.grpc.framesize",
+          4 * 1024 * 1024 + ""));
+      ipythonClient = new IPythonClient(ManagedChannelBuilder.forAddress("127.0.0.1", ipythonPort)
+          .usePlaintext(true).maxInboundMessageSize(framesize));
```
felixcheung: does this `.usePlaintext(true)` change the behavior?

zjffdu:
@Tagar Can you paste the full log ? I'd like to see the full stacktrace.

Tagar:
> ERROR [2018-02-14 10:39:10,922] ({grpc-default-executor-3} IPythonClient.java[onError]:138) - Fail to call IPython grpc
> io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: io.grpc.netty.NettyClientTransport$3: Frame size 216695976 exceeds maximum: 4194304.
>         at io.grpc.Status.asRuntimeException(Status.java:543)
>         at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:395)
>         at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:426)
>         at io.grpc.internal.ClientCallImpl.access$100(ClientCallImpl.java:76)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:512)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$700(ClientCallImpl.java:429)
>         at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:544)
>         at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:52)
>         at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:117)
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
>         at java.lang.Thread.run(Thread.java:748)
>  INFO [2018-02-14 10:39:10,923] ({grpc-default-worker-ELG-1-2} AbstractClientStream2.java[inboundDataReceived]:249) - Received data on closed stream
>  INFO [2018-02-14 10:39:10,924] ({grpc-default-worker-ELG-1-2} AbstractClientStream2.java[inboundDataReceived]:249) - Received data on closed stream
>  INFO [2018-02-14 10:39:10,925] ({grpc-default-worker-ELG-1-2} AbstractClientStream2.java[inboundDataReceived]:249) - Received data on closed stream
> 

@zjffdu that's what I have.. is this helpful? It seems contain full exception stack.

zjffdu:
@Tagar Have you tried this PR ? Does this PR solve your issue ?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
zjffdu: No, it is the same as previous behavior.

Tagar:
thank you @zjffdu - I will check this today.

python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
```diff
@@ -131,11 +132,18 @@ public void onNext(ExecuteResponse executeResponse) {
...
         } catch (IOException e) {
           LOGGER.error("Unexpected IOException", e);
         }
         LOGGER.error("Fail to call IPython grpc", throwable);
+        finalResponseBuilder.setStatus(ExecuteStatus.ERROR);
+
+        completedFlag.set(true);
+        synchronized (completedFlag) {
+          completedFlag.notify();
+        }
```
Tagar: that's great! just tested and glad to see errors are now getting its way to the fronted, without making spark interpreter appear "hanging".

Tagar:
@zjffdu I just verified this new commit and it works great.
Glad to see exceptions in ipython backend are now being captured correctly. 

I wasn't able to actually test frame size problem, as it's hard to reproduce.

As @ejona86 said today in https://github.com/grpc/grpc-java/issues/4086#issuecomment-367053465

> Note that something else is broken if you're seeing that error. You should not be receiving a frame that large. (Chances are it's not actually a frame length and we're seeing garbage.)

So I think that frame size error might be caused by some other problem.
But it's great we can make frame size configurable now.

Tagar:
LGTM

thanks a lot @zjffdu

python/src/main/java/org/apache/zeppelin/python/IPythonClient.java
zjffdu: Thanks @Tagar for verifying and review.

ejona86:
@Tagar, I mis-read the error earlier. That message is just "gRPC message too large." It'd probably be good to talk about "messages" instead of "frames" here.

zjffdu:
@ejona86 The error from grpc is about `Frame size exceed`

ejona86:
I just changed them to be more clear in grpc/grpc-java#4106. Each message gets its own gRPC frame, but the very fact there are gRPC frames is more of an internal detail; users shouldn't need to know about such things.

zjffdu:
Thanks @ejona86

zjffdu:
rename the property in #2813

PR #2805, opened 2018-02-16
[ZEPPELIN-3240] Zeppelin server fail to start if interpreter has mixed properties

prabhjyotsingh:
### What is this PR for?
I found that Zeppelin server is failing to start when there is a mix of both kind of properties in interpreter;

```
"properties": {
        "shell.command.timeout.millisecs": {
          "type": "string",
          "name": "shell.command.timeout.millisecs",
          "value": "60000"
        },
        "shell.working.directory.user.home": {
          "type": "checkbox",
          "name": "shell.working.directory.user.home",
          "value": false
        },
        "zeppelin.shell.auth.type": "KERBEROS",
        "zeppelin.shell.keytab.location": "/etc/security/keytabs/zeppelin.server.kerberos.keytab",
        "zeppelin.shell.principal": "zeppelin@EXAMPLE.COM"
      }
```

Mix as in these two kind of properties:

 - Where key is string and value is object
```
"shell.command.timeout.millisecs": {
          "type": "string",
          "name": "shell.command.timeout.millisecs",
          "value": "60000"
        }
```

 - Where key and value both are string
```
"zeppelin.shell.auth.type": "KERBEROS"
```

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3240](https://issues.apache.org/jira/browse/ZEPPELIN-3240)

* Place this [interpreter.json](https://issues.apache.org/jira/secure/attachment/12910862/interpreter.json) in Z-server conf folder and try to start Z-Server, with this patch it should start.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -26,7 +26,6 @@ import com.google.gson.JsonObject;
 import com.google.gson.annotations.SerializedName;
 import com.google.gson.internal.StringMap;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.dep.Dependency;
 import org.apache.zeppelin.dep.DependencyResolver;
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
@zjffdu  can you help review this.

Tagar:
Are those kind of properties mix in interpreter could happen because of Zeppelin upgrade?
I.e. old properties and new properties mix?

prabhjyotsingh:
@Tagar  Yes, this can happen in Zeppelin upgrade, not in all upgrades, but in some trivial case hard to reproduce.

PR #2806, opened 2018-02-16
[ZEPPELIN-3241] fix NPE when restarting Interpreter

weand:
### What is this PR for?
We often (but not always) face an NPE when restarting Spark interpreter. Our Spark interpreter run in scoped mode per note. The restart then is not possible, only restart of whole zeppelin daemon helps out.

This PR fixes this NPE.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[https://issues.apache.org/jira/browse/ZEPPELIN-3241](https://issues.apache.org/jira/browse/ZEPPELIN-3241)

- manual

### Questions:


zjffdu:
LGTM

PR #2807, opened 2018-02-18
ZEPPELIN-3234. z.show() compatibility with previous release

zjffdu:
### What is this PR for?
Enhance the ZeppelinContext in IPySparkInterpreter

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3234

* Unit test is added

### Questions:

### DIFF
--- a/spark/interpreter/src/main/resources/python/zeppelin_ipyspark.py
+++ b/spark/interpreter/src/main/resources/python/zeppelin_ipyspark.py
@@ -51,3 +51,17 @@ if intp.isSpark2():
     sqlContext = sqlc = __zSqlc__ = __zSpark__._wrapped
 else:
     sqlContext = sqlc = __zSqlc__ = SQLContext(sparkContext=sc, sqlContext=intp.getSQLContext())
+
+class IPySparkZeppelinContext(PyZeppelinContext):
+
+    def __init__(self, z):
+        super(IPySparkZeppelinContext, self).__init__(z)
+
+    def show(self, obj):
+        from pyspark.sql import DataFrame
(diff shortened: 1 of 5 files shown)


spark/interpreter/src/main/resources/python/zeppelin_ipyspark.py
```diff
@@ -51,3 +51,17 @@
...
+
+    def __init__(self, z):
+        super(IPySparkZeppelinContext, self).__init__(z)
+
+    def show(self, obj):
+        from pyspark.sql import DataFrame
+        if isinstance(obj, DataFrame):
+            print(self.z.showData(obj._jdf))
+        else:
+            super(IPySparkZeppelinContext, self).show(obj)
```
felixcheung: why not just call `super(IPySparkZeppelinContext, self).show(obj)` always?

zjffdu: Here PySparkZeppelinContext enhance and extend PyZeppelinContext so that PySparkZeppelinContext can not only display pandas dataframe but also spark dataframe.

zjffdu: Dataset is not supported in PySpark IIUC. There's no dataset related api in pyspark.

felixcheung:
Ah yes. There is however a bit of fuzzy line with in mixed language mode (ie. Dataset created in Scala and then turn to Python)

PR #2808, opened 2018-02-18
[ZEPPELIN-3243] fix IndexOutOfBoundsException when 'No interpreter is binded to this note'

weand:
### What is this PR for?
I observed IndexOutOfBoundsException when using zeppelin locally on an existing notebook repository and with fresh interpreter.json file.
The default Zeppelin behavior when using existing notebooks is the manual binding of the interpreter settings, which works fine. This PR simply avoids the IndexOutOfBoundsException.

### What type of PR is it?
Bug Fix

### Todos
-

### What is the Jira issue?
[https://issues.apache.org/jira/browse/ZEPPELIN-3243](https://issues.apache.org/jira/browse/ZEPPELIN-3243)

- manual

### Questions:


zjffdu:
LGTM

Tagar:
I think I've seen this behavior too. Thanks for fixing this.

weand:
We did not face any other IndexOutOfBoundException so far. So I would propose to merge that first.

zjffdu:
LGTM

PR #2809, opened 2018-02-20
[ZEPPELIN-3249] Add support for streaming table

prabhjyotsingh:
### What is this PR for?
Since Zeppelin support streaming from various backends, I think it will be useful if even tables (and later graphs) can also be streamed.

### What type of PR is it?
[Improvement | Feature]

### Todos
* [x] - At times it fails with `Uncaught TypeError: Cannot read property 'p20180220_113300_1663645286_0_table_gridApi' of undefined`, have to fix this error.

### What is the Jira issue?
* [ZEPPELIN-3249](https://issues.apache.org/jira/browse/ZEPPELIN-3249)

I have done it using shell interpreter, but this should work for all other backends as well
```
%sh
echo "%table"
echo "Col1	Col2"
echo "1	2"
sleep 1
echo "3	4"
echo "5	6"
sleep 2
echo "7	8"
sleep 3
echo "9	10"
echo "11	12"
sleep 4
echo "12	13"
```

zeppelin-3249

### Questions:


prabhjyotsingh:
Here is another example for python.
```
%python
print "%table col1	col2"
print "i 1	i 1"
wait_time = 2
import time
time.sleep(1)
for x in range(2, 200):
    print "i",x,"	i",x
    if x == wait_time:
        time.sleep(2)
        wait_time = wait_time*2
```

zeppelin-3249-python

prabhjyotsingh:
@r-kamath can you help review this PR

Tagar:
I think as a nice side effect, this PR also fixes https://issues.apache.org/jira/browse/ZEPPELIN-3238 - see comments there

r-kamath:
Tested. LGTM.

weand:
seems like it has changed table behavior when height of paragraph has been changed.
https://issues.apache.org/jira/browse/ZEPPELIN-3289
@prabhjyotsingh can you fix that bug?

prabhjyotsingh:
Sure, let me take a look.

PR #2814, opened 2018-02-24
ZEPPELIN-3255. Can not run spark1 and spark2 in one zeppelin instance

zjffdu:
### What is this PR for?
Although #2750 enable the support of spark 2.3, it breaks the support of spark 1.6. Users have to build zeppelin against spark 1.6 to make zeppelin work with that. But previous one zeppelin instance can work with multiple versions of spark. This PR introduce spark shims module which is to resolve the api incompatible issue between different versions of spark, so that one zeppelin instance can work with multiple versions of spark. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* https://issues.apache.org/jira/browse/ZEPPELIN-3254  Although zeppelin should support to run multiple versions of spark in one instance, but our travis test doesn't cover it, ZEPPELIN-3254 would do that.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3255

* CI pass

### Questions:


zjffdu:
@Leemoonsoo @felixcheung Could you help review it ? Thanks

zjffdu:
Will merge it soon as it block 0.8 release.

spark/spark2-shims/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.zeppelin</groupId>
+  <artifactId>spark2-shims</artifactId>
+  <version>0.9.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <name>Zeppelin: Spark2 Shims</name>
+
+  <properties>
+    <scala.binary.version>2.11</scala.binary.version>
+    <spark.version>2.1.2</spark.version>
```
felixcheung: instead of hardcoded here, should this be in the parent, spark-parent pom?

also this should change to 2.2.1?

zjffdu: spark 2.2.1 requires jdk8 while in travis we still use jdk7. The final goal is to make one build of zeppelin to support multiple versions of spark instead of building zeppelin against different version of spark. So finally I would remove all the profiles for different versions of spark. This would be done in https://issues.apache.org/jira/browse/ZEPPELIN-3254

PR #2815, opened 2018-02-24
[ZEPPELIN-3262] Make install_external_dependencies.sh work on macOS

mephi42:
### What is this PR for?
This original purpose of this script is to setup Travis environment,
but it's also useful for local development.

Also make it possible to run this script multiple times.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3262

* Run ./testing/install_external_dependencies.sh

### Questions:

### DIFF
--- a/testing/install_external_dependencies.sh
+++ b/testing/install_external_dependencies.sh
@@ -16,7 +16,7 @@
 # limitations under the License.
 #
 
-# Script for installing R / Python dependencies for Travis CI
+# Script for installing R / Python dependencies for Travis CI and local development environments
 set -ev
 touch ~/.environ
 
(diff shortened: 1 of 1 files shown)


testing/install_external_dependencies.sh
```diff
@@ -35,15 +35,28 @@ fi
...
+  fi
   echo "export PATH='$HOME/miniconda/bin:$PATH'" >> ~/.environ
   source ~/.environ
   hash -r
   conda config --set always_yes yes --set changeps1 no
   conda update -q conda
   conda info -a
   conda config --add channels conda-forge
-  conda install -q matplotlib pandasql ipython=5.4.1 jupyter_client ipykernel matplotlib bokeh=0.12.10
+  conda install -q matplotlib pandasql ipython=5.4.1 jupyter_client ipykernel matplotlib bokeh=0.12.10 py4j
```
felixcheung: not sure we should install py4j here - spark for example carries its own version of it

testing/install_external_dependencies.sh
```diff
@@ -35,15 +35,28 @@ fi
...
+    ;;
+  *)
+    os=Linux
+    ;;
+  esac
+  if [ ! -e miniconda.sh ]; then
+    wget "https://repo.continuum.io/miniconda/Miniconda${PYTHON}-4.2.12-$os-$(uname -m).sh" -O miniconda.sh.part
+    mv miniconda.sh.part miniconda.sh
+  fi
+  if [ ! -e "$HOME/miniconda" ]; then
```
felixcheung: shouldn't it check for `if [ ! -e "$HOME/miniconda" ]; then` before downloading it?

testing/install_external_dependencies.sh
```diff
@@ -35,15 +35,28 @@ fi
...
+  Darwin)
+    os=MacOSX
+    ;;
+  *)
+    os=Linux
+    ;;
+  esac
+  if [ ! -e miniconda.sh ]; then
+    wget "https://repo.continuum.io/miniconda/Miniconda${PYTHON}-4.2.12-$os-$(uname -m).sh" -O miniconda.sh.part
+    mv miniconda.sh.part miniconda.sh
```
felixcheung: why this?

testing/install_external_dependencies.sh
mephi42: Good idea, I will add this

testing/install_external_dependencies.sh
mephi42: In case download is Ctrl+C-ed and partial file remains.

testing/install_external_dependencies.sh
mephi42: Local tests didn't work without it (even though I have Spark installed). But I don't want to mess up Travis setup, so I will add a separate "if" statement for it.

jongyoul:
ping @mephi42

PR #2818, opened 2018-02-26
ZEPPELIN-3246. Need option for automatically restart the livy interpreter automatically as zeppelin does not start new Livy session if yarn livy session application is killed

zjffdu:
### What is this PR for?
Add one new property `zeppelin.livy.restart_dead_session` to allow livy session to be created automatically. By default it is false, because there's many reason that session is dead, it is encouraged to ask user to check why it is dead and restart interpreter by themselves. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3246

* Manually tested. 

image

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -93,6 +93,7 @@ public abstract class BaseLivyInterpreter extends Interpreter {
   private int sessionCreationTimeout;
   private int pullStatusInterval;
   protected boolean displayAppInfo;
+  private boolean restartDeadSession;
   protected LivyVersion livyVersion;
   private RestTemplate restTemplate;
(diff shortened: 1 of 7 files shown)


bikassaha:
Can this be specified on a per notebook basis using the %conf interpreter? It may be difficult to turn this on globally because of the various reasons why sessions may die.

Would it be better for the error message to ask the user to restart the interpreter from the notebook page itself?

zjffdu:
For whatever interpreter setting, user can set it via %conf.

When `zeppelin.livy.restart_dead_session` is turned on. dead livy session will be restarted automatically.
When it is turned off, proper message will be displayed in frontend, user can restart livy interpreter by himself after seeing this message.

bikassaha:
zeppelin.livy.restart_dead_session - Is this an interpreter level setting? Will it start only that users livy session?

zjffdu:
It would only restart that users livy session.

PR #2821, opened 2018-02-27
[ZEPPELIN-3271] Option for disabling scheduler

prabhjyotsingh:
### What is this PR for?
Zeppelin server should have an option to enable/disable cron scheduler from Zeppelin config.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3271](https://issues.apache.org/jira/browse/ZEPPELIN-3271)

* On adding below mentioned property in `zeppelin-site.xml`, this feature should get disabled from both web-socket and REST-APIs.

```
<property>
  <name>zeppelin.notebook.cron.enable</name>
  <value>false</value>
  <description>Notebook enable cron scheduler feature</description>
</property>
```

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -381,6 +381,7 @@ public class NotebookRestApi {
 
     note.setName(noteName);
     note.persist(subject);
+    note.setCronSupported(notebook.getConf());
     notebookServer.broadcastNote(note);
     notebookServer.broadcastNoteList(subject, SecurityUtils.getRoles());
     return new JsonResponse<>(Status.OK, "", note.getId()).build();
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -848,11 +848,9 @@ public class NotebookRestApi {
   public Response registerCronJob(@PathParam("noteId") String noteId, String message)
       throws IOException, IllegalArgumentException {
     LOG.info("Register cron job note={} request cron msg={}", noteId, message);
-    ZeppelinConfiguration conf = notebook.getConf();
-    if (conf.isZeppelinNotebookCronEnable()) {
+    Note note = notebook.getNote(noteId);
+    if (!(Boolean) note.getConfig().get("isZeppelinNotebookCronEnable")) {
       CronRequest request = CronRequest.fromJson(message);
-
-      Note note = notebook.getNote(noteId);
       checkIfNoteIsNotNull(note);
       checkIfUserCanRun(noteId, "Insufficient privileges you cannot set a cron job for this note");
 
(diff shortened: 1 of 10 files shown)


prabhjyotsingh:
@r-kamath @zjffdu  can you help review this PR.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -847,23 +848,27 @@ public Response stopParagraph(@PathParam("noteId") String noteId,
   public Response registerCronJob(@PathParam("noteId") String noteId, String message)
       throws IOException, IllegalArgumentException {
     LOG.info("Register cron job note={} request cron msg={}", noteId, message);
+    ZeppelinConfiguration conf = notebook.getConf();
+    if (conf.isZeppelinNotebookCronEnable()) {
```
felixcheung: we should log a warning I think if cron is disabled but someone is trying to register ?

mebelousov:
@prabhjyotsingh , I dreamed about this feature: [dev mailing list](https://lists.apache.org/thread.html/eb13b9dbd4a57c5e1e93297d7bde012e7f975bd2a0eb2a2a18c2d0ce@<dev.zeppelin.apache.org>)!

Now I think it would be useful define folders in which notes can be scheduled.
By default notes can be scheduled everywhere and the config parameter can limit scheduling to only special folders, for example "System/" or "Testing/". What do you think about such parameter?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -887,6 +887,11 @@ private void updateNote(NotebookSocket conn, HashSet<String> userAndRoles, Noteb
 
     Note note = notebook.getNote(noteId);
     if (note != null) {
+      if (!notebook.getConf().isZeppelinNotebookCronEnable()) {
+        if (config.get("cron") != null) {
+          config.remove("cron");
```
felixcheung: hmm, not sure - is this going to modify the notebook? even if cron is disabled it shouldn't modify the content permanently - if cron is enabled again the note should work as-is as before?

prabhjyotsingh: The `config` object is coming from WebSocket's `fromMessage`, over here I'm trying to put a check if someone tries to forcefully inject "cron" (or say WebServer is updated but UI is still running on an old version) then remove it (silently).

prabhjyotsingh:
@mebelousov Help me understand what you are saying, say if `zeppelin.notebook.cron.enable` (defined in this PR) is set to `true` and `zeppelin.notebook.cron.folders` is set to `System, Testing` then cron will only be run on these two folders.

My general opinion about cron is around either user should have permission or not, not at notebook level.

mebelousov:
It's sufficient to have only `zeppelin.notebook.cron.folders`.
If this parameter is set to `*` - scheduling can be in all folders. If `System/*, Testing/*` - only in those folders.
Thus if one will copy notes from production to another Zeppelin server and will change `zeppelin.notebook.cron.folders` to specific folder, he could test and use this server without removing of cron in notes.

prabhjyotsingh:
@mebelousov have pushed a commit with your suggested feedback.

zjffdu:
@prabhjyotsingh  Could we add any unit test for this feature ?

prabhjyotsingh:
@zjffdu  Sure added it.

prabhjyotsingh:
Have pushed another commit that disables Cron/scheduler by default, and user needs this feature, will need to with config.

zjffdu2:
@prabhjyotsingh Shouldn't it be turned on by default for backward compatibility ?

prabhjyotsingh:
Yes, I've updated the doc (https://github.com/apache/zeppelin/pull/2821/files#diff-bed1be4fb1f7f6c8e3bd6db18d72dc44R54). and yes, I agree we should document this in our release notes as well.

PR #2822, opened 2018-02-27
ZEPPELIN-3242. Listener threw an exception java.lang.NPEat o.a.zeppelin.spark.Utils.getNoteId(Utils.java:156)

zjffdu:
### What is this PR for?
This issue also cause spark url can not be displayed in frontend. The root cause is that PySparkInterpreter/IPySparkInterpreter doesn't set JobGroup correctly. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3242

* CI pass and also manually verified. 

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
@@ -23,6 +23,7 @@ import org.apache.zeppelin.interpreter.BaseZeppelinContext;
 import org.apache.zeppelin.interpreter.Interpreter;
 import org.apache.zeppelin.interpreter.InterpreterContext;
 import org.apache.zeppelin.interpreter.InterpreterException;
+import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.LazyOpenInterpreter;
 import org.apache.zeppelin.interpreter.WrappedInterpreter;
 import org.apache.zeppelin.python.IPythonInterpreter;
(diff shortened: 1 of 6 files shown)


zjffdu:
@Tagar  Could you help try this PR ? Thanks

Tagar:
thank you @zjffdu - I will give this a try today

spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
```diff
@@ -385,5 +396,7 @@ public void onUpdate(int index, InterpreterResultMessageOutput out) {
               }
             })
     );
+    context.setClient(mockRemoteEventClient);
```
Tagar: Is this a good idea to have a similar test for OldSparkInterpreter too?

zjffdu: OldSparkInterpreter already has this kind of testing (but doesn't use mock). https://github.com/apache/zeppelin/blob/master/spark/interpreter/src/test/java/org/apache/zeppelin/spark/OldSparkInterpreterTest.java#L340

zjffdu:
Thanks for testing @Tagar , will merge it soon

spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
```diff
@@ -98,6 +99,16 @@ public BaseZeppelinContext buildZeppelinContext() {
...
   }
 
+  @Override
+  public InterpreterResult interpret(String st, InterpreterContext context) {
+    InterpreterContext.set(context);
+    sparkInterpreter.populateSparkWebUrl(context);
+    String jobGroupId = Utils.buildJobGroupId(context);
+    String jobDesc = "Started by: " + Utils.getUserName(context.getAuthenticationInfo());
+    String setJobGroupStmt = "sc.setJobGroup('" +  jobGroupId + "', '" + jobDesc + "')";
+    return super.interpret(setJobGroupStmt +"\n" + st, context);
```
Tagar: @zjffdu This brought in a minor side effect described in https://issues.apache.org/jira/browse/ZEPPELIN-3484

PR #2823, opened 2018-03-01
ZEPPELIN-3273. SparkRInterpreter doesn't work in yarn mode

zjffdu:
### What is this PR for?
The root cause is the sparkr package is not distributed correctly. This PR correct the distributed file name and also add unit test for this. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3273

* Unit test added

### Questions:


zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/SparkInterpreterModeTest.java
```diff
@@ -137,6 +149,10 @@ public void testYarnClusterMode() throws IOException, YarnException, Interrupted
     interpreterSettingManager.close();
   }
 
+  private boolean isSpark2(String sparkVersion) {
+    return sparkVersion.startsWith("2.");
```
felixcheung: use SparkVersion.java?

zjffdu: SparkVersion.java is in spark module, can not be used in zeppelin-zengine, I don't want to make zeppelin-zengine depends on spark module just for this kind of trivial thing.

PR #2825, opened 2018-03-02
[ZEPPELIN-1967]: New pull request

sanjaydasgupta:
### What is this PR for?
The code in this PR enables embedding/interpolating Z variables into command strings passed to Spark's SQL and Shell interpreters. It implements the functionality described in issue [ZEPPELIN-1967](https://issues.apache.org/jira/browse/ZEPPELIN-1967)

This PR resumes a fresh effort while taking into consideration all of the discussion in the context of the earlier [PR-2502](https://github.com/apache/zeppelin/pull/2502). The earlier PR-2502 was closed due to a corruption in my repo that could not be corrected.

The code in this PR resolves all of the discussion and suggestions in the body of the earlier [PR-2502](https://github.com/apache/zeppelin/pull/2502). The following description is a summary of the current implementation: 

Patterns of the form `{var-name}` within commands will be interpolated only if a predefined object of the specified name exists in `z`. Additionally, **all** such patterns within the command line should also be translatable for any interpolation to occur. Partial translation of a command line (where only some of the patterns are translated, and others are not) is never performed.

Patterns of the form `{{any-text}}` are translated into `{any-text}`. 

These translations are performed only when all occurrences of {, }, {{, and }} in any command string confirm to one of the two forms described above. A command containing { and/or } characters used in any other way (than `{var-name}` and `{{any-text}}` described above) is used as-is without any changes. 
No error is flagged in any case. This behavior is identical to the implementation of a similar feature in 
Jupyter's shell invocation using the ! magic command.

At present only the SQL and Shell interpreters support object interpolation. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-1967

A new unit-test class ([ZeppCtxtVariableTest.java](https://github.com/apache/zeppelin/compare/master...sanjaydasgupta:master#diff-ad942153fbbad8ac1954b1063eabf920)) has been added. The attached screenshots also show tests of the functionality.

figure-1
figure-2
figure-3
figure-4
figure-5

### Questions:
* Does this needs documentation? Yes, detailed documentation has been added to the part describing ZeppelinContext variables.

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -365,6 +365,47 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
   </div>
 </div>
 
+### Object Interpolation
+Some interpreters can interpolate object values from `z` into the command string by using the 
+`{variable-name}` syntax. The value of any object previously `put` into `z` can be 
+interpolated into a command string by using such a pattern containing the object's name. 
+The following example shows one use of this facility:
+
+####In Scala cell:
+```
+z.put("minAge", 35)
+```
+
+####In later SQL cell:
+```
+%sql select * from members where age >= {minAge}
+```
+
+The interpolation of a `{var-name}` pattern is performed only when `z` contains an object with the specified name.
+But the pattern is left unchanged if the named object does not exist in `z`.
+Further, all `{var-name}` patterns within the command must must be translatable for any interpolation to occur -- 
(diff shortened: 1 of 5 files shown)


sanjaydasgupta:
Created from wrong branch in source repo. Need to correct branch and reissue.

PR #2834, opened 2018-03-05
[ZEPPELIN-1967] Passing Z variables to Shell and SQL Interpreters

sanjaydasgupta:
### What is this PR for?

The code in this PR enables embedding/interpolating Z variables into command strings passed to Spark's SQL and Shell interpreters. It implements the functionality described in issue [ZEPPELIN-1967](https://issues.apache.org/jira/browse/ZEPPELIN-1967)

This PR resumes a fresh effort while taking into consideration all of the discussion in the context of the earlier [PR-2502](https://github.com/apache/zeppelin/pull/2502). The earlier PR-2502 was closed due to a corruption in my repo that could not be corrected.

The code in this PR resolves all of the discussion and suggestions in the body of the earlier [PR-2502](https://github.com/apache/zeppelin/pull/2502). The following description is a summary of the current implementation:

Patterns of the form `{var-name}` within commands will be interpolated only if a predefined object of the specified name exists in `z`. Additionally, all such patterns within the command line should also be translatable for any interpolation to occur. Partial translation of a command line (where only some of the patterns are translated, and others are not) is never performed.

Patterns of the form `{{any-text}}` are translated into `{any-text}`. This feature is an escaping mechanism that allows `{` and `}` characters to be passed into a command without invoking the interpolation mechanism.

The translations described above are performed only when all occurrences of `{`, `}`, `{{`, and `}}` in any command string conform to one of the two forms described above. A command that contains `{` and/or `}` characters used in any other way (than `{var-name}` and `{{any-text}}` as described above) will be used as-is without attempting any translations whatsoever -- even if the command also contains legal, translate-able and/or escape-able, constructs of the above two forms.

No error is flagged in any case. This behavior is identical to the implementation of a similar feature in
Jupyter's shell invocation using the ! magic command.

At present only the SQL and Shell interpreters support object interpolation.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-1967

A new unit-test class ZeppCtxtVariableTest.java (see below) has been added. The attached screenshots below also show tests of the functionality.

figure-1
figure-2
figure-3
figure-4
figure-5

### Questions:
* Does this needs documentation? Yes, and detailed documentation has been added to the part describing ZeppelinContext variables (see file spark.md below).


zeppelin-web/package-lock.json
```diff
@@ -3669,910 +3668,6 @@
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8=",
       "dev": true
     },
-    "fsevents": {
```
Tagar: Is this package-lock.json change related to ZEPPELIN-1967 or it's something else?

Tagar:
Great improvement! Many of our users were asking for this since day one they started using Zeppelin.
Like, "how do I pass a variable from %pyspark paragraph to %sh one.. "

Left a minor comment there - but wouldn't be able to do a thorough code review here.
Hopefully one of @felixcheung @zjffdu @Leemoonsoo can review this soon?
Thanks!!

zeppelin-web/package-lock.json
sanjaydasgupta: Thanks @Tagar -  this was indeed an unintended change that crept in by mistake. 

I have reversed the changes out of the PR now.

Tagar:
What happens when a notebook had code {something} and "something" doesn't exist as a 'z' variable?

Would it be possible to review that comment I left earlier in 
https://github.com/apache/zeppelin/pull/2502#issuecomment-324095161

Should we have a Zeppelin option that controls this behavior?

> Option 1 if used - it'll completely follow Jupyter bahavior (no error, {..} block return unchanged).
> Option 2 - fail explicitly if variable in curly braces can't be found.
> That being said, I think most of folks may actually prefer option 1, as introduction of { .. } may be a breaking change for some notebooks' code. So I think Option 1 should be default.

Because this change may break code in some users' notebooks (if they had curly braces), it would be nice to let users choose if they want to use Option 1 (so their code will be compatible / not broken by this PR), or Option 2 which would tell them explicitly that variable isn't found.

sanjaydasgupta:
Here is one approach: define two new [zeppelin configurable properties](http://zeppelin.apache.org/docs/latest/install/configuration.html) as follows:

1) A boolean property named _zeppelin.interpreter.interpolation_ which must be set to _true_ to enable this feature at all. The default can be _false_, to guarantee that unaware users will never receive rude shocks.
2) Another boolean property named _zeppelin.interpreter.interpolation.silent_ which when set to _true_ causes the behavior in your _Option 1_, and when set to _false_ causes the behavior in _Option 2_. The default can be _false_ to enable a gentle transition into the feature.

I am thinking if we need (2) at all, since (1) alone is enough for allowing users to start using the new feature in a safe and controlled way.

Please let me know what you think. If you have a simpler or more effective approach please let me know.

zjffdu:
Thanks for the contribution. Regarding the interpolation property, I think it might be better to use paragraph level property instead of interpreter level property.  e.g. here's what I imagine
```
%sql(interpolation=true)
select * from table where length > ${min_length}
```

2 reasons I introduce paragraph level property
* interpolation has overhead, we should avoid it as possible as we can. 
* paragraph level property is more readable for users that once they notice this paragraph level property, it would be easy for them to know that there's interporpeter in code. In contrast, few users would notice the interpreter level setting. 

Will back to review more after 0.8 release.

Tagar:
@zjffdu Great idea on paragraph-level properties! There are so many possible uses for this feature like ones you explained below.

For interpolation I think there should be a way to do it globally too, to match Jupyter's behavior. And it could be also optionally enabled at paragraph level like you're describing. Not sure I understood on overhead though - since it will be only applied to code in the paragraphs.

@sanjaydasgupta 
> I am thinking if we need (2) at all, since (1) alone is enough for allowing users to start using the new feature in a safe and controlled way.

I agree, thank you.

@zjffdu, what @sanjaydasgupta is saying allows to switch this feature off at global level, so we can maintain compatibility with previous version, and only switch interpolation when users choose to enable it. It looks like a good idea to me. We could add enabling this at paragraph-level as part of another PR later, when this new facility to have paragraph-level properties will be available. What do you guys think?

zjffdu:
Right, it is 2 level of property (global level and paragraph level). paragraph level can override global level. I think the paragraph level setting would be in 0.9 as 0.8 is close to release, it is a little risky to add new feature now.

sanjaydasgupta:
I have added code, documentation, and configuration-template files for a single environment-variable named `ZEPPELIN_INTERPRETER_INTERPOLATION` (and a corresponding java-property `zeppelin.interpreter.interpolation`) for globally enabling/disabling interpolation.

The feature is disabled by default, guaranteeing backward-compatibility for existing code, while allowing users who want the feature to switch it on when needed. 

As suggested by @zjffdu above, more fine-grained configuration can be added in future when the paragraph-level configuration becomes available in 0.9.0.

conf/zeppelin-env.sh.template
```diff
@@ -48,6 +48,7 @@
 
 # export ZEPPELIN_IDENT_STRING   		# A string representing this instance of zeppelin. $USER by default.
 # export ZEPPELIN_NICENESS       		# The scheduling priority for daemons. Defaults to 0.
+# export ZEPPELIN_INTERPRETER_INTERPOLATION         # Enable ZeppelinContext variable interpolation into command string (FALSE by default)
```
felixcheung: can you align the comment `#`?

docs/interpreter/spark.md
```diff
@@ -365,6 +365,51 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
   </div>
 </div>
 
+### Object Interpolation
+Some interpreters can interpolate object values from `z` into the command string by using the
```
felixcheung: question: perhaps this is talked about but the meaning of "command string" isn't very clear to me as an user.
Are we referring to the paragraph text?

docs/interpreter/spark.md
```diff
@@ -365,6 +365,51 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
...
+forms described above. A command containing `{` and/or `}` characters used in any other way 
+(than `{var-name}` and `{{any-text}}`) is used as-is without any changes. 
+No error is flagged in any case. This behavior is identical to the implementation of a similar feature in 
+Jupyter's shell invocation using the `!` magic command.
+
+This feature is disabled by default, and must be explicitly turned on by setting either the environment variable 
+`ZEPPELIN_INTERPRETER_INTERPOLATION` or the Java property `zeppelin.interpreter.interpolation` to `true`. 
+As usual, if both are defined, the environment variable's setting has priority.
+
+At present only the SQL and Shell interpreters support object interpolation.
```
felixcheung: should this go to a different doc page? Shell is not part of spark, and sql can be other sql interpreters like hive or postgresql?

conf/zeppelin-env.sh.template
sanjaydasgupta: Done

docs/interpreter/spark.md
sanjaydasgupta: The following changes have been made:

1) added a short description of this feature in the documentation for the shell interpreter.
2) changed `SQL` here to `SparkSQL`, but have not removed the reference to shell (see explanation below) 

The reason why the description of interpolation is here is that it is one of the features of `ZeppelinContext` - which was already described here before I added the interpolation part. Perhaps in a future release we should reorganize this document (spark.md), and remove `ZeppelinContext` into a different place in the documentation.

docs/interpreter/spark.md
sanjaydasgupta: "command" and "command string" used everywhere in my original documentation has been changed to "paragraph text".

sanjaydasgupta:
Thanks for your review @felixcheung 

The 3 points have been addressed as described individually above.

docs/interpreter/spark.md
felixcheung: ok

Tagar:
@zjffdu, @felixcheung would be exciting to have this feature be part of the 0.8.0 release.
Since there is a way to maintain backward compatibility, it seems to be a low hanging fruit 
as it's a great improvement in Zeppelin functionality. Thanks @sanjaydasgupta and everyone!

sanjaydasgupta:
Hello reviewers, this PR has accumulated many ideas from the wisdom of its past and present reviewers (@zjffdu @felixcheung @Tagar @Leemoonsoo ), and it would be interesting to see how the user community leverages it. Requesting you to consider adding it to release 0.8.0.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -780,6 +780,7 @@ public String getZeppelinNotebookCronFolders() {
         "zeppelin.interpreter.lifecyclemanager.timeout.checkinterval", 6000L),
     ZEPPELIN_INTERPRETER_LIFECYCLE_MANAGER_TIMEOUT_THRESHOLD(
         "zeppelin.interpreter.lifecyclemanager.timeout.threshold", 3600000L),
+    ZEPPELIN_INTERPRETER_INTERPOLATION("zeppelin.interpreter.interpolation", false),
```
zjffdu: It it not necessary to put it in ZeppelinConfiguration & zeppelin-env.sh. Properties in them are global properties, since it is interpreter level property, we only need to put it in interpreter-setting.json

zjffdu:
@sanjaydasgupta Overall It LGTM, just left one minor comment.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
sanjaydasgupta: Thanks @zjffdu that was very helpful. 

The following changes have been made:
1) Removed global property definitions from _ZeppelinConfiguration_, _zeppelin-env.sh_ template, and _zeppelin-site.xml_ template.
2) Also removed values of _env-name_ and _property-name_ (replaced both with `null`) from both _interpreter-setting.json_ files.
3) Renamed the two properties to `zeppelin.spark.sql.interpolation` (for SparkSQL interpreter) and `zeppelin.shell.interpolation` (for Shell interpreter) 
4) Updated documentation files for _configuration_, _spark_, and _shell_.

sanjaydasgupta:
@zjffdu Thanks for your review - that was very helpful.

Have fixed all the issues - please see [details added after your comment](https://github.com/apache/zeppelin/pull/2834/files/1e2c87dd36dc091ca898baf8e9f178d6d1a5e600#r176944574).

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/ZeppCtxtVariableTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @Test
+  public void randomBracesThree() {
+    String result = interpreter.interpolate("Paired { begin an escaped sequence", resourcePool);
+    assertTrue("Random braces - three", "Paired { begin an escaped sequence".equals(result));
+  }
+
+  @Test
+  public void randomBracesFour() {
+    String result = interpreter.interpolate("Paired } end an escaped sequence", resourcePool);
```
felixcheung: could you add some tests for multiline string - and paired/unpaired  `{` and `{{` or `}` `}}`

docs/interpreter/shell.md
```diff
@@ -82,3 +87,8 @@ export LAUNCH_KERBEROS_REFRESH_INTERVAL=4h
 # Change kinit number retries (default value is 5), which means if the kinit command fails for 5 retries consecutively it will close the interpreter. 
 export KINIT_FAIL_THRESHOLD=10
 ```
+
+## Object Interpolation
+The shell interpreter also supports interpolation of `ZeppelinContext` objects into the paragraph text.
+Details of this feature can be found in the Spark interpreter documentation under 
+[Object Interpolation](spark.html#object-interpolation)
```
felixcheung: since this points to spark doc, perhaps add a quick example here with shell?

felixcheung: actually, let's have to example in spark because..

docs/interpreter/spark.md
```diff
@@ -365,6 +370,52 @@ myScalaDataFrame = DataFrame(z.get("myScalaDataFrame"), sqlContext)
...
+(than `{var-name}` and `{{any-text}}`) is used as-is without any changes. 
+No error is flagged in any case. This behavior is identical to the implementation of a similar feature in 
+Jupyter's shell invocation using the `!` magic command.
+
+This feature is disabled by default, and must be explicitly turned on for each interpreter by setting the value of 
+an interpreter-specific property to `true`. Consult the configuration section of each interpreter's documentation 
+to find out if object interpolation is implemented, and the name of the parameter that must be set to `true` to 
+enable the feature.
+
+At present only the SparkSQL and Shell interpreters support object interpolation.
```
felixcheung: so can the user uses the same variable in sql and then in shell?
let's add an example here?

shell/src/main/resources/interpreter-setting.json
```diff
@@ -38,6 +38,13 @@
         "defaultValue": "",
         "description": "Kerberos principal",
         "type": "string"
+      },
+      "zeppelin.shell.interpolation": {
+        "envName": null,
+        "propertyName": null,
```
felixcheung: doesn't this need a propertyName?

spark/interpreter/src/main/resources/interpreter-setting.json
```diff
@@ -108,6 +108,13 @@
         "description": "Show full exception stacktrace for SQL queries if set to true.",
         "type": "checkbox"
       },
+      "zeppelin.spark.sql.interpolation": {
+        "envName": null,
+        "propertyName": null,
```
felixcheung: ditto

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/ZeppCtxtVariableTest.java
felixcheung: eg.
```
this is a {
test}
```

or

```
this is a
test }}
```

zjffdu:
Yes, it is safe for me merge it into branch-0.8

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/ZeppCtxtVariableTest.java
sanjaydasgupta: Have expanded the set of tests for multi-line - both success and failures. There are now 18 unit tests.

docs/interpreter/shell.md
sanjaydasgupta: I have expanded and clarified the description of the feature here, and hope it is not confusing for readers now. Also explained that object interpolation is implemented per-interpreter, can be enabled/disabled independently per-interpreter, and has a different parameter name for enabling the feature in each interpreter. The parameter name is documented in the _Configuration_ part of both the implementing interpreters (SparkSQL, and Shell)

docs/interpreter/spark.md
sanjaydasgupta: No, the parameter name is different for each interpreter as the implementation is per-interpreter, and the feature can be enabled/disabled independently for each interpreter. Have expanded the description and have included the parameter names for the two interpreters.

shell/src/main/resources/interpreter-setting.json
sanjaydasgupta: Since this parameter is interpreter-level, not global, the _envName_ and _propertyName_ are not needed. I think the _envName_ is for values to be parsed from the file _zeppelin-env.sh_, and the _propertyName_ is for values to be parsed from the file _zeppelin-site.xml_

But request @zjffdu to confirm my assumptions above.

spark/interpreter/src/main/resources/interpreter-setting.json
sanjaydasgupta: Refer to my response immediately above

docs/interpreter/spark.md
sanjaydasgupta: In your question, if you meant _Z-variable_ name, then the answer is *yes*. There is an example added in both document files (for SparkSQL and Shell) illustrating this now.

felixcheung: yes, for variable

shell/src/main/resources/interpreter-setting.json
felixcheung: envName yes, but I think it should have propertyName? it's at least consistent to all other properties.
I'll defer to @zjffdu though

zjffdu: envName is not necessary. It is not used by zeppelin-env.sh. Actually it is never used, I planned to remove it from interpreter-setting.json.  BTW the structure of `interpreter-setting.json` (property of each interpreter) and `interpreter.json` (all the interpreter saved configuration) is not consistent, we may need to redesign the structure to make them consistent.

sanjaydasgupta: @felixcheung the _propertyName_ field in both _interpreter-setting.json_ files have been restored. Thanks for the catch.

@zjffdu thanks for the clarification.

I read the code and documents for the interpreter configuration area, and believe both need a closer look and many improvements are possible. If you are planning an effort for this after 0.8.0, I would love to help. My strengths are mainly Scala and Java, and I have also written professionally before.

sanjaydasgupta:
@felixcheung  following [the earlier discussion](https://github.com/apache/zeppelin/pull/2834#discussion_r176976263), the _propertyName_ field in both _interpreter-setting.json_ files have been restored. Thanks for the catch.

zjffdu:
Thanks for the contribution @sanjaydasgupta will merge if no more comments

PR #2836, opened 2018-03-06
ZEPPELIN-3291. Throw exception instead of return null for InterpreterFactory.getInterpreter

zjffdu:
### What is this PR for?
Minor code refactoring to throw exception when no interpreter is found

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3291

* CI pass

### Questions:


felixcheung:
@Leemoonsoo

PR #2837, opened 2018-03-06
ZEPPELIN-3279. [FlakyTest] NotebookTest.testPerSessionInterpreterCloseOnUnbindInterpreterSetting

zjffdu:
### What is this PR for?
The root cause of this flaky test is that hashCode doesn't represent an unique id of object. This PR just use IdGenerator for unique id and also combine process id. 
This PR also fix the flaky test caused by matplotlib. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3279

* CI Pass

### Questions:

### DIFF
--- a/testing/install_external_dependencies.sh
+++ b/testing/install_external_dependencies.sh
@@ -44,6 +44,6 @@ if [[ -n "$PYTHON" ]] ; then
   conda update -q conda
   conda info -a
   conda config --add channels conda-forge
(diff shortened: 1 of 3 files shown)


felixcheung:
@Leemoonsoo

PR #2843, opened 2018-03-07
ZEPPELIN-3299. Combine spark integration test with its unit test in one build

zjffdu:
### What is this PR for?

In ZEPPELIN-3254, we introduce 2 build for spark integration test. And we can run unit test with these system test together so that we could save 2 build in travis and also save time to set up enviroment. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3299

* Travis CI Pass

### Questions:


felixcheung:
@Leemoonsoo

PR #2847, opened 2018-03-07
ZEPPELIN-3303. Avoid download spark for travis build

zjffdu:
### What is this PR for?

It is not necessary to download spark distribution for several travis build. This PR would remove that to save time.

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3303

* CI pass

### Questions:


felixcheung:
@Leemoonsoo

PR #2848, opened 2018-03-07
[Zeppelin-3307] - Improved shared browsing/editing for the note

Savalek:
### What is this PR for?
Now if the note is opened in several tabs (or several users are watching or editing it), then there may be problems. Loss of code entered by the user, reset the cursor position.
This PR adds a basic opportunity for collaborative editing. For the organization of joint editing, the library diff-match-patch is used.
PR does not change the logic of operation if the note is used by one person.
Also, maybe this will solve the problem with [ZEPPELIN-3131](https://issues.apache.org/jira/browse/ZEPPELIN-3131).

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-3307](https://issues.apache.org/jira/browse/ZEPPELIN-3307)

gif

### Questions:


Savalek:
@prabhjyotsingh @r-kamath
could you help with the review?

mebelousov:
Zeppelin has websockets and two users can view a note simultaneously, but there is a bug.
- Open the note in two tabs.
- Write anything in first tab for second.
- Wait for 10 seconds.
- Repeat the last two steps several times.
- Check that written code is lost.

Thus paired coding is painful.
Please review this PR.

jongyoul:
I think this PR might be not bad for a small number of users, but, I'm not sure users increase. What do you think of it? It looks like calling websockets everytime whenever all users make changes

Savalek:
@jongyoul, yes sending any change to the server is a big load. But this will happen only if the notebook is opened by several users. The current implementation is subject to errors in consequence of which users may lose some of the code. Even if the second user just open the tab in the browser. This is very annoying.

zjffdu:
@prabhjyotsingh @r-kamath Could you help take a look at it ? It seems a necessary feature.

jongyoul:
@Savalek I agree with unexpected ignorances feel bad. Can you add test cases both of backend and frontend? And please make sure the relationship with `personalized mode`.

This is an idea but needs to think of how we improve this feature? If this mode is disabled, all writers can write in the same way? otherwise, prohibit it except owner? When `personalized mode` is enabled, how does this feature behave?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1284,6 +1306,42 @@ private void updateParagraph(NotebookSocket conn, HashSet<String> userAndRoles,
...
+    DiffMatchPatch dmp = new DiffMatchPatch();
+    String patchText = (String) fromMessage.get("patch");
+
+    // javascript add "," if change contains several patches
+    patchText = patchText.replace(",@@", "@@");
+    LinkedList<DiffMatchPatch.Patch> patches =
+        (LinkedList<DiffMatchPatch.Patch>) dmp.patchFromText(patchText);
+
+    String paragraphText = p.getText();
+    paragraphText = (String) dmp.patchApply(patches, paragraphText)[0];
```
jongyoul: This, too.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1284,6 +1306,42 @@ private void updateParagraph(NotebookSocket conn, HashSet<String> userAndRoles,
...
+
+    final Note note = notebook.getNote(noteId);
+    Paragraph p = note.getParagraph(paragraphId);
+
+    DiffMatchPatch dmp = new DiffMatchPatch();
+    String patchText = (String) fromMessage.get("patch");
+
+    // javascript add "," if change contains several patches
+    patchText = patchText.replace(",@@", "@@");
+    LinkedList<DiffMatchPatch.Patch> patches =
```
jongyoul: `dm.patchFromText()` returns `List<DiffMatchPathc.Path>` type. Is there any specific reason to use `LinkedList`?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -702,9 +705,24 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
...
     if ($scope.dirtyText !== $scope.originalText) {
-      $scope.startSaveTimer();
+      if ($scope.collaborativeMode) {
+        $scope.sendPatch();
+      } else {
+        $scope.startSaveTimer();
+      }
     }
     setParagraphMode(session, dirtyText, editor.getCursorPosition());
+    if ($scope.cursorPosition && $scope.cursorPosition && $scope.cursorPosition) {
```
jongyoul: I don't know well about front, but what is this for?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1263,6 +1282,11 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
       $route.current.pathParams.noteId);
   };
 
+  const patchParagraph = function(paragraph, patch) {
+    const id = paragraph.id;
```
jongyoul: It looks unnecessary as `id` is used only once. and this function is only used inside `sendPatch`. Isn't it ok to merge this function into that?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1562,6 +1586,24 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
...
+      if (!$scope.paragraph.text || $scope.paragraph.text === undefined) {
+        $scope.paragraph.text = '';
+      }
+      $scope.paragraph.text = $scope.diffMatchPatch.patch_apply(patch, $scope.paragraph.text)[0];
+      $scope.originalText = angular.copy($scope.paragraph.text);
+      let newPosition = $scope.editor.getCursorPosition();
+      if (newPosition && newPosition.row && newPosition.column) {
+        $scope.cursorPosition = $scope.editor.getCursorPosition();
+      } else {
+        console.log('NOOOOOOO Save');
```
jongyoul: Can you change the level of the log with a specific comment?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
Savalek: patchApply(LinkedList<DiffMatchPatch.Patch> patches, String text)
First argunment requires LinkedList<DiffMatchPatch.Patch>

Savalek:
@jongyoul, I have corrected.

r-kamath:
@Savalek Thanks for this new feature. IMHO it will be nice to have an option to turn on/off the collaborative mode and a way to show the number of users collaborating.
something like users with a proper tooltip.

Savalek:
@r-kamath, what do you mean by turn off/on option? 
This option will be global and enable/disable will only be available for the server administrator?

Savalek:
@jongyoul, I do not like it when `try {...} catch {...}` is used many times in the code. Therefore, I slightly changed the handling of `ClassCastException`. Could you look at the new implementation and say what you think about it?

zeppelin-web/src/app/notebook/notebook-actionBar.html
```diff
@@ -248,6 +248,16 @@ <h3>
         </button>
       </span>
 
+      <span class="labelBtn" style="vertical-align:middle; display:inline-block;">
+        <button type="button"
+                class="btn btn-default btn-xs"
+                ng-show="collaborativeMode"
+                tooltip-placement="bottom" uib-tooltip="Users who watch this note: {{collaborativeModeUsers.join(', ')}}"
```
r-kamath: 👍

jongyoul:
@Savalek Thanks!! That’s what I want to do exactly. I also don’t like to use `try` and `catch` several times.

Savalek:
@jongyoul, I added tests.

conf/zeppelin-site.xml.template
```diff
@@ -67,6 +67,12 @@
   <description>hide homescreen notebook from list when this value set to true</description>
 </property>
 
+<property>
+  <name>zeppelin.notebook.collaborative.mode.enable</name>
+  <value>true</value>
+  <description>Enable basic opportunity for collaborative editing. Does not change the logic of operation if the note is used by one person.</description>
```
r-kamath: MINOR: Ideal to have a long description in the documentation. "Enable collaborative mode feature" is should be sufficient here.

jongyoul:
@Savalek Thanks for adding e2e test but can you add also functional test for your backend code? It would be better to make a mock for `NotebookServer` to check if `broadcast` is called properly.

Savalek:
@r-kamath, @prabhjyotsingh, i fixed the description

Savalek:
@jongyoul, I added tests for backend.

jongyoul:
@Savalek BTW, https://travis-ci.org/Savalek/zeppelin/builds/380644746 Can you check your Travis? Zeppelin has a lot of tricky tests but, in your case, it's a bit different. Do you know the reason?

Savalek:
@jongyoul, I run two tests. "master" and "zeppelin-3307 merge master." Travis shows the same result. Errors in this tests identical.

travis

jongyoul:
Will merge it if there's no further discussion

Savalek:
Fixed conflicts. What about merge?

jongyoul:
This feature could replace "personalized mode". I will create PR to remove "personalized mode". Thank you @Savalek again.

mebelousov:
@jongyoul As I understand personal mode allows users run paragraphs and have different views and different results due user chosen values in dynamic forms.
I'm against the removal of personal mode.

jongyoul:
@mebelousov The main purpose of "personalized mode" is to keep the current user's view. But if the user refreshes the browser, it changes the newest one. Do you think it's enough?

mebelousov:
@jongyoul I see next usecase. 10 users open the note, put client ID in dynamic form, refresh note, get and process data  result. In this case getting of default note view is OK.
That is in personal mode we may not save note updates.

jongyoul:
Git it. Thanks. I thought it might not be useful. but I might be wrong. Thank you.

PR #2850, opened 2018-03-08
ZEPPELIN-3099 Livy Interpreter doesn't support German Special Character - Encoding Issue

Akhilsnaik:
### What is this PR for?

ZEPPELIN-3099 Livy Interpreter doesn't support German Special Character - Encoding Issue

### What type of PR is it?
BUG FIX for : ZEPPELIN-3099 
### What is the Jira issue?
**https://issues.apache.org/jira/browse/ZEPPELIN-3099**
test with German Special Characters like ß, ü, Ü, ä, Ä, ö, Ö .

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -19,6 +19,7 @@ package org.apache.zeppelin.livy;
 
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.nio.charset.Charset;
 import java.security.KeyStore;
 import java.security.Principal;
 import java.util.ArrayList;
@@ -68,6 +69,7 @@ import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.http.converter.StringHttpMessageConverter;
 import org.springframework.security.kerberos.client.KerberosRestTemplate;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.HttpServerErrorException;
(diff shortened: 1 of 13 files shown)


Akhilsnaik:
@prabhjyotsingh  @zjffdu 

Can you pleaes help me review this change

zjffdu:
@Akhilsnaik  Could you add test for it ? You can add it in this method, https://github.com/apache/zeppelin/blob/master/livy/src/test/java/org/apache/zeppelin/livy/LivyInterpreterIT.java#L128

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -118,6 +120,7 @@ public BaseLivyInterpreter(Properties property) {
     this.pullStatusInterval = Integer.parseInt(
         property.getProperty("zeppelin.livy.pull_status.interval.millis", 1000 + ""));
     this.restTemplate = createRestTemplate();
+    this.restTemplate.getMessageConverters().add(0, new StringHttpMessageConverter(Charset.forName("UTF-8")));
```
zjffdu: It is better to this in method `createRestTemplate`

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -653,18 +655,24 @@ public Principal getUserPrincipal() {
...
       }
     }
     if (httpClient == null) {
-      return new RestTemplate();
+      restTemplate = new RestTemplate();
     } else {
-      return new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient));
+      restTemplate = new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient));
     }
+    if (null != restTemplate) {
```
zjffdu: What case would restTemplate be null ?

Akhilsnaik: it cannot be null in any scenario , but i made a null check because the variable restTemplate is defined as null and i dont want a nullpointerException to be thrown in this part of code anytime near future .

PR #2852, opened 2018-03-08
[ZEPPELIN-3264] Notebook Snapshot feature.

SleepyThread:
### What is this PR for?
Currently, a user can share the notebook with other users. There is a usage pattern where data analyst creates notebook report and then share with other stakeholders. Meanwhile, they would like to update/change the report and re-publish the results.  

We need a shareable url like,
 /notebook/{**Notebook ID**}/view/{**Snapshot Name**}

which can be shared to other users and can be updated with a logical sharable report. 

### What type of PR is it?
[ Feature | Documentation]

### Todos
* [ ] - CSS and assets changes. 

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3264

* Create git based Notebook. 
* Create 1st commit. 
* Add a snapshot ID. 
* View the snapshot. 

create_snapshot

### Questions:
 - No
 - No.
 - Yes. Already updated.


SleepyThread:
@Leemoonsoo @zjffdu friendly reminder regarding this pull request.

zjffdu:
> Meanwhile, they would like to update/change the report and re-publish the results.

If they update/change the report, should they create another snapshot ?

Here's my understanding of this scenario. The snapshot is only readable, and user can only see this report. (maybe they can update it, but only via dynamic forms, so the paragraph text is readable, only the result can be changed).

If the above scenario is reasonable,then I think we can just create git tag for snapshot instead of saving the snapshot into file. And only retrieve that tag from git repo when user open that snapshot.

mebelousov:
Will the git storage allow to delete older snapshots?
I see the scenario: every day create new snapshot and delete the oldest snapshot.

SleepyThread:
@zjffdu Implementation using Git Tag is also another approach but updating/delete Git tag regularly is not something feels right [ Usually in practice Git Tag are to be done only once ]. Hence, I have used file-based approach for storing a mapping of Snapshot ID => Revision ID.

Let me know if you feel strongly about the other approach, and I work on the Git Tag based implementation.

zjffdu:
>  Git tag regularly is not something feels right [ Usually in practice Git Tag are to be done only once ].

That is my understanding of the snapshot scenario. Snapshot is only readable, but not writable. When user view snapshot of one note, zeppelin just fetch the note at that tag. Let me know if my understanding of your snapshot scenario is not correct.

jongyoul:
Can someone tell me the status of this issue?

PR #2858, opened 2018-03-12
ZEPPELIN-3315. Merge beforeStatusChange and afterStatusChange to onStatusChange

zjffdu:
### What is this PR for?
The signature of `beforeStatusChange` & `afterStatusChange` include both the status of before and after, so it is not necessary to create both `beforeStatusChange` & `afterStatusChange`. Only one method `onStatusChange` is sufficient. 

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3315

* CI pass

### Questions:


zjffdu:
And reading the code snippet below, I don't see any senacairo we need to have 2 methods in `setStatus` of Job.java. Most of the implementation just left `beforeStatusChange` empty. 

https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java#L156

PR #2861, opened 2018-03-13
ZEPPELIN-3322. Update interpreterBind when restarting zeppelin server

zjffdu:
### What is this PR for?
When loading interpreter.json, also update interpreterBindings as we change change interpreter setting id in ZEPPELIN-3208. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3322

* Unit test is added. 

### Questions:


felixcheung:
@Leemoonsoo

PR #2871, opened 2018-03-15
ZEPPELIN-3337. Add more test to SparkRInterpreter

zjffdu:
### What is this PR for?
Add more test for SparkRInterpreter, and also some code refactoring in SparkRInterpreter.  Also fix one bug of SparkRInterpreter that it can be cancelled. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3337

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
@@ -46,8 +46,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class SparkRInterpreter extends Interpreter {
   private static final Logger logger = LoggerFactory.getLogger(SparkRInterpreter.class);
 
-  private static String renderOptions;
+  private String renderOptions;
   private SparkInterpreter sparkInterpreter;
+  private boolean isSpark2;
   private ZeppelinR zeppelinR;
   private AtomicBoolean rbackendDead = new AtomicBoolean(false);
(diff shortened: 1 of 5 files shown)


felixcheung:
@Leemoonsoo

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
```diff
@@ -101,37 +104,28 @@ public void open() throws InterpreterException {
...
     if (useKnitr()) {
       zeppelinR.eval("library('knitr')");
     }
-    renderOptions = getProperty("zeppelin.R.render.options");
-  }
-
-  String getJobGroup(InterpreterContext context){
-    return "zeppelin-" + context.getParagraphId();
+    renderOptions = getProperty("zeppelin.R.render.options",
+        "out.format = 'html', comment = NA, echo = FALSE, results = 'asis', message = F, " +
```
felixcheung: how do we find this default value?

zjffdu: The default value is in interpreter-setting.json. It it better to load default properties instead of specify it explicitly like this (there's many such code in zeppelin). This could be done in another PR in future.

felixcheung:
Agreed. This is a bit uncomfortablely long to be hard coded.

PR #2875, opened 2018-03-16
ZEPPELIN-3343. Interpreter Hook is broken

zjffdu:
### What is this PR for?
Currently the interpreter hook mechanism is broken. This PR fix this issue and also add unit test for interpreter hook, (including global hook and note level hook)

### What type of PR is it?
[Bug Fix | Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3343

* CI pass

### Questions:

### DIFF
--- a/python/src/main/resources/grpc/python/zeppelin_python.py
+++ b/python/src/main/resources/grpc/python/zeppelin_python.py
@@ -113,6 +113,29 @@ class PyZeppelinContext(object):
     #)
     body_buf.close(); header_buf.close()
 
+  def registerHook(self, event, cmd, replName=None):
+    if replName is None:
+      self.z.registerHook(event, cmd)
+    else:
+      self.z.registerHook(event, cmd, replName)
+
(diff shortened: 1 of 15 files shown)


python/src/main/resources/grpc/python/zeppelin_python.py
```diff
@@ -113,6 +113,29 @@ def show_dataframe(self, df, show_index=False, **kwargs):
     #)
     body_buf.close(); header_buf.close()
 
+  def registerHook(self, event, cmd, replName=None):
```
felixcheung: hmm, I think we have doc on the python z context API, let's make sure we have them

felixcheung: also new api is better to be in a separate PR

felixcheung:
@Leemoonsoo

PR #2878, opened 2018-03-17
ZEPPELIN-3331. Use NullLifecycleManager by default

zjffdu:
### What is this PR for?

Just keep the behavior consistent with previous version. So use NullLifecycleManager by default, user need to configure zeppelin-site.xml to enable TimeoutLifecycleManager

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3331

* CI pass

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -434,6 +434,7 @@
   <description>Enable directory listings on server.</description>
 </property>
 
+<!--
 <property>
   <name>zeppelin.interpreter.lifecyclemanager.class</name>
   <value>org.apache.zeppelin.interpreter.lifecycle.TimeoutLifecycleManager</value>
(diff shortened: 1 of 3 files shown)


felixcheung:
@Leemoonsoo

jturcotte:
The documentation is wrong though:

> For now, there are two implementations: NullLifecycleManager and TimeoutLifecycleManager which is the default. 

I ended up here trying to figure out why it didn't work when I looked at the zeppelin-site.xml without syntax highlighting (I didn't notice that this section was commented out).

Reamer:
It would be nice if you could open a PR that would solve this documentation problem.
Your changes should be made in https://github.com/apache/zeppelin/blob/master/docs/usage/interpreter/overview.md.

jturcotte:
Sure, here is a quick and dirty MR using GitHub's web editor: https://github.com/apache/zeppelin/pull/3962

PR #2881, opened 2018-03-19
ZEPPELIN-3313 z.getInterpreterContext().out().clear() fails to clear table data

r-kamath:
### What is this PR for?
z.getInterpreterContext().out().clear() fails to clear table data

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Fix visualizations in a separate ticket

### What is the Jira issue?
[ZEPPELIN-3313](https://issues.apache.org/jira/browse/ZEPPELIN-3313)

see jira description

#### before
before

#### after
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-table.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-table.js
@@ -288,6 +288,51 @@ export default class TableVisualization extends Visualization {
 
     const config = this.config;
     const self = this; // for closure
+    const scope = this.getScope();
+    // set gridApi for this elem
+    const gridApiId = this.getGridApiId();
+    const gridOptions = this.createGridOptions(tableData, onRegisterApiCallback, config);
+
+    const onRegisterApiCallback = (gridApi) => {
+      scope[gridApiId] = gridApi;
+      // should restore state before registering APIs
+
+      // register callbacks for change evens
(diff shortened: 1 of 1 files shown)


zjffdu:
Thanks @r-kamath , I try another example, but the result seems unstable (sometimes it would display 2 elements), see screenshot.
table_streaming

Another issue is that if I select another chart type, the result would be even more unstable.

r-kamath:
@zjffdu the issue in the second screenshot will not occur when viz switching is enabled, I'm working on it

r-kamath:
@zjffdu currently data update and refresh timing in ui grid is not guaranteed. This issue persists only when the interval is below 1sec. I modified your example with 2sec delay and the rendering is fine.

However, frequent refreshing of ui table is causing too many
`[Violation] 'setTimeout' handler took <N>ms` and
`[Violation] Added non-passive event listener to a scroll-blocking <some> event. Consider marking event handler as 'passive' to make the page more responsive. See <URL>`

About support for Passive Event Listeners: https://stackoverflow.com/a/39187679

r-kamath:
> @zjffdu the issue in the second screenshot will not occur when viz switching is enabled, I'm working on it

Patch will be send in a different PR.

zjffdu:
LGTM

r-kamath:

PR #2883, opened 2018-03-20
ZEPPELIN-3355. Support inline configuration for user session

zjffdu:
### What is this PR for?
Generic ConfInterpreter only support process level configuration, this PR is trying to introduce UserSessionConfInterpreter which can customize session level configuration, One example is for livy interpreter that can customize livy session. 

### What type of PR is it?
[Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3355

* Ci pass and manually test livy interpreter 

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -689,7 +689,16 @@ public void clearNoteIdAndParaMap() {
       LOGGER.info("Interpreter {} created for user: {}, sessionId: {}",
           interpreter.getClassName(), user, sessionId);
     }
-    interpreters.add(new ConfInterpreter(intpProperties, interpreterGroupId, this));
+
+    // TODO(zjffdu) this kind of hardcode is ugly. For now SessionConfInterpreter is used
+    // for livy, we could add new property in interpreter-setting.json when there's new interpreter
+    // require SessionConfInterpreter
+    if (group.equals("livy")) {
```
felixcheung: yes... rather we don't hardcode livy...

PR #2885, opened 2018-03-21
[ZEPPELIN-3001] Disable ace editor code completion when backend interpreter support completion

prabhjyotsingh:
### What is this PR for?
This is what I see in zeppelin and jupyter. The difference is due to zeppelin also add code completion of ace editor. I think code completion of ace editor should only be enabled when the backend interpreter doesn't support completion.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3001](https://issues.apache.org/jira/browse/ZEPPELIN-3001)

### Questions:


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1393,7 +1393,13 @@ private void completion(NotebookSocket conn, HashSet<String> userAndRoles, Noteb
     }
 
     final Note note = notebook.getNote(getOpenNoteId(conn));
-    List<InterpreterCompletion> candidates = note.completion(paragraphId, buffer, cursor);
```
prabhjyotsingh: this used to throw exception, I think its a better idea to log that error and send empty array to frontend,

spark/interpreter/src/main/resources/interpreter-setting.json
```diff
@@ -227,7 +232,8 @@
     },
     "editor": {
       "language": "r",
-      "editOnDblClick": false
+      "editOnDblClick": false,
+      "completionSupport": true
```
zjffdu: Should be false for SparkRInterpreter, as the completion method of SparkRInterpreter always return empty list

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -963,8 +963,73 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
     }
   };
 
+  // ref: https://github.com/ajaxorg/ace/blob/5021d0193d9f2bba5a978d0b1d7a4f73d18ce713/lib/ace/autocomplete.js#L454
+  const completionSupportWithoutBackend = function(str) {
```
prabhjyotsingh: This function is copied from the above link, and the only difference between this and below function is 
```
if (item.meta !== '') {
  return false;
}
```

spark/interpreter/src/main/resources/interpreter-setting.json
prabhjyotsingh: Sure, totally missed it, let me make that change.

zjffdu:
@prabhjyotsingh There's no response when I type tab, anything broken ?

prabhjyotsingh:
That sounds like nothing came back from the backend, can you check your WebSocket messages for "op= COMPLETION_LIST", then look under data.completions, that should be an empty array. 

Right now this PR is suppressing the default (BasicAutocompletion) completion if `completionSupport` is set to `true` for a given interpreter, should I change this to, if the return list is empty, then show the basic autocompletion?

prabhjyotsingh:
@felixcheung agreed, but at least it gives back something basic.

zjffdu:
@prabhjyotsingh What I see is that the paragraphId sent from frontend is incorrect which cause the issue I mention in the last comment. When I type tab in the first paragraph, it would send the last pargraphId to backend. could you help check that ?

zjffdu:
@prabhjyotsingh I try the latest PR, but it doesn't work when I enable zeppelin.spark.useNew. I can confirm the completion api return the same thing no matter whether zeppelin.spark.useNew is true or false. Could you help check what's difference in frontend ? Thanks

zjffdu:
ping @prabhjyotsingh

prabhjyotsingh:
I didn't know this, there is a difference is spark, let me try it out.

prabhjyotsingh:
@zjffdu I was able to reproduce the issue that you mentioned, have made a fix for the same.

zjffdu:
Thanks @prabhjyotsingh , but it still doesn't work for me.

zjffdu:
Never mind, it works for me now. Thanks @prabhjyotsingh  +1

prabhjyotsingh:

PR #2888, opened 2018-03-22
[minor] Escape string before insertion it into HTML

prabhjyotsingh:
### What is this PR for?
In current implementation some of the unescaped HTML get passed to frontend via BootstrapDialog, this PR is to escape those string (and sanitize the output).

### What type of PR is it?
[Improvement]

### Questions:


prabhjyotsingh:
@r-kamath @zjffdu can you help review this.

prabhjyotsingh:

prabhjyotsingh:
Angular template take care of these. In future, we need to watch for any javascript variables that can directly render html, like in this case it was most of the BootstrapDialog calls.

PR #2889, opened 2018-03-23
ZEPPELIN-3359. Move config storage & recovery storage to plugins

zjffdu:
### What is this PR for?
This PR move config storage & recovery storage into plugins. It is followup of ZEPPELIN-3196, the final goal is to move hadoop out of zeppelin-zengine, so that it is easy for us to load different hadoop version jars at runtime via plugins. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3359

* CI pass

### Questions:


Reamer:
@zjffdu Interesting PR. Any motivation to keep working on it?

zjffdu:
I may not have time to continue this work at least in the coming 2 months.

PR #2891, opened 2018-03-23
ZEPPELIN-1070: Inject Credentials in any Interpreter-Code

pellmont:
### What is this PR for?
This PR enables a generic syntax for inserting credentials. A username can be inserted by $[user.entry] where "entry" is the name of the credential. A password can be inserted by $[password.entry].
To avoid output of the password all occurences of the password-String in the Interpreter-output will be replaced by "###". This should not be a really secure feature (since the runner of the notebook knows the password anyway), but it should avoid accidential exposure of the used passwords by any sort of interpreter echo.

### What type of PR is it?
Feature

### Todos
* [ ] - Documentation

### What is the Jira issue?
ZEPPELIN-1070 Enable data source authentication in each community interpreters

* unit tests available

### Questions:
* Is there breaking changes for older versions? The new syntax could break notebooks which contain the patterns $[user.*] or $[password.*]. But I assume this pattern is not very common...
* Does this needs documentation? yes, but I didn't find the proper place to document it yet.

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/CredentialInjector.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/CredentialInjector.java
@@ -53,9 +53,11 @@ class CredentialInjector {
     while (matcher.find()) {
       String key = matcher.group(1);
       UsernamePassword usernamePassword = creds.getUsernamePassword(key);
-      String value = usernamePassword == null ? "undef" : usernamePassword.getUsername();
-      replaced = matcher.replaceFirst(value);
-      matcher = userpattern.matcher(replaced);
+      if (usernamePassword != null) {
+        String value = usernamePassword.getUsername();
+        replaced = matcher.replaceFirst(value);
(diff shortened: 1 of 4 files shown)


felixcheung:
thanks, sounds like that could be useful.
what do you think about "secret" management?
also for template I think we should match the other way, where we have a longer discussion: #2834

pellmont:
nice if I can contribute :-)

I think the current API for the credentials makes sense, of course it would be nice if the credentials could be shared across users without exposing the underlying passwords. But I think as long as there are different interpreters (especially those with "free-form-connections" like spark, livy, etc) it's impossible without implementing quite some code for each of the interpreters. Ending up with limited possibilities to cpnnect to a datasource.

What do you exactly mean by "match the other way"? Using the same syntax? Curly braces to be replaced and double-curly-braces to escape? Or do you mean by the configuration option(s)? Or both?

felixcheung:
Curly braces

rpastrana:
@pellmont we've been looking for a feature just like this one, are there any plans to finalize this PR? Can we help in anyway?

PR #2896, opened 2018-03-27
ZEPPELIN-3347. Fix "PYTHONPATH" for spark.pyspark

mebelousov:
### What is this PR for?
Use system PYTHONPATH in spark.pyspark interpreter, if PYTHONPATH already exists in environment variables.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3347

* Add some directories to PYTHONPATH
`export PYTHONPATH="/opt/python/dir1:/opt/python/dir2:$PYTHONPATH"`
* Build branch "ZEPPELIN-3347"
* Run in %spark.pyspark
```
import sys
print('\n'.join(sys.path))
```

* Check that the directories are in the output list.
Compare with the result in %python.

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
@@ -22,6 +22,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.io.File;
 import java.io.FilenameFilter;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/PythonUtils.java
```diff
@@ -89,6 +91,9 @@ public boolean accept(File dir, String name) {
       }
     }
 
+    if (pythonPathEnv.length() > 0) {
```
zjffdu: pythonPathEnv might be null I think

mebelousov:
@zjffdu has closed this issue in #2901

PR #2899, opened 2018-03-28
ZEPPELIN-3366 Removal of spark dependency when packaging it

jongyoul:
### What is this PR for?
Removing downloading spark, extracting some files, and injecting into Zeppelin binary

### What type of PR is it?
[Improvement]

### Todos
* [x] - Remove spark-dependencies module
* [x] - Introduce new variable of `SPARK_DISABLE_DOWNLOAD` to control downloading Spark

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3366

This PR must not break any of current behaviors. Build and test Spark scripts and queries

### Questions:
* Does this needs documentation? Yes, for Windows users. I've already updated


jongyoul:
@zjffdu Could you please review this PR?

bin/zeppelin-daemon.sh
```diff
@@ -98,6 +98,21 @@ function initialize_default_directories() {
...
     $(mkdir -p "${ZEPPELIN_PID_DIR}")
   fi
+
+  # TODO(jl): We need to discuss if we keep this feature or not for the future. See ZEPPELIN-3366
+  if [[ -z "${SPARK_HOME}" ]] && [[ "${SPARK_DISABLE_DOWNLOAD}" != "true" ]]; then
+    # I didn't make it configurable because it's not decided to be open to public
+    SPARK_DIST_DIR="${ZEPPELIN_HOME}/.spark-dist"
+    if [[ ! -d ${SPARK_DIST_DIR} ]]; then
+      mkdir -p ${SPARK_DIST_DIR}
+      URL=$(curl -fsSL https://www.apache.org/dyn/closer.lua/spark/spark-2.3.0/spark-2.3.0-bin-hadoop2.7.tgz | grep -o '<strong>[^<]*</strong>' | sed 's/<[^>]*>//g' | head -1)
```
felixcheung: how do we handle different version of spark?

spark/interpreter/pom.xml
```diff
@@ -388,48 +388,6 @@
...
-            <id>zip-pyspark-files</id>
-            <phase>generate-resources</phase>
-            <goals>
-              <goal>run</goal>
-            </goals>
-            <configuration>
-              <target>
-                <delete dir="../../interpreter/spark/pyspark" />
-                <copy file="${project.build.directory}/${spark.archive}/python/lib/py4j-${py4j.version}-src.zip" todir="${project.build.directory}/../../../interpreter/spark/pyspark" />
-                <zip basedir="${project.build.directory}/${spark.archive}/python" destfile="${project.build.directory}/../../../interpreter/spark/pyspark/pyspark.zip" includes="pyspark/*.py,pyspark/**/*.py" />
```
felixcheung: does pyspark work without this?

bin/zeppelin-daemon.sh
jongyoul: My intention was it's only for those who don't want to download Spark by themselves. We could make it configurable as same as the way to set `SPARK_VER` and `HADOOP_VER` like `testing/downlaodSpark.sh`. Do you think it's possible to be configurable?

spark/interpreter/pom.xml
jongyoul: Yes, there was a bug to use `SPARK_HOME`'s files, but it's a bug and will fix it. Basically, We don't download them if we use `SPARK_HOME`

bin/zeppelin-daemon.sh
felixcheung: hmm, also if I have SPARK_HOME I don't need to have .spark-dist, right?

jongyoul: Sure, To keep current behavior, this PR will download spark distribution. If you set one of `SPARK_HOME` and `SPARK_DISABLE_DOWNLOAD`, the download step will be ignored.

jongyoul: In case of downloading Spark distribution, `.spark-dist` will be used as a default directory to store it.

felixcheung: ok got it

felixcheung: I think the problem isn't multiple version, but with a hardcoded version + favor - it's easy for us to forget about this when we move to spark 2.4? or 2.3.1?

jongyoul: Yes, right. I'm also a bit afraid of it. This kind of hardcoded script doesn't look better. Let me find another way of handling it.

zjffdu: Downloading spark in zeppelin-daemons.h would make zeppelin server start pretty slow I think, and users may think zeppelin fails to start.

spark/interpreter/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterTest.java
```diff
@@ -59,6 +59,9 @@ private static Properties getPySparkTestProperties() throws IOException {
     p.setProperty("zeppelin.dep.localrepo", tmpDir.newFolder().getAbsolutePath());
     p.setProperty("zeppelin.pyspark.useIPython", "false");
     p.setProperty("zeppelin.spark.test", "true");
+    if (null != System.getenv("SPARK_HOME")) {
```
zjffdu: Where SPARK_HOME is set ? I don't see you set it in pom.xml

.travis.yml
```diff
@@ -66,14 +66,14 @@ matrix:
     - sudo: required
       jdk: "oraclejdk8"
       dist: trusty
-      env: BUILD_PLUGINS="true" PYTHON="3" SPARKR="true" PROFILE="-Pspark-2.2 -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="install -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/SparkIntegrationTest.java,**/ZeppelinSparkClusterTest.java,**/org/apache/zeppelin/spark/*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
+      env: BUILD_PLUGINS="true" PYTHON="3" SPARKR="true" SPARK_VER="2.2.0" HADOOP_VER="2.6" PROFILE="-Pspark-2.2 -Pscalding -Phelium-dev -Pexamples -Pscala-2.11" BUILD_FLAG="install -Pbuild-distr -DskipRat" TEST_FLAG="verify -Pusing-packaged-distr -DskipRat" MODULES="-pl ${INTERPRETERS}" TEST_PROJECTS="-Dtests.to.exclude=**/SparkIntegrationTest.java,**/ZeppelinSparkClusterTest.java,**/org/apache/zeppelin/spark/*,**/HeliumApplicationFactoryTest.java -DfailIfNoTests=false"
```
zjffdu: Why set SPARK_VER here ?

spark/interpreter/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterTest.java
jongyoul: `testing/downloadSpark.sh` sets it.

.travis.yml
jongyoul: I actually set it if it has `-Pspark-*`. I'll check it.

jongyoul:
@zjffdu I have several bugs that we use internal and external Spark at the same time. I'm also fixing those issues. It would help to remove embedded Spark.

bin/zeppelin-daemon.sh
jongyoul: Actually, `bin/zeppelin.sh` and `bin/zeppelin-daemon.sh` will show the message Spark is downloading. In case of advanced users and those who already have their own Spark package, they can set `SPARK_DISABLE_DOWNLOAD` into zeppelin-env.sh, which helps to avoid downloading Spark distribution when starting it. BTW, do you have any idea of downloading Spark distribution?

zjffdu: How about downloading spark when user start spark interpreter at the first time ? AFAIK, some users don't use spark in zeppelin, they may only use jdbc or other interpreters.

spark/interpreter/src/test/java/org/apache/zeppelin/spark/PySparkInterpreterTest.java
zjffdu: This is what I concern, because in that case it is difficult for user to run unit test in local enviroment if they don't know the internal mechanism .

zjffdu:
Thanks @jongyoul Overall I have 2 concerns.

1. Downloading spark in zeppelin-deamons.sh doesn't make sense to me. Because some users may not use spark in zeppelin, they may only use jdbc or other interpreters. And even for some users that use spark, they may not have internet access, then they will fail to start zeppelin. Personally I prefer to defer the downloading. E.g. we can download it at the first time of running spark interpreter. If user don't specify SPARK_HOME, then we either can download it in backend and display proper message in frontend or just display error message in frontend that asking user to download spark manually and specify SPARK_HOME.

2. Running unit test for PySparkInterpreter & SparkRInterpreter. Running them require correct configuration of PYTHONPATH (pyspark.zip & py4j) and SparkRLibPath (sparkr.zip). For now, we downloading spark in pom.xml and set them properly in pom.xml. But in your PR, you set it in travis script, this would make running these unit test locally difficult, especially for new contributor if they don't know the underlying mechanisms.  My suggestion is that we can download it programatically [SparkDownloadUtils](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/SparkDownloadUtils.java) and set java property `spark.home` to pass to PySparkInterpreter & SparkRInterpreter and we treat the same `spark.home` & `SPARK_HOME` in [PytonUtils](https://github.com/apache/zeppelin/blob/master/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PythonUtils.java) (Because we can not set enviroment variable programmatically). In this case, we can run unit test of PySparkInterpreter & SparkRInterpreter easily without any extra configuration. 

\cc @felixcheung @Leemoonsoo

jongyoul:
@zjffdu I understood what you're worried about. Concerning downloading Spark distribution for the first time, I'll test and implement it. About testing Z, I think your idea is more proper. I also want contributors to test z easily.

felixcheung:
maybe a downloadspark script that can launch from https://github.com/apache/zeppelin/blob/master/bin/interpreter.sh#L116?

I am not sure if it address the usability concerns fully - user can still be confused why the spark interpreter takes so long to start? right, some sort of messaging will be nice.

jongyoul:
@felixcheung That's a proper position and I've tried to implement the way you mentioned. But as you're worried about, users might be confused that kind of slow starting of Spark Interpreter, and even we have no way to show a proper message that we are downloading spark to users. In a current implementation, RuntimeException will be shown to FE side. I've talked with Jeff and am trying to find a proper position to locate code to download Spark.

BTW, while I've struggled this issue, I thought it might take almost same time to implement 'dynamic download function' :)

felixcheung:
haha yes :)
sounds good, let's see if there is a better way

PR #2905, opened 2018-04-03
ZEPPELIN-3382 Installing interpreter from UI

jongyoul:
### What is this PR for?
Installing a new interpreter from helium's UI. You just can click enable button to download a new interpreter you want to install

### What type of PR is it?
[Improvement]

### Todos
* [x] - Changed frontend side to enable users to click an install button
* [x] - Added status message for installation
* [x] - Added Backend logic to handle an interpreter installation

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3382

1. Make sure you don't have `solr` interpreter in an interpreter create menu
1. Move to Helium/Interpreter menu and click the `solr` enable button
1. See the messages for downloading it and having downloaded it
1. Check interpreter creation menu to make sure `solr` exists

image
image
image

### Questions:
* Does this needs documentation? Might be


jongyoul:
@felixcheung @zjffdu @Leemoonsoo Please review this PR.

zjffdu:
Not sure whether periodically thread checking is a good idea. What will happen if the checking thread run in the middle of interpreter copying.

@jongyoul What is your scenario of adding new interpreter ? Another option is adding some kind of trigger in UI (maybe a button) so that admin can trigger the interpreter checking via this trigger.

So this is what I imagine:
1. First copy new interpreter artifact to interpeter folder
2. Admin trigger the interpreter checking by clicking some kind of UI in frontend.

jongyoul:
@zjffdu Yes, right. I thought about the copying interpreter into the directory. I'll fix it. Actually, I thought this would help to add a new interpreter by the shell script of "install-interpreter.sh" or a simillar way. To go further, I made a method to update it and will add a rest point to update immediately. I'll add a rest endpoint to call register interpreter immediately.

jongyoul:
@zjffdu @felixcheung Can you take a look at this PR? I've changed the goal of this PR so that Installing a new interpreter become possible in helium/interpreter tab

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.zeppelin.socket.NotebookServer;
+import org.apache.zeppelin.utils.ThreadFactoryFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sonatype.aether.RepositoryException;
+
+/**
+ * This class handles all of business logic for {@link org.apache.zeppelin.rest.InterpreterRestApi}
+ */
+public class InterpreterService {
```
zjffdu: Do you plan to move all the functions of InterpreterRestApi to InterpreterService ?

zeppelin-server/src/main/java/org/apache/zeppelin/utils/ThreadFactoryFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.utils;
+
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class ThreadFactoryFactory {
```
zjffdu: You can use ThreadFactoryBuilder of guava which should also do what you want

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -364,19 +390,19 @@ private boolean registerInterpreterFromResource(ClassLoader cl, String interpret
...
+  private boolean registerInterpreterFromPath(String interpreterDir, String interpreterJson,
+      boolean override) throws IOException {
 
     Path interpreterJsonPath = Paths.get(interpreterDir, interpreterJson);
     if (Files.exists(interpreterJsonPath)) {
       LOGGER.debug("Reading interpreter-setting.json from file {}", interpreterJsonPath);
       List<RegisteredInterpreter> registeredInterpreterList =
           getInterpreterListFromJson(new FileInputStream(interpreterJsonPath.toFile()));
-      registerInterpreterSetting(registeredInterpreterList, interpreterDir);
+      registerInterpreterSetting(registeredInterpreterList, interpreterDir, override);
```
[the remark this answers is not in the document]
jongyoul: Currently, Z's initialization logic of interpreterSettings forcibly overrides. But I'm not sure we can override when installing a new interpreter with the same group by different package name. In that case, I thought it's better not to override current setting.

zeppelin-server/src/main/java/org/apache/zeppelin/utils/ThreadFactoryFactory.java
jongyoul: Thanks. Let me check it

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
jongyoul: Yes, I think it's more general structure of web services

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
```diff
@@ -281,11 +300,31 @@ public Response removeRepository(@PathParam("repoId") String repoId) {
...
+
+  /**
+   * Install interpreter
+   */
+  @POST
+  @Path("install")
+  @ZeppelinApi
+  public Response installInterpreter(@NotNull String message) {
+    logger.info("Install interpreter: {}", message);
+    InterpreterInstallationRequest request = gson
```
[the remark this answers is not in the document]
jongyoul: Yes, I'm willing to do it. But personally, I don't think it's a good way because we need to keep several gson instances. BTW I'm willing to follow the current way in this time.

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
zjffdu: Great, that is what I intended to do. See https://issues.apache.org/jira/browse/ZEPPELIN-3288
Zeppelin has 2 approach to call backend (rest api & web socket ). I think we need to refactor them to call the same api. Take this as example, all the interpreter operation should go to InterpreterService

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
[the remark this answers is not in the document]
zjffdu: What do you mean ` keep several gson instances` ? It is static instance, so should be only one instance per class

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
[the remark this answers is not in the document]
zjffdu: Sorry, I didn't get it. In what kind of scenario that user would override the same group by different package name.  Can you give me a concrete example ?

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
zjffdu: Since this is a service class, I don't think it make sense to put websocket stuff here. websocket should be the called of this service class. I think we can provide call back interface for this service class, then the caller can call the callback function when the service operation is done

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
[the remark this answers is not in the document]
jongyoul: Not a big problem. I mean I think we can reduce more of redundant gson objects, even it has only a few numbers of instances per class type. Never mind. I'll post this issue after I clarify it.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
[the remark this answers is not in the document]
jongyoul: Let me explain it more detail. In a current situation, if users enable -Pr or use R interpreter, some spark interpreter properties will be overridden by later one. I changed it to override option and made it `true` when initialized. And also set `false` when installing a new interpreter by UI.

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
jongyoul: Agreed. Actually, I also thought it's not good. I'll think of it to make a better structure.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -425,10 +451,11 @@ private void registerInterpreterSetting(List<RegisteredInterpreter> registeredIn
         .setIntepreterSettingManager(this)
         .create();
 
-    LOGGER.info("Register InterpreterSettingTemplate: {}",
-        interpreterSettingTemplate.getName());
-    interpreterSettingTemplates.put(interpreterSettingTemplate.getName(),
```
jongyoul: @zjffdu I hope this line will help you to understand `override` issue

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
jongyoul: I've introduced `MessageCallback` to handle it

jongyoul:
@zjffdu Can you check it again?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
```diff
@@ -56,18 +63,32 @@
...
 
-  private InterpreterSettingManager interpreterSettingManager;
-  private NotebookServer notebookServer;
+  private final InterpreterService interpreterService;
+  private final InterpreterSettingManager interpreterSettingManager;
+  private final NotebookServer notebookServer;
+  private final Gson gson;
 
   public InterpreterRestApi() {
+    //TODO(jl): Is this constructor needed?
```
zjffdu: If it is not used, then I think we can remove it.

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
```diff
@@ -0,0 +1,155 @@
...
+      DependencyResolver dependencyResolver,
+      Path interpreterDir,
+      MessageCallback startedMessage,
+      MessageCallback finishedMessage) {
+    Message m = new Message(OP.INTERPRETER_INSTALL_STARTED);
+    Map<String, Object> result = Maps.newHashMap();
+    try {
+      logger.info("Start to download a dependency: {}", request.getName());
+      result.put("result", "Starting");
+      result.put("message", "Starting to download " + request.getName() + " interpreter");
```
zjffdu: Thanks for introducing Callback, but I think we can do more.
1. How about put them in one callback class.
```
  interface MessageCallback {
     onStart();
     onSuccess();
     onFailure();
  }
```

2. All the message stuff should be put in the callback instead of putting them in the Service class.  E.g. 
Here we can constructing the result HashMap in callback interface. The purpose is that Service class just do the service work, and don't need to care about what do do when service is succeeded or failure, it is callback's responseentity to that do.

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
```diff
@@ -0,0 +1,155 @@
...
+      m.data = result;
+      if (null != startedMessage) {
+        startedMessage.broadcastMessage(m);
+      }
+      dependencyResolver.load(request.getArtifact(), interpreterDir.toFile());
+      interpreterSettingManager.refreshInterpreterTemplates();
+      logger.info("Finish downloading a dependency: {}", request.getName());
```
zjffdu: Can we also log the downloaded folder ?

zjffdu:
@jongyoul I tried this PR and download the pig interpreter, and found several issues.
1. The helium.json still use plugin of 0.7.2, but my zeppelin server is on 0.9 (master). Zeppelin should display the valid interpreter plugins.
2. I enable zeppelin-pig plugin, and it is downloaded to my local folder interpreter/zeppelin-pig, but I don't think it is correct, because it should be interpreter/pig

zeppelin-server/src/main/java/org/apache/zeppelin/rest/InterpreterRestApi.java
jongyoul: Fixed

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
jongyoul: Fixed

zeppelin-server/src/main/java/org/apache/zeppelin/service/InterpreterService.java
jongyoul: Agreed. Fixed

jongyoul:
@zjffdu 

1. It's what we need to discuss. I, basically, make a new PR for handing version issue for it, but we need to discuss how to manage interpreters' version. Before that, we need to provide a script to get valid helium menu. I plan to make a new issue for handling this.
1. This is also similar issue the above one. We, currently, don't have any rule for matching installed one and installable one because in case of `pig`, there's no strict naming rule for pig directory, but you just coded to copy files into `interpreter/pig` directory. In this time, I agree that I should map heuristic to prevent well-known interpreters from installing twice in a different directory.
1. I know this issue. To handle this one, we need to solve the first and second one. How about solving this issue with seperate PR?

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+  InterpreterService interpreterService;
+
+  @Before
+  public void setUp() throws Exception {
+    temporaryDir = Files.createTempDirectory("tmp");
+    interpreterDir = Files.createTempDirectory(temporaryDir, "interpreter");
+    localRepoDir = Files.createTempDirectory(temporaryDir, "local-repo");
+
+    when(mockZeppelinConfiguration.getInterpreterDir()).thenReturn(interpreterDir.toString());
```
zjffdu: We can set java properties via System.setProperty instead of usign mock ZeppelinConfiguration here

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+  @Test
+  public void downloadInterpreter() throws IOException {
+    String interpreterName = "test-interpreter";
+    String artifactName = "junit:junit:4.11";
+    Path specificInterpreterPath =
+        Files.createDirectory(Paths.get(interpreterDir.toString(), interpreterName));
+    DependencyResolver dependencyResolver = new DependencyResolver(localRepoDir.toString());
+
+    doNothing().when(mockInterpreterSettingManager).refreshInterpreterTemplates();
```
zjffdu: Need to verify that refreshInterpreterTemplates is called when download is succeeded

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
jongyoul: Yes, right. But in case of `InterpreterService`, ZeppelinConfiguration is passed as a parameter of a constructor, - personally, I think a static method is hard to test it -, thus I used mocking object of ZeppelinConfiguration.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/MessageCallback.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.socket;
+
+/** This will be used by some services to pass messages to frontend via WebSocket */
+public interface MessageCallback {
```
zjffdu: Can we rename it to ServiceCallback which I think more proper ?

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
zjffdu: Not sure what you mean. I don't think here we need static method. Just call System.setProperty in setUp should work. Or do I miss anything ?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/MessageCallback.java
jongyoul: I like it. Thanks to more proper name. Actually, I wonder if it's good.

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
jongyoul: Yes, Yes, What I mean is in most cases, ZeppelinConfiguration is created inside any class by `ZeppelinConfiguration.create()`. But I don't think it's a good pattern to test it, so I added ZeppelinConfiguration as a parameter and injected it when creating `InterpreterService`.

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
jongyoul: Let me try to solve it :-)

zeppelin-server/src/main/java/org/apache/zeppelin/socket/MessageCallback.java
jongyoul: Changed

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
jongyoul: Changed test case to check callback's message

zeppelin-server/src/test/java/org/apache/zeppelin/service/InterpreterServiceTest.java
zjffdu: Not sure whether I understand you. Here you can just do it as following which looks more easier and don't need mock. 
```
System.setProperty(ConfVars.ZEPPELIN_INTERPRETER_DIR.getVarName(), interpreterDir.toString());
System.setProperty(ConfVars.ZEPPELIN_INTERPRETER_LOCALREPO.getVarName(), localRepoDir.toString());
ZeppelinConfiguration zConf = new ZeppelinConfiguration();
interpreterService =
       new InterpreterService(zConf, mockInterpreterSettingManager);
```

jongyoul: Yes, right. We can do it. Regardless of static method, as I think, it's more fragile to use `System.setProperty()` because it might be a problem if we missed to clear or restore the value. Isn't it?

zjffdu: hmm, right. Never mind, this is not a critical issue, I am fine with your current approach.

jongyoul: Yes, it's not a big problem. You and I might have talked about some different approaches to seeing the code.

jongyoul: I feel a bit sorry because I should have explained it more details at the first time. I assumed you might know my thoughts a bit.

jongyoul:
Concerning what you mentioned the above, I've implemented the second one to check different names with same interpreters. To clarify it, if you install `zeppelin-pig` in a UI, the logic would check if `zeppelin-pig` and `pig` exist. If both don't exist, the logic would install `zeppelin-pig` in the directory of `interpreter/zeppelin-pig`. That's because I don't think it's good to change interpreter name with our current unruled format. currently, `zeppelin-pig` is extracted by maven artifact name. If we made a helium file with interpreter name 'pig', we could install `interpreter/pig`. I think it's one of the realistic solutions to handle the name of interpreters. WDYT?

And about the first and third one, I think it would be better to handle separate issues.

jongyoul:
@zjffdu I addressed all of what you commented before and test passed. https://travis-ci.org/jongyoul/zeppelin/builds/369969538

zjffdu:
Thanks @jongyoul  LGTM, and please create apache ticket for the follow up work

jongyoul:
- Helium menu: https://issues.apache.org/jira/browse/ZEPPELIN-3424
- Enable/Disable: https://issues.apache.org/jira/browse/ZEPPELIN-3425

PR #2907, opened 2018-04-03
ZEPPELIN-3347 Fix "PYTHONPATH" for spark.pyspark in branch-0.8

mebelousov:
### What is this PR for?
Use system PYTHONPATH in spark.pyspark interpreter, if PYTHONPATH already exists in environment variables.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3347

* Add some directories to PYTHONPATH
`export PYTHONPATH="/opt/python/dir1:/opt/python/dir2:$PYTHONPATH"`
* Build branch "ZEPPELIN-3347_0.8"
* Run in %spark.pyspark
```
import sys
print('\n'.join(sys.path))
```
* Check that the directories are in the output list.
Compare with the result in %python.

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
@@ -204,12 +204,17 @@ public class PySparkInterpreter extends Interpreter implements ExecuteResultHand
 
     // only set PYTHONPATH in local or yarn-client mode.
     // yarn-cluster will setup PYTHONPATH automatically.
-    SparkConf conf = getSparkConf();
+    SparkConf conf = null;
+    try {
+      conf = getSparkConf();
+    } catch (InterpreterException e) {
+      throw new IOException(e);
+    }
     if (!conf.get("spark.submit.deployMode", "client").equals("cluster")) {
       if (!env.containsKey("PYTHONPATH")) {
         env.put("PYTHONPATH", PythonUtils.sparkPythonPath());
       } else {
-        env.put("PYTHONPATH", PythonUtils.sparkPythonPath());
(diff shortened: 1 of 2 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -204,33 +204,35 @@ private Map setupPySparkEnv() throws IOException, InterpreterException {
 
     // only set PYTHONPATH in local or yarn-client mode.
     // yarn-cluster will setup PYTHONPATH automatically.
-    SparkConf conf = getSparkConf();
+    SparkConf conf = null;
+    try {
+      conf = getSparkConf();
+    } catch (InterpreterException e) {
+      throw new IOException(e);
```
felixcheung: hmm, why should this be IOException and not InterpreterException?

spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -204,33 +204,35 @@ private Map setupPySparkEnv() throws IOException, InterpreterException {
 
     // only set PYTHONPATH in local or yarn-client mode.
     // yarn-cluster will setup PYTHONPATH automatically.
-    SparkConf conf = getSparkConf();
+    SparkConf conf = null;
+    try {
+      conf = getSparkConf();
```
felixcheung: what's the reason that getSparkConf can fail here?

mebelousov: I'm just copied from #3374. 
I can simplify if necessary.
@zjffdu

zjffdu:
I have merged it, @mebelousov Could you close it ?

PR #2908, opened 2018-04-04
[HOTFIX] Fix ClassCastException when reading interpreter.json

zjffdu:
### What is this PR for?
Fix the bug reported in https://github.com/apache/zeppelin/commit/b49b3f46f91d06c4d497f8f11cbc334fd480e539#commitcomment-28397460

### What type of PR is it?
[Hot Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* NO Jira created

* CI pass

### Questions:


zjffdu:
```
I'm not 100% sure when map.get(key) is a StringMap?
```
This is due to the schema changes of interpeter.json, and we need to support different schemas for backward compatibility.

jongyoul:
Oh. This is the same one with my PR #2910. I've checked it's not always a string type. @zjffdu Could you please check my PR and merge both of them?

PR #2909, opened 2018-04-04
[ZEPPELIN-3384] Reduce the number of calls that are made to `loginUserFromKeytab` in JDBC interpreter

prabhjyotsingh:
### What is this PR for?
In JDBC interpreter every time `getConnection` has requested a call is made to `UserGroupInformation.loginUserFromKeytab` this PR is created to reduce the number of time it gets called.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3384](https://issues.apache.org/jira/browse/ZEPPELIN-3384)

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
@@ -46,10 +46,16 @@ public class JDBCSecurityImpl {
         conf.set("hadoop.security.authentication", KERBEROS.toString());
         UserGroupInformation.setConfiguration(conf);
         try {
-          UserGroupInformation.loginUserFromKeytab(
-              properties.getProperty("zeppelin.jdbc.principal"),
-              properties.getProperty("zeppelin.jdbc.keytab.location")
-          );
+          // Check TGT before calling login
+          // Ref: https://github.com/apache/hadoop/blob/release-3.0.1-RC1/hadoop-common-project/
+          // hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L1232
+          if (!UserGroupInformation.isSecurityEnabled()
+              || UserGroupInformation.getCurrentUser().getAuthenticationMethod() != KERBEROS
+              || !UserGroupInformation.isLoginKeytabBased()) {
+            UserGroupInformation.loginUserFromKeytab(
+                properties.getProperty("zeppelin.jdbc.principal"),
+                properties.getProperty("zeppelin.jdbc.keytab.location"));
+          }
         } catch (IOException e) {
           LOGGER.error("Failed to get either keytab location or principal name in the " +
               "interpreter", e);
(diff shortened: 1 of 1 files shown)


prabhjyotsingh:
@zjffdu @felixcheung  Can you help review this.

jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
```diff
@@ -46,10 +46,16 @@ public static void createSecureConfiguration(Properties properties,
...
+          // Check TGT before calling login
+          // Ref: https://github.com/apache/hadoop/blob/release-3.0.1-RC1/hadoop-common-project/
+          // hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L1232
+          if (!UserGroupInformation.isSecurityEnabled()
+              || UserGroupInformation.getCurrentUser().getAuthenticationMethod() != KERBEROS
+              || !UserGroupInformation.isLoginKeytabBased()) {
+            UserGroupInformation.loginUserFromKeytab(
+                properties.getProperty("zeppelin.jdbc.principal"),
+                properties.getProperty("zeppelin.jdbc.keytab.location"));
+          }
```
felixcheung: should we log something if the conditions are not met?

jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
```diff
@@ -46,10 +46,16 @@ public static void createSecureConfiguration(Properties properties,
...
-          UserGroupInformation.loginUserFromKeytab(
-              properties.getProperty("zeppelin.jdbc.principal"),
-              properties.getProperty("zeppelin.jdbc.keytab.location")
-          );
+          // Check TGT before calling login
+          // Ref: https://github.com/apache/hadoop/blob/release-3.0.1-RC1/hadoop-common-project/
+          // hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L1232
+          if (!UserGroupInformation.isSecurityEnabled()
+              || UserGroupInformation.getCurrentUser().getAuthenticationMethod() != KERBEROS
+              || !UserGroupInformation.isLoginKeytabBased()) {
```
zjffdu: Is the if condition logic correct ? will login keytab when security is not enabled ?

jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
prabhjyotsingh: Sure, let me add a logger in the else condition.

jdbc/src/main/java/org/apache/zeppelin/jdbc/security/JDBCSecurityImpl.java
prabhjyotsingh: The first condition (`case KERBEROS:`)  indicates that user wanted to enable Kerberos, and this second condition (combination of three flags) indicates that it is not done yet.

prabhjyotsingh:
Will merge this to master and branch-0.8 if no more discussion.

PR #2915, opened 2018-04-09
ZEPPELIN-3391 Incorrect status shown for '%livy2.conf' and %spark2.conf' interpreters

r-kamath:
### What is this PR for?
Incorrect status shown for '%livy2.conf' and %spark2.conf' interpreters

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3391](https://issues.apache.org/jira/browse/ZEPPELIN-3391)

* see jira ticket description

### Questions:


r-kamath:

zjffdu:
Thanks @r-kamath , but I don't think this is the fix. As you can see we have unit test to cover this. https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/ConfInterpreterTest.java#L62

This  should due to something bug else

r-kamath:
@zjffdu thanks for the pointer. I didn't check out the tests. My bad!

r-kamath:
right fix is in #2926

PR #2919, opened 2018-04-11
ZEPPELIN-3375: Make PySparkInterpreter extends PythonInterpreter

zjffdu:
### What is this PR for?
This PR is trying the remove the code duplication between PySparkInterpreter and PythonInterpreter. So here's the main things this PR did:
* PySparkInterpreter extends PythonInterpreter
* PySparkInterpreterTest extends PythonInterpreterTest so that we can verify PySparkInterpreter can do whatever PythonInterpreter can do
* Move interpreter/lib/python/backend_zinline.py and interpreter/lib/python/mpl_config.py into python module, so that python module can ship these resources together. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3375

* CI pass

### Questions:


zjffdu:
@felixcheung @Leemoonsoo Could you help review it ? Thanks

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -243,16 +243,21 @@ private void setupJVMGateway(int jvmGatewayPort) throws IOException {
...
       URL url = getClass().getClassLoader().getResource("grpc/python"
           + "/" + ipythonScript);
-      FileUtils.copyURLToFile(url, new File(tmpPythonScriptFolder, ipythonScript));
+      FileUtils.copyURLToFile(url, new File(pythonWorkDir, ipythonScript));
     }
 
+    //TODO(zjffdu) don't do hard code on py4j here
+    File py4jDestFile = new File(pythonWorkDir, "py4j-src-0.9.2.zip");
+    FileUtils.copyURLToFile(getClass().getClassLoader().getResource(
+        "python/py4j-src-0.9.2.zip"), py4jDestFile);
```
felixcheung: yeah, 2.3 is running with Py4J 0.10.6

python/src/main/java/org/apache/zeppelin/python/PythonCondaInterpreter.java
```diff
@@ -392,27 +382,50 @@ public Scheduler getScheduler() {
...
+    Process process = Runtime.getRuntime().exec(commands.toArray(new String[0]));
+    StreamGobbler errorGobbler = new StreamGobbler(process.getErrorStream());
+    StreamGobbler outputGobbler = new StreamGobbler(process.getInputStream());
+    errorGobbler.start();
+    outputGobbler.start();
+    if (process.waitFor() != 0) {
+      throw new IOException("Fail to run shell commands: " + StringUtils.join(commands, " "));
+    }
+    logger.info("Complete shell commands: " + StringUtils.join(commands, " "));
+    return outputGobbler.getOutput();
```
felixcheung: I thought we had some launcher wrapper for something like this?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
```diff
@@ -161,4 +161,17 @@ public boolean equals(Object o) {
...
     return id != null ? id.hashCode() : 0;
   }
+
+  public void close() {
+    for (List<Interpreter> session : sessions.values()) {
+      for (Interpreter interpreter : session) {
+        try {
+          interpreter.close();
+        } catch (InterpreterException e) {
+          LOGGER.warn("Fail to close interpreter: " + interpreter.getClassName());
```
felixcheung: would the exception stack be useful?
just to LOGGER.warn( .... , e);?

spark/interpreter/src/main/resources/python/zeppelin_pyspark.py
```diff
@@ -15,150 +15,43 @@
...
 java_import(gateway.jvm, "org.apache.spark.mllib.api.python.*")
 
-intp = gateway.entry_point
-output = Logger()
-sys.stdout = output
-sys.stderr = output
-
-jsc = intp.getJavaSparkContext()
-
-if sparkVersion.isImportAllPackageUnderSparkSql():
```
felixcheung: why not keeping this?

spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -159,474 +129,81 @@ public void open() throws InterpreterException {
...
-        !sparkInterpreter.isYarnMode()) {
-      String sparkSubmitJars = conf.get("spark.jars").replace(",", ":");
-      if (!StringUtils.isEmpty(sparkSubmitJars)) {
-        env.put("PYTHONPATH", env.get("PYTHONPATH") + ":" + sparkSubmitJars);
+    if (!useIPython()) {
+      // Initialize Spark in Python Process
+      try {
+        bootstrapInterpreter("python/zeppelin_pyspark.py");
+      } catch (IOException e) {
+        e.printStackTrace();
```
felixcheung: log?

python/src/main/resources/python/zeppelin_python.py
```diff
@@ -142,19 +152,33 @@ def handler_stop_signals(sig, frame):
...
+          # only call it when it is not for code completion. code completion will call it in
+          # PythonCompletion.getCompletion
+          intp.setStatementsFinished("", False)
+      except Py4JJavaError:
+        # raise it to outside try except
+        raise
       except:
-        raise Exception(traceback.format_exc())
+        if not isForCompletion:
+          exception = traceback.format_exc()
```
felixcheung: add a comment what this is looking for and what it looks like typically?

python/src/main/resources/python/zeppelin_python.py
```diff
@@ -47,46 +35,80 @@ def reset(self):
...
+      self.interpreter.setStatementsFinished(result, False)
+
+host = sys.argv[1]
+port = int(sys.argv[2])
+
+client = GatewayClient(address=host, port=port)
+gateway = JavaGateway(client, auto_convert = True)
+intp = gateway.entry_point
+# redirect stdout/stderr to java side so that PythonInterpreter can capture the python execution result
+output = Logger()
```
felixcheung: there are problems reported with these names too common, conflict with existing variables, output, gateway, client etc, we should name these uniquely if we could

python/src/main/resources/python/zeppelin_python.py
```diff
@@ -47,46 +35,80 @@ def reset(self):
...
+    execResult = locals()
+    try:
+      exec("{} = dir({})".format("objectDefList", objName), _zcUserQueryNameSpace, execResult)
+    except:
+      self.interpreter.logPythonOutput("Fail to run dir on " + objName)
+      self.interpreter.logPythonOutput(traceback.format_exc())
+      return None
+    else:
+      objectDefList = execResult['objectDefList']
+      return [completion for completion in execResult['objectDefList'] if completion.startswith(methodName)]
```
felixcheung: `startswith ` -I think a lot times partial match - not necessarily from the beginning, can be good?

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
felixcheung: should this detect any mismatch here? check spark version or something?

zjffdu: It is fine to use py4j 0.9.2 here for IPythonInterpreter, as for IPySparkInterpreter it would use the py4j of spark instead of py4j 0.9.2

python/src/main/resources/python/zeppelin_python.py
zjffdu: It if fine to use them, because they are not in the namespace of user code.

python/src/main/resources/python/zeppelin_python.py
zjffdu: Might be, there's a lot work to do for improving the code completion. This PR is large, I don't want to cover too much thing in this single PR.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterGroup.java
zjffdu: Fixed

spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
zjffdu: Fixed

python/src/main/resources/python/zeppelin_python.py
zjffdu: Fixed

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
Tagar: I wonder why Spark doesn't ship py4j zip file version-agnostic? Filed https://issues.apache.org/jira/browse/SPARK-23965

python/src/main/resources/python/zeppelin_python.py
felixcheung: I'm not sure - this sets globally right? for example `z` is accessible from user code.

zjffdu: User code is in namespace `_zcUserQueryNameSpace` instead of global namespace https://github.com/zjffdu/zeppelin/blob/ZEPPELIN-3375/python/src/main/resources/python/zeppelin_python.py#L91

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
HyukjinKwon: I don't think this is a strong reason to rename or make a link for Spark's Py4J. Also, to be clear, I think It's an orthogonal issue with the current change here, if I am not mistaken.

zjffdu:
Will merge if no more comment

python/src/main/resources/python/zeppelin_python.py
felixcheung: got it

PR #2920, opened 2018-04-11
[ZEPPELIN-2679] JDBC. precode for session

mebelousov:
### What is this PR for?
We have a precode for interpreter (all interpreters) which executes once after start interpreter. With it we can create shared code that will be available in the paragraph, create tables in the database and etc.
This precode for jdbc only which executes before run paragraph in same session (sql). In this precode we can declare session variables that will be available when you run paragraph (any sql which will be executed before the paragraph). For example `set application_name='#{user};`
Previous discussion in #2442 

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2679

(jdbc interpreter, postgres)
1. set property **default.statementPrecode** `set application_name='#{user}';`
2. run paragraph `select current_setting('application_name');`

### Questions:


mebelousov:
@zjffdu 
Another repo and another PR, but the same error
https://travis-ci.org/TinkoffCreditSystems/zeppelin/builds/365002216

PR #2923, opened 2018-04-12
ZEPPELIN-3312 Add option to convert username to lower case

VipinRathor:
### What is this PR for?
This PR introduces a new configuration property to convert username to lower case. This is useful when the users (from external sources like AD/LDAP) are coming in with mixed-case names and Hadoop services (like Hive) can't authorize them correctly because Hadoop services recognize users only in lower case (like Linux). 

Adding a new config option "zeppelin.username.force.lowercase" to handle such scenarios.

Behavior without this PR:
Access is denied to CaMel case user while running a Hive paragraph

Behavior with this PR:
User is allowed to run query when proposed configuration set to true.
By default, keeping zeppelin.username.force.lowercase=false to retain the current behavior.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Unit test

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3312

* Travis CI should pass
* Manual steps to test:
1. Configure Zeppelin with Active Directory authentication
2. Login to Zeppelin as a CaMel case user
3. Try to run a simple JDBC note with a Hive query (like a select * query). This would fail with "user [CaMel] does not have proper privileges to [USE] operation" error message.
4. Now set zeppelin.username.force.lowercase=true in custom zeppelin-site.xml configuration.
5. Once again, login as CaMel case user. This time the same Hive query would run as expected. Because the username is now passed in lower case.
6. Also notice that after successful login, the login username (in the top-right corner) will be in lower case too.

* Login as CaMel case user:
screen shot 2018-04-12 at 3 33 01 am
* Notice the converted username post login:
screen shot 2018-04-12 at 3 33 43 am

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -411,6 +411,12 @@
 </property>
 
 <property>
+  <name>zeppelin.username.force.lowercase</name>
+  <value>false</value>
+  <description>Force convert username case to lower case, useful for Active Directory. Default is not to change case</description>
+</property>
+
+<property>
   <name>zeppelin.notebook.default.owner.username</name>
   <value></value>
   <description>Set owner role by default</description>
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -413,7 +413,7 @@
 <property>
   <name>zeppelin.username.force.lowercase</name>
   <value>false</value>
-  <description>Force convert username case to lower case, useful for Active Directory. Default is not to change case</description>
+  <description>Force convert username case to lower case, useful for Active Directory/LDAP. Default is not to change case</description>
 </property>
 
 <property>
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
@@ -94,7 +94,7 @@ public class SecurityUtils {
       principal = extractPrincipal(subject);
       if (ZeppelinServer.notebook.getConf().isUsernameForceLowerCase()) {
         log.debug("Converting principal name " + principal
-          + " to lower case:" + principal.toLowerCase());
+            + " to lower case:" + principal.toLowerCase());
         principal = principal.toLowerCase();
       }
     } else {
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
@@ -44,6 +44,7 @@ import javax.naming.NamingException;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.realm.ActiveDirectoryGroupRealm;
 import org.apache.zeppelin.realm.LdapRealm;
+import org.apache.zeppelin.server.ZeppelinServer;
 
 /**
  * Tools for securing Zeppelin.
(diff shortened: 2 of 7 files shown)


VipinRathor:
@prabhjyotsingh @felixcheung @zjffdu Can you please help review? Thanks.

Tagar:
Great idea.. I had a similar jira https://issues.apache.org/jira/browse/ZEPPELIN-2886 
Thanks!

VipinRathor:
All Travis CI builds are failing for some npm error. Looks unrelated but I'm not sure. I need some help in getting this through.
CC: @prabhjyotsingh @r-kamath

zjffdu:
LGTM @prabhjyotsingh could you help take a look at the npm error in travis ? Some other contributors also hit this same in other PR recently.

r-kamath:
@VipinRathor can you try removing `npm cache clear` from .travis.yml ? otherwise add `--force`.

npm doc says.
> All data that passes through the cache is fully verified for integrity on both insertion and extraction. Cache corruption will either trigger an error, or signal to pacote that the data must be refetched, which it will do automatically. For this reason, it should never be necessary to clear the cache for any reason other than reclaiming disk space, thus why clean now requires --force to run.
[details](https://docs.npmjs.com/cli/cache#details)

r-kamath:
@VipinRathor #2928

conf/zeppelin-site.xml.template
```diff
@@ -410,6 +410,12 @@
   <description>Anonymous user allowed by default</description>
 </property>
 
+<property>
+  <name>zeppelin.username.force.lowercase</name>
+  <value>false</value>
+  <description>Force convert username case to lower case, useful for Active Directory. Default is not to change case</description>
```
felixcheung: could you elaborate more on `, useful for Active Directory`? say `as Active Directory/LDAP are case-insensitive`?
also, isn't this applicable to LDAP too?

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -91,6 +92,11 @@ public static String getPrincipal() {
     String principal;
     if (subject.isAuthenticated()) {
       principal = extractPrincipal(subject);
+      if (ZeppelinServer.notebook.getConf().isUsernameForceLowerCase()) {
+        log.debug("Converting principal name " + principal
+          + " to lower case:" + principal.toLowerCase());
```
felixcheung: log.warn()? since this is possibly security related.

conf/zeppelin-site.xml.template
VipinRathor: yes, this is very much applicable to LDAP. I'll update in the next commit.

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
VipinRathor: Can do but then there will be lot noise at info log level. Are we sure that we want to have this?

VipinRathor:
5/11 builds are still failing but due to unrelated errors. Tried restarting failed one multiple times. All 5 are showing different errors. We can either ignore them or fix them in separate issues.

VipinRathor:
After few more retries, only 2/11 are failing now. Just restarted them again with the hope that those 2 will succeed too.

Tagar:
I was thinking we should also lower-case usernames that are stored in `notebook-authorization.json`, when doing authorization. Otherwise users wouldn't be able to see their own notebooks when this option is enabled and they used case inconsistently when were logging into Zeppelin before.

VipinRathor:
@Tagar, me & @prabhjyotsingh did discuss notebook authorization changes but somehow I missed to include them. I committed those changes yesterday. Now notebook authorization will also be mindful of username case conversion. Thanks for pointing out!

Today, we also discovered that similar to notebook authorization, interpreter "set permission" is another place where case conversion will be required. I've just committed those changes as well.

prabhjyotsingh:
https://travis-ci.org/VipinRathor/zeppelin/builds/369033290 looks flaky, works well on local.

Merging this to master and branch-0.8 if no more discussion.

Tagar:
Can you guys please have a look at https://issues.apache.org/jira/browse/ZEPPELIN-3719 in case if this change caused that problem ?

PR #2924, opened 2018-04-12
[ZEPPELIN-3356] Kerberos ticket still expire after 7 days

zjffdu:
### What is this PR for?

The root cause is that we may do UserGroupInformation.loginUserFromKeytab multiple times if we use hdfs for several places, e.g. for notebook repo and zeppelin config. This PR fix the bug by only doing UserGroupInformation.loginUserFromKeytab one time.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3356

Manually tested
* Add the following properties to /etc/krb5.conf
   renew_lifetime = 10m
  ticket_lifetime = 5m
* Add the following properties to /var/kerberos/krb5kdc/kdc.conf
  max_renewable_life = 10m
  max_life = 5m

Without this PR, will hit the ticket expire after 10 minutes. With this PR, the ticket is still valid after 10 minutes 

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
@@ -30,9 +30,31 @@ public class FileSystemStorage {
 
   private static Logger LOGGER = LoggerFactory.getLogger(FileSystemStorage.class);
 
+  // only do UserGroupInformation.loginUserFromKeytab one time, otherwise you will still get
+  // your ticket expired.
+  static {
+    if (UserGroupInformation.isSecurityEnabled()) {
+      ZeppelinConfiguration zConf = ZeppelinConfiguration.create();
+      String keytab = zConf.getString(
+          ZeppelinConfiguration.ConfVars.ZEPPELIN_SERVER_KERBEROS_KEYTAB);
+      String principal = zConf.getString(
+          ZeppelinConfiguration.ConfVars.ZEPPELIN_SERVER_KERBEROS_PRINCIPAL);
+      if (StringUtils.isBlank(keytab) || StringUtils.isBlank(principal)) {
+        throw new RuntimeException("keytab and principal can not be empty, keytab: " + keytab
+            + ", principal: " + principal);
+      }
+      try {
+        UserGroupInformation.loginUserFromKeytab(principal, keytab);
+      } catch (IOException e) {
+        throw new RuntimeException("Fail to login via keytab:" + keytab +
+            ", principal:" + principal, e);
+      }
+    }
+  }
+
   private ZeppelinConfiguration zConf;
   private Configuration hadoopConf;
-  private boolean isSecurityEnabled = false;
+  private boolean isSecurityEnabled;
   private FileSystem fs;
 
   public FileSystemStorage(ZeppelinConfiguration zConf, String path) throws IOException {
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -30,9 +30,31 @@
...
+  static {
+    if (UserGroupInformation.isSecurityEnabled()) {
+      ZeppelinConfiguration zConf = ZeppelinConfiguration.create();
+      String keytab = zConf.getString(
+          ZeppelinConfiguration.ConfVars.ZEPPELIN_SERVER_KERBEROS_KEYTAB);
+      String principal = zConf.getString(
+          ZeppelinConfiguration.ConfVars.ZEPPELIN_SERVER_KERBEROS_PRINCIPAL);
+      if (StringUtils.isBlank(keytab) || StringUtils.isBlank(principal)) {
+        throw new RuntimeException("keytab and principal can not be empty, keytab: " + keytab
+            + ", principal: " + principal);
```
felixcheung: is `static Logger LOGGER` created? could we log here?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -30,9 +30,31 @@
...
+      }
+      try {
+        UserGroupInformation.loginUserFromKeytab(principal, keytab);
+      } catch (IOException e) {
+        throw new RuntimeException("Fail to login via keytab:" + keytab +
+            ", principla:" + principal);
```
felixcheung: wrap e

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
zjffdu: It would be logged by outside code  (caller of FileSystemStorage)

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
zjffdu: Fixed

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -30,9 +30,31 @@
...
+          ZeppelinConfiguration.ConfVars.ZEPPELIN_SERVER_KERBEROS_PRINCIPAL);
+      if (StringUtils.isBlank(keytab) || StringUtils.isBlank(principal)) {
+        throw new RuntimeException("keytab and principal can not be empty, keytab: " + keytab
+            + ", principal: " + principal);
+      }
+      try {
+        UserGroupInformation.loginUserFromKeytab(principal, keytab);
+      } catch (IOException e) {
+        throw new RuntimeException("Fail to login via keytab:" + keytab +
+            ", principla:" + principal, e);
```
felixcheung: ops, didn't catch the typo  `principal`

zjffdu: Fixed

zjffdu:

PR #2914, opened 2018-04-09
[ZEPPELIN-3350] Don't allow set cronExecutionUser

zjffdu:
### What is this PR for?
This PR just does a quick fix this security issue. 
1. Remove the setting cron user in frontend
2. Run the note via owner.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3350

### How should this be tested?
* Manually tested

### Screenshots (if appropriate)
screen shot 2018-04-09 at 3 04 06 pm

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #2925, opened 2018-04-13
ZEPPELIN-3404. Fail to run cronjob when user doesn't run it manually before cronjob

zjffdu:
### What is this PR for?

This bug is introduced by #2914, this PR will set authenticationInfo using its user as we will store user into note.json.  

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3404

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -447,16 +447,14 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
       if (!$scope.note.config.cronExecutingUser) {
         $scope.note.config.cronExecutingUser = $rootScope.ticket.principal;
       }
+      if (!$scope.note.config.cronExecutingRoles) {
+        $scope.note.config.cronExecutingRoles = $rootScope.ticket.roles;
+      }
     } else {
       $scope.note.config.cronExecutingUser = '';
+      $scope.note.config.cronExecutingRoles = '';
     }
     $scope.note.config.cron = cronExpr;
-    $scope.setConfig();
-  };
-
-  /** Set the username of the user to be used to execute all notes in notebook **/
-  $scope.setCronExecutingUser = function(cronExecutingUser) {
-    $scope.note.config.cronExecutingUser = cronExecutingUser;
     $scope.setConfig();
   };
 
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -21,6 +21,7 @@ import static java.lang.String.format;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.io.IOException;
@@ -31,6 +32,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -645,7 +647,16 @@ public class Note implements ParagraphJobListener, JsonSerializable {
    * Run all paragraphs sequentially. Only used for CronJob
    */
   public synchronized void runAll() {
-    runAll(null, true);
+    String cronExecutingUser = (String) getConfig().get("cronExecutingUser");
+    String cronExecutingRoles = (String) getConfig().get("cronExecutingRoles");
+    if (null == cronExecutingUser) {
+      cronExecutingUser = "anonymous";
+    }
+    AuthenticationInfo authenticationInfo = new AuthenticationInfo(
+        cronExecutingUser,
+        StringUtils.isEmpty(cronExecutingRoles) ? null : cronExecutingRoles,
+        null);
+    runAll(authenticationInfo, true);
   }
 
   public void runAll(AuthenticationInfo authenticationInfo, boolean blocking) {
@@ -653,9 +664,7 @@ public class Note implements ParagraphJobListener, JsonSerializable {
       if (!p.isEnabled()) {
         continue;
       }
-      if (authenticationInfo != null) {
-        p.setAuthenticationInfo(authenticationInfo);
-      }
+      p.setAuthenticationInfo(authenticationInfo);
       if (!run(p.getId(), blocking)) {
         logger.warn("Skip running the remain notes because paragraph {} fails", p.getId());
         break;
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/user/AuthenticationInfo.java
@@ -76,6 +76,10 @@ public class AuthenticationInfo implements JsonSerializable {
     this.roles = roles;
   }
 
+  public void setRoles(String roles) {
+    this.roles = gson.fromJson(roles, ArrayList.class);
+  }
+
   public List<String> getUsersAndRoles() {
     List<String> usersAndRoles = new ArrayList<>();
     if (roles != null) {


zjffdu:
@felixcheung Mind to help review it ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -655,6 +655,13 @@ public void runAll(AuthenticationInfo authenticationInfo, boolean blocking) {
       }
       if (authenticationInfo != null) {
         p.setAuthenticationInfo(authenticationInfo);
+      } else if (p.getUser() != null){
+        // set AuthenticationInfo of this paragraph using its user when authenticationInfo is null.
+        // authenticationInfo will be null if cronjob started before user run this
+        // paragraph manually.
+        p.setAuthenticationInfo(new AuthenticationInfo((p.getUser())));
+      } else {
+        p.setAuthenticationInfo(AuthenticationInfo.ANONYMOUS);
```
felixcheung: looks ok, is it better to make sure authenticationInfo is set when scheduling the cronjob?

r-kamath:
LGTM

weand:
so running via cron will use the last user that executed a paragraph. that does not fit exactly to the docu change from original PR:
> Cron executing user (It is removed from 0.8 where it **enforces the cron execution user to be the note owner** for security purpose)
[https://github.com/apache/zeppelin/pull/2914/files#diff-bed1be4fb1f7f6c8e3bd6db18d72dc44](https://github.com/apache/zeppelin/pull/2914/files#diff-bed1be4fb1f7f6c8e3bd6db18d72dc44)

zjffdu:
Thanks @weand , even I didn't realize the user in note.json means the last user run the paragraph (I thought it is the owner). In that case, it seems meaningless to store user in note.json IMHO.  
Regarding the issue, I will update it soon

zjffdu:
Update the PR to run the cron job as the note owner

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -655,6 +656,16 @@ public void runAll(AuthenticationInfo authenticationInfo, boolean blocking) {
...
         p.setAuthenticationInfo(authenticationInfo);
+      } else {
+        // set AuthenticationInfo of this paragraph using its owner when authenticationInfo is null.
+        // authenticationInfo will be null if cronjob started before user run this
+        // paragraph manually.
+        Set<String> owners = NotebookAuthorization.getInstance().getOwners(id);
+        if (owners.isEmpty()) {
+          p.setAuthenticationInfo(AuthenticationInfo.ANONYMOUS);
+        } else {
+          p.setAuthenticationInfo(new AuthenticationInfo(owners.iterator().next()));
```
felixcheung: should we check if `owners.iterator().next()` is an empty string?

mebelousov:
Groups can be note owners. As I see `AuthenticationInfo` is not valid for internal shiro groups.

How can I find the user which scheduled note?

zjffdu:
@mebelousov Currently I would only allow note owner to set the schedule and run cron job via the owner. But seems zeppelin allow multiple owners for one note (it doesn't make sense to me), this is only what I can do the eliminate the security issue.

mebelousov:
@zjffdu I have checked with group as owner. This is ok.
Why do you think that use of owner is more secure than cronExecutingUser?

weand:
@zjffdu is there a chance to set the cronExecutionUser implictly to the user who activates the cron schedule. 
and in addtion only allow owners (which can be defined by username or groupnames) to activate the cron schedule.

weand:
and optionally (or alternatively) introduce a config parameter for a static cronExecutionUser (kind of service user) to be defined by ops team. when this config parameter is set, cron schedules will always use that user.

zjffdu:
@mebelousov What do you mean `group as owner` ? And how do you use that ?

mebelousov:
@weand cronExecutionUser is set to the user who set the cron string inside the note.
@zjffdu I define groups in shiro.ini.
The user sets groupname in owner field. Then the note runs on cron from this groupname.

image

zjffdu:
@mebelousov Your solution depends on shiro setting, we  could not assume that in code.

zjffdu:
I am saying using group is not good, it just depends on user's setting. BTW, it looks like this PR also works for your scenario. In this PR, I would always choose the owner to run the cronjob. And in your case, you just set the owner as group. Do I understand it correctly ? :)

weand:
@mebelousov I fully agree that running document as the group is not what a user would expect.

That's why I'm proposing another approach again:
1) prior to #2914 the cronExecutionUser config property was added to org.apache.zeppelin.notebook.Note.config when activating the cron and entering a username
2) that property is not set anymore. this PR here only sets the authentication info at running time to something (a principal) from the permission tab (which can be user or groups, which can not be determined properly)
3) so why not again set that cronExecutionUser config property from 1) (under the hood): with the user who activates the cron schedule expression. when only the owner can enable cron expressions, it will always be the username regardless of what principals exactly are configured in the permissions tab.

zjffdu:
Thanks @weand , I agree with you that your proposal seems much easier. I have updated the PR, please help verify it, thanks

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -645,17 +646,21 @@ private void setParagraphMagic(Paragraph p, int index) {
...
 
   public void runAll(AuthenticationInfo authenticationInfo, boolean blocking) {
     for (Paragraph p : getParagraphs()) {
       if (!p.isEnabled()) {
         continue;
       }
-      if (authenticationInfo != null) {
-        p.setAuthenticationInfo(authenticationInfo);
-      }
+      p.setAuthenticationInfo(authenticationInfo);
```
felixcheung: yap, I like this better

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -645,17 +646,21 @@ private void setParagraphMagic(Paragraph p, int index) {
    * Run all paragraphs sequentially. Only used for CronJob
    */
   public synchronized void runAll() {
-    runAll(null, true);
+    String cronExecutingUser = (String) getConfig().get("cronExecutingUser");
+    if (null == cronExecutingUser) {
```
felixcheung: not sure it is matters here, check for empty or blank string?

weand:
The notebook now runs successfully after restart with the user who enabled the scheduler.

There is one more valid use case left in my mind (dealing with permissions):
- interpreter permissions have been set only for groups (e.g. DepartmentA should be allowed to run spark interpreter)
- user created note with %spark interpreter and schedules a cron
- now when cron runs, only username is added to authentication info. groups of the user (e.g. DepartmentA in that example) are omitted and the notebook fails with an error, that user 'xyz' has no permission for running spark interpreter.

any chance to fix that case as well? 
I think of something like introducing cronExecutingRoles (in addition to  cronExecutingUser) and adding all roles of the user who enables the cron to that config property. And then when a cron runs, all that roles are added to the corresponding authentication info.

weand:
The notebook now runs successfully after restart with the user who enabled the scheduler.

2 remaining issues:

1) user1 schedules the cron. then the note is executed via cron with 'user1' only as long as nobody else runs the note. as soon as 'user2' runs the note manually, the cron will also use 'user2' from that time on. Endusers would expect the cron will always run as 'user1'.

2) One more valid but yet uncovered use case left in my mind (dealing with permissions):
- interpreter permissions have been set only for groups (e.g. DepartmentA should be allowed to run spark interpreter)
- user created note with %spark interpreter and schedules a cron
- now when cron runs, **only username is added to authentication info. groups of the user (e.g. DepartmentA in that example) are omitted** and the notebook fails with an error, that **user 'xyz' has no permission for running the interpreter**.

any chance to fix that case as well? 
I think of something like introducing cronExecutingRoles (in addition to  cronExecutingUser) and adding all roles of the user who enables the cron to that config property. And then when a cron runs, all that roles are added to the corresponding authentication info.

mebelousov:
@weand Thank you!
Some addition to 2.
Over time user may not belong to group.
At first we could store cronExecutingRoles and in future it's to be good to check groups on the fly.

weand:
@zjffdu uhm, I can't confirm it. Today after checking out this PR from scratch my first point now works correctly. sorry for the spam.

created issue [ZEPPELIN-3427](https://issues.apache.org/jira/browse/ZEPPELIN-3427) for the second point. is that really so complex to wait for 0.8.1 or 0.9.0 ?

Part of that change would be:
1) moving/adding logic of setting note.config.cronExecutionUser + cronExecutionRoles to org.apache.zeppelin.socket.NotebookServer.isCronUpdated(Map<String, Object>, Map<String, Object>) when _cronUpdated = true_.
2) Then evaluate those config properties in org.apache.zeppelin.notebook.Note.runAll() when authenticationInfo is created

zjffdu:
Thanks for the hint @weand , I am not familiar with this component and didn't realize it is so not difficult to do that. 

PR is updated, please help try that. Thanks again.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -645,17 +647,29 @@ private void setParagraphMagic(Paragraph p, int index) {
    * Run all paragraphs sequentially. Only used for CronJob
    */
   public synchronized void runAll() {
-    runAll(null, true);
+    String cronExecutingUser = (String) getConfig().get("cronExecutingUser");
+    String cronExecutingRoles = (String) getConfig().get("cronExecutingRoles");
+    if (null == cronExecutingUser) {
+      cronExecutingUser = "anonymous";
+    }
+    AuthenticationInfo authenticationInfo = new AuthenticationInfo();
```
weand: Parsing roles here with split(...) does not work anymore, when having more than one role.
There were recent changes to parsing the roles, see:
https://github.com/apache/zeppelin/commit/ad77265c69bf2ea5fe82fa8c15c7a3d3ea2178e3#diff-4a78303414018c7d08a044687ab6afd4R57

This code instead worked locally. AuthenticationInfo constructor will properly parse the roles:
```
    ...
    AuthenticationInfo authenticationInfo = new AuthenticationInfo(
        cronExecutingUser, 
        StringUtils.isEmpty(cronExecutingRoles) ? null : cronExecutingRoles,
        null);
    runAll(authenticationInfo, true);
```

zjffdu: Thanks, fixed

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -447,19 +447,16 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
       if (!$scope.note.config.cronExecutingUser) {
         $scope.note.config.cronExecutingUser = $rootScope.ticket.principal;
       }
+      if (!$scope.note.config.cronExecutingRoles) {
+        $scope.note.config.cronExecutingRoles = $rootScope.ticket.roles;
+      }
     } else {
       $scope.note.config.cronExecutingUser = '';
```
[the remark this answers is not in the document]
zjffdu: fixed

zjffdu:
Thanks @weand for review, comments are addressed

weand:
great. LGTM :+1: please merge to master and branch-0.8

PR #2926, opened 2018-04-13
ZEPPELIN-3391. Incorrect status shown for '%livy2.conf' and %spark2.conf' interpreters

zjffdu:
### What is this PR for?
The root cause of this bug is that the final status of paragraph is not only determined by interpreter, but also by its scheduler. Here the bug is in FIFOScheduler. I just fix it as a quick solution in this PR. But for a long term solution, I think we may need to do code refactoring in the scheduler component. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3391

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
@@ -23,6 +23,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 
+import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.scheduler.Job.Status;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


zjffdu:
@felixcheung @r-kamath Could you help review it ? Thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
```diff
@@ -137,15 +138,26 @@ public void run() {
               listener.jobStarted(scheduler, runningJob);
             }
             runningJob.run();
+            Object jobResult = runningJob.getReturn();
             if (runningJob.isAborted()) {
               runningJob.setStatus(Status.ABORT);
+              LOGGER.debug("Job Aborted, " + runningJob.getId() + ", " +
+                  runningJob.getErrorMessage());
```
felixcheung: is there an error message in case of abort?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
```diff
@@ -137,15 +138,26 @@ public void run() {
...
+            } else if (runningJob.getException() != null) {
+              LOGGER.debug("Job Error, " + runningJob.getId() + ", " +
+                  runningJob.getReturn());
```
felixcheung: shouldnt this be `runningJob.getErrorMessage()`

felixcheung: LOGGER.error()?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
```diff
@@ -137,15 +138,26 @@ public void run() {
...
+              runningJob.setStatus(Status.ERROR);
+            } else if (jobResult != null && jobResult instanceof InterpreterResult
+                && ((InterpreterResult) jobResult).code() == InterpreterResult.Code.ERROR) {
+              LOGGER.debug("Job Error, " + runningJob.getId() + ", " +
+                  runningJob.getReturn());
```
felixcheung: is there more information about the ERROR that we could log here?
from jobResult?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
zjffdu: Job error (Paragraph error) is normal for zeppelin, so I don't think we need to make it LOGGEr.error here.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
zjffdu: This if condition correpend to user code error, so the job return value could give sufficient info.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/FIFOScheduler.java
felixcheung: ok

PR #2929, opened 2018-04-17
ZEPPELIN-3412 Enable query prefix syntax in bigquery interpreter

iijima-satoshi:
### What is this PR for?
- interpreter
Enable query prefix syntax in bigquery interpreter.
Bigquery has a convenient query prefix syntax #standardSQL and #legacySQL.
This enable to switch between standardSQL and legacySQL without changing UseLegacySql option.
Currently (v0.8.0 or later), bigquery interpreter definitely runs with UseLegacySql option.
As described in the [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql?hl=en#sql-prefix), when this option is set to true or false, the query prefix cannot be used.

- documentation
Removed some contents from README.md. They are duplicated with interpreter documentation.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3412

- Unit test should pass

### Questions:

### DIFF
--- a/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
+++ b/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
@@ -226,19 +226,15 @@ public class BigQueryInterpreter extends Interpreter {
     String projId = getProperty(PROJECT_ID);
     long wTime = Long.parseLong(getProperty(WAIT_TIME));
     long maxRows = Long.parseLong(getProperty(MAX_ROWS));
-    String sqlDialect = getProperty(SQL_DIALECT, "");
-    Boolean useLegacySql;
-    switch (sqlDialect) {
-      case "standardSQL":
-        useLegacySql = false;
-        break;
-      case "legacySQL":
-        useLegacySql = true;
-        break;
-      default:
-        // Enable query prefix like '#standardSQL' if specified
-        useLegacySql = null;
+    String sqlDialect = getProperty(SQL_DIALECT, "").toLowerCase();
+    // Query prefix like '#standardSQL' can be used in case of null
+    Boolean useLegacySql = null;
+    if (sqlDialect.contains("standard")) {
+      useLegacySql = false;
+    } else if (sqlDialect.contains("legacy")) {
+      useLegacySql = true;
     }
+
     Iterator<GetQueryResultsResponse> pages;
     try {
       pages = run(sql, projId, wTime, maxRows, useLegacySql);
--- a/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
+++ b/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
@@ -89,7 +89,7 @@ public class BigQueryInterpreter extends Interpreter {
   static final String PROJECT_ID = "zeppelin.bigquery.project_id";
   static final String WAIT_TIME = "zeppelin.bigquery.wait_time";
   static final String MAX_ROWS = "zeppelin.bigquery.max_no_of_rows";
-  static final String LEGACY_SQL = "zeppelin.bigquery.use_legacy_sql";
+  static final String SQL_DIALECT = "zeppelin.bigquery.sql_dialect";
 
   private static String jobId = null;
   private static String projectId = null;
@@ -226,8 +226,19 @@ public class BigQueryInterpreter extends Interpreter {
     String projId = getProperty(PROJECT_ID);
     long wTime = Long.parseLong(getProperty(WAIT_TIME));
     long maxRows = Long.parseLong(getProperty(MAX_ROWS));
-    String legacySql = getProperty(LEGACY_SQL);
-    boolean useLegacySql = legacySql == null ? true : Boolean.parseBoolean(legacySql);
+    String sqlDialect = getProperty(SQL_DIALECT, "");
+    Boolean useLegacySql;
+    switch (sqlDialect) {
+      case "standardSQL":
+        useLegacySql = false;
+        break;
+      case "legacySQL":
+        useLegacySql = true;
+        break;
+      default:
+        // Enable query prefix like '#standardSQL' if specified
+        useLegacySql = null;
+    }
     Iterator<GetQueryResultsResponse> pages;
     try {
       pages = run(sql, projId, wTime, maxRows, useLegacySql);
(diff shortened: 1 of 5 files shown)


iijima-satoshi:
https://travis-ci.org/iijima-satoshi/zeppelin/builds/367533713
I tried to run travis ci several times. But #n.4 always failed with below errors.
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```
or
```
Done. Your build exited with 1.
grep: write error: No space left on device
```

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
```diff
@@ -89,7 +89,7 @@
   static final String PROJECT_ID = "zeppelin.bigquery.project_id";
   static final String WAIT_TIME = "zeppelin.bigquery.wait_time";
   static final String MAX_ROWS = "zeppelin.bigquery.max_no_of_rows";
-  static final String LEGACY_SQL = "zeppelin.bigquery.use_legacy_sql";
+  static final String SQL_DIALECT = "zeppelin.bigquery.sql_dialect";
```
felixcheung: removing a config would break existing users- how about adding the new sql_dialect but also keep the use_legacy_sql one?

docs/interpreter/bigquery.md
```diff
@@ -48,6 +48,11 @@ limitations under the License.
     <td>100000</td>
     <td>Max result set size</td>
   </tr>
+  <tr>
+    <td>zeppelin.bigquery.sql_dialect</td>
+    <td></td>
+    <td>Bigquery SQL dialect (standardSQL or legacySQL). If empty, query prefix like '#standardSQL' can be used.</td>
```
felixcheung: is there a link we could add to the details of these dialects?

zjffdu:
I am OK with backporting to 0.8 since this is bigquery specific fix, should not affect the zeppelin core.

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
felixcheung: that https://github.com/apache/zeppelin/pull/2929/files#r182642493 was what I meant.

but if use_legacy_sql is new and never released than I'm good with removing it

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
```diff
@@ -226,8 +226,19 @@ private InterpreterResult executeSql(String sql) {
     String projId = getProperty(PROJECT_ID);
     long wTime = Long.parseLong(getProperty(WAIT_TIME));
     long maxRows = Long.parseLong(getProperty(MAX_ROWS));
-    String legacySql = getProperty(LEGACY_SQL);
-    boolean useLegacySql = legacySql == null ? true : Boolean.parseBoolean(legacySql);
+    String sqlDialect = getProperty(SQL_DIALECT, "");
+    Boolean useLegacySql;
+    switch (sqlDialect) {
+      case "standardSQL":
```
felixcheung: is this going to be case sensitive?

docs/interpreter/bigquery.md
felixcheung: yes, please add them into this PR.

also should we quote the values, ie. "standardSQL" and if they are case sensitive

iijima-satoshi:
@felixcheung I addressed your comments. Case sensitive, and furthermore exact match are not required. thanks.

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
```diff
@@ -226,19 +226,15 @@ private InterpreterResult executeSql(String sql) {
...
-      case "legacySQL":
-        useLegacySql = true;
-        break;
-      default:
-        // Enable query prefix like '#standardSQL' if specified
-        useLegacySql = null;
+    String sqlDialect = getProperty(SQL_DIALECT, "").toLowerCase();
+    // Query prefix like '#standardSQL' can be used in case of null
+    Boolean useLegacySql = null;
+    if (sqlDialect.contains("standard")) {
```
felixcheung: I'd prefer exact match - is there a reason to prefer substring/contains?

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
```diff
@@ -226,19 +226,15 @@ private InterpreterResult executeSql(String sql) {
...
-        break;
-      case "legacySQL":
-        useLegacySql = true;
-        break;
-      default:
-        // Enable query prefix like '#standardSQL' if specified
-        useLegacySql = null;
+    String sqlDialect = getProperty(SQL_DIALECT, "").toLowerCase();
+    // Query prefix like '#standardSQL' can be used in case of null
+    Boolean useLegacySql = null;
```
felixcheung: what would be the behavior to call QueryRequest .setUseLegacySql(null)?

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
iijima-satoshi: ok, I'd defer to you.

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
iijima-satoshi: I verified the behavior including other cases. 
- setUseLegacySql(null) : Default query dialect is legacySQL and query prefix can be used (the same as default behavior in BigQuery)
- setUseLegacySql(false): Query dialect is only standardSQL and query prefix cannot be used
- setUseLegacySql(true) : Query dialect is only legacySQL and query prefix cannot be used

iijima-satoshi:
ready to review

bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
felixcheung: k thanks

iijima-satoshi:
@felixcheung Fixed. Thank you for the review.

zjffdu:
LGTM, will merge if no more comment

PR #2930, opened 2018-04-18
ZEPPELIN-3414. branch-0.8 - download py4j from main repo

mebelousov:
### What is this PR for?
Now py4j-0.9.2 is downloaded from "pypi.python.org". Seems that "pypi.python.org" needs new version of OpenSSL.
There are some troubles to build Zeppelin on old CentOS/RHEL because is not easy to update OpenSSL.
It's good to change repo for py4j.

### What type of PR is it?
Small Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3414

* Tests pass

### Questions:

### DIFF
--- a/python/pom.xml
+++ b/python/pom.xml
@@ -40,8 +40,6 @@
         **/PythonInterpreterPandasSqlTest.java,
         **/PythonInterpreterMatplotlibTest.java
     </python.test.exclude>
-    <pypi.repo.url>https://pypi.python.org/packages</pypi.repo.url>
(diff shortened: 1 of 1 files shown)


zjffdu:
@mebelousov I don't think this will work for 0.8. In 0.9, I just put py4j in the code base of zeppelin, that's why I can avoid downloading it in pom.

zjffdu:
@mebelousov I don't think this will work for 0.8. In 0.9, I just put py4j in the code base of zeppelin, that's why I can avoid downloading it in pom.

mebelousov:
Jeff, you are right. Need rework.

PR #2931, opened 2018-04-18
ZEPPELIN-3405 Zeppelin fails to display the User home page if user belongs to roles with space in its name

r-kamath:
- escape roles value in login and ticket json response
- fix method name typo

### What is this PR for?
Fix string escape issue in roles. More details in JIRA description.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3405

see JIRA description

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
@@ -16,6 +16,7 @@
  */
 package org.apache.zeppelin.rest;
 
+import com.google.gson.Gson;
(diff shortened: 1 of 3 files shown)


zjffdu:
LGTM

r-kamath:

weand:
@prabhjyotsingh @zjffdu 
:-1: this actually blew up the permission checks with roles :-1:

When...
- enabling shiro.ini
- enabling permission on an interpreter for _role1_
- executing note with that interpreter via _user1_

...execution fails with __user1 has no permission for spark__

I'm wondering if there is no test for this use case. When reverting this PR the execution works fine.

PR #2935, opened 2018-04-22
ZEPPELIN-3411 Long running logic inside synchronized block in InterpreterSettingManager

jongyoul:
### What is this PR for?
Removing redundant synchronized code to avoid blocking other logics.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Change synchronized block to read/write lock

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3411

* Current tests should be passed

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -24,7 +24,6 @@ import com.google.common.collect.Maps;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
@@ -65,7 +64,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -921,8 +919,7 @@ public class InterpreterSettingManager {
 
   public void close() {
     List<Thread> closeThreads = new LinkedList<>();
-    Collection<InterpreterSetting> intpSettings = interpreterSettings.values();
-    for (final InterpreterSetting intpSetting : intpSettings) {
+    for (final InterpreterSetting intpSetting : interpreterSettings.values()) {
       Thread t =
           new Thread() {
             public void run() {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -24,6 +24,7 @@ import com.google.common.collect.Maps;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


jongyoul:
This is not a critial issue when running normally. But in case of problemistic situation like restarting forcibly to stop long running job of SparkInterpreter, some logic hangs inside synchronized block.

zjffdu:
Do you have jstack so that I can understand what the exact problem is

jongyoul:
@zjffdu Let me make it

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -103,6 +104,7 @@
    */
   private final Map<String, InterpreterSetting> interpreterSettings =
       Maps.newConcurrentMap();
+  private final ReentrantReadWriteLock interpreterSettingsLock = new ReentrantReadWriteLock();
```
zjffdu: Actually interpreterSettings is ConcurrentHashMap, I think we don't need extra lock here.

jongyoul: Yes, I thought it and wanted to confirm it. Then, I'll remove all of the locks related to interpreterSettings in this class.

jongyoul: Fixed

jongyoul:
Removed all of the synchronized blocks from `InterpreterSettingManager` for `interpreterSettings`

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -940,17 +921,16 @@ public void close(String settingId) {
...
-        Thread t = new Thread() {
-          public void run() {
-            intpSetting.close();
-          }
-        };
-        t.start();
-        closeThreads.add(t);
-      }
+    Collection<InterpreterSetting> intpSettings = interpreterSettings.values();
+    for (final InterpreterSetting intpSetting : intpSettings) {
```
zjffdu: Merge them into one line
```
for (InterpreterSetting intpSetting : interpreterSettings.values()) {
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -24,6 +24,7 @@
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
```
zjffdu: this could be removed

zjffdu:
THanks @jongyoul Just several minor issues, rest look good to me.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
jongyoul: Fixed

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
jongyoul: Fixed

jongyoul:
After finishing CI, I'll merge it

jongyoul:
Test passed. https://travis-ci.org/jongyoul/zeppelin/builds/369961193 Will merge it

weand:
are you going to merge that into branch-0.8 as well?

jongyoul:
@weand Hi, I’d like to do it but need to check the timeline of the release date because the two following PRs should be implemented. Thus I think it would be better to merge it after the rest parts would be implemented. WDYT? @zjffdu when is the target date for the next release?

jongyoul:
@weand Oops. I was confused with another issue. I’ll cherry-pick this into 0.8

PR #2941, opened 2018-04-24
[ZEPPELIN-3311] Zeppelin should ask user to re-login after JWT cookie expires (KnoxJwtRealm)

prabhjyotsingh:
### What is this PR for?
Zeppelin should ask user to re-login after JWT cookie expires.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3311](https://issues.apache.org/jira/browse/ZEPPELIN-3311)

Steps to reproduce :
1) Enable SSO for zeppelin
2) Login to zeppelin using SSO.
3) Now wait till JWT expires(knoxsso.token.ttl under Advanced knoxsso-topology could be set to lower values to get a short lived cookie)
4) Refresh zeppelin. Zeppelin doesn't ask for re-login, though JWT is already expired and not valid anymore.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
@@ -16,6 +16,7 @@
  */
 package org.apache.zeppelin.realm.jwt;
 
+import java.util.Date;
 import org.apache.commons.io.FileUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.Groups;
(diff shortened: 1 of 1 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
```diff
@@ -182,6 +193,33 @@ protected boolean validateSignature(SignedJWT jwtToken) {
...
+   * subclasses in order to customize the expiration validation behavior.
+   *
+   * @param jwtToken
+   *            the token that contains the expiration date to validate
+   * @return valid true if the token has not expired; false otherwise
+   */
+  protected boolean validateExpiration(SignedJWT jwtToken) {
+    boolean valid = false;
+    try {
+      Date expires = jwtToken.getJWTClaimsSet().getExpirationTime();
```
zjffdu: Does LOCALE matter here ?

prabhjyotsingh: It's taken from https://github.com/apache/ranger/blob/master/security-admin/src/main/java/org/apache/ranger/security/web/filter/RangerSSOAuthenticationFilter.java#L467-L491.
ExpirationTime is passed as an epoch, so it should not matter.

prabhjyotsingh: Yes, that I understood LOCALE for this context is time zone, yes it should not matter.

r-kamath:
LGTM

zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
zjffdu: Thanks, LGTM

prabhjyotsingh:

PR #2944, opened 2018-04-25
[ZEPPELIN-3429] Fix python version on dev/merge_zeppelin_pr.py

jwagun:
### What is this PR for?
Current merge_zeppelin_pr.py shebang has user env dependent.
Which means python version would be different from each user's environment.

However this script is unavailable with python 3.
Therefore I think it needs to be fixed with available python version.

 ### What type of PR is it?
[Bug]

 ### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3429

Run merge_zeppelin_pr.py on the system using default python version 3.

 ### Questions:

### DIFF
--- a/dev/merge_zeppelin_pr.py
+++ b/dev/merge_zeppelin_pr.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python2
 
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more


dev/merge_zeppelin_pr.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python2
```
prabhjyotsingh: This will not work in all the cases. For example this is what I see on my local.
```
python             python2.7          python3            python3.6          python3.6m         pythonw            
python-config      python2.7-config   python3-config     python3.6-config   python3.6m-config  pythonw2.7
```

And now when I try to run this it gives me an error `/usr/bin/python2: bad interpreter: No such file or directory`

r-kamath:
@jwagun can you try adding `alias python='python2'` in your bash profile?

zjffdu:
How about to make it support both python2 and python3 ?

r-kamath:
@zjffdu Need to refactor to support python 2 and 3 simultaneously.
IMHO alias to right version of python 2.x should make things run without any failure.

jwagun:
@r-kamath @prabhjyotsingh @zjffdu
Right. Supporting both version is the better solution. I will close this pr. Thanks.

PR #2945, opened 2018-04-26
[ZEPPELIN-3430] fix logic of loading githubnotebookrepo

cvaliente:
### What is this PR for?
loadNotebookRepo(String notebookRepoClassName) does not actually load notebookRepoClassName but instead it loads the first NotebookRepo class that it finds on the plugins classpath. In case of NotebookRepos that depend/extend other NotebookRepos, this results in the wrong class being loaded.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3430

* set up notebook that uses githubrepo and check if the class actually gets initialized (look for debug log entry)

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/plugin/PluginManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/plugin/PluginManager.java
@@ -83,8 +83,13 @@ public class PluginManager {
       return null;
     }
     URLClassLoader classLoader = new URLClassLoader(urls.toArray(new URL[0]));
-    Iterator<NotebookRepo> iter = ServiceLoader.load(NotebookRepo.class, classLoader).iterator();
-    NotebookRepo notebookRepo = iter.next();
+    NotebookRepo notebookRepo = null;
+    try {
+      notebookRepo = (NotebookRepo) (Class.forName(notebookRepoClassName, true, classLoader)).newInstance();
+    } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+      LOGGER.warn("Fail to instantiate notebookrepo from plugin classpath:" + notebookRepoClassName, e);
+    }
+
     if (notebookRepo == null) {
       LOGGER.warn("Unable to load NotebookRepo Plugin: " + notebookRepoClassName);
     }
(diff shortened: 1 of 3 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/plugin/PluginManager.java
```diff
@@ -83,8 +83,13 @@ public NotebookRepo loadNotebookRepo(String notebookRepoClassName) throws IOExce
       return null;
     }
     URLClassLoader classLoader = new URLClassLoader(urls.toArray(new URL[0]));
-    Iterator<NotebookRepo> iter = ServiceLoader.load(NotebookRepo.class, classLoader).iterator();
-    NotebookRepo notebookRepo = iter.next();
+    NotebookRepo notebookRepo = null;
+    try {
+      notebookRepo = (NotebookRepo) (Class.forName(notebookRepoClassName, true, classLoader)).newInstance();
```
zjffdu: Could you still load the notebookRepo via ServiceLoader and just verify the classname equal to  notebookRepoClassName ?

myuwono: @cvaliente thanks for this fix and for highlighting the issue. We also ran to this classpath loading issue and applying this change seems to have fixed it! Prior to this ServiceLoader seems to find VFSNotebookRepo instead of the desired GitNotebookRepo.

Tagar:
This PR fixes the issue I described in https://github.com/apache/zeppelin/pull/2760#issuecomment-386148566
Now the git/versioning subpanel is back!
Thanks.

Tagar:
@zjffdu, is it possible to commit this into the 0.8 branch? thanks.

zjffdu:
@Tagar It only affect 0.9, there's no plugin in 0.8

Tagar:
got it, thanks @zjffdu

zeppelin-zengine/src/main/java/org/apache/zeppelin/plugin/PluginManager.java
zjffdu: hmm, @cvaliente You are right, ServiceLoader is not necessary here. I will remove ServiceLoader mechanism in another PR.

zjffdu:
LGTM, will merge to master soon

PR #2950, opened 2018-04-27
ZEPPELIN-3432 Fix results object when throwing exception while running job

jongyoul:
### What is this PR for?
Handling the issue that `InterpretJob.jobRun()` throws an `InterpreterException`

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Change the logic to create `InterpreterResult` when throwing `InterpreterException`

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3432

1. Enable `zeppelin.spark.sql.stacktrace` option
2. %sql select invalid query

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
@@ -22,6 +22,7 @@ import java.util.Date;
 import java.util.Map;
 
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.zeppelin.interpreter.InterpreterException;
 import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.InterpreterResult.Code;
 import org.slf4j.Logger;
@@ -202,7 +203,11 @@ public abstract class Job {
   }
 
   private synchronized void completeWithError(Throwable error) {
-    setResult(new InterpreterResult(Code.ERROR, error.getMessage()));
+    // TODO(jl): Remove this trick. Most of error are covered by InterpreterException
+    if (error instanceof InterpreterException && null != error.getCause()) {
+      error = error.getCause();
+    }
+    setResult(new InterpreterResult(Code.ERROR, getStack(error)));
     setException(error);
     dateFinished = new Date();
   }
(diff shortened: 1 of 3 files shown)


zjffdu:
LGTM

jongyoul:
image

I've tested it but found it didn't show proper messages. I'll fix it until this weekend

zjffdu:
@jongyoul Sorry, it seems my mistake in #2926 , revert the changes in RemoteInterpreterServer.java should be able to fix this.

jongyoul:
@zjffdu Thanks. I've checked that issue and added the original logic to `Job` class as it looks more clear to handle it in one class. After merging this PR, others' classes don't have to care the case that `results` is not `InterpreterResult` class.

image

zjffdu:
LGTM Thanks @jongyoul

jongyoul:
@zjffdu Thanks. I've fixed tests for adopting changes. I'll merge it after CI finishes.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
```diff
@@ -200,7 +203,11 @@ private synchronized void completeWithSuccess(Object result) {
   }
 
   private synchronized void completeWithError(Throwable error) {
-    setResult(error.getMessage());
+    // TODO(jl): Remove this trick. Most of error are covered by InterpreterException
+    if (error instanceof InterpreterException && null != error.getCause()) {
+      error = error.getCause();
+    }
+    setResult(new InterpreterResult(Code.ERROR, getStack(error)));
```
felixcheung: do we need to return the whole stack as the result? that seems a bit much to the user. how about just the message as before and log the exception?

jongyoul: I'm trying to check if `getMessage()` is enough or not.

jongyoul: BTW, AFAIK, our past behaviours was also showing all stacktraces.

zjffdu: I would suggest to use full stacktrace for now, as this is the behavior we used before. Although I think we should not output the full stacktrace, we need to investigate it more to make sure the error message is clear and sufficient for users. Job is one place, there's other places like (e.g. when invalid interpreter is specified )

jongyoul: image

It think it's enough for the simple error but not sure for other cases. And it would be better to keep it this time because this PR is kind of howfixes :-)

jongyoul:
@zjffdu I've changed not to unwrap `InterpreterException` as I've found some problems when I run an unavailable script at the first time. I need to check it finally and will tell you the results

jongyoul:
I've checked this PR finally.

jongyoul:
Will merge it today.

PR #2959, opened 2018-05-04
ZEPPELIN-3443 - Fix regex to support optional space before ;

alexott:
### What is this PR for?

Almost all regular expressions, except for DESCRIBE MATERIALIZED VIEW in the CQL interpreter allow to have an optional space before closing ; character.  There is some code that uses these spaces before ; character, so the command fails when you copy/past commands.

This PR fixes it

### What type of PR is it?
Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-3443

Fix is trivial and doesn't require unit test, or extra testing


doanduyhai:
The fix is trivial, looks good to me

alexott:
Build failed because of problems of `urllib3`/`pyopenssl`...

r-kamath:

PR #2964, opened 2018-05-09
ZEPPELIN-2221 Show `Jobs` page when `jobId` is missing

jongyoul:
### What is this PR for?
Because of yarn's bug, Spark's `jobId` is not passed. This causes some Spark UI link looks broken. In this kind of case, showing `Jobs` page looks reasonable. Yarn's bug is already fixed with the latest version only, so we need to handle it in Zeppelin side.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Return `Jobs` page when `Job` page is not available

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2221

1. Install outdated yarn
2. Run script
3. Click `Spark UI` to show `Jobs` page instead of 404

### Questions:

### DIFF
--- a/spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
+++ b/spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
@@ -19,6 +19,12 @@
 package org.apache.zeppelin.spark;
 
 
+import java.io.IOException;
+import org.apache.commons.httpclient.HttpStatus;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.DefaultRedirectStrategy;
+import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.zeppelin.interpreter.BaseZeppelinContext;
 import org.apache.zeppelin.interpreter.remote.RemoteEventClientWrapper;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
```diff
@@ -87,10 +101,31 @@ protected String getParagraphId(String jobgroupId) {
...
+  protected void setHttpClient(HttpClient httpClient) {
+    this.httpClient = httpClient;
+  }
+
   protected void buildSparkJobUrl(String sparkWebUrl, int jobId, Properties jobProperties) {
     String jobGroupId = jobProperties.getProperty("spark.jobGroup.id");
     String uiEnabled = jobProperties.getProperty("spark.ui.enabled");
     String jobUrl = sparkWebUrl + "/jobs/job?id=" + jobId;
+    // See ZEPPELIN-2221 which fixes an issue passing wrong parameters. It's related to YARN-6615.
+    HttpGet httpGetRequest = new HttpGet(jobUrl);
```
zjffdu: Could you use org.apache.hadoop.util.VersionInfo to get the hadoop version info in SparkShims ? Otherwise invoking http request per spark job is very costly.

jongyoul: I think so. But is `VersionInfo` getting the version dynamically?

jongyoul: OH!!! I've tested it would get the current version. I'll fix it

jongyoul: Fixed

jongyoul:
Changed it and added related tests

jongyoul:
Merging it if there's no further discussion

PR #2903, opened 2018-04-02
[ZEPPELIN-3377] Passing Z variables to JDBC interpreter

sanjaydasgupta:
### What is this PR for?
This PR enables the interpolation of ZeppelinContext objects into the paragraph text of JDBC cells. It also introduces a new interpreter-level configuration parameter named _zeppelin.jdbc.interpolation_. This new parameter is _false_ by default, and must be set to _true_ to enable object interpolation. The default value of _false_ guarantees backward compatibility for users who are not aware of the new feature.

The implementation takes the same approach that was followed in [PR-2898](https://github.com/apache/zeppelin/pull/2898).

I have also taken the liberty to correct a preexisting error in the description of the use of Dynamic Forms in the associated documentation (_jdbc.md_).

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3377

### How should this be tested?
CI Pass. 

The code in this PR merely causes the JDBC interpreter to "opt-in" to the implementation already existing in the `Interpreter` base class - described in [PR-2898](https://github.com/apache/zeppelin/pull/2898). The unit tests necessary are already present in PR-2898

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? Yes, documentation has been added to the file _jdbc.md_. I have also taken the liberty to correct a preexisting error in the description of the use of Dynamic Forms in the associated documentation.

PR #2970, opened 2018-05-16
ZEPPELIN-3379 Refactoring of JDBC autocomplete.

mebelousov:
### What is this PR for?
Current implementation of autocomplete allows only completion in `schema.table.column` scenario. Such scenario exclude tables for default schema.
The PR fixes the scenario with tables in default schema and alias processing.
Also I want to add feature to complete columns for used tables in statement from blank cursor.

The autocomplete was tested for Greenplum 4.3 (PostgreSQL), Oracle 11g, MySQL (Hive Metastore) and Hive.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3379

* Manually
* Tests pass

mysql

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -810,7 +810,7 @@ public class InterpreterSetting {
       public void run() {
         try {
           // dependencies to prevent library conflict
-          File localRepoDir = new File(conf.getInterpreterLocalRepoPath() + "/" + getId());
+          File localRepoDir = new File(conf.getInterpreterLocalRepoPath() + "/" + id);
           if (localRepoDir.exists()) {
             try {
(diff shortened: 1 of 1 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -783,9 +784,7 @@ private String replaceReservedChars(String str) {
   }
 
   @Override
-  public InterpreterResult interpret(String originalCmd, InterpreterContext contextInterpreter) {
```
sanjaydasgupta: Hi @mebelousov the change proposed in these lines removes the capability described in [ZEPPELIN-3377](https://issues.apache.org/jira/browse/ZEPPELIN-3377) (Passing Z variables to JDBC interpreter) and the corresponding [PR-2903](https://github.com/apache/zeppelin/pull/2903). 

Is it necessary to remove ZEPPELIN-3377 capability to implement this PR?

PR #2971, opened 2018-05-17
[ZEPPELIN-3466] Table export to excel is not working due to missing dependencies

prabhjyotsingh:
### What is this PR for?
Table export to excel is not working due to missing dependencies.
angular-ui-grid 4.4.7 onwards comes pre-packaged with jszip and excel-builder, but have few noticeable bugs hence not upgrading to latest.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3466](https://issues.apache.org/jira/browse/ZEPPELIN-3466)

* Export to excel should work.

### Questions:

### DIFF
--- a/zeppelin-web/bower.json
+++ b/zeppelin-web/bower.json
@@ -34,7 +34,9 @@
     "MathJax": "2.7.0",
     "ngclipboard": "^1.1.1",
     "jsdiff": "3.3.0",
-    "ngInfiniteScroll": "^1.3.4"
+    "ngInfiniteScroll": "^1.3.4",
+    "jszip": "2.6.1",
+    "excel-builder-js": "excelbuilder#2.0.0"
   },
   "devDependencies": {
     "angular-mocks": "1.5.7"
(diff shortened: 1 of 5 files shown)


prabhjyotsingh:
Yes, this affects branch-0.8. I mean prior to ZEPPELIN-3289 there was no "Table export to excel". 
Let me merge this soon so that this also becomes part of the Zeppelin-0.8 release.

Tagar:
Could you please also have a look at https://issues.apache.org/jira/browse/ZEPPELIN-3478
CSV/TSV exports seem to be broken too.

Tagar:
How to activate XLS/XLSX download option? 
It doesn't show up for me with this PR applied.

image

prabhjyotsingh:
@Tagar  It shows up on the table (comes as a part of angular-ui-grid)
screenshot from 2018-05-22 22-35-07

Tagar:
Got it @prabhjyotsingh - thanks.

So there are two separate ways to export data now - [1] is the newer way and [2] is the older way.
What's advantage of having two export options? 
Should we leave just [1] if this newer export through Angular-ui-grid seems to be better?
For example, I think it might be a little bit confusing to users to have two different ways to exprot csv.

Thanks.

[1]
export1

[2]

export2

prabhjyotsingh:
Yeah, I know it looks redundant and might be confusing, both have their ups and downs, we can choose to hide one.

PR #2974, opened 2018-05-18
ZEPPELIN-3472 No interpreter status is shown after restarting the individual interpreter.

sameer79:
### What is this PR for?
Shows the interpreter status after restarting the individual interpreter.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3472](https://issues.apache.org/jira/browse/ZEPPELIN-3472)

Click on setting icon on note-book interpreter panel will be shown. click on refresh icon for the specific interpreter from the list user will promt to confirm do he want's to restart the interpreter. If yes depending on the api response the status has been shown in the form of (circle) color mapping below.

- GREEN -> Interpreter successfully restarted
- RED -> Error has been occure while restarting

on hovering the (circle) you will see success / error  message for the respective interpreter.

zeppelin-3472

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.html
+++ b/zeppelin-web/src/app/notebook/notebook.html
@@ -61,6 +61,18 @@ limitations under the License.
                 </span>
               </small>
             </div>
+            <small ng-switch="item.status">
+              <small ng-switch-when="READY">
+                <i style="color: green; margin-right: 3px;" class="fa fa-circle"
+                   uib-tooltip="Interpreter restart successfully">
+                </i>
+              </small>
+              <small ng-switch-when="ERROR">
+                <i style="color: red; cursor: pointer" class="fa fa-circle"
+                   uib-tooltip="{{item.errorReason}}">
+                </i>
+              </small>
+            </small>
           </div>
         </div>
       </div>
--- a/zeppelin-web/src/app/notebook/notebook.html
+++ b/zeppelin-web/src/app/notebook/notebook.html
@@ -72,6 +72,11 @@ limitations under the License.
                    uib-tooltip="{{item.errorReason}}">
                 </i>
               </small>
+              <small ng-switch-when="YELLOW">
+                <i style="color: yellow; cursor: pointer" class="fa fa-circle"
+                   uib-tooltip="{{item.errorReason}}">
+                </i>
+              </small>
             </small>
           </div>
         </div>
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -1034,12 +1034,13 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
               checkInterpreterStatus(data.body);
               thisConfirm.close();
             }).error(function(data, status, headers, config) {
+              checkInterpreterStatus({
+                id: interpreter.id,
+                errorReason: 'Not able to restart the interpreter.',
+                status: 'YELLOW',
+              });
               thisConfirm.close();
               console.log('Error %o %o', status, data.message);
-              BootstrapDialog.show({
-                title: 'Error restart interpreter.',
-                message: _.escape(data.message),
-              });
             });
           return false;
         }
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -1002,6 +1002,14 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
     $timeout(makeSearchBoxVisible());
   });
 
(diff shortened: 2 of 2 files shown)


Tagar:
I think it's a great idea to add something like this to also show status if an interpreter is running. 
so show a green icon if interpreter is running/ or had been restarted successfully. 
no icon - if it wasn't started.
red icon - if interpreter had issues restarting, or if it was terminated / killed through Lifecycle management..
what do you think?

zjffdu:
+1 for the idea or icon for interpreter process status

sameer79:
@Tagar  Thanks for the comment, it will be handled with the new JIRA

zeppelin-web/src/app/notebook/notebook.html
```diff
@@ -61,6 +61,23 @@ <h5>Interpreter binding</h5>
                 </span>
               </small>
             </div>
+            <small ng-switch="item.status">
+              <small ng-switch-when="READY">
+                <i style="color: green; margin-right: 3px;" class="fa fa-circle"
```
r-kamath: is `margin-right` required here ?

sameer79: @r-kamath  Thanks for taking time to reviewing the code, as per the margin-right concern it provides the breathing space between the interpreter list items and the (circle) which we show for interpreter status.

bikassaha:
Lets say I restarted the interpreter. Between that time and the interpreter actually starting, what icon is shown for the "start in progress state?" A blinking green?

Also is this for the logical interpreter or the interpreter JVM process? E.g. in scoped mode interpreter (like Livy) will this be green if the Livy interpreter JVM is running or if the actual interpreter session for the that scope?

bikassaha:
I am guessing that would depend on whether we want to user to continue doing other stuff or make them wait for the interpreter to start/restart.

The interpreter page is kind of a global config place. For scoped mode interpreters or isolated mode or per note interpreters, each user/note could have their own interpreter instance and so the interpreter page may not be the best place to capture that. With %conf interpreter, now the admin could also consider locking down the interpreter page in order to prevent global interpreter restarts etc. With admin/user level settings locking down that page would be more important. So user visible setting may be better off where they are most applicable. On the notebook. What do you think? @felixcheung  @sameer79

felixcheung:
hmm, but I can think of the opposite case too - that the interpreter is per-note and the user wants to have access to restart it, and not get locked down. sounds like a broader issue with "what interpreter controls should go"

bikassaha:
Thats a good point. Though I thought interpreter restart was available from the note itself. Is that not the case? Agree interpreter controls/settings between admin/user needs to be looked at in general.

sameer79:
@bikassaha  @felixcheung  This is an extension to https://github.com/apache/zeppelin/pull/1155 

The green icon does not mean the interpreter is running, it only indicates the interpreter is ready to be run
 - green - ready to run
 - red - some error like download dependencies failed
 - spin animation - it is in process of restarting
 to make the UI consistent with the interpreter setting page

let me know your thought/improvements if required.

bikassaha:
Does "ready to run" mean code can be executed in the note?

sameer79:
@bikassaha  "Yes," The interpreter has successfully restarted.

sameer79:
@r-kamath @prabhjyotsingh @bikassaha  Awaiting for a review.

zjffdu:
It looks like there's many misunderstandings between people about what the status icon really means.  Personally I would suggest to abandon this PR for now, as I think end users would also has the confusion about what that icon really means. status icon is nice to have but not must to have, so before we nail down what that status icon represent, we should not deliver such kind of feature. 
What user really care about is that when the interpreter fails to start, zeppelin frontend should tell the end user what's wrong and what should be done to fix that.

zjffdu:
@sameer79 Or we could make it simple for now that the status icon means whether the interpreter restarting operation is successful just as what this PR did now IIUC.  One concern is that when the restarting operation is failed, what user would get in frontend besides the red status icon. Because I think red status icon is not sufficient for users, user don't know what to do for this kind of failure. At least, frontend should tell user what's wrong in restarting interpreter.

bikassaha:
The description says "on hovering the (circle) you will see success / error message for the respective interpreter." I wonder if that is going to show whats wrong in starting/restarting the interpreter.

zjffdu:
hmm, I missed that statement in description. But I think it would be better to pop up a error message to users when it fails to restart. I don't think user would know to hover over that icon to see the error message. 

BTW, another thing I'd like to clarify is that the restarting button doesn't means `restarting interpreter` for now. It is just `terminate the interpreter but not start it`.  Interpreter starting happens when user run paragraph.This is a legacy issue we should fix.

felixcheung:
I think the green light can be easily mis-interpreted. If you have any other suggestion on the icon to use perhaps we could revisit.

PR #2977, opened 2018-05-22
ZEPPELIN-3482 Incorrect user is picked up by Zeppelin during relogin after Knox SSO token expiry

r-kamath:
### What is this PR for?
Incorrect user is picked up by Zeppelin during relogin after Knox SSO token expiry

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3482

* steps in JIRA description.

### Questions:


prabhjyotsingh:

PR #2978, opened 2018-05-22
[ZEPPELIN-3467] two-step, atomic configuration file writes

sanjaydasgupta:
### What is this PR for?
This PR prevents the total loss of configuration information that happens when zeppelin attempts to write out modified configuration information into an already full file-system. The approach taken here is to first write the modified information to a temporary file (in the same directory), and then to rename the successfully written file to its eventual name. 

The rename operation is not attempted if there is an error while writing the temporary file, thus leaving the pre-existing configuration file untouched.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3467

CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
@@ -18,6 +18,7 @@
 package org.apache.zeppelin.storage;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.FileUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.interpreter.InterpreterInfoSaving;
 import org.apache.zeppelin.notebook.NotebookAuthorizationInfoSaving;
@@ -51,7 +52,7 @@ public class LocalConfigStorage extends ConfigStorage {
   @Override
   public void save(InterpreterInfoSaving settingInfos) throws IOException {
     LOGGER.info("Save Interpreter Setting to " + interpreterSettingPath.getAbsolutePath());
-    writeToFile(settingInfos.toJson(), interpreterSettingPath);
+    atomicWriteToFile(settingInfos.toJson(), interpreterSettingPath);
   }
 
   @Override
@@ -68,7 +69,7 @@ public class LocalConfigStorage extends ConfigStorage {
   @Override
   public void save(NotebookAuthorizationInfoSaving authorizationInfoSaving) throws IOException {
     LOGGER.info("Save notebook authorization to file: " + authorizationPath);
-    writeToFile(authorizationInfoSaving.toJson(), authorizationPath);
+    atomicWriteToFile(authorizationInfoSaving.toJson(), authorizationPath);
   }
 
   @Override
@@ -95,17 +96,21 @@ public class LocalConfigStorage extends ConfigStorage {
   @Override
   public void saveCredentials(String credentials) throws IOException {
     LOGGER.info("Save Credentials to file: " + credentialPath);
-    writeToFile(credentials, credentialPath);
+    atomicWriteToFile(credentials, credentialPath);
   }
 
   private String readFromFile(File file) throws IOException {
     return IOUtils.toString(new FileInputStream(file));
   }
 
-  private void writeToFile(String content, File file) throws IOException {
-    FileOutputStream out = new FileOutputStream(file);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    String fileName = file.getName() + ".tmp";
+    File tempFile = new File(directory, fileName);
+    FileOutputStream out = new FileOutputStream(tempFile);
     IOUtils.write(content, out);
     out.close();
+    FileUtils.moveFile(tempFile, file);
   }
 
 }
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
@@ -18,7 +18,6 @@
 package org.apache.zeppelin.storage;
 
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.io.FileUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.interpreter.InterpreterInfoSaving;
 import org.apache.zeppelin.notebook.NotebookAuthorizationInfoSaving;
@@ -29,7 +28,11 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.FileSystems;
+import java.nio.file.FileSystem;
+import java.nio.file.StandardCopyOption;
 
 /**
  * Storing config in local file system
@@ -105,12 +108,15 @@ public class LocalConfigStorage extends ConfigStorage {
 
   private void atomicWriteToFile(String content, File file) throws IOException {
     File directory = file.getParentFile();
-    String fileName = file.getName() + ".tmp";
-    File tempFile = new File(directory, fileName);
+    String fileNameParts[] = file.getName().split("\\.");
+    File tempFile = File.createTempFile(fileNameParts[0], fileNameParts[1], directory);
     FileOutputStream out = new FileOutputStream(tempFile);
     IOUtils.write(content, out);
     out.close();
-    FileUtils.moveFile(tempFile, file);
+    FileSystem defaultFileSystem = FileSystems.getDefault();
+    Path tempFilePath = defaultFileSystem.getPath(tempFile.getCanonicalPath());
+    Path destinationFilePath = defaultFileSystem.getPath(file.getCanonicalPath());
+    Files.move(tempFilePath, destinationFilePath, StandardCopyOption.ATOMIC_MOVE);
   }
 
 }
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +96,21 @@ public String loadCredentials() throws IOException {
...
-  private void writeToFile(String content, File file) throws IOException {
-    FileOutputStream out = new FileOutputStream(file);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    String fileName = file.getName() + ".tmp";
+    File tempFile = new File(directory, fileName);
+    FileOutputStream out = new FileOutputStream(tempFile);
     IOUtils.write(content, out);
     out.close();
+    FileUtils.moveFile(tempFile, file);
```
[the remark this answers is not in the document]
sanjaydasgupta: @Tagar thanks for pointing this out.

The destination file will of course exist, so this feature of `org.apache.commons.io.FileUtils.moveFile(...)` seems to be very inconvenient for this use. We are after all trying to remove the old config file and replace it with the new config file.

Perhaps this calls for `java.io.File.renameTo(...)`? 

Let me know what you think.

Thanks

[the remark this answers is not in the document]
Tagar: Should we use https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move%28java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...%29 
with ATOMIC_MOVE instead?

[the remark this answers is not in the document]
sanjaydasgupta: Done. Also used more formal way of creating the temporary file.

Thanks for the insight.

sanjaydasgupta:
Done. Also used more formal way of creating the temporary file. 

I've tested the integrated code on Ubuntu 16.04 - success case only :-) but am not sure how things will work on Windows. Much of the confusion over atomicity seems based on the different platform-specific behaviors (UNIX vs Windows) of the _rename_ function in the underlying C Standard library.

Thanks for the insight @Tagar

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,24 @@ public String loadCredentials() throws IOException {
...
   private String readFromFile(File file) throws IOException {
     return IOUtils.toString(new FileInputStream(file));
   }
 
-  private void writeToFile(String content, File file) throws IOException {
-    FileOutputStream out = new FileOutputStream(file);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    String fileNameParts[] = file.getName().split("\\.");
+    File tempFile = File.createTempFile(fileNameParts[0], fileNameParts[1], directory);
```
felixcheung: when does the temp file get removed, if the write fails?

sanjaydasgupta: Fixed. Please see comments in body of PR.

sanjaydasgupta:
Added code to delete the temp file if the write fails. Re-throwing the exception serves two purposes:

1) Ensure that the caller sees the same exception when the write fails, and
2) Prevent calling the subsequent rename operation (which would fail)

Thanks for pointing out this case @felixcheung

Tagar:
Thanks @sanjaydasgupta. I will give this a try today.

Tagar:
What's difference between LocalConfigStorage and FileSystemConfigStorage?
I don't know Zeppelin internals that well.
Wonder if this has to be fixed FileSystemConfigStorage.java as well?

https://github.com/sanjaydasgupta/zeppelin/blob/36698a9a6a45610b409fa456f94255b32b746db8/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/FileSystemConfigStorage.java#L66

zjffdu:
LocalConfigStorage is the old implantation of config storage just for backward compatibility. FileSystemConfigStorage is for hadoop compatible filesystem which is new in 0.8

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,29 @@ public String loadCredentials() throws IOException {
...
+    FileOutputStream out = new FileOutputStream(tempFile);
+    try {
+      IOUtils.write(content, out);
+    } catch (IOException iox) {
+      tempFile.delete();
+      throw iox;
+    }
     out.close();
```
felixcheung: thanks, how about:

```
    File tempFile = File.createTempFile(fileNameParts[0], fileNameParts[1], directory);
    FileOutputStream out = null;
    try {
      out = new FileOutputStream(tempFile);
      IOUtils.write(content, out);
    } catch (IOException iox) {
      if (out != null) out.close();
      tempFile.delete();
      throw iox;
    }
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,29 @@ public String loadCredentials() throws IOException {
...
   private String readFromFile(File file) throws IOException {
     return IOUtils.toString(new FileInputStream(file));
   }
 
-  private void writeToFile(String content, File file) throws IOException {
-    FileOutputStream out = new FileOutputStream(file);
-    IOUtils.write(content, out);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    String fileNameParts[] = file.getName().split("\\.");
```
felixcheung: isn't this split on `\` platform specific?

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,29 @@ public String loadCredentials() throws IOException {
...
+      IOUtils.write(content, out);
+    } catch (IOException iox) {
+      tempFile.delete();
+      throw iox;
+    }
     out.close();
+    FileSystem defaultFileSystem = FileSystems.getDefault();
+    Path tempFilePath = defaultFileSystem.getPath(tempFile.getCanonicalPath());
+    Path destinationFilePath = defaultFileSystem.getPath(file.getCanonicalPath());
+    Files.move(tempFilePath, destinationFilePath, StandardCopyOption.ATOMIC_MOVE);
```
felixcheung: in my earlier question - what if move throws? should it clean up the tempFilePath?

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
sanjaydasgupta: Agreed, done.

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
sanjaydasgupta: The original idea was to split on the dot separator between the filename and the `json` type at the end of the name. It turns out that this separation (hence _split_) is not necessary as the following invocation of `File.createTempFile(...)` is perfectly happy with the entire filename (including separator and type) passed in one chunk.

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
sanjaydasgupta: I've added a try-catch here.

sanjaydasgupta:
@felixcheung have made the suggested changes. But am not sure how to add tests for this. Any hints will be useful.

@zjffdu, @Tagar do you recommend similar treatment for FileSystemConfigStorage also?

Tagar:
@sanjaydasgupta 

> do you recommend similar treatment for FileSystemConfigStorage also?

yep, HDFS although is not a posix filesystem, but renaming a file is still implemented as an atomic operation. 

might want to consider moving atomicWriteToFile() to ConfigStorage so both LocalConfigStorage  and FileSystemConfigStorage can use it.

sanjaydasgupta:
Hi @zjffdu 

While reading the code of `FileSystemConfigStorage.java`, I noticed that the method `save(...)` delegates its work to `FileSystemStorage.writeFile(...)`. Interestingly, this `writeFile(...)` method has an _unused_ boolean parameter named `writeTempFileFirst`. 

My guess is that this unused boolean parameter `writeTempFileFirst` is in fact a _ToDo_ that has been left here with the intention of adding the same kind of protective logic that we are discussing here.

Please let me know if that is the case, and if so whether I can go ahead and implement the required logic in the `FileSystemStorage.writeFile(...)` method.

Thanks for your help.

sanjaydasgupta:
Hi @zjffdu, can you please confirm if anything more remains to be done in [FileSystemStorage.writeFile](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java#L146)?

Thanks for your insight.

zjffdu:
There's nothing needs to be done in FileSystemStorage which has already use tmp file

sanjaydasgupta:
Hi @felixcheung, @Tagar 

Any suggestions on how to push this forward from here? 

Thanks

Tagar:
@sanjaydasgupta, `writeFile` in `FileSystemStorage` drops file first, and then renames temp file. 
It's not an atomic write, and also leaves a chance to loosing file that is being written. 
I'd say approach in this PR is more formalized and a better way to go.

https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java#L155

sanjaydasgupta:
There is a problem in my repo that needs correction.

I am closing this PR now, and will reopen it later if possible, or return with a new PR.

PR #2980, opened 2018-05-23
ZEPPELIN-3489. Yarn cluster mode doesn't work for multiple node cluster

zjffdu:
### What is this PR for?
Change the scope of zeppelin-interpreter to compile, so that it will be packaged into the spark interpreter jar

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3489

### Questions:

### DIFF
--- a/interpreter-parent/pom.xml
+++ b/interpreter-parent/pom.xml
@@ -38,7 +38,6 @@
         <groupId>${project.groupId}</groupId>
         <artifactId>zeppelin-interpreter</artifactId>
         <version>${project.version}</version>
-        <scope>provided</scope>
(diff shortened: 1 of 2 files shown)


zjffdu:
@jongyoul Can you help review and verify it ?

jongyoul:
minor comment. This patch increases the interpreter packages about 10M each. In the next release, we might find a better solution to handle this issue.

jongyoul:
Let's merge it into master and 0.8 branch as well.

PR #2984, opened 2018-05-24
ZEPPELIN-3497. Improve error message when livy session is failed to create

zjffdu:
### What is this PR for?
The current behavior we will only return 10 lines of log to frontend which is not enough for user. This PR would invoke another livy api to get more logs (by default it is 1000 lines)

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3497

* Manually tested

Before
screen shot 2018-05-24 at 2 09 09 pm

After
screen shot 2018-05-24 at 4 25 51 pm

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
@@ -94,6 +94,7 @@ public abstract class BaseLivyInterpreter extends Interpreter {
   private String livyURL;
   private int sessionCreationTimeout;
   private int pullStatusInterval;
+  private int maxLogLines;
   protected boolean displayAppInfo;
   private boolean restartDeadSession;
   protected LivyVersion livyVersion;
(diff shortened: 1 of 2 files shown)


zjffdu:
@prabhjyotsingh @felixcheung Could you help review it ? Thanks

livy/src/main/java/org/apache/zeppelin/livy/BaseLivyInterpreter.java
```diff
@@ -362,6 +365,11 @@ private SessionInfo getSessionInfo(int sessionId) throws LivyException {
     return SessionInfo.fromJson(callRestAPI("/sessions/" + sessionId, "GET"));
   }
 
+  private SessionLog getSessionLog(int sessionId) throws LivyException {
+    return SessionLog.fromJson(callRestAPI("/sessions/" + sessionId + "/log?size=" + maxLogLines,
+        "GET"));
```
felixcheung: is there a way to indicate the result is truncated?

livy/src/main/resources/interpreter-setting.json
```diff
@@ -97,6 +97,12 @@
         "description": "The interval for checking paragraph execution status",
         "type": "number"
       },
+      "zeppelin.livy.maxLogLines": {
+        "propertyName": "zeppelin.livy.maxLogLines",
+        "defaultValue": "1000",
+        "description": "Max number of lines of logs",
```
felixcheung: document this?

PR #2986, opened 2018-05-25
ZEPPELIN-2035. BI directional RPC framework between ZeppelinServer and InterpreterProcess on top of thrift

zjffdu:
### What is this PR for?

This PR introduce the bi directional RPC framework between ZeppelinServer and InterpreterProcess on top of thrift, see the design doc here. https://docs.google.com/document/d/1B_4b7UuvE7RwbfpTjxV2RSxdDeF-p9vXVYC8ace76-Q/edit?usp=sharing

This PR is pretty big, most of the code changes is due to regeneration of thrift code. 
Another improvement is introduce the builder pattern of InterpreterContext as the arguments of InterpreterContext constructor is too many. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2035

* CI pass

### Questions:


Leemoonsoo:
Thanks @zjffdu for the improvement. 

Overall looks good to me. I'd like to see CI goes green before merge while it change code a lot.

One question. Let's say in the future, zeppelin have capability to run interpreter as a container in the remote machine for scalability. For example, run interpreter in in k8s cluster or docker host in remote machine. Initiating communication between ZeppelinServer and interpreter will require firewall setup, port export/mapping on container service.

What would be the simplest connectivity model in this case?

I guess initiating one connection from Interpreter -> ZeppelinServer is simplest way? Because of interpreter running in container on remote host doesn't need to allow incoming connection (no firewall, container port mapping). Only single known port of ZeppelinServer need to accept the connection from all interpreters running remotely. Will there be other connectivity model keep network configuration simple?

With the consideration with connectivity model that minimize user's network configuration, what would be the future direction of bidirectional rpc between ZeppelinServer and Interpreter?

zjffdu:
There's 2 properties (`zeppelin.server.rpc.portRange` and `zeppelin.interpreter.rpc.portRange`) that could be used to constrain the portRange of zeppelin server rpc and interpreter process rpc. And user can config these 2 properties to use a fixed port so that user can just expose this fixed port on containers.  @Leemoonsoo  Is this what you concern ?

Leemoonsoo:
I was thinking the cases like a (or more) firewalls, NATs between Server and Interpreters.
And accepting range of ports on each side is more difficult to configure than accepting a single port in one side. 

But for now, Interpreter runs in the same host where Server runs. So not a concern unless Zeppelin have ability to run Interpreter in remote machines.

zjffdu:
For now yarn-cluster mode of spark interpreter is the only use case that run interpreter process in remote host. And we use property to config the portRange for the remote spark interpreter connect back to zeppelin server. In future we may runt all the other interpreters in containers which is another scenario of running interpreter in remote host. For this scenario, I think it is easy to expose one port when starting this container which is not difficult IIUC based on my limited knowledge on containers. BI directional RPC is very common pattern in distributed computing, I think it is worth to use Bi directional RPC in zeppelin with regarding its benefits mentioned int he design doc.  @Leemoonsoo  What do you think ?

zjffdu:
Thanks @Leemoonsoo Any more comment ？

zjffdu:
@Leemoonsoo If no more comments, I will merge it to continue the rest of the work

PR #2988, opened 2018-05-28
ZEPPELIN-3506. DepInterpreter is broken

zjffdu:
### What is this PR for?
The bug is due to getInterpreterInTheSameSessionByClassName doesn't find the correct DepInterpreter. This PR fix this issue. The unit test fails due to classpath issue, will enable it later.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3506

* CI pass and manually tested 

screen shot 2018-05-28 at 11 49 33 am

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/AbstractSparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/AbstractSparkInterpreter.java
@@ -31,6 +31,8 @@ import java.util.Properties;
  */
 public abstract class AbstractSparkInterpreter extends Interpreter {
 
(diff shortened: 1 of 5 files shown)


alonshoham:
I am trying to upgrade Zeppelin to 0.8.2 and having problems with depInterpreter. I am loading jars with depInterpreter and using them as dependencies in a spark paragraph. The paragraph fails beacause it does not have the jars in its classpath. By setting the "zeppelin.spark.useNew" property to false the spark paragraph succeeds. Could this be related also to [this](https://github.com/apache/zeppelin/pull/3353) change in 0.8.2?

PR #2990, opened 2018-05-29
[HOTFIX] Don't copy dependencies to target/lib folder

zjffdu:
### What is this PR for?
It is not necessary to copy dependencies jars to target/lib folder

### What type of PR is it?

### Todos
* [ ] - Task

### What is the Jira issue?
* No jira created

* CI pass

### Questions:

### DIFF
--- a/interpreter-parent/pom.xml
+++ b/interpreter-parent/pom.xml
@@ -78,7 +78,7 @@
           <artifactId>maven-dependency-plugin</artifactId>
           <executions>
             <execution>
-              <id>copy-interpreter-dependencies</id>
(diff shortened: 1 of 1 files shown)


interpreter-parent/pom.xml
```diff
@@ -78,7 +78,7 @@
           <artifactId>maven-dependency-plugin</artifactId>
           <executions>
             <execution>
-              <id>copy-interpreter-dependencies</id>
+              <id>copy-dependencies</id>
```
zjffdu: Just override the id of parent pom file

PR #2993, opened 2018-05-31
ZEPPELIN-3517 Remove zeppelin.pyspark.python in PySparkInterpreter

zjffdu:
### What is this PR for?
`zeppelin.pyspark.python` should be removed as it is zeppelin specific property, and only affect the driver, but not on executor. So we should use spark property instead. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3517

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
@@ -180,17 +180,22 @@ public class PySparkInterpreter extends PythonInterpreter {
 
   // Run python shell
   // Choose python in the order of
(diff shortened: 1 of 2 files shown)


r-kamath:
LGTM

zjffdu:
@felixcheung The only place to do mapping is in zeppelin server side, and I it is not to put any interpreter specific login in zeppelin server side. And I think it's better for zeppelin to use the same configuration as spark instead of introducing zeppelin specific configuration.

zjffdu:
Thanks @felixcheung . I still think it is better to remove it. As it is difficult to deprecate this and let users to aware of that. Zeppelin is a notebook instead of library. If we don't remove it and just deprecate it in documentation, I believe most of users won't be know that and would continue to use it and may ask question in user mail list why they hit python version mismatch issue between driver and executors.

PR #3000, opened 2018-06-01
[ZEPPELIN-3467] two-step, atomic configuration file 

sanjaydasgupta:
### What is this PR for?
This PR continues the discussion and implementation in [PR-2978](https://github.com/apache/zeppelin/pull/2978) which had to be closed earlier.

This PR prevents the total loss of configuration information that happens when zeppelin attempts to write out modified configuration information into an already full file-system. The approach taken here is to first write the modified information to a temporary file (in the same directory), and then to rename the successfully written file to its eventual name.

The rename operation is not attempted if there is an error while writing the temporary file, thus leaving the pre-existing configuration file untouched.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3467

CI Pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
@@ -28,7 +28,11 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.FileSystems;
+import java.nio.file.FileSystem;
+import java.nio.file.StandardCopyOption;
 
 /**
  * Storing config in local file system
@@ -51,7 +55,7 @@ public class LocalConfigStorage extends ConfigStorage {
   @Override
   public void save(InterpreterInfoSaving settingInfos) throws IOException {
     LOGGER.info("Save Interpreter Setting to " + interpreterSettingPath.getAbsolutePath());
-    writeToFile(settingInfos.toJson(), interpreterSettingPath);
+    atomicWriteToFile(settingInfos.toJson(), interpreterSettingPath);
   }
 
   @Override
@@ -68,7 +72,7 @@ public class LocalConfigStorage extends ConfigStorage {
   @Override
   public void save(NotebookAuthorizationInfoSaving authorizationInfoSaving) throws IOException {
     LOGGER.info("Save notebook authorization to file: " + authorizationPath);
-    writeToFile(authorizationInfoSaving.toJson(), authorizationPath);
+    atomicWriteToFile(authorizationInfoSaving.toJson(), authorizationPath);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


felixcheung:
@zjffdu

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,34 @@ public String loadCredentials() throws IOException {
...
-    FileOutputStream out = new FileOutputStream(file);
-    IOUtils.write(content, out);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    File tempFile = File.createTempFile(file.getName(), null, directory);
+    FileOutputStream out = new FileOutputStream(tempFile);
+    try {
+      IOUtils.write(content, out);
+    } catch (IOException iox) {
+      if (out != null) out.close();
```
zjffdu: out should be never null since it is created out of the try catch block

zjffdu:
One minor comment, rest LGTM

sanjaydasgupta:
Agreed @zjffdu.

I also made two other (similar) changes. If the `tempFile.delete()` fails, then `tempFile.deleteOnExit()` is invoked to request the system to handle it if possible on exit.

Thanks.

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
```diff
@@ -95,17 +99,35 @@ public String loadCredentials() throws IOException {
...
-    IOUtils.write(content, out);
+  private void atomicWriteToFile(String content, File file) throws IOException {
+    File directory = file.getParentFile();
+    File tempFile = File.createTempFile(file.getName(), null, directory);
+    FileOutputStream out = new FileOutputStream(tempFile);
+    try {
+      IOUtils.write(content, out);
+    } catch (IOException iox) {
+      if (!tempFile.delete())
+        tempFile.deleteOnExit();
```
Tagar: deleteOnExit() triggers only on jvm exit, and I'm not sure if that's the case here .. 
https://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()
"Deletion will be attempted only for normal termination of the virtual machine, as defined by the Java Language Specification. Once deletion has been requested, it is not possible to cancel the request. This method should therefore be used with care."
I would ignore if temp file deletion failed.. not a big deal. We will overwrite that anyway in next write operation attempt.

sanjaydasgupta:
@Tagar thanks for your comments.

The thought here was to make the best of a bad situation. If and when the `delete()` fails, no recovery action is possible at the user level. So the only further action left is to notify the run-time-library that we are no longer interested in the file, and that the RTL should try to delete it if possible at some logical point in time (process exit). 

The orphaned file will not be automatically overwritten later because `File.createTempFile(...)` will always attempt to create a new filename that is unused in the directory.

This is probably an obscure corner case without major negative consequences when handled either way. It is also likely impossible to test in a repeatable way. 

My only motivation for this addition was to silence the Intellij IDE's warning that the return value of `delete()` had been ignored, but I think there are no technical errors in the overall logic.

Tagar:
> The orphaned file will not be automatically overwritten later because File.createTempFile(...) will always attempt to create a new filename that is unused in the directory.

That makes now. Thanks for the detailed response.

sanjaydasgupta:
ping @felixcheung

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
felixcheung: style - I'd put single line statement in `{}`
like
```
if (!tempFile.delete()) {
        tempFile.deleteOnExit();
}
```

sanjaydasgupta:
Have fixed the style issues @felixcheung.

One of the tests is failing--despite rebase and restart--but appears to be unrelated to the code change.

zeppelin-zengine/src/main/java/org/apache/zeppelin/storage/LocalConfigStorage.java
sanjaydasgupta: Done. Thanks for spotting these issues @felixcheung

sanjaydasgupta:
I've fixed the style issues pointed out.

One test still fails on travis--despite a rebase and restart. But the failure seems unrelated to the changes made in this PR.

jongyoul:
@sanjaydasgupta Could you please rebase this PR from the current master? merging master results as same but it's not human readable.

sanjaydasgupta:
@jongyoul the merge occurred by mistake, and I am not being able to revert it using the `git revert -m 1 <commit>` command. 

I've tried looking up all help on the net, but am not getting anywhere. Please let me know if you have any ideas that may help.

I can create a clean new PR if that will be better.

sanjaydasgupta:
@jongyoul please ignore my previous comments (now deleted)

The rebase has been done.

jongyoul:
Thanks. BTW, please check your CI and restart the failed ones

https://travis-ci.org/sanjaydasgupta/zeppelin/builds/401742347

sanjaydasgupta:
Hi @jongyoul one test is still red. I have restarted it 3 times, but it has failed with the same result each time. Please let me know if there is anything else to try.

zjffdu:
@sanjaydasgupta If you see the following error in travis build. Please clean travis cache first, then retrigger the build. 

```
[ERROR] error: error while loading <root>, error in opening zip file
[ERROR] error: scala.reflect.internal.MissingRequirementError: object scala.runtime in compiler mirror not found.
[ERROR] 	at scala.reflect.internal.MissingRequirementError$.signal(MissingRequirementError.scala:16)
```

sanjaydasgupta:
Cleaning the cache helped. Thanks for the hint @zjffdu 

All green now @jongyoul

PR #3001, opened 2018-06-01
[ZEPPELIN-3524] Replace the Amazon json libs with org.json

arminwasicek:
### What is this PR for?
Currently, Zeppelin uses a JSON library that Amazon deprecated in subsequent releases. This PR replaces Amazon's wrapped library with the original from org.json.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Testing

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3524

Compiles locally, no unit test performed.

### Questions:
No, license is "The Software shall be used for Good, not Evil."

This should be backward compatible

Nope, this is just replacing a wrapped library with its original.


zeppelin-web/package-lock.json
```diff
@@ -154,21 +154,23 @@
...
+        "angular": {
+          "version": "1.6.10",
+          "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.10.tgz",
+          "integrity": "sha512-PCZ5/hVdvPQiYyH0VwsPjrErPHRcITnaXxhksceOXgtJeesKHLA7KDu4X/yvcAi+1zdGgGF+9pDxkJvghXI9Wg=="
+        }
       }
     },
     "angular-viewport-watch": {
-      "version": "github:shahata/angular-viewport-watch#182923b3934e63817b6fc7b640ecb5c4a011f74c"
+      "version": "github:wix/angular-viewport-watch#182923b3934e63817b6fc7b640ecb5c4a011f74c"
```
felixcheung: why these changes?

zeppelin-zengine/pom.xml
```diff
@@ -257,6 +258,12 @@
       <version>1.5</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>${org.json.version}</version>
```
felixcheung: is this the code/license?

https://github.com/stleary/JSON-java/blob/master/LICENSE

sounds like we need to include in the LICENSE file

zeppelin-web/package-lock.json
arminwasicek: These are unintended. I'll remove them from the PR.

zeppelin-zengine/pom.xml
arminwasicek: Thanks, I'll add that.

jongyoul:
BTW, this module already has `gson` as a dependency. Do you have any reason adding it?

jongyoul:
@arminwasicek ping

arminwasicek:
The story is that Amazon deprecated their use of json.org and moved to jackson. The Zeppelin code in question uses on json.org through the import of the Amazon lib. The change I'm proposing is to explicitly include json.org, so everything else can stay the same. 

Using gson (or jackson via amazon for that matter) would make sense, but this entails changing more code.

jongyoul:
Sorry for the late reply. As you mentioned, Zeppelin already has two or more json dependencies at the same time, but I don't think it's good and we can fix it when we can do and have a chance. How about fixing it in this time?

Reamer:
Amazon's JSON library has been replaced by org.codehaus.jettison.

PR #3002, opened 2018-06-04
ZEPPELIN-3520. Use the first one as the default value of dynamic form dropdown list

zjffdu:
### What is this PR for?
This PR would use the first one as the default value of dynamic form dropdown list. Otherwise user will get error when the return value of null. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3520
* CI pass

### Questions:

### DIFF
--- a/zeppelin-integration/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
+++ b/zeppelin-integration/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
@@ -571,7 +571,7 @@ public class ParagraphActionsIT extends AbstractZeppelinIT {
       waitForParagraph(1, "FINISHED");
       collector.checkThat("Output text should not display any of the options in select form",
               driver.findElement(By.xpath(getParagraphXPath(1) + "//div[contains(@class, 'text plainTextContent')]")).getText(),
-              CoreMatchers.equalTo("Howdy "));
+              CoreMatchers.equalTo("Howdy 1"));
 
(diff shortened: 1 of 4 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/display/GUI.java
```diff
@@ -91,11 +91,15 @@ public Object textbox(String id) {
   }
 
   public Object select(String id, Object defaultValue, ParamOption[] options) {
+    if (defaultValue == null) {
+      defaultValue = options[0].getValue();
```
felixcheung: check if options is empty?

zjffdu: Fixed

zeppelin-integration/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
```diff
@@ -674,7 +674,7 @@ public void testMultipleDynamicFormsSameType() throws Exception {
     }
   }
 
-  @Test
+//  @Test
```
felixcheung: are these tests broken? where are they disabled?

zjffdu: oops, I mix this PR with another PR for ZEPPELIN-3535. Will revert these changes

zjffdu:

PR #3003, opened 2018-06-05
[ZEPPELIN-3526] Zeppelin auth mechanisms (LDAP or password based) should be mutually exclusive

prabhjyotsingh:
### What is this PR for?
Problem:
When any external authentication (like LDAP/AD) is enabled for Zeppelin, the default password-based authentication could still be configured in addition to that. This makes space for backdoor in Zeppelin where the user can still get in using the local username/password.

Proposed Solution:
Zeppelin shouldn't allow specifying [users] section in shiro.ini when it is configured to authenticate with LDAP/AD.

### What type of PR is it?
[Bug Fix | Feature ]

### Todos
* [x] - Add documentation 

### What is the Jira issue?
* [ZEPPELIN-3526](https://issues.apache.org/jira/browse/ZEPPELIN-3526)

If both [users] and [main] for example activeDirectoryRealm section enabled in shiro, Zeppelin server should not start.

### DIFF
--- a/docs/setup/security/shiro_authentication.md
+++ b/docs/setup/security/shiro_authentication.md
@@ -104,6 +104,9 @@ To learn more about Apache Shiro Realm, please check [this documentation](http:/
 
 We also provide community custom Realms.
 
+**Note**: When using any of the below realms the default 
+      password-based (IniRealm) authentication needs to be disabled.
+
 ### Active Directory
 
 ```
@@ -267,6 +270,7 @@ If you want to grant this permission to other users, you can change **roles[ ]**
 
 ### Apply multiple roles in Shiro configuration
 By default, Shiro will allow access to a URL if only user is part of "**all the roles**" defined like this:
+
 ```
 [urls]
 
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
@Leemoonsoo @felixcheung @zjffdu  can you help review this.

zjffdu:
LGTM, one suggestion is that I think it is time for us to do refactoring for the authentication component, we should put it in one component and provide interface to be used by other components. Instead of putting logic in zeppelin server like this PR.

prabhjyotsingh:
Agreed we should definitely do something like zeppelin-plugins for authentication components as well in Zeppelin-0.9

mebelousov:
@prabhjyotsingh I think it's good to document only one way for authentification. What do you think about it?

docs/setup/security/shiro_authentication.md
```diff
@@ -267,13 +268,15 @@ If you want to grant this permission to other users, you can change **roles[ ]**
 
 ### Apply multiple roles in Shiro configuration
 By default, Shiro will allow access to a URL if only user is part of "**all the roles**" defined like this:
+
```
prabhjyotsingh: Added empty lines for http://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/setup/security/shiro_authentication.html#apply-multiple-roles-in-shiro-configuration

prabhjyotsingh:
@mebelousov sure I've added a doc.

PR #3004, opened 2018-06-05
[ZEPPELIN-2913] Support for both user and role for AuthorizationFilter

prabhjyotsingh:
### What is this PR for?
Currently, Zeppelin only supports roles for AuthorizationFilter, but there can be a condition as described in https://issues.apache.org/jira/browse/ZEPPELIN-2913 where Zeppelin's user does not belong to a group/role, and the administrator wants to have control using user only.

### What type of PR is it?
[Feature]

### Todos
* [x] - Add documentation

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2913 

add the following in shiro.ini:
```
[main]
...
anyofroles = org.apache.zeppelin.utils.AnyOfRolesUserAuthorizationFilter

[urls]
...
/api/interpreter/** = authc, anyofroles[admin, user1]
/api/configurations/** = authc, roles[admin]
/api/credential/** = authc, roles[admin]
```
With the above config both user (user1) and users the belong to role admin will have access to interpreter setting page.


prabhjyotsingh:
@zjffdu @VipinRathor  can you help review this PR

zeppelin-server/src/main/java/org/apache/zeppelin/utils/AnyOfRolesUserAuthorizationFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.shiro.subject.Subject;
+import org.apache.shiro.web.filter.authz.RolesAuthorizationFilter;
+
+/**
+ * Allows access if current user has at least one role of the specified list.
+ * <p>
+ * Basically, it's the same as {@link RolesAuthorizationFilter} but using {@literal OR} instead
+ * of {@literal AND} on the specified roles or user.
+ */
+public class AnyOfRolesUserAuthorizationFilter extends RolesAuthorizationFilter {
```
VipinRathor: @prabhjyotsingh how about updating AnyOfRolesAuthorizationFilter class instead of adding a new AnyOfRolesUserAuthorizationFilter?

prabhjyotsingh: Sure, I've made the suggested changes.

VipinRathor: thanks!

VipinRathor:
LGTM

PR #3005, opened 2018-06-05
[ZEPPELIN-3528] Ordering by interpreter name in same interpreter group

woowahan-jaehoon:
### What is this PR for?
In 'Create new note' diagram, 'Default Interpreter' is ordered by only interpreter groups.

But interpreter is not ordered in same interpreter group.
(It seems to be ordered by hashcode of 'InterpreterSetting')

When there are many interpreter in same group and this group is first,
user cannot predict what interpreter is showed first.

So I think that interpreter need to be ordered by name.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Modify Code

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3528

* Click 'create new note'

### Questions:
    * No
    * No
    * No


zjffdu:
LGTM

PR #3006, opened 2018-06-05
[ZEPPELIN-3529] Add uName into jobGroupId for checking user

woowahan-jaehoon:
### What is this PR for?
Need to check user in jobGroupId
because when some paragraph is running,
who run paragraph cannot be found in spark application web UI.

I just add to userId.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Modify Code

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3529

* Check jobGroupId in Spark Application Web UI of Zeppelin Spark Interpreter

### Questions:
    * No
    * No
    * No

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
@@ -149,7 +149,11 @@ class Utils {
   }
   
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    String uName = "anonymous";
+    if (context.getAuthenticationInfo() != null) {
+      uName = getUserName(context.getAuthenticationInfo());
+    }
+    return "zeppelin-" + uName + "-" + context.getNoteId() + "-" + context.getParagraphId();
   }
 
   public static String buildJobDesc(InterpreterContext context) {


spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -149,7 +149,7 @@ static boolean isSpark2() {
   }
   
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    return context.getAuthenticationInfo().getUser() + "-" + context.getNoteId() + "-" + context.getParagraphId();
```
zjffdu: Is it possible `context.getAuthenticationInfo()` is null or `context.getAuthenticationInfo().getUser()` is null

woowahan-jaehoon:
@felixcheung, you're right. I modified.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -149,7 +149,12 @@ static boolean isSpark2() {
   }
   
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    String userIdAndHyphen = "";
+    if (context.getAuthenticationInfo() != null && context.getAuthenticationInfo().getUser() != null) {
```
felixcheung: as in [here](https://github.com/apache/zeppelin/blob/master/spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java#L172) perhaps you should check if getUser() returns isEmpty()

woowahan-jaehoon: Oh, I'm sorry. I missed.
Thank you.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -149,7 +149,12 @@ static boolean isSpark2() {
   }
   
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    String userIdAndHyphen = "";
```
zjffdu: The default value should be `anonymous`.

woowahan-jaehoon: I think that if `context.getAuthenticationInfo()` is null, before rule of jobGroupId is better.

So, I didn't write hyphen in 157 line.
Is it better to add userId always?

zjffdu: To me, the user would either be `anonymous` or the real user name. Otherwise there must be some bugs in zeppelin.

woowahan-jaehoon:
It tested successful.
https://travis-ci.org/woowabros/zeppelin/builds/412124390

spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -149,7 +149,14 @@ static boolean isSpark2() {
   }
   
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    String uName = "";
```
felixcheung: maybe you can initialize to `"anonymous"`?

woowahan-jaehoon:
Modified and passed tests.
https://travis-ci.org/woowabros/zeppelin/builds/412253387

woowahan-jaehoon:
Is it not enough?

zjffdu:
Thanks for late response, +1, will merge it soon

PR #3007, opened 2018-06-05
zeppelin 0.8 version pyspark interpreter error when using ipython

lizhenhuan:
### What is this PR for?
zeppelin 0.8 version pyspark interpreter error when using ipython
change python project grpc version to 1.10.0 to fix this bug

### What type of PR is it?
Bug Fix 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3530

### Questions:
NO
NO
Yes, 
https://zeppelin.apache.org/docs/0.8.0-SNAPSHOT/interpreter/python.html#ipython-support
install ipython 
should also install protobuf

### DIFF
--- a/python/pom.xml
+++ b/python/pom.xml
@@ -42,7 +42,7 @@
     </python.test.exclude>
     <pypi.repo.url>https://pypi.python.org/packages</pypi.repo.url>
     <python.py4j.repo.folder>/64/5c/01e13b68e8caafece40d549f232c9b5677ad1016071a48d04cc3895acaa3</python.py4j.repo.folder>
-    <grpc.version>1.4.0</grpc.version>
+    <grpc.version>1.10.0</grpc.version>
     <plugin.shade.version>2.4.1</plugin.shade.version>
   </properties>
 


zjffdu:
What protobuf version do you use ? I think it is due to mismatch between protobuf and grpc

lizhenhuan:
There are two protobuf.version variable in pom.xml
First one in spark/pom.xml is 2.5.0
Second on in spark-dependencies/pom.xml is 2.4.1.
I also install python protobuf by run command : pip install protobuf
python version 2.7.5

zjffdu:
@lizhenhuan the protobuf in spark project is not related. The version you see in pom file is the protobu java version which is not the same as the python package version. I guess you hit this issue because that you install another version of protobuf before installing grpc. Could you try using virtualenv to create a clean enviroment ? You don't need to install protobuf expclitly, it will be installed automatically as the dependency of grpc.

jongyoul:
Let me close it as it's too outdated to handle.

PR #3009, opened 2018-06-06
[ZEPPELIN-3513] Zeppelin Note Creation Failure is not shown on UI

prabhjyotsingh:
### What is this PR for?
When creating a note from Zeppelin UI which fails due to permission denied exception, there is no failure/error message shown to a user on UI which is causing confusion.

This happens only in the case when HDFS is configured as the default storage for storing notebook/config.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3513](https://issues.apache.org/jira/browse/ZEPPELIN-3513)

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -30,7 +30,6 @@ import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.vfs2.FileSystemException;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.conf.ZeppelinConfiguration.ConfVars;
 import org.apache.zeppelin.display.*;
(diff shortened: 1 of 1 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1052,7 +1051,7 @@ private void createNote(NotebookSocket conn, HashSet<String> userAndRoles, Noteb
       note.persist(subject);
       addConnectionToNote(note.getId(), (NotebookSocket) conn);
       conn.send(serializeMessage(new Message(OP.NEW_NOTE).put("note", note)));
-    } catch (FileSystemException e) {
+    } catch (IOException e) {
```
prabhjyotsingh: Changing this to IOException as Hadoop throws org.apache.hadoop.security.AccessControlException

prabhjyotsingh:
@zjffdu @r-kamath please help review this.

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
zjffdu: LGTM

prabhjyotsingh:
Thanks for the review will merge this soon.

PR #3011, opened 2018-06-08
[HOTFIX][ZEPPELIN-3526] fix when no shiro.ini exists

prabhjyotsingh:
### What is this PR for?
This is a side effect of ZEPPELIN-3526, occurs when there's no shiro.ini in the classpath.

### What type of PR is it?
[Hot Fix]

* CI should be green

### Questions:


zjffdu:
LGTM, wait for CI pass

PR #3012, opened 2018-06-09
[SECURITY] Upgrade py4j version to 0.10.7

jongyoul:


zjffdu:
@jongyoul There's more thing needs to be done for upgrading py4j to 0.10.7. py4j 0.10.7 introduce authentication mechanism for connecting gateway server, so we also need to update PySparkInterperter & PythonInterpreter.  I am doing that now, if you don't mind, could you close this PR.

PR #3024, opened 2018-06-15
[ZEPPELIN-3545] save all tables to ResourcePool

Savalek:
### What is this PR for?
Now if paragraph's output contains more than one table in ResourcePool saves only last table. 
It would be desirable that in ResoursePool stores all tables.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-3545](https://issues.apache.org/jira/browse/ZEPPELIN-3545)

### Screenshots
target p
result ps

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
@@ -649,14 +649,17 @@ public class RemoteInterpreterServer extends Thread
         }
         // put result into resource pool
         if (resultMessages.size() > 0) {
-          int lastMessageIndex = resultMessages.size() - 1;
-          if (resultMessages.get(lastMessageIndex).getType() == InterpreterResult.Type.TABLE) {
-            context.getResourcePool().put(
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks @Savalek for this contribution, but I think putting all tables into ResourcePool doesn't make sense. As it would occupy lots of memory. I plan to introduce paragraph level properties (ZEPPELIN-3348), so that user can control whether to put the interpreter result into ResourcePool.

mebelousov:
@zjffdu thank you about ResourcePool impoving.

Share please your vision how it would be. For example, a paragraph has 5 table results. How user will define which of them would be added into ResourcePool?

mebelousov:
@zjffdu 
I support adding only selected table results to Resource Pool.
As paragraph can have multiple results than I propose to add result level properties.

zjffdu:
@mebelousov There're many options for how to specify which result to be stored into resource pool.
e.g.

```
%spark(saveToResourcePool=1,2,4)
```
Or
```
%spark(1.saveToResourcePool=true, 2.saveToResourcePool=true, 4.saveToResourcePool=true)
```

We can discuss more about which is the best approach, the key point here is to allow user to customize it via paragraph levle properties.

PR #3027, opened 2018-06-18
[ZEPPELIN-2605] Import notebook has 1MB size limit but bigger notebooks can be saved without error or warning

sanjaydasgupta:
### What is this PR for?
Notebooks of any size can be exported and saved, but the import facility only accepts notebooks up to 1Mb in size. Allowing notebooks bigger than 1Mb to be saved silently (without any warning or error notification) is therefore a dangerous trap for users.

It has been argued that notebooks bigger than 1Mb are rare and unusual, but a saved notebook also contains the displayed output - and can contain large volumes of tabular data, and even images from matplotlib output. Large notebooks can be common when zeppelin is used for presentation of visual analysis.

This PR adds a popup dialog which is displayed when the user attempts to export a notebook that has a size exceeding 1Mb. A screenshot of the popup is shown below.

popup-dialog

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-2605

CI pass
Manual testing. A notebook that is small when outputs are cleared, but becomes bigger than the limit when all cells are run is included (big-output-tester.json).

[big-output-tester.json.zip](https://github.com/apache/zeppelin/files/2110281/big-output-tester.json.zip)

See above.

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -220,9 +220,29 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
   };
 
   // Export notebook
+  let limit = 0;
+
+  websocketMsgSrv.listConfigurations();
+  $scope.$on('configurationsInfo', function(scope, event) {
+    limit = event.configurations['zeppelin.websocket.max.text.message.size'];
+  });
+
   $scope.exportNote = function() {
     let jsonContent = JSON.stringify($scope.note);
-    saveAsService.saveAs(jsonContent, $scope.note.name, 'json');
+    if (jsonContent.length > limit) {
+      BootstrapDialog.confirm({
+        closable: true,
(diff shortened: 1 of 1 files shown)


mebelousov:
@sanjaydasgupta thank you!
LGTM

zjffdu:
But what is the meaning of this pop up window ? As a user, it's not clear to me what is the impact if I import this kind of notebook

sanjaydasgupta:
Hi Jeff,

The window is popped up only when the user tries to export a notebook that
would be more than 1Mb in size.

The problem with such a notebook (greater than 1Mb in size) is that it is
not allowed to be imported again. So allowing such a big notebook to be
exported without any warning is a bug.

Regards,
- Sanjay

zjffdu:
Thanks @sanjaydasgupta .  LGTM

sanjaydasgupta:
ping @jongyoul

PR #3033, opened 2018-06-19
ZEPPELIN-3551. Upgrade Scala to 2.11.12

zjffdu:
### What is this PR for?
This is Just to update scala to 2.11.12 which to be consistent with spark (SPARK-24418)

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3551

* CI pass

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,13 +86,13 @@ matrix:
     - sudo: required
       jdk: "oraclejdk8"
       dist: trusty
-      env: BUILD_PLUGINS="true" PYTHON="2" PROFILE="-Pspark-2.2" SPARKR="true" BUILD_FLAG="install -DskipTests -DskipRat -am" TEST_FLAG="test -DskipRat -am" MODULES="-pl zeppelin-zengine,zeppelin-server,spark/interpreter,spark/spark-dependencies" TEST_PROJECTS="-Dtest=ZeppelinSparkClusterTest,org.apache.zeppelin.spark.* -DfailIfNoTests=false"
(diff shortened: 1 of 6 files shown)


zjffdu:
@Leemoonsoo @felixcheung Could you help review it ? Thanks

zjffdu:

Leemoonsoo:
LGTM

felixcheung:
specifically in SPARK-24418 - there code changes needed but not release yet (2.4?)

zjffdu:
I tested it against spark master branch, and it works. What kind of problems does spark have on 2.11.12 ?  I notice SPARK-24418 is already merged.

felixcheung:
something in SparkILoop. given that I don't think we can/should upgrade scala broadly - it might only work for Spark 2.4.0 (unreleased) but not older/current releases

zjffdu:
Thanks @felixcheung , I will revert the change in the root pom file and only update scala version in spark module.

felixcheung:
? it looks like spark.version is 2.2.0.
as far as I know SPARK-24418 is fixed only in (unreleased) master branch

zjffdu:
It is to fine to compile with spark 2.2, the main change is on SparkScala211Interpreter.scala which is due to scala repl api changes in scala 2.11.12.

felixcheung:
Perhaps you could elaborate on why we should do this now since Spark doesn’t actually support it? Is there a reason that I have missed?

zjffdu:
This is for the next release of zeppelin which I suppose should supported the latest spark 2.4.0 which would use scala 2.11.12. Just want to catch up with spark.

felixcheung:
well, ok. though might be hard to tell - we are juggling multiple releases 2.1.3, 2.2.2, 2.3.2 - might take some time to get to 2.4.0.

zjffdu:
Anyway, this PR is just for the next release of spark that supports scala 2.11.12, I can hold it if you have any concern

HyukjinKwon:
Hey @zjffdu, busy? I can take this over - looks there's only minor conflict.

zjffdu:
Thanks @HyukjinKwon  feel free to take over this

PR #3039, opened 2018-06-21
ZEPPELIN-3348. Enable paragraph level properties

zjffdu:
### What is this PR for?
This PR is trying to introduce paragraph level properties so that user can customize the interpreter behavior in paragraph level.  User can specify the paragraph level properties after the repl name (e.g. `%spark(pool=p1, imageWidth=200)`)

The jdbc interpreter is another example (like `%jdbc(hive)`, but we use other kind of hack method for that before. This PR also update jdbc interpreter to adopt the paragraph level properties. 

This PR just allow user to specify paragraph level properties and pass them to interpreter. It is interpreter's responsibility to use these paragraph level properties. So there will be some follow up work to customize interpreter at paragraph level. E.g. allow spark interpreter to specify the queue for paragraph. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3348

* CI pass

### Questions:


zjffdu:
@Leemoonsoo @prabhjyotsingh @felixcheung @jongyoul Could you help review it ? Thanks

mebelousov:
@zjffdu great powerful feature!
When the parameters will be a lot, it will be not easy to find and write the parameter name. I wish it would be more user-friendly with drop-down panel.
Also parameters must be explicitly displayed for easy debug as in your case.

Does make sense to add some user functionality?

zjffdu:
Thanks @mebelousov for the feedback. We could do that in a followup PR to improve  user friendliness. Because currently I can not imagine in what kind scenario user would add lots of parameters.  I would be happy to see the real use case for that.

jongyoul:
Hi, basically, parsing properties with every execution may not affect much in performance on interpreter side. But It might affect the behaviours of restarting interpreters. e. g. every restarting while running spark Interpreter with different paragraphs. Does it make sense as well? Otherwise, do you have any further steps to handle this kind of behaviours?

jongyoul:
Specifically in this PR, LGTM though

zjffdu:
Sorry @jongyoul I don't understand that. Why do you think it would affect restarting interpreters ? Do you have any more detail examples ?

jongyoul:
In my guess, you might want to change some spark properties with executing Spark paragraph. In this case, we might need to restart SparkInterpreter to adopt some properties like `spark.executor.memory` and so on.

zjffdu:
No，this is not this ticket's purpose. If user want to change interpreter level properties, they need inline conf interpreter. This ticket is to allow customize paragraph level behavior, e.g. pool of spark job, image size of current paragraph. So user can not specify `spark.executor.memory` as paragraph level properties, because it is interpreter level property.

jongyoul:
@zjffdu Thanks for explaining. I thought it could change all of the properties by this settings.

zjffdu:
Thanks for review, will merge if no more comments

PR #3043, opened 2018-06-25
ZEPPELIN-3218. Plugins for Interpreter Launcher

zjffdu:
### What is this PR for?
Move launcher into zeppelin plugins

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3218

* CI pass

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -665,6 +660,13 @@ public void clearNoteIdAndParaMap() {
     runtimeInfosToBeCleared = null;
   }
 
+  public String getLauncherPlugin() {
```
Leemoonsoo: Cool!  Do you have any idea how can we make InterpreterLauncher user configurable?

zjffdu: I was thinking to make it configurable in `interpreter-setting.json` and user could also customize it in UI. But I havn't nailed down the final design. Would do it in a separated PR after I complete the whole design.

felixcheung: group isn't guaranteed to be "spark" right?

zjffdu: It is always `spark` if it is spark interpreter even when there's multiple spark interpreters. We use field `name` to differentiate them

zjffdu:

PR #3044, opened 2018-06-28
ZEPPELIN-3563. Add pool to paragraph property that use spark interpreter

zjffdu:
### What is this PR for?
Allow user to specify the pool when running spark sql in concurrent approach. 
e.g.
```
%spark.sql(pool=pool_1)

sql statement
```

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3563

* CI pass

### Questions:

### DIFF
--- a/docs/interpreter/spark.md
+++ b/docs/interpreter/spark.md
@@ -121,6 +121,11 @@ You can also set other Spark properties which are not listed in the table. For a
     <td>Execute multiple SQL concurrently if set true.</td>
   </tr>
   <tr>
+    <td>zeppelin.spark.concurrentSQL.max</td>
+    <td>10</td>
+    <td>Max number of SQL concurrently executed</td>
+  </tr>
+  <tr>
     <td>zeppelin.spark.maxResult</td>
     <td>1000</td>
     <td>Max number of Spark SQL result to display.</td>
@@ -332,6 +337,21 @@ utilizing Zeppelin's built-in [Angular Display System](../usage/display_system/a
 
 <img class="img-responsive" src="{{BASE_PATH}}/assets/themes/zeppelin/img/docs-img/matplotlibAngularExample.gif" />
 
+## Running spark sql concurrently
+By default, each sql statement would run sequentially in `%spark.sql`. But you can run them concurrently by following setup.
+
+1. set `zeppelin.spark.concurrentSQL` to true to enable the sql concurrent feature, underneath zeppelin will change to use fairscheduler for spark. And also set `zeppelin.spark.concurrentSQL.max` to control the max number of sql statements running concurrently.
+2. configure pools by creating `fairscheduler.xml` under your `SPARK_CONF_DIR`, check the offical spark doc [Configuring Pool Properties](http://spark.apache.org/docs/latest/job-scheduling.html#configuring-pool-properties)
+3. set pool property via setting paragraph property. e.g.
+
+```
+%spark(pool=pool1)
+
+sql statement
+```
+
+This feature is available for both all versions of scala spark, pyspark. For sparkr, it is only available starting from 2.3.0.
+ 
 ## Interpreter setting option
 
 You can choose one of `shared`, `scoped` and `isolated` options wheh you configure Spark interpreter.
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
@@ -1041,8 +1041,7 @@ public class OldSparkInterpreter extends AbstractSparkInterpreter {
     synchronized (this) {
       z.setGui(context.getGui());
       z.setNoteGui(context.getNoteGui());
-      String jobDesc = "Started by: " + Utils.getUserName(context.getAuthenticationInfo());
-      sc.setJobGroup(Utils.buildJobGroupId(context), jobDesc, false);
+      sc.setJobGroup(Utils.buildJobGroupId(context), Utils.buildJobDesc(context), false);
       InterpreterResult r = interpretInput(lines, context);
       sc.clearJobGroup();
       return r;
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
@@ -123,8 +123,7 @@ public class SparkRInterpreter extends Interpreter {
       throws InterpreterException {
 
     String jobGroup = Utils.buildJobGroupId(interpreterContext);
-    String jobDesc = "Started by: " +
-       Utils.getUserName(interpreterContext.getAuthenticationInfo());
+    String jobDesc = Utils.buildJobDesc(interpreterContext);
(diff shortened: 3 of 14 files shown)


Tagar:
Would it be possible to set `spark.scheduler.pool` to authenticated user name as well, 
as described in https://issues.apache.org/jira/browse/ZEPPELIN-3334
I think it's a much better default than setting it manually in each paragraph. 
(still a great feature of having ability to set pool at a paragraph level too)

zjffdu:
@Tagar What do you mean authenticated user name ? Do you mean impersonation in one spark app ? I don't believe it is possible in spark.

Tagar:
I meant Zeppelin would set `spark.scheduler.pool` to username of the user that was used to authenticate into Zeppelin.

zjffdu:
We could do that but it depends on the `farischeduler.xml` that user specified. If user didn't specify the user name as pool name in `fairscheduler.xml`, then it doesn't make sense to do that in zeppelin side.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
```diff
@@ -1041,8 +1041,7 @@ public InterpreterResult interpret(String[] lines, InterpreterContext context) {
     synchronized (this) {
       z.setGui(context.getGui());
       z.setNoteGui(context.getNoteGui());
-      String jobDesc = "Started by: " + Utils.getUserName(context.getAuthenticationInfo());
-      sc.setJobGroup(Utils.buildJobGroupId(context), jobDesc, false);
+      sc.setJobGroup(Utils.buildJobGroupId(context), Utils.buildJobDesc(context), false);
```
felixcheung: should check if buildJobDesc returns nothing? anonymous?

docs/interpreter/spark.md
```diff
@@ -332,6 +337,19 @@ utilizing Zeppelin's built-in [Angular Display System](../usage/display_system/a
 
 <img class="img-responsive" src="{{BASE_PATH}}/assets/themes/zeppelin/img/docs-img/matplotlibAngularExample.gif" />
 
+## Running spark sql concurrently
+By default, each sql statement would run sequentially in `%spark.sql`. But you can run them concurrently by following setup.
+
+1. set `zeppelin.spark.concurrentSQL` to true to enable the sql concurrent feature. And also set `zeppelin.spark.concurrentSQL.max` to control the max number of sql statements running concurrently.
```
felixcheung: should add that by setting this it would also change to the fair scheduler?

zjffdu: Actually this is a bug of existing zeppelin. `zeppelin.spark.concurrentSQL` won't make spark run under fair scheduler.  What I did in this PR is that if user set `zeppelin.spark.concurrentSQL` to be true, then I will set spark conf `spark.scheduler.mode` to `FAIR` https://github.com/apache/zeppelin/pull/3044/files#diff-421cf195e8f8d49c7ca358768c9f56ceR96

spark/interpreter/src/main/java/org/apache/zeppelin/spark/OldSparkInterpreter.java
zjffdu: It won't be null even it is anonymous, https://github.com/apache/zeppelin/pull/3044/files#diff-25ff02ea9a7b4a31efaf02e8b401ea5bR156

docs/interpreter/spark.md
zjffdu: Sorry, I misunderstand your question. Will add that

Tagar:
> We could do that but it depends on the farischeduler.xml that user specified. If user didn't specify the user name as pool name in fairscheduler.xml, then it doesn't make sense to do that in zeppelin side.

Yep, I understand that. They have to specify a pool name in `farischeduler.xml` either way - if it's given manually through paragraph property as proposed in this PR, or if it's taken from authenticated username as proposed in https://issues.apache.org/jira/browse/ZEPPELIN-3334 .

Thanks.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -88,25 +82,17 @@ public void close() {}
...
     SparkContext sc = sqlc.sparkContext();
-    if (concurrentSQL()) {
-      sc.setLocalProperty("spark.scheduler.pool", "fair");
-    } else {
-      sc.setLocalProperty("spark.scheduler.pool", null);
-    }
-
-    String jobDesc = "Started by: " + Utils.getUserName(context.getAuthenticationInfo());
-    sc.setJobGroup(Utils.buildJobGroupId(context), jobDesc, false);
+    sc.setLocalProperty("spark.scheduler.pool", context.getLocalProperties().get("pool"));
```
Tagar: Do we have to check here if setLocalProperty() is only available from spark 2.3.0?

zjffdu: setLocalProperty is on available for sparkr, but it is available for scala api in very early version.

zjffdu:
@Tagar One approach I can think of is to support variable substitution. 

e.g.

```
%spark(pool=${name})

...
```

Tagar:
Thanks @zjffdu - that looks good to me. What about {username} instead of {name} ?

zjffdu:
Will merge it if no more comments.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkRInterpreter.java
```diff
@@ -156,7 +155,15 @@ public InterpreterResult interpret(String lines, InterpreterContext interpreterC
           "\", \"" + jobDesc + "\", TRUE)";
     }
     lines = setJobGroup + "\n" + lines;
-
+    if (sparkInterpreter.getSparkVersion().newerThanEquals(SparkVersion.SPARK_2_3_0)) {
```
felixcheung: what happens when the spark version is not newer than 2.3.0? would it make the pooling not work?

felixcheung: then, we should document this instead of dropping silently

zjffdu: Sure, will add doc

docs/interpreter/spark.md
```diff
@@ -332,6 +337,21 @@ utilizing Zeppelin's built-in [Angular Display System](../usage/display_system/a
...
+2. configure pools by creating `fairscheduler.xml` under your `SPARK_CONF_DIR`, check the offical spark doc [Configuring Pool Properties](http://spark.apache.org/docs/latest/job-scheduling.html#configuring-pool-properties)
+3. set pool property via setting paragraph property. e.g.
+
+```
+%spark(pool=pool1)
+
+sql statement
+```
+
+This feature is available for both all versions of scala spark, pyspark. For sparkr, it is only available starting from 2.3.0.
```
felixcheung: nit, really would be good to fix capitalization in this doc. same for L343
spark -> Spark
pyspark -> PySpark
sparkr -> sparkR or SparkR

PR #3046, opened 2018-07-02
Minor fix "ipykernel" package name shown in logs when ipython prerequisite isn't met

khalidhuseynov:
### What is this PR for?
minor fix to show correct package name `ipykernel` in logs when prerequisite for ipython interpreter isn't met.

### What type of PR is it?
[ Improvement ]

### Todos
* [x] - fix naming

### What is the Jira issue?

if `ipykernel` isn't installed then it would show correct message in logs when running ipython interpreter

### Questions:


zjffdu:
LGTM

PR #3055, opened 2018-07-06
ZEPPELIN-3587. Interpret paragarph text as whole code statement instead of breaking into lines

zjffdu:
### What is this PR for?

Just to make this kind of code to work.  

```
sc
.range(1,10
.sum()
```

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3587

* CI pass

### Questions:


jongyoul:
I think the most simple solution is to copy from the old interpret method. We discussed several times and fixed several times as well.

zjffdu:
@felixcheung  I am afraid we have to break paragraph text. Zeppelin interpreter is different from scala shell. scala shell will execute the code when user type a complete code, while in zeppelin there may be multiple complete scala code.

e.g. If we type the following code in scala-shell, it will execute `sc.version`, then you can type `1+1`. While in zeppelin spark interpreter, we submit the whole code to spark interpreter which will break it and execute it via scala repl api.

```
sc.version
1+1
```

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -79,7 +79,35 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+    Console.withOut(if (context != null) context.out else Console.out) {
+      interpreterOutput.ignoreLeadingNewLinesFromScalaReporter()
+      val lastStatus = scalaInterpret(code) match {
+        case scala.tools.nsc.interpreter.IR.Success =>
+          InterpreterResult.Code.SUCCESS
+        case error@scala.tools.nsc.interpreter.IR.Error =>
+          InterpreterResult.Code.ERROR
+        case scala.tools.nsc.interpreter.IR.Incomplete =>
+          // add print("") at the end in case the last line is comment which lead to INCOMPLETE
+          scalaInterpret(code + "\nprint(\"\")") match {
```
felixcheung: looks like we are re-running (scalaInterpret) the code in this case? I'm not sure about this - since the code might not be idempotent?

zjffdu: It won't rerun the code,  see the test here. https://github.com/apache/zeppelin/pull/3055/files#diff-b3caeb6afe474aec6d5352ac3ee3a02aR114

zjffdu:
@felixcheung @jongyoul I have updated the PR to run the paragraph text once instead of breaking it into lines. But it leads one behavior change: the order of output is changed compared to previous implementation. Basically there're two kinds of outputs: repl output and code execution output. e.g.g

```
val a = "hello world"
println(a)
```
Output
```
hello world                         (This is code execution output)
a: String = hello world       (This is repl output)
```

Code execution output will be displayed while code is running, but repl output is only displayed after scala repl complete the code execution. That's why you will see that the repl output is always after the code execution output, that's why I update some test code in this PR.

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -79,9 +79,49 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+
+        val status = scalaInterpret(code) match {
+          case success@scala.tools.nsc.interpreter.IR.Success =>
+            success
+          case scala.tools.nsc.interpreter.IR.Error =>
+            if (new String(interpreterOutput.getInterpreterOutput.toByteArray)
+              .contains("value toDF is not a member of org.apache.spark.rdd.RDD")) {
+              // prepend "import sqlContext.implicits._" due to
+              // https://issues.scala-lang.org/browse/SI-6649
+              _interpret("import sqlContext.implicits._\n" + code)
```
zjffdu: It is fine to re-interpret the code, as the Error happens for compilation stage, so the code won't run twice.

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -79,9 +79,49 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+            } else {
+              scala.tools.nsc.interpreter.IR.Error
+            }
+          case scala.tools.nsc.interpreter.IR.Incomplete =>
+            // add print("") at the end in case the last line is comment which lead to INCOMPLETE
+            scalaInterpret(code + "\nprint(\"\")")
```
zjffdu: It is fine to re-interpret the code, as the InComplete happens for compilation stage, so the code won't run twice.

zjffdu:
ping @felixcheung

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
zjffdu: Here's the test. https://github.com/apache/zeppelin/pull/3055/files#diff-b3caeb6afe474aec6d5352ac3ee3a02aR114

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -171,6 +211,9 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
     interpret("import sqlContext.implicits._")
     interpret("import sqlContext.sql")
     interpret("import org.apache.spark.sql.functions._")
+    // print empty string otherwise the last statement's output of this method
+    // (aka. import org.apache.spark.sql.functions._) will mix with the output of user code
+    interpret("print(\"\")")
```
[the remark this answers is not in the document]
zjffdu: println will have output '\n' while print has no side effect in output.

zjffdu:
Actually we already import this when initialize spark interpreter. https://github.com/apache/zeppelin/blob/master/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala#L171

The problem is that we would still hit this kind of error when run the tutorial note of bank data. This is due to an issue scala 2.10 (SI-6649)[https://issues.scala-lang.org/browse/SI-6649]. This error doesn't' happen before just because we split code into lines and interpret it line by line. But as I mentioned before, interpret it line by line is not a correct approach. 
The good news is that it only affect scala 2.10.

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -79,9 +79,49 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+
+        val status = scalaInterpret(code) match {
+          case success@scala.tools.nsc.interpreter.IR.Success =>
+            success
+          case scala.tools.nsc.interpreter.IR.Error =>
+            if (new String(interpreterOutput.getInterpreterOutput.toByteArray)
+              .contains("value toDF is not a member of org.apache.spark.rdd.RDD")) {
+              // prepend "import sqlContext.implicits._" due to
+              // https://issues.scala-lang.org/browse/SI-6649
+              scalaInterpret("import sqlContext.implicits._\n" + code)
```
zjffdu: Fixed

felixcheung: IIRC, importing implicits has a bunch of [side effect to the code](http://spiridonov.pro/2015/10/14/scala-implicit-hell/)

also, we should try to avoid referencing sqlContext

why not only do this when needed, scala 2.10? maybe this will mask user code errors too.

felixcheung: except when in newer spark it should use sparksession instead of sqlcontext, probably will cause some code to fail

anthonycorbacho:
@zjffdu do we have an expected date for this fix?

jongyoul:
??

PR #3059, opened 2018-07-10
ZEPPELIN-3595. Remove interpreter binding in backend

zjffdu:
### What is this PR for?

This PR is trying to remove backend of interpreter binding related code in zeppelin server side. The next step is to put default interpreterSetting into note.json. 

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3595

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
@@ -165,13 +165,10 @@ public class Message implements JsonSerializable {
     LIST_NOTE_JOBS,               // [c-s] get note job management information
     LIST_UPDATE_NOTE_JOBS,        // [c-s] get job management information for until unixtime
     UNSUBSCRIBE_UPDATE_NOTE_JOBS, // [c-s] unsubscribe job information for job management
-                                  // @param unixTime
-    GET_INTERPRETER_BINDINGS,     // [c-s] get interpreter bindings
-                                  // @param noteId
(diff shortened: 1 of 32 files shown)


jongyoul:
Five hundred lines' difficult and sophisticated code will be removed. :-)

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/socket/Message.java
```diff
@@ -165,13 +165,11 @@
     LIST_NOTE_JOBS,               // [c-s] get note job management information
     LIST_UPDATE_NOTE_JOBS,        // [c-s] get job management information for until unixtime
     UNSUBSCRIBE_UPDATE_NOTE_JOBS, // [c-s] unsubscribe job information for job management
-                                  // @param unixTime
+
     GET_INTERPRETER_BINDINGS,     // [c-s] get interpreter bindings
```
zjffdu: need to remove this message after frontend change

jongyoul:
@felixcheung It's not for now if you set `spark interpreter group` as the first interpreter group in your `zeppelin-site.xml`. I think it needs more implementation even in front side as well to keep the default interpreter setting per note. We need to know whole plan to remove interpreter binding option

Tagar:
@jongyoul I like the idea of having a default interpreter (and not a default interpreter group ) **per note**. 
+1

zjffdu:

PR #3060, opened 2018-07-10
[ZEPPELIN-3600] Add REST API to change logger level dynamically

jongyoul:
### What is this PR for?
Enabling users to change the level of a specific logger to investigate Zeppelin server's status.

### What type of PR is it?
[Feature]

### Todos
* [x] - Add api to change the level of logger
* [x] - Add exception handler

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3600

1. http --session ~/session.txt --form POST http://localhost:8080/api/login userName=admin password=password1
1. http --session ~/session.txt POST http://localhost:8080/api/admin name=org.apache.zeppelin.rest.AdminRestApi level=DEBUG

### Questions:
* Does this needs documentation? May be

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/GsonProvider.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/GsonProvider.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.server;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyReader;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+import org.apache.log4j.Logger;
+import org.apache.zeppelin.rest.message.LoggerRequest;
+import org.apache.zeppelin.rest.message.gson.LoggerSerializer;
+
+@Provider
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class GsonProvider<T> implements MessageBodyReader<T>, MessageBodyWriter<T> {
+  private final Gson gson;
+
+  public GsonProvider() {
+    GsonBuilder gsonBuilder = new GsonBuilder().enableComplexMapKeySerialization();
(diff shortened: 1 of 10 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/server/GsonProvider.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.zeppelin.rest.message.LoggerRequest;
+import org.apache.zeppelin.rest.message.gson.LoggerSerializer;
+
+@Provider
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class GsonProvider<T> implements MessageBodyReader<T>, MessageBodyWriter<T> {
+  private final Gson gson;
+
+  public GsonProvider() {
```
zjffdu: What's the purpose of this class ?

jongyoul: Its for serializing json from vanilla object. We already has `JsonResponse` to handle jaon output. But it handles both of error and normal case, and even manages cookie value. I don’t think it’s good and I want to divide them and reduce all of code related to normal cases. I’ll replace all of `JsonResponse` to `provider` which is supported by `jax-rs`

jongyoul:
Will merge if there's no discussion

https://travis-ci.org/jongyoul/zeppelin/builds/402472726

felixcheung:
maybe it's better to clarify that we are adding the Admin REST API in this change and also allow logging lever to be changed?

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public org.apache.log4j.Logger getLogger(String name) {
+    return LogManager.getLogger(name);
+  }
+
+  public void setLoggerLevel(LoggerRequest loggerRequest) throws BadRequestException {
+    try {
+      Class.forName(loggerRequest.getName());
+    } catch (Throwable ignore) {
+      throw new BadRequestException("The name of the logger is incorrect class name");
```
felixcheung: include the bad class name in the exception?

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    org.apache.log4j.Logger logger = LogManager.getLogger(loggerRequest.getName());
+    if (null == logger) {
+      throw new BadRequestException("The name of the logger is wrong");
+    }
+
+    org.apache.log4j.Level level = org.apache.log4j.Level.toLevel(loggerRequest.getLevel(), null);
```
felixcheung: should be validate what getLevel() returns as a value?

felixcheung: also is there a chance to DOS zeppelin by say setting netty log level to INFO or DEBUG?

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
jongyoul: Yes, it helps to avoid typos when typing class name

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
jongyoul: This is also the helper that users don't set a wrong value. And this api is protected by shiro authentication if enabled. And we can also set the permission who can access this api as well.

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
felixcheung: I mean "include the bad class name in the exception message"?

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
felixcheung: ok

felixcheung: ah cool. what's the default `/** = authc` ? so all authencated user can set log level in this case then? could we follow up with permission default to admin?

jongyoul: Yes, correct, agreed to change that permission to admin

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
jongyoul: Fixed

zeppelin-server/src/main/java/org/apache/zeppelin/service/AdminService.java
jongyoul: Added

jongyoul:
Will merge it

PR #3064, opened 2018-07-11
[ZEPPELIN-3596] Saving resources from pool to SQL

oxygen311:
Now you can access to `ResourcePool` data from SQL query. For example, if you write `SELECT * FROM {ResourcePool.note_id=SOME_NOTE_ID.paragraph_id=SOME_PARAGRAPH_ID};` and it will be correct SQL query.
How it works:
1. Creates and fills a table named like a `PARAGRAPH_ID`;
2. Expression `{*}` replaces with actual table name;
3. Updated SQL query is running.

Also, you can not specify the note_id, then the note_id of the current notebook will be used. For example, `SELECT * FROM {ResourcePool.paragraph_id=SOME_PARAGRAPH_ID};`.

### What type of PR is it?
Improvement

### What is the Jira issue?
[Zeppelin 3596](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3596?filter=allopenissues)

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
@@ -16,7 +16,7 @@ package org.apache.zeppelin.jdbc;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.interpreter.InterpreterContext;
-import org.apache.zeppelin.interpreter.InterpreterOutput;
+import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.InterpreterResultMessage;
 import org.apache.zeppelin.resource.ResourcePool;
 import org.apache.zeppelin.resource.WellKnownResourceName;
@@ -25,7 +25,6 @@ import org.apache.zeppelin.tabledata.InterpreterResultTableData;
 import org.apache.zeppelin.tabledata.Row;
 import org.apache.zeppelin.tabledata.TableData;
 
-import java.io.IOException;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Arrays;
@@ -35,10 +34,11 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 public class JDBCPoolManager {
-  private static final String paragraphIdKey = "paragraph_id=";
-  private static final String noteIdKey = "note_id=";
+  private static final String PARAGRAPH_ID_KEY = "paragraph_id=";
+  private static final String NOTE_ID_KEY = "note_id=";
+  // the table name must begin with a letter
+  private static final String SQL_NAME_PREFIX = "p";
 
-  // todo: smarter parser
   private static String getPoolExpression(String sqlReq) {
     final int poolReqBegIndex = sqlReq.indexOf("{ResourcePool");
     final int poolReqEndIndex = sqlReq.indexOf("}", poolReqBegIndex);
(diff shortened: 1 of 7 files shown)


zjffdu:
Thanks @oxygen311 for the contribution, but I think it is already done in ZEPPELIN-3377 via ZeppelinContext variable interpolation

oxygen311:
@zjffdu variable interpolation is about just using text variable. But with this feature you can transform `%table` text from ResourcePool to SQL table.

zjffdu:
@oxygen311 I am not sure how much useful and important this senior is for users. It looks like a little weird to me that you have to insert the paragraph result into database and then execute the query. 
Can you share more details about your scenario ?

BTW, I raise one thread in mail list to discuss about sharing data in zeppelin, welcome any comments on that. 
https://lists.apache.org/thread.html/36d5c74bc946df2ed2ab456227251e94c1d88635c6acbc69d9a8c4f1@%3Cdev.zeppelin.apache.org%3E

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.stream.Collectors;
+
+public class JDBCPoolManager {
+  private static final String PARAGRAPH_ID_KEY = "paragraph_id=";
+  private static final String NOTE_ID_KEY = "note_id=";
+  // the table name must begin with a letter
+  private static final String SQL_NAME_PREFIX = "p";
+
+  private static String getPoolExpression(String sqlReq) {
+    final int poolReqBegIndex = sqlReq.indexOf("{ResourcePool");
```
felixcheung: this isn't very safe - the sql query could contain the same text

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private static String getSqlCreateTableReq(List<String> columDefs, String tableName) {
+    return "CREATE TABLE " + tableName + " (" + StringUtils.join(columDefs, ", ") + ");";
+  }
+
+  private static List<String> getSqlInsertReqs(TableData tableData, String tableName) {
+    List<String> reqs = new LinkedList<>();
+    for (Iterator<Row> iterator = tableData.rows(); iterator.hasNext(); ) {
+      Row row = iterator.next();
+      reqs.add("INSERT INTO " + tableName + " VALUES ("
```
felixcheung: is this going to be SQL syntax independent?

zjffdu:
ping @oxygen311

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
oxygen311: We rely on the [SQL standard](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) from 1992. If database does not support this standard, this feature will not work.

oxygen311:
@zjffdu 
The user can transfer data not only from one database to another, but from flat files and interpreters, which are not available in the python (e.g. SAP).
Sending data through `ResourcePool` is not as effective as native methods, but can cover up to 80% of user needs.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
oxygen311: Common SQL syntax almost not contain figure brackets. Especially with word "ResourcePool" in start of it.

zjffdu:
@oxygen311 Could you tell more details about your scenario ? What is the interpreter of your source data (ResourcePool.note_id=SOME_NOTE_ID.paragraph_id=SOME_PARAGRAPH_ID)

oxygen311:
Scenario: get a list of client IDs from one source (flat file, base, SAP), send the list to the database and perform complex analytics with it.
The data in the database is large and there is no way to pull everything into Python to run the analytics inside the Python.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCPoolManager.java
felixcheung: what I'm saying is we could be more strict/careful in parsing the sql, not just indexOf

oxygen311: Fixed

PR #3065, opened 2018-07-12
ZEPPELIN-3617. Allow to specify saving resourceName as paragraph property

zjffdu:
### What is this PR for?
This is allow user to specify resourceName when they want to save the paragraph result into ResourcePool. Before this PR, user don't have control on what name of the saving resource name. It is associated with noteId and paragraphId, but this is not a good solution. Because when you clone the note, noteId will be changed, and you have to change the noteId in code as well. This PR is trying to allow user to set resource Name for the saving paragraph result. 

### What type of PR is it?
[ Feature |g]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3617

* CI pass

screen shot 2018-07-12 at 2 45 54 pm

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -559,6 +559,8 @@ public class PythonInterpreter extends Interpreter {
           InterpreterContext.get());
       if (result.code() != Code.SUCCESS) {
         throw new IOException("Fail to run bootstrap script: " + resourceName + "\n" + result);
+      } else {
+        LOGGER.debug("Bootstrap python successfully.");
       }
     } catch (InterpreterException e) {
       throw new IOException(e);
--- a/python/src/main/resources/python/zeppelin_context.py
+++ b/python/src/main/resources/python/zeppelin_context.py
@@ -61,6 +61,14 @@ class PyZeppelinContext(object):
     def get(self, key):
         return self.__getitem__(key)
 
+    def getAsDataFrame(self, key):
+        value = self.get(key)
+        try:
+            import pandas as pd
+        except ImportError:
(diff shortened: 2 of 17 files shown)


zjffdu:
@mebelousov  This is related with ZEPPELIN-3545, feel free to try this PR and provide comments.

mebelousov:
@zjffdu Nice!
How about adding to ResourcePool only if resourceName specified in paragraph properties?

sanjaydasgupta:
This will also make it possible to access paragraph results easily using the z-variable interpolation feature of various interpreters. So a simple command like `%sh echo {population-table}` will retrieve and display a previously saved (population-table) result (if interpolation is enabled for shell).

But a crafty user will be able to disrupt https://issues.apache.org/jira/browse/ZEPPELIN-3596 by using a resourceName compatible with ZEPPELIN-3596's syntax for {...}

mebelousov:
@zjffdu WDYT?

zjffdu:
Sorry for late response @mebelousov , this is what I planed to do before. The reason I didn't remove this piece of code https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java#L653 is because I am not sure how many use this approach, I just don't want to break users's existing code. But if no one use it, I would sure prefer to add to ResourcePool only if resourceName specified in paragraph properties

mebelousov:
@zjffdu First version with ResourcePool was released a month ago (thanks to you ;) )
Now is good moment to decide future limitations.

zjffdu:
@mebelousov Do you mean you have already used this feature in your company a month ago ?

mebelousov:
@zjffdu Our team do not use actively ResourcePool. We only plan to use.

zjffdu:

PR #3067, opened 2018-07-13
[ZEPPELIN-3609][PART-1] Refactor ZeppelinServer class

jongyoul:
### What is this PR for?
Refactoring ZeppelinServer to add new feature easily and understand more clearly. And I'd like to refactor with several commits not to keep apart from the current master.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Intorduce @Inject annotation for server component

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3609

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -177,6 +176,16 @@ public InterpreterSettingManager(ZeppelinConfiguration conf,
     this.interpreterEventServer = new RemoteInterpreterEventServer(conf, this);
     this.interpreterEventServer.start();
     init();
+
+    Runtime.getRuntime()
+        .addShutdownHook(
+            new Thread(
```
felixcheung: so then this would run asynchronously (because new thread) in the shutdown hook? doesn't that mean it is possible that it hasn't shutdown completely then?

felixcheung: don't we need to call .start() on the new thread?

jongyoul: Yes, we don't have to do.

jongyoul:
I decided to make each change as small as possible. This is the first part.
CI passed: https://travis-ci.org/jongyoul/zeppelin/builds/416895950

jongyoul:
Rebased from current master. Will merge it if CI passed

PR #3068, opened 2018-07-13
[ZEPPELIN-3438] Passing Z variables to BigQuery 

sanjaydasgupta:
### What is this PR for?
This PR enables the interpolation of ZeppelinContext objects into the paragraph text of BigQuery cells. It also introduces a new interpreter-level configuration parameter named `zeppelin.bigquery.interpolation`. This new parameter is `false` by default, and must be set to `true` to enable object interpolation. The default value of `false` guarantees backward compatibility for users who are not aware of the new feature.

The implementation takes the same approach that was followed in PR [2898](https://github.com/apache/zeppelin/pull/2898) and [2903](https://github.com/apache/zeppelin/pull/2903).

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3438

Two new unit test classes have been added: [BigQueryInterpolationDisabled.java](https://github.com/sanjaydasgupta/zeppelin/blob/572c2aaa7a96eca98cdef4c77044dfc7a359f89f/bigquery/src/test/java/org/apache/zeppelin/bigquery/BigQueryInterpolationDisabled.java), and [BigQueryInterpolationEnabled.java](https://github.com/sanjaydasgupta/zeppelin/blob/572c2aaa7a96eca98cdef4c77044dfc7a359f89f/bigquery/src/test/java/org/apache/zeppelin/bigquery/BigQueryInterpolationEnabled.java).

Also, the code in this PR merely causes the BigQuery interpreter to "opt-in" to the implementation of z-variable interpolation already existing in the Interpreter base class - described in [PR-2898](https://github.com/apache/zeppelin/pull/2898). The unit tests necessary for testing z-variable interpolation in detail are already present in PR-2898

### Questions:
* Is there breaking changes for older versions?  No
* Does this needs documentation?  Yes, documentation changes provided in [bigquery.md](https://github.com/sanjaydasgupta/zeppelin/blob/572c2aaa7a96eca98cdef4c77044dfc7a359f89f/docs/interpreter/bigquery.md)

### DIFF
--- a/docs/interpreter/bigquery.md
+++ b/docs/interpreter/bigquery.md
@@ -53,6 +53,11 @@ limitations under the License.
     <td></td>
     <td>BigQuery SQL dialect (standardSQL or legacySQL). If empty, [query prefix](https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql#sql-prefix) like '#standardSQL' can be used.</td>
   </tr>
+  <tr>
+    <td>zeppelin.bigquery.interpolation</td>
+    <td></td>
+    <td>Enables ZeppelinContext variable interpolation into paragraph text. Default value is false.</td>
+  </tr>
 </table>
 
 
@@ -118,6 +123,28 @@ ORDER BY
 LIMIT
   40
 ```
+
+## Object Interpolation
+The BigQuery interpreter also supports interpolation of `ZeppelinContext` objects into the paragraph text.
+The following example shows one use of this facility:
+
+####In Scala cell:
+```
+z.put("country_code", "KR")
+    // ...
+```
+
+
+```bash
+%bigquery.sql
+    select * from patents_list where 
+    priority_country = '{country_code}' and filing_date like '2015-%'
+```
+
+Object interpolation is disabled by default, and can be enabled  by 
+setting the value of the property `zeppelin.bigquery.interpolation` to `true` (see _More Properties_ above). 
+More details of this feature can be found in the Spark interpreter documentation under 
+[Object Interpolation](spark.html#object-interpolation)
 
 ## Technical description
 
--- a/docs/interpreter/bigquery.md
+++ b/docs/interpreter/bigquery.md
@@ -55,8 +55,8 @@ limitations under the License.
   </tr>
   <tr>
     <td>zeppelin.bigquery.interpolation</td>
-    <td></td>
-    <td>Enables ZeppelinContext variable interpolation into paragraph text. Default value is false.</td>
+    <td>false</td>
+    <td>Enables ZeppelinContext variable interpolation into paragraph text.</td>
   </tr>
 </table>
 
@@ -143,8 +143,7 @@ z.put("country_code", "KR")
 
 Object interpolation is disabled by default, and can be enabled  by 
 setting the value of the property `zeppelin.bigquery.interpolation` to `true` (see _More Properties_ above). 
-More details of this feature can be found in the Spark interpreter documentation under 
-[Object Interpolation](spark.html#object-interpolation)
+More details of this feature can be found under _Usage_/_Other Features_/[Zeppelin-Context](../usage/other_features/zeppelin_context.html)
 
 ## Technical description
 
--- a/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
+++ b/bigquery/src/main/java/org/apache/zeppelin/bigquery/BigQueryInterpreter.java
@@ -289,7 +289,9 @@ public class BigQueryInterpreter extends Interpreter {
   }
 
   @Override
-  public InterpreterResult interpret(String sql, InterpreterContext contextInterpreter) {
+  public InterpreterResult interpret(String originalSql, InterpreterContext contextInterpreter) {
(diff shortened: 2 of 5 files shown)


jongyoul:
LGTM

docs/interpreter/bigquery.md
```diff
@@ -119,6 +124,28 @@ LIMIT
...
+```bash
+%bigquery.sql
+    select * from patents_list where 
+    priority_country = '{country_code}' and filing_date like '2015-%'
+```
+
+Object interpolation is disabled by default, and can be enabled  by 
+setting the value of the property `zeppelin.bigquery.interpolation` to `true` (see _More Properties_ above). 
+More details of this feature can be found in the Spark interpreter documentation under 
+[Object Interpolation](spark.html#object-interpolation)
```
felixcheung: is this the right link? it's been changed in other cases
https://github.com/apache/zeppelin/commit/941ccbbbdb6c2ca11f09f54bf0ddd31d6ba25d1c

sanjaydasgupta:
Thanks for pointing out the issue @felixcheung 

I've corrected the link as well as some of the preceding words as the ZeppelinContext documentation is now in its own page - not under Spark Interpreter. This correction (the words preceding the link) will have to be made in the Spark and JDBC interpreter documentation too, and I will bundle it later with another PR. 

I have also made another correction in the configuration table - the default value `false` has been moved into the middle column (where it should have been) from its previous location at the end of the description in the last column.

zjffdu:
@sanjaydasgupta I was thinking would it better to put it in Interpreter base class instead of doing it in every interpreter implementation.

sanjaydasgupta:
Thanks for the suggestion @zjffdu. 

The same idea was given earlier by Leemoonsoo [here](https://github.com/apache/zeppelin/pull/2502#issuecomment-323272122), and I had then [moved all of the required code](https://github.com/apache/zeppelin/pull/2834/commits/28688256f3b11e696c08ed10875e0596db8c0463#diff-f9d22f7302ae356454cdcc637942856f) into the `interpolate(...)` method of the base class.

So, the PR for each interpreter now contains the following 4 elements:

1) The configuration changes (defining the `zeppelin.???.interpolate` parameter) in the associated `interpreter-setting.json` file (seen [here](https://github.com/apache/zeppelin/pull/3068/files#diff-1081c34c222dd47418550b2660730df6)),
2) The documentation changes in the associated `docs/interpreter/???.md` file (seen [here](https://github.com/apache/zeppelin/pull/3068/files#diff-dae0efb0a1eda2c293e809843480c781)), 
3) A 2-line change to the interpreter's `interpret()` method causing it to invoke `interpolate(...)` in the base class if `zeppelin.???.interpolate` is found to be true (seen [here](https://github.com/apache/zeppelin/pull/3068/files#diff-1c3f4657786b890b0c786adaec9b4004)), and
4) The specific unit-test classes as applicable (seen [here](https://github.com/apache/zeppelin/pull/3068/files#diff-d117ea24a43ae0fd6db3151c50c658dc) and [here](https://github.com/apache/zeppelin/pull/3068/files#diff-4388c6558b95f9b2944130ed7b0ab5a5))

Keeping this consistent set of files together in one PR helps the review process IMHO, but I am open to any suggestions for further improvement.

sanjaydasgupta:
This is a great idea @zjffdu. 

There are also a few other things to think of - like automatically generating the name of the enabler parameter on a per-interpreter basis (`zeppelin.shell.interpolation` in your sample code above). This can be done by adding a static map to `interpreter.java` that translates from interpreter class name to parameter name, but I think we will obtain greater mileage from the following 2-part strategy:

1) refactor the `interpret(...)` method in all of the existing interpreters once to introduce the 2 lines that have to change in each interpreter (no need to introduce a new method like `interpreter2(...)`
2) add a map to `interpreter.java` that translates from interpreter class name to parameter name, for use in the updated code in (1) above. This map could be filled  with unused names in advance without doing any harm.

This is more or less what I had in mind during [this earlier discussion](https://github.com/apache/zeppelin/pull/2502#issuecomment-323272122), but I did not want to start it then as there was other interpreter refactoring work going on. Perhaps this is a better time to resume that thought :-) Allow me to ask the opinions of a wider group of reviewers - what do you think @felixcheung @jongyoul @zjffdu ?

felixcheung:
do we really need this for all interpreters though? for those we don't have active maintainer it might better not to change them

zjffdu:
We won't enable it for all interpreters. By default we could disable it, interpreter can enable it via setting `zeppelin.interpreter.interpolation` to `true` to enable it. The purpose is just to provide hook to allow interpreter have chance to update code before doing real interpretation.

richiesgr:
Hi this PR have never been merged.
So simple question is variable interpolation for bigquery interpreter is working or not at all ?
trying to use z.put(var,val) then use in %%bigquery block but no interpolation the string the placeholder is not replaced
Thanks

zjffdu:
@richiesgr Yes, I don't think interpolation is working for bigquery interpreter right now. The easy way is just to make BigQuery extends AbstarctInterpreter. so that user can enable interpolation via local properties, e.g.

```
%bigquery(interpolate=true)

...
```

PR #3079, opened 2018-07-18
ZEPPELIN-3635. Drop support for spark before 1.6

zjffdu:
### What is this PR for?
This PR is to remove legacy code to drop support for spark before 1.6

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3635

* Ci pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
@@ -23,14 +23,8 @@ import org.slf4j.LoggerFactory;
  * Provide reading comparing capability of spark version returned from SparkContext.version()
  */
 public class SparkVersion {
-  Logger logger = LoggerFactory.getLogger(SparkVersion.class);
+  private static final Logger logger = LoggerFactory.getLogger(SparkVersion.class);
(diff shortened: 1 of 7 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
```diff
@@ -86,37 +80,14 @@ public static SparkVersion fromVersionString(String versionString) {
...
-
-  public boolean oldLoadFilesMethodName() {
-    return this.olderThan(SPARK_1_3_0);
-  }
-
-  public boolean oldSqlContextImplicits() {
-    return this.olderThan(SPARK_1_3_0);
-  }
-
   public boolean isSpark2() {
```
felixcheung: should we check if older than 1.6 and fail?

felixcheung:
looks like we don't test older spark versions..

jongyoul:
BTW, we could also remove profiles for spark 1.4 - 1.5

zjffdu:
Thanks @jongyoul  I have removed these profiles

PR #3081, opened 2018-07-18
ZEPPELIN-3575. Add 'Copy Column Name' button in visualization table

mebelousov:
### What is this PR for?
Add button to drop-down list in visualization table to copy column name.
I know that column name could be obtained in a different way, for example, by auto-complete.
Nevertheless I think this feature is comfortable.

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-3575](https://issues.apache.org/jira/browse/ZEPPELIN-3575)

* Manually

zeppelin-3575

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-table.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-table.js
@@ -256,6 +256,15 @@ export default class TableVisualization extends Visualization {
             return this.context.col.colDef.type === TableColumnType.DATE;
           },
         },
+        {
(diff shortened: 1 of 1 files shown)


mebelousov:
@jongyoul I also haven't good idea to write small and useful test.
It's possible to rewrite copyStringToClipboard as pass to function the column number and then put in clipboard column name, although it's not straightforward way.

jongyoul:
Sorry for the late reply. How about adding an integration test?

mebelousov:
Will try to revert with tests.

PR #3084, opened 2018-07-19
[zeppelin-3639] Add Ipython interpreter prerequisite check for protobuf

khalidhuseynov:
### What is this PR for?
This is to protobuf prerequisite check when starting ipython kernel. This is required since from grpcio 1.12.0 it's  independent from protobuf

### What type of PR is it?
[ Improvement ]

### Todos
* [x] - add protobuf prerequisite check
* [x] - add to ipython docs

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3639

1. remove grpcio and protobuf 
2. install latest grpcio and try to run ipython interpreter - should fail
3. then install protobuf and try again - should work

### Questions:

### DIFF
--- a/docs/interpreter/python.md
+++ b/docs/interpreter/python.md
@@ -258,8 +258,9 @@ IPython is more powerful than the default python interpreter with extra function
    
     - Jupyter `pip install jupyter`
     - grpcio `pip install grpcio`
+    - protobuf `pip install protobuf`
 
-If you already install anaconda, then you just need to install `grpcio` as Jupyter is already included in anaconda.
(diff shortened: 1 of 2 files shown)


zjffdu:
LGTM

Tagar:
Would be nice to check freeze output for tornado version.
tornado 5.x is broken for Zeppelin. https://issues.apache.org/jira/browse/ZEPPELIN-3505

khalidhuseynov:
@Tagar thanks for suggestion, but I think it's better not to include this check for tornado in this function since it's not really pre-requisite check. if we add it, then we need to keep track it, and restore back when the issue with tornado is fixed.

Tagar:
that makes sense. thanks @khalidhuseynov

PR #3087, opened 2018-07-20
[ZEPPELIN-3644]: Adding SPARQL query language support to the editor.

padmacs:
### What is this PR for?
* Adding SPARQL query language support to zeppelin notebook editor

### What type of PR is it?
* [Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3644

### Questions:

### DIFF
--- a/zeppelin-web/bower.json
+++ b/zeppelin-web/bower.json
@@ -49,6 +49,7 @@
         "src-noconflict/mode-scala.js",
         "src-noconflict/mode-python.js",
         "src-noconflict/mode-sql.js",
+        "src-noconflict/mode-sparql.js",
         "src-noconflict/mode-markdown.js",
         "src-noconflict/mode-pig.js",
         "src-noconflict/mode-sh.js",


zjffdu:
@padmacs Can you describe more details about this PR ? What features does this PR provide ? Screenshot will be helpful.

padmacs:
@zjffdu  Actually we are working on an interpreter that submits SPARQL query to our system through GRPC calls currently. But we can work with community to make to more generalized, so that queries can be executed against any of the SPARQL Endpoints.

This PR is specifically about getting an SPARQL query language support into Zeppelin-web module. So that the Notebook editor identifies and displays the SPARQL query with proper color codes for syntax and keywords.

padmacs:
@felixcheung Can you elaborate on which type of hook up you mean?

felixcheung:
this is just for syntax highlighting, there is no interpreter code here.
also even for syntax the ACE editor should be set with the language of choice - this PR does not have either of those.

padmacs:
@felixcheung Yes. This PR will just make the support available for highlighting the SPARQL query syntax, so that any interpreter can use it as language of choice. The interpreter development is still in progress and  is bit specific to the SPARQL server implementation that we have, hence we are working on it to make it more generic so can be used for any SPARQL endpoint.

NicolasRouquette:
FYI: It is possible to use the R SPARQL package:

https://cran.r-project.org/web/packages/SPARQL/index.html

For example, here's the web interface to ACM's SPARQL endpoint:
http://acm.rkbexplorer.com/sparql/

With Zeppelin, we can do the same thing with two paragraphs:

```
%spark.r
library(SPARQL)

endpoint <- "http://acm.rkbexplorer.com/sparql/"
 
query <-
'
PREFIX id:   <http://acm.rkbexplorer.com/id/>
PREFIX rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX akt:  <http://www.aktors.org/ontology/portal#>
PREFIX owl:  <http://www.w3.org/2002/07/owl#>
PREFIX akt:  <http://www.aktors.org/ontology/portal#>
PREFIX akts: <http://www.aktors.org/ontology/support#>

SELECT * WHERE { ?s rdfs:label ?o }
LIMIT 10
'

qd <- SPARQL(endpoint,query)
r <- qd$results
df <- as.DataFrame(r)
registerTempTable(df, "ACM")
```

and:

```
%sql
select * from ACM
```

Obviously, this solution lacks SPARQL-language specific syntax highlighting or completion;
however, it is very convenient for executing SPARQL queries.

PR #3088, opened 2018-07-23
[HOTFIX] Disable some flaky tests

jongyoul:
### What is this PR for?
Some tests become failed due to https://issues.apache.org/jira/browse/ZEPPELIN-3593 BTW, it looks like a timing issue. I saw some test was passed if I added some debug logs. Thus I think it's better to make them to functional tests as it doesn't have to test under integration tests

### What type of PR is it?
[Hot Fix]

### Todos
* [x] - Disable flaky tests

### Questions:


jongyoul:
@zjffdu https://travis-ci.org/jongyoul/zeppelin/builds/406977293 Jenkins seems to become bad

PR #3090, opened 2018-07-23
[ZEPPELIN-3645] Add LSP Protocol completion support

oxygen311:
### What is this PR for?
Add support of request autocomplete with LSP Protocol through TCP.

### What type of PR is it?
Feature

### What is the Jira issue?
[ZEPPELIN-3645](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3645?filter=allopenissues)

### Demo
peek 2018-07-23 14-40

### Questions:
* Does the licenses files need update? Yes, updated
* Does this needs documentation? Yes, updated

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -468,13 +468,15 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
                                                 InterpreterContext interpreterContext)
       throws InterpreterException {
 
-    final List <InterpreterCompletion> completions = LSPUtils.
-        getLspServerComplitions(buf, cursor,
-            getProperty("zeppelin.python.lspHost", "localhost"),
-            Integer.parseInt(getProperty("zeppelin.python.lspPort", "2087")),
-            LSP_LANG_ID);
-    if (!completions.isEmpty()) {
-      return completions;
+    if (Boolean.parseBoolean(getProperty("zeppelin.python.useLsp", "false"))) {
+      final List<InterpreterCompletion> completions = LSPUtils
+          .getLspServerCompletion(buf, cursor,
+              getProperty("zeppelin.python.lspHost", "localhost"),
+              Integer.parseInt(getProperty("zeppelin.python.lspPort", "2087")),
+              LSP_LANG_ID);
+      if (!completions.isEmpty()) {
+        return completions;
+      }
     }
 
     if (iPythonInterpreter != null) {
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -28,6 +28,7 @@ import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.exec.environment.EnvironmentUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.zeppelin.completer.LSPUtils;
 import org.apache.zeppelin.interpreter.BaseZeppelinContext;
 import org.apache.zeppelin.interpreter.Interpreter;
 import org.apache.zeppelin.interpreter.InterpreterContext;
@@ -64,6 +65,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public class PythonInterpreter extends Interpreter implements ExecuteResultHandler {
   private static final Logger LOGGER = LoggerFactory.getLogger(PythonInterpreter.class);
   private static final int MAX_TIMEOUT_SEC = 30;
+  private static final String LSP_LANG_ID = "python";
 
   private GatewayServer gatewayServer;
   private DefaultExecutor executor;
@@ -79,6 +81,7 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
   private BaseZeppelinContext zeppelinContext;
   private String condaPythonExec;  // set by PythonCondaInterpreter
   private boolean usePy4jAuth = false;
+
 
   public PythonInterpreter(Properties property) {
     super(property);
@@ -464,12 +467,23 @@ public class PythonInterpreter extends Interpreter implements ExecuteResultHandl
   public List<InterpreterCompletion> completion(String buf, int cursor,
                                                 InterpreterContext interpreterContext)
       throws InterpreterException {
+
+    final List <InterpreterCompletion> completions = LSPUtils.
+        getLspServerComplitions(buf, cursor,
+            getProperty("zeppelin.python.lspHost", "localhost"),
+            Integer.parseInt(getProperty("zeppelin.python.lspPort", "2087")),
+            LSP_LANG_ID);
+    if (!completions.isEmpty()) {
+      return completions;
+    }
+
     if (iPythonInterpreter != null) {
       return iPythonInterpreter.completion(buf, cursor, interpreterContext);
     }
     if (buf.length() < cursor) {
       cursor = buf.length();
     }
+
     String completionString = getCompletionTargetString(buf, cursor);
     String completionCommand = "__zeppelin_completion__.getCompletion('" + completionString + "')";
     LOGGER.debug("completionCommand: " + completionCommand);
--- a/zeppelin-interpreter/pom.xml
+++ b/zeppelin-interpreter/pom.xml
@@ -45,6 +45,7 @@
     <maven.aeither.provider.version>3.0.3</maven.aeither.provider.version>
     <wagon.version>1.0</wagon.version>
     <jline.version>2.12.1</jline.version>
+    <powermock.version>1.7.4</powermock.version>
 
     <!--plugin versions-->
     <plugin.shade.version>2.3</plugin.shade.version>
@@ -252,9 +253,35 @@
       <artifactId>mockito-all</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+       <groupId>org.powermock</groupId>
+       <artifactId>powermock-module-junit4</artifactId>
+       <version>${powermock.version}</version>
+       <scope>test</scope>
+    </dependency>
+
+    <dependency>
+       <groupId>org.powermock</groupId>
+       <artifactId>powermock-api-mockito</artifactId>
+       <version>${powermock.version}</version>
+       <exclusions>
+          <exclusion>
+             <artifactId>hamcrest-core</artifactId>
+             <groupId>org.hamcrest</groupId>
+          </exclusion>
+          <exclusion>
+             <artifactId>objenesis</artifactId>
+             <groupId>org.objenesis</groupId>
+          </exclusion>
+        </exclusions>
+       <scope>test</scope>
+    </dependency>
+
+
   </dependencies>
 
-  <build>
+    <build>
     <resources>
       <resource>
         <directory>src/main/resources</directory>
--- a/zeppelin-interpreter/pom.xml
+++ b/zeppelin-interpreter/pom.xml
@@ -37,6 +37,7 @@
 
   <properties>
     <!--library versions-->
+    <lsp4j.version>0.4.1</lsp4j.version>
     <commons.pool2.version>2.3</commons.pool2.version>
     <commons.exec.version>1.3</commons.exec.version>
     <maven.plugin.api.version>3.0</maven.plugin.api.version>
@@ -50,6 +51,21 @@
   </properties>
 
   <dependencies>
+
+    <!-- https://mvnrepository.com/artifact/org.eclipse.lsp4j/org.eclipse.lsp4j -->
+    <dependency>
+      <groupId>org.eclipse.lsp4j</groupId>
+      <artifactId>org.eclipse.lsp4j</artifactId>
+      <version>${lsp4j.version}</version>
+    </dependency>
+
+    <!-- https://mvnrepository.com/artifact/org.eclipse.lsp4j/org.eclipse.lsp4j.jsonrpc -->
+    <dependency>
+      <groupId>org.eclipse.lsp4j</groupId>
+      <artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
+      <version>${lsp4j.version}</version>
+    </dependency>
+
 
     <dependency>
       <groupId>org.apache.thrift</groupId>
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
@@ -64,7 +64,7 @@ public class LSPUtils {
    * Return empty list if there're nothing to return or exception occurred.
    */
   public static List<InterpreterCompletion> getLspServerCompletion(String buf, int cursor,
-                                                                   String host, int port, String langId) {
+                                                          String host, int port, String langId) {
     Logger.info("Trying to complete on {} with cursor {}", buf, cursor);
 
     List<InterpreterCompletion> list = Collections.emptyList();
@@ -111,18 +111,18 @@ public class LSPUtils {
         new Position(line, character));
   }
 
-  private static InterpreterCompletion castCompletionItemToInterpreterCompletion(
+  static InterpreterCompletion castCompletionItemToInterpreterCompletion(
       CompletionItem item) {
     return new InterpreterCompletion(item.getLabel(), item.getLabel(), "");
   }
 
-  private static List<InterpreterCompletion> castList(List<CompletionItem> items) {
+  static List<InterpreterCompletion> castList(List<CompletionItem> items) {
     return items.stream()
         .map(LSPUtils::castCompletionItemToInterpreterCompletion)
         .collect(Collectors.toList());
   }
 
-  private static List<InterpreterCompletion> eitherToCompletionList(
+  static List<InterpreterCompletion> eitherToCompletionList(
       Either<List<CompletionItem>, CompletionList> either) {
     if (either.isLeft()) {
       return castList(either.getLeft());
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
@@ -63,8 +63,8 @@ public class LSPUtils {
    * @return list of possible completion.
    * Return empty list if there're nothing to return or exception occurred.
    */
-  public static List<InterpreterCompletion> getLspServerComplitions(String buf, int cursor,
-        String host, int port, String langId) {
+  public static List<InterpreterCompletion> getLspServerCompletion(String buf, int cursor,
+                                                                   String host, int port, String langId) {
     Logger.info("Trying to complete on {} with cursor {}", buf, cursor);
 
     List<InterpreterCompletion> list = Collections.emptyList();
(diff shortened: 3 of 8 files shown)


zjffdu:
Is there any document of how to use this ? Like how to configure and start the LSP server

oxygen311:
Now this feature is enabled only in python interpreter. So you can specify `zeppelin.python.lspHost` and `zeppelin.python.lspPort` properties and use it.
Server is a separate process. I have been using [python-language-server](https://github.com/palantir/python-language-server). But there are much more [server implementations](https://microsoft.github.io/language-server-protocol/implementors/servers/) that can be used with any language.

zjffdu:
Could you add instruction in python.md ? Otherwise it is hard for users to know how to enable this feature except you.

oxygen311:
@zjffdu Sound like good idea)
Done

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
```diff
@@ -0,0 +1,143 @@
...
+
+      server.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(
+          new TextDocumentItem(ANY_URI, langId, ANY_VERSION, buf)
+      ));
+
+      CompletableFuture future =
+          server.getTextDocumentService().completion(getCompletionParams(buf, cursor));
+
+      Either<List<CompletionItem>, CompletionList> either =
+          (Either<List<CompletionItem>, CompletionList>) future.get(3, TimeUnit.SECONDS);
```
jongyoul: BTW, if not set LSP, it waits for 3 seconds every time, doesn't it?

zeppelin-interpreter/src/test/java/org/apache/zeppelin/completer/LSPUtilsTest.java
```diff
@@ -0,0 +1,72 @@
...
+  private static final String LANG_ID = "python";
+  private static final int CURSOR = 0;
+
+  @Test
+  public void getLspServerComplitionsTest() throws Exception{
+
+    try (ServerSocket serverSocket = new ServerSocket(0)) {
+      final int port = serverSocket.getLocalPort();
+
+      Thread t = new Thread(() -> {
```
jongyoul: Making thread has timing issue sometimes. How about using `@Mock`?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
jongyoul: In my opinion, how about adding one more option to enable/disable it?

zeppelin-interpreter/pom.xml
```diff
@@ -51,6 +52,21 @@
 
   <dependencies>
 
+    <!-- https://mvnrepository.com/artifact/org.eclipse.lsp4j/org.eclipse.lsp4j -->
+    <dependency>
+      <groupId>org.eclipse.lsp4j</groupId>
```
jongyoul: image

It has `guava` as a dependency. I'm worried that it might make a problem for other interpreters.

zjffdu: We should not put lsp4j in zeppelin-interpreter. It is better to put it into python module

zeppelin-interpreter/pom.xml
```diff
@@ -51,6 +52,21 @@
...
+      <artifactId>org.eclipse.lsp4j</artifactId>
+      <version>${lsp4j.version}</version>
```
zjffdu: What's the licence of this dependency ?

zeppelin-interpreter/pom.xml
jongyoul: I agree to move it under python. Even if LSP supports most of the languages, `zeppelin-interpreter` influences many of interpreters. We should be careful when we add code in `zeppelin-interpreter`.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @return list of possible completion.
+   * Return empty list if there're nothing to return or exception occurred.
+   */
+  public static List<InterpreterCompletion> getLspServerComplitions(String buf, int cursor,
+        String host, int port, String langId) {
+    Logger.info("Trying to complete on {} with cursor {}", buf, cursor);
+
+    List<InterpreterCompletion> list = Collections.emptyList();
+    try (Socket socket = new Socket(host, port)) {
```
zjffdu: It doesn't seems efficient to create new connection for each getCompletion call. getCompletion will be a very frequent call sometimes.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
zjffdu: Right, agree wth @jongyoul Interpreter should know whether LSP is enabled or not.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
```diff
@@ -0,0 +1,143 @@
...
+  private static int countLines(String str){
+    return str.split("\r\n|\r|\n").length;
+  }
+
+  private static CompletionParams getCompletionParams(String buf, int cursor) {
+    final int actualCursor = Math.min(cursor, buf.length());
+    final String beforeCursor = buf.substring(0, actualCursor);
+    final int line = countLines(beforeCursor) - 1;
+    final int character = beforeCursor.length() - beforeCursor.lastIndexOf("\n") - 1;
+    Logger.info("Line: {}, character: {} from actual cursor: ", line, character, cursor);
```
zjffdu: change it to debug level

zjffdu:
@oxygen311 I can not install it. Let me know if there any other dependency I need to install

```
➜  zeppelin pr/3090 ✗ pip install pyls
Collecting pyls
  Using cached https://files.pythonhosted.org/packages/de/b7/5a056c2cac8fd19c905ae7451c2a60b9195531959c3770a49d6be4d905e3/pyls-0.1.3.zip
    Complete output from command python setup.py egg_info:
    Downloading http://pypi.python.org/packages/source/d/distribute/distribute-0.6.14.tar.gz
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 143, in use_setuptools
        raise ImportError
    ImportError

    During handling of the above exception, another exception occurred:

      File "<string>", line 1, in <module>
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup.py", line 31, in <module>
        main()
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup.py", line 11, in main
        distribute_setup.use_setuptools()
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 145, in use_setuptools
        return _do_download(version, download_base, to_dir, download_delay)
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 124, in _do_download
        to_dir, download_delay)
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 193, in download_setuptools
        src = urlopen(url)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 223, in urlopen
        return opener.open(url, data, timeout)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 532, in open
        response = meth(req, response)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 642, in http_response
        'http', request, response, code, msg, hdrs)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 570, in error
        return self._call_chain(*args)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 504, in _call_chain
        result = func(*args)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 650, in http_error_default
        raise HTTPError(req.full_url, code, msg, hdrs, fp)
    urllib.error.HTTPError: HTTP Error 403: SSL is required

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/
```

zjffdu:
@oxygen311 I can not install it. Let me know if there any other dependency I need to install

```
➜  zeppelin pr/3090 ✗ pip install pyls
Collecting pyls
  Using cached https://files.pythonhosted.org/packages/de/b7/5a056c2cac8fd19c905ae7451c2a60b9195531959c3770a49d6be4d905e3/pyls-0.1.3.zip
    Complete output from command python setup.py egg_info:
    Downloading http://pypi.python.org/packages/source/d/distribute/distribute-0.6.14.tar.gz
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 143, in use_setuptools
        raise ImportError
    ImportError

    During handling of the above exception, another exception occurred:

      File "<string>", line 1, in <module>
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup.py", line 31, in <module>
        main()
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup.py", line 11, in main
        distribute_setup.use_setuptools()
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 145, in use_setuptools
        return _do_download(version, download_base, to_dir, download_delay)
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 124, in _do_download
        to_dir, download_delay)
      File "/private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/setup_utils/distribute_setup.py", line 193, in download_setuptools
        src = urlopen(url)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 223, in urlopen
        return opener.open(url, data, timeout)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 532, in open
        response = meth(req, response)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 642, in http_response
        'http', request, response, code, msg, hdrs)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 570, in error
        return self._call_chain(*args)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 504, in _call_chain
        result = func(*args)
      File "/Users/jzhang/anaconda3/lib/python3.6/urllib/request.py", line 650, in http_error_default
        raise HTTPError(req.full_url, code, msg, hdrs, fp)
    urllib.error.HTTPError: HTTP Error 403: SSL is required

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/dp/hmchg5dd3vbcvds26q91spdw0000gp/T/pip-install-7veml3im/pyls/
```

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -464,12 +467,23 @@ public int getProgress(InterpreterContext context) throws InterpreterException {
   public List<InterpreterCompletion> completion(String buf, int cursor,
                                                 InterpreterContext interpreterContext)
       throws InterpreterException {
+
+    final List <InterpreterCompletion> completions = LSPUtils.
+        getLspServerComplitions(buf, cursor,
```
felixcheung: this doesn;t seem like the style for multi - line code
```
LSPUtils.
getLspServerComplitions
```

also `Complitions`?

zeppelin-interpreter/pom.xml
```diff
@@ -37,6 +37,7 @@
 
   <properties>
     <!--library versions-->
+    <lsp4j.version>0.4.1</lsp4j.version>
```
felixcheung: need to add to LICENSE?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
```diff
@@ -0,0 +1,143 @@
...
+        String host, int port, String langId) {
+    Logger.info("Trying to complete on {} with cursor {}", buf, cursor);
+
+    List<InterpreterCompletion> list = Collections.emptyList();
+    try (Socket socket = new Socket(host, port)) {
+      LanguageServer server = connectToLanguageServer(socket);
+      server.initialize(new InitializeParams());
+
+      server.getTextDocumentService().didOpen(new DidOpenTextDocumentParams(
+          new TextDocumentItem(ANY_URI, langId, ANY_VERSION, buf)
```
felixcheung: what is version = ANY_VERSION for?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
felixcheung: what if the user try to auto complete before the LSP connects? "hanging" for 3 seconds when the user press tab seems pretty bad

oxygen311: This "hanging" for 3 seconds only exists only if the connection to the server is successful. But i think it's good idea to add property for enable/disable this whole feature.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/completer/LSPUtilsTest.java
oxygen311: `@Mock` does not work with static methods. For this, `powermock` library is needed which will be additional dependency.

zeppelin-interpreter/pom.xml
oxygen311: It's right of course.
But this code seems reusable in other interpreters. And it would be strange to add it in python module and use it in other interpreters like method from python module.

zeppelin-interpreter/pom.xml
oxygen311: > LSP4J is published under two licenses:
>  * [Eclipse Public License v1.0](https://www.eclipse.org/legal/epl-v10.html)
>  * [Eclipse Distribution License - v 1.0](https://www.eclipse.org/org/documents/edl-v10.php) [A BSD 3-clause License](https://opensource.org/licenses/BSD-3-Clause)

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
oxygen311: Yes, but we can't keep this connection because we can connect to server from different JVMs. And if one JVM process occupies connection, other JVMs will not be able to connect to server.

oxygen311:
@zjffdu 
Sorry, correct line is `pip install python-language-server`, i will fix it.

oxygen311:
@felixcheung 
I have added property `zeppelin.python.useLsp` which is disabled by default. It does not seem unsecured for me now. We also can specify host and port by ourselves with `zeppelin.python.lspHost` and `zeppelin.python.lspPort` properties.

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
oxygen311: Fixed

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
oxygen311: This is described in the const definition.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/completer/LSPUtilsTest.java
jongyoul: Even if we should `powermock`, it looks better than making tests fragile, in my opinion.

zeppelin-interpreter/pom.xml
oxygen311: Added

zeppelin-interpreter/src/test/java/org/apache/zeppelin/completer/LSPUtilsTest.java
oxygen311: You are right, fragile test is bad practice. I changed it to PowerMock.

zjffdu:
1. Do you have any examples to show how much better LSP is compared to the current approach ?
2. Does the auto completion works across paragraphs ? e.g. I create string variable `a` in p1, can I get correct completion in p2 ?
3. What is the scope of LSP completion ? Is every call to LSP isolated ?

oxygen311:
1. LSP is usable and powerful solution:
- LSP can understand imports in the start of paragraph;
- We can use already made solution for every language and don't write new one;
- LSP server contains much more information than we process now. 
E.g. it's able to show documentation of a method, reformat code, find references etc.

2-3. Now every paragraph is isolated, so `a` will not be shown.
It's made because in current implementation we can not keep connection to the server (because a lot of JVMs).
But maybe someone has better architecture ideas how to connect zeppelin to LSP.

zjffdu:
Then it looks like LSP can only work in paragraph level. This seems not workable for users. Let's assume we also want to make PySpark support LSP. Then I guess if I type `sc.`, I will get nothing for completion, Because `sc` is created by zeppelin not by any paragraph. Is that correct ?

zjffdu:
Is there any session concept in LSP ? so that multiple LSP request could belong to the same session

oxygen311:
@zjffdu 
If LSP has no complete for `sc.`, iPython complete will be shown.
LSP does not support sessions, but it can keep connection.

Tagar:
Very interesting improvements. Thanks for working on this. 
Shouldn't LSP server be embedded into Python / PySpark interpreter itself and not be a separate process?
This would address both security concerns and ability to grasp code across paragraphs like in the above example.

zjffdu:
@oxygen311 This is what I concern about the behavior consistency. Let's assume user use PySpark (ipython is disabled). User may find code completion works pretty well in p1, but works badly in p2 because LSP only works in paragraph level. 

BTW, it is weird to me that LSP don't support concept of session. That would mean If I am writing a long code (1000 lines of code), I need to send all these code to LSP for code completion each time, this seems very inefficient.  But I may be wrong. I am so familiar with LSP. Just want to confirm whether there's more efficient way to use LSP

zeppelin-interpreter/src/main/java/org/apache/zeppelin/completer/LSPUtils.java
felixcheung: hm, it just says it is not used `Zeppelin don't support file versions, so we can pass any version` - but doesn't say if it does support version what it will do. Actually zeppelin does have notebook git support etc, and could have iteration or revision support.

felixcheung:
open ip and port to connect to has become a huge problem recently, so unless LSP has some sort of authentication story, my vote would be "no" even if this is disabled by default, because this can become a support issue when users do have it enable.

https://github.com/apache/zeppelin/pull/3090#pullrequestreview-139750764

mebelousov:
@felixcheung I agree that support of LSP in Zeppelin is unmature idea.

Share please undesired cases with open ip and port.

oxygen311:
Solution must be improved to use by real users.

PR #3091, opened 2018-07-24
ZEPPELIN-3643. Put Shiro related thing in one place

zjffdu:
### What is this PR for?
Move all shiro related thing in one place (`SecurityUtils`)

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3643

* CI pass

### Questions:


zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -179,12 +274,210 @@ public static Collection getRealmsList() {
...
+      }
+
+      if (StringUtils.isBlank(username) || StringUtils.isBlank(tablename)) {
+        return userlist;
+      }
+
+      userquery = String.format("SELECT %s FROM %s", username, tablename);
+    } catch (IllegalAccessException e) {
+      LOGGER.error("Error while accessing dataSource for JDBC Realm", e);
+      return null;
```
felixcheung: seems like we should return userlist (would be empty) here?
or L475 should also return null?

might just need to be consistent

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
```diff
@@ -179,12 +274,210 @@ public static Collection getRealmsList() {
...
+        if (attrs.get(userAttribute) != null) {
+          String currentUser;
+          if (r.getUserLowerCase()) {
+            LOGGER.debug("userLowerCase true");
+            currentUser = ((String) attrs.get(userAttribute).get()).toLowerCase();
+          } else {
+            LOGGER.debug("userLowerCase false");
+            currentUser = (String) attrs.get(userAttribute).get();
+          }
+          if (LOGGER.isDebugEnabled()) {
```
felixcheung: btw, why it is calling `isDebugEnabled`?

zjffdu: I only did refactoring in this PR, didn't change it. It is as before, but I can improve it.

zeppelin-server/src/main/java/org/apache/zeppelin/utils/SecurityUtils.java
zjffdu: will do it

PR #3092, opened 2018-07-24
[ZEPPELIN-3653] - New Java interpreter

selvinsource:
### What is this PR for?
**New Java interpreter**
There are several Java libraries that could be used to leverage the Data Visualization & Collaboration features of Zeppelin hence the need of a Java interpreter to run any java code / library with no further dependencies.
Dependencies to any java library can be added by the end users in the Java Zeppelin interpreter Dependencies settings section.

### What type of PR is it?
* Feature

### Todos
* Any feedback from reviewers

### What is the Jira issue?
* [ZEPPELIN-3653]

Manually
* Start the Zeppelin server
* Create a new note with the java interpreter binding
* Write some java code as per documentation (docs/interpreter/java.md)
Unit tests
* Run unit tests (JavaInterpreterTest.java and JavaInterpreterUtilsTest.java)

### Questions:
No, the dependency on com.thoughtworks.qdox was already added as part of the Beam Interpreter.
No.
Yes, it has been added to the PR, see docs/interpreter/java.md.

### DIFF
--- a/conf/interpreter-list
+++ b/conf/interpreter-list
@@ -17,23 +17,23 @@
 #
 # [name]  [maven artifact]  [description]
 
-alluxio         org.apache.zeppelin:zeppelin-alluxio:0.7.0              Alluxio interpreter
-angular         org.apache.zeppelin:zeppelin-angular:0.7.0              HTML and AngularJS view rendering
-beam            org.apache.zeppelin:zeppelin-beam:0.7.0                 Beam interpreter
-bigquery        org.apache.zeppelin:zeppelin-bigquery:0.7.0             BigQuery interpreter
-cassandra       org.apache.zeppelin:zeppelin-cassandra_2.11:0.7.0       Cassandra interpreter built with Scala 2.11
-elasticsearch   org.apache.zeppelin:zeppelin-elasticsearch:0.7.0        Elasticsearch interpreter
-file            org.apache.zeppelin:zeppelin-file:0.7.0                 HDFS file interpreter
-flink           org.apache.zeppelin:zeppelin-flink_2.11:0.7.0           Flink interpreter built with Scala 2.11
-hbase           org.apache.zeppelin:zeppelin-hbase:0.7.0                Hbase interpreter
-ignite          org.apache.zeppelin:zeppelin-ignite_2.11:0.7.0          Ignite interpreter built with Scala 2.11
-jdbc            org.apache.zeppelin:zeppelin-jdbc:0.7.0                 Jdbc interpreter
-kylin           org.apache.zeppelin:zeppelin-kylin:0.7.0                Kylin interpreter
-java            org.apache.zeppelin:zeppelin-java:0.7.0                 Java interpreter
-lens            org.apache.zeppelin:zeppelin-lens:0.7.0                 Lens interpreter
-livy            org.apache.zeppelin:zeppelin-livy:0.7.0                 Livy interpreter
-md              org.apache.zeppelin:zeppelin-markdown:0.7.0             Markdown support
-pig             org.apache.zeppelin:zeppelin-pig:0.7.0                  Pig interpreter
-python          org.apache.zeppelin:zeppelin-python:0.7.0               Python interpreter
-scio            org.apache.zeppelin:zeppelin-scio_2.11:0.7.0            Scio interpreter
-shell           org.apache.zeppelin:zeppelin-shell:0.7.0                Shell command
+alluxio         org.apache.zeppelin:zeppelin-alluxio:0.9.0              Alluxio interpreter
+angular         org.apache.zeppelin:zeppelin-angular:0.9.0              HTML and AngularJS view rendering
+beam            org.apache.zeppelin:zeppelin-beam:0.9.0                 Beam interpreter
+bigquery        org.apache.zeppelin:zeppelin-bigquery:0.9.0             BigQuery interpreter
+cassandra       org.apache.zeppelin:zeppelin-cassandra_2.11:0.9.0       Cassandra interpreter built with Scala 2.11
+elasticsearch   org.apache.zeppelin:zeppelin-elasticsearch:0.9.0        Elasticsearch interpreter
+file            org.apache.zeppelin:zeppelin-file:0.9.0                 HDFS file interpreter
+flink           org.apache.zeppelin:zeppelin-flink_2.11:0.9.0           Flink interpreter built with Scala 2.11
+hbase           org.apache.zeppelin:zeppelin-hbase:0.9.0                Hbase interpreter
+ignite          org.apache.zeppelin:zeppelin-ignite_2.11:0.9.0          Ignite interpreter built with Scala 2.11
+jdbc            org.apache.zeppelin:zeppelin-jdbc:0.9.0                 Jdbc interpreter
+kylin           org.apache.zeppelin:zeppelin-kylin:0.9.0                Kylin interpreter
+java            org.apache.zeppelin:zeppelin-java:0.9.0                 Java interpreter
+lens            org.apache.zeppelin:zeppelin-lens:0.9.0                 Lens interpreter
+livy            org.apache.zeppelin:zeppelin-livy:0.9.0                 Livy interpreter
+md              org.apache.zeppelin:zeppelin-markdown:0.9.0             Markdown support
+pig             org.apache.zeppelin:zeppelin-pig:0.9.0                  Pig interpreter
+python          org.apache.zeppelin:zeppelin-python:0.9.0               Python interpreter
+scio            org.apache.zeppelin:zeppelin-scio_2.11:0.9.0            Scio interpreter
+shell           org.apache.zeppelin:zeppelin-shell:0.9.0                Shell command
--- a/conf/interpreter-list
+++ b/conf/interpreter-list
@@ -29,6 +29,7 @@ hbase           org.apache.zeppelin:zeppelin-hbase:0.7.0                Hbase in
 ignite          org.apache.zeppelin:zeppelin-ignite_2.11:0.7.0          Ignite interpreter built with Scala 2.11
 jdbc            org.apache.zeppelin:zeppelin-jdbc:0.7.0                 Jdbc interpreter
 kylin           org.apache.zeppelin:zeppelin-kylin:0.7.0                Kylin interpreter
+java            org.apache.zeppelin:zeppelin-java:0.7.0                 Java interpreter
 lens            org.apache.zeppelin:zeppelin-lens:0.7.0                 Lens interpreter
 livy            org.apache.zeppelin:zeppelin-livy:0.7.0                 Livy interpreter
 md              org.apache.zeppelin:zeppelin-markdown:0.7.0             Markdown support
--- a/docs/interpreter/java.md
+++ b/docs/interpreter/java.md
@@ -1,0 +1,113 @@
+---
+layout: page
+title: Java interpreter in Apache Zeppelin
+description: Run Java code and any distributed java computation engine by importing the dependencies in the interpreter configuration.
+group: interpreter
+---
+<!--
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+{% include JB/setup %}
+
+# Java interpreter for Apache Zeppelin
+
+<div id="toc"></div>
+
+## How to use
+Basically, you can write normal java code. You should write the main method inside a class because the interpreter invoke this main to execute the code. Unlike Zeppelin normal pattern, each paragraph is considered as a separate job, there isn't any relation to any other paragraph.
+
+TODO: UPDATE EXAMPLE BELOW WITH A JAVA ONE... TRY IT IN THE INTERPRETER FIRST
+The following is a demonstration of a word count example with data represented in array of strings
+But it can read data from files by replacing `Create.of(SENTENCES).withCoder(StringUtf8Coder.of())` with `TextIO.Read.from("path/to/filename.txt")`
+
+```java
+%java
+
+// most used imports
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.transforms.Create;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+import java.util.ArrayList;
+import org.apache.beam.runners.direct.*;
+import org.apache.beam.sdk.runners.*;
+import org.apache.beam.sdk.options.*;
+import org.apache.beam.runners.flink.*;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.io.TextIO;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.Count;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.MapElements;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.SimpleFunction;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+public class MinimalWordCount {
+  static List<String> s = new ArrayList<>();
+  
+  static final String[] SENTENCES_ARRAY = new String[] {
+    "Hadoop is the Elephant King!",
+    "A yellow and elegant thing.",
+    "He never forgets",
+    "Useful data, or lets",
+    "An extraneous element cling!",
+    "A wonderful king is Hadoop.",
+    "The elephant plays well with Sqoop.",
+    "But what helps him to thrive",
+    "Are Impala, and Hive,",
+    "And HDFS in the group.",
+    "Hadoop is an elegant fellow.",
+    "An elephant gentle and mellow.",
+    "He never gets mad,",
+    "Or does anything bad,",
+    "Because, at his core, he is yellow",
+	};  
+  static final List<String> SENTENCES = Arrays.asList(SENTENCES_ARRAY);
+  public static void main(String[] args) {
+    PipelineOptions options = PipelineOptionsFactory.create().as(PipelineOptions.class);
+    options.setRunner(FlinkRunner.class);
+    Pipeline p = Pipeline.create(options);
+    p.apply(Create.of(SENTENCES).withCoder(StringUtf8Coder.of()))
+         .apply("ExtractWords", ParDo.of(new DoFn<String, String>() {
+           @ProcessElement
+           public void processElement(ProcessContext c) {
+             for (String word : c.element().split("[^a-zA-Z']+")) {
+               if (!word.isEmpty()) {
+                 c.output(word);
+               }
+             }
+           }
+         }))
+        .apply(Count.<String> perElement())
+        .apply("FormatResults", ParDo.of(new DoFn<KV<String, Long>, String>() {
+          @ProcessElement
+          public void processElement(DoFn<KV<String, Long>, String>.ProcessContext arg0)
+            throws Exception {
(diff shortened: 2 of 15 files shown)


conf/interpreter-list
```diff
@@ -29,6 +29,7 @@ hbase           org.apache.zeppelin:zeppelin-hbase:0.7.0                Hbase in
 ignite          org.apache.zeppelin:zeppelin-ignite_2.11:0.7.0          Ignite interpreter built with Scala 2.11
 jdbc            org.apache.zeppelin:zeppelin-jdbc:0.7.0                 Jdbc interpreter
 kylin           org.apache.zeppelin:zeppelin-kylin:0.7.0                Kylin interpreter
+java            org.apache.zeppelin:zeppelin-java:0.7.0                 Java interpreter
```
zjffdu: The version here seems outdated, I created https://issues.apache.org/jira/browse/ZEPPELIN-3660 for it

conf/zeppelin-site.xml.template
```diff
@@ -331,13 +331,13 @@
...
   <description>Comma separated interpreter configurations. First interpreter become a default</description>
 </property>
 
 <property>
   <name>zeppelin.interpreter.group.order</name>
-  <value>spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase,bigquery,beam,groovy</value>
+  <value>spark,md,angular,sh,livy,alluxio,file,psql,flink,python,ignite,lens,cassandra,geode,kylin,elasticsearch,scalding,jdbc,hbase,bigquery,beam,groovy,java</value>
```
zjffdu: I also concern whether we should keep this property. I believe zeppelin.interpreter.default should be sufficient. We only care about the default interpreter.

zjffdu:
Thanks @selvinsource for the contribution. Have you thought about other implementation of java interpreter. like https://github.com/albertlatacz/java-repl and JShell of Java 9

selvinsource:
@zjffdu thanks for reviewing the PR.

Let me give you a bit of background before I answer your question on using another java interpreter.

First of all, I found lately Zeppelin an amazing tool to work with and in particular using the Spark/scala interpreter.
However I also started evaluating / using Hazelcast Jet, which is a similar streaming/batch engine, entirely written in Java, and I immediately had a need to have that in Zeppelin to do some visualization of the results.

I therefore started to look into how to create an interpreter for Jet and went through all the existing interpreter to get some ideas. I then realized that the Beam interpreter is just a java interpreter apart from the many dependencies on Beam/Google libraries.
A solution could have been to clone the Beam interpreter to create a Jet one with dependencies only on Hazelcast Jet but that didn't seem the best solution (when you start cloning code).

Hence my idea to create a Java interpreter (using what has been done and tested already with Beam) and have Beam (and eventually Hazelcast Jet) interpreters to simply extends the Java one with engine specific dependencies.

So, this PR is a start, to have a common Java interpreter that other Java API/engine interpreters can extend.

I agree the current implementation is not ideal as a whole class with a main method is needed to run some simple java code, the better approach is to use a Java REPL API.

The obvious choice is Java 9 JShell as it is the only one that is going to be supported in the long run and no need of other dependencies as it is shipped as part of the JDK. I don't think it's wise to use any other Java REPL as they won't compete with JShell which will be the standard that every IDE will integrate with.

However, the question back to you, is Zeppelin ready to be compiled with Java 9 and requiring Java 9 run time? Is it wise at this stage?
Talking from experience, in production environments I have seen rarely Java 9, the Java community is waiting for the next LTS release which is Java 11 (due in Sep 2018).

I would say to start with a Java interpreter like the one I proposed (re-using what's good was done for Beam), have Beam and eventually Jet extends the Java interpreter. Once Zeppelin will run on Java 11, the Java interpreter can be easily upgraded to use the JShell API therefore all other java based interpreters will have the same benefits (Beam, Jet, etc...).

Definitely the Java interpreter will be much more usable and immediate with a REPL, it just need to wait for the JShell to be available.

If you want I can already create a JIRA ticket to eventually update the Java interpreter to JShell as soon as this is available and I am happy to contribute to it: having seen the JShell API is going to be even simpler than the current REPL implementation that has been used in the Beam interpreter (but let's not forget that it works just fine!).

zjffdu:
Thanks @selvinsource Very glad that you willing to improve the current java interpreter when JShell is available in Java 11. It make sense to use the current java repl that Beam use, and improve it later. 
BTW, do you plan to create Hazelcast Jet interpreter later after this PR is completed ?

selvinsource:
@zjffdu as soon as this PR is merged in the master I will update the upstream master and start working on Jet interpreter, I have created a Jira ticket already (with a dependency on the Java interpreter).

Hazelcast Jet interpreter will be similar to the Beam one in this PR, it will simply implement JavaInterpreter and it will have the dependencies required (hazelcast-jet jar). What's more I am planning to add some more utils for Hazelcast for iMap/iCache/DAG/etc, see the JIRA description https://issues.apache.org/jira/browse/ZEPPELIN-3654.

When is the next zeppelin release due (0.9)? I can try to put both in it before that.

zjffdu:
There's no exact plan for 0.9.0, my rough plan is to release 0.9.0 in the next half year of 2018.

pom.xml
```diff
@@ -821,6 +821,13 @@
       </modules>
     </profile>
 
+    <profile>
```
zjffdu: Could you just add java module directly instead of using profile ? We plan to remove all these interpreter specific profiles.

java/src/main/java/org/apache/zeppelin/java/JavaInterpreter.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.List;
+import java.util.Properties;
+import java.util.UUID;
+
+/**
+ * Java interpreter
+ */
+public class JavaInterpreter extends Interpreter {
+
+  Logger logger = LoggerFactory.getLogger(JavaInterpreter.class);
```
zjffdu: Use final static

java/README.md
```diff
@@ -0,0 +1,24 @@
+# Overview
+Java interpreter for Apache Zeppelin
+
+# Architecture
+Current interpreter implementation supports the static repl. It compiles the code in memory, execute it and redirect the output to zeppelin.
+
+## Building the Java Interpreter
+You have to first build the Java interpreter by enable the **java** profile as follows:
```
zjffdu: Update this after you removed java profile

selvinsource:
@zjffdu 
I made the requested changes except zeppelin.interpreter.group.order, should I just remove the whole property? I can't see where it's used to be honest.

java/src/main/java/org/apache/zeppelin/java/JavaInterpreterUtils.java
```diff
@@ -0,0 +1,44 @@
...
+public class JavaInterpreterUtils {
+
+  /**
+   * Convert a map to %table display system to leverage Zeppelin's built in visualization
+   * @param keyName Key column name
+   * @param valueName Value column name
+   * @param rows Map of keys and values
+   * @return Zeppelin %table
+   */
+  public static String displayTableFromSimpleMap(String keyName, String valueName, Map<?, ?> rows){
```
zjffdu: It is fine to use JavaInterpreterUtils for now. But I think it is better to support ZeppelinContext in JavaInterpreter as well. So that user can use z.show to leverage the builtin visualization of zeppelin and also other features provided by ZeppelinContext.

zjffdu:
@selvinsource I have removed property `zeppelin.interpreters`, could you do rebase ?

selvinsource:
@zjffdu I have done the rebase.

java/src/main/java/org/apache/zeppelin/java/JavaInterpreter.java
```diff
@@ -0,0 +1,99 @@
...
+  public void open() {
+
+  }
+
+  @Override
+  public void close() {
+    File dir = new File(".");
+    // delete all .class files created while compilation process
+    for (int i = 0; i < dir.list().length; i++) {
+      File f = dir.listFiles()[i];
```
zjffdu: move dir.listFiles out of this for loop, otherwise listFiles will be called multiple times.

java/src/main/java/org/apache/zeppelin/java/JavaInterpreterUtils.java
zjffdu: Sound good plan. Thanks

java/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    <!--library versions-->
+    <interpreter.name>java</interpreter.name>
+  </properties>
+  
+  <dependencies>
+
+    <dependency>
+      <groupId>com.thoughtworks.qdox</groupId>
+      <artifactId>qdox</artifactId>
+      <version>2.0-M3</version>
```
felixcheung: LICENSE update?

docs/interpreter/java.md
```diff
@@ -0,0 +1,54 @@
...
+-->
+
+{% include JB/setup %}
+
+# Java interpreter for Apache Zeppelin
+
+<div id="toc"></div>
+
+## How to use
+Basically, you can write normal java code. You should write the main method inside a class because the interpreter invoke this main to execute the code. Unlike Zeppelin normal pattern, each paragraph is considered as a separate job, there isn't any relation to any other paragraph.
```
felixcheung: perhaps clarify " each paragraph is considered as a separate job, there isn't any relation to any other paragraph."
does it mean I can't share variables in different paragraphs? I think it will greatly help if this is documented

java/README.md
```diff
@@ -0,0 +1,17 @@
+# Overview
+Java interpreter for Apache Zeppelin
+
+# Architecture
+Current interpreter implementation supports the static repl. It compiles the code in memory, execute it and redirect the output to zeppelin.
```
felixcheung: `repl` -> `REPL`

java/pom.xml
felixcheung: ok I saw the PR comment on this

selvinsource: Yes, it was already updated as part of the Beam interpreter.

docs/interpreter/java.md
selvinsource: At the moment there isn't a way to share variable unless you use a clustered environment.
For example, when I will implement the Hazelcast interpreter, one could store a variable in the hazelcast cluster as part of one job and retrieve the same variable as part of another job.

As in previous comments with @zjffdu when we will move to JShell it will be easier to introduce the support for ZeppelinContext which will allow also to share variables.

java/README.md
selvinsource: Ok I will update this.

java/src/main/java/org/apache/zeppelin/java/JavaInterpreter.java
selvinsource: OK, good point, I will change it shortly.

selvinsource:
@zjffdu @felixcheung I have pushed the requested changes.

zjffdu:
LGTM, will merge soon

felixcheung:
did you add the behavior in doc?

selvinsource:
@zjffdu thanks

@felixcheung should I mention explicitly that one cannot share variables between paragraph? I think other interpreters do not allow that either but it's not mentioned. Instead, in the interpreter that is supported (like Spark) in the doc it mentions one could use the ZeppelinContext.
Also can you confirm that when you say "share variable" you are talking about the ZeppelinContext and its dynamic form support?
Let me know, I can explicitly add that if you prefer.

felixcheung:
no, I mean

paragraph 1
```
%java

Map<String, Long> counts = new HashMap<>();
```

paragraph 2
```
%java

counts.put("key", 1);
```

selvinsource:
@felixcheung 
I see what you mean now and I added explicitly to the Java Interpreter doc that the Java interpreter do not support it.
Eventually, when we move the Java interpreter implementation to JShell we can look into adding a JavaContext that allows to share variables between paragraph as you mentioned.

@zjffdu 
I have rebased once more for an easy merge to the master.

zjffdu:
@selvinsource What @felixcheung mentioned is not sharing varaibles via JavaContext, he mean share variable in the same java repl as spark repl did right now.

selvinsource:
@zjffdu yeah I think I understood and JShell should be able to achieve this.
I meant as JavaContext a way to maintain the same JShell instance to run multiple snipper across paragraph in a similar way (conceptually) the Spark interpreter does by using the SparkContext.
Definitely the aim when introducing JSchell is to support both sharing variables as in the example from @felixcheung as well as using the ZeppelinContext for Zeppelin input/output support. At that point the Java interpreter will have comparable features to the other interpreters. As I said I will add a Jira for this. My next step is to add the HazelCast Jet interpreter once this PR is merged.

zjffdu:
@selvinsource Travis still fails, I will commit after travis is passed

```

[ERROR] Failed to execute goal on project zeppelin-beam: Could not resolve dependencies for project org.apache.zeppelin:zeppelin-beam:jar:0.9.0-SNAPSHOT: Could not find artifact org.apache.zeppelin:zeppelin-java:jar:0.9.0-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
```

felixcheung:
right. just want to clarify that it is perfectly fine to run Scala, Python etc without Spark. The session context is maintained in the REPL implementation and not in SparkContext or ZeppelinContext. I think they are orthogonal.

zjffdu:
@selvinsource I think you need to update this line of .travis.xml  https://github.com/apache/zeppelin/blob/master/.travis.yml#L44

selvinsource:
@zjffdu I have made the change, if you can re-run Jenkins please.

zjffdu:
@selvinsource Do you set up travis ? https://zeppelin.apache.org/contribution/contributions.html

selvinsource:
@zjffdu yes I did before opening the pull request.

I manually re-run it and it's fine now
https://travis-ci.org/selvinsource/zeppelin/builds/411153675

zjffdu:
merging

selvinsource:
@zjffdu @felixcheung thanks for reviewing this PR and merging it in the master. I am going to work on the Hazelcast Jet interpreter next and send a new PR once ready.

mebelousov:
Seems this PR broke build for scala 2.11

After execution of ` ./dev/change_scala_version.sh 2.11` I have got error 
```
[ERROR] Failed to execute goal on project zeppelin-beam: Could not resolve dependencies for project org.apache.zeppelin:zeppelin-beam:jar:0.9.0-SNAPSHOT: Could not find artifact org.apache.zeppelin:zeppelin-scio_2.10:jar:0.9.0-SNAPSHOT in apache.snapshots (http://repository.apache.org/snapshots) -> [Help 1]```

@selvinsource could check maven build?

**Comment by selvinsource** [comment]
@mebelousov 

If you see
https://travis-ci.org/selvinsource/zeppelin/builds/411153675
it seems fine with SCALA_VER="2.11".

What options are you running maven build with after the change of scala version? I could try to re-create in dev environment.

**Comment by selvinsource** [comment]
@zjffdu 
the issue reported by @mebelousov may have been caused by https://issues.apache.org/jira/browse/ZEPPELIN-3557
and not this PR, as I only added in the beam interpreter the dependency on the java interpreter.

If you look at scio/pom.xml it has
```
  <artifactId>zeppelin-scio_2.10</artifactId>
  <packaging>jar</packaging>
  <version>0.9.0-SNAPSHOT</version>
  <name>Zeppelin: Scio</name>
  <description>Zeppelin Scio support</description>
```
So maven will always only build the 2.10 version for scio.

At the same time the beam/pom.xml
has dependency to
```
 <dependency>
      <artifactId>zeppelin-scio_${scala.binary.version}</artifactId>
    </dependency>
```
which do not exists for scala 2.11.

So the solution is either that scio is also built for 2.11 or beam to reference 2.10 no matter what the scala version is.

I think you never noticed it before because the beam interpreter was "opt in" so perhaps nobody noticed this as they only ever built the beam interpreter (-Pbeam) with scala 2.10 and not 2.11.

But now that all profiles have been removed as per ZEPPELIN-3557, the beam interpreter needs the correct scio version.

selvinsource:
@zjffdu @mebelousov I sent you a comment before, please ignore, I thought that scio was only built for 2.10 but I noticed that changing scala version using the script does update the pom for scio to 2.11.

I still think is somehow related to https://issues.apache.org/jira/browse/ZEPPELIN-3557 as before people wouldn't build using profiles by default so there may be some issue uncovered now that all interpreters are built all the times.

In this PR I only changed beam/pom.xml to have a dependency on java interpreter, nothing was changed related to scio or scala versions.

Will investigate further but it would be good to know which options @mebelousov is using to build to re-create locally.

selvinsource:
Found something that could cause the issue.
in beam/pom.xml there is
```<beam.beam.version>2.0.0</beam.beam.version>```

then 
     ``` <artifactId>beam-runners-flink_${scala.binary.version}</artifactId>```

But looking at
https://mvnrepository.com/artifact/org.apache.beam/beam-runners-flink
there isn't a beam-runners-flink_2_10 for beam version 2.0.0.

So even in Zeppelin 0.8.0 if you try to switch to scala 2.11 and run install -DskipTests -Pbeam you will get an error but perhaps that's something nobody ever tried (to compile beam on 2.11).

However after removing all profiles, that's the case now and a solution is needed.

Either the beam version is updated or the beam-runners flink can be 2_10 even for scala 2.11.


@zjffdu what do you think about my analysis? Am I making sense?

zjffdu:
Thanks @selvinsource for the investigation, the analysis seems correct to me.

felixcheung:
so we need to open a JIRA then?

selvinsource:
@felixcheung yes I think so as it's not something new, it has been there since the beam interpreter was created but nobody noticed until now that all interpreter specific profiles have been removed.

@mebelousov for now the workaround is simple to change beam/pom.xml to
```<artifactId>beam-runners-flink_2.10</artifactId>```
before you build zeppelin.

selvinsource:
@felixcheung @zjffdu 
FYI, I have added https://issues.apache.org/jira/browse/ZEPPELIN-3742 based on the discussion we had in this PR.

selvinsource:
@zjffdu @felixcheung do you have any estimate on when version 0.9 is going to be released? Looking forwawrd to see the Java interpreter included in the downloadable version.

felixcheung:
oh wait, I didn't realize this comment https://github.com/apache/zeppelin/pull/3092#issuecomment-410957256 is not related to ZEPPELIN-3742 

do we need a fix for `beam-runners-flink_2.10`

selvinsource:
> oh wait, I didn't realize this comment [#3092 (comment)](https://github.com/apache/zeppelin/pull/3092#issuecomment-410957256) is not related to ZEPPELIN-3742
> 
> do we need a fix for `beam-runners-flink_2.10`

As I mentioned before, that's nothing new, it's been there since the Beam interpreter was added, it never supported scala 2.11, see the notes on
https://github.com/apache/zeppelin/blob/master/beam/README.md
"Flink runner comes with binary compiled for scala 2.10. So, currently we support only Scala 2.10".

By doing a quick search somebody else already created JIRAs for this
https://issues.apache.org/jira/browse/ZEPPELIN-3903
https://issues.apache.org/jira/browse/ZEPPELIN-3899

@novelari perhaps is in the best position to update the beam-runners-flink to a most up to date version to support both scala 2.10 and 2.11, I am not that familiar with Beam/Flink.

A solution could be to use profiles in beam/pom.xml to activate either 
https://mvnrepository.com/artifact/org.apache.beam/beam-runners-flink_2.10/2.2.0
or
https://mvnrepository.com/artifact/org.apache.beam/beam-runners-flink_2.11/2.3.0
depending on the scala version.

For example
<code>
<profiles>
    <profile>
      <id>scala-2.10</id>
      <activation>
        <activeByDefault>true</activeByDefault>
      </activation>
        <dependencies>
<dependency>
    <groupId>org.apache.beam</groupId>
    <artifactId>beam-runners-flink_2.10</artifactId>
    <version>2.2.0</version>
</dependency>
        </dependencies>
    </profile>
    <profile>
      <id>scala-2.11</id>
        <dependencies>
<dependency>
    <groupId>org.apache.beam</groupId>
    <artifactId>beam-runners-flink_2.11</artifactId>
    <version>2.3.0</version>
</dependency>
        </dependencies>
    </profile>
</profiles>
</code>

felixcheung:
ah yes, thanks for looking into this. it's a long standing problem with scala version in scala builds. I think it's a good idea to upgrade beam

sgloutnikov:
@selvinsource thank. you! Saved me a good amount of time investigating why I couldn't build with Scala 2.11. Was able to build fine with `beam-runners-flink_2.11:2.3.0`.

PR #3093, opened 2018-07-24
[ZEPPELIN-3655] Add missing roles information to api/notebook/job/ method

oxygen311:
### What is this PR for?
Fix bugs with private access interpreter launched with REST by adding missing roles information to `AuthenticationInfo`.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3655](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3655)

### Questions:


zjffdu:
Can you add unit test ?

oxygen311:
It's don't easy to check this bug with a unit test. Even in Paragraph is not running because of  "user has no permission for interpreter" response in rest api would be "OK"

zjffdu:
@oxygen311  Could you resolve conflict ?

bin/common.sh
```diff
@@ -16,6 +16,10 @@
 # limitations under the License.
 #
 
+# -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+# ZEPPELIN_JAVA_OPTS
+export ZEPPELIN_JAVA_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005"
```
zjffdu: Please remove it

oxygen311:
@zjffdu done

zjffdu:
It looks like all the public api in NotebookRestApi miss the role info. Do you see any issue when roles is missing ? I think it is better to set roles properly in some kind of Filter.

oxygen311:
@zjffdu 
Also this bug may occur in run paragraph methods, it's also fixed now.
Other methods don't miss role info because they don't need to pass it farther. And `checkIfUserCanWrite` and `checkIfUserCanWrite` which is used in most rest api calls are take into roles information.

zjffdu:
Thanks @oxygen311 LGTM

PR #3095, opened 2018-07-25
ZEPPELIN-3652. Remove reflection in SparkInterpreter

zjffdu:
### What is this PR for?
This is a followup ticket of ZEPPELIN-3635. This PR remove some legacy code about old spark versions where we have to use reflection at that time.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3652

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
@@ -19,23 +19,27 @@ package org.apache.zeppelin.spark;
 
 import com.google.common.collect.Lists;
 import org.apache.spark.SparkContext;
-import org.apache.spark.sql.SQLContext;
-import org.apache.spark.sql.catalyst.expressions.Attribute;
 import org.apache.zeppelin.annotation.ZeppelinApi;
 import org.apache.zeppelin.display.AngularObjectWatcher;
-import org.apache.zeppelin.display.Input;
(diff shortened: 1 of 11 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkZeppelinContext.java
```diff
@@ -91,88 +92,8 @@ public SparkZeppelinContext(
...
-    msg.append("\n");
-
-    // ArrayType, BinaryType, BooleanType, ByteType, DecimalType, DoubleType, DynamicType,
-    // FloatType, FractionalType, IntegerType, IntegralType, LongType, MapType, NativeType,
-    // NullType, NumericType, ShortType, StringType, StructType
-
-    try {
-      for (int r = 0; r < maxResult && r < rows.length; r++) {
-        Object row = rows[r];
-        Method isNullAt = row.getClass().getMethod("isNullAt", int.class);
```
felixcheung: null check is gone in the new version?

zjffdu: Not necessary. Row.get(i) will return null directly. The unit test here cover this case. https://github.com/apache/zeppelin/pull/3095/files#diff-b3caeb6afe474aec6d5352ac3ee3a02aR195

zjffdu:
Will merge if no more comment

PR #3098, opened 2018-07-26
ZEPPELIN-3664: ActiveDirectoryGroupRealm returns "cn" instead of "userPrincipalName" for note permission auto completion

florian-trimble:
* Return "userPrincipalName" from ActiveDirectoryGroupRealm when searching for users.
* Added unit test for search functionality.

### What is this PR for?
Fix for https://issues.apache.org/jira/browse/ZEPPELIN-3664
Return the "userPrincipalName" instead of the "cn" (common name) from ActiveDirectoryGroupRealm when searching users to set note permissions.

### What type of PR is it?
[Bug Fix]

### Todos
* -

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3664

Steps to reproduce
1. Configure connection to AD with org.apache.zeppelin.realm.ActiveDirectoryGroupRealm
2. Login to Zeppelin
3. Create new notebook
4. Edit the permissions and start typing the name of a AD user to one of the permission fields
5. Select one of the suggestions and save
6. Check the content of "notebook-authorization.json"

Expected result
* The "userPrincipalName" of the user is shown in the suggestion box and written to "notebook-authorization.json" (see screenshot attached).

Actual result
* The "cn" (common name) of the user is shown in the suggestion box and written to "notebook-authorization.json".

### Screenshots
notepermission

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
@@ -265,7 +265,7 @@ public class ActiveDirectoryGroupRealm extends AbstractLdapRealm {
         NamingEnumeration ae = attrs.getAll();
         while (ae.hasMore()) {
           Attribute attr = (Attribute) ae.next();
-          if (attr.getID().toLowerCase().equals("cn")) {
+          if (attr.getID().toLowerCase().equals("userprincipalname")) {
             userNameList.addAll(LdapUtils.getAllAttributeValues(attr));
           }
         }
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealmTest.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealmTest.java
@@ -1,0 +1,50 @@
+package org.apache.zeppelin.realm;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.List;
+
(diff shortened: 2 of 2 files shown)


mebelousov:
@fsteinle our team use authorization with parameter 
`activeDirectoryRealm.principalSuffix = @company.com`
Thus email is not necessary. Could it possible to use mailNickName, for example?

BTW, `userPrincipalName` is good.

fsteinle:
@mebelousov: Basically this is not related to emails but to the userPrincipalName. The principalSuffix parameter makes sense, but doesn't address this issue. In our case we wanna keep the suffix to be sure it's really unique. Not all users of the organization are in the same AD branch.

zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
```diff
@@ -265,7 +265,7 @@ protected AuthorizationInfo buildAuthorizationInfo(Set<String> roleNames) {
         NamingEnumeration ae = attrs.getAll();
         while (ae.hasMore()) {
           Attribute attr = (Attribute) ae.next();
-          if (attr.getID().toLowerCase().equals("cn")) {
+          if (attr.getID().toLowerCase().equals("userprincipalname")) {
```
prabhjyotsingh: Should we make this (cn/userprincipalname/or any other parameter) configurable?

fsteinle: From my point of view "cn" just doesn't work. So I don't see a need having this option.
When adding a new notebook also the upn is set as owner (or also for readers/writers/runners depending on the config). So why should one use the "cn" here when searching?
It also actually is the upn we're searching for in this method ("userPrincipalName=*" + containString + "*").

But maybe I do not have the full picture...

zjffdu:
@fsteinle Could you update the PR title ?

PR #3100, opened 2018-07-27
ZEPPELIN-3557. Remove interpreter profiles

zjffdu:
### What is this PR for?
Just remove all interpreter profiles, instead of disable them by default. So that we won't forget to build them in travis. 

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3557

* CI

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,14 @@ addons:
 env:
   global:
     # Interpreters does not required by zeppelin-server integration tests
-    - INTERPRETERS='!hbase,!pig,!jdbc,!file,!flink,!ignite,!kylin,!lens,!cassandra,!elasticsearch,!bigquery,!alluxio,!scio,!livy,!groovy,!sap'
+    - INTERPRETERS='!beam,!hbase,!pig,!jdbc,!file,!flink,!ignite,!kylin,!lens,!cassandra,!elasticsearch,!bigquery,!alluxio,!scio,!livy,!groovy,!sap,!r'
 
 matrix:
   include:
     # Test License compliance using RAT tool
     - jdk: "oraclejdk8"
       dist: trusty
-      env: BUILD_PLUGINS="false" SCALA_VER="2.11" PROFILE="-Prat -Pr" BUILD_FLAG="clean" TEST_FLAG="org.apache.rat:apache-rat-plugin:check" TEST_PROJECTS=""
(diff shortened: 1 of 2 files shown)


zjffdu:
@jongyoul Could you help review it ? Thanks

jongyoul:
What I thought is to move all interpreters modules to optional. For instance, running `mvn package` will create only zeppelin-* modules, and if users want to build spark, users should do add `-P spark` as an option like `mvn package -Pspark`. What do you think of it? The main reason why I'd like to do it is to make Zeppelin's code more friendly for IDE. WDYT?

zjffdu:
I am afraid currently we can not do this. Because some unit test in zeppelin-zengine & zeppelin-server depends on spark interpreter (ZeppelinSparkClusterTest).

jongyoul:
Agreed. Let's merge it and proceed them with another PR.

jongyoul:
BTW, did you test if sparkR works with `-Pr`?

zjffdu:
Actually this is my long time confusion part. `-Pr` would copy its artifact to `interpreter/spark` folder but it never used. I don't know the history of `-Pr`, is it an existing issue ?

jongyoul:
@zjffdu Yes, it has a very long story. I, personally, don't want to keep both of them, but didn't find any solution to fit both requirements. `-Pr` is pure R interpreter which supported - I didn't test it - spark inside it. AFAIK, no one tests if SparkR works for now with `-Pr`.

zjffdu:
Right, I think no one is using the pure r interpreter. I think we can use it 2 ways.
1. Create a pure r interpreter without spark support.
2. Use it another r implementing of spark and make option to allow user to choose whatever r implementation they want to use. (Because I don't know which r implementation is better, so I don't want to remove anyone of them)

jongyoul:
We've tried to option 1 before but failed. We'd better give options to users like option 2

zjffdu:
Thanks @jongyoul I created ZEPPELIN-3676 for this. and will merge this ticket first.

egorklimov:
@zjffdu, please look, smth went wrong: https://issues.apache.org/jira/browse/ZEPPELIN-3682

PR #3102, opened 2018-07-30
[WIP][ZEPPELIN-3671] Add info about running interpreters PID to API and JMX

egorklimov:
### What is this PR for?
It would be nice if we could get PID of running interpreter, and group of paragraphs that running under that interpreter, using API and JMX.

Using this feature it will be easy to check CPU and memory load, etc. 
This PR adds:
* API method to get info about running interpreters (`/api/interpreter/running`)
* API method to get info about running paragraphs grouped by interpreters (`/api/notebook/jobmanager/running`)
* Few JMX methods which do the same as API.
* Template for simple running paragraphs analysis using API.

### What type of PR is it?
Improvement

### What is the Jira issue?
Issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-3671

* CI is green

Example of tables built on response data:
* Interpreters:
intp-table-example
* Paragraphs:
running-paragraphs-info
* Count of paragraphs belongs to each interpreter:
paragraph-example

Example of cpu and memory load analysis:
* Pie chart for memory load
memory-example
* Pie chart for cpu load
cpu-example

Tree of processes associated with running:
tree

The same data using JMX (viewed in jconsole):

* Running interpreters:
jconsole2
* Running paragraphs
jconsole-example

### Questions:
* Does this needs documentation? Yes, info about API should be updated


zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -205,7 +205,7 @@ public ZeppelinServer() throws Exception {
     this.interpreterService = new InterpreterService(conf, interpreterSettingManager);
 
     // Register MBean
-    if ("true".equals(System.getenv("ZEPPELIN_ENABLE_JMX"))) {
+    if ("true".equals(System.getenv("ZEPPELIN_JMX_ENABLE"))) {
```
zjffdu: It looks like we didn't put it zeppelin-env.sh.template

egorklimov: @zjffdu It was a typo in system variable name defined in zeppelin-env.sh.template.

egorklimov:
There are some troubles during processing `run` folder

felixcheung:
cool!

zeppelin-server/pom.xml
```diff
@@ -275,6 +276,13 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
```
zjffdu: What is the license of this dependency ? You may need to update the LICENCE file.

zjffdu:
Thanks @egorklimov , this is an interesting feature. 
This assume all the interpreter process will generate pid file, but this assumption is not true. There's 2 exceptions at least for now. One is  yarn-cluster mode of spark where the interpreter runs in remote node of yarn cluster.  Another is running interpreter in container which is on our roadmap. Do you have any ideas of how to handle these 2 scenarios ?

zeppelin-server/pom.xml
egorklimov: You are right, I fixed it.

egorklimov:
@zjffdu If I'm not mistaken according to `interpreter.sh`:
```
HOSTNAME=$(hostname)
ZEPPELIN_SERVER=org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer

INTERPRETER_ID=$(basename "${INTERPRETER_DIR}")
ZEPPELIN_PID="${ZEPPELIN_PID_DIR}/zeppelin-interpreter-${INTERPRETER_ID}-${ZEPPELIN_IDENT_STRING}-${HOSTNAME}-${PORT}.pid"
ZEPPELIN_LOGFILE="${ZEPPELIN_LOG_DIR}/zeppelin-interpreter-${INTERPRETER_SETTING_NAME}-"
``` 
Pid file generates every time (except case when interpreter process didn't start successfully). But in yarn scenario it will be hard to analyze CPU and memory load because resources will be consumed on each node, maybe someone will add one more template for that case. 

In case of container, I suppose we could generate other info in run folder.

Now my tests fail third Travis test, trying to fix that scenario ;)

PR #3103, opened 2018-07-31
ZEPPELIN-3666. Use zeppelin.interpreter.default to replace zeppelin.interpreter.group.order

zjffdu:
### What is this PR for?

Actually user only care about the default interpreter rather than the whole interpreter order, so that we can use zeppelin.interpreter.default to replace zeppelin.interpreter.group.order, otherwise we may miss to update zeppelin.interpreter.group.order when adding new interpreter

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3666

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -701,9 +701,7 @@ public class ZeppelinConfiguration extends XMLConfiguration {
         "http://repo1.maven.org/maven2/"),
     ZEPPELIN_INTERPRETER_CONNECT_TIMEOUT("zeppelin.interpreter.connect.timeout", 60000),
     ZEPPELIN_INTERPRETER_MAX_POOL_SIZE("zeppelin.interpreter.max.poolsize", 10),
-    ZEPPELIN_INTERPRETER_GROUP_ORDER("zeppelin.interpreter.group.order", "spark,md,angular,sh,"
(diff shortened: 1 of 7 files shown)


Tagar:
@zjffdu, do I understand correctly that `zeppelin.interpreter.default` would choose default interpreter, and not just interpreter group?
If so, it would also resolve https://issues.apache.org/jira/browse/ZEPPELIN-3282
Thanks!

Tagar:
Thanks @zjffdu - that's a great improvement as ordering wasn't used in zeppelin.
Would it be possible to introduce a similar knob to choose a default interpreter, not just default interpreter group?

Tagar:
As discussed in https://github.com/apache/zeppelin/pull/3059 it seems a good idea to have a **default interpreter group** set globally, and **default interpreter set at note level**. What do you think? 
Thanks!

zjffdu:
@Tagar Allow user to customize default interpreter make sense to me. Maybe this ticket is a little misleading. Let me clarify if more.  Here the property `zeppelin.interpreter.default` is to set the default interpreter group when user creating note. But user can still choose other interpreter group when creating note. 
screen shot 2018-08-01 at 8 24 28 am

zjffdu:
I change it to `zeppelin.interpreter.group.default` to avoid misunderstanding

zjffdu:
@felixcheung  This is for default interpreter group, I have updated it.

felixcheung:
I see, would be good to update the title - I think that goes into the actual commit

zjffdu:
git commit message has been updated.

PR #3104, opened 2018-07-31
ZEPPELIN-3668. Can't hide Spark Jobs (Spark UI) button

zjffdu:
### What is this PR for?

This is to fix the bug of unable to hide spark jobs. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3668

* Unit test added
* Verify it manually as well.

### Questions:

### DIFF
--- a/spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
+++ b/spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
@@ -44,7 +44,6 @@ import java.net.URL;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -53,6 +52,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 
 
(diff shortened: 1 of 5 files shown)


zjffdu:

maziyarpanahi:
Hi @zjffdu,
Is it possible to have `spark.ui.enabled: true`, but hide the `Spark Jobs` in UI? Or should we not mention `spark.ui.enabled` at all if we wish not to see the `Spark Jobs` button in UI?

Many thanks

zjffdu:
Why would you want to have spark.ui.enabled: true but hide the spark jobs in UI ?

maziyarpanahi:
We don’t provide proxy or any multi-user Spark UI solutions to our data scientists. But as a SysAdmin I need to see the jobs etc. in order to diagnose the problems they have time to time. Also I use it to see if there is other problem with the rest of the cluster (YARn/HDFS).

In Zeppelin 0.7.3 I never mentioned anything about Spark UI and the Zeppelin UI didn’t display this button at all. I know it is nice to have this option by default as people try it on localhost and the UI is useful and accessible but this is not the case in multi-user scenarios as Spark is not offering any multi-user option for its UI right out of the box.
 That is why I think it is important to have a setting for Zeppelin to allow admins to turn this feature off (hidden) if they only use Spark UI for debugging. (This can be zeppelin.spark.ui.hidden:false by default for backward compatibility) 
What do you think?

zjffdu:
Thanks @maziyarpanahi It makes sense to me. I created ZEPPELIN-3679 for it

zjffdu:
@maziyarpanahi I make a pr for this #3109

maziyarpanahi:
Many thanks @zjffdu

PR #3106, opened 2018-08-01
[ZEPPELIN-3677] SAP Interpreter improvement

oxygen311:
### What is this PR for?
Improvements:
- An empty result processing, message about no data will be shown;
- Dimension's attributes auto complete and processing;
- Property `universe.queryTimeout` for socket and connection timeouts;
- Interpolation support and property for it;
- `completionSupport` flag fix;
- Support of `distinct` keyword (written after `select`);
- Support of `limit %d` keyword (written in the end of query);
- Add concurrency support, relevant properties are `universe.concurrent.use` and `universe.concurrent.maxConnection`;
- `;` is adding to the end of query if it has not it.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-3677](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3677?filter=allopenissues)

### Questions:
* Does this need documentation? Updated

### DIFF
--- a/docs/interpreter/sap.md
+++ b/docs/interpreter/sap.md
@@ -123,7 +123,7 @@ where
 ```
 
 ## Object Interpolation
-The JDBC interpreter also supports interpolation of `ZeppelinContext` objects into the paragraph text.
+The SAP interpreter also supports interpolation of `ZeppelinContext` objects into the paragraph text.
 To enable this feature set `universe.interpolation` to `true`. The following example shows one use of this facility:
 
 ####In Scala cell:
--- a/docs/interpreter/sap.md
+++ b/docs/interpreter/sap.md
@@ -121,3 +121,21 @@ where
 
   and [Folder1].[Dimension5] in ('Value1', 'Value2');
 ```
+
+## Object Interpolation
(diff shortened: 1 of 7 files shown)


docs/interpreter/sap.md
```diff
@@ -121,3 +121,21 @@ where
...
 ```
+
+## Object Interpolation
+The SAP interpreter also supports interpolation of `ZeppelinContext` objects into the paragraph text.
+To enable this feature set `universe.interpolation` to `true`. The following example shows one use of this facility:
+
+####In Scala cell:
+
+```scala
+z.put("curr_date", "2018-01-01 00:00:00")
```
zjffdu: In you scenario, do you just need scala repl or needs spark support ? The reason I ask is that I am thinking whether we should introduce pure scala interpreter

mebelousov: Tinkoff team use only spark (pyspark) interpreter.
I don't feel the necessary of pure scala interpreter.

zjffdu:
LGTM

PR #3108, opened 2018-08-02
ZEPPELIN-3678. Change SparkZeppelinContext to scala implementation

zjffdu:
### What is this PR for?

This PR just to change SparkZeppelinContext to scala implementation, so that it is easy for us to implement SparkZeppelinContext via scala instead of java. This also make SparkZeppelinContext more readable and maintainable.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3678

* CI pass

### Questions:


spark/interpreter/src/main/scala/org/apache/zeppelin/spark/SparkZeppelinContext.scala
```diff
@@ -0,0 +1,148 @@
...
+                           val sparkShims: SparkShims,
+                           val hooks2: InterpreterHookRegistry,
+                           val maxResult2: Int) extends BaseZeppelinContext(hooks2, maxResult2) {
+
+  private val interpreterClassMap = Map(
+    ("spark", "org.apache.zeppelin.spark.SparkInterpreter"),
+    ("sql", "org.apache.zeppelin.spark.SparkSqlInterpreter"),
+    ("dep", "org.apache.zeppelin.spark.DepInterpreter"),
+    ("pyspark", "org.apache.zeppelin.spark.PySparkInterpreter")
+  )
```
sanjaydasgupta: In favor of more idiomatic Scala:

    "spark" -> "org.apache.zeppelin.spark.SparkInterpreter",
    "sql" -> "org.apache.zeppelin.spark.SparkSqlInterpreter",
    "dep" -> "org.apache.zeppelin.spark.DepInterpreter",
    "pyspark" -> "org.apache.zeppelin.spark.PySparkInterpreter"

spark/interpreter/src/main/scala/org/apache/zeppelin/spark/SparkZeppelinContext.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  try
+    supportedClasses += Class.forName("org.apache.spark.sql.DataFrame")
+  catch {
+    case e: ClassNotFoundException =>
+
+  }
+  if (supportedClasses.isEmpty) throw new RuntimeException("Can not load Dataset/DataFrame class")
+
```
sanjaydasgupta: Since both the supported classes are part of the Scala SDK, should it be safe to just define it like this, and avoid the `forName` calls:

 ```
 val supportedClasses: scala.collection.mutable.Seq[Class[_]] = scala.collection.mutable.Seq(
      classOf[org.apache.spark.sql.Dataset[_]], classOf[org.apache.spark.sql.DataFrame])

```

sanjaydasgupta:
Hi @zjffdu I have added a couple of simple suggestions to the Scala code.

spark/interpreter/src/main/scala/org/apache/zeppelin/spark/SparkZeppelinContext.scala
zjffdu: I am afraid not, because zeppelin needs to support both spark 1.x and spark 2.x. And Dataset is not available in spark 1.x

spark/interpreter/src/main/scala/org/apache/zeppelin/spark/SparkZeppelinContext.scala
zjffdu: Fixed

PR #3109, opened 2018-08-02
ZEPPELIN-3679. Provide option to allow hide spark ui in frontend

zjffdu:
### What is this PR for?
Introducing new property `zeppelin.spark.ui.hidden` to control whether hide spark ui in frontend. By default it is false.

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3679

* Unit test is added

### Questions:


zjffdu:

PR #3110, opened 2018-08-03
[ZEPPELIN-3671] Add info about running interpreters PIDs to API and JMX

egorklimov:
### What is this PR for?
This is a continuation of [3102](https://github.com/apache/zeppelin/pull/3102)

It would be nice if we could get PID of running interpreter, and group of paragraphs that running under that interpreter, using API and JMX.

Using this feature it will be easy to check CPU and memory load, etc. 

This PR adds:
* API method to get info about running interpreters (`/api/interpreter/running`);
* API method to get info about running paragraphs grouped by interpreters (`/api/notebook/jobmanager/running`);
* Few JMX methods which do the same as API;
* Template for simple running paragraphs analysis using API.

Part of discussion from previous PR:
> author: @zjffdu 
> Thanks @egorklimov , this is an interesting feature.
> This assume all the interpreter process will generate pid file, but this assumption is not true. There's 2 exceptions at least for now. One is yarn-cluster mode of spark where the interpreter runs in remote node of yarn cluster. Another is running interpreter in container which is on our roadmap. Do you have any ideas of how to handle these 2 scenarios ?

> author: @egorklimov
> If I'm not mistaken according to `bin/interpreter.sh:234-238`:
> ```
> if [[ -z "${pid}" ]]; then
>   exit 1;
> else
>   echo ${pid} > ${ZEPPELIN_PID}
> fi
> ``` 
> Pid file generates every time (except case when interpreter process didn't start successfully). But in yarn scenario it will be hard to analyze CPU and memory load because resources will be consumed on each node, maybe someone will add one more template for that case.
> 
> In case of container, I suppose we could generate other info in run folder.

### What type of PR is it?
Improvement

### What is the Jira issue?
Issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-3671

* CI failed on [third test](https://travis-ci.org/TinkoffCreditSystems/zeppelin/jobs/411783078), because of `org.apache.zeppelin.rest.ZeppelinRestApiTest.testJobs()` with job time limit error;
* Tests added.

### Screenshots
Example of tables built on response data:
* Interpreters:
intp-table-example
* Paragraphs:
running-paragraphs-info

Tree of processes associated with running:
tree

The same data using JMX (viewed in jconsole):

* Running interpreters:
jconsole2
* Running paragraphs
jconsole-example

### Questions:
* Does the licenses files need update? Yes, Common Public License Version 1.0 added
* Does this needs documentation? Yes, info about API should be updated


zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -205,7 +205,7 @@ public ZeppelinServer() throws Exception {
     this.interpreterService = new InterpreterService(conf, interpreterSettingManager);
 
     // Register MBean
-    if ("true".equals(System.getenv("ZEPPELIN_ENABLE_JMX"))) {
+    if ("true".equals(System.getenv("ZEPPELIN_JMX_ENABLE"))) {
```
egorklimov: It was a typo in system variable name defined in zeppelin-env.sh.template.

egorklimov:
CI pass: https://travis-ci.org/TinkoffCreditSystems/zeppelin/builds/416708180 
(with cherry picked [ZEPPELIN-3715](https://github.com/apache/zeppelin/pull/3148) otherwise it [falls](https://travis-ci.org/TinkoffCreditSystems/zeppelin/builds/416708280) because of REST API bug)

egorklimov:
@zjffdu Thank you for review in [3102](https://github.com/apache/zeppelin/pull/3102), I fixed tests.

Our team is going to use this feature in our fork. If it  can not be merged in master because cluster mode is not supported, then I'll close this PR.

PR #3113, opened 2018-08-06
[ZEPPELIN-3616] fix editor sections auto-collapse

Savalek:
# Description
This PR change logic of add new event listener on frontend. Now then add new event listener, also added destroy listener. Destroy listener attach to some DOM element and then this element destroyed listener drop event listener. 
Since the old listeners were not deleted when changing the size of the paragraph / window, an error occurred. Because of this error, updating the text in the paragraph did not happen until the end.

### List of all resize event (after some time of work)
<p align="center">before (77 lis.)</p>

before

<p align="center">after (16 lis.)</p>

after

### Questions:


zjffdu:
ping @prabhjyotsingh @r-kamath

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -755,8 +755,13 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
       }
 
       autoAdjustEditorHeight(_editor);
-      angular.element(window).resize(function() {
-        autoAdjustEditorHeight(_editor);
+
```
prabhjyotsingh: @Savalek instead of this; could we have done `angular.element(window).unbind('resize');` in `$scope.$on('$destroy', function() {`, would it yield same result?

Savalek: @prabhjyotsingh, in this case, the replacement of this option is possible. Change it to your version?

prabhjyotsingh: Sure, as I think this is bit more readable.

prabhjyotsingh:
Apart from minor comment, rest LGTM

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
Savalek: Сhanged it.

Savalek:
What about merge?

Tagar:
cc @zjffdu @Leemoonsoo @prabhjyotsingh can you please merge this? 

Thank you.

prabhjyotsingh:
Will merge this, if no more discussion.

PR #3116, opened 2018-08-08
[ZEPPELIN-3690] display all column with the same name in ui-grid

Savalek:
### Description
This PR allows you to correctly display tables with the same column names.
Also removes a little unnecessary code.

### Info
PR type:   `Bug Fix`
Jira issue [`ZEPPELIN-3690`](https://issues.apache.org/jira/browse/ZEPPELIN-3690)
paragraph's text : [`paragraph-text.txt`](https://github.com/apache/zeppelin/files/2270553/paragraph-text.txt)

### Screenshots
<p align="center">Before PR</p>

before

<p align="center">After PR</p>

after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/tabledata/tabledata.js
+++ b/zeppelin-web/src/app/tabledata/tabledata.js
@@ -32,7 +32,6 @@ export default class TableData extends Dataset {
 
     let columnNames = [];
     let rows = [];
-    let array = [];
     let textRows = paragraphResult.msg.split('\n');
     let comment = '';
     let commentRow = false;
@@ -54,7 +53,6 @@ export default class TableData extends Dataset {
       }
       let textCols = textRow.split('\t');
       let cols = [];
-      let cols2 = [];
       for (let j = 0; j < textCols.length; j++) {
         let col = textCols[j];
         if (i === 0) {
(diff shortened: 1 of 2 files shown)


zeppelin-web/src/app/tabledata/tabledata.js
```diff
@@ -64,12 +62,10 @@ export default class TableData extends Dataset {
             col = valueOfCol;
           }
           cols.push(col);
-          cols2.push({key: (columnNames[i]) ? columnNames[i].name : undefined, value: col});
```
felixcheung: you need to handle the unset columnNames[i] case in visualization-table.js?

Savalek: No. I deleted these lines because `array` is not used anywhere.

felixcheung: I mean, it seems `columnNames[i]` can be null or unset.

Savalek: Yes. It would be correct to use `j` instead of `i`

zjffdu:
ping @prabhjyotsingh @r-kamath

aka-demik:
> ping @prabhjyotsingh @r-kamath @zjffdu 

Any progress?

prabhjyotsingh:
Tested on local.
LGTM

PR #3117, opened 2018-08-08
[ZEPPELIN-3691] Scheduled note continues to work in the trash

oxygen311:
### What is this PR for?
Fix bug with scheduled note working in the trash.

How to repeat bug:
1. Make scheduled note with cron;
2. Move it to trash;
3. Restart zeppelin.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3691](https://issues.apache.org/jira/browse/ZEPPELIN-3691)

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -1020,6 +1020,9 @@ public void removeCron(String id) {
     } catch (SchedulerException e) {
       logger.error("Can't remove quertz " + id, e);
     }
+    Note notebook = getNote(id);
+    Map<String, Object> config = notebook.getConfig();
+    config.put("cron", null);
```
felixcheung: why is the note still getting run if `quartzSched.deleteJob` is called?

oxygen311: @felixcheung 
When zeppelin is restarting it bypass all notes and look at them's cron config. So note again become cronable.

felixcheung: when the note is deleted it should go to trash, right? is there something I've missed here? it's still found when zeppelin starts again?

oxygen311: @felixcheung 
The difference with trashed note is only `~Trash/` prefix. So zeppelin can found it.

oxygen311:
@felixcheung 
I have amended commit for checking is note in trash.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -1020,6 +1020,11 @@ public void removeCron(String id) {
     } catch (SchedulerException e) {
       logger.error("Can't remove quertz " + id, e);
     }
+    Note note = getNote(id);
+    if (note.isTrash()) {
+      Map<String, Object> config = note.getConfig();
+      config.remove("cron");
```
zjffdu: Shouldn't we check whether it is in trash when running cron job ? Otherwise if we restore the note from trash, the cron config is lost

oxygen311:
@zjffdu 
So, I moved this check to `refreshCron()` method

zjffdu:
Thanks @oxygen311  Could you add any unit test ?

oxygen311:
@zjffdu 
Added

zjffdu:
LGTM, @oxygen311 Can you paste your travis link ? I can not find it, maybe there's something wrong with the integration of jenkins.

zjffdu:
@oxygen311 Could you fix the conflict ?

oxygen311:
@zjffdu resolved

zjffdu:
Travis fails, could you fix that ?

oxygen311:
@zjffdu 
Looks like it's some bug with dependencies in master: `[ERROR] SLF4J: Class path contains multiple SLF4J bindings.`

oxygen311:
@zjffdu 
Sorry, you are right. Fixed

PR #3118, opened 2018-08-08
[zeppelin-3693] Option to toggle chart settings of paragraph

saravsars:
### What is this PR for?
To add an option to toggle chart settings of paragraph

Reasons:
* Reducing the paragraph width less than 4 breaks UI
* Hiding the chart settings will give dashboard feel
* Chart selector length increases when multiple helium packages are enabled

### What type of PR is it?
Improvement

### Todos
* Add integration test

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3693

* CI Pass
* Yet to add integration test

before:
before_1

after:
after_1

### Questions:
* Does this needs documentation? Basic documentation added

### DIFF
--- a/docs/quickstart/explore_ui.md
+++ b/docs/quickstart/explore_ui.md
@@ -107,6 +107,7 @@ On the top-right corner of each paragraph there are some commands to:
 * execute the paragraph code
 * hide/show `code section`
 * hide/show `result section`
+* hide/show `chart settings section`
 * configure the paragraph
 
 To configure the paragraph, just click on the gear icon:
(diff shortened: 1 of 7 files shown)


mebelousov:
HI @saravsars !
Could you check `report` display mode described at last element in [Note toolbar](http://zeppelin.apache.org/docs/0.8.0/quickstart/explore_ui.html#note-toolbar)?

saravsars:
Hi @mebelousov 
We are already using `reports` mode but we want to add this option because `report` mode is very plain and we love the card layout in `default` or `simple` mode

mebelousov:
@saravsars 
I'm not sure that your feature will merge.

How about to introduce new display mode?

felixcheung:
I'd agree, this seems like the intent of the report mode. maybe you can add a option to report mode instead to keep the frame for the chart?

saravsars:
@mebelousov @felixcheung thanks for the suggestions.
will make the changes and reopen it.

PR #3120, opened 2018-08-09
ZEPPELIN-3697. Remove interpreter completion for magic part

zjffdu:
### What is this PR for?
Trivial PR to remove interpreter completion for magic part. Code completion is sufficient for users, I don't think user use code completion for magic part (repl name). 

### What type of PR is it?
[ Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3697

* CI pass

### Questions:


zjffdu:
@felixcheung @jongyoul Could you help review it ? Thanks

zjffdu:
Thanks @jongyoul will do that, actually the interpreter completion never works, I don't think people used it before.

PR #3134, opened 2018-08-13
ZEPPELIN-3703. Introduce ConfigurationService

zjffdu:
### What is this PR for?
This is sub task of ZEPPELIN-3288.

* Code refactoring of ZeppelinConfiguration, use Java 8 lambda function`Predicate` instead of ConfigurationKeyPredicate
* Introduce `AbstractRestApi` which has the common functions

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3703

* CI pass

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/AbstractRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/AbstractRestApi.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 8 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/rest/AbstractRestApi.java
```diff
@@ -0,0 +1,52 @@
...
+    return new ServiceContext(authInfo, userAndRoles);
+  }
+
+  public static class RestServiceCallback<T> extends SimpleServiceCallback<T> {
+
+    @Override
+    public void onFailure(Exception ex, ServiceContext context) throws IOException {
+      super.onFailure(ex, context);
+      if (ex instanceof WebApplicationException) {
+        throw (WebApplicationException) ex;
```
felixcheung: why change the type?

zjffdu: Just to match the `IOException` type that this method could throw

PR #3136, opened 2018-08-13
ZEPPELIN-3699. Remove the logic of converting single result to multiple results

zjffdu:
### What is this PR for?
It is a refactoring PR. Zeppelin introduce multiple interpreter result in 0.7, and also introduce logic of converting single result to multiple results. I think it it good time to remove that for 0.9, as I believe no one is using zeppelin before 0.7 for now.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3699

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/Job.java
@@ -38,7 +38,6 @@ import java.util.Map;
  * Changing/adding/deleting non transitive field name need consideration of that.
  */
 public abstract class Job<T> {
-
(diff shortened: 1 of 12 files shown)


zjffdu:
Which api do you mean ? @felixcheung

felixcheung:
Paragraph or REST API.
though looks like it will break all existing notebook saved since it changes the persistent json. is there a way to make them compatible?

zjffdu:
It doesn't change the persistent json. In the current code, we have already save the note json via multiple interpreter result. The side effect of this PR is that zeppelin 0.9 can not read the note json format before 0.6

zjffdu:

PR #3139, opened 2018-08-14
[ZEPPELIN-3712] Add `maxConnLifetime` parameter to JDBC Interpreter

oxygen311:
### What is this PR for?
This parameter is necessary in that case:
1) Zeppelin open connection and keep it in pool;
2) Database close this connection after some time of inactivity;
3) User get an error and necessity to restart interpreter when he runs a paragraph.
 
The default value of this parameter is `-1` which means the connection has an infinite lifetime.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-3712](https://issues.apache.org/jira/browse/ZEPPELIN-3712)

### Questions:

### DIFF
--- a/jdbc/src/main/resources/interpreter-setting.json
+++ b/jdbc/src/main/resources/interpreter-setting.json
@@ -115,6 +115,13 @@
         "defaultValue": false,
         "description": "Enable ZeppelinContext variable interpolation into paragraph text",
         "type": "checkbox"
+      },
+      "zeppelin.jdbc.maxConnLifetime": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.maxConnLifetime",
+        "defaultValue": "-1",
+        "description": "Maximum of connection lifetime in milliseconds. A value of zero or less means the connection has an infinite lifetime.",
+        "type": "number"
       }
     },
     "editor": {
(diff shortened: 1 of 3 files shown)


jdbc/src/main/resources/interpreter-setting.json
```diff
@@ -115,6 +115,13 @@
         "defaultValue": false,
         "description": "Enable ZeppelinContext variable interpolation into paragraph text",
         "type": "checkbox"
+      },
+      "zeppelin.jdbc.maxConnLifetime": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.maxConnLifetime",
+        "defaultValue": -1,
+        "description": "Maximum of connection lifetime in milliseconds. A value of zero or less means the connection has an infinite lifetime.",
```
felixcheung: I thought other parameter/config is documented. could you check if you should update the jdbc.md file

jdbc/src/main/resources/interpreter-setting.json
```diff
@@ -115,6 +115,13 @@
         "defaultValue": false,
         "description": "Enable ZeppelinContext variable interpolation into paragraph text",
         "type": "checkbox"
+      },
+      "zeppelin.jdbc.maxConnLifetime": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.maxConnLifetime",
+        "defaultValue": -1,
```
mebelousov: Please add double quotes near `-1`.
Otherwise it will be converted in "-1.0" and will cause NumberFormatException.

oxygen311: Fixed

jdbc/src/main/resources/interpreter-setting.json
oxygen311: Updated

felixcheung:
LGTM

PR #3146, opened 2018-08-16
[ZEPPELIN-3724] fix `run all paragraph`

Savalek:
### Description
This PR fix 'run all paragraph'.
Also now the execution of all paragraphs does not stop if among them there is a paragraph with disabled run.

### Info
Type: `Bug Fix`
Jira issue: [`ZEPPELIN-3724`](https://issues.apache.org/jira/browse/ZEPPELIN-3724)

### Questions:


Deegue:
Maybe sometimes just want to stop the specific one while using 'run all paragraph' ,  I don't know if it would be better making a new button named 'stop all paragraph' ?

Savalek:
Perhaps soon such an opportunity will be added. But already in the framework of another task.

zjffdu:
Thanks @Savalek it will be fixed in ZEPPELIN-3702

PR #3119, opened 2018-08-09
ZEPPELIN-3681. Introduce NotebookService

zjffdu:
### What is this PR for?
This the first phase of refactoring work of ZEPPELIN-3288. The background of ZEPPELIN-3288 is that zeppelin provides 2 kinds of api (rest api and websocket), both of them invoke the same backend logic.  This PR put all the notebook related operation into class NotebookService which called by both NotebookServer & NotebookRestApi.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3288

### How should this be tested?
* CI pass

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #3148, opened 2018-08-16
[ZEPPELIN-3715] Fix text & titile passing in NotebookRestApi.runParagraph

egorklimov:
### What is this PR for?

All paragraphs processing as blank, because text string is empty.
```
@POST
@Path("job/{noteId}/{paragraphId}")
@ZeppelinApi
public Response runParagraph(@PathParam("noteId") String noteId,
@PathParam("paragraphId") String paragraphId, String message)
throws IOException, IllegalArgumentException {

...

notebookService.runParagraph(
    noteId, paragraphId, "", "", params,
    new HashMap<String, Object>(), false, getServiceContext(), new RestServiceCallback<>());
return new JsonResponse<>(Status.OK).build();
}
```

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* [issue](https://issues.apache.org/jira/browse/ZEPPELIN-3715)

* [CI pass](https://travis-ci.org/TinkoffCreditSystems/zeppelin/builds/416706975)
* Test updated

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -751,14 +751,20 @@ public class NotebookRestApi extends AbstractRestApi {
       throws IOException, IllegalArgumentException {
     LOG.info("run paragraph job asynchronously {} {} {}", noteId, paragraphId, message);
 
+    Note note = notebook.getNote(noteId);
+    checkIfNoteIsNotNull(note);
+    Paragraph paragraph = note.getParagraph(paragraphId);
+    checkIfParagraphIsNotNull(paragraph);
+
     Map<String, Object> params = new HashMap<>();
     if (!StringUtils.isEmpty(message)) {
       RunParagraphWithParametersRequest request =
           RunParagraphWithParametersRequest.fromJson(message);
       params = request.getParams();
     }
-    notebookService.runParagraph(noteId, paragraphId, "", "", params,
-        new HashMap<>(), false, false, getServiceContext(), new RestServiceCallback<>());
+    notebookService.runParagraph(noteId, paragraphId, paragraph.getTitle(),
+            paragraph.getText(), params, new HashMap<>(),
+            false, false, getServiceContext(), new RestServiceCallback<>());
     return new JsonResponse<>(Status.OK).build();
   }
 


zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -770,14 +770,21 @@ public Response runParagraph(@PathParam("noteId") String noteId,
       throws IOException, IllegalArgumentException {
     LOG.info("run paragraph job asynchronously {} {} {}", noteId, paragraphId, message);
 
+    Note note = notebook.getNote(noteId);
+    checkIfNoteIsNotNull(note);
+    checkIfUserCanRun(noteId, "Insufficient privileges you cannot run job for this note");
```
felixcheung: I'm not sure I fully understand `you cannot run job for this note` - could we clarify?

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
```diff
@@ -107,14 +108,24 @@ public void testRunParagraphJob() throws IOException {
...
-    p.setText("test");
+    p.setText("%spark.pyspark\nprint(\"test\")");
     post = httpPost("/notebook/job/" + note1.getId() + "/" + p.getId(), "");
     assertThat(post, isAllowed());
     resp = gson.fromJson(post.getResponseBodyAsString(),
             new TypeToken<Map<String, Object>>() {}.getType());
     assertEquals(resp.get("status"), "OK");
     post.releaseConnection();
-    assertNotEquals(p.getStatus(), Job.Status.READY);
+    long startTimeInMins = (System.currentTimeMillis() / (1000 * 60)) % 60;
```
jongyoul: You’d better move this logic under the test of `notebookservice`

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
egorklimov: I brought back a log from [[ZEPPELIN-2848]](https://github.com/apache/zeppelin/pull/2526) which was deleted in [[ZEPPELIN-3681]](https://github.com/apache/zeppelin/pull/3119/files#diff-dff367ac423a9016688dc86ee27e4929R773), and didn't notice that user privileges will be checked in `NotebookService`, so It is enough to check for null value.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
egorklimov: `NotebookService.runParagraph` is already tested in" NotebookServiceTest.testParagraphOperations ", so I decided to test exactly the REST API method, because even if ` NotebookService.runParagraph` works, the REST API method could fail because of wrong argument preprocessing for `NotebookService.runParagraph`.

egorklimov: @jongyoul Do you agree with me?

jongyoul: Sorry for the late response. BTW, if this is already tested in `NotebookServiceTest`, it would be ok to check all parameters were passed exactly with mock object. WDYT? Personally, I like this kind of conservative code but at the same time, I should consider the flakiness of all tests. This assumes one minute but it could toke more time with some reason which is not related to actual failure.

egorklimov:
CI pass: https://travis-ci.org/TinkoffCreditSystems/zeppelin/builds/422448244

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
egorklimov: @jongyoul I agree with you, so I reverted this changes, maybe I'll refactor them later using PowerMock.

egorklimov:
@zjffdu Seems that this PR can be merged, look, please

PR #3152, opened 2018-08-20
[ZEPPELIN-3736] Helium page's UI to show loading screen

prabhjyotsingh:
### What is this PR for?
At times Helium page's API takes time to return the result, the idea of this JIRA is to show a loading screen while it takes time.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* [ZEPPELIN-3736](https://issues.apache.org/jira/browse/ZEPPELIN-3736)

* Refer screen shot

Before:
before

After:
after

### Questions:

### DIFF
--- a/zeppelin-web/src/app/helium/helium.controller.js
+++ b/zeppelin-web/src/app/helium/helium.controller.js
@@ -15,7 +15,7 @@
 import {HeliumType} from './helium-type';
 
 export default function HeliumCtrl($scope, $rootScope, $sce,
-                                   baseUrlSrv, ngToast, heliumService) {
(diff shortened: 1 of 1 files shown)


zjffdu:
Any reason why helium response slowly ?

prabhjyotsingh:
Possibly because `/api/helium/package` API checks for both local and remote (by default configure to https://s3.amazonaws.com/helium-package/helium.json) helium.json.

prabhjyotsingh:
Update:
Option 2 that can be used alternatively over here:
option2

zjffdu:
Then it seems a bug to me.

PR #3155, opened 2018-08-21
[ZEPPELIN-3741] Do not clear "Authorization" header if Z-server is running behind proxy

prabhjyotsingh:
### What is this PR for?
There can be a case where Zeppelin-Sever is running as Form-Based-Authentication, however, it can be running behind a proxy which may be requiring Authorization header.
The idea of this PR is to not clear that header when it behind a proxy and control it with config.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Add documentaion

### What is the Jira issue?
* [ZEPPELIN-3741](https://issues.apache.org/jira/browse/ZEPPELIN-3741)

* Configure Nginx to run with `auth_basic` option
* Start Zeppelin server behind a proxy server like Nginx 
* Make sure that `shiro.ini` is configured to run with `/** = authc`
* In `zeppelin-site.xml` configure `zeppelin.server.authorization.header.clear` as `false`
Now on logout from Zeppelin-Server should not clear *Authorization* header of Nginx

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -571,6 +571,10 @@ public class ZeppelinConfiguration extends XMLConfiguration {
     return getInt(ConfVars.ZEPPELIN_SERVER_JETTY_REQUEST_HEADER_SIZE);
   }
 
+  public Boolean isClearAuthorizationHeader() {
+    return getBoolean(ConfVars.ZEPPELIN_SERVER_CLEAR_AUTHORIZATION_HEADER);
+  }
+
 
   public String getXFrameOptions() {
     return getString(ConfVars.ZEPPELIN_SERVER_XFRAME_OPTIONS);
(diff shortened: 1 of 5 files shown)


prabhjyotsingh:
@zjffdu @felixcheung @r-kamath can you help review this.

I've tested this on latest versions of Chrome (Mac), Firefox (Mac), Edge (Windows 10).

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -759,6 +763,7 @@ public String getZeppelinSearchTempPath() {
     ZEPPELIN_SERVER_XFRAME_OPTIONS("zeppelin.server.xframe.options", "SAMEORIGIN"),
     ZEPPELIN_SERVER_JETTY_NAME("zeppelin.server.jetty.name", null),
     ZEPPELIN_SERVER_JETTY_REQUEST_HEADER_SIZE("zeppelin.server.jetty.request.header.size", 8192),
+    ZEPPELIN_SERVER_CLEAR_AUTHORIZATION_HEADER("zeppelin.server.clear.authorization.header", true),
```
zjffdu: I think it is better to be named as `zeppelin.server.authorization.header.clear`

prabhjyotsingh:
Will merge this to master if no more discussion.

PR #3156, opened 2018-08-21
[ZEPPELIN-3610] Cluster Raft module design

liuxunorg:
### What is this PR for?

By using the Raft protocol, multiple Zeppelin-Server groups are built into a Zeppelin cluster, the cluster State Machine is maintained through the Raft protocol, and the services in the cluster are agreed upon. The Zeppelin-Server and Zeppelin-Interperter services and processes are stored in the Cluster MetaData. Metadata information;

### What type of PR is it?
[Feature]

### Todos
* [x] add raft algorithm atomix jar
* [x] add cluster state machine
* [x] add state machine query command
* [x] add state machine delete command
* [x] add state machine put command
* [x] Isolate the netty JAR package introduced by atomix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3610

[CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/418742522)

### Questions:


liuxunorg:
@jongyoul @zjffdu ,I have followed the advice you put forward last time https://github.com/apache/zeppelin/pull/3074 . Change COPYCAT to ATOMIX library, please review my code, thank you!

bin/common.sh
```diff
@@ -90,6 +90,62 @@ function addJarInDirForIntp() {
...
+    netty-buffer-4.1.11.Final.jar
+    netty-codec-4.1.11.Final.jar
+    netty-common-4.1.11.Final.jar
+    netty-handler-4.1.11.Final.jar
+    netty-resolver-4.1.11.Final.jar
+    netty-transport-4.1.11.Final.jar
+    netty-transport-native-epoll-4.1.11.Final-linux-x86_64.jar
+    netty-transport-native-unix-common-4.1.11.Final.jar
+    objenesis-2.5.1.jar
+  )
```
zjffdu: Specify excluded jar doesn't seems a good approach for me, it makes maintenance difficult.  It would be very easy to break if we update dependency in pom file but forget to update it here.

zjffdu: Specify excluded jar doesn't seems a good approach for me, it makes maintenance difficult.  It would be very easy to break if we update dependency in pom file but forget to update it here.

zjffdu:
Thanks @liuxunorg , but the shade part does not seem correct to me. I make a simple PR to implement that. https://github.com/zjffdu/zeppelin/commit/a82dc24adace704979b866f36459df61a70d0632

I think we need to introduce another module zeppelin-interpreter-api which shade all dependencies of zeppelin-interpreter, and all the interpreter modules depends on this module instead of zeppelin-interpreter so that we can avoid the potential conflict of some interpreter module and zeppelin-interpreter

liuxunorg:
Thanks @zjffdu , This zeppelin-interpreter-api module is very much needed to solve the problem of package conflicts. I will wait for you to submit to the master. I am re-editing my code.

PR #3157, opened 2018-08-21
[ZEPPELIN-3654] - New Hazelcast Jet interpreter

selvinsource:
### What is this PR for?
**New Hazelcast Jet interpreter**

Zeppelin has interpreters for different data processing systems like Flink, Spark, Kylin, Ignite, Geode, Beam, etc.

Hazelcast Jet is a general purpose distributed data processing engine, built on top of Hazelcast for stream/batch processing, comparable if not better in terms of performance to the engine supported by Zeppelin therefore a perfect candidate for a Zeppelin interpreter.

Part of the interpreter is to have a set of utility methods that print out Hazelcast data structures (IMap and ICache) and leverage Zeppelin's built in visualization (%table).

What's more, a nice addition is to have the Hazelcast Jet DAG of the pipeline displayed as a network graph using %network display system.

### What type of PR is it?
* Feature 

### Todos
* Utility method to display Hazelcast Jet DAG  using %network
* Review LICENCE file
* Any feedback from reviewers

### What is the Jira issue?
* [ZEPPELIN-3654]

Manually
* Start the Zeppelin server
* Create a new note with the hazelcastjet interpreter binding
* Write an Hazelcast Jet job as per documentation (docs/interpreter/hazelcastjet.md)
Unit tests
* Run unit tests (HazelcastJetInterpreterTest.java)

### Questions:
Yes, added Apache License for Hazelcast Jet.
No.
Yes, it has been added to the PR, see docs/interpreter/hazelcastjet.md.


selvinsource:
@felixcheung @zjffdu 
I have just opened the PR for the Hazelcast Jet interpreter as discussed before.

Regarding the LICENSE file, I have a question. Do I need to put the license for all the jars used by Hazelcast Jet (https://github.com/hazelcast/hazelcast-jet/blob/master/pom.xml)? 
So far I only added the Apache License for Hazelcast Jet in  zeppelin-distribution/src/bin_license/LICENSE.

I will be working on the utility method to display Hazelcast Jet DAG using %network to complete ZEPPELIN-3654.

zjffdu:
@selvinsource I don't see any difference between Hazelcast Jet interpreter & java interpreter except extra dependency on hazelcast-jet. So why not use Hazelcast Jet just by adding hazelcast-jet as dependency of java interpreter ?

selvinsource:
@zjffdu this PR is a start. As I said I still need to add Utils methods specific to Jet, in particular the Utils that convert the Jet DAG pipeline to a network graph using Zeppelin display system. If we don't have an Hazelcast Jet interpreter where would we put that code? It doesn't belong to the Java interpreter.

Also, when the Java interpreter will be moved to JShell API (for which I created a JIRA issue), there is a lot more that can be added to the Hazelcast Jet interpreter like configuring the JetInstance (being a client or a node of the cluster) in the interpreter properties so the user don't need to do that for every note or paragraph.

You could say the same for the Beam interpreter, right now it extends the Java interpreter and simply add some dependencies, however I still see value in that so that the end user can just code beam without worrying about the so many dependencies required.

Let me know what you think before I proceed in working on the Jet specific Utils methods.

zjffdu:
Thanks @selvinsource , it make sense to me.

zjffdu:
Thanks @selvinsource Could you fix the conflict ?

selvinsource:
@zjffdu I fixed the conflict and also updated the pom.xml for hazelcastjet due to ZEPPELIN-3774.

zjffdu:
Thanks @selvinsource will merge it soon

PR #3158, opened 2018-08-22
[ZEPPELIN-3740] Adopt `google-java-format` and `fmt-maven-plugin`

jongyoul:
### What is this PR for?
Avoiding outdated checkstyle error.

To check format
`mvn validate`

To fix non-complying files
`mvn fmt:format`

### What type of PR is it?
[Improvement | Refactoring]

### Todos
* [x] - Adopt `fmt-maven-plugin`

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3740

CI passs

### Questions:
* Does this needs documentation? Yes for contributors


jongyoul:
https://travis-ci.org/jongyoul/zeppelin/builds/419215810

4th profiles is unstable now for the current master.

Will merge it if there's no more discussion

felixcheung:
what's wrong with `maven-checkstyle-plugin`?

jongyoul:
@felixcheung IntelliJ's lambda format was not matched in our current checkstyle. I've tried to fix it by changing checkstyle, but I felt depressed because we should do every time whenever this inconsistency occurs. So I thought it would be better to change IDE's style and maven formatter to the same tool. That's the basic reason why I'd like to change style-related plugins

felixcheung:
I see. might be good to get some consensus first - we seem to be doing quite a bit of style changes in the last few months, it would make maintenance or backporting harder, for example.

jongyoul:
Agreed to maintain harder. This would be one of criteria. Actually, this formatter doesn't have to fix format manually, just run `mvn fmt:format` before rebasing it. I also started a discussion in dev@

felixcheung:
ok

jongyoul:
Will merge if there’s no more discussion

PR #3162, opened 2018-08-24
ZEPPELIN-3745. Disable CRC check of hadoop FileSystem

zjffdu:
### What is this PR for?
Trivial PR to disable CRC check of hadoop FileSystem, otherwise we may get error if we manually change the file content of interpreter-setting.json

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3745

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
@@ -62,6 +62,8 @@ public class FileSystemStorage {
     this.hadoopConf = new Configuration();
     // disable checksum for local file system. because interpreter.json may be updated by
     // non-hadoop filesystem api
+    // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks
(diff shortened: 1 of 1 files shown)


bikassaha:
+1 lgtm

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -62,6 +62,8 @@ public FileSystemStorage(ZeppelinConfiguration zConf, String path) throws IOExce
     this.hadoopConf = new Configuration();
     // disable checksum for local file system. because interpreter.json may be updated by
     // non-hadoop filesystem api
+    // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks
+    this.hadoopConf.setBoolean("fs.file.impl.disable.cache", true);
```
felixcheung: shouldn't this be configurable instead of hard coded?

PR #3163, opened 2018-08-24
ZEPPELIN-2619. Save note in ${notename_noteid}.zpln instead of ${noteid}/note.json

zjffdu:
### What is this PR for?

This PR has serverla main changes:
1. `NoteManager` is the most importance class that would build the filesystem tree structure i memory and all note related operation is handled by this class.
2. `NotebookRepo` add several new methods for folder operations, like moving folder, removing folder.
3. The semantics of `name` of `Note` is changed. It is changed to the note name instead of the full note path, there's one transiant field `path` which represent the path of note. 

Design doc
https://docs.google.com/document/d/126egAQmhQOL4ynxJ3AQJQRBBLdW8TATYcGkDL1DNZoE/edit?usp=sharing

### What type of PR is it?
[Improvement |  Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://github.com/zjffdu/zeppelin/tree/ZEPPELIN-2619

* Ci pass

### Questions:


jongyoul:
I have a very simple request about a format of `.zpln`. Could you please add version number of note format? e.g.  `"version": "0"`. It would help to parse `.zpln` in the future

mebelousov:
@zjffdu 
I had a little test.
- Create note with name "Folder/Name 1"
- It was saved on disc correctly.
- Try to edit note name: folder path is absent.

zjffdu:
@jongyoul You can check the PR description for more details. Let me know what else you need to know. 

@mebelousov The issue is fixed.

zeppelin-web/src/components/websocket/websocket-message.service.js
```diff
@@ -93,8 +93,12 @@ function WebsocketMessageService($rootScope, websocketEvents) {
       websocketEvents.sendNewEvent({op: 'NOTE_RENAME', data: {id: noteId, name: noteName}});
     },
 
-    renameFolder: function(folderId, folderName) {
-      websocketEvents.sendNewEvent({op: 'FOLDER_RENAME', data: {id: folderId, name: folderName}});
+    renameNote: function(noteId, noteName, relative) {
+      websocketEvents.sendNewEvent({op: 'NOTE_RENAME', data: {id: noteId, name: noteName, relative: relative}});
+    },
```
mebelousov: Duplicate key 'renameNote'

zjffdu:
@jongyoul @felixcheung I will merge it if no more comments, so that I can continue the follow up improvement.

jongyoul:
@zjffdu Basically, I agree with this idea. BTW, do you have any plan to migrate from old format to new format?

mebelousov:
@zjffdu 
- note can't be cloned
- need to deny folder name as ".." (double dot), because of after restart note will be in another folder
- note with only one empty paragraph doesn't have any paragraph after Zeppelin reload.

zjffdu:
Thanks @mebelousov for verifying this PR, I have addressed the issue.

mebelousov:
@zjffdu 
- Need change the behavior for button "Move note to Trash" for notes in Trash:
- * Need show another tooltip
- * And delete note, but not add another "~Trash" to note path
- Please add check for doubles of note ID. It's good to add warning to log about second note with the same ID.

I will continue testing. Please wait my approve :smiley:

zjffdu:
Thanks @mebelousov for testing.
* Need show another tooltip
Sorry I don't get it, what another tooltip do you mean ? 
* And delete note, but not add another "~Trash" to note path
Do you mean to delete note permanently instead of moving to Trash ? 

* noteId is unique, if there's duplicates, then it should fail to create new note.

mebelousov:
@zjffdu 
* Now usual note has tooltip "Move this note to trash" on trash button. Trashed note has tooltip "Remove this note permanently".
* Trashed note must be deleted on "Trash" button click.
* I know that noteId is unique. But what if somebody create "zpln" file in "notebook" folder and duplicate ID?

mebelousov:
@zjffdu 
- In note tree plus sing near folder must create note in this folder not at root.
- It's seems double dots as folder name is still not fixed.
- Can we avoid CR/LF chars as names? Notes are created, but have "?" in file system name.
- Need to limit the name length for 255 symbols. For long names user get (tested on Ubuntu 16.04) 
`Failed to create note.`
`IOException: Fail to create Note`

```python
import requests
import json

zeppelin_user, zeppelin_password = 'user1' , ''
zeppelin_host, zeppelin_port = 'localhost', 8080

class Izpepelin:
    def __init__(self):
        self.session = requests.Session()
        self.session.auth = (zeppelin_user, zeppelin_password)
        self.base_url = 'http://' + zeppelin_host + ':' + str(zeppelin_port)
        
    def create_note(self, note_name):
        body = {"name": note_name}
        return self.session.post(self.base_url + '/api/notebook/', json = body).json()


super_names = ['Folder1/../double dot' '''Folder1/New 
line''', 'Folder1/1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111']

for name in super_names:
    zepp.create_note(name)
```

zjffdu:
@mebelousov The issues you mentioned are resolved except the tooltip. I think the tooltip is correct, this is what zeppelin right now, first move note to trash, then user can remove it from trash permanently.

mebelousov:
@zjffdu, I'm back

1. I get IOException if folder name equals to "..". The same need for name "." (single dot).
1. Try to create 2 notes with the same path and name, then press "Move folder to Trash". On disk folder will be moved to trash, but note tree will contain bronen note link.
1. Still Trash button just renames note if it already in Trash. See first GIF.
1. In branch-0.8 only owners can rename note or change path, in your branch writers can change path and can trash note. (GIF2)
1. Note search (note-name-query form on main page) is broken. (GIF2)
1. Button "Clone this note" clones note to root folder. I expect cloning to the same folder. (GIF2)

trash_button

renaming_cloning

zjffdu:
@mebelousov Do you see popup windows saying `Fail to create note, cause: Note name can not contain '..'` ? This is by design. 

Regarding 2, 3, this is a known issue, this is due to whether we should allow duplicate note name under folder. I plan to discuss it in community, will fix it in follow up PR.

Regarding 6,7, I plan to fix it in follow up PR as they are not trivial to fix.

mebelousov:
@zjffdu 
No, I didn't see the popup with error explanation for user.

I have tested a lot with Russian symbols. Seems that Zeppelin works fine.

Thank you for this improvement. PR could be merged.

zjffdu:
Thanks all for reviewing and testing, will merge it soon

jongyoul:
Wow!! Thank you guys for all your efforts @zjffdu @mebelousov

maziyarpanahi:
Hi guys,
I really like the idea of other tools recognizing `.zpln` like Jupyter and displaying it. However, at the moment the only way to export a note is to use the export button which produces `.json` which is a universal format that can be imported in older versions as well.
Should we add another option if someone wants to export it in `.zpln` format?

PR #3168, opened 2018-08-31
[ZEPPELIN-3753] Fix indent with TAB

oxygen311:
### What is this PR for?
Now when you select multiline text and press TAB, text replaces with "\t" char.
With this PR text just shift right if TAB have been pressed.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
 [ZEPPELIN-3753](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3753)

### Questions:


felixcheung:
merging if no more comment

PR #3170, opened 2018-09-06
ZEPPELIN-3689. Shade all dependencies of zeppelin-interpreter

zjffdu:
### What is this PR for?
This PR introduce new module zeppelin-interpreter-api which will shade all dependencies of zeppelin-interpreter, this is just to avoid the conflict with the custom interpreter implementation which may use jars that conflicts with dependency of zeppelin-interpreter.

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3689

* CI pass

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ addons:
 env:
   global:
     # Interpreters does not required by zeppelin-server integration tests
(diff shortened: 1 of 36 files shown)


liuxunorg:
Thanks @zjffdu , This design perfectly solves the problem of dependency package conflicts.
If I need to add a dependency package to zeppelin-interpreter, Is the dependency package added in the zeppelin-interpreter/pom.xml, and then excluded in the zeppelin-interpreter-api/pom.xml, right?

zjffdu:
That's correct @liuxunorg

PR #3173, opened 2018-09-09
[ZEPPELIN-3772] Allow definition of http proxy for default interpreter repository

baunz:
### What is this PR for?
When running zeppelin behind a firewall the default maven central _without_ proxy is always checked first, which makes dependency downloading unusable - for every pom, jar the timeout of the request against maven central adds up to the total time of downloading the dependencies, even if a repository with proxy is added (made possible in ZEPPELIN-1376)

This commit adds additional properties / environment variables that set the proxy on the default repository  

### What type of PR is it?
Improvement
### Todos
* [ ] - Is the modification of org.apache.zeppelin.conf.ZeppelinConfiguration really necessary? The properties are never referenced anywhere and not used in Booter 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3772

* In a proxy environment

### Questions:
  no
  no
  Currently the default repository is not documented somewhere. Maybe this should be changed? Besides this, the environment template and site file have comments

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -494,6 +494,18 @@ public class ZeppelinConfiguration extends XMLConfiguration {
     return getString(ConfVars.ZEPPELIN_INTERPRETER_DEP_MVNREPO);
   }
 
+  public String getInterpreterMvnRepoProxyHost() {
+    return getString(ConfVars.ZEPPELIN_INTERPRETER_DEP_MVNREPO_PROXYHOST);
+  }
+
+  public int getInterpreterMvnRepoProxyPort() {
+    return getInt(ConfVars.ZEPPELIN_INTERPRETER_DEP_MVNREPO_PROXYHOST);
+  }
+
+  public String getInterpreterMvnRepoProxyType(){
+    return getString(ConfVars.ZEPPELIN_INTERPRETER_DEP_MVNREPO_PROXYHOST);
+  }
+
   public String getRelativeDir(ConfVars c) {
     return getRelativeDir(getString(c));
   }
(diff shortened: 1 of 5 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -697,6 +709,12 @@ public String getZeppelinSearchTempPath() {
     ZEPPELIN_INTERPRETER_LOCALREPO("zeppelin.interpreter.localRepo", "local-repo"),
     ZEPPELIN_INTERPRETER_DEP_MVNREPO("zeppelin.interpreter.dep.mvnRepo",
         "http://repo1.maven.org/maven2/"),
+    ZEPPELIN_INTERPRETER_DEP_MVNREPO_PROXYHOST("zeppelin.interpreter.dep.mvnRepo.proxyHost",
+            null),
+    ZEPPELIN_INTERPRETER_DEP_MVNREPO_PROXYPORT("zeppelin.interpreter.dep.mvnRepo.proxyPort",
+            null),
```
felixcheung: should port default to null?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/Booter.java
```diff
@@ -71,16 +73,47 @@ static String resolveLocalRepoPath(String localRepoPath) {
...
   }
 
+  /**
+   * Check for environment variable and java property with precedence of the environment variable
+   * @param environmentVariableName
+   * @param javaPropertyName
+   * @return
+   * The environment variable, if not set the system property, or null
+   */
+  private static String getEnvironmentVariable(
```
felixcheung: perhaps a different name? this isn't just getting from environment

baunz: You are right. There is similar functionality in ZepperlinConfiguration https://github.com/apache/zeppelin/blob/5c94d7550aa569361d5c5569004917ee9bd82c70/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L187 but only availabe on the config. Would something like "getEnviromentVariableOrProperty()" be okay?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
felixcheung: I guess what the mvn port default is...?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/dep/Booter.java
felixcheung: sure

baunz:
@felixcheung I adressed your comments, but to be honest I am not sure we should go forward with this.
We solved our problem by using our internal nexus as a maven proxy and I don't know how big the value of this is to other users. Given this, I would rather close this pull request if you are fine with it to not make the configuration files more complex

baunz:
As suggested this is not really relevant for most users anymore

PR #3174, opened 2018-09-10
[ZEPPELIN-3773] add check permission on write

Savalek:
### What is this PR for?
Sometimes when drawing the result of a paragraph, a call is made commit paragraph. And if the user does not have permission on write paragraph a window appears with a warning.
This PR fix it.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3773](https://issues.apache.org/jira/browse/ZEPPELIN-3773)

### Questions:


Savalek:
It is necessary to refine

PR #3176, opened 2018-09-11
ZEPPELIN-3558. Refactor Scheduler component

zjffdu:
### What is this PR for?

This is a refactoring PR for refactoring scheduler component.  This PR introduce AbstractScheduler which have many common code, FIFOScheduler and ParallelSchduler become much clean after this PR

### What type of PR is it?
[ Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3558

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/AbstractScheduler.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/AbstractScheduler.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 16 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/scheduler/AbstractScheduler.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @Override
+  public void run() {
+    while (!terminate) {
+      Job runningJob = null;
+      try {
+        runningJob = queue.take();
+      } catch (InterruptedException e) {
+        LOGGER.warn("FIFOScheduler is interrupted", e);
```
aka-demik: It seems that the name of the scheduler needs to be corrected after cut+paste.
May be:
```java
LOGGER.warn("{} is interrupted", getClass().getSimpleName(), e);
```

zjffdu: Thanks @aka-demik  I have fixed it

PR #3179, opened 2018-09-12
[ZEPPELIN-3773] - add check permission on write.

Savalek:
### What is this PR for?
Sometimes when drawing the result of a paragraph, a call is made commit paragraph. And if the user does not have permission on write paragraph a window appears with a warning.
This PR fix it.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3773](https://issues.apache.org/jira/browse/ZEPPELIN-3773)

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -1200,6 +1200,31 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
     return true;
   };
 
+  $scope.userHasWritePermission = function() {
+    let owners = $scope.permissions.owners;
+    let writers = $scope.permissions.writers;
+
+    if (owners.length === 0 || writers.length === 0) {
+      return true;
+    }
+
+    let userName = $rootScope.ticket.principal;
+    let userRoles = $rootScope.ticket.roles;
+
+    userRoles = userRoles.substr(1, userRoles.length - 2).replace(/"/g, '').split(',');
+    let userNameAndRoles = [];
+    userNameAndRoles.push(userName);
+    if (userRoles !== null) {
+      userNameAndRoles = userRoles.concat(userNameAndRoles);
+    }
+
+    if (owners.concat(writers).some((name) => userNameAndRoles.includes(name))) {
+      return true;
+    }
+
+    return false;
+  };
+
   $scope.toggleNotePersonalizedMode = function() {
     let personalizedMode = $scope.note.config.personalizedMode;
     if ($scope.isOwner) {
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -759,6 +759,9 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
   };
 
   $scope.sendPatch = function() {
+    if (!$scope.userHasWritePermission()) {
+      return;
+    }
     $scope.originalText = $scope.originalText ? $scope.originalText : '';
     let patch = $scope.diffMatchPatch.patch_make($scope.originalText, $scope.dirtyText).toString();
     $scope.originalText = $scope.dirtyText;
@@ -1126,6 +1129,7 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
         $scope.$on('editorSetting', function(event, data) {
           if (paragraph.id === data.paragraphId) {
             deferred.resolve(data);
+            $scope.editor.setReadOnly(!$scope.userHasWritePermission());
           }
         }
       ), 1000);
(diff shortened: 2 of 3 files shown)


mebelousov:
@Savalek 
I see an infinite number of GET request like http://localhost:8080/api/helium/suggest/2DMKVSPYC/20180807-154514_2063688624 with response "401 Unauthorized" is case if user don't have permission for write. Could you fix?

Savalek:
@mebelousov, Perhaps you have incorrectly configured shiro.ini.

Savalek:
@felixcheung, there is already a check on the server. Due to some errors, a commit of the paragraph went to the server and an error occurred.

mebelousov:
@Savalek 
You're right. I fix my shiro.ini.

But I still see a huge amount of background queries (up to 100 per second) connected with Helium.
I can reproduce the issue in development mode.
Tested on Ubuntu 16.04, Chromium (Version 69.0.3497.92) and Firefox Quantum 62.0
Check please.

deradam:
This seems to be the bug that we are also facing currently. I have a notebook with some users that only have read permissions. These users receive an error message that the do not have the update permission on the notebook and therefore cannot open it.

Any estimate when this will be merged and released?

Savalek:
Need someone else to check this PR. Because I could not reproduce this error.

Savalek:
I'm fixed problem with "infinite number of GET request".
Please review this pr somebody.

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -759,6 +759,9 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
   };
 
   $scope.sendPatch = function() {
+    if (!$scope.userHasWritePermission()) {
```
felixcheung: should this save $scope.userHasWritePermission() and not called it multiple times?
what if the return from `$scope.userHasWritePermission()` changes?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -1200,6 +1200,31 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
     return true;
   };
 
+  $scope.userHasWritePermission = function() {
+    let owners = $scope.permissions.owners;
+    let writers = $scope.permissions.writers;
+
+    if (owners.length === 0 || writers.length === 0) {
```
felixcheung: owners.length shouldn't be empty??

felixcheung: if writers.length == 0 shouldn't this return false?

zeppelin-web/src/app/notebook/notebook.controller.js
```diff
@@ -1200,6 +1200,31 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
...
+      return true;
+    }
+
+    let userName = $rootScope.ticket.principal;
+    let userRoles = $rootScope.ticket.roles;
+
+    userRoles = userRoles.substr(1, userRoles.length - 2).replace(/"/g, '').split(',');
```
felixcheung: do you have an example of what this looks like?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
```diff
@@ -1126,6 +1129,7 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
         $scope.$on('editorSetting', function(event, data) {
           if (paragraph.id === data.paragraphId) {
             deferred.resolve(data);
+            $scope.editor.setReadOnly(!$scope.userHasWritePermission());
```
felixcheung: I worry this would leave it in an inconsistent state - UI showing the changed value, but it isn't saved

zeppelin-web/src/app/notebook/notebook.controller.js
Savalek: If the field is empty, it means that anyone can take this action.
I understand correctly?

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
Savalek: But permissions can change while the notebook is open. Or do you propose to calculate userHasWritePermission() only when the permissions change?

zeppelin-web/src/app/notebook/notebook.controller.js
Savalek: userRoles = ["role1","role2","role3","admin"] - (String)
.substr(1, userRoles.length - 2) ---> "role1","role2","role3","admin"
.replace(/"/g, '') ---> role1,role2,role3,admin
.split(',') ---> ["role1", "role2", "role3", "admin"]

zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
Savalek: $scope.editor.setReadOnly(...) - blocks user input completely.
Or do you mean something else?

Savalek:
@felixcheung, could you review, please?

felixcheung:
this is a bit tricky, could @prabhjyotsingh review?

PR #3182, opened 2018-09-19
[ZEPPELIN-3761] Aliasing is not working with JDBC interperter

prabhjyotsingh:
### What is this PR for?
Using aliasing to rename the column name is not really working. For instance, a SELECT like this renames "aircraft" column in output to "something" in MySQL terminal, but not in Zeppelin:

SELECT aircraft AS something FROM birdstrikes LIMIT 5;

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-3761](https://jira.apache.org/jira/browse/ZEPPELIN-3761)

* Running SQL query with or without alisa should result with the expected column name.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -542,7 +542,11 @@ public class JDBCInterpreter extends KerberosInterpreter {
       if (i > 1) {
         msg.append(TAB);
       }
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
@zjffdu @astroshim   can you help review this.

zjffdu:
LGTM, can we add unit test for this ?

prabhjyotsingh:
Sure, thanks for the review have added it.

prabhjyotsingh:
Will merge this soon.

PR #3189, opened 2018-09-26
[ZEPPELIN-3758]. Convert old note file note.json to new style

zjffdu:
### What is this PR for?
This PR is follow up of ZEPPELIN-261, it is to convert old note file note.json to new style when user upgrading zeppelin before 0.9. 2 properties are introduced.
* `zeppelin.notebook.new_format.convert`, by default it is false. When enabled, zeppelin will first find all the old note file note.json and then convert it into new style.
* `zeppelin.notebook.new_format.delete_old`, by default it is false, when setting true, zeppelin will delete the old note file after converting it into new style.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3758

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -732,6 +732,8 @@ public class ZeppelinConfiguration extends XMLConfiguration {
     ZEPPELIN_NOTEBOOK_STORAGE("zeppelin.notebook.storage",
         "org.apache.zeppelin.notebook.repo.GitNotebookRepo"),
     ZEPPELIN_NOTEBOOK_ONE_WAY_SYNC("zeppelin.notebook.one.way.sync", false),
+    ZEPPELIN_NOTEBOOK_NEW_FORMAT_CONVERT("zeppelin.notebook.new_format.convert", false),
+    ZEPPELIN_NOTEBOOK_NEW_FORMAT_DELETE_OLD("zeppelin.notebook.new_format.delete_old", false),
     // whether by default note is public or private
     ZEPPELIN_NOTEBOOK_PUBLIC("zeppelin.notebook.public", true),
     ZEPPELIN_INTERPRETER_REMOTE_RUNNER("zeppelin.interpreter.remoterunner",
--- a/zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
+++ b/zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
@@ -1,0 +1,92 @@
+package org.apache.zeppelin.notebook.repo;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.zeppelin.conf.ZeppelinConfiguration;
+import org.apache.zeppelin.notebook.FileSystemStorage;
+import org.apache.zeppelin.notebook.Note;
+import org.apache.zeppelin.notebook.NoteInfo;
+import org.apache.zeppelin.notebook.OldNoteInfo;
+import org.apache.zeppelin.user.AuthenticationInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * NotebookRepos for hdfs.
+ *
+ * Assume the notebook directory structure is as following
+ * - notebookdir
+ *              - noteId/note.json
+ *              - noteId/note.json
+ *              - noteId/note.json
+ */
+public class OldFileSystemNotebookRepo implements OldNotebookRepo {
+  private static final Logger LOGGER = LoggerFactory.getLogger(FileSystemNotebookRepo.class);
+
+  private FileSystemStorage fs;
+  private Path notebookDir;
+
+  public OldFileSystemNotebookRepo() {
+
+  }
+
+  public void init(ZeppelinConfiguration zConf) throws IOException {
+    this.fs = new FileSystemStorage(zConf, zConf.getNotebookDir());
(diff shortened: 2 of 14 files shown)


prabhjyotsingh:
Also, since we are doing this; should there be a command line utility as well? which on run from command-line can convert from old to new format.

zjffdu:
Good idea, I can do it in a follow up PR.

zjffdu:
Many classes are old existing code which I just rename them.  e.g. `OldAzureNotebookRepo` is renamed from the old version of `AzureNotebookRepo`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -83,6 +84,27 @@ public void init(ZeppelinConfiguration conf) throws IOException {
...
+
+    // convert old note file (noteId/note.json) to new note file (note_name_note_id.zpln)
+    boolean convertToNew = conf.getBoolean(ConfVars.ZEPPELIN_NOTEBOOK_NEW_FORMAT_CONVERT);
+    boolean deleteOld = conf.getBoolean(ConfVars.ZEPPELIN_NOTEBOOK_NEW_FORMAT_DELETE_OLD);
+    if (convertToNew) {
+      NotebookRepo newNotebookRepo = repos.get(0);
+      OldNotebookRepo oldNotebookRepo =
+          PluginManager.get().loadOldNotebookRepo(newNotebookRepo.getClass().getCanonicalName());
+      oldNotebookRepo.init(conf);
+      List<OldNoteInfo> oldNotesInfo = oldNotebookRepo.list(AuthenticationInfo.ANONYMOUS);
```
felixcheung: would accessing as anonymous here causes some old note to be skipped?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
```diff
@@ -83,6 +84,27 @@ public void init(ZeppelinConfiguration conf) throws IOException {
...
+      LOGGER.info("Convert old note file to new style, note count: " + oldNotesInfo.size());
+      for (OldNoteInfo oldNoteInfo : oldNotesInfo) {
+        Note note = oldNotebookRepo.get(oldNoteInfo.getId(), AuthenticationInfo.ANONYMOUS);
+        note.setPath(note.getName());
+        newNotebookRepo.save(note, AuthenticationInfo.ANONYMOUS);
+        if (deleteOld) {
+          oldNotebookRepo.remove(note.getId(), AuthenticationInfo.ANONYMOUS);
```
felixcheung: log when deleting?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
zjffdu: No, anonymous can access all notes.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -732,6 +732,8 @@ public String getZeppelinSearchTempPath() {
     ZEPPELIN_NOTEBOOK_STORAGE("zeppelin.notebook.storage",
         "org.apache.zeppelin.notebook.repo.GitNotebookRepo"),
     ZEPPELIN_NOTEBOOK_ONE_WAY_SYNC("zeppelin.notebook.one.way.sync", false),
+    ZEPPELIN_NOTEBOOK_NEW_FORMAT_CONVERT("zeppelin.notebook.new_format.convert", false),
+    ZEPPELIN_NOTEBOOK_NEW_FORMAT_DELETE_OLD("zeppelin.notebook.new_format.delete_old", false),
```
Tagar: If zeppelin.notebook.new_format.convert=true and zeppelin.notebook.new_format.delete_old=false, would Zeppelin try to convert same old notes many times, after each restart?

zjffdu: That's right, user need to set `zeppelin.notebook.new_format.convert` to false when converting is done. so that next time zeppelin won't do the conversion again.

zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
```diff
@@ -0,0 +1,92 @@
...
+  @Override
+  public Note get(final String noteId, AuthenticationInfo subject) throws IOException {
+    String content = this.fs.readFile(
+        new Path(notebookDir.toString() + "/" + noteId + "/note.json"));
+    return Note.fromJson(content);
+  }
+
+  @Override
+  public void save(final Note note, AuthenticationInfo subject) throws IOException {
+    this.fs.writeFile(note.toJson(),
```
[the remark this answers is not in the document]
zjffdu: It will write to a temp file first. See here https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java#L178

maziyarpanahi:
Does this work out of the box if the notes are in HDFS? I am facing a similar situation which I am guessing is related to this issue:
https://jira.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3987

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
maziyarpanahi: If the user forgot to change it back, every restart would convert an old version of an old format (json) and overwrites it on a new version of a newer version note. I would go with the script, but it's not easy when you have 100 notes in HDFS :(

varadskarmarkar:
Hi, this property should be set either in zeppelin-site.xml or in zeppelin-env.sh, correct? I have tried both and neither has converted the notebooks.

PR #3192, opened 2018-09-27
ZEPPELIN-3768. run all paragraphs sequential mode when using ZeppelinContext api

zjffdu:
### What is this PR for?

This PR is to fix the bugs of ZeppelinContext.run api. It should run paragraph in blocking mode rather than non-blocking mode. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3768

* CI pass

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ addons:
 env:
   global:
     # Interpreters does not required by zeppelin-server integration tests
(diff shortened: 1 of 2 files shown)


zjffdu:

PR #3198, opened 2018-09-30
ZEPPELIN-3801. Code refactoring of org.apache.zeppelin.interpreter.remote

zjffdu:
### What is this PR for?
This PR is a refactoring PR which to refactor the code of package `org.apache.zeppelin.interpreter.remote`.
* Remove `InterpreterClient`, and replace it with `RemoteInterpreterProcess`.
* Use lambda function as possible as we can.

### What type of PR is it?
[ Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3801

* CI pass

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteRemoteInterpreterManagedProcess.java
```diff
@@ -39,10 +38,10 @@
 /**
  * This class manages start / stop of remote interpreter process
  */
-public class RemoteInterpreterManagedProcess extends RemoteInterpreterProcess
```
liuxunorg: `RemoteRemoteInterpreterManagedProcess`, What is the meaning of the name of this class? Could it be a little confusing？

PR #3201, opened 2018-10-03
[ZEPPELIN-3805] Don't distribution shade jar to lib/interpreter for zeppelin-interpreter

liuxunorg:
The zeppelin-interpreter-api module shades the conflicting package, Need to remove the package that issued these shades in zeppelin-distribution.

[Bug Fix]

* [x] Exclude shade JAR packages from lib/interpreter 
* [x] Add zeppelin-interpreter-api JAR package to lib/interpreter 

* https://issues.apache.org/jira/browse/ZEPPELIN-3805

[CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/436144528)

### DIFF
--- a/zeppelin-distribution/src/assemble/distribution.xml
+++ b/zeppelin-distribution/src/assemble/distribution.xml
@@ -97,6 +97,37 @@
     <fileSet>
       <outputDirectory>/lib/interpreter</outputDirectory>
       <directory>../zeppelin-interpreter/target/lib</directory>
+      <excludes>
+        <exclude>asm-5.0.4.jar</exclude>
+        <exclude>atomix-cluster-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-primary-backup-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-utils-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-primitive-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-raft-3.0.0-rc4.jar</exclude>
+        <exclude>atomix-storage-3.0.0-rc4.jar</exclude>
+        <exclude>config-1.3.2.jar</exclude>
+        <exclude>fast-classpath-scanner-2.21.jar</exclude>
+        <exclude>guava-22.0.jar</exclude>
+        <exclude>kryo-4.0.2.jar</exclude>
+        <exclude>minlog-1.3.0.jar</exclude>
+        <exclude>netty-buffer-4.1.27.Final.jar</exclude>
+        <exclude>netty-common-4.1.27.Final.jar</exclude>
+        <exclude>netty-codec-4.1.27.Final.jar</exclude>
+        <exclude>netty-handler-4.1.27.Final.jar</exclude>
+        <exclude>netty-resolver-4.1.27.Final.jar</exclude>
+        <exclude>netty-transport-4.1.27.Final.jar</exclude>
+        <exclude>netty-transport-native-unix-common-4.1.27.Final.jar</exclude>
+        <exclude>netty-transport-native-epoll-4.1.27.Final-linux-x86_64.jar</exclude>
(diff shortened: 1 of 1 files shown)


zeppelin-distribution/src/assemble/distribution.xml
```diff
@@ -97,6 +97,37 @@
     <fileSet>
       <outputDirectory>/lib/interpreter</outputDirectory>
       <directory>../zeppelin-interpreter/target/lib</directory>
+      <excludes>
```
zjffdu: I think we can just remove this line. https://github.com/apache/zeppelin/blob/master/bin/interpreter.sh#L88

shaded jar (zeppelin-interpreter-api) is only used for interpreter process. Zeppelin server process still use the unshaded jar (zeppelin-interpreter)

liuxunorg: 1. The "${ZEPPELIN_HOME}/zeppelin-interpreter-api/target" path only exists when running travis CI, so bin/interpreter.sh allows the CI to pass the test.
2. There is no "${ZEPPELIN_HOME}/zeppelin-interpreter-api/target" path in zeppelin-0.X.tar.gz released by zeppelin-distribution.
3. If you do not modify zeppelin-distribution/src/assemble/distribution.xml, there will be no zeppelin-interpreter-api-XXX.jar in zeppelin-0.X.tar.gz

zeppelin-distribution/src/assemble/distribution.xml
```diff
@@ -97,6 +97,37 @@
...
+        <exclude>netty-transport-4.1.27.Final.jar</exclude>
+        <exclude>netty-transport-native-unix-common-4.1.27.Final.jar</exclude>
+        <exclude>netty-transport-native-epoll-4.1.27.Final-linux-x86_64.jar</exclude>
+        <exclude>objenesis-2.5.1.jar</exclude>
+      </excludes>
+    </fileSet>
+    <fileSet>
+      <outputDirectory>/lib/interpreter</outputDirectory>
+      <directory>../zeppelin-interpreter-api/target</directory>
+      <includes>
```
liuxunorg: If you do not modify zeppelin-distribution/src/assemble/distribution.xml, there will be no zeppelin-interpreter-api-XXX.jar in zeppelin-0.X.tar.gz

zeppelin-distribution/src/assemble/distribution.xml
liuxunorg: OK.

liuxunorg:
@zjffdu , Because the interpreter process requires the zeppelin-interpreter-api.jar file,
So in the `zeppelin-distribution/src/assemble/distribution.xml` only include 
```
<fileSet>
      <outputDirectory>/lib/interpreter</outputDirectory>
      <directory>../zeppelin-interpreter-api/target</directory>
        <include>zeppelin-interpreter-api-*.jar</include>
       </includes>
</fileSet>
```
include, delete the excludes part.
```
<excludes>
        <exclude>asm-5.0.4.jar</exclude>
         <exclude>atomix-cluster-3.0.0-rc4.jar</exclude>
         ......
</excludes>
```
Also delete the `https://github.com/apache/zeppelin/blob/master/bin/interpreter.sh#L88` line, publish `zeppelin-interpreter-api.jar` to the lib/interpreter/ directory, and pass Travis CI .

However, the LIB file directory in zeppelin-xxx.tar.gz released by zeppelin-distribution is `https://issues.apache.org/jira/secure/attachment/12942346/zeppelin-lib-tree.txt`.
Both shaded jars (zeppelin-interpreter-api) also include unshaded jars (zeppelin-interpreter).

Is `zeppelin-interpreter-api.jar` not published to the `lib/interpreter` directory?

zjffdu:
It is not necessary to publish zeppelin-interpreter-api.jar to lib/interpreter. zeppelin interpreter api jar is already included to all interpreter jars under lib folder.

liuxunorg:
Tanks @zjffdu , Indeed zeppelin-interpreter-api.jar is already included in every interpreter JAR package.

PR #3221, opened 2018-11-09
ZEPPELIN-3858. Code completion doesn't work for spark interpreter when cursor is in the middle of code

zjffdu:
### What is this PR for?
It is a trivial fix for the code completion of spark interpreter when the cursor is in the middle of code.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3858

* CI pass

### Questions:

### DIFF
--- a/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
+++ b/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
@@ -138,7 +138,7 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
   protected def completion(buf: String,
                            cursor: Int,
                            context: InterpreterContext): java.util.List[InterpreterCompletion] = {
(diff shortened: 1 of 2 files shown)


spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -138,7 +138,7 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
   protected def completion(buf: String,
                            cursor: Int,
                            context: InterpreterContext): java.util.List[InterpreterCompletion] = {
-    val completions = scalaCompleter.complete(buf, cursor).candidates
+    val completions = scalaCompleter.complete(buf.substring(0, cursor), cursor).candidates
```
felixcheung: does buf contain text from the earlier line etc?
eg.
```
foo
sc.ra
```
or
```
foo sc.ra
```

PR #3228, opened 2018-11-16
ZEPPELIN-3865. Paragraph cancel is broken

zjffdu:
### What is this PR for?
Paragraph cancel is broken due to previous code refactoring, This PR fix it in straightforward way and add system test as well.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3865

* CI pass

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
@@ -136,8 +136,18 @@ public class ZeppelinSparkClusterTest extends AbstractTestRestApi {
 
   private void waitForFinish(Paragraph p) {
     while (p.getStatus() != Status.FINISHED
-        && p.getStatus() != Status.ERROR
-        && p.getStatus() != Status.ABORT) {
+            && p.getStatus() != Status.ERROR
+            && p.getStatus() != Status.ABORT) {
+      try {
+        Thread.sleep(100);
+      } catch (InterruptedException e) {
+        LOG.error("Exception in WebDriverManager while getWebDriver ", e);
+      }
+    }
+  }
+
+  private void waitForRunning(Paragraph p) {
+    while (p.getStatus() != Status.RUNNING) {
       try {
         Thread.sleep(100);
       } catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -461,9 +461,10 @@ protected boolean jobAbort() {
     if (interpreter == null) {
       return true;
     }
-    Scheduler scheduler = interpreter.getScheduler();
-    if (scheduler == null) {
-      return true;
+    try {
+      interpreter.cancel(getInterpreterContext(null));
+    } catch (InterpreterException e) {
+      throw new RuntimeException(e);
```
felixcheung: why can't we throw InterpreterException here?

zjffdu: This is due to signature of class `Job`, change the signature of class `Job` will result in many changes in its callers (I will do this kind of refactoring later),  so I just throw RuntimeException here, but the exception would still be caught by ZeppelinSever.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/ZeppelinSparkClusterTest.java
```diff
@@ -176,6 +176,17 @@ public void scalaOutputTest() throws IOException {
     // test code completion
     List<InterpreterCompletion> completions = note.completion(p.getId(), "sc.", 2);
     assertTrue(completions.size() > 0);
+
+    // test cancel
+    p.setText("%spark sc.range(1,10).map(e=>{Thread.sleep(1000); e}).collect()");
+    note.run(p.getId(), false);
+    // wait for 3 seconds so that job is started before cancel
+    Thread.sleep(3000);
```
aka-demik: Can you use busy waiting, event or latch here? In the current implementation, this test lasts for six seconds, even if it is not necessary.
Unfortunately, the tests are already slow = (

zjffdu: Thanks @aka-demik , I have fixed it

PR #3230, opened 2018-11-17
[ZEPPELIN-3847] Duplicate results in notebooks due to WS interruption

Leemoonsoo:
### What is this PR for?
Duplicate results in notebooks after ws reconnect.
The problem occurs via duplicated 'setConnectedStatus' event handler in notebook.controller.js.
The same function exists at https://github.com/Leemoonsoo/zeppelin/blob/dbbf0436490941d117fd4c06b9da5b6cb47697ff/zeppelin-web/src/app/notebook/notebook.controller.js#L99-L104.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3847

Disconnect and reconnect websocket connection to browser, and check if result is duplicated.

### Questions:


zjffdu:
LGTM

PR #3233, opened 2018-11-21
[ZEPPELIN-3875] Groups are not derived when using ActiveDirectory

zlosim:
### What is this PR for?
ActiveDirectoryGroupRealm is using UserPrincipalName for finding user and deriving groups that he belongs to. However this is not mandatory nor always corresponds to the REALM that user belongs to.
This PR changed userPrincipalName to sAMAccountName when deriving groups user belongs to

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-3875](https://issues.apache.org/jira/browse/ZEPPELIN-3875)

* CI pass

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
@@ -291,11 +291,11 @@ public class ActiveDirectoryGroupRealm extends AbstractLdapRealm {
     SearchControls searchCtls = new SearchControls();
     searchCtls.setSearchScope(SearchControls.SUBTREE_SCOPE);
     String userPrincipalName = username;
-    if (this.principalSuffix != null && userPrincipalName.indexOf('@') < 0) {
-      userPrincipalName += principalSuffix;
+    if (this.principalSuffix != null && userPrincipalName.indexOf('@') > 1) {
+      userPrincipalName = userPrincipalName.split("@")[0];
     }
 
-    String searchFilter = "(&(objectClass=*)(userPrincipalName=" + userPrincipalName + "))";
+    String searchFilter = "(&(objectClass=*)(sAMAccountName=" + userPrincipalName + "))";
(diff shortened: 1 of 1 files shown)


zjffdu:
ping @prabhjyotsingh

zeppelin-server/src/main/java/org/apache/zeppelin/realm/ActiveDirectoryGroupRealm.java
```diff
@@ -291,11 +291,11 @@ protected AuthorizationInfo buildAuthorizationInfo(Set<String> roleNames) {
...
     searchCtls.setSearchScope(SearchControls.SUBTREE_SCOPE);
     String userPrincipalName = username;
-    if (this.principalSuffix != null && userPrincipalName.indexOf('@') < 0) {
-      userPrincipalName += principalSuffix;
+    if (this.principalSuffix != null && userPrincipalName.indexOf('@') > 1) {
+      userPrincipalName = userPrincipalName.split("@")[0];
     }
 
-    String searchFilter = "(&(objectClass=*)(userPrincipalName=" + userPrincipalName + "))";
+    String searchFilter = "(&(objectClass=*)(sAMAccountName=" + userPrincipalName + "))";
```
prabhjyotsingh: This `(userPrincipalName=?)`  is used in two places `public List<String> searchForUserName` and `private Set<String> getRoleNamesForUser`.
Also, can you make to property configurable like https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/realm/LdapRealm.java#L107

zlosim: sorry @prabhjyotsingh for the late reply, you were right, the other search was used for searching when setting note permissions and i fixed that one also
also userSearchParameter is configurable via shiro.ini from now but uses default if not set

please let me know i i should update doc somewhare as I was not able to find any resources

zlosim:
Ready to review

prabhjyotsingh:

PR #3234, opened 2018-11-22
ZEPPELIN-3876. Unable to rename note

zjffdu:
### What is this PR for?

This is trivial PR for fixing the issue of unable to rename note.  The root cause is that the notePath may have 2 leading `/` which cause the error.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3876

* CI pass

### Questions:


zjffdu:
Test is added, will merge it no more comments

PR #3235, opened 2018-11-22
[ZEPPELIN-3864] Fix Travis tests

Savalek:
### What is this PR for?
Maven `frontend plugin v1.4` incorrectly processed `npm` messages with error.
This PR update `frontend plugin` to 1.6 from 1.4.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-3864](https://issues.apache.org/jira/browse/ZEPPELIN-3864),
[ZP-70]

### Questions:


felixcheung:
this test failed? https://travis-ci.org/TinkoffCreditSystems/zeppelin/jobs/458386760

Savalek:
@felixcheung, yes.

Savalek:
Any progress here?

Savalek:
@felixcheung, these tests should fall, because `maven` incorrectly handled errors when testing through `npm`. This PR is only the first part of fixing problems. The second part is the correction of those tests that have been falling for a long time, but because of the bug of the `frontend maven plugin 1.4`, `travis` shows that the `npm` tests have passed, but in fact this is not the case.

felixcheung:
Can you please fix the tests then. We can’t merge anything that will cause all PRs to fail in Travis.

Even when they “should” be

Savalek:
@felixcheung, I'm trying to fix the tests but I can't even run `npm run test` in `zeppelin-web` because the file `webpack.config.js` made for version `webpack 1` but the current version `4`. 
They are not compatible between. 
In addition, I counted at least 33 broken tests.
And i'm not strong in the frontend.

Savalek:
@felixcheung, i managed to fix the tests.

Savalek:
What about merge?

PR #3238, opened 2018-11-23
ZEPPELIN-3870. Configuration in spark-defaults.conf doesn't take effect

zjffdu:
### What is this PR for?

By default, spark.executor.memory is not set in spark's interpreter setting, if I set it in spark-defaults.xml, then it won't take effect. Because spark.executor.memory in spark's interpreter setting will override that value in spark-defaults.xml. This doesn't make sense, we should only override property in spark-default.xml when you have non-empty value.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3870

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
@@ -19,6 +19,7 @@ package org.apache.zeppelin.interpreter.remote;
 
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.thrift.TException;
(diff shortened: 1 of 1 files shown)


weand:
@zjffdu 

i tried to build the latest branch-0.8 where this PR has been merged recently.

build fails with `RemoteInterpreterServer.java:[22,32] package org.apache.commons.lang3 does not exist`

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project zeppelin-interpreter: Compilation failure
[ERROR] /home/zeppelin/zeppelin-distro/work/zeppelin-repo/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java:[22,32] package org.apache.commons.lang3 does not exist
```

PR #3239, opened 2018-11-23
Zeppelin 3879: create "maxRows" and "rowsFetchSize" values in interpreter/jdbc/interpreter-setting.json

monsieurp:
### What is this PR for?
This PR introduces code to configure manually the "maxRows" and "rowsFetchSize" Statement settings directly from the interpreter/jdbc/interpreter-setting.json file. The default value for these two settings is 1000 which can slow down a Zeppelin instance quite a lot if a SQL query returns a big chunk of data. See [1] and [2] for more info about these function calls.

[1]: https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setFetchSize-int-
[2]: https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setMaxRows-int-

### What type of PR is it?
Improvement.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3879

* CI pass

### Questions:
* Does this needs documentation? Yes. For the time being, I'd like the Zeppelin authors input whether such a change makes sense.

Thanks!

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -136,12 +136,14 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private static final String CONCURRENT_EXECUTION_COUNT =
           "zeppelin.jdbc.concurrent.max_connection";
   private static final String DBCP_STRING = "jdbc:apache:commons:dbcp:";
+  private static final String MAX_ROWS_KEY = "zeppelin.jdbc.maxRows";
 
   private final HashMap<String, Properties> basePropretiesMap;
   private final HashMap<String, JDBCUserConfigurations> jdbcUserConfigurationsMap;
   private final HashMap<String, SqlCompleter> sqlCompletersMap;
 
   private int maxLineResults;
+  private int maxRows;
 
   public JDBCInterpreter(Properties property) {
     super(property);
@@ -209,6 +211,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
     logger.debug("JDBC PropretiesMap: {}", basePropretiesMap);
 
     setMaxLineResults();
+    setMaxRows();
   }
 
   protected boolean isKerboseEnabled() {
@@ -226,6 +229,14 @@ public class JDBCInterpreter extends KerberosInterpreter {
         basePropretiesMap.get(COMMON_KEY).containsKey(MAX_LINE_KEY)) {
       maxLineResults = Integer.valueOf(basePropretiesMap.get(COMMON_KEY).getProperty(MAX_LINE_KEY));
     }
+  }
+
+  /**
+   * Fetch MAX_ROWS_KEYS value from property file and set it to
+   * "maxRows" value.
+   */
+  private void setMaxRows() {
+    maxRows = Integer.valueOf(getProperty(MAX_ROWS_KEY, "1000"));
   }
 
   private SqlCompleter createOrUpdateSqlCompleter(SqlCompleter sqlCompleter,
@@ -705,7 +716,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
         // fetch n+1 rows in order to indicate there's more rows available (for large selects)
         statement.setFetchSize(getMaxResult());
-        statement.setMaxRows(getMaxResult() + 1);
+        statement.setMaxRows(maxRows);
 
         if (statement == null) {
           return new InterpreterResult(Code.ERROR, "Prefix not found.");
--- a/jdbc/src/main/resources/interpreter-setting.json
+++ b/jdbc/src/main/resources/interpreter-setting.json
@@ -122,6 +122,13 @@
         "defaultValue": "-1",
         "description": "Maximum of connection lifetime in milliseconds. A value of zero or less means the connection has an infinite lifetime.",
         "type": "number"
+      },
+      "zeppelin.jdbc.maxRows": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.maxRows",
+        "defaultValue": "1000",
+        "description": "Maximum number of rows fetched from the query.",
+        "type": "number"
       }
     },
     "editor": {


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -704,8 +726,8 @@ private InterpreterResult executeSql(String propertyKey, String sql,
         statement = connection.createStatement();
 
         // fetch n+1 rows in order to indicate there's more rows available (for large selects)
-        statement.setFetchSize(getMaxResult());
-        statement.setMaxRows(getMaxResult() + 1);
+        statement.setFetchSize(getRowsFetchSize());
```
zjffdu: getRowFetchSize -> rowsFetchSize
getMaxRows -> maxRows

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -878,6 +900,17 @@ public int getMaxResult() {
...
+	return maxRows;
+  }
+
+  /**
+   * @return the rows fetch size
+   */
+  public int getRowsFetchSize() {
+	return rowsFetchSize;
+  }
+
```
zjffdu: Since these 2 fields are not used out of `JDBCInterpreter.java`, we can remove these 2 get methods

zjffdu:
Thanks @monsieurp for the contribution, I left a few comments, otherwise LGTM

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -228,6 +234,22 @@ private void setMaxLineResults() {
     }
   }
 
+  /**
+   * Fetch ROWS_FETCH_SIZE_KEY value from property file and set it to
+   * "rowsFetchSize" value.
+   */
+  private void setRowsFetchSize() {
+    rowsFetchSize = Integer.valueOf(getProperty(ROWS_FETCH_SIZE_KEY));
```
aka-demik: I think to preserve backward compatibility, it is better to first check for the presence of `ROWS_FETCH_SIZE_KEY / MAX_ROWS_KEY` in the properties. If there is no parameter, specify a reasonable default.
Examples:

- [JDBCInterpreter.java:L259 defaultIfEmpty](https://github.com/apache/zeppelin/blob/58068816a341c96aa96a08663a16eb89a7d85904/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java#L259)
- [JDBCInterpreter.java:L221 if (!isEmpty(getProperty](https://github.com/apache/zeppelin/blob/58068816a341c96aa96a08663a16eb89a7d85904/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java#L221)

zjffdu: Regarding the default, you can use `getProperty(key, defaultValue)`

zjffdu:
@monsieurp This might be flaky test, let me take a look at it.

jdbc/src/main/resources/interpreter-setting.json
```diff
@@ -122,6 +122,20 @@
         "defaultValue": "-1",
         "description": "Maximum of connection lifetime in milliseconds. A value of zero or less means the connection has an infinite lifetime.",
         "type": "number"
+      },
+      "zeppelin.jdbc.rowsFetchSize": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.rowsFetchSize",
+        "defaultValue": "1000",
+        "description": "Number of rows fetched from the database.",
```
felixcheung: instead of "Number of rows fetched from the database.", should it be "Number of rows fetched from executing the query."

jdbc/src/main/resources/interpreter-setting.json
```diff
@@ -122,6 +122,20 @@
...
+        "type": "number"
+      },
+      "zeppelin.jdbc.maxRows": {
+        "envName": null,
+        "propertyName": "zeppelin.jdbc.maxRows",
+        "defaultValue": "1000",
+        "description": "Maximum number of rows fetched from the database.",
```
felixcheung: ditto about database

jdbc/src/main/resources/interpreter-setting.json
monsieurp: Thanks! Fixed.

jdbc/src/main/resources/interpreter-setting.json
monsieurp: Also fixed.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
monsieurp: Problem addressed. See the latest commit.

monsieurp:
Is there anything else I can improve?

zjffdu:
@monsieurp Could you do a rebase and retrigger the travis build ?

monsieurp:
@zjffdu: Sorry for taking so long to respond. It's been a busy week. I've rebased my PR as requested.

zjffdu:
@monsieurp I though about it again, and feel that providing 2 properties is very confusing to users. I think it is better to give one property (zeppelin.jdbc.maxRows) what is user care about. `zeppelin.jdbc.rowsFetchSize` is internal property that zeppelin should configure to users. Here's one post explain the difference between `setFetchSize` and `setMaxRows` 
https://stackoverflow.com/questions/32399546/what-is-the-difference-between-statement-setmaxrows-vs-statement-setfetchsize-in?lq=1

zjffdu:
ping @monsieurp

monsieurp:
@zjffdu: Hi! Fair enough. What do you want me to do then? Should I remove the `rowsFetchSize` property?

zjffdu:
It is weird that I see another property `common.max_count` but it is never used in `JdbcInterpreter`. I think we just need to keep one and remove others.   Personally I prefer `zeppelin.jdbc.maxRows`

monsieurp:
@zhongneu: OK, done.

monsieurp:
@zffdu: Sorry for not responding earlier. IMHO, deleting `common.max_count` should be dealt with in a separate PR / bug report since it is completely unrelated to the feature here. Until then, can we merge this PR? Thanks!

zjffdu:
No problem, Thanks @monsieurp , will merge this PR soon. BTW, could you help create a ticket for removing `common.max_count` and welcome to contribute again 😄

zjffdu:
@monsieurp Is `Patrice Clement` your jira account ? I'd like to assign this ticket to you.

monsieurp:
Yes it is! :)

PR #3242, opened 2018-11-26
[ZEPPELIN-3882] Neo4jInterpreter - Support Point and Date Types

conker84:
### What is this PR for?
Add the support for Point and Date data types introduced since Neo4j 3.4

### What type of PR is it?
[Improvement] In order to allow users to use the Neo4j Interpreter with the last introduced Data types

### Todos
* [x] - Bumped Neo4j dependecies version
* [x] - Added new data types to test

### What is the Jira issue?
* Put link here, and add [ZEPPELIN-3882](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-3882)

* Execute the unit tests

### Questions:

### DIFF
--- a/docs/interpreter/neo4j.md
+++ b/docs/interpreter/neo4j.md
@@ -26,6 +26,9 @@ limitations under the License.
 ## Overview
 [Neo4j](https://neo4j.com/product/) is a native graph database, designed to store and process graphs from bottom to top.
 
+### Supported Version
+
+The Neo4j Interpreter supports all Neo4j versions since v3 via the official [Neo4j Java Driver](https://github.com/neo4j/neo4j-java-driver)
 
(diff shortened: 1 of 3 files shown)


conker84:
Hi @felixcheung the driver is compatible with all Neo4j version >= 3 (the last one is the 3.5). Actually in the documentation there is no mention about the minimum supported version. Do you want that I'll update it?

felixcheung:
perhaps we should have a min version compatible in the interpreter doc?
otherwise I think it's fine not to have tested version, supported version etc in doc since then we'd need to keep updating them

conker84:
Documentation updated!

felixcheung:
can you check the failed tests?

conker84:
Now there are two failing that I think are not related to this PR

conker84:
I found a bug, I'll keep you posted.

PR #3243, opened 2018-11-29
[ZEPPELIN-3886] Remove dependency on flatmap-stream 0.1.1

drod331:
### What is this PR for?
Updated event-stream version to 4.0.1 in zeppelin-web, due to version 3.3.6 containing a malicious dependency that was removed from npmjs.

### What type of PR is it?
[Hot Fix] A fix for the zeppelin-web module so that it won't fail on the (now) non-existent dependencies.

### Todos
* [X] - Update event-stream version to 4.0.1.
* [X] - Update event-stream resolved to .../event-stream-4.0.1.tgz.
* [X] - Update event-stream integrity to the SHA512 key on registry.npmjs.org.
* [X] - Remove flatmap-stream 0.1.1 from the requires list.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3886

### How should this be tested?
Execute the unit tests

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #3246, opened 2018-12-03
ZEPPELIN-3892 Make SecurityUtils as a service

jongyoul:
### What is this PR for?
Making SecurityService more configurable. Currently, SecurityUtils consists of static methods, and which means it's not extendable and pluggable. It would be better if users could implement new security features easily.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Introduce `SecurityService` as an interface to support security features
* [x] - Move `SecurityUtils` to `ShiroSecurityService` as an implementation of `SecurityService`

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3892

* This is a refactoring. It should pass all of the current tests

### Questions:


jongyoul:
Ready to review. Failed tests might be related to #3243

jongyoul:
https://travis-ci.org/jongyoul/zeppelin/builds/462832712

PR #3249, opened 2018-12-04
[ZEPPELIN-3898] Adding repo in testDelRepo so test can run by itself

OrDTesters:
### What is this PR for?
Test DependencyResolverTest.testDelRepo fails when run on its own. The test needs a repo to already be added so it can delete it. If the test is run on its own, the repo is not added to start and so the test fails. The proposed PR fixes this issue by having the test add the repo at the beginning.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3898

### DIFF
--- a/zeppelin-interpreter/src/test/java/org/apache/zeppelin/dep/DependencyResolverTest.java
+++ b/zeppelin-interpreter/src/test/java/org/apache/zeppelin/dep/DependencyResolverTest.java
@@ -66,6 +66,7 @@ public class DependencyResolverTest {
 
   @Test
   public void testDelRepo() {
+    resolver.addRepo("securecentral", "https://repo1.maven.org/maven2", false);
     int reposCnt = resolver.getRepos().size();
     resolver.delRepo("securecentral");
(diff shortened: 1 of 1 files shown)


zjffdu:
@OrDTesters What do you mean "run on its own" ? Do you mean to run it in IDE ?

OrDTesters:
If I run the command `mvn test -Dtest=DependencyResolverTest#testDelRepo` to run the test method by itself, the test fails. The proposed patch that has the test add the repo beforehand leads to the command to pass.

zjffdu:
Thanks @OrDTesters LGTM

OrDTesters:
From what I can tell, the build is failing on Travis due to one job taking too long to run. I do not believe it to be related to my change.

felixcheung:
@OrDTesters could you take a look https://travis-ci.org/OrDTesters/zeppelin/jobs/463188385

zjffdu:

PR #3263, opened 2018-12-13
ZEPPELIN-3907. Add code statement into Spark JobGroup Id for SparkInterpreter

zjffdu:
### What is this PR for?
Straightforward change to add code statement into Spark JobGroupId, so that we can find the associated spark job in spark ui easier. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3907

* Manually tested.

image

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
@@ -103,7 +103,7 @@ class Utils {
   static boolean isScala2_11() {
     return !isScala2_10();
   }
-  
+
   static boolean isCompilerAboveScala2_11_7() {
     if (isScala2_10() || SCALA_COMPILER_VERSION == null) {
       return false;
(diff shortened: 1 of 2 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -147,9 +147,11 @@ static boolean isSpark2() {
       return false;
     }
   }
-  
+
   public static String buildJobGroupId(InterpreterContext context) {
-    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId();
+    return "zeppelin-" + context.getNoteId() + "-" + context.getParagraphId() + "\n" +
```
felixcheung: it looks like the newline doesn't show up in the web interface?

FireArrow:
While this is a really nice feature to have, it does introduce some risk of information leakage in an organization. Take for example `select * from tableofsecrets where secretcolumn = 'supersecret'`.
What about adding a configuration parameter for activating this feature?

Tagar:
> While this is a really nice feature to have, it does introduce some risk of information leakage in an organization. Take for example `select * from tableofsecrets where secretcolumn = 'supersecret'`.
> What about adding a configuration parameter for activating this feature?

That's a good point. 

Although I think it's possible to lock down Spark Driver UI through `spark.ui.filters`?

Also it might be possible to redact out passwords and other sensitive information in Zeppelin - something similar to `spark.redaction.regex`.

Maybe this change should be an option and off by default?

Tagar:
Another option is to set Spark `JobGroup Id` to paragraph's title ..

PR #3264, opened 2018-12-14
Fix Date Sorting issue

fred521:
### What is this PR for?
When plotting a histogram, Zeppelin does not order the x variable correctly, making the plot incorrect.

I have attached screenshots of both histograms below.

It seems that Zeppelin is putting numbers with no decimal places to the left (as if the were smaller) than numbers with decimal places. I tried casting the x axis value to a float but the problem persists.ns.html

### What type of PR is it?
[Bug Fix]

### Todos
N/A

### What is the Jira issue?
[ZEPPELIN-3908]
https://issues.apache.org/jira/browse/ZEPPELIN-3908

It is in the Jira

### Questions:

### DIFF
--- a/zeppelin-web/src/app/visualization/builtins/visualization-nvd3chart.js
+++ b/zeppelin-web/src/app/visualization/builtins/visualization-nvd3chart.js
@@ -164,6 +164,18 @@ export default class Nvd3ChartVisualization extends Visualization {
     let rowIndexValue = {};
 
     for (let k in rows) {
+    let rowKeys = Object.keys(rows);
+    let isNumberKey = false;
+
+    rowKeys.sort(function(a, b) {
+      if (!isNaN(a) && !isNaN(b) && typeof Number(a) === 'number' && typeof Number(b) === 'number') {
+        isNumberKey = true;
+        return a-b;
+      }
+    });
+    rowKeys = isNumberKey ? rowKeys : Object.keys(rows);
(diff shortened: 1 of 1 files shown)


fred521:
> is there js test we could add?

I agree, just will take more time to write the test framework for this bug.

kennyishihara:
Any update to this?

fred521:
the problem of the code is for loop in the line 166 can't sort for number,
have to add extra check the variable is number or string

"for (let k in rows)"

zeppelin-web/src/app/visualization/builtins/visualization-nvd3chart.js
```diff
@@ -164,6 +164,18 @@ export default class Nvd3ChartVisualization extends Visualization {
     let rowIndexValue = {};
 
     for (let k in rows) {
+    let rowKeys = Object.keys(rows);
+    let isNumberKey = false;
+
+    rowKeys.sort(function(a, b) {
+      if (!isNaN(a) && !isNaN(b) && typeof Number(a) === 'number' && typeof Number(b) === 'number') {
```
lleohao: `typeof Number(any)` constant equal 'number'

lleohao: The rowKeys is string array and its value is not empty. You can judge this way
```javascript
if (!isNaN(a) && !isNaN(b)) {...}
```

PR #3267, opened 2018-12-17
[ZEPPELIN-3609] Refactoring `ZeppelinServer` to adopt DI

jongyoul:
### What is this PR for?
Clarifying the structure of ZeppelinServer and related classes.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Adopt DI on Zeppelin server

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3609

All CI should be passed

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,13 +90,13 @@ matrix:
     - sudo: required
       jdk: "oraclejdk8"
       dist: trusty
-      env: PYTHON="2" PROFILE="-Pspark-2.2" SPARKR="true" BUILD_FLAG="package install -DskipTests -DskipRat -am -amd" TEST_FLAG="test -DskipRat -am" MODULES="-pl zeppelin-plugins,zeppelin-zengine,zeppelin-server,spark/interpreter,spark/spark-dependencies" TEST_PROJECTS="-Dtest=ZeppelinSparkClusterTest,org.apache.zeppelin.spark.* -DfailIfNoTests=false"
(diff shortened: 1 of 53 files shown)


felixcheung:
Squash and merge seems like the right option,
https://help.github.com/articles/merging-a-pull-request/

jongyoul:
CI Passed. https://travis-ci.org/jongyoul/zeppelin/builds/471361767. Ready to review.

bhavikpatel9977:
@jongyoul:  Is there any specific reason for moving logic (initialization of classes) from the default constructor to the main method? 

The reason behind asking this is, that I am working on [ZEPPELIN-4356](https://issues.apache.org/jira/browse/ZEPPELIN-4356), Zeppelin Server should stop when it can't create/access a notebook or if **Shiro.ini** file is not configured properly (WIP [pr-3488](https://github.com/apache/zeppelin/pull/3488)).

So to achieve this I have used Immediate scope while binding the classes in ServiceLocator and successfully able to stop the Zeppelin Server, if it can't access/create a notebook & if Shiro configuration is wrong. Also, the exception is thrown when _realms are null_. Because of these changes few test-cases are failing and it throws the exception "realm is null" but it should not be null.
While debugging, I found that the **ShiroValidationService** class with **Immediate** scope is executed before the **Shiro environment gets initialized** due to multithreading, and that results in an error of "_realm was found null_" but normally it should not be null.

Also, I have checked your commits initially you have also used **Immediate** scope for **ShiroAuthenticationService**, later on, you have reverted that changes and used **Singleton** scope; is there any specific reason for that or any issue that you have observed?

PR #3271, opened 2018-12-21
[ZEPPELIN-3919] Paragraph config default value can be customized

liuxunorg:
### What is this PR for?

The title of the paragraph is not displayed by default, and the runOnSelectionChange check option is `ON` by default.
For added flexibility, Add the config configuration section in the interpreter-setting.json file of the interpreter. You can modify the default values with the following four options.

```
"config": {
  "fontSize": 9,
  "colWidth": 12,
  "runOnSelectionChange": false,
  "title": true
}
```

In the `hadoop submarine interpreter`, each paragraph of a notebook represents a complete module of a machine learning algorithm. The name of the algorithm name needs to be filled in through the title of the paragraph. 

**Notebook paragraph display title (Optional)**
The notebook paragraph does not display the title by default.
You can display the title by default with `config.title = true`.

**Notebook run on selection change (Optional)**
The dynamic form in the notebook triggers execution when the selection is change.
You can make the dynamic form in the notebook not trigger execution after selecting the modification by setting `config.runOnSelectionChange=false`.

### What type of PR is it?
[Improvement]

### Todos
* [x] Add paragraph config default configuration in interpreter-setting.json

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3919

[CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/470852962)

alt text

### Questions:

### DIFF
--- a/docs/development/writing_zeppelin_interpreter.md
+++ b/docs/development/writing_zeppelin_interpreter.md
@@ -118,6 +118,12 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
       "language": "your-syntax-highlight-language",
       "editOnDblClick": false,
       "completionKey": "TAB"
+    },
+    "config": {
+      "fontSize": 9,
+      "colWidth": 12,
+      "runOnSelectionChange": true/false,
+      "title": true/false
     }
   },
   {
@@ -183,6 +189,25 @@ Currently `TAB` is only available option.
 }
 ```
 
+### Notebook paragraph display title (Optional)
+The notebook paragraph does not display the title by default.
+You can have the title of the notebook display the title by `config.title=true`.
+
+```json
+"config": {
+  "title": true  # default: false
+}
+```
+
+### Notebook run on selection change (Optional)
+The dynamic form in the notebook triggers execution when the selection is modified.
+You can make the dynamic form in the notebook not trigger execution after selecting the modification by setting `config.runOnSelectionChange=false`.
+
+```json
+"config": {
+  "runOnSelectionChange": false # default: true
+}
+```
 
 ## Install your interpreter binary
 
--- a/docs/development/writing_zeppelin_interpreter.md
+++ b/docs/development/writing_zeppelin_interpreter.md
@@ -120,8 +120,6 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
       "completionKey": "TAB"
     },
     "config": {
-      "fontSize": 9,
-      "colWidth": 12,
       "runOnSelectionChange": true/false,
       "title": true/false
     }
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -400,7 +400,8 @@ public class InterpreterSettingManager implements InterpreterSettingManagerMBean
       //TODO(zjffdu) merge RegisteredInterpreter & InterpreterInfo
       InterpreterInfo interpreterInfo =
           new InterpreterInfo(registeredInterpreter.getClassName(), registeredInterpreter.getName(),
-              registeredInterpreter.isDefaultInterpreter(), registeredInterpreter.getEditor());
+              registeredInterpreter.isDefaultInterpreter(), registeredInterpreter.getEditor(),
+              registeredInterpreter.getConfig());
       interpreterInfo.setConfig(registeredInterpreter.getConfig());
       group = registeredInterpreter.getGroup();
       runner = registeredInterpreter.getRunner();
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -495,20 +495,22 @@ public class InterpreterSettingManager implements InterpreterSettingManagerMBean
     return editor;
   }
 
+  // Get configuration parameters from `interpreter-setting.json`
+  // based on the interpreter group ID
   public Map<String, Object> getConfigSetting(String interpreterGroupId) {
     InterpreterSetting interpreterSetting = get(interpreterGroupId);
     if (null != interpreterSetting) {
       List<InterpreterInfo> interpreterInfos = interpreterSetting.getInterpreterInfos();
+      int infoSize = interpreterInfos.size();
       for (InterpreterInfo intpInfo : interpreterInfos) {
-        if (intpInfo.isDefaultInterpreter() || (interpreterInfos.size() == 1)) {
-          if (intpInfo.getConfig() != null) {
-            return intpInfo.getConfig();
-          }
+        if ((intpInfo.isDefaultInterpreter() || (infoSize == 1))
+            && (intpInfo.getConfig() != null)) {
+          return intpInfo.getConfig();
         }
       }
     }
 
-    return new HashMap<String, Object>();
+    return new HashMap<>();
   }
 
   public List<ManagedInterpreterGroup> getAllInterpreterGroup() {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -401,6 +401,7 @@ public class InterpreterSettingManager implements InterpreterSettingManagerMBean
       InterpreterInfo interpreterInfo =
           new InterpreterInfo(registeredInterpreter.getClassName(), registeredInterpreter.getName(),
               registeredInterpreter.isDefaultInterpreter(), registeredInterpreter.getEditor());
+      interpreterInfo.setConfig(registeredInterpreter.getConfig());
       group = registeredInterpreter.getGroup();
       runner = registeredInterpreter.getRunner();
       // use defaultOption if it is not specified in interpreter-setting.json
@@ -492,6 +493,22 @@ public class InterpreterSettingManager implements InterpreterSettingManagerMBean
       LOGGER.debug("Couldn't get interpreter editor setting");
     }
     return editor;
+  }
+
+  public Map<String, Object> getConfigSetting(String interpreterGroupId) {
+    InterpreterSetting interpreterSetting = get(interpreterGroupId);
+    if (null != interpreterSetting) {
+      List<InterpreterInfo> interpreterInfos = interpreterSetting.getInterpreterInfos();
+      for (InterpreterInfo intpInfo : interpreterInfos) {
+        if (intpInfo.isDefaultInterpreter() || (interpreterInfos.size() == 1)) {
+          if (intpInfo.getConfig() != null) {
+            return intpInfo.getConfig();
+          }
+        }
+      }
+    }
+
+    return new HashMap<String, Object>();
   }
 
   public List<ManagedInterpreterGroup> getAllInterpreterGroup() {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -248,6 +248,10 @@ public class Note implements JsonSerializable {
     this.interpreterSettingManager = interpreterSettingManager;
   }
 
+  InterpreterSettingManager getInterpreterSettingManager() {
+    return this.interpreterSettingManager;
+  }
+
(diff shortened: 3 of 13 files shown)


jongyoul:
Looks fine for me

docs/development/writing_zeppelin_interpreter.md
```diff
@@ -118,6 +118,12 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
       "language": "your-syntax-highlight-language",
       "editOnDblClick": false,
       "completionKey": "TAB"
+    },
+    "config": {
+      "fontSize": 9,
+      "colWidth": 12,
+      "runOnSelectionChange": true/false,
+      "title": true/false
```
zjffdu: `displayTitle` is more meaning IMO

zjffdu:
@liuxunorg What is the behavior if I changed interpreter in one paragraph ? e.g. I change the interpreter from submarine to python. Would the font size change if these 2 interpreter has different config for font size ? If yes, when that change happens ?

liuxunorg:
> @liuxunorg What is the behavior if I changed interpreter in one paragraph ? e.g. I change the interpreter from submarine to python. Would the font size change if these 2 interpreter has different config for font size ? If yes, when that change happens ?

@zjffdu , Thank you are asking is a good question!
Now if you are using a default interpreter with the config attribute set (for example: submarine), adding a paragraph of any other type of interpreter to this note will follow the submarine config configuration, since submarine is the default interpreter.

I will improve the implementation of this function, and also control the display of the front-end paragraphs in the interpreter process according to the configuration in `interpreter-setting.json`.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -492,6 +493,22 @@ public ManagedInterpreterGroup getInterpreterGroupById(String groupId) {
     return editor;
   }
 
+  public Map<String, Object> getConfigSetting(String interpreterGroupId) {
+    InterpreterSetting interpreterSetting = get(interpreterGroupId);
+    if (null != interpreterSetting) {
+      List<InterpreterInfo> interpreterInfos = interpreterSetting.getInterpreterInfos();
+      for (InterpreterInfo intpInfo : interpreterInfos) {
+        if (intpInfo.isDefaultInterpreter() || (interpreterInfos.size() == 1)) {
+          if (intpInfo.getConfig() != null) {
```
felixcheung: merge L502 and 501?

liuxunorg:
@zjffdu , @jongyoul , @felixcheung  Can now support any time, no matter which default interpreter you use, You can dynamically modify the interpreter type when you execute it. :-)

alt text

docs/development/writing_zeppelin_interpreter.md
liuxunorg: "fontSize", "colWidth", "runOnSelectionChange", "title"
These four parameter names need to be consistent with the variable names in paragraph-control.html.

https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/app/notebook/paragraph/paragraph-control.html#L176

docs/development/writing_zeppelin_interpreter.md
```diff
@@ -118,6 +118,12 @@ Here is an example of `interpreter-setting.json` on your own interpreter.
       "language": "your-syntax-highlight-language",
       "editOnDblClick": false,
       "completionKey": "TAB"
+    },
+    "config": {
+      "fontSize": 9,
+      "colWidth": 12,
```
zjffdu: Do we need to expose `fontSize` and `colWidth` to user to customize ? I don't see you need it here. If not, we could remove them, and expose them later in future if it is really necessary.  Because once we expose something to users, we have to maintain it and keep compatibility.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -401,6 +401,7 @@ private void registerInterpreterSetting(List<RegisteredInterpreter> registeredIn
       InterpreterInfo interpreterInfo =
           new InterpreterInfo(registeredInterpreter.getClassName(), registeredInterpreter.getName(),
               registeredInterpreter.isDefaultInterpreter(), registeredInterpreter.getEditor());
+      interpreterInfo.setConfig(registeredInterpreter.getConfig());
```
zjffdu: Put it into constructor.

docs/development/writing_zeppelin_interpreter.md
liuxunorg: Ok, I will remove it first, If need it later, can add it again.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -365,6 +369,13 @@ public void fireParagraphUpdateEvent(Paragraph p) throws IOException {
    */
   public Paragraph insertNewParagraph(int index, AuthenticationInfo authenticationInfo) {
     Paragraph paragraph = new Paragraph(this, paragraphJobListener);
+    if (null != interpreterSettingManager) {
+      // Set the default parameter configuration for the paragraph
+      // based on `interpreter-setting.json` config
+      Map<String, Object> config =
+          interpreterSettingManager.getConfigSetting(defaultInterpreterGroup);
+      paragraph.setConfig(config);
+    }
```
zjffdu: Currently, the interpreter of new paragraph is not the default interpreter of note. It is the interpreter of the paragraph before the current paragraph. (we can discuss whether this is correct behavior). So in that case, I don't think you can set the default config of the default note interpreter.

docs/development/writing_zeppelin_interpreter.md
liuxunorg: Ok, after I finish this PR, I will open a PR and deal with it.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -450,6 +459,21 @@ protected InterpreterResult jobRun() throws Throwable {
...
+      if (null != intpSettingManager) {
+        InterpreterGroup intpGroup = interpreter.getInterpreterGroup();
+        if (intpGroup instanceof ManagedInterpreterGroup) {
+          String name = ((ManagedInterpreterGroup) intpGroup).getInterpreterSetting().getName();
+          Map<String, Object> config
+              = intpSettingManager.getConfigSetting(name);
+          applyConfigSetting(config);
+        }
+      }
+
```
zjffdu: What if user change the config ? e.g. User change the `displayTitle` to false in frontend for submarine interpreter, then after user execute the paragraph, the config will be reset ? Will `displayTitle` become true again ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -569,6 +593,45 @@ public void setConfig(Map<String, Object> config) {
...
+  // When creating a paragraph, it will update some of the configuration
+  // parameters of the paragraph from the web side.
+  // Need to deal with 2 situations
+  // 1. The interpreter does not have a config configuration set,
+  //    so newConfig is equal to null, Need to be processed using the
+  //    default parameters of the interpreter
+  // 2. The user manually modified the  interpreter types of this paragraph.
+  //    Need to delete the existing configuration of this paragraph,
+  //    update with the specified interpreter configuration
+  public void applyConfigSetting(Map<String, Object> newConfig) {
```
zjffdu: Can we add unit test for this ? Since this is not trivial

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
liuxunorg: This is very important because:
When you create a new paragraph, you can directly display the custom configuration, for example, directly display the title bar, otherwise you must run this paragraph to display the title bar.
For example: create a new submarine interpreter note, the paragraph can directly display the title bar to facilitate the user to enter the algorithm name.

zjffdu: I understand this. But as I said before, the default interpreter of new paragraph may not be the default interpreter of note. e.g. You can may get the config of `submarine` but the magics part could be `%python`. So I would suggest we first change the behavior of how to set the default interpreter of new paragraph. This is currently what zeppelin to do. https://github.com/apache/zeppelin/pull/3271/files#diff-0f40daff726ba29c9a004b342d422dfeR380

liuxunorg: There should be no problem. It has been shown in the GIF. You can modify the magic at any time. 
Assuming `submarine` sets the display title, `python` does not set the display title.
In the note created with submarine, all newly created paragraphs will display the title by default using the submarine configuration.
If you use `%python` to modify the default interpreter in a paragraph, Paragraph::applyConfigSetting() will re-use python's configuration to take effect in this paragraph.

The default interpreter behavior for each note, which was previously discussed in the email, may be a more complicated issue.

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -1082,6 +1083,50 @@ public void testAuthorizationRoles() throws IOException {
     assertEquals(user2Notes.get(0).getId(), note.getId());
   }
 
+  @Test
+  public void testInterpreterSettingConfig() {
```
liuxunorg: Added test case. Can check the configuration changes after the interpreter type changes.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
liuxunorg: Added test case. in NotebookTest::testInterpreterSettingConfig()

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
liuxunorg: done.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -450,6 +463,24 @@ protected InterpreterResult jobRun() throws Throwable {
         p.settings.setParams(settings.getParams());
       }
 
+      // After the paragraph is executed,
+      // need to apply the paragraph to the configuration in the
+      // `interpreter-setting.json` config
+      if (this.configSettingNeedUpdate) {
```
liuxunorg: The default configuration is used when the `intpText` changes.
The configuration modified by the user on the paragraph is not overwritten by the interpreter default configuration.

liuxunorg:
@zjffdu, @felixcheung , @jongyoul , Please help me review code. Thanks! :-)

zjffdu:
Thanks @liuxunorg for this PR. I look at the PR again, and have some concerns about it. 

Overall I agree that we should allow user to custom configuration of paragraph. But from my understanding, it is more proper to take it as note configuration when creating Note instead of interpreter config. That means when user create new note, they can specify the default paragraph configuration in this note. And this kind of default configuration will be passed to Paragraph when it is created.  What do you think ?

liuxunorg:
Tanks @zjffdu the question raised, in fact, I have considered this issue for a long time.

I think, Each interpreter is dedicated.
Just like the markdown interpreter, `editOnDblClick:true`, when the markdown syntax is executed, the code editing area will be hidden to directly display the HTML after Markdown rendering. Double-clicking will display the code editing area.
Similarly, in some interpreters such as SubmarineInterpreters, the title bar is displayed by default, which is convenient for the user to enter the algorithm module name.
When the user selects the Submarine interpreter for use, each paragraph will display the title bar by default as a clear behavior, which will not cause confusion for the user.

If the configuration does not take effect when the paragraph is created, the user must execute the code for this paragraph, passing through the interpreter process.
```
public InterpreterResult interpret(String script, InterpreterContext context) {
    context.getConfig().put("title", true);
}
```
Changes the configuration of the front-end paragraph. It is technically feasible, but such an interpreter already explicitly needs to display the title, but it is not displayed by default, which makes the user inconvenient to use and prone to confusion.

And now the configuration of the paragraph has been supported for dynamic modification. Dynamically modifying the interpreter in different paragraphs in the same note will re-enable the paragraph configuration according to the configuration of the corresponding interpreter. To some extent, make up for your concerns.

zjffdu:
Thanks @liuxunorg  LGTM

PR #3283, opened 2019-01-15
[ZEPPELIN-3949] Add HADOOP_CLASSPATH in ZEPPELIN_CLASSPATH when starting Zeppelin

mcapuccini:
### What is this PR for?
Add `HADOOP_CLASSPATH` to `ZEPPELIN_CLASSPATH` when starting Zeppelin. This adds some important deps that Zeppelin may need. For instance when storing the notebooks in [Swift](https://wiki.openstack.org/wiki/Swift) (an Hadoop-compatible storage system) Zeppelin need to use a driver that is generally included in `HADOOP_CLASSPATH`, but not in `HADOOP_CONF_DIR`.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3949

The Travis build should run just fine.

### Questions:
Nope.

Nope.

Nope. Adding the HADOOP_CLASSPATH would be the expected behaviour.

### DIFF
--- a/bin/zeppelin.sh
+++ b/bin/zeppelin.sh
@@ -77,6 +77,10 @@ if [[ -n "${HADOOP_CONF_DIR}" ]] && [[ -d "${HADOOP_CONF_DIR}" ]]; then
   ZEPPELIN_CLASSPATH+=":${HADOOP_CONF_DIR}"
 fi
 
+if [[ -n "${HADOOP_CLASSPATH}" ]] && [[ -d "${HADOOP_CLASSPATH}" ]]; then
+  ZEPPELIN_CLASSPATH+=":${HADOOP_CLASSPATH}"
+fi
+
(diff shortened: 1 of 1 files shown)


mcapuccini:
Hi @felixcheung. I've tried to set `CLASSPATH=$CLASSPATH:$HADOOP_CLASSPATH` before running `bin/zeppelin.sh`, but that causes Zeppelin to crash (I was getting errors like "class not found"). Setting `ZEPPELIN_CLASSPATH` worked just fine for me.

zjffdu:
@mcapuccini Plugin use separate ClassLoader which is different from that of ZeppelinServer.

D01B:
I think you should try to use ZEPPELIN_CLASSPATH_OVERRIDES instead of ZEPPELIN_CLASSPATH(on your own risk) As @zjffdu said, additional jars in base classpath may caused jar conflicts and troubles in runtime.

PR #3286, opened 2019-01-21
Fix abstractions. Removing Message class usage from NotebookService.

D01B:
### What is this PR for?
Removing Message class usage from NotebookService. Using DTO object in services broke abstraction.

### What type of PR is it?
[Improvement | Refactoring]

### What is the Jira issue?
* [ZEPPELIN-3958]
* [ZP-104]

* Travis CI pass

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -29,7 +29,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.CountDownLatch;
 import javax.inject.Inject;
 import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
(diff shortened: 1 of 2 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -955,37 +952,46 @@ public void emptyTrash(ServiceContext context,
...
   public void spell(String noteId,
-                    Message message,
+                    String paragraphId,
+                    String title,
+                    String text,
+                    Object results,
+                    String errorMessage,
+                    String dateStarted,
+                    String dateFinished,
+                    String status,
```
felixcheung: should this be typed Job.Status?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -955,37 +952,46 @@ public void emptyTrash(ServiceContext context,
   }
 
   public void spell(String noteId,
-                    Message message,
+                    String paragraphId,
+                    String title,
+                    String text,
+                    Object results,
```
felixcheung: InterpreterResult?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
D01B: Object type or a variable name?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
D01B: yes, it's InterpreterResult, I think we should do as much as possible logic inside the service class, not in a facade.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
D01B: I think it should be Object, in case if we can't parse status, it may be set with a default value or  be gotten from another source

D01B: if you want I could change the type of status.

zjffdu:
Could you add Jira Id in the PR title ?

PR #3288, opened 2019-01-24
[ZEPPELIN-2274] - update cron UI

Savalek:
### About PR
This PR change cron ui on more user-friendly.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
[ZEPPELIN-2274](https://issues.apache.org/jira/browse/ZEPPELIN-2274)

peek 2019-01-24 11-47

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -886,6 +886,23 @@ public class NotebookRestApi extends AbstractRestApi {
   }
 
   /**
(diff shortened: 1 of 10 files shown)


liuxunorg:
The effect is good. It means that the color matching of the interface is not well coordinated with zeppelin, and the background color of the editing area on the right may be better.

prabhjyotsingh:
The theme does not match our notebook UI, otherwise looks good.

felixcheung:
@liuxunorg can you consider an update on the UI color scheme?

felixcheung:
ping @Savalek

Savalek:
I'm not strong in the front end)
Do you want me to replace all selections with input?

PR #3296, opened 2019-01-31
ZEPPELIN-3974. Reorganize zeppelin integration test

zjffdu:
### What is this PR for?
I create a new module zeppelin-interpreter-integration which contains all the interpreter related integration test. The reason I didn't put it into zeppelin-integration is that zeppelin-integration module's selenium dependency has transitive guava dependency which conflicts with that of zeppelin zengine.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3974

* CI pass

### Questions:


Leemoonsoo:
LGTM!

PR #3306, opened 2019-02-13
[ZEPPELIN-4001] zeppelin-interpreter-integration is ignored due to wrong folder structure

zjffdu:
### What is this PR for?

zeppelin-interpreter-integration is ignored due to wrong folder structure, that means all the integration of this module is ignored. This PR fix this issue, rename `zeppelin-interpreter-integration/src/main/test` to  `zeppelin-interpreter-integration/src/test`. Besides that, I split `SparkIntegrationTest` & `ZeppelinSparkClusterTest` into each spark version and each travis build will only run integration test and unit test for each spark version.  This could save travis build time as each build will download one spark distribution one time.

### What type of PR is it?
[Bug Fix | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4001

* CI pass

### Questions:


.travis.yml
```diff
@@ -37,9 +37,13 @@ addons:
     - r-source
     - sourceline: 'deb http://cran.rstudio.com/bin/linux/ubuntu trusty/'
       key_url: 'keyserver.ubuntu.com/pks/lookup?op=get&search=0x51716619E084DAB9'
+    - r-packages-trusty
+    - mysql-5.7-trusty
     packages:
     - r-base
     - r-base-dev
+    - mysql-server
+    - mysql-client
```
felixcheung: why is this needed?

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
```diff
@@ -86,7 +86,7 @@ public boolean isSpark2() {
   }
 
   public boolean isSecretSocketSupported() {
-    return this.newerThanEquals(SPARK_2_3_1);
+    return this.newerThanEquals(SPARK_2_3_1) || this.equals(SparkVersion.fromVersionString("2.1.2"));
```
felixcheung: I don't think we should hardcode this? why not newerThanEquals(SPARK_2_1_2)?

.travis.yml
```diff
@@ -87,35 +91,43 @@ matrix:
...
+
+    # ZeppelinSparkClusterTest16, SparkIntegrationTest16, Unit test of Spark 1.6
+    - sudo: required
+      jdk: "oraclejdk8"
       dist: trusty
-      env: PYTHON="2" SCALA_VER="2.10" PROFILE="-Pspark-1.6 -Phadoop2 -Pscala-2.10" SPARKR="true" BUILD_FLAG="install -DskipTests -DskipRat -am" TEST_FLAG="test -DskipRat -am" MODULES="-pl spark/interpreter,spark/spark-dependencies" TEST_PROJECTS="-Dtest=org.apache.zeppelin.spark.*,org.apache.zeppelin.rinterpreter.*,org.apache.spark.api.r.* -DfailIfNoTests=false"
+      env: BUILD_PLUGINS="true" PYTHON="3" SCALA_VER="2.10" PROFILE="-Pspark-1.6 -Phadoop2 -Pscala-2.10 -Pintegration" SPARKR="true" BUILD_FLAG="install -DskipTests -DskipRat -am" TEST_FLAG="test -DskipRat -am" MODULES="-pl zeppelin-interpreter-integration,zeppelin-web,spark/spark-dependencies" TEST_PROJECTS="-Dtest=ZeppelinSparkClusterTest16,SparkIntegrationTest16,org.apache.zeppelin.spark.* -DfailIfNoTests=false"
```
felixcheung: honestly, I think we need to deprecate support for spark 2.2.3 and/or earlier - which is End of Life

.travis.yml
zjffdu: This is for `JdbcIntegrationTest`, otherwise this test will fail

.travis.yml
zjffdu: I do think we could deprecate support for some old version of spark, but I am afraid there're still some users using that. Let me raise a thread in user mail-list to discuss it. @felixcheung Where can I find the deprecation info for spark ? I didn't follow spark community for a while.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
zjffdu: This is by accident, I have removed it.

zjffdu:
Thanks @felixcheung  will merge it if no more comments

.travis.yml
felixcheung: bottom of https://spark.apache.org/versioning-policy.html

PR #3307, opened 2019-02-16
ZEPPELIN-4005. SparkRInterpreter is broken for spark 2.1.3 and 2.2.2

zjffdu:
### What is this PR for?

The security patch about the security communication between JVM and R process is backported to spark 2.13 and 2.2.2. And Zeppelin didn't catch that. This PR fix it and also update the test case. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4005

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
@@ -37,6 +37,9 @@ public class SparkVersion {
   public static final SparkVersion UNSUPPORTED_FUTURE_VERSION = SPARK_3_0_0;
 
   private int version;
+  private int majorVersion;
+  private int minorVersion;
+  private int patchVersion;
   private String versionString;
 
   SparkVersion(String versionString) {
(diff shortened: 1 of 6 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkVersion.java
```diff
@@ -86,7 +86,11 @@ public boolean isSpark2() {
   }
 
   public boolean isSecretSocketSupported() {
-    return this.newerThanEquals(SPARK_2_3_1);
+    return this.newerThanEquals(SPARK_2_3_1) ||
+            (this.newerThanEquals(SparkVersion.fromVersionString("2.2.2")) &&
+                    this.olderThan(SparkVersion.fromVersionString("2.3.0"))) ||
+            (this.newerThanEquals(SparkVersion.fromVersionString("2.1.3")) &&
+                    this.olderThan(SparkVersion.fromVersionString("2.2.0")));
```
felixcheung: I think this works, but if and only if they are evaluated in the given order, which is a bit error prone. might be worthwhile to add a `newerThanEqualsPatchVersion` for `2.2.[2+]` type of check

zeppelin-interpreter-integration/src/test/java/org/apache/zeppelin/integration/ZeppelinSparkClusterTest21.java
```diff
@@ -34,7 +34,7 @@ public ZeppelinSparkClusterTest21(String sparkVersion) throws Exception {
   @Parameterized.Parameters
   public static List<Object[]> data() {
     return Arrays.asList(new Object[][]{
-            {"2.1.2"}
+            {"2.1.3"}
```
felixcheung: can this be passed in thru travis? 

and should these be updated https://github.com/apache/zeppelin/blob/master/.travis.yml#L136 "2.1.0"

https://github.com/apache/zeppelin/blob/master/.travis.yml#L142 "2.2.0"

zjffdu: These 2 are livy interpreter, not related here

felixcheung: ok for livy is there new version support https://github.com/apache/zeppelin/blob/master/.travis.yml#L142?

zjffdu: 0.5.0 is the latest livy release, livy community is discussing release of 0.6.0

PR #3308, opened 2019-02-16
[ZEPPELIN-3986]. Cannot access any JAR in yarn cluster mode

zjffdu:
### What is this PR for?

User specified jars is missing in yarn-cluster mode due to we didn't detect the user jar correctly. This PR fix the detecting jar logic in `BaseSparkScalaInterpreter`. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3986

* System integration test is added into SparkIntegrationTest, we tested the case of `spark.jars` & `spark.jars.packages`

### Questions:

### DIFF
--- a/spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
+++ b/spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
@@ -370,34 +370,6 @@ public class NewSparkInterpreterTest {
     interpretThread.join();
   }
 
-  @Test
-  public void testDependencies() throws IOException, InterpreterException {
-    Properties properties = new Properties();
-    properties.setProperty("spark.master", "local");
-    properties.setProperty("spark.app.name", "test");
-    properties.setProperty("zeppelin.spark.maxResult", "100");
-    properties.setProperty("zeppelin.spark.useNew", "true");
-    // disable color output for easy testing
-    properties.setProperty("zeppelin.spark.scala.color", "false");
-
-    // download spark-avro jar
-    URL website = new URL("http://repo1.maven.org/maven2/com/databricks/spark-avro_2.11/3.2.0/spark-avro_2.11-3.2.0.jar");
-    ReadableByteChannel rbc = Channels.newChannel(website.openStream());
-    File avroJarFile = new File("spark-avro_2.11-3.2.0.jar");
-    FileOutputStream fos = new FileOutputStream(avroJarFile);
-    fos.getChannel().transferFrom(rbc, 0, Long.MAX_VALUE);
-
-    properties.setProperty("spark.jars", avroJarFile.getAbsolutePath());
-
-    interpreter = new SparkInterpreter(properties);
-    assertTrue(interpreter.getDelegation() instanceof NewSparkInterpreter);
-    interpreter.setInterpreterGroup(mock(InterpreterGroup.class));
-    interpreter.open();
-
-    InterpreterResult result = interpreter.interpret("import com.databricks.spark.avro._", getInterpreterContext());
-    assertEquals(InterpreterResult.Code.SUCCESS, result.code());
-  }
-
   //TODO(zjffdu) This unit test will fail due to classpath issue, should enable it after the classpath issue is fixed.
   @Ignore
   public void testDepInterpreter() throws InterpreterException {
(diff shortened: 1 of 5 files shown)


spark/interpreter/src/test/java/org/apache/zeppelin/spark/NewSparkInterpreterTest.java
```diff
@@ -370,34 +370,6 @@ public void run() {
...
-    properties.setProperty("spark.jars", avroJarFile.getAbsolutePath());
-
-    interpreter = new SparkInterpreter(properties);
-    assertTrue(interpreter.getDelegation() instanceof NewSparkInterpreter);
-    interpreter.setInterpreterGroup(mock(InterpreterGroup.class));
-    interpreter.open();
-
-    InterpreterResult result = interpreter.interpret("import com.databricks.spark.avro._", getInterpreterContext());
-    assertEquals(InterpreterResult.Code.SUCCESS, result.code());
-  }
```
zjffdu: We can not test user jar case in unit test (MutableURLClassLoader only exist in real spark app which is launched via spark-submit) . So remove it here.

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -370,20 +374,25 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+    var extraJars = Seq.empty[String]
+    while (classLoader != null) {
+      if (classLoader.getClass.getCanonicalName ==
+        "org.apache.spark.util.MutableURLClassLoader") {
+        extraJars = classLoader.asInstanceOf[URLClassLoader].getURLs()
+          // Check if the file exists.
+          .filter { u => u.getProtocol == "file" && new File(u.getPath).isFile }
+          // Some bad spark packages depend on the wrong version of scala-reflect. Blacklist it.
+          .filterNot { u =>
+          Paths.get(u.toURI).getFileName.toString.contains("org.scala-lang_scala-reflect")
```
felixcheung: can we make this configurable?

zjffdu: User jar is configurable via `spark.jars` or `spark.jars.packages`. Here's the internal mechnism at runtime of detecting what user jars user has been specified

zjffdu:
@felixcheung kafka jar is for unit testing. https://github.com/apache/zeppelin/pull/3308/files#diff-338173a3721f4282e638ac5145b4bfe7R88

felixcheung:
We really shouldn’t include binary jar in the source code repo.

Other projects do so by building a test jar in test from source.

zjffdu:
@felixcheung I have removed the kafka jar and use the zeppelin-interpreter-integration jar

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
felixcheung: I mean the `org.scala-lang_scala-reflect` 

also indentation seems off

zeppelin-interpreter-integration/src/test/java/org/apache/zeppelin/integration/SparkIntegrationTest.java
```diff
@@ -80,7 +85,14 @@ public static void tearDown() throws IOException {
     }
   }
 
-  private void testInterpreterBasics() throws IOException, InterpreterException {
+  private void testInterpreterBasics() throws IOException, InterpreterException, XmlPullParserException {
+    // add jars & packages for testing
+    InterpreterSetting sparkInterpreterSetting = interpreterSettingManager.getInterpreterSettingByName("spark");
+    sparkInterpreterSetting.setProperty("spark.jars.packages", "com.maxmind.geoip2:geoip2:2.5.0");
```
felixcheung: use something in org.apache.zeppelin instead?

zjffdu: It may already shipped to spark driver and put under classpath. So it is better to use some other libraries.

zjffdu:

maziyarpanahi:
Hi @zjffdu,
Is this being also added to 0.8.2 or just the master? 

Many thanks.

zjffdu:
Both master & 0.8.2

maziyarpanahi:
Thakns @zjffdu,
The reason I asked was that it still only works if I mention the JARs in `spark.jars` in UI but not with 

```bash
export SPARK_SUBMIT_OPTIONS="--jars "
```

I clone the latest git and checkout to 0.8 so it has this pull request merged already.

maziyarpanahi:
On top of this, `spark.jars.packages` also doesn't work. It shows the JARs in UI but in import, it can't find them.

PR #3312, opened 2019-02-19
[ZEPPELIN-4004] Fix RemoteResource.invokeMethod()

Leemoonsoo:
### What is this PR for?
RemoteResource is a representation of object in ResourcePool in Interpreter running on another process.

RemoteResource provides a invokeMethod() to call method of an object remotely, which is not working now. This PR provide a fix for that. Also improve the test case to cover this case.

Also provide small improvement to inference parameter type on remote invocation.

### What type of PR is it?
Bug Fix, Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4004

Unittest included. CI pass.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/resource/DistributedResourcePoolTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/resource/DistributedResourcePoolTest.java
@@ -20,6 +20,7 @@ import com.google.gson.Gson;
 import org.apache.zeppelin.interpreter.AbstractInterpreterTest;
 import org.apache.zeppelin.interpreter.InterpreterContext;
 import org.apache.zeppelin.interpreter.InterpreterException;
+import org.apache.zeppelin.interpreter.InterpreterOption;
 import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.InterpreterSetting;
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreter;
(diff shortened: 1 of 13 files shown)


zeppelin-zengine/src/test/java/org/apache/zeppelin/resource/DistributedResourcePoolTest.java
```diff
@@ -44,8 +46,9 @@
   public void setUp() throws Exception {
     super.setUp();
     InterpreterSetting interpreterSetting = interpreterSettingManager.getByName("mock_resource_pool");
+    interpreterSetting.getOption().setPerNote(ISOLATED);
     intp1 = (RemoteInterpreter) interpreterSetting.getInterpreter("user1", "note1", "mock_resource_pool");
-    intp2 = (RemoteInterpreter) interpreterSetting.getInterpreter("user2", "note1", "mock_resource_pool");
+    intp2 = (RemoteInterpreter) interpreterSetting.getInterpreter("user2", "note2", "mock_resource_pool");
```
Leemoonsoo: While `intp1` and `intp2` were configured to the same Note `note1` with `SHARED` mode, the test was instantiating `intp1` and `intp2` in the same process and didn't test access to the remote object in the remote process.

Therefore, changed interpreter option to per note `ISOLATED` and configured `intp1` for `note1` and `intp2 for `note2`, make sure `intp1` and `intp2` runs in different process.

In this way, it can reproduce error described in ZEPPELIN-4004

zjffdu:
Thanks @Leemoonsoo Is there any integration test for it ?

Leemoonsoo:
Thanks @felixcheung @zjffdu for taking a look.
Regarding integration test, I believe DistributedResourcePoolTest.java pretty much do the job. It creates two interpreter processes and test using them.

Leemoonsoo:
Merge to master and branch-0.8 if no further comment.

PR #3313, opened 2019-02-21
[ZEPPELIN-4004] add a systemd unit file to launch the Zeppelin daemon via systemd commands

monsieurp:
### What is this PR for?

Here's a possible systemd unit file to control the Zeppelin Java process using systemd commands. Typically on a Ubuntu Linux server for instance. I've also written a very short README to get noobs started.

### What type of PR is it?
Feature

### Todos
* Review code.
* Merge.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4015

* Install systemd unit file.
* Test.
* Report back.

* None required.

### Questions:
* Does this needs documentation? Yes. Written in a README file for the time being. Might need to write a lengthy documentation page if contribution is accepted.

### DIFF
--- a/bin/zeppelin-systemd-service.sh
+++ b/bin/zeppelin-systemd-service.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# description: Enable/disable the Zeppelin systemd service.
+#
+
+# Directory in which the systemd unit files sit.
+SYSTEMD_DIR=/etc/systemd/system
+
+function enable_systemd_service()
+{
+    # Where are we in the fs?
+    OLD_PWD=$(pwd)
+
+    # Work out where the script is run from and cd into said directory.
+    cd "$(dirname "${BASH_SOURCE[0]}")"
+
+    # Work out the current directory.
+    MY_PWD=$(readlink -f .)
+
+    # Work out the Zeppelin source directory (go up a directory actually).
+    ZEPPELIN_DIR=$(dirname "${MY_PWD}")
+
+    # Copy the unit file.
+    cp "${ZEPPELIN_DIR}"/scripts/systemd/zeppelin.systemd "${SYSTEMD_DIR}"
+
+    # Swap the template variable with the right directory path.
+    sed -i -e "s#%ZEPPELIN_DIR%#${ZEPPELIN_DIR}#g;" \
+        "${SYSTEMD_DIR}"/zeppelin.systemd
+
+    # Set up the unit file.
+    systemctl daemon-reload
+    systemctl enable zeppelin.service
+
+    # Display a help message.
+    echo "To start Zeppelin using systemd, simply type:
+# systemctl start zeppelin
+
+To check the service health:
+# systemctl status zeppelin"
+
+    # Go back where we came from.
+    cd "${OLD_PWD}"
+}
+
+function disable_systemd_service()
+{
+    # Let's mop up.
+    systemctl stop zeppelin.service
+    systemctl disable zeppelin.service
+    rm "${SYSMTED_DIR}"/zeppelin.systemd
+    systemctl daemon-reload
+    systemctl reset-failed
+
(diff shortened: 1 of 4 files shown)


scripts/systemd/zeppelin.systemd
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Apache Zeppelin daemon
+After=syslog.target network.target
+
+[Service]
+Type=oneshot
+ExecStart=/opt/zeppelin-0.8.1-bin-all/bin/zeppelin-daemon.sh start
```
liuxunorg: Hi,  @monsieurp  The path may not be `/opt/zeppelin-0.8.1-bin-all/`?

monsieurp: Hi @liuxunorg! Thanks for reviewing.

You are absolutely right: the directory in which Zeppelin files are to be installed can sit anywhere on the filesystem. I tend to install Zeppelin in `/opt` but that's just my personal preference. Of course, someone else may install Zeppelin in a different place and therefore, the path may not be `/opt/...` in this case. What do you suggest then? Because a path has to be specified in the systemd unit file.

monsieurp: I have made this assumption very clear in the `README.md` file. Please have a look.

liuxunorg: Perhaps you can add a script for `add_daemon.sh` to `zeppelin/bin/`.
Use this script to generate the correct path to the `zeppelin.service` file in the `/etc/systemd/system` directory.

But you need to pay attention to the following points:
1. You need to determine whether the execution user is `root` in `add_daemon.sh`
2. You need to determine the version of the operating system in `add_daemon.sh`, because the path of the `systemd` folder of different operating systems may be different.
3. You need to determine the version of the operating system in `add_daemon.sh`, because the `systemctl` commands may be different for different operating systems.

monsieurp:
@liuxunorg: Thanks for your input. How about now?

liuxunorg:
1. You need to determine whether the execution user is `root`, Because only `root` can copy files to `/etc/systemd/system`.
```
function check_user()
{
  if [[ $(id -u) -ne 0 ]];then
    echo "This script must be run with a ROOT user!"
    exit
  fi
}
```

2. Need to determine whether the operating system is using `systemd` and `systemctl` for service startup (for example, debian 6 does not have `systemd`), Because there is no `systemd` and `systemctl`, executing the `systemctl` command will fail, so you need to give a prompt information.

3. Combine the `disable_zeppelin_systemd_service.sh` and `enable_zeppelin_systemd_service.sh` files into one file (for example, zeppelin_systemd_service.sh), `disable` and `enable` by executing parameters, Would it be better?

Because not all operating systems use `systemd` for self-starting services. So maybe you can support the operating system with `systemd` installed first? 
This is just my suggestion, See if other people have any opinions? 
anyway, Thank you for your contribution! :-)

monsieurp:
@felixcheung I wondered about the same thing... :)

zjffdu:
We do need to doc for it. @monsieurp

monsieurp:
Hi guys. Thanks for the feedback. I've added the needed and requested documentation. Please have a look and review. Cheers.

monsieurp:
Hi guys. Thanks for the feedback. I've added the needed and requested documentation. Please have a look and review. Cheers.

monsieurp:
@zjffdu @felixcheung: Hi guys! Can someone take a look at this PR? Cheers!

monsieurp:
ping?

monsieurp:
@felixcheung @zjffdu ping!

zjffdu:
@liuxunorg  Could you help review it ? Thanks

bin/zeppelin-systemd-service.sh
```diff
@@ -0,0 +1,101 @@
...
+
+function check_user()
+{
+    # Are we root?
+    if [[ $(id -u) -ne 0 ]]; then
+        echo "Please run this script as root!"
+        exit -1
+    fi
+}
+
```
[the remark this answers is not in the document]
monsieurp: I must disagree unfortunately: your assertation that the `systemctl` command is only available on CentOS is wrong. I use Ubuntu Linux 16.04 on a daily basis and the `systemctl` command is widely available, even on a fresh installation. Try for yourself and you will see.

You are however right that the script should test whether this command is available and exit if that's not the case.

bin/zeppelin-systemd-service.sh
```diff
@@ -0,0 +1,101 @@
...
+USAGE="usage: zeppelin-systemd-service.sh {enable|disable}
+
+  enable:    enable Zeppelin systemd service.
+  disable:   disable Zeppelin systemd service.
+"
+
+# Main method starts from here downwards.
+check_user
+
```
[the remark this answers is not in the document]
monsieurp: I wrote a similar function: `check_systemctl`. Please take a look.

monsieurp:
@zjffdu @liuxunorg: PR updated with your suggestions. Please review.

bin/zeppelin-systemd-service.sh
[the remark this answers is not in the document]
liuxunorg: Test if the command `systemctl` is available in the script, and if not, exit.
It is a good idea. 👍

liuxunorg:
LGTM, thanks for the contribution @monsieurp :-)

monsieurp:
Hi! Can this PR be merged now? @zjffdu @felixcheung?

PR #3314, opened 2019-02-26
[ZEPPELIN-4019] Interpreter process get note from server

liuxunorg:
### What is this PR for?
The interpreter is executed according to the paragraph. Click the run button in the paragraph, and only the code content in the current paragraph can be obtained in the interpreter process.

To pass the Dashboard paragraph in the submarine interpreter, you need to submit all the code content in this note to tensorflow for execution.

So, I think need to get the note data in sever through `RemoteInterpreterEventServer` in the interpreter process.

I implemented two other methods, so that the interpreter can get the note in the server, the actual running effect is not the best solution.

**1) By having the interpreter directly access the external storage of the note repository, such as HDFS, S3.**

1. But this limits the need to store notes in an external storage system.
2. In hdfs with kerberos authentication system enabled, not all users can access the note folder.

**2) Get the note from the server from the REST client in the interpreter.**

1. Need to let zeppelin/api/notes intercept in shrio.ini.
2. Or let the interpreter access the REST interface of the server through the user and password, but it is not safe.

In addition, in [JIRA-4018](https://issues.apache.org/jira/browse/ZEPPELIN-4018) Workflow and orchestration, it is also necessary to query all the notes and paragraphs of the user in the workflow paragraph for the execution of the note.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4019

ShellInterpreter::internalInterpret(String cmd, InterpreterContext contextInterpreter)
```
String noteJson = contextInterpreter.getIntpEventClient().getNoteFromServer(
    cmd,
    contextInterpreter.getAuthenticationInfo(), true);
if (null != noteJson) {
  LOGGER.info(noteJson);
  return new InterpreterResult(Code.SUCCESS, noteJson);
}

return new InterpreterResult(Code.ERROR, "can not get note :" + cmd);
```

getnotefromserver

[CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/498581356)


zjffdu:
@liuxunorg  Currently interpreter process can get all the data of paragraph(text, config, and etc) and run note via id. Could you be more specific about what kind of note data you need to obtain  and for what purpose ?

liuxunorg:
## Submarine interpreter
The use of the submarine interpreter to write the tensorflow algorithm follows such a rule. **One Note One Algorithms Project**
Because a tensorflow algorithm may have multiple python modules. That is, multiple python files.

1. The user writes all the modules of the algorithm in a note.
2. Each module is written in a separate paragraph.
3. Click the `JOB RUN` button in the submarine dashboard to submit all the algorithms in the entire note at one time.

So when the entire paragraph of the submarine dashboard is executed, you need to get the data from all the paragraphs in the entire note, and create a python file for the algorithm in each paragraph.

zepl-submarine-interactive-dev

1. The newly added `RemoteInterpreterEventClient::getNoteFromServer()` has checked the permissions of the noteId. Only the `user's own note`, the note `shared with the user's read permission`, can be successfully obtained.

2. And support json data compressed note from the server to the interpreter process.

## Workflow and orchestration
Workflow and orchestration that may be developed in the future, Also need to get the user's list of notes and paragraphs  in the interpreter process.

Because the user needs to be in the Workflow dashboard,Use the `drop-down box` control to display a list of notes and paragraphs that you have permission to operate.

Workflow orchestration of notes and paragraphs by way of selection. The user experience will be better and not easy to make mistakes.

Of course, Workflow is another topic.

felixcheung:
cool!

zjffdu:
@liuxunorg I just have concern that allowing getting note json is too flexible for users, and it cause risk once the note json format has changes over time. I would suggest to add api like following:

```
public String getParagraphText(String noteId, String paragraphId)
```

liuxunorg:
@zjffdu , Ok!  
Because I need to get all the paragraphs in the note, i modified it to 2 interface functions.
1. Get a list of paragraphIds from noteId
```
public List<String> getParagraphList(String noteId)
```

2. Get the paragraph script from paragraphId
```
public String getParagraphText(String noteId, String paragraphId)
```

zjffdu:
@liuxunorg How about just using one api such as getParagraphsTextList otherwise you need to call these api multiple times

liuxunorg:
@zjffdu ,  Redesigned the call interface, `authenticationInfo` can be used to check if the user has read permission to get the note. Return to ParagraphInfo list once.

```
struct ParagraphInfo {
  1: string noteId,
  2: string paragraphId,
  3: string paragraphTitle,
  4: string paragraphText
}

list<ParagraphInfo> getParagraphList(1: string authenticationInfo, 2: string noteId);
```

liuxunorg:
@zjffdu , CI pass. https://travis-ci.org/liuxunorg/zeppelin/builds/501127271
Please help me review the code. Thanks!

zeppelin-interpreter/src/main/thrift/RemoteInterpreterEventService.thrift
```diff
@@ -107,4 +114,6 @@ service RemoteInterpreterEventService {
   list<string> getAllResources(1: string intpGroupId);
   binary getResource(1: string resourceIdJson);
   binary invokeMethod(1: string intpGroupId, 2: string invokeMethodJson);
+
+  list<ParagraphInfo> getParagraphList(1: string authenticationInfo, 2: string noteId);
```
zjffdu: I think just user should be enough.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterEventClient.java
```diff
@@ -86,6 +88,17 @@ public synchronized ResourceSet getAllResources() {
     }
   }
 
+  public synchronized List<ParagraphInfo> getParagraphList(AuthenticationInfo authInfo, String noteId) {
+    try {
+      String authInfoJson = gson.toJson(authInfo);
+      List<ParagraphInfo> paragraphList = intpEventServiceClient.getParagraphList(authInfoJson, noteId);
```
zjffdu: Does intpEventServiceClient.getParagraphList throw exception when the current user doesn't have permission to read this note ?

liuxunorg: This should only catch the Thrift call exception.
If the user does not have permission to get this note, the function returns `null`.

zjffdu: I think we should throw exception to tell user expclitly that he doesn't have permission on this note. Otherwise it would be very difficult to know why user get null.

zeppelin-interpreter/src/main/thrift/RemoteInterpreterEventService.thrift
liuxunorg: Done.

liuxunorg:
1. Users can only get own notes and notes shared by others.
2. When you don't have read access to the note, There will be an exception prompt.
3. When the note does not exist, There will be an exception prompt.

getparagraphlist

liuxunorg:
@zjffdu , CI pass. https://travis-ci.org/liuxunorg/zeppelin/builds/501368001
Please help me review the code. Thanks!

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1844,6 +1849,61 @@ public void onParaInfosReceived(String noteId, String paragraphId,
...
+      throws TException, ServiceException {
+    try {
+      AuthenticationInfo authInfo = new AuthenticationInfo();
+      authInfo.setUser(user);
+
+      Set<String> userAndRoles = new HashSet<>();
+      userAndRoles.add(user);
+
+      ServiceContext serviceContext = new ServiceContext(authInfo, userAndRoles);
+      Note note = getNotebookService().getNote(noteId, serviceContext,
```
zjffdu: NotebookService is usually for external call (such as websocket or rest api). Because only websocket & rest api needs callback. Here we can directly call notebook() to get Note and check permission.

liuxunorg: @zjffdu  Done. CI pass. https://travis-ci.org/liuxunorg/zeppelin/builds/501825407

zjffdu:
LGTM

PR #3316, opened 2019-03-02
[ZEPPELIN-3985]. Move note permission from notebook-authorization.json to note file

zjffdu:
### What is this PR for?

This PR move note permission from `notebook-authorization.json` to note json file. `AuthorizationService` is responsible for authorization operations, including setting permission and querying permissions. I keep `NoteAuthorization` and will create another PR to upgrade notes.

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-3985

* CI pass

### Questions:

### DIFF
--- a/zeppelin-plugins/notebookrepo/git/src/test/java/org/apache/zeppelin/notebook/repo/NotebookRepoSyncTest.java
+++ b/zeppelin-plugins/notebookrepo/git/src/test/java/org/apache/zeppelin/notebook/repo/NotebookRepoSyncTest.java
@@ -36,10 +36,10 @@ import org.apache.zeppelin.helium.ApplicationEventListener;
 import org.apache.zeppelin.interpreter.InterpreterFactory;
 import org.apache.zeppelin.interpreter.InterpreterSettingManager;
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreterProcessListener;
+import org.apache.zeppelin.notebook.AuthorizationService;
 import org.apache.zeppelin.notebook.Note;
 import org.apache.zeppelin.notebook.NoteInfo;
 import org.apache.zeppelin.notebook.Notebook;
-import org.apache.zeppelin.notebook.NotebookAuthorization;
 import org.apache.zeppelin.notebook.Paragraph;
 import org.apache.zeppelin.search.SearchService;
 import org.apache.zeppelin.storage.ConfigStorage;
@@ -64,9 +64,9 @@ public class NotebookRepoSyncTest {
   private InterpreterFactory factory;
   private InterpreterSettingManager interpreterSettingManager;
   private SearchService search;
-  private NotebookAuthorization notebookAuthorization;
   private Credentials credentials;
   private AuthenticationInfo anonymous;
+  private AuthorizationService authorizationService;
   private static final Logger LOG = LoggerFactory.getLogger(NotebookRepoSyncTest.class);
 
   @Before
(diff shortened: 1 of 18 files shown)


zeppelin-plugins/notebookrepo/git/src/test/java/org/apache/zeppelin/notebook/repo/NotebookRepoSyncTest.java
```diff
@@ -1,417 +1,414 @@
...
-      if (files != null && files.length > 0) {
-        for (File f : files) {
-          delete(f);
-        }
-      }
-      file.delete();
-    }
-  }
-}
+///*
```
felixcheung: some info on this perhaps for reviewers ? why is this commented out?

liuxunorg:
This improvement will be very helpful. To solve the problem of a large number of notes, The `notebook-authorization.json` file is too large for performance issues. 👍

zjffdu:
@felixcheung Could you help review it ? Thanks

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
```diff
@@ -0,0 +1,249 @@
...
+
+  private ZeppelinConfiguration conf;
+  private Notebook notebook;
+  /*
+   * contains roles for each user
+   */
+  private Map<String, Set<String>> userRoles = new HashMap<>();
+
+  @Inject
+  public AuthorizationService(Notebook notebook) {
```
jongyoul: It would be better to have `ZeppelinConfiguration` injected as a parameter of this constructor.

felixcheung: hmm.. why does AuthorizationService need a Notebook? it sounds a bit odd

zjffdu: @felixcheung AuthorizationService needs to use Notebook to fetch Note and set permission on it.  https://github.com/apache/zeppelin/pull/3316/files#diff-97c5f98883ecf7a1eb2dadcfaa684496R68
@jongyoul I can make `ZeppelinConfiguration` as injected, but Notebook already has reference on it, I don't have strong preference to inject or not.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
```diff
@@ -0,0 +1,249 @@
...
+    return isReader(noteId, userAndRoles);
+  }
+
+  public boolean isPublic() {
+    return conf.isNotebookPublic();
+  }
+
+  public void setRoles(String user, Set<String> roles) {
+    if (StringUtils.isBlank(user)) {
+      LOGGER.warn("Setting roles for empty user");
```
felixcheung: shouldn't this fail?

felixcheung:
one minor comment. LGTM

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
jongyoul: Yes, right. We could use it by calling from `notebook` but to make dependencies of classes clear, I recommend using injections explocitly, and removing `getConf()` in the class of `Notebook`. Before using injection, adding parameters in a constructor is quite bothersome but using injection made it easy.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
zjffdu: This is from the old implementation: `NotebookAuthorization`, I guess it is to handle some corner case when user is empty.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
zjffdu: Thanks @jongyoul I have fixed it

zjffdu:

PR #3321, opened 2019-03-07
[ZEPPELIN-3487]. Spark SQL (%sql) paragraphs for DDLs should suppress datagrid-ui

zjffdu:
### What is this PR for?

This PR will return empty string when the returned DataFrame of sql is empty, but it is not true for spark 1.x, so for spark 1.x, user would still see empty table in frontend.  

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-3487

* CI pass

image

### Questions:

### DIFF
--- a/spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
+++ b/spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
@@ -56,6 +56,10 @@ public class Spark1Shims extends SparkShims {
     if (obj instanceof DataFrame) {
       DataFrame df = (DataFrame) obj;
       String[] columns = df.columns();
+      // DDL will empty DataFrame
+      if (columns.length == 0) {
+        return "";
+      }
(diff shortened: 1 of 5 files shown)


spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
```diff
@@ -56,6 +56,10 @@ public String showDataFrame(Object obj, int maxResult) {
     if (obj instanceof DataFrame) {
       DataFrame df = (DataFrame) obj;
       String[] columns = df.columns();
+      // DDL will empty DataFrame
```
felixcheung: is row count doesnt matter here?

felixcheung: I mean what if row count == 0?

zjffdu: row count doesn't matter. Because if column count is 0, then row count must be 0

PR #3325, opened 2019-03-08
[ZEPPELIN-4030] spark-scala-2.10 won't be built under some profile

zjffdu:
### What is this PR for?
Before this PR, module spark-scala-2.10 is only built under spark profile which support both scala 2.10 and 2.11. But it should not be depend on which version of spark to build with. The spark profile should only affect the embedded spark distribution of spark interpreter, it should support multiple spark version no matter which profile it is built against. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4030

* CI pass

### Questions:


spark/interpreter/pom.xml
```diff
@@ -63,6 +61,12 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.zeppelin</groupId>
+      <artifactId>spark-scala-2.10</artifactId>
+      <version>${project.version}</version>
```
felixcheung: doesn't this break if/when spark 2.4.0 doesn't have a spark-scala-2.10?

zjffdu: spark-scala-2.10 will always compile with spark 2.2 and scala 2.10. The profile only affect the embedded spark version

zjffdu: spark-scala-2.10 will always compile with spark 2.2 and scala 2.10. The profile only affect the embedded spark version

felixcheung: well, ok, but that should go to spark 2.4 at one point? or it should be under a profile so people can choose the version of spark they want in embedded?

zjffdu: People can still choose profile spark-2.4 so that the embedded spark version will be 2.4

spark/spark-scala-parent/pom.xml
```diff
@@ -34,6 +34,11 @@
     <packaging>pom</packaging>
     <name>Zeppelin: Spark Scala Parent</name>
 
+    <properties>
+        <!-- -->
+        <spark.runtime.version>2.2.0</spark.runtime.version>
```
felixcheung: in else where the default  is 2.4.0? should this be also to match?

zjffdu: I changed it to 2.4.0, but actually it doesn't matter which version I specified here. Because type of spark-scala-parent is pom and the spark version will be override by its child module scala-2.10 and scala-2.11.

PR #3326, opened 2019-03-08
[ZEPPELIN-4028]. Pop up error message if user try to create duplicated note

zjffdu:
### What is this PR for?
This PR fix the backend by disallowing create duplicated note. If user try to create duplicate note, it throw pop up one error message in fronend.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4028

* Unit test is added, and also manually verify it

image

### Questions:


felixcheung:
PR title - is that right? doesn't seem to match the content

zjffdu:
@felixcheung What's wrong with the PR title ?

felixcheung:
it looks like you are changing tests and error handling, I guess dup note name is the trigger. how about describing the fix instead? I assume it also handles other cases better?

PR #3333, opened 2019-03-13
[ZEPPELIN-4053]. Implement impersonation via c native api

zjffdu:
### What is this PR for?

This PR use a native c code to implement impersonation. The previous approach use ssh which require user to enable password-less for each user. While this approach is much easier for user. zeppelin admin only needs to compile this c code it to enable impersonation.

### What type of PR is it?
[Improvement | Feature | Documentation ]

### Todos
* [ ] - Task

### What is the Jira issue?

* Tested manually on shell, spark and python interpreter

image

### Questions:


liuxunorg:
For security, YARN uses the yarn user group to execute `/etc/yarn/sbin/Linux-amd64-64/container-executor`, which requires special settings for the `container-executor` file.

1. Running zeppelin with root can switch users, But this is not safe enough.

2. So need to create a zeppelin user group in the operating system (for example: zeppelin-group). The user running the zeppelin service belongs to this group zeppelin-group.
After performing the following operations on the file using the root account, the execution user of the zeppelin service can also switch users.

```
chown root:${zeppelin-group} zeppelin/bin/execute-as-user
chmod 6050 zeppelin/bin/execute-as-user
```

So need to add some instructions for use.

liuxunorg:
@zjffdu , I submit a PR in your branch（https://github.com/zjffdu/zeppelin/pull/6）,
Add one to the Zeppelin plugin User-impersonation native module,
Can be compiled when zeppelin, Compile this module with the `mvn -Pnative` parameter
Please review and merge that PR.

PR #3335, opened 2019-03-20
[MINOR] Refactor CronJob class

jongyoul:
### What is this PR for?
Making cronjob disabled by configuration clearly. This is the first step to clear code only. In the later step, I'll make a PR because it might not be that simple.

### What type of PR is it?
[Refactoring]

### Todos
* [X] - Removed uncleared code

### What is the Jira issue?
* No

* CI

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -32,13 +32,11 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
-import org.apache.commons.lang.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.conf.ZeppelinConfiguration.ConfVars;
 import org.apache.zeppelin.display.AngularObject;
 import org.apache.zeppelin.display.AngularObjectRegistry;
 import org.apache.zeppelin.interpreter.Interpreter;
-import org.apache.zeppelin.interpreter.InterpreterException;
 import org.apache.zeppelin.interpreter.InterpreterFactory;
 import org.apache.zeppelin.interpreter.InterpreterGroup;
 import org.apache.zeppelin.interpreter.InterpreterNotFoundException;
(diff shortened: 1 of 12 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -414,7 +423,8 @@ public Note loadNoteFromRepo(String id, AuthenticationInfo subject) {
         for (AngularObject object : objectList) {
           SnapshotAngularObject snapshot = angularObjectSnapshot.get(object.getName());
           if (snapshot == null || snapshot.getLastUpdate().before(lastUpdatedDate)) {
-            angularObjectSnapshot.put(object.getName(),
+            angularObjectSnapshot.put(
```
zjffdu: Are these changes caused by IDE automatically ?

jongyoul: Yes, it is.

zjffdu: Could you remove these changes caused by IDE ? Otherwise it is difficult to track the commit history.

jongyoul: I don't set any special things except setting google-java-format. Could you try to do it with the same setting?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -64,11 +65,12 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Sets;
```
felixcheung: tbh we keep moving these imports in various PRs. should we just adopt a style that could work consistently in IDE?

jongyoul: Yes, for that kind of code style issue, I've talked with @zjffdu and concluded not to change this time. :-) Thanks for letting me know that.

felixcheung: yes, but we just need to all use the same style, then it will stop moving

jongyoul: You’re right. Someday we need to adopt same style at a time. And I’m worried about that code style issue always become low priority because it affected a lot but there’s no new feature. It’s a bit hard to make consensus for that reason. :-)

jongyoul:
To review easily, this PR prevent from running quartz thread even if we don't set and use scheduler features. Applying this PR will lead a quartz schedule to be initialized only if we enable this feature in a configuration.

jongyoul:
CI passed: https://travis-ci.org/jongyoul/zeppelin/builds/511850669

jongyoul:
Will merge it if there's no more discussion

jongyoul:
Rebased. Will merge it

PR #3337, opened 2019-03-20
[ZEPPELIN-4090] Ipython queue performance

AyWa:
### What is this PR for?

The pr is to fix a bug that will make the **ipython** queue listener, overuse cpu. After this fix, cpu usage should be way lower. 
Also there is a bit of refactor to use only one queue to ensure message will be order even with a sleep.
Also, the refactor is reducing the number of line and code duplication

### What type of PR is it?
Bug Fix / performance improvement

### Todos
* [x] - Performance improvement

### What is the Jira issue?
It is one part of the jira issue.
https://issues.apache.org/jira/browse/ZEPPELIN-4090

### Questions:

### DIFF
--- a/python/src/main/resources/grpc/python/ipython_server.py
+++ b/python/src/main/resources/grpc/python/ipython_server.py
@@ -122,7 +122,7 @@ class IPython(ipython_pb2_grpc.IPythonServicer):
             if not self.isKernelAlive() or t.is_alive():
                 yield ipython_pb2.ExecuteResponse(status=ipython_pb2.ERROR,
                                                   type=ipython_pb2.TEXT,
-                                                  output="Ipython kernel has been stopped. It might be because of an out of memory issue")
+                                                  output="Ipython kernel has been stopped. Please check logs. It might be because of an out of memory issue.")
         if payload_reply:
             result = []
             for payload in payload_reply[0]['content']['payload']:
--- a/python/src/main/resources/grpc/python/ipython_server.py
+++ b/python/src/main/resources/grpc/python/ipython_server.py
@@ -53,38 +53,45 @@ class IPython(ipython_pb2_grpc.IPythonServicer):
         print("execute code:\n")
         print(request.code.encode('utf-8'))
         sys.stdout.flush()
-        stderr_queue = queue.Queue(maxsize = 10)
-        text_queue = queue.Queue(maxsize = 10)
-        png_queue = queue.Queue(maxsize = 5)
-        jpeg_queue = queue.Queue(maxsize = 5)
-        html_queue = queue.Queue(maxsize = 10)
-
+        stream_reply_queue = queue.Queue(maxsize = 20)
+        payload_reply = []
         def _output_hook(msg):
             msg_type = msg['header']['msg_type']
             content = msg['content']
             print("******************* CONTENT ******************")
             print(str(content)[:400])
+            outstatus, outtype, outtype = ipython_pb2.SUCCESS, None, None
             if msg_type == 'stream':
-                text_queue.put(content['text'])
-            elif msg_type in ('display_data', 'execute_result'):
-                if 'text/html' in content['data']:
-                    html_queue.put(content['data']['text/html'])
-                elif 'image/png' in content['data']:
-                    png_queue.put(content['data']['image/png'])
-                elif 'image/jpeg' in content['data']:
-                    jpeg_queue.put(content['data']['image/jpeg'])
-                elif 'text/plain' in content['data']:
-                    text_queue.put(content['data']['text/plain'])
-                elif 'application/javascript' in content['data']:
-                    print('add to html queue: ' + str(content)[:100])
-                    html_queue.put('<script> ' + content['data']['application/javascript'] + ' </script>\n')
-                elif 'application/vnd.holoviews_load.v0+json' in content['data']:
-                    html_queue.put('<script> ' + content['data']['application/vnd.holoviews_load.v0+json'] + ' </script>\n')
-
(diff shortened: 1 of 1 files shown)


python/src/main/resources/grpc/python/ipython_server.py
```diff
@@ -52,24 +52,19 @@ def execute(self, request, context):
...
-                stdout_queue.put(content['data'].get('text/plain', ''))
+                stream_reply_queue.put(ipython_pb2.ExecuteResponse(status=ipython_pb2.SUCCESS, type=ipython_pb2.TEXT, output=content['data'].get('text/plain', '')))
                 if 'image/png' in content['data']:
-                    image_queue.put(content['data']['image/png'])
-            elif msg_type == 'error':
-                stderr_queue.put('\n'.join(content['traceback']))
-
-
-        payload_reply = []
+                    stream_reply_queue.put(ipython_pb2.ExecuteResponse(status=ipython_pb2.SUCCESS, type=ipython_pb2.IMAGE, output=content['data']['image/png']))
```
felixcheung: this is a bit long and hard to read? consider refactor `stream_reply_queue.put` into a separate line

python/src/main/resources/grpc/python/ipython_server.py
```diff
@@ -52,24 +52,19 @@ def execute(self, request, context):
         print("execute code:\n")
         print(request.code.encode('utf-8'))
         sys.stdout.flush()
-        stdout_queue = queue.Queue(maxsize = 10)
-        stderr_queue = queue.Queue(maxsize = 10)
-        image_queue = queue.Queue(maxsize = 5)
-
+        stream_reply_queue = queue.Queue(maxsize = 20)
```
felixcheung: should maxsize be a bit more?

python/src/main/resources/grpc/python/ipython_server.py
AyWa: Sure let me update that.

python/src/main/resources/grpc/python/ipython_server.py
AyWa: What about 30 ? 
The queue is blocking if full, so should be fine isnt ?

felixcheung: ok, what happens when the queue is full?

AyWa: When the queue is full , Add element to the queue is a blocking call, so message will just be delay. 
It could be also related to how the `execute_interactive` from `jupyter_client` is implemented. But there is no timeout so far: https://github.com/jupyter/jupyter_client/blob/44980c13680f4e4226cf25f199ce4e4bb6e11296/jupyter_client/blocking/client.py#L338
and they are using queue under the hood too. So even with many message to handle, it will just bring some slow down.

(relative slow down, because `while not stream_reply_queue.empty():` is likely to never stop when there is a lot of message to process.) So thats why there is at most `0.05sc` delay compare to before. (but with less cpu usage, so likely to be faster at the end)

AyWa:
Will try to rebase and fix this pr in the coming week :)

zjffdu:
ping @AyWa

AyWa:
@zjffdu thank you to ping me, working on it and just rebase and fixed the new conflict. 
I will ping you when test are passing

AyWa:
@zjffdu it was a bit longer than I expected, because I failed to rebased properly, but now everything is alright and ready to be reviewed.

I think all test are passing now https://travis-ci.org/AyWa/zeppelin/builds/583492141 (only one is still in progress but should be alright)

python/src/main/resources/grpc/python/ipython_server.py
```diff
@@ -94,69 +105,24 @@ def execute_worker():
...
-
-        # if kernel is not alive (should be same as thread is still alive), means that we face
-        # an unexpected issue.
-        if not self.isKernelAlive() or t.is_alive():
-            yield ipython_pb2.ExecuteResponse(status=ipython_pb2.ERROR,
-                                                type=ipython_pb2.TEXT,
-                                                output="Ipython kernel has been stopped. Please check logs. It might be because of an out of memory issue.")
-            return
-
-        while not text_queue.empty():
```
zjffdu: Why removing these ? Will this cause data lose if this queue is not empty ?

AyWa: So the refactor I did, move all the logic in a single loop condition. So we do not need to have multiple time the similar code (in the main loop and outside of it)

It is line https://github.com/apache/zeppelin/pull/3337/files#diff-9abe0e14503875941f3276b5f7743c46R113

until the queue is not empty we will not leave the main loop: `while (t.is_alive() and self.isKernelAlive()) or not stream_reply_queue.empty():
 `

zjffdu: Thanks for the explanation, the changes LGTM

AyWa:
test are successful https://travis-ci.org/AyWa/zeppelin/builds/583492141

zjffdu:
Awesome, will merge it soon.

zjffdu:
Another favor needs from you @AyWa Could you check whether you configure your travis according this guideline ? Because I don't see your travis build link here. 
https://zeppelin.apache.org/contribution/contributions.html#continuous-integration

AyWa:
So I double check, it really seems to be correct, next pr I made I will ensure the CI is shown in the pr.

zjffdu:
That's fine, let's look at this issue next time.

AyWa:
Thank you.

PR #3339, opened 2019-03-21
[ZEPPELIN-4089] handle ipython kernel crash

AyWa:
### What is this PR for?

The pr is the last fix of the `ZEPPELIN-4089` It will aims to notify user when ipython kernel crash. Indeed, right now, execution will be stuck forever when ipython kernel crash.

### What type of PR is it?
Bug Fix

### Todos
* [x] - logic to check kernel status 
* [ ] - decide what to do when we spot the failure
* [x] - test

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4089

### Questions:


python/src/main/resources/grpc/python/ipython_server.py
```diff
@@ -146,10 +160,16 @@ def serve(port):
     server.start()
     ipython.start()
     try:
-        while True:
-            time.sleep(_ONE_DAY_IN_SECONDS)
+        while ipython.isKernelAlive():
```
AyWa: We have multiple options when we figure out that the kernel die:
- return an error then kill the process (this pr is doing that)
- restart the kernel ?

The challenge with 1, is that we maybe need to add a new status to the `status` grpc method. like `shutdown`. We can check this status before running like in the `checkIpythonProcessAndRestartIfDead` 

The challenge with 2 is that our execute thread will always be stuck. and we might have no way to kill it.

zjffdu:
@AyWa Do you know how to create such scenario (ipython server not die but ipython kernel die) ?

AyWa:
Yeah an out of memory will do the trick. I Could reproduce with a docker running the ipython server with a memory limit. I will push it tomorrow.

AyWa:
@zjffdu Hi, 
I just push an example with docker. https://github.com/apache/zeppelin/commit/c9ec0335cd2fd93476cee31ba99b6dc157face0d

So if you run that:
in `zeppelin/python/src/main/resources/grpc` run `docker build -t my-python-app . && docker run -m 1g --memory-swap 1g -p 50053:50053 my-python-app`
It will install the dependencies of the zeppelin ipython server, and then run the grpc server on port `50053`. And we put a docker limit of `1g` so we can test the `OOM`
Then run in `zeppelin/python/src/main/resources/grpc`:
`python3 python/ipython_client_OOM.py`

With this pr you should see something like: (both client and server terminate)
```
output:14

output:15

output:16

output:17

output:18

output:Ipython kernel has been stopped. Please check logs. It might be because of an out of memory issue.
status: ERROR
output: "Ipython kernel has been stopped. Please check logs. It might be because of an out of memory issue."
```

On master you will have: (both client and server hanging)
```
output:18
```

zjffdu:
Thanks @AyWa , but it seems not easy to do unit test. I find this post which might be helpful. Do you mind to try this ? https://stackoverflow.com/questions/16779497/how-to-set-memory-limit-for-thread-or-process-in-python

AyWa:
Yeah indeed, it require docker to do so :/

Before using docker I tried to so similar things but without success.  I think the problem is that the memory issue is coming from the ipython kernel. (so it will not be the same process as far as I understand and I didnt find option to set memory limit in the kernel)

However let me try with the link you shared.

AyWa:
Just think that maybe we could just kill the ipython kernel, like finding pid. Might be easier for unit test ?

AyWa:
@zjffdu Hi, 
I just found a way to test it: https://github.com/apache/zeppelin/pull/3339/commits/f37d5f95dcaab07fdab08d9e1f03265896d02e68

I find the kernel pid and kill it from the "running" paragraph.

the test succeed on this branch but hang / fail on master. 

This is the only way I found (without docker)

zjffdu:
Thanks @AyWa The unit test is very helpful

zjffdu:
@AyWa Before I merge it, could you create a dedicated ticket for this PR. I found you have 2 PRs for the same ticket ZEPPELIN-4078

AyWa:
Yeah sure. I will create subtask for my 3 pr. (for now they all share the same jira issue )

AyWa:
@zjffdu updated the jira for this pr.

PR #3341, opened 2019-03-25
ZEPPELIN-4038. Deprecate spark 2.2 and earlier

zjffdu:
### What is this PR for?
This PR is to deprecate spark 2.2 and earlier. It would display a warning message in frontend, and user can disable this message by configure `zeppelin.spark.deprecatedMsg.show` to false, by default it is true.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4038

* CI pass

image

### Questions:


spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
```diff
@@ -178,4 +184,20 @@ public static String getUserName(AuthenticationInfo info) {
...
+                                            InterpreterContext context,
+                                            Properties properties) throws InterpreterException {
+    context.out.clear();
+    if (sparkVersion.olderThan(SparkVersion.SPARK_2_2_0)
+            && Boolean.parseBoolean(
+                    properties.getProperty("zeppelin.spark.deprecatedMsg.show", "true"))) {
+      try {
+        context.out.write(DEPRRECATED_MESSAGE);
+        context.out.write("%text ");
+      } catch (IOException e) {
```
felixcheung: kinda odd to get IOException here?

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -95,7 +95,6 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
         System.setOut(Console.out)
         interpreterOutput.setInterpreterOutput(context.out)
         interpreterOutput.ignoreLeadingNewLinesFromScalaReporter()
-        context.out.clear()
```
felixcheung: is there any other cases of `context.out.clear()`?

zjffdu: I move the clear into other places. https://github.com/apache/zeppelin/pull/3341/files#diff-25ff02ea9a7b4a31efaf02e8b401ea5bR191
Otherwise, the deprecated warning message will disappear after the paragraph is finished.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/Utils.java
zjffdu: Indeed, but I have to catch this message here.

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
felixcheung: ah, but I meant if there might be other calls to clear that will also undo printDeprecateMessage

zjffdu:
Thanks @felixcheung `spark.md` is updated

Leemoonsoo:
I've seen companies still depends on spark 1.6 and spark 2.2. And guess many of them can not upgrade spark easily.

How about make longer term plan of spark support, depreciation, and publish it on website or documentation? e.g. Support last N spark releases. Deprecates after X.
I think sharing long term plan will help users to have the right expectations on using Zeppelin with Spark.

zjffdu:
Thanks @Leemoonsoo , it's a good point, my initial plan is to support them before zeppelin 1.0
The final not-supported date is not determined yet.

felixcheung:
that's a fair point. maybe we should keep for 0.9 and 0.8 and remove only in 1.0

PR #3342, opened 2019-03-25
[ZEPPELIN-4031] Fixed Unable to detect that the interpreter process was killed manually

liuxunorg:
### What is this PR for?
Zeppelin-server can't perceive, When a network exception occurs Or a state where a program exception causes the interpreter process to be unavailable, Cause the user to fail to perform the task all the time. You need to restart the interpreter process to get the user's interpreter back to normal, Cause a bad user experience.

By detecting the state of the remote interpreter process in Zeppelin server, When an interpreter process exception is found, By cleaning up the session of this interpreter, Let the interpreter regain its availability, Improves the user experience, It also reduces the operation and maintenance burden of the system.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4031

```
RemoteInterpreterTest::testDetectIntpProcessKilled()
```

#### **Before bug fix**
4031-BUG

#### **After bug fix**
4031-Fixed

### Questions:


liuxunorg:
@zjffdu, @felixcheung , @jongyoul , Please help me review the code, Thanks!

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -103,7 +103,7 @@ public String getSessionId() {
   }
 
   public synchronized RemoteInterpreterProcess getOrCreateInterpreterProcess() throws IOException {
-    if (this.interpreterProcess != null) {
+    if (this.interpreterProcess != null && interpreterProcess.isRunning()) {
       return this.interpreterProcess;
     }
     ManagedInterpreterGroup intpGroup = getInterpreterGroup();
```
felixcheung: before creating a new one, if `this.interpreterProcess != null` should we close it or something even when it is not running now?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
```diff
@@ -141,17 +165,27 @@ private void close(Collection<Interpreter> interpreters) {
...
     Scheduler scheduler = interpreter.getScheduler();
 
-    for (final Job job : scheduler.getAllJobs()) {
-      job.abort();
-      job.setStatus(Job.Status.ABORT);
-      LOGGER.info("Job " + job.getJobName() + " aborted ");
-    }
+    // Need to abort the task being executed
+    // when actively shutting down the remote interpreter
+    if (false == remoteIntpProcessIsShutdown()) {
```
felixcheung: this isn't quite I'd expected... if the process is created but it's not running, yes it could be being shutdown else where but it could also be that it was killed or crashed or something. in the latter case, shouldn't it go ahead to abort the jobs anyway?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
```diff
@@ -58,10 +58,19 @@ public InterpreterSetting getInterpreterSetting() {
   public synchronized RemoteInterpreterProcess getOrCreateInterpreterProcess(String userName,
                                                                              Properties properties)
       throws IOException {
+    if (remoteIntpProcessIsShutdown()) {
+      LOGGER.info("Check whether the InterpreterProcess has been shutdown.");
+      // clean invalid session and dirty data of interpreterSetting
+      close();
```
felixcheung: same q about whether to close iff process is not running

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
liuxunorg: The original code is not perfect.
Because `RemoteInterpreter.java::getOrCreateInterpreterProcess()` is called, an available RemoteInterpreter process must be returned.

`interpreterProcess.isRunning()` Will call `RemoteInterpreterUtils.java::checkIfRemoteEndpointAccessible()` Check if the remote interpreter is available through the socket.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
liuxunorg: `remoteIntpProcessIsShutdown()==true`, Found that the remote interpreter process is unavailable, clean invalid session by close() function.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
liuxunorg: ### 1. If interpreter process is normal, isRunning() must equal true
`isRunning()` is not a flag, It's a function call.
`interpreterProcess.isRunning()` Will call `RemoteInterpreterUtils.java::checkIfRemoteEndpointAccessible()` , Check if the remote interpreter is available through the socket.

### 2. Avoid invalid calls
`Job.abort()` Will go to the remote to call the interpreter, `Interpreter.cancel(getInterpreterContext(null));` function. If the interpreter is no longer connected, The `interpreter.cancel()` function cannot be called correctly.

But if you find that the remote interpreter is not available in `job.abort()`, Will pass again
`RemoteInterpreter.java::getOrCreateInterpreterProcess()`, Trying to create an interpreter process, This will fail due to a 30 second timeout.

liuxunorg:
This bug, It's not easy to get through the code review to understand the situation.
Better way, It is verified by testing.

felixcheung:
ok thanks for explaining. im ok with this. IMO might be worthwhile to refactor the code to make it more straightforward perhaps?

liuxunorg:
### No need to click the second time to execute successfully.  Restore the interpreter process immediately
4031-Fixed-2

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -199,13 +199,31 @@ public Void call(Client client) throws Exception {
...
+
   @Override
   public InterpreterResult interpret(final String st, final InterpreterContext context)
       throws InterpreterException {
     if (LOGGER.isDebugEnabled()) {
       LOGGER.debug("st:\n{}", st);
     }
 
+    // Detecting an invalid interpreter process
+    resurrectionInvalidIntpProcess();
```
liuxunorg: Every time the user executes a note, Called
`RemoteInterpreter.java::interpret(final String st, final InterpreterContext context)` function,
Check if the remote interpreter process is valid, If the remote interpreter is invalid,
In the `RemoteInterpreter.java`, It is very easy to perform session recycling for the interpreter process,
And recreate the interpreter. Conforms to zeppelin's original process of creating an interpreter.

liuxunorg:
CI Pass https://travis-ci.org/liuxunorg/zeppelin  
@zjffdu , @felixcheung , Please help me review code.
In the second commit, I adjusted the way to recreate the invalid remote interpreter process. It is now very elegant to fix remote invalid interpreters, It can be used immediately without any manual intervention. 
See the GIF effect: https://github.com/apache/zeppelin/pull/3342#issuecomment-477046336

PR #3343, opened 2019-03-25
[ZEPPELIN-4063] Don't include noteId for constructing Interpreter GroupId when under isolated per user mode

zjffdu:
### What is this PR for?
It is a straightforward fix to correct the groupId in different isolated mode.  

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4063

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
@@ -149,7 +149,7 @@ public class InterpreterOption {
     return ISOLATED.equals(perNote);
   }
 
-  public boolean isProcess() {
+  public boolean isIsolated() {
     return perUserIsolated() || perNoteIsolated();
   }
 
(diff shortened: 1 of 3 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/InterpreterOption.java
```diff
@@ -149,7 +149,7 @@ public boolean perNoteIsolated() {
     return ISOLATED.equals(perNote);
   }
 
-  public boolean isProcess() {
+  public boolean isSolated() {
```
felixcheung: should `isSolated` be `isIsolated`? is that what you mean?

zjffdu: Correct, it's a typo, I have fixed it

zjffdu:
Will merge if no more comment

PR #3344, opened 2019-03-31
[ZEPPELIN-4102] fix "getJSONObject()" to "optJSONObject()" to avoid crash

bd2019us:
### What is this PR for?
Fix [#ZEPPELIN-4102](https://issues.apache.org/jira/browse/ZEPPELIN-4102), which changes "getJSONObject()" to "optJSONObject()" to avoid program crash.

### What type of PR is it?
[Bug Fix]


liuxunorg:
@bd2019us , thanks for the contribution! :-)
The next time you mention PR, Please add a point to solve the problem description. This will help others review the code. And merge it in as soon as possible.

### When the return value is not a JSONObject object, the return value is null, and no exception is thrown;
```
public JSONObject optJSONObject(String name) {
  Object object = opt(name);
  return object instanceof JSONObject ? (JSONObject) object : null;
}
```

### Throws an exception when the return value is not a JSONObject object;
```
public JSONObject getJSONObject(String name) throws JSONException {
        Object object = get(name);
        if (object instanceof JSONObject) {
            return (JSONObject) object;
        } else {
            throw JSON.typeMismatch(name, object, "JSONObject");
        }
    }
```

liuxunorg:

PR #3346, opened 2019-04-02
[ZEPPELIN-4066]. Introduce ProcessLauncher to encapsulate process launch

zjffdu:
### What is this PR for?
There're several places in zeppelin to launch processes, such as interpreter process launch, python process launch.  Sometimes it is very hard to diagnose if the process is failed to launch or some error happens during its lifecycle. This PR introduce ProcessLauncher to encapsulate process launch. ProcessLauncher will do most of the process launching stuff but also allow for customization. E.g. how to check the process is started properly. And when error happens, ProcessLauncher will capture the necessary error output for further diagnose. 

### What type of PR is it?
[ Improvement | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4066

* Ci pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/util/ProcessLauncher.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/util/ProcessLauncher.java
@@ -43,7 +43,8 @@ public abstract class ProcessLauncher implements ExecuteResultHandler {
     NEW,
     LAUNCHED,
     RUNNING,
-    TERMINATED
+    TERMINATED,
+    COMPLETED
   }
 
   private CommandLine commandLine;
@@ -68,17 +69,22 @@ public abstract class ProcessLauncher implements ExecuteResultHandler {
     executor.setWatchdog(watchdog);
     try {
       executor.execute(commandLine, envs, this);
-      state = State.LAUNCHED;
+      transition(State.LAUNCHED);
       LOGGER.info("Process is launched: {}", commandLine);
     } catch (IOException e) {
       this.processOutput.stopCatchLaunchOutput();
       LOGGER.error("Fail to launch process: " + commandLine, e);
-      state = State.TERMINATED;
+      transition(State.TERMINATED);
       errorMessage = e.getMessage();
     }
   }
 
   public abstract void waitForReady(int timeout);
+
+  public void transition(State state) {
+    this.state = state;
+    LOGGER.info("Process state is transitioned to " + state);
+  }
 
   public void onTimeout() {
     LOGGER.warn("Process launch is time out.");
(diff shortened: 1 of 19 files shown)


zjffdu:
@felixcheung @liuxunorg Could you help review it ? Thanks

zeppelin-interpreter/src/main/java/org/apache/zeppelin/util/ProcessLauncher.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Map;
+
+/**
+ * Abstract class for launching java process.
+ */
+public abstract class ProcessLauncher implements ExecuteResultHandler {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ProcessLauncher.class);
+
+  public enum State {
```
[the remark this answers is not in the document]
zjffdu: timeout is not a state. Process won't be in timeout state, because once we detect it is timeout, we will kill the process and move it to EXITED.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
```diff
@@ -235,52 +186,64 @@ public boolean isUserImpersonated() {
...
+    }
 
-      if (out != null) {
+    @Override
+    public void onProcessComplete(int exitValue) {
+      LOGGER.warn("Process is exited with exit value " + exitValue);
+      // For yarn-cluster mode, client process will exit with exit value 0
+      // after submitting spark app. So don't move to EXITED state when exitValue is 0.
+      if (exitValue != 0) {
+        state = State.EXITED;
```
felixcheung: so EXITED is unexpected termination? perhaps this should be TERMINATED?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
```diff
@@ -235,52 +186,64 @@ public boolean isUserImpersonated() {
...
-        synchronized (this) {
-          if (out != null) {
-            out.write(b);
+    public void waitForReady(int timeout) {
+      synchronized (this) {
+        if (state != State.RUNNING) {
+          try {
+            wait(timeout);
+          } catch (InterruptedException e) {
+            LOGGER.error("Remote interpreter is not accessible");
```
felixcheung: LOGGER.error("". e)

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
```diff
@@ -171,38 +140,20 @@ public Void call(RemoteInterpreterService.Client client) throws Exception {
           }
         });
       } catch (Exception e) {
-        logger.warn("ignore the exception when shutting down");
+        LOGGER.warn("ignore the exception when shutting down");
```
felixcheung: I think we should still log the error LOGGER.warn("", e)

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
zjffdu: Make sense, I have renamed it.

felixcheung:
I think one of comment might have disappear: should there be a TERMINATED state and a COMPLETED state to represent respectively successful, and unexpected ?

(btw, I think it will much easier to review if you don't force push..)

zjffdu:
Thanks for the review and suggestion, I have addressed the comments

zjffdu:
Will merge if no more comment

PR #3355, opened 2019-04-25
[ZEPPELIN-4133]. Idle sessions are no longer being closed even though TimeoutLifecycleManagement is configured properly

zjffdu:
### What is this PR for?
The root cause is that we use Timer to scheduler periodical interpreter checking task. But java Timer has one critical issue that once the Timer thread is crashed, subsequent task will be suppressed. This PR use ScheduledExecutorService instead to fix this issue. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4133

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
@@ -7,9 +7,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 
 /**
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
```diff
@@ -30,29 +32,30 @@
...
           Long lastTimeUsing = entry.getValue();
-          if ((now - lastTimeUsing) > timeoutThreshold )  {
+          if ((now - lastTimeUsing) > timeoutThreshold) {
             LOGGER.info("InterpreterGroup {} is timeout.", interpreterGroup.getId());
             interpreterGroup.close();
             interpreterGroups.remove(entry.getKey());
           }
         }
+      } catch (Exception e) {
+        LOGGER.warn("Fail to run periodical checking task", e);
```
felixcheung: perhaps put try catch around `interpreterGroup.close();`? it looks to me like other calls inside should not fail in most cases. if `close()` is expensive or could fail, then a try catch just around it could let other groups a chance to get clean up, instead of skipping all of the rest this is doing now.

zjffdu: Make sense to me. I updated it.

felixcheung:
LG

zjffdu:

PR #3357, opened 2019-04-29
[ZEPPELIN-4136] -Use a match expression to determine SparkR Security

RussellSpitzer:
### What is this PR for?
Previously the code dependend on hard coded Spark Versions, this made
some distributions of Spark incompatible because they included security
backports not listed in the official distribution versions. This instead
checks the class itself to determine whether or not the new Secret
method should be utilized

### What type of PR is it?
[Bug Fix | Refactoring]

### Todos
Change from version check to Match

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4136

Gonna try out CI

### Questions:
* Does the licenses files need update? Nope
* Is there breaking changes for older versions? Nope
* Does this needs documentation? Nope

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
@@ -50,8 +50,7 @@ public class IPySparkInterpreter extends IPythonInterpreter {
         getInterpreterInTheSameSessionByClassName(PySparkInterpreter.class, false);
     setProperty("zeppelin.python", pySparkInterpreter.getPythonExec());
     sparkInterpreter = getInterpreterInTheSameSessionByClassName(SparkInterpreter.class);
-    setProperty("zeppelin.py4j.useAuth",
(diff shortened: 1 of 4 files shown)


zjffdu:
Thanks for the contribution @RussellSpitzer , could you remove SparkVersion#isSecretSocketSupported ?

RussellSpitzer:
Of course

felixcheung:
this is the same for pyspark, right?

RussellSpitzer:
@felixcheung From what I can tell, pyspark just assumes that the secret is always passed https://github.com/apache/zeppelin/blob/8fa534343beccb797f9f1850ec43673ac830b73e/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java#L121

RussellSpitzer:
Not sure why i'm having such pain with this type matching :/

RussellSpitzer:
@felixcheung I read the code wrong, to avoid the version check for Python I have it check whether py4j supports the auth token method.

RussellSpitzer:
I'm really not liking what i'm doing here. I think i'm going to try to just push the version number in our distribution.

PR #3359, opened 2019-05-13
[ZEPPELIN-4131] Add %sh.terminal interpreter

liuxunorg:
### What is this PR for?
Realize the same function as the terminal, The shell-by-command can be executed by the shell interpreter.

《Refactoring zeppelin shell interpreter Design Document》 : https://docs.google.com/document/d/1-XUkXY06cMtEUBxVHVyqQPkyCkv2BjYhAQj-W1CBxNI/edit

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4131

* [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/531594343)

shell-pr

### Questions:
* Does this needs documentation? YES


liuxunorg:
@zjffdu , @felixcheung , Please help me review code. Thanks! :-)

zjffdu:
@liuxunorg Is it possible to download hterm_all.js at building time instead of put into repo ? And could you write a simple design doc for it since it is a redesign of shell interpreter.

liuxunorg:
@zjffdu , 《Refactoring zeppelin shell interpreter Design Document》 : https://docs.google.com/document/d/1-XUkXY06cMtEUBxVHVyqQPkyCkv2BjYhAQj-W1CBxNI/edit

liuxunorg:
@zjffdu , `hterm_all.js` is exported via the chromium libapps project. `hterm_all.js` relies on several submodules in chromium libapps. `hterm_all.js` cannot be downloaded directly.
Therefore, Need to perform the following operations in the official documentation to generate `hterm_all.js`.

## Get the code

The “libapps” git repository contains hterm and its dependencies. Clone this repo into a parent directory of your choice. In this example we'll create `~/src/libapps/`:

```
$ mkdir ~/src
$ cd ~/src
$ git clone https://chromium.googlesource.com/apps/libapps
```

## Build hterm

The build process bundles some resources as JavaScript source and concatenates the JavaScript into a single file (`hterm_all.js`).

```
$ cd ~/src/libapps
$ ./hterm/bin/mkdist.sh
```

The files are all written to `./hterm/dist/js/`. Copy the `hterm_all.js` file into your project.

shell/src/main/resources/ui_templates/terminal-dashboard.jinja
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-#}
+<div class="row" style="width:100%">
+  <div class="col-md-12" ng-switch="TERMINAL_SOCKET_STATUS">
+    <b>Terminal - {{ HOST_NAME }} ({{ HOST_IP }})</b>
+    <b style="color: #999; margin-left:10px;">STATUS: </b>
+    <i ng-switch-when="TERMINAL_SOCKET_CONNECT" ng-style="{'color':'green'}">CONNECT</i>
```
Tagar: CONNECTED?

shell/src/main/resources/ui_templates/terminal-dashboard.jinja
```diff
@@ -0,0 +1,40 @@
...
+    <i ng-switch-when="TERMINAL_SOCKET_CLOSE" ng-style="{'color':'red'}">DISCONNECT</i>
```
Tagar: disconnectED

Tagar:
Very cool.
Perhaps this should be a new interpreter called Terminal? 
And leave Shell interpreter as is?

shell/src/main/resources/ui_templates/terminal-dashboard.jinja
liuxunorg: Ok, I am correcting it.

shell/src/main/resources/ui_templates/terminal-dashboard.jinja
liuxunorg: Ok, I am correcting it.

liuxunorg:
@Tagar This is a good question, @zjffdu  Is there any special use of the shell interpreter that needs to be retained?

Leemoonsoo:
I can think a use case, such as

```
%sh
# download some data
curl -O http://....

# extract
tar -xzf ....
```

```
%python

# load downloaded data
df = pd.read_csv("...")
...
```

This case, %sh interpreter may better have a script to run, rather than expecting interactive user input.
Is this new %sh interpreter implementation capable of doing this as well?

FireArrow:
We are using %sh pretty much exactly like @Leemoonsoo described it above. I think this is a really cool feature, but I don't think it should replace the %sh magic, but rather get its own, new, magic (as @Tagar suggested above). `%terminal` or `%shell` perhaps?

liuxunorg:
# The function is exactly the same as the terminal emulator
## curl + tar ScreenRecord
shell-curl-tar

## python + gcc ScreenRecord
shell-python-gcc

## call script ScreenRecord
shll-script

Tagar:
As part of Zeppelin notes we have static %sh paragraphs (non-interactive) with 
or without variable embeddings (variable interpolations ) like described in 
https://zeppelin.apache.org/docs/0.8.1/interpreter/shell.html#object-interpolation 

Would this new implementation keep that possible? (without typing in a command each time)

I totally agree it's a very cool feature, but we may need to preserve old good `%sh` interpreter
if this new approach takes some of its functionality away. 

Thoughts?

Thank you @liuxunorg

liuxunorg:
@Tagar , this new shell interpreter support `Kerberos auto refresh`, but not support `interpolations`.

I think either use this new interpreter with the `%sh.terminal` keyword? What do you think?

zjffdu:
@liuxunorg Based on the above comments, I think we should keep the old shell interpreter as the batch mode and the new shell interpreter as interactive mode.  Personally I prefer to keep them in one interpreter group so that they can be launched in one JVM and collaborate with each other.

Regarding the `interpolations`, it should already support, as it inherits `AbstractInterpreter` which support `interpolations`.

liuxunorg:
[CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/535725474)  
The changes have been completed in accordance with the https://github.com/apache/zeppelin/pull/3359#issuecomment-492935457.

@zjffdu , Please review the code, Thanks!

sh-terminal

FireArrow:
Nice!
One thought: Is the "any chars" needed for some common guard against running empty paragraphs, or could any chars actually be no chars?
I feel it is a bit confusing for end users to have to provide some text that doesn't matter just to start the interpreter.

liuxunorg:
@FireArrow , Yes, Because the paragraph limits the input you must perform.
So `%sh.terminal` needs to enter a few char to execute.

I also considered this question:
1. Is it necessary to control whether a paragraph must have input to execute?
2. If there is input, it can only be in the interpreter's configuration file.
Add a configuration item that the specified interpreter can execute without entering characters.

Which way is better? I can open another issue to add this feature.

liuxunorg:
#### `%sh.terminal` runs directly, No need to enter any extraneous characters. 
[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/538090348)

terminal

liuxunorg:

shell/pom.xml
```diff
@@ -38,22 +38,79 @@
     <!--library versions -->
     <commons.lang3.version>3.4</commons.lang3.version>
     <commons.exec.version>1.3</commons.exec.version>
-    <interpreter.name>sh</interpreter.name>
+    <pty4j.version>0.9.3</pty4j.version>
+    <jinjava.version>2.4.0</jinjava.version>
+    <guava.version>20.0</guava.version>
   </properties>
 
+  <!-- pty4j library not in travis central (http://repo.maven.apache.org/maven2) -->
```
zjffdu: It is not travis central, it is maven central repository

shell/src/main/java/org/apache/zeppelin/shell/TerminalInterpreter.java
```diff
@@ -0,0 +1,221 @@
...
+      try {
+        terminalPort = RemoteInterpreterUtils.findRandomAvailablePortOnAllLocalInterfaces();
+        terminalThread = new TerminalThread(terminalPort);
+        terminalThread.start();
+      } catch (IOException e) {
+        LOGGER.error(e.getMessage(), e);
+        return new InterpreterResult(Code.ERROR, e.getMessage());
+      }
+
+      for (int i = 0; i < 5 && !terminalThread.isRunning(); i++) {
```
zjffdu: What happens if it is still not running after 5 loops ? We should consider to add timeout mechanism here and allow the timeout threshold configurable.

liuxunorg: Ok, I am adding a timeout error log.
But I want to change 5 to 10, wait a total of 5 seconds, it should be enough. Zeppelin has a lot of configuration items. right?

shell/pom.xml
liuxunorg: Ok, I am going to modify the text description.

liuxunorg:
@felixcheung , Can you tell me how to solve these problems?
> 1. for all dependencies changes, they should listed/updated in LICENSE file

I do not quite understand, Can you tell me how to amend it?

> 2. we can't include .so or any binary files in source repo. This is against release policy in ASF.

I will deal with this problem.

> 3. certain file seems like in category X (eg. GPL) that we, as a policy in ASF, cannot release https://apache.org/legal/resolved.html#category-x

I do not quite understand, Can you tell me how to amend it?

felixcheung:
1. for the new dependencies, in the pom file https://github.com/apache/zeppelin/pull/3359/files#diff-ba629e2dbade35aef7a022f6bc745bc1 check their license (see the source repo) against  https://apache.org/legal/resolved.html#category-x

2. .so, .exe, .dll etc cannot be in the git source repo...

3. I think this was largely the binaries

Tagar:
@felixcheung 

`libpty` is distributed under EPL license 
https://github.com/JetBrains/pty4j#license

It falls into "WEAK COPYLEFT" LICENSES according to the link you posted above
https://apache.org/legal/resolved.html#weak-copyleft-licenses 

So it's a `Category B` open-source software 
https://apache.org/legal/resolved.html#category-b

And according to the same ASF link it has to be included in binary and not source code form -
```
Unless otherwise specified, all Category B licensed works should be included 
in binary-only form in Apache Software Foundation 
convenience binaries (and not source code).
```

zjffdu:
Thanks for the info. @Tagar 
I think we could include the binary distribution of `libpty` as long as meet the `APPROPRIATELY LABELLED CONDITION` by declaring it clearly in `README` file.

liuxunorg:
I add License description & add download-libpyt shell script, [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/541561476)

Please help me review, Thank you!

liuxunorg:

felixcheung:
To be clear the ASF page clearly states:

BINARY-ONLY INCLUSION CONDITION
Unless otherwise specified, all Category B licensed works should be included in binary-only form in Apache Software Foundation convenience binaries (and not source code).

This means we should not have cat B binaries in git repo which is the source code.

We need to find another way to package them into binary release without having them in source release (or source code/ source repo)

https://apache.org/legal/resolved.html#binary-only-inclusion-condition


liuxunorg:
@felixcheung , thanks for your reminder. :-)
If remove the binary from the source, When zeppelin was released, by downloading the binary to the zeppelin package, Does this meet the requirements?

If this is possible, I will open a new ticket, the follow-up to complete the job.

felixcheung:
if they are cat b, it's ok in the build process they are downloaded and packaged into the -bin binary package

please check the license for these dependencies

PR #3360, opened 2019-05-13
[ZEPPELIN-4152]. Use spark 2.2 as the default profile

zjffdu:
### What is this PR for?
Trivial PR to use spark 2.2 as the default profile so that the default build command can pass

```
mvn clean package -DskipTests
```

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4152

* CI pass

### Questions:

### DIFF
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -196,9 +196,6 @@
                 <protobuf.version>2.5.0</protobuf.version>
                 <py4j.version>0.10.7</py4j.version>
             </properties>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
         </profile>
 
         <profile>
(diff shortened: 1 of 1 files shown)


zjffdu:
Unfortunately, spark 2.3 doesn't support scala 2.10 which is default profile.

felixcheung:
hm. I guess we need a plan to move to 2.11

zjffdu:
I tried to do that, but some other interpreters don't support scala-2.11. So it would take much more time to support scala-2.11 fully in zeppelin. I plan to move some interpreter out of zeppelin project to subproject, so that each interpreter can has its own build environment and dependency. This PR is a temporary solution before that

felixcheung:
sure, this could be a big problem...

felixcheung:
LGTM

PR #3385, opened 2019-06-14
[ZEPPELIN-4195] Fixed deleted note does not take effect in the file system

liuxunorg:
### What is this PR for?
Because the latest note storage is saved in a real file directory.
When deleting the note,
Move /note-storage-dir/note1 directly to /note-storage-dir/~Trash/note1,
If the /note-storage-dir/~Trash directory does not exist, then moving the note will fail.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4195

* [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/545631231)
### Questions:


zjffdu:
Can you add some unit test to cover this ?

liuxunorg:
There are already test cases,  https://github.com/apache/zeppelin/blob/master/zeppelin-plugins/notebookrepo/filesystem/src/test/java/org/apache/zeppelin/notebook/repo/FileSystemNotebookRepoTest.java#L111

But `FileSystem` has some differences in the two different file systems, the `local file system` and `HDFS`. 

### FileSystem::move(src, dest)
+ In the local file system, the parent directory of dest can be created automatically if it does not exist.
+ In the HDFS file system, the parent directory of dest will fail if it does not exist.

zjffdu:
I see, thanks @liuxunorg +1

PR #3389, opened 2019-06-30
[ZEPPELIN-4209] Fixed the ClusterInterpreterLauncher can not listen cluster events

liuxunorg:
### What is this PR for?
1. **ClusterInterpreterLauncher can not listen cluster events**
In Cluster mode, The `ClusterInterpreterLauncher` needs to listen for cluster events, Accept requests from other zeppelin servers to create interpreter processes.
Since the `ClusterInterpreterLauncher` is lazy, dynamically generated, So in cluster mode, when the zeppelin service starts, Create a `ClusterInterpreterLauncher` object, This allows the `ClusterInterpreterLauncher` to listen for cluster events.

2. **Replace `RemoteInterpreterManagedProcess` with `ClusterInterpreterProcess` for local interpreter process creation**
Because in cluster mode, `RemoteInterpreterServer` performs service discovery by registering thrift's `ip` and `port` into the cluster metadata, not through the thrift interface `callback`.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4209

* [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/552320897)

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/cluster/src/main/java/org/apache/zeppelin/interpreter/launcher/ClusterInterpreterCheckThread.java
+++ b/zeppelin-plugins/launcher/cluster/src/main/java/org/apache/zeppelin/interpreter/launcher/ClusterInterpreterCheckThread.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.interpreter.launcher;
+
+import org.apache.zeppelin.cluster.ClusterManagerServer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+
+import static org.apache.zeppelin.cluster.meta.ClusterMeta.INTP_TSERVER_HOST;
+import static org.apache.zeppelin.cluster.meta.ClusterMeta.INTP_TSERVER_PORT;
+import static org.apache.zeppelin.cluster.meta.ClusterMetaType.INTP_PROCESS_META;
+
+// Metadata registered in the cluster by the interpreter process,
+// Keep the interpreter process started
+public class ClusterInterpreterCheckThread extends Thread {
+  private static final Logger LOGGER
+      = LoggerFactory.getLogger(ClusterInterpreterCheckThread.class);
+
+  private ClusterInterpreterProcess intpProcess;
+
+  ClusterInterpreterCheckThread(ClusterInterpreterProcess intpProcess) {
+    this.intpProcess = intpProcess;
+  }
+
+  @Override
+  public void run() {
+    LOGGER.info("ClusterInterpreterCheckThread run() >>>");
+
+    ClusterManagerServer clusterServer = ClusterManagerServer.getInstance();
+
+    String intpGroupId = intpProcess.getInterpreterGroupId();
+
+    HashMap<String, Object> intpMeta = clusterServer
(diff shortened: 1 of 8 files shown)


liuxunorg:

zjffdu:
@liuxunorg Could you update the PR title to describe what kind of issue you try to fix ? The current title is a little vague.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -210,6 +211,18 @@ private void initInterpreterSetting(InterpreterSetting interpreterSetting) {
...
         .setInterpreterEventServer(interpreterEventServer)
         .postProcessing();
+    
+    // Since the ClusterInterpreterLauncher is lazy, dynamically generated, So in cluster mode,
+    // when the zeppelin service starts, Create a ClusterInterpreterLauncher object,
+    // This allows the ClusterInterpreterLauncher to listen for cluster events.
+    if (conf.isClusterMode()) {
+      try {
+        String pluginName = interpreterSetting.getLauncherPlugin();
+        PluginManager.get().loadInterpreterLauncher(pluginName, recoveryStorage);
```
zjffdu: Why init IntepreterLauncher here instead of InterpreterSetting ?

zeppelin-plugins/launcher/cluster/src/main/java/org/apache/zeppelin/interpreter/launcher/ClusterInterpreterCheckThread.java
```diff
@@ -0,0 +1,86 @@
...
+        break;
+      }
+    }
+
+    if (null == intpMeta || !intpMeta.containsKey(INTP_TSERVER_HOST)
+        || !intpMeta.containsKey(INTP_TSERVER_PORT)) {
+      LOGGER.error("Can not found interpreter meta!");
+    }
+
+    LOGGER.info("ClusterInterpreterCheckThread run() <<<");
```
zjffdu: What happens if any exception happens in this thread ? Might be better to add try catch ?

liuxunorg: Already added try catch at https://github.com/apache/zeppelin/pull/3389/files/b1cdcaba70aa34f2263c81a20bb7432b748f3fb2#diff-8b0ca33dee568da61a22747f9c13fbcaR59

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
liuxunorg: Modify `ClusterInterpreterLauncher` initialization in zeppelin server.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -358,6 +361,22 @@ private static void setupClusterManagerServer(ServiceLocator serviceLocator) {
...
+      // when the zeppelin service starts, Create a ClusterInterpreterLauncher object,
+      // This allows the ClusterInterpreterLauncher to listen for cluster events.
+      try {
+        InterpreterSettingManager intpSettingManager = sharedServiceLocator.getService(InterpreterSettingManager.class);
+        RecoveryStorage recoveryStorage = ReflectionUtils.createClazzInstance(
+                conf.getRecoveryStorageClass(),
+                new Class[] {ZeppelinConfiguration.class, InterpreterSettingManager.class},
+                new Object[] {conf, intpSettingManager});
+        recoveryStorage.init();
+        PluginManager.get().loadInterpreterLauncher(InterpreterSetting.CLUSTER_INTERPRETER_LAUNCHER_NAME, recoveryStorage);
```
zjffdu: Launcher will be created here https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java#L294
Why you do it again here ?

liuxunorg: Since the `ClusterInterpreterLauncher` is lazy, dynamically generated, So in cluster mode, 
When the user needs to create an interpreter on the cluster (for example: `node A`)
If the `A node` has never created an interpreter and never initialized the `ClusterInterpreterLauncher` object, then the `ClusterInterpreterLauncher` object on the a node will not be able to listen to the create interpreter event sent to it by the cluster.

Therefore, in cluster mode, the `ClusterInterpreterLauncher` object needs to be initialized in advance when the zeppelin server is started.

liuxunorg:

liuxunorg:
[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/554527481)

PR #3391, opened 2019-07-02
[ZEPPELIN-4189] Record INTERPRETER_RUN_COMMAND execution log in the interpreter.sh

xunliu:
### What is this PR for?
Record the `INTERPRETER_RUN_COMMAND` execution log in the interpreter.sh file. Can help find the reason why the startup interpreter failed to execute. 
Especially in the container environment, it can play a relatively large role. When creating an interpreter process in the container, if it fails, you can use this log to re-execute and check the problem.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4189

[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/553071915) 

### Questions:


liuxunorg:

PR #3393, opened 2019-07-02
[ZEPPELIN-4214]. Spark Web UI is displayed in the wrong paragraph

zjffdu:
### What is this PR for?

This PR fix the issue that spark job url is displayed in the wrong paragraph. Unit test is added in SparkInterpreterTest.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4214

* CI pass

Before

image

After
image

### Questions:

### DIFF
--- a/spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
+++ b/spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
@@ -96,35 +96,42 @@ public abstract class SparkShims {
   public abstract String showDataFrame(Object obj, int maxResult);
 
 
-  protected String getNoteId(String jobgroupId) {
-    int indexOf = jobgroupId.indexOf("-");
-    int secondIndex = jobgroupId.indexOf("-", indexOf + 1);
-    return jobgroupId.substring(indexOf + 1, secondIndex);
-  }
-
-  protected String getParagraphId(String jobgroupId) {
-    int indexOf = jobgroupId.indexOf("-");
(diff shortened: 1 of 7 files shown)


zjffdu:
@felixcheung @liuxunorg @jongyoul Could you help review it ? Thanks

spark/spark-shims/src/main/scala/org/apache/zeppelin/spark/SparkShims.java
```diff
@@ -96,37 +96,44 @@ public abstract void setupSparkListener(String master,
...
       jobUrl = sparkWebUrl + "/jobs";
     }
+    String jobGroupId = jobProperties.getProperty("spark.jobGroup.id");
 
     Map<String, String> infos = new java.util.HashMap<String, String>();
     infos.put("jobUrl", jobUrl);
     infos.put("label", "SPARK JOB");
     infos.put("tooltip", "View in Spark web UI");
-    infos.put("noteId", context.getNoteId());
-    infos.put("paraId", context.getParagraphId());
```
liuxunorg: Because each time execute a paragraph, The paragraph id in the InterpreterContext is the currently executed paragraph ID. The resulting BUG?

zjffdu: The root cause is that InterpreterContext is never updated, it is always the first paragraph's InterpreterContext, so the spark job url is always displayed in the first paragraph.

zjffdu:

PR #3397, opened 2019-07-05
[ZEPPELIN-4208] Cluster synchronize InterpreterSetting

liuxunorg:
### What is this PR for?
In cluster mode, The user creates, modifies, and deletes the InterpreterSetting on any of the zeppelin servers.
All need to be notified to all the zeppelin servers in the cluster to synchronize the update of InterpreterSetting. Failure to do so will result in the user not being able to continue while switching to another server.

1. Listen for note update events
  Listen for the NEW_INTERPRETER_SETTING、DEL_INTERPRETER_SETTING、UPDATE_INTERPRETER_SETTING ... event of the InterpreterSetting in the InterpreterSettingManager#onMessage() function.

2. Broadcast note update event
  The note is refreshed by notifying the event to all zeppelin servers in the cluster via messagingService.

### What type of PR is it?
[Feature]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4208

* [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/554484474)

Sync-InterpreterSetting

### Questions:


liuxunorg:

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -978,4 +1041,85 @@ public void onParagraphUpdate(Paragraph p) throws IOException {
...
+    Map<String, InterpreterProperty> properties
+        = (Map<String, InterpreterProperty>) intpSetting.getProperties();
+    InterpreterOption intpOption = intpSetting.getOption();
+
+    HashMap<String, String> params = new HashMap<>();
+    String jsonDep = gson.toJson(dependencies, new TypeToken<List<Dependency>>() {
+    }.getType());
+    String jsonProps = gson.toJson(properties, new TypeToken<Map<String, InterpreterProperty>>() {
+    }.getType());
+    params.put("id", intpSetting.getId());
```
zjffdu: How about create toJson in InterpreterSetting.java, so that you don't to create toJson in Dependency and InterpreterOption ? I believe we may need to serialization interpreter setting elsewhere, although now we do it via InterpreterSettingInfoSaving, but this is due to legacy reason. We should unify the serialization of InterpreterSetting.

liuxunorg: If I replace the entire `InterpreterSetting` with json, I am worried that the data is too large, and I only need data for `dependencies` and `InterpreterProperty`, and I don't need the full `InterpreterSetting` data.

liuxunorg:
Any more comments?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
zjffdu: It won't be too large. Many fields of InterpreterSetting is transient, that would be serialized.

liuxunorg:
InterpreterSetting support `JsonSerializable` add test case.
[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/555851743)

liuxunorg:
@zjffdu , Please help review code, thanks!

zjffdu:
Thanks @liuxunorg  LGTM
BTW, another thing I am thinking about is the integration test of Cluster Manager. Maybe we could simulate that by starting 2 zeppelin server in one machine.

liuxunorg:
I added a few test cases that can run several nodes on a server to create a zeppelin cluster. They are

### One zeppelin-sever + Two ClusterManagerServer -> Compose a zeppelin cluster
> Can simulate cluster events for zeppelin clusters.

https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/test/java/org/apache/zeppelin/cluster/ClusterEventTest.java

### One or Three ClusterManagerServer -> Compose a zeppelin cluster
> Metadata test used to check the cluster itself.

https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/ClusterMultiNodeTest.java

https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/ClusterSingleNodeTest.java

PR #3398, opened 2019-07-05
[ZEPPELIN-4219] User can start interpreter when interpreter dependencies jars are not downloaded

zjffdu:
### What is this PR for?
Now interpreter can start even when it is downloading dependencies. It will cause ClassNotFound exception. The root cause is that the status of InterpreterSetting is incorrectly set to `READY`.  This PR fix this issue.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4219

* CI pass

### Questions:


zjffdu:
That's right @felixcheung

PR #3400, opened 2019-07-05
[ZEPPELIN-4217]. Remove semicolon at the end of sql statement for JdbcInterpreter

zjffdu:
### What is this PR for?
It is very possible for user to add semicolon at the end of sql statement. Some jdbc engine will handle the semicolon for users, while others won't, e.g. hive. This PR will remove the semicolon at the end of sql. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4217

* CI pass

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -710,9 +710,13 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
     try {
       List<String> sqlArray;
+      sql = sql.trim();
       if (splitQuery) {
         sqlArray = splitSqlQueries(sql);
       } else {
+        if (sql.endsWith(";")) {
+          sql = sql.substring(0, sql.length() - 1);
+        }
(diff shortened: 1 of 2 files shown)


zjffdu:
@prabhjyotsingh @felixcheung Could you help review it ?

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -713,6 +713,9 @@ private InterpreterResult executeSql(String propertyKey, String sql,
       if (splitQuery) {
         sqlArray = splitSqlQueries(sql);
       } else {
+        if (sql.trim().endsWith(";")) {
+          sql = sql.trim().substring(0, sql.trim().length() - 1);
+        }
```
felixcheung: nit: why not sql.trim() in all other cases?

zjffdu: Fixed

PR #3404, opened 2019-07-13
[ZEPPELIN-4235] Pass environment variables for pyspark with impersonate.

woowahan-jaehoon:
### What is this PR for?
Pass 2 environment variable for pyspark with impersonate
- PYTHONPATH
- ZEPPELIN_HOME

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4235

* Run paragraph that use pyspark interpreter with impersonate setting without `PYTHONPATH` in `zeppelin-env.sh`.

### Questions:
    - No
    - No
    - No


woowahan-jaehoon:
> I thought source to inherent the environment from the current one?

@felixcheung, Sorry but I didn't understand your question.

Without `PYTHONPATH` in `zeppelin-env.sh` and before this patch, 
pyspark can run without impersonate setting but with impersonate, pyspark cannot run.

Because PYTHONPATH is made by `interpreter.sh`
but when impersonate is on, PYTHONPATH wasn't exist each system account so spark-submit cann't use `PYTHONPATH`.

`ZEPPELIN_HOME` also be not passed. but I think that it can be important information for something because it is base path.

PR #3405, opened 2019-07-15
[ZEPPELIN-4236] Cluster mode support interpreter running in docker

liuxunorg:
### What is this PR for?
Now that zeppelin is set up in cluster mode, the interpreter will not work in docker mode.
We need to mix these two modes together.
This will allow more users to use the interpreter container.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4236

* [CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/557797483)

### Questions:


liuxunorg:
Add `Docker Mode` & `Local Mode` test case.
And [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/558774215)

liuxunorg:

PR #3407, opened 2019-07-15
[ZEPPELIN-4240] Fixed the date format error in cluster management page

liuxunorg:
### What is this PR for?
The startup time of the zeppelin service and the interpreter process in the cluster metadata. The heartbeat time is the LocalDateTime.
The correct format is not displayed in the cluster management page.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4240

* [CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/558822999)

# Before fixed
cluster-web-page-error-time

# After fixed
FormattedDateTime

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/ClusterRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/ClusterRestApi.java
@@ -134,8 +134,13 @@ public class ClusterRestApi {
       if (properties.containsKey(ClusterMeta.SERVER_START_TIME)) {
         // format LocalDateTime
         Object serverStartTime = properties.get(ClusterMeta.SERVER_START_TIME);
-        String dateTime = formatDateTime(serverStartTime);
-        sortProperties.put(ClusterMeta.SERVER_START_TIME, dateTime);
+        if (serverStartTime instanceof LocalDateTime) {
+          LocalDateTime localDateTime = (LocalDateTime) serverStartTime;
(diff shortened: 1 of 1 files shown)


liuxunorg:
@zjffdu , @felixcheung Please help me review code, Thanks!

zeppelin-server/src/main/java/org/apache/zeppelin/rest/ClusterRestApi.java
```diff
@@ -152,8 +160,15 @@ public Response getClusterNodes(){
     return new JsonResponse(Response.Status.OK, "", nodes).build();
   }
 
-  private String formatIntpLink(String intpName) {
-    return String.format("<a href=\"/#/cluster/%s\">%s</a>", intpName, intpName);
+  private String formatDateTime(Object localDateTime) {
```
felixcheung: i think the function should be strongly types? one approach is `private String formatDateTime(LocalDateTime localDateTime)`

and then `String dateTime = formatDateTime((LocalDateTime)serverStartTime);`

I think this will be more maintainable.

liuxunorg: done. and [CI pass](https://travis-ci.org/liuxunorg/zeppelin/builds/559279220)

liuxunorg:

PR #3411, opened 2019-07-22
ZEPPELIN-4272: Zeppelin fails to compile when hadoop3 is enabled

bhavikpatel9977:
### What is this PR for?
To compile and build Zeppelin with hadoop3 added couple of missing dependencies

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-4272](https://issues.apache.org/jira/browse/ZEPPELIN-4272)

* CI green

apache_zepp

### Questions:
* Is there breaking changes for older versions? Yes, added 'kerberos-client' dependency for hadoop3

### DIFF
--- a/zeppelin-zengine/pom.xml
+++ b/zeppelin-zengine/pom.xml
@@ -48,6 +48,8 @@
     <aws.sdk.s3.version>1.10.62</aws.sdk.s3.version>
     <commons.vfs2.version>2.2</commons.vfs2.version>
     <eclipse.jgit.version>4.5.4.201711221230-r</eclipse.jgit.version>
+    <jettison.version>1.4.0</jettison.version>
+    <kerberos-client.version>2.0.0-M15</kerberos-client.version>
     <!--test library versions-->
     <google.truth.version>0.27</google.truth.version>
     <google.testing.nio.version>0.32.0-alpha</google.testing.nio.version>
@@ -157,12 +159,6 @@
       <groupId>org.apache.lucene</groupId>
       <artifactId>lucene-highlighter</artifactId>
       <version>${lucene.version}</version>
-    </dependency>
-
-    <dependency>
-      <groupId>com.amazonaws</groupId>
-      <artifactId>aws-java-sdk-s3</artifactId>
-      <version>${aws.sdk.s3.version}</version>
     </dependency>
 
     <dependency>
@@ -312,6 +308,12 @@
       </activation>
 
       <dependencies>
+        <dependency>
+          <groupId>com.amazonaws</groupId>
+          <artifactId>aws-java-sdk-s3</artifactId>
+          <version>${aws.sdk.s3.version}</version>
+        </dependency>
+
         <dependency>
           <groupId>org.apache.hadoop</groupId>
           <artifactId>hadoop-client</artifactId>
@@ -650,6 +652,21 @@
           <version>${hadoop.version}</version>
           <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>org.codehaus.jettison</groupId>
+          <artifactId>jettison</artifactId>
+          <version>${jettison.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-aws</artifactId>
+          <version>${hadoop.version}</version>
+        </dependency>
+		<dependency>
+		  <groupId>org.apache.directory.server</groupId>
+		  <artifactId>kerberos-client</artifactId>
+		  <version>${kerberos-client.version}</version>
+	    </dependency>
       </dependencies>
     </profile>
   </profiles>
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
@@ -29,6 +29,8 @@ import java.util.List;
 public class FileSystemStorage {
 
   private static Logger LOGGER = LoggerFactory.getLogger(FileSystemStorage.class);
+  private static final String S3A = "s3a";
+  private static final String FS_DEFAULTFS = "fs.defaultFS";
 
   // only do UserGroupInformation.loginUserFromKeytab one time, otherwise you will still get
   // your ticket expired.
@@ -60,18 +62,42 @@ public class FileSystemStorage {
   public FileSystemStorage(ZeppelinConfiguration zConf, String path) throws IOException {
     this.zConf = zConf;
     this.hadoopConf = new Configuration();
-    // disable checksum for local file system. because interpreter.json may be updated by
-    // non-hadoop filesystem api
-    // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks
-    this.hadoopConf.setBoolean("fs.file.impl.disable.cache", true);
-    this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
-    this.isSecurityEnabled = UserGroupInformation.isSecurityEnabled();
+    URI zepConfigURI;
+    URI defaultFSURI;
 
     try {
-      this.fs = FileSystem.get(new URI(path), this.hadoopConf);
+      zepConfigURI = new URI(path);
     } catch (URISyntaxException e) {
+      LOGGER.error("Failed to get Zeppelin config URI");
       throw new IOException(e);
     }
+    // disable checksum for local file system. because interpreter.json may be updated by
+    // non-hadoop filesystem api
+    // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks.
+
+    this.hadoopConf.setBoolean("fs.file.impl.disable.cache", true);
+    String defaultFS = this.hadoopConf.get(FS_DEFAULTFS);
+    try {
+      defaultFSURI = new URI(defaultFS);
+    } catch (URISyntaxException e) {
+      LOGGER.error("Failed to get defaultFS URI");
+      throw new IOException(e);
+    }
+
+    // to check whether underlying fileSystemStorage is S3A or not
+    if (!isS3AFileSystem(defaultFSURI, zepConfigURI)) {
+      this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
+    }
+
+    this.isSecurityEnabled = UserGroupInformation.isSecurityEnabled();
+
+    this.fs = FileSystem.get(zepConfigURI, this.hadoopConf);
+  }
+  
+  public boolean isS3AFileSystem(URI defaultFSURI, URI zepConfigURI) {
+    return defaultFSURI.getScheme().equals(S3A)
+      || (StringUtils.isNotEmpty(zepConfigURI.getScheme())
+      && zepConfigURI.getScheme().equals(S3A));
   }
 
   public FileSystem getFs() {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/zeppelinhub/websocket/ZeppelinhubClient.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/zeppelinhub/websocket/ZeppelinhubClient.java
@@ -37,6 +37,9 @@ import org.apache.zeppelin.notebook.repo.zeppelinhub.websocket.utils.Zeppelinhub
 import org.apache.zeppelin.notebook.socket.Message;
 import org.apache.zeppelin.notebook.socket.Message.OP;
 import org.apache.zeppelin.ticket.TicketContainer;
+import org.codehaus.jettison.json.JSONArray;
+import org.codehaus.jettison.json.JSONException;
+import org.codehaus.jettison.json.JSONObject;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.client.ClientUpgradeRequest;
@@ -44,9 +47,6 @@ import org.eclipse.jetty.websocket.client.WebSocketClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.amazonaws.util.json.JSONArray;
-import com.amazonaws.util.json.JSONException;
-import com.amazonaws.util.json.JSONObject;
 import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;


zeppelin-server/pom.xml
```diff
@@ -109,6 +110,12 @@
       </exclusions>
     </dependency>
 
+        <dependency>
+          <groupId>org.apache.directory.server</groupId>
+          <artifactId>kerberos-client</artifactId>
+          <version>${kerberos-client.version}</version>
```
zjffdu: Why we need kerberos-client in zeppelin-server ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -60,16 +63,42 @@
...
     // non-hadoop filesystem api
     // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks
+
     this.hadoopConf.setBoolean("fs.file.impl.disable.cache", true);
-    this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
+    String defaultFS = this.hadoopConf.get(FS_DEFAULTFS);
+    defaultFSURI = getURI(defaultFS, "Failed to get defaultFS URI");
+
+    if (isS3AFileSystem(defaultFSURI, zepConfigURI)) {
+      this.hadoopConf.set("fs.s3a.impl", S3AFileSystem.class.getName());
```
zjffdu: Why we have to set it explicitly? Hadoop can not detect the s3a filesystem ?

zeppelin-server/pom.xml
bhavikpatel9977: When i tried to build with hadoop3 profile, the build was failed and it throws an error `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project zeppelin-server: Compilation failure: Compilation failure:
[ERROR] /home/bhavik/workspace/apache_zep/zeppelin/zeppelin-server/src/main/java/org/apache/zeppelin/realm/kerberos/KerberosUtil.java:[41,58] package org.apache.directory.server.kerberos.shared.keytab does not exist`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
bhavikpatel9977: I have tried without setting s3a filesystem but it was not working, so i had set it explicitly; still i will give one more try without setting s3a filesystem.

zeppelin-server/pom.xml
zjffdu: Then I believe you need to put it under profile hadoop3

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
zjffdu: I doubt why hadoop doesn't support it, ideally hadoop should support this otherwise any downstream project has to set s3 filesystem explicitly which is annoying.

bhavikpatel9977: if we remove explicitly setting of fs.s3a.impl as S3AFileSystem, then it throws the class cast exception as it sets fs.file.impl as RawLocalFileSystem. So do we also need to remove the explicitly setting of fs.file.impl?

zjffdu: I am suggesting to remove it. I just want to confirm why hadoop didn't detect the right FileSystem implementation. As my understanding, hadoop should be able to detect the right FileSystem implementation based on the filesystem schema (e.g. file://, hdfs:// and etc). So could you help investigate why it doesn't work for s3. Thanks

bhavikpatel9977: After removing 'fs.file.impl' it works with all the filesystem schema (file:///, hdfs:///, s3a://)

bhavikpatel9977: @zjffdu : If anything is not missing then can I merge?

zjffdu: I think you need to update this ticket and PR description. There's no issue to connect to s3a. Only the compilation issue with hadoop3 is fixed in this PR IIUC.

bhavikpatel9977: @zjffdu I have updated ticket and PR description.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -60,16 +59,25 @@
...
     this.hadoopConf = new Configuration();
+    URI zepConfigURI;
+
+    zepConfigURI = getURI(path, "Failed to get URI path");
     // disable checksum for local file system. because interpreter.json may be updated by
     // non-hadoop filesystem api
     // disable caching for file:// scheme to avoid getting LocalFS which does CRC checks
+
     this.hadoopConf.setBoolean("fs.file.impl.disable.cache", true);
-    this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
```
zjffdu: Don't remove this line, see the comment above.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -60,16 +59,25 @@
   public FileSystemStorage(ZeppelinConfiguration zConf, String path) throws IOException {
     this.zConf = zConf;
     this.hadoopConf = new Configuration();
+    URI zepConfigURI;
```
zjffdu: Is the change in `FileSstemStorage.java` necessary ? IIUC, it is not necessary to do any change for this class.

zeppelin-server/pom.xml
```diff
@@ -515,6 +516,18 @@
...
+        <hadoop.version>3.0.0</hadoop.version>
+	  </properties>
+	  <dependencies>
+		<dependency>
+		  <groupId>org.apache.directory.server</groupId>
+		  <artifactId>kerberos-client</artifactId>
+		  <version>${kerberos-client.version}</version>
+		</dependency>
+	  </dependencies>
+	</profile>
```
zjffdu: Why add hadoop3 profile in zeppelin-server ? It is alrready in zeppelin-zengine. And only zeppelin-zengine use hadoop.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
bhavikpatel9977: We have set **zeppelin.notebook.dir** and **zeppelin.config.fs.dir** to s3a FileSystemStorage.

If we don't remove this line then it throws exception
``` WARN [2019-07-29 22:47:49,335] ({main} ContextHandler.java[log]:2359) - unavailable
MultiException stack 1 of 1
java.lang.ClassCastException: org.apache.hadoop.fs.RawLocalFileSystem cannot be cast to org.apache.hadoop.fs.LocalFileSystem
        at org.apache.hadoop.fs.FileSystem.getLocal(FileSystem.java:447)
        at org.apache.hadoop.fs.LocalDirAllocator$AllocatorPerContext.confChanged(LocalDirAllocator.java:301)
        at org.apache.hadoop.fs.LocalDirAllocator$AllocatorPerContext.getLocalPathForWrite(LocalDirAllocator.java:378)
        at org.apache.hadoop.fs.LocalDirAllocator$AllocatorPerContext.createTmpFileForWrite(LocalDirAllocator.java:461)
        at org.apache.hadoop.fs.LocalDirAllocator.createTmpFileForWrite(LocalDirAllocator.java:200)
        at org.apache.hadoop.fs.s3a.S3AFileSystem.createTmpFileForWrite(S3AFileSystem.java:753)
        at org.apache.hadoop.fs.s3a.S3ADataBlocks$DiskBlockFactory.create(S3ADataBlocks.java:811)
        at org.apache.hadoop.fs.s3a.S3ABlockOutputStream.createBlockIfNeeded(S3ABlockOutputStream.java:190)
        at org.apache.hadoop.fs.s3a.S3ABlockOutputStream.<init>(S3ABlockOutputStream.java:168)
        at org.apache.hadoop.fs.s3a.S3AFileSystem.create(S3AFileSystem.java:1044)
        at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1130)
        at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:1110)
        at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:999)
        at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:987)
```

zeppelin-server/pom.xml
bhavikpatel9977: This Jar file require only when hadoop3 is enabled, so we have added this dependency for hadoop3 profile.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
zjffdu: I am afraid this approach will introduce another bug which is mentioned in the above PR. how about setting it in your `hdfs-site.xml`?

bhavikpatel9977: Sorry I did not get which bug you are mentioning?

zjffdu: See this comment https://github.com/apache/zeppelin/pull/3411/files#diff-6b930e3e153d78dacf931917ff89587cR65

bhavikpatel9977: > how about setting it in your hdfs-site.xml?
Which property we have to set?

zjffdu: set `fs.file.impl` to `org.apache.hadoop.fs.LocalFileSystem`

bhavikpatel9977: After setting the above property still it throws same exception.

zjffdu: hmm, right, it will be overwritten here. Let me think about it

bhavikpatel9977: @zjffdu Did you get a chance to look at it this issue

zjffdu: @bhavikpatel9977 One approach I can think of is adding one extra configuration such as `zeppelin.config.storage.useS3a`, by default it is false. 
```
if (!useS3a) {
  this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
}
```

This way we can not only solve your issue but also won't break other user cases.

bhavikpatel9977: Ok. We are addding `zeppelin.config.storage.useS3A` property in conf/zeppelin-site.xml.template with default value as false.

bhavikpatel9977: Rather then adding new extra property we can also do something like this 
```if (!isS3AFileSystem(defaultFSURI, zepConfigURI)) {
  this.hadoopConf.set("fs.file.impl", RawLocalFileSystem.class.getName());
}```

``` public boolean isS3AFileSystem(URI defaultFSURI, URI zepConfigURI) {
    return defaultFSURI.getScheme().equals(S3A)
        || (StringUtils.isNotEmpty(zepConfigURI.getScheme())
        && zepConfigURI.getScheme().equals(S3A));
  }```

Comment (zjffdu) [reply zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java]: hmm, this approach seems better, as it doesn't introduce new configuration. BTW, please add a comment to explain this change.

File: zeppelin-server/pom.xml (hunk 5 above)
Comment (zjffdu) [reply zeppelin-server/pom.xml]: What I mean is that zeppelin-server doesn't require hadoop3, it is zeppelin-zengine that needs hadoop3. So you should put it in zeppelin-zengine

File: zeppelin-zengine/pom.xml [hunk 6]
```diff
@@ -650,6 +645,21 @@
...
           <scope>test</scope>
         </dependency>
+            <dependency>
+          <groupId>org.codehaus.jettison</groupId>
+          <artifactId>jettison</artifactId>
+          <version>${jettison.version}</version>
+        </dependency>
+            <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-aws</artifactId>
```
risdenk: It would probably be better to use `hadoop-cloud-storage` here since that would not only get s3, azure, and gcp as well. No need to individually pull these in.

Also I would hope that don't need kerberos client explicitly here.

bhavikpatel9977: Currently zeppelin support hadoop-3.0.0 version and if we want to support hadoop-cloud-storage then we have to bump-up the hadoop version.

zjffdu: I am fine to upgrade to hadoop 3.1.x

bhavikpatel9977: Shall we add a new profile? or do we have to update hadoop-version from 3.0.0 to 3.1.1

bhavikpatel9977: @zjffdu / @risdenk : I think we have to create separate jira to bump-up hadoop version.

zjffdu: Make sense, let's do it in separate PR

bhavikpatel9977: If everything is fine for you then can you please give +1.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/FileSystemStorage.java
```diff
@@ -60,16 +62,39 @@
...
+    URI defaultFSURI;
+
+    zepConfigURI = getURI(path, "Failed to get URI path");
```
zjffdu: It is not a good practice to pass error message to function. And is `getURI` necessary ? I see it only create new URI but do nothing else.

bhavikpatel9977: I have pushed requested changes.

bhavikpatel9977: @zjffdu can you please take a look and everything is ok then we can merge it.

bhavikpatel9977: @zjffdu Did you get a chance to look at it?

zjffdu:
@bhavikpatel9977 The CI is failed. Could you do rebased and retrigger the CI ?

zeppelin-zengine/pom.xml
```diff
@@ -312,6 +308,12 @@
       </activation>
 
       <dependencies>
+        <dependency>
+          <groupId>com.amazonaws</groupId>
```
prabhjyotsingh: Any specific reason this is moved from there to only `hadoop2` profile? Hope this does not cause any problem with `hadoop3` profile.

bhavikpatel9977: Initially this dependency is common for hadoop2 & hadoop3 but for hadoop3 we have added `hadoop-aws` dependency which clone all the required aws, s3 related classes and there are some conflict between the multiple versions of jar so we have removed `aws-java-sdk-s3` but forgot to added it in hadoop2 and that was the main reason of failing CI builds.

bhavikpatel9977: > Any specific reason this is moved from there to only `hadoop2` profile? Hope this does not cause any problem with `hadoop3` profile.

 Initially this dependency is common for hadoop2 & hadoop3 but for hadoop3 we have added `hadoop-aws` dependency which clone all the required aws, s3 related classes and there are some conflict between the multiple versions of jar so we have removed `aws-java-sdk-s3` but forgot to added it in hadoop2 and that was the main reason of failing CI builds.

prabhjyotsingh:
Other than one minor comments LGTM!
I see one of Travis profile is failing which is unrelated

bhavikpatel9977:
> @bhavikpatel9977 The CI is failed. Could you do rebased and retrigger the CI ?

I have rebased and also retriggered the CI, build got successful, Could please give us the +1 so we can merge it

zjffdu:
LGTM， will merge if no more comments

PR #3414, opened 2019-07-26
[ZEPPELIN-4280] Fix an issue that makes an infinite loop while switching the visualizations

Byeol:
### What is this PR for?
Fix an issue that makes an infinite loop while switching the visualizations.

An infinite loop occurs because the front-end web application tries to commit the paragraph even if there are no changes in the visualization configuration.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4280

If you switch quickly between different visualizations multiple times, the infinite loops should not happen.

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
@@ -812,6 +812,9 @@ function ResultCtrl($scope, $rootScope, $route, $window, $routeParams, $location
         newConfig.graph.values = newConfig.graph.commonSetting.pivot.values;
         delete newConfig.graph.commonSetting.pivot;
       }
+      if (angular.equals($scope.config, newConfig)) {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


zjffdu:
Thanks for the contribution @Byeol Could you add one line of comment so that it will be easier for others to understand it when reading zeppelin code.

zjffdu:
ping @Byeol

Leemoonsoo:
LGTM. Like @zjffdu mentioned, one line comment with link to the issue will help.

zjffdu:
ping @Byeol

Leemoonsoo:
ping @Byeol

Leemoonsoo:
I reproduced this problem and verified this patch fixes the problem
I'd like to include the fix in 0.8.2. Since @Byeol is not responding, I'll merge this PR to master and branch-0.8 and create a followup PR to add comment on the code.

PR #3421, opened 2019-08-02
ZEPPELIN-4227. Move StandardInterpreterLauncher and SparkInterpreterLauncher into zeppelin-zengine

zjffdu:
### What is this PR for?
We should include these default interpreter launcher into zeppelin-zengine, so that user can use zeppelin even without building zeppelin-plugins module. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4227

* CI pass

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/plugin/PluginManager.java
```diff
@@ -140,6 +140,19 @@ public synchronized InterpreterLauncher loadInterpreterLauncher(String launcherP
...
+    // If fails, then try to load it from plugin folder.
+    try {
+      InterpreterLauncher launcher = (InterpreterLauncher)
+              (Class.forName("org.apache.zeppelin.interpreter.launcher." + launcherPlugin))
+                      .getConstructor(ZeppelinConfiguration.class, RecoveryStorage.class)
+                      .newInstance(zConf, recoveryStorage);
+      return launcher;
+    } catch (InstantiationException | IllegalAccessException | ClassNotFoundException
+            | NoSuchMethodException | InvocationTargetException e) {
+      LOGGER.warn("Fail to instantiate InterpreterLauncher from classpath directly:" + launcherPlugin);
```
felixcheung: log e too

zjffdu: Addressed

PR #3425, opened 2019-08-14
[ZEPPELIN-4301]  js double precision loss problem

curlsli:
### What is this PR for?
JS double precision loss problem.

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Fix

### What is the Jira issue?
*  [https://issues.apache.org/jira/browse/ZEPPELIN-4301](https://issues.apache.org/jira/browse/ZEPPELIN-4301)

* CREATE TABLE 'student' ('id' BIGINT, 'age' INT)
* insert into student values (6703273802681287331, 2)
* select * from student

image

### Questions:
      No
      No
     No

### DIFF
--- a/zeppelin-web/src/app/tabledata/tabledata.js
+++ b/zeppelin-web/src/app/tabledata/tabledata.js
@@ -1,4 +1,8 @@
 /*
+ * @Author: curlsli@tencent.com
+ * @LastEditors: curlsli@tencent.com
+ */
+/*
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -60,7 +64,7 @@ export default class TableData extends Dataset {
         } else {
           let valueOfCol;
           if (!(col[0] === '0' || col[0] === '+' || col.length > float64MaxDigits)) {
-            if (!isNaN(valueOfCol = parseFloat(col)) && isFinite(col)) {
+            if (!isNaN(valueOfCol = col) && isFinite(col)) {
               col = valueOfCol;
             }
           }
--- a/zeppelin-web/src/app/tabledata/tabledata.js
+++ b/zeppelin-web/src/app/tabledata/tabledata.js
@@ -1,8 +1,4 @@
 /*
- * @Author: curlsli@tencent.com
- * @LastEditors: curlsli@tencent.com
- */
-/*
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at


curlsli:
> hi @curlsli, Can you attach a screenshot of the system after modifying the code in the PR?
> This way you can see the effect of the run.

okey

curlsli:
> ### What is this PR for?
> JS double precision loss problem.
> 
> ### What type of PR is it?
> [Bug Fix]
> 
> ### Todos
> * [ ]  - Fix
> 
> ### What is the Jira issue?
> * https://issues.apache.org/jira/browse/ZEPPELIN-4301
> 
> ### How should this be tested?
> * CREATE TABLE 'student' ('id' BIGINT, 'age' INT)
> * insert into student values (6703273802681287331, 2)
> * select * from student
> 
> ### Screenshots (if appropriate)
> image

> ### Questions:
> * Does the licenses files need update?
>   No
> * Is there breaking changes for older versions?
>   No
> * Does this needs documentation?
>   No

zeppelin-web/src/app/tabledata/tabledata.js
```diff
@@ -1,3 +1,7 @@
+/*
+ * @Author: curlsli@tencent.com
+ * @LastEditors: curlsli@tencent.com
```
felixcheung: in ASF, we discourage edit or author tag. commit log already shows who did the work

curlsli: Thanks @felixcheung ，the comment has been deleted.

zjffdu:
Thanks @curlsli 
Is it possible to add unit test to avoid regression ?

zeppelin-web/src/app/tabledata/tabledata.js
```diff
@@ -60,7 +60,7 @@ export default class TableData extends Dataset {
         } else {
           let valueOfCol;
           if (!(col[0] === '0' || col[0] === '+' || col.length > float64MaxDigits)) {
-            if (!isNaN(valueOfCol = parseFloat(col)) && isFinite(col)) {
+            if (!isNaN(valueOfCol = col) && isFinite(col)) {
```
prabhjyotsingh: Will changing this (removing `parseFloat`) matter in this case? If I understand it right, ideally it should not even come to this if.
Which makes me wonder what is `typeof col`, ideally that should be string, and if it was a string and `col.length` in this case should have been 19, which is greater than 16 (float64MaxDigits). 
Also if it is not a string then that a limitation of JS itself, as this is what I see on my browser console:
```
let col=6703273802681287331;
console.log(col);
6703273802681288000
```

Will try to repro this using JDBC interpreter,  what I used to test is the following paragraph:
```
%sh
echo -e "%table id\tage"
echo -e "6703273802681287331\t2"
```

curlsli: I think there is no problem. Because JS is to show that the real operations are run in the interpreter. This is also the simplest solution.

zjffdu:
Thanks @curlsli will merge if no more comment

PR #3427, opened 2019-08-18
ZEPPELIN-4304 Allow to override useNativeGit property of the gitcommitid plugin

alexott:
## What is this PR for?

It's now possible to override value of the `useNativeGit` configuration property of the
git commit id plugin with `plugin.gitcommitid.useNativeGit` property, like this:

```
mvn package -Dplugin.gitcommitid.useNativeGit=true
```

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-4304

### Questions:
No
No
No

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,8 @@
     <plugin.deploy.version>2.8.2</plugin.deploy.version>
     <plugin.shade.version>3.1.1</plugin.shade.version>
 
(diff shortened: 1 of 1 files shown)


alexott:
This may not be required after the next version of plugin will be released - it will allow to override this setting via command-line

zjffdu:
LGTM, will merge if no more comments

PR #3430, opened 2019-08-23
[ZEPPELIN-4312] enable livy to run as remote interpreter in docker

zlosim:
### What is this PR for?
This pull request enables livy interpreter to run as remote interpreter in docker. This is (hopefully) improvement of @liuxunorg `s work on running remote interpreters in docker

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4312

### Questions:
No
No
No

### DIFF
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -404,6 +404,7 @@ public class DockerInterpreterProcess extends RemoteInterpreterProcess {
       LOGGER.warn("{} file not found, Did not upload the krb5.conf to the container!", krb5conf);
     }
 
+    // TODO: Interpreter specific settings, we should consider general property or some other more elegant solution
     // 3) Get the keytab file in each interpreter properties
     // Upload Keytab file to container, Keep the same directory as local host
     // 3.1) shell interpreter properties keytab file
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -194,6 +194,7 @@ public class DockerInterpreterProcess extends RemoteInterpreterProcess {
     // Create container with exposed ports
     final ContainerConfig containerConfig = ContainerConfig.builder()
         .hostConfig(hostConfig)
+        .hostname(this.zeppelinServiceHost)
         .image(containerImage)
         .workingDir("/")
         .env(listEnv)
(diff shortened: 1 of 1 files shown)


zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -416,14 +417,18 @@ private void copyRunFileToContainer(String containerId)
       intpKeytab = properties.getProperty("submarine.hadoop.keytab", "");
     }
     if (StringUtils.isBlank(intpKeytab)) {
-      // 3.4) jdbc interpreter properties keytab file
+      // 3.4) livy interpreter properties keytab file
+      intpKeytab = properties.getProperty("zeppelin.livy.keytab", "");
+    }
+    if (StringUtils.isBlank(intpKeytab)) {
+      // 3.5) jdbc interpreter properties keytab file
```
zjffdu: This seems work, but it is weird to see interpreter specific thing in `DockerInterpreterProcess`. We might need to think about other approach for it. @liuxunorg

liuxunorg: Because the key table property names set by each interpreter are different, in `DockerInterpreterProcess.java`, the keytab file is now copied to the container in this way.
Let the interpreter in the container also be authenticated properly.

zlosim: so is there anything I can improve?

zlosim:
@zjffdu @liuxunorg kindly ask for feedback for this

zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
zjffdu: I am fine with current implementation also not so elegant. Maybe in future we should consider about a general property for interpreter keytab, e.g. `zeppelin.interpreter.keytab`. But it is beyond the scope of this PR, and could be done in another ticket.

zjffdu: Could you add a TODO comment so that we won't miss this issue ?

zlosim: Yes, I added todo.

felixcheung: what @zjffdu said.

zjffdu:
Thanks @zlosim  will merge if no more comments

PR #3431, opened 2019-08-26
[ZEPPELIN-2680] Allow opening notebook as a reader

redsk:
What is this PR for?
--------------------

When users only have reader permissions over a notebook AND the notebook
contains a chart, the app sends a COMMIT_PARAGRAPH for each paragraph,
which causes a write permission popup to show
("Insufficient privileges to write note").

Ths issue is with function commitParagraphResult in `result.controller.js`.

The original [fix](https://github.com/apache/zeppelin/pull/2439)
introduced a nasty bug that dramatically increased CPU usage and was never
merged.

This PR fixes the bug. It also automatically switches to `report` mode
as it is pointless to show UI features that the user is not allowed
to control.

What type of PR is it?
----------------------
Bug Fix

What is the Jira issue?
-----------------------
[ZEPPELIN-2680](https://issues.apache.org/jira/browse/ZEPPELIN-2680)

Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/notebook.controller.js
+++ b/zeppelin-web/src/app/notebook/notebook.controller.js
@@ -809,7 +809,7 @@ function NotebookCtrl($scope, $route, $routeParams, $location, $rootScope,
         minimumInputLength: 3,
       };
 
-      $scope.setIamOwner();
+      $scope.setMyPermissions();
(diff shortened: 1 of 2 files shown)


zjffdu:
LGTM

prabhjyotsingh:
LGTM, will merge this to master if no more discussion.

PR #3434, opened 2019-08-29
[ZEPPELIN-4281] Fixed unusable after cluster mode restarts interpreter

liuxunorg:
### What is this PR for?
In cluster mode,
After restarting the interpreter,
The cluster metadata sets the process state of the interpreter to offline.
The interpreter could not be executed again due to a judgment error.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4281

[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/577750017) And completed 4 test cases

1. ClusterInterpreterLauncherTest::testConnectExistOnlineIntpProcess()
2. ClusterInterpreterLauncherTest::testConnectExistOfflineIntpProcess()
3. ClusterInterpreterLauncherTest::testCreateIntpProcessDockerMode()
4. ClusterInterpreterLauncherTest::testCreateIntpProcessLocalMode()

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
@@ -31,7 +31,6 @@ import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
 import org.apache.zeppelin.cluster.ClusterManagerClient;
 import org.apache.zeppelin.cluster.meta.ClusterMeta;
-import org.apache.zeppelin.cluster.meta.ClusterMetaType;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.dep.DependencyResolver;
 import org.apache.zeppelin.display.AngularObject;
@@ -99,6 +98,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+
+import static org.apache.zeppelin.cluster.meta.ClusterMetaType.INTP_PROCESS_META;
 
 /**
  * Entry point for Interpreter process.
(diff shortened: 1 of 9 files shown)


liuxunorg:
Please help me review the code, Thanks!

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -238,6 +239,9 @@ public void run() {
   @Override
   public void shutdown() throws TException {
     logger.info("Shutting down...");
+    // delete interpreter cluster meta
+    deleteClusterMeta();
```
zjffdu: What happens if interpreter process is killed manually (e.g. via kill -9, or killed by yarn) in which case shutdown is not called.

liuxunorg: To determine if an interpreter is available, there are two ways to cluster.
1. The interpreter process exits normally, where the interpreter actively deletes the information in the cluster metadata.
2. The interpreter exits abnormally. zeppelin-server periodically checks if the interpreter is available. If it is not available, it will be set to offline.
https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/cluster/ClusterMonitor.java#L165

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -340,7 +344,21 @@ private void putClusterMeta() {
...
+
+  private void deleteClusterMeta() {
+    if (!zconf.isClusterMode()){
+      return;
+    }
+
+    try {
+      // delete interpreter cluster meta
+      clusterManagerClient.deleteClusterMeta(INTP_PROCESS_META, interpreterGroupId);
+      Thread.sleep(300);
```
felixcheung: why a sleep here?

liuxunorg: When `clusterManagerClient` operation metadata operations, it is asynchronous, so it can be more reliable.

liuxunorg:

PR #3442, opened 2019-09-10
[ZEPPELIN-4283] Choose which server to run the interpreter process

xunliu:
### What is this PR for?
Currently creating an interpreter in cluster mode,
Run on a server that has the least memory resources in all servers in the cluster,
However, jdk does not accurately obtain memory usage.
So, modify to the average policy and start the same number of interpreters in each server.

Obtaining the operating system's memory usage through the OperatingSystemMXBean is quite different from the actual physical memory usage.
The CPU information is accurate.

```
OperatingSystemMXBean operatingSystemMXBean
    = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);

// Returns the amount of free physical memory in bytes.
long freePhysicalMemorySize = operatingSystemMXBean.getFreePhysicalMemorySize();

// Returns the total amount of physical memory in bytes.
long totalPhysicalMemorySize = operatingSystemMXBean.getTotalPhysicalMemorySize();

// Returns the "recent cpu usage" for the whole system.
double systemCpuLoad = operatingSystemMXBean.getSystemCpuLoad();

int process = Runtime.getRuntime().availableProcessors();
```

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4283

[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/583024446)

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/cluster/ClusterManagerServer.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/cluster/ClusterManagerServer.java
@@ -48,6 +48,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
 
+import static org.apache.zeppelin.cluster.meta.ClusterMetaType.INTP_PROCESS_META;
 import static org.apache.zeppelin.cluster.meta.ClusterMetaType.SERVER_META;
 
 /**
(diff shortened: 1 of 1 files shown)


liuxunorg:
Can anyone help me review the code, Thank you!

zjffdu:
> So, modify to the average policy and start the same number of interpreters in each server.

Then it will require all the machines in the cluster to be the same (with same memory and cpu cores), do I understand it correctly ?

liuxunorg:
Not that, Because java can't get the actual physical memory usage of the server correctly.
So now change it, Created on the server with the least number of interpreter processes.

e.g., have A, B, C three zeppelin server,
A server: have one interpreter process.
B server: have two interpreter process.
C server: have two interpreter process.

If need create new interpreter process, The new interpreter processes will create by A server. because A server has the least interpreter processes number.

zjffdu:
This is what I concern. Why do you think choosing is correct ? IIUC, we have to assume that A,B,C have the same computing resources (cpu cores and memory) and besides we also assume all the interpreter processes consume the same amount of memory. But I don't think these assumption is true in reality. 

Do you know how much gap between the actual memory and the number calculated by java ? It is weird to me that we can not get the free memory programmatically,  it is a very common problem.

liuxunorg:
The reality is that, The memory deviation obtained by java is very large.
Causes the interpreters to be concentrated on one server.
Server resource differences generally do not exceed 30%.
So scheduling by quantity is better than following the way of resources.

zjffdu:
> So scheduling by quantity is better than following the way of resources.
This is based on the 2 assumptions I mentioned above, right ?

liuxunorg:
In two cases:

1. Zeppelin cluster + intp on localhsot mode: There are 2 problems. 

  + One is that the memory usage obtained by JAVA is very inaccurate. In our production environment, the intp allocation skew was very serious.

  + The second problem, after the intp starts, the user can sometimes closes him, so there is no strict and fair scheduling.

Our production environment is used, and the quantity distribution is the most ideal state.

2. Zeppelin cluster + intp on docker(or yarn, k8s) mode: Scheduling according to the number of intp is an ideal mode.

PR #3445, opened 2019-09-15
[ZEPPELIN-4263] Fixed cluster docker mode cannot exit the container of the remote interpreter

liuxunorg:
### What is this PR for?
Because docker first used the `while :; do sleep 1; done` script to start the container.

Execute `bin/interpreter.sh` to execute the interpreter startup script in the container.

When zeppelin closes or restarts the interpreter in docker, the interpreter process in the container can exit normally, but the container's startup shell does not exit, causing the container to continue running.

So we need to determine if the interpreter process exits in the container's startup script.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4263

[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/585091106)

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -45,6 +45,7 @@ import com.spotify.docker.client.DockerClient;
 import com.spotify.docker.client.LogStream;
 import com.spotify.docker.client.ProgressHandler;
 import com.spotify.docker.client.exceptions.DockerException;
+import com.spotify.docker.client.messages.Container;
 import com.spotify.docker.client.messages.ContainerConfig;
 import com.spotify.docker.client.messages.ContainerCreation;
 import com.spotify.docker.client.messages.ExecCreation;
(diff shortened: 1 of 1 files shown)


zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -340,16 +353,28 @@ public Void call(RemoteInterpreterService.Client client) throws Exception {
...
 
     // Close the docker client
     docker.close();
   }
 
+  private void removeExistContainer(String containerName) {
+    try {
+      docker.killContainer(containerName);
+    } catch (DockerException | InterruptedException e) {
+      LOGGER.error(e.getMessage(), e);
```
felixcheung: should not `error` if docker doesn't exist?

zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -340,16 +353,28 @@ public Void call(RemoteInterpreterService.Client client) throws Exception {
...
+    } finally {
+      try {
+        docker.removeContainer(containerName);
```
felixcheung: same here

zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -162,6 +162,8 @@ public String getInterpreterSettingName() {
   public void start(String userName) throws IOException {
     docker = DefaultDockerClient.builder().uri(URI.create(DOCKER_HOST)).build();
 
+    removeExistContainer(containerName);
```
zjffdu: Could you add comment here why need to remove exist container ?

zjffdu: I mean adding comment above this line of code, otherwise others will confuse to see that why needs to removeExistContainer

liuxunorg: done.

zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
liuxunorg: done.
Increase the judgment of whether there is a container with the same name.

liuxunorg:
[CI Pass](https://travis-ci.org/liuxunorg/zeppelin/builds/585958134)

liuxunorg:

PR #3446, opened 2019-09-15
ZEPPELIN-4329 fix building Scio with Scala 2.11

alexott:
### What is this PR for?

Goal is to make Scio & Beam interpreters buildable with Scala 2.11 profile.

Now both Scio & Beam interpreters explicitly use Scala 2.10, so it's possible
to compile both if Scala 2.11 profile is enabled.

Also was need to add missing dependency for Beam interpreter, otherwise it didn't work.

Plus small rework of the properties, so the Scala 2.10 version could be explicitly used in
the interpreters. (Also, tried Scala 2.10.7, and it didn't work, so we must stay on 2.10.5).

### What type of PR is it?
Bug Fix 

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4329

* Tested manually by building both default & Scala 2.10 profiles 
* Travis CI build: https://travis-ci.org/alexott/zeppelin/builds/585186691

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,8 +95,10 @@
   <properties>
     <!-- language versions -->
     <java.version>1.8</java.version>
-    <scala.version>2.10.5</scala.version>
+    <scala.2.10.version>2.10.5</scala.2.10.version>
+    <scala.version>${scala.2.10.version}</scala.version>
     <scala.binary.version>2.10</scala.binary.version>
+    <scala.2.11.version>2.11.8</scala.2.11.version>
     <scalatest.version>3.0.7</scalatest.version>
     <scalacheck.version>1.12.5</scalacheck.version>
 
@@ -752,7 +754,7 @@
         <activeByDefault>true</activeByDefault>
       </activation>
       <properties>
-        <scala.version>2.10.5</scala.version>
+        <scala.version>${scala.2.10.version}</scala.version>
         <scala.binary.version>2.10</scala.binary.version>
       </properties>
     </profile>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -95,8 +95,10 @@
   <properties>
     <!-- language versions -->
     <java.version>1.8</java.version>
-    <scala.version>2.10.5</scala.version>
+    <scala.2.10.version>2.10.5</scala.2.10.version>
```
zjffdu: What is the reason of introducing `scala.2.10.version` and `scala.2.10.version`  instead of using just `scala.version` ?

alexott: @zjffdu Scio & Beam interpreters need to stay on the 2.10 because for 2.11 we need to bump Beam dependencies, etc. So when compiling with `scala-2.11` profile, we need somehow specify Scala 2.10 version, and it was easier to take it from the top-level project.

zjffdu: Do you mean even I specify `-Pscala-2.11` scio module still use scala-2.10 ?

alexott: Yes, because the scio/beam dependencies are mess :-(
While Scio has both 2.10 & 2.11 dependencies, the Beam has either 2.10 or 2.11 dependencies, so it would require to switch Beam dependencies to higher versions when compiling with 2.11.

So I decided to fix the issue this way - just to fix compilation problems, etc., and then try to think how to make it more generic when working on JIRA about removing Scala 2.10 & adding Scala 2.12, etc. But that could be still a problem with Beam as it looks like that they aren't doing the releases that depend on different Scala versions.

zjffdu: I see, glad to know that you plan to fix it in a followup PR. This PR LGTM

PR #3448, opened 2019-09-17
[ZEPPELIN-4322] Flaky test: NotebookRestApiTest#testRunWithServerRestart

zjffdu:
### What is this PR for?
The root cause of this flaky test is due to that the note may be removed when paragraph job is still running. And the job will still send rpc message to zeppelin-server side which may cause unexpected exception. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4322

* CI pass

### Questions:


Leemoonsoo:
LGTM

PR #3451, opened 2019-09-18
[ZEPPELIN-4226] Fix "View in Spark web UI" in kubernetes mode

Leemoonsoo:
### What is this PR for?
When Zeppelin is running in Kubernetes, "View in Spark web UI" gives internal address, instead of address defined in SERVICE_DOMAIN.

I think this problem is side effect of https://github.com/apache/zeppelin/pull/3375 and this PR includes fix and updated unittest.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4226

Run Zeppelin on kubernetes, and run spark job, click "View in Spark web UI" button.

### Questions:

### DIFF
--- a/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
+++ b/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
@@ -23,6 +23,7 @@ import java.net.URLClassLoader
 import java.nio.file.Paths
 import java.util.concurrent.atomic.AtomicInteger
 
+import org.apache.commons.lang3.StringUtils
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.zeppelin.interpreter.util.InterpreterOutputStream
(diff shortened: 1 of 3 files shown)


spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -303,7 +303,11 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
 
   protected def createZeppelinContext(): Unit = {
     val sparkShims = SparkShims.getInstance(sc.version, properties)
-    sparkShims.setupSparkListener(sc.master, sparkUrl, InterpreterContext.get)
+    var webUiUrl = properties.getProperty("zeppelin.spark.uiWebUrl");
+    if (webUiUrl == null || webUiUrl.trim().length() == 0) {
```
zjffdu: Maybe use `StringUtils.isBlank()` which is more clean

Leemoonsoo: Updated!

zjffdu:
LGTM

Leemoonsoo:
Merge to master

PR #3454, opened 2019-09-19
[ZEPPELIN-4337] Make Zeppelin runs on kubernetes with different domain suffix

Leemoonsoo:
### What is this PR for?

A couple of hardcoded dns suffix in the source code (cluster.local). Remove them so Zeppelin have better compatibility with Kubernetes cluster with dns name.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4337

Should work on Kubernetes cluster with different cluster dns name than 'cluster.local'

### Questions:


Leemoonsoo:
Thanks @felixcheung. Merge to master.

PR #3458, opened 2019-09-24
[ZEPPELIN-4347]: add seperate shortcuts for run all above/below

santosh1994:
### What is this PR for?
Currently same shortcut is configured for run all above and run all below, clicking on shortcut will again show popup to choose run all above or run all below. With this additional action required from user it doesn't makes sense to use shortcut.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4347

* Ctrl + Shift + UP shortcut should run all the paragraphs above
* Ctrl  + Shift + Down shortcut should run all the paragraphs below and the current one

### Questions:

### DIFF
--- a/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
+++ b/zeppelin-web/src/app/notebook/paragraph/paragraph.controller.js
@@ -509,33 +509,18 @@ function ParagraphCtrl($scope, $rootScope, $route, $window, $routeParams, $locat
     $scope.$emit('runAllBelowAndCurrent', paragraph, true);
   };
 
-  $scope.runAllToOrFromThis = function(paragraph) {
-    BootstrapDialog.show({
-      message: 'Run paragraphs:',
-      title: '',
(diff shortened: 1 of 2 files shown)


santosh1994:
travis build - https://travis-ci.org/santosh1994/zeppelin/builds/598482665

Leemoonsoo:
Thanks, @santosh1994 for the improvement.

Just curious how did you come up with shortcut key combination `Ctrl + Shift + UP` and `Ctrl + Shift + DOWN`. Any reason behind of it? or is it a combination that you like (which is totally fine)?

santosh1994:
No reason as such, based on usage on cursor move and text selection with UP and DOWN arrow felt using it.
If we are following any command patter like `a` for above and `b` for below we can use the same combination

Leemoonsoo:
Thanks for the explanation.
I'm merging it to master.

PR #3474, opened 2019-10-04
[ZEPPELIN-4359] Interpreter list is incorrect after recent changes

alexott:
### What is this PR for?

Recently, the names of the artifacts were changed to exclude explicit Scala version. This PR fixes this problem. Also, missing interpreters were added to list.

### What type of PR is it?
Bug Fix & Improvement

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-4359


zjffdu:
LGTM

PR #3478, opened 2019-10-09
[ZEPPELIN-4369] Redundant line separator for multiple text output

zjffdu:
### What is this PR for?

There's redundant line separator when there's multiple text output. The root cause is that each `%text ` will append new line separator in frontend. This PR fix this issue by removing `%text ` is the previous output type is also `%text `.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4369

* CI pass

Before
before

After

after

### Questions:


prabhjyotsingh:
LGTM

PR #3483, opened 2019-10-12
[ZEPPELIN-4377] Interpreter doesn't start if name has space

alexott:
### What is this PR for?

We can create new interpreters based on existing ones, but we don't enforce its name, so
it's possible to enter it with space and other characters, that aren't handled correctly
by Bash.

This PR fixes this problem for master.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* ZEPPELIN-4377

* Tested manually by creating a new JDBC interpreter with name containing space, and executing it


zjffdu:
I don't think we should allow space in interpreter name, frontend would be able recognize such interpreter, such as `%a b`. So we should not allow user to create interpreter with space in name.

alexott:
I agree - but we still need to make interpreter more robust...

I'll create a separate JIRA to fix the UI to allow only specific characters - I need to look how the frontend extracts interpreter name right now. 

is this approach ok for you?

alexott:
@zjffdu I've pushed UI validation in this branch as well...

epugh:
I think I got bit by this!!!  Glad to know what issue was.

zjffdu:
Thanks @alexott  PR LGTM

zjffdu:
@alexott Could you paste the travis build link here ? Need to make sure CI pass before merging.

alexott:
https://travis-ci.org/alexott/zeppelin/builds/597228348 - please notice that it's failing in other's code. I can of course rebase, but we need to make sure that Docker launcher is fixed.

zjffdu:
Could you do a rebase ? I believe the error in travis build is already fixed

alexott:
Rebased, new build just started: https://travis-ci.org/alexott/zeppelin/builds/599109513

zjffdu:
@alexott Some tests are failed, probably due to flaky test, could you re-trigger them ?

PR #3485, opened 2019-10-13
[ZEPPELIN-3548] KSQL Interpreter for Zeppelin

conker84:
### What is this PR for?
This PR adds the support to [KSQL](https://docs.confluent.io/current/ksql/docs/developer-guide/api.html)

### What type of PR is it?
[Feature]

### Todos
* [x] - Created the interpreter

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3548

You can use the following guide in order to spin-up the KSQL stack.
https://github.com/confluentinc/demo-scene/blob/master/ksql-intro/demo_ksql-intro.adoc

ksql

### Questions:
* Does the licenses files need update? No, I didn't add any new dependency into the project.


conker84:
@rmoff please look also here

docs/interpreter/ksql.md
```diff
@@ -0,0 +1,77 @@
...
+  <tbody>
+    <tr>
+      <td>ksql.url</td>
+      <td>http://localhost:8080</td>
+      <td>The KSQL Endpoint base URL</td>
+    </tr>
+  </tbody>
+</table>
+
+N.b. The interpreter supports all the KSQL properties, i.e. `ksql.streams.auto.offset.reset`.
```
zjffdu: Could you add a link to refer all the ksql stream configuration ?

ksql/src/main/java/org/apache/zeppelin/ksql/BasicKSQLHttpClient.java
```diff
@@ -0,0 +1,173 @@
...
+    this.type = type;
+    this.headers = headers;
+    this.timeout = timeout;
+    this.connected = false;
+  }
+
+  @Override
+  public void close() throws IOException {
+    connected = false;
+    connection.disconnect();
```
zjffdu: It is possible that `connection` is null. Add following
```
if (connection != null) {
```

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  // VisibleForTesting
+  public KSQLInterpreter(Properties properties, KSQLRestService ksqlRestService) {
+    super(properties);
+    this.ksqlRestService = ksqlRestService;
+  }
+
+  @Override
+  public void open() throws InterpreterException {}
```
zjffdu: It is better to create and initialize `ksqlRestService` in `open` method.

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
```diff
@@ -0,0 +1,170 @@
...
+    try {
+      if (data instanceof Collection || data instanceof Map) {
+        return json.writeValueAsString(data);
+      }
+      if (data instanceof String) {
+        return (String) data;
+      }
+      return String.valueOf(data);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
```
zjffdu: use `IOException` ?

zjffdu:
Thanks for the contribution @conker84 , I am curious to know what will happen when the ksql return infinite data, will it cause frontend crash ?

docs/interpreter/ksql.md
conker84: Done

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
conker84: Can we left it as is? It's better for testing so I can inject the mocked version, and the behaviour si lazy for each paragraph

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
conker84: I use it as method reference at line 132 so I can't use a checked exception

ksql/src/main/java/org/apache/zeppelin/ksql/BasicKSQLHttpClient.java
conker84: done

conker84:
> Thanks for the contribution @conker84 , I am curious to know what will happen when the ksql return infinite data, will it cause frontend crash ?

great question @zjffdu , the output gets truncated when the `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT` is reached.

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
zjffdu: I see

ksql/src/main/java/org/apache/zeppelin/ksql/KSQLInterpreter.java
zjffdu: OK

zjffdu:
> > Thanks for the contribution @conker84 , I am curious to know what will happen when the ksql return infinite data, will it cause frontend crash ?
> 
> great question @zjffdu , the output gets truncated when the `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT` is reached.

Do you have any idea to improve it ? It is not a specific problem for ksql, but a general issue for all streaming job.

conker84:
> > > Thanks for the contribution @conker84 , I am curious to know what will happen when the ksql return infinite data, will it cause frontend crash ?
> > 
> > 
> > great question @zjffdu , the output gets truncated when the `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT` is reached.
> 
> Do you have any idea to improve it ? It is not a specific problem for ksql, but a general issue for all streaming job.

Do you mean for the back-end?
I should take a deep look into the Interpreter lifecycle management, but if we want to do it with the KISS principle maybe by simply adding a boolean property like `ZEPPELIN_INTERPRETER_STOP_ON_OUTPUT_LIMIT_REACHED`, that automatically stops the paragraph when the limit is reached, could be a flexible and "simple" solution.

zjffdu:
No, I mean the frontend. Ideally backup should always send streaming data to frontend, but frontend should do the truncation. e.g. frontend would always keep the latest 1000 records.

conker84:
Oh I see. Well I agree with you about showing the latest 1k (configurable?!) records

conker84:
I can open an issue about that

conker84:
Is there anything else that I can do for this PR?

Tagar:
@conker84 great improvement !

@zjffdu  yep would be great to have an option - if `ZEPPELIN_INTERPRETER_OUTPUT_LIMIT` is reached, for steaming data like KSQL, just remove first/oldest records in the output, so a paragraph can show live streaming data until it is explicitly stopped. again, this could be an option (e.g. `ZEPPELIN_INTERPRETER_STOP_ON_OUTPUT_LIMIT_REACHED`) like @conker84 has mentioned, perhaps made in a follow up change. 

thanks

zjffdu:
Thanks @conker84 @Tagar will merge this PR if no more comment

PR #3489, opened 2019-10-17
[ZEPPELIN-4387] Run travis_check.py using Github Actions instead of Jenkins

Leemoonsoo:
### What is this PR for?
Full CI migration to Github Actions will take time. ([ZEPPELIN-4385](https://issues.apache.org/jira/browse/ZEPPELIN-4385))
Meanwhile, we can make Github Actions do the same job Jenkins used to do, so we can have CI status in all contributors pull requests.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4387

### Questions:

### DIFF
--- a/.github/workflows/travis_check.yml
+++ b/.github/workflows/travis_check.yml
@@ -26,6 +26,9 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
+      - name: Install python dependencies
+        run: |
+          pip install requests
       - name: check
         run: |
           git log -n 1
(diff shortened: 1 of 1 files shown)


Leemoonsoo:
Hmm.. This does not brings Github Actions on, and I still see 

image

Looks like we can not use Github Actions?

zjffdu:
@Leemoonsoo I can see github action on pull request for my repo, but unable to trigger github action for apache repo. I think github action is enabled for apache project, so I suspect maybe it is due to zeppelin itself. Do you know how the jenkins job is triggered before ? Maybe it causes the github action untouched.

Reamer:
we are using Github Actions now. Closed.

PR #3491, opened 2019-10-24
[ZEPPELIN-4390]. ExecutorService is not properly shutdown

zjffdu:
### What is this PR for?

`ExecutorService` is not properly shutdown due to we didn't use the correct api. We should use `shutdownNow` instead of `shutdown`. See https://stackoverflow.com/questions/11520189/difference-between-shutdown-and-shutdownnow-of-executor-service.

The effect of this issue will cause thread resource leakage. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4390

* CI pass

### Questions:


zjffdu:

PR #3494, opened 2019-10-27
[ZEPPELIN-4393] Support for '--' comment in Cassandra interpreter

alexott:
### What is this PR for?

Cassandra Query Language (CQL) supports 2 types of end-of-line comments - `//` and `--`, but Cassandra interpreter supports only first one, resulting in error when using `--` comment in the query.

### What type of PR is it?
Bug

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-4393

* 2 unit tests were added - for both comments types
* Tested manually
* Travis build: https://travis-ci.org/alexott/zeppelin/builds/603532433

### Questions:
* Does this needs documentation? Documentation was updated


zjffdu:
Thanks @alexott  LGTM

PR #3498, opened 2019-10-31
[ZEPPELIN-4409] Set spark.app.name to know which user is running which notebook

amakaur:
### What is this PR for?
This PR sets spark.app.name to "Zeppelin Notebook xxxxx for user test" which makes it helpful to know which notebooks are currently being used and how much resources are being allocated by which user and notebook. 

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4409

- Create a notebook with spark interpreter
- Run println("hello")
- spark app name will be set to user name and note id 

#
image

## Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -725,13 +725,14 @@ public class InterpreterSetting {
   ///////////////////////////////////////////////////////////////////////////////////////
   // This is the only place to create interpreters. For now we always create multiple interpreter
   // together (one session). We don't support to create single interpreter yet.
-  List<Interpreter> createInterpreters(String user, String interpreterGroupId, String sessionId) {
+  List<Interpreter> createInterpreters(String user, String interpreterGroupId, String sessionId, String noteId) {
     List<Interpreter> interpreters = new ArrayList<>();
     List<InterpreterInfo> interpreterInfos = getInterpreterInfos();
     Properties intpProperties = getJavaProperties();
+    intpProperties.put( "spark.app.name", "Zeppelin Notebook " + noteId + " for user " + user );
     for (InterpreterInfo info : interpreterInfos) {
       Interpreter interpreter = new RemoteInterpreter(intpProperties, sessionId,
-          info.getClassName(), user, lifecycleManager);
+          info.getClassName(), user, lifecycleManager, noteId);
       if (info.isDefaultInterpreter()) {
         interpreters.add(0, interpreter);
       } else {
@@ -773,7 +774,7 @@ public class InterpreterSetting {
     Preconditions.checkNotNull(interpreterGroup, "No InterpreterGroup existed for user {}, " +
         "noteId {}", user, noteId);
     String sessionId = getInterpreterSessionId(user, noteId);
-    return interpreterGroup.getOrCreateSession(user, sessionId);
+    return interpreterGroup.getOrCreateSession(user, sessionId, noteId);
   }
 
   public Interpreter getDefaultInterpreter(String user, String noteId) {
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
@@ -158,11 +158,11 @@ public class ManagedInterpreterGroup extends InterpreterGroup {
     SchedulerFactory.singleton().removeScheduler(scheduler.getName());
   }
 
-  public synchronized List<Interpreter> getOrCreateSession(String user, String sessionId) {
+  public synchronized List<Interpreter> getOrCreateSession(String user, String sessionId, String noteId) {
     if (sessions.containsKey(sessionId)) {
       return sessions.get(sessionId);
     } else {
-      List<Interpreter> interpreters = interpreterSetting.createInterpreters(user, id, sessionId);
+      List<Interpreter> interpreters = interpreterSetting.createInterpreters(user, id, sessionId, noteId);
       for (Interpreter interpreter : interpreters) {
         interpreter.setInterpreterGroup(this);
       }
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
@@ -62,6 +62,7 @@ public class RemoteInterpreter extends Interpreter {
   private String className;
   private String sessionId;
   private FormType formType;
+  private String noteId;
 
   private RemoteInterpreterProcess interpreterProcess;
   private volatile boolean isOpened = false;
@@ -76,12 +77,14 @@ public class RemoteInterpreter extends Interpreter {
                            String sessionId,
                            String className,
                            String userName,
-                           LifecycleManager lifecycleManager) {
+                           LifecycleManager lifecycleManager,
+                           String noteId) {
     super(properties);
     this.sessionId = sessionId;
     this.className = className;
     this.setUserName(userName);
     this.lifecycleManager = lifecycleManager;
+    this.noteId = noteId;
   }
 
   public boolean isOpened() {
@@ -125,7 +128,7 @@ public class RemoteInterpreter extends Interpreter {
         // depends on other interpreter. e.g. PySparkInterpreter depends on SparkInterpreter.
         // also see method Interpreter.getInterpreterInTheSameSessionByClassName
         for (Interpreter interpreter : getInterpreterGroup()
-                                        .getOrCreateSession(this.getUserName(), sessionId)) {
+                                        .getOrCreateSession(this.getUserName(), sessionId, noteId)) {
           try {
             if (!(interpreter instanceof ConfInterpreter)) {
               ((RemoteInterpreter) interpreter).internal_create();
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroupTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroupTest.java
@@ -63,18 +63,18 @@ public class ManagedInterpreterGroupTest {
     assertEquals(0, interpreterGroup.getSessionNum());
 
     // create session_1
-    List<Interpreter> interpreters = interpreterGroup.getOrCreateSession("user1", "session_1");
+    List<Interpreter> interpreters = interpreterGroup.getOrCreateSession("user1", "session_1", "notebook_1");
     assertEquals(3, interpreters.size());
     assertEquals(EchoInterpreter.class.getName(), interpreters.get(0).getClassName());
     assertEquals(DoubleEchoInterpreter.class.getName(), interpreters.get(1).getClassName());
     assertEquals(1, interpreterGroup.getSessionNum());
 
     // get the same interpreters when interpreterGroup.getOrCreateSession is invoked again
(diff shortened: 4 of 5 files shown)


zjffdu:
Thanks for the contribution @amakaur 
It makes sense to include user name and note id in the spark app name. But one easier approach is to set `spark.app.name` in `SparkInterpreterLauncher`, you can set `spark.app.name` to be `interpreterGroupId` which is consisted of note id and user name. Here's the logic how `interpreterGroupId` is generated.  https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java#L406

amakaur:
Thanks for the suggestion @zjffdu , i tried setting spark.app.name in SparkInterpreterLauncher#buildEnvFromProperties through env variable "SPARK_APP_NAME" and via spark property "spark.app.name", but didn't seem to work. It sets spark.app.name to default value "Zeppelin".

zjffdu:
@amakaur How do you set that ? Can you show me the code ?

amakaur:
@zjffdu  here's two ways that i tried. 

``` @Override
  public Map<String, String> buildEnvFromProperties(InterpreterLaunchContext context) throws IOException {
    Map<String, String> env = super.buildEnvFromProperties(context);
    Properties sparkProperties = new Properties();
    String sparkMaster = getSparkMaster(properties);
    for (String key : properties.stringPropertyNames()) {
      if (RemoteInterpreterUtils.isEnvString(key)) {
        env.put(key, properties.getProperty(key));
      }
      if (isSparkConf(key, properties.getProperty(key))) {
        sparkProperties.setProperty(key, toShellFormat(properties.getProperty(key)));
      }
    }

    env.put("SPARK_APP_NAME", context.getInterpreterGroupId() + "_testing");
    //sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId() + "_testing");

    setupPropertiesForPySpark(sparkProperties);
    setupPropertiesForSparkR(sparkProperties);
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      env.put("ZEPPELIN_SPARK_YARN_CLUSTER", "true");
      sparkProperties.setProperty("spark.yarn.submit.waitAppCompletion", "false");
    }

    StringBuilder sparkConfBuilder = new StringBuilder();
    if (sparkMaster != null) {
      sparkConfBuilder.append(" --master " + sparkMaster);
    }
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      if (sparkProperties.containsKey("spark.files")) {
        sparkProperties.put("spark.files", sparkProperties.getProperty("spark.files") + "," +
            zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      } else {
        sparkProperties.put("spark.files", zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      }
      sparkProperties.put("spark.yarn.maxAppAttempts", "1");
    }```

- by either setting SPARK_APP_NAME OR by setting spark property spark.app.name

**Comment by zjffdu** [comment]
@amakaur `SPARK_APP_NAME` is not valid, you need to use `spark.app.name` which is what spark expect  ( refer [here](http://spark.apache.org/docs/latest/configuration.html) )

**Comment by amakaur** [comment]
@zjffdu Thanks for trying it out. Also, let me try again, but i had spark.app.name as commented out in the code that i posted just to show two different ways i was using to get it to work. 

public Map<String, String> buildEnvFromProperties(InterpreterLaunchContext context) throws IOException {
    Map<String, String> env = super.buildEnvFromProperties(context);
    Properties sparkProperties = new Properties();
    String sparkMaster = getSparkMaster(properties);
    for (String key : properties.stringPropertyNames()) {
      if (RemoteInterpreterUtils.isEnvString(key)) {
        env.put(key, properties.getProperty(key));
      }
      if (isSparkConf(key, properties.getProperty(key))) {
        sparkProperties.setProperty(key, toShellFormat(properties.getProperty(key)));
      }
    }

    //env.put("SPARK_APP_NAME", context.getInterpreterGroupId() + "_testing");
    sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId() + "_testing");

    setupPropertiesForPySpark(sparkProperties);
    setupPropertiesForSparkR(sparkProperties);
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      env.put("ZEPPELIN_SPARK_YARN_CLUSTER", "true");
      sparkProperties.setProperty("spark.yarn.submit.waitAppCompletion", "false");
    }

    StringBuilder sparkConfBuilder = new StringBuilder();
    if (sparkMaster != null) {
      sparkConfBuilder.append(" --master " + sparkMaster);
    }
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      if (sparkProperties.containsKey("spark.files")) {
        sparkProperties.put("spark.files", sparkProperties.getProperty("spark.files") + "," +
            zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      } else {
        sparkProperties.put("spark.files", zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      }
      sparkProperties.put("spark.yarn.maxAppAttempts", "1");
    }```

**Comment by amakaur** [comment]
Still not working for me, but will debug to see if i can find something. 

 ```@Override
  public Map<String, String> buildEnvFromProperties(InterpreterLaunchContext context) throws IOException {
    Map<String, String> env = super.buildEnvFromProperties(context);
    Properties sparkProperties = new Properties();
    String sparkMaster = getSparkMaster(properties);
    for (String key : properties.stringPropertyNames()) {
      if (RemoteInterpreterUtils.isEnvString(key)) {
        env.put(key, properties.getProperty(key));
      }
      if (isSparkConf(key, properties.getProperty(key))) {
        sparkProperties.setProperty(key, toShellFormat(properties.getProperty(key)));
      }
    }

    sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId());

    setupPropertiesForPySpark(sparkProperties);
    setupPropertiesForSparkR(sparkProperties);
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      env.put("ZEPPELIN_SPARK_YARN_CLUSTER", "true");
      sparkProperties.setProperty("spark.yarn.submit.waitAppCompletion", "false");
    }

    StringBuilder sparkConfBuilder = new StringBuilder();
    if (sparkMaster != null) {
      sparkConfBuilder.append(" --master " + sparkMaster);
    }
    if (isYarnMode() && getDeployMode().equals("cluster")) {
      if (sparkProperties.containsKey("spark.files")) {
        sparkProperties.put("spark.files", sparkProperties.getProperty("spark.files") + "," +
            zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      } else {
        sparkProperties.put("spark.files", zConf.getConfDir() + "/log4j_yarn_cluster.properties");
      }
      sparkProperties.put("spark.yarn.maxAppAttempts", "1");
    }```

**Comment by zjffdu** [comment]
Try this branch which works for me.
https://github.com/zjffdu/zeppelin/commit/47a50c0ae73c9fc8c5441572cee0d9164106a60f

And set spark.app.name to be empty in interpreter setting first.

**Comment by amakaur** [comment]
Sure, let me try.

**Comment by amakaur** [comment]
After debugging i realized that the reason my changes weren't working was because spark.app.name was getting overridden by the default value set in interpreter-setting. 

i did checkout your branch and got the following result.
image

**Comment by zjffdu** [comment]
Awesome, would you update this PR and add unit test ?

**Comment by zjffdu** [comment]
BTW, is the above UI some kind of interpreter process monitoring page ? It is pretty useful for zeppelin, do you have plan to contribute to zeppelin ?

**Comment by amakaur** [comment]
Yup will do once i figure out why user name is spark instead of an actual user name such as `amandeep.kaur`. 

I'm running Zeppelin on Mesos mode. The above is Mesos UI. Pretty useful.

**Comment by amakaur** [comment]
@zjffdu after debugging a bit this new approach might not work for me because user is set as `spark` instead of LDAP user.

**Comment by zjffdu** [comment]
@amakaur The user should be the zeppelin login user name. Do you enable LDAP in zeppelin ?

**Comment by amakaur** [comment]
Yeah LDAP is enabled. It works as expected w/ original solution.

**Comment by zjffdu** [comment]
Oh, you might forget one thing, you need to set spark interpreter as user isolated.

**Comment by amakaur** [comment]
I've it set as per note in isolated process

**Comment by zjffdu** [comment]
Still doesn't work for you in per user isolated mode ? What do you see in the spark app name ?

**Comment by amakaur** [comment]
`spark.app.name` is set as `spark-2ESRB4NZ1`  . 

The following is an ex from the logs 
```Create Session: shared_session in InterpreterGroup: spark-2ENBYUKC2 for user: amandeep.kaur``` where user is LDAP user BUT `interpreterGroupId` contains `spark as a user value`

I would like to set spark.app.name to `amandeep.kaur-2ENBYUKC2`.

**Comment by zjffdu** [comment]
Ah, you are using per note isolated, you should use `per user isolated` in which case the interpreterGroupId will include user name.

**Comment by amakaur** [comment]
I'm not sure if i'll be able to use `per user isolated` for zeppelin in prod because we would like each note to have its own JVM. Would using `per used isolated` create a single shared interpreter process for all of the notebooks that are under same user?

**Comment by zjffdu** [comment]
In that case, I think you can use per user isolated & per note isolated. Then the spark name would be `interpreter_name` + `user_name` + `noteId`

**Comment by iamabug** [comment]
> Try this branch which works for me.
> [zjffdu@47a50c0](https://github.com/zjffdu/zeppelin/commit/47a50c0ae73c9fc8c5441572cee0d9164106a60f)
> 
> And set spark.app.name to be empty in interpreter setting page first.

@zjffdu Is this branch enough to achieve the goal ?

**Comment by zjffdu** [comment]
@iamabug I don't think so, check my comments above https://github.com/apache/zeppelin/pull/3498#issuecomment-548179978. I believe the right approach is to update `SparkInterpreterLauncher`.

**Comment by iamabug** [comment]
@zjffdu I believe the commit I mentioned is updating `SparkInterpreterLauncher`:

```
    if (!sparkProperties.containsKey("spark.app.name") ||
            StringUtils.isBlank(sparkProperties.getProperty("spark.app.name"))) {
      sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId());
    }
```
here `spark.app.name` is set to `interpreterGroupId` which is generated as we want, am I right ?

zjffdu:
Yes, maybe it is better to add prefix `Zeppelin_`

iamabug:
And this is specific to Spark interpreter, I am not sure whether we need to consider other interpreters, such as Flink.

zjffdu:
If flink has such configuration, we could do it as well. But it could be done in another PR.

iamabug:
If `spark.app.name` is already set, should we leave it as it is or add username and note id also ? if we leave it, all users and notes shall have the same name, i don't think that's what we want to see. So I think the logic here could be like this:
1. if `spark.app.name` is not set or set to blank, we set it to `Zeppelin_` + `interpreter_name` + `username` + `noteId`, and `interpreter_name` is actually `spark`, so it is `Zeppelin_spark_` + `username` + `noteId`.
2. if `spark.app.name` is already set to a string, noted by `APP_NAME`, in this case, we set it to `APP_NAME` + `username` + `noteId`.

zjffdu:
> If `spark.app.name` is already set, should we leave it as it is or add username and note id also ? if we leave it, all users and notes shall still have the same name, i don't think that's what we want to see. So I think the logic here could be like this:
> 
> 1. if `spark.app.name` is not set or set to blank, we set it to `Zeppelin_` + `interpreter_name` + `username` + `noteId`, and `interpreter_name` is actually `spark`, so it is `Zeppelin_spark_` + `username` + `noteId`.
> 2. if `spark.app.name` is already set to a string, noted by `APP_NAME`, in this case, we set it to `APP_NAME` + `username` + `noteId`.
>    @zjffdu

1. We could use `Zeppelin_` + interpreter_group_name which may contains note id and user name which depends on the interpreter mode. (See https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java#L422)
2. I think it is fine to leave `spark.app.name` as what it is set when user set it explicitly.

PR #3499, opened 2019-10-31
[ZEPPELIN-4410] Added exception handling when converting old notebooks to newer format

amakaur:
### What is this PR for?
Added exception handling when creating a Note object from a json string. 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4410

- Create a notebook on zeppelin 0.7.3
- Create a spark interpreter 
- Run paragraph that creates a runtime exception
- Run upgrade-note.sh on the created notebook
- This will crash the script because it'll not be able to construct a java object from the given note.json

### Questions:


zjffdu:
LGTM

PR #3500, opened 2019-10-31
[ZEPPELIN-4411] Added current notebook location and name to ui header 

amakaur:
This makes it easier to find the notebook since multiple notebooks can have same name under different path

### What is this PR for?
This makes it easier to know which notebook you're currently running and the location of the notebook. 

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4411

- Create a notebook under a directory
- location of the notebook and name will be on the top 

image

### Questions:


zjffdu:
LGTM

PR #3501, opened 2019-10-31
[ZEPPELIN-4412] Added REST Api to shut down spark context for notebook

amakaur:
### What is this PR for?
This provides the ability to shut down interpreter for a specific notebook.

This makes it easier to release any allocated resources or close spark context when a paragraph is stuck on pending state.

This helps with resetting the state of the notebook. 

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4412

- Create a notebook with spark interpreter 
- Make a call to http://[zeppelin-server]:[zeppelin-port]/api/notebook/close/[noteid]

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -704,6 +704,28 @@ public class NotebookRestApi extends AbstractRestApi {
   }
 
   /**
+   * Shut down notebook interpreter REST API.
+   *
+   * @param noteId ID of Note
+   * @return JSON with status.OK
+   * @throws IOException
+   * @throws IllegalArgumentException
+   */
+  @GET
+  @Path("close/{noteId}")
+  @ZeppelinApi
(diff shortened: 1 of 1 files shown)


zjffdu:
I have a little concern about this PR. It looks like some workaround for issue of paragraph stuck in pending state, am I right ?

amakaur:
@zjffdu Yes, that's correct. We've ran into couple of issues with zeppelin 0.9 where this PR was necessary. Paragraph(s) would get stuck into a pending state so shutting down the interpreter would help bring back the notebook into working state.

zjffdu:
@amakaur I think this is a bug of Zeppelin. Could you let me know when you will hit this kind of issue, is it reproducible ?

amakaur:
@zjffdu its reproducible, but doesn't happen all of the time. 

Currently we instantiate spark interpreter per note in isolated mode for each user.
Change one of the spark interpreter settings from the ui and restart the interpreter.

After the restart the notebooks get stuck in pending state. This happens sometimes not all of the time.

PR #3502, opened 2019-10-31
[ZEPPELIN-4413] Added a close note button on zepp ui to shut down sc and release roso…

amakaur:
### What is this PR for?
Adds the ability to shut down interpreter and free allocated resources for a notebook via zeppelin ui. 

This makes it easy to close the notebook by clicking on one button

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4413

- Create a new notebook and use spark interpreter
- Click on close button between "clear output" and "clone this note"

image

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -131,6 +131,23 @@ public class NotebookService {
     return note;
   }
 
+  public boolean closeNote(String noteId,
+                           ServiceContext context,
+                           ServiceCallback<Note> callback) throws IOException{
+    Note note = null;
+    if ( noteId != null ) {
+      note = notebook.getNote(noteId);
+    }
+
+    if (note == null) {
+      callback.onFailure(new NoteNotFoundException(noteId), context);
+      return false;
+    }
+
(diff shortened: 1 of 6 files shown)


zjffdu:
You can release the interpreter resource by clicking restart interpreters in the note page. Or even you can set the LifecycleManager to be `TimeoutLifecycleManager` so that by default interpreter will be closed after 1 hour idle.

amakaur:
@zjffdu 1. We're already using TimeoutLifecycleManager to release the resources automatically after 1 hour, but there's an issue with closing notebook this way where https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java#L88 `close(sessionId)` wouldn't get triggered for some of the notebooks randomly and spark context would be alive for couple of days until zeppelin is restarted or notebook is closed using the feature in this PR. 

Also, we would like the ability to close the notebook at any moment without depending on the timeout.

2. Restarting the interpreter causes notebooks to get stuck in a pending state a lot of the times. So this helps to completely shut down the spark context and bring back the notebook in a fresh clean state.

3. This is similar to https://github.com/apache/zeppelin/pull/3501 but requires less work from the user to close the note. This makes it easy for any user to reset the state of their notebook.

PR #3510, opened 2019-11-04
[ZEPPELIN-4416]. Don't merge properties from inteprreter-setting.json if it already exists

zjffdu:
### What is this PR for?

Currently zeppelin will merge the setting in `interpreter-setting.json` with setting in `interpreter.json`, usually this is not what user expect. E.g. if user delete one property defined in `interpreter-setting.json` and then restart zeppelin, he would see this property again as we merged it from `interpreter-setting.json`. So this PR is to remove the logic of merging properties from interpreter-setting.json if it already exists.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4416

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -245,26 +245,13 @@ public class InterpreterSettingManager implements NoteEventListener, ClusterEven
 
       InterpreterSetting interpreterSettingTemplate =
           interpreterSettingTemplates.get(savedInterpreterSetting.getGroup());
-      // InterpreterSettingTemplate is from interpreter-setting.json which represent the latest
+      // InterpreterSettingTemplate is from interpreter-setting.json which represent the initialized
       // InterpreterSetting, while InterpreterSetting is from interpreter.json which represent
       // the user saved interpreter setting
       if (interpreterSettingTemplate != null) {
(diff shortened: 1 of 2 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -245,26 +245,11 @@ private void loadFromFile() throws IOException {
 
       InterpreterSetting interpreterSettingTemplate =
           interpreterSettingTemplates.get(savedInterpreterSetting.getGroup());
-      // InterpreterSettingTemplate is from interpreter-setting.json which represent the latest
+      // InterpreterSettingTemplate is from interpreter-setting.json which represent the initialized
       // InterpreterSetting, while InterpreterSetting is from interpreter.json which represent
       // the user saved interpreter setting
       if (interpreterSettingTemplate != null) {
-        savedInterpreterSetting.setInterpreterDir(interpreterSettingTemplate.getInterpreterDir());
```
liuxunorg:
```
savedInterpreterSetting.setInterpreterDir(interpreterSettingTemplate.getInterpreterDir());
```
This `setInterpreterDir()` property is not copied in `InterpreterSetting.java` above.
Can this line of code be deleted?

zjffdu: Correct, I should keep it. will fix it

PR #3513, opened 2019-11-06
[ZEPPELIN-4431]. submarine fail to build due to missing jar

zjffdu:
### What is this PR for?

Add repository in pom.xml so that it can download dependency pty4j

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4431

* CI pass

### Questions:

### DIFF
--- a/submarine/pom.xml
+++ b/submarine/pom.xml
@@ -130,6 +130,13 @@
     </dependency>
   </dependencies>
 
+  <repositories>
+    <repository>
+      <id>jerbrain</id>
+      <url>https://dl.bintray.com/jetbrains/pty4j/</url>
+    </repository>
+  </repositories>
+
   <build>
     <testResources>
       <testResource>


zjffdu:
@liuxunorg Could you help review it ?

submarine/pom.xml
```diff
@@ -130,6 +130,13 @@
     </dependency>
   </dependencies>
 
+  <repositories>
+    <repository>
+      <id>jerbrain</id>
+      <url>https://dl.bintray.com/jetbrains/pty4j/</url>
```
liuxunorg: `Pty4j` is the dependency used in the shell-terminal interpreter. The `pty4j` package should not be used in the submarine. Will it not compile now?

zjffdu: It looks like due to my environment issue, will close this PR

PR #3515, opened 2019-11-08
[ZEPPELIN-4396]. HvPlot is broken

zjffdu:
### What is this PR for?

HvPlot is broken due to bug in `ipython_server.py`. IPython may produce multiple types of output for one single python object. And the orders of these output matters, we should output html type over plain text type. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4396

* CI pass

image

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ matrix:
     # Test interpreter modules
     - jdk: "openjdk8"
       dist: xenial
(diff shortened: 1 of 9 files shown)


zjffdu:
Will merge if  no more comment

PR #3516, opened 2019-11-10
[ZEPPELIN-4383] Notebook Permalink Support

malayhm:
### What is this PR for?
Support permalink of a note so that it can be shared with others

### What type of PR is it?
Improvement

### Todos
* [ ] - Forward button should also change the active tab

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4383

* Manually - open note, it should change the URL to reflect the note id

NA

### Questions:

### DIFF
--- a/zeppelin-web-vue/src/App.vue
+++ b/zeppelin-web-vue/src/App.vue
@@ -9,7 +9,7 @@
           <LeftSidebar />
       </SplitArea>
       <SplitArea :size="75" id="content-pane" class="scrollbar">
-        <router-view />
+        <router-view :key="$route.path"/>
       </SplitArea>
     </Split>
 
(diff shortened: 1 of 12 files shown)


zjffdu:
Is the note link not permalink before this PR ?  Is the note link the same as now ?

malayhm:
@zjffdu 

Before this PR, there was no permalink support in vue.js implementation.
 > Is the note link not permalink before this PR ?

Today, we have http://localhost:8080/#/notebook/<note_id> as permalink in Angular implementation, but with vue implementation, it will be http://localhost:8080/#/<note_id>.
Should we change it to http://localhost:8080/#/notebook/<note_id> in this PR as well?
 > Is the note link the same as now ?

zjffdu:
Right, let's keep that the same as before, this is important for users to keep compatibility IMO.

PR #3523, opened 2019-11-18
[ZEPPELIN-4326][spark]Spark Interpreter restart failed when no suffic…

MOBIN-F:
### What is this PR for?
The pr is the last fix of the [ZEPPELIN-4326](https://issues.apache.org/jira/browse/ZEPPELIN-4326) The user can still restart the spark interpreter successfullywhen no sufficient yarn queue resources

   When the queue resource is insufficient, the spark task will remain in the ACCEPTED state. If the spark interpreter is restarted at this time, the restart interface will be stuck[**thread blocking  in the LazyOpenInterpreter.open method, this is a synchronized method**]. Restarting the spark interpreter fails. only use the yarn applicaiton -kill appID command or the kill -9 SparkSubmit process to restart the spark interpreter., which is very unfriendly to the user.
   At present, I use the system.exit(1) rough way to exit the JVM to achieve the purpose of restarting successfully, similar to the task of ending the ACCEPTED state by using Ctrl+C in the spark-shell.Spark only has **cancelJob** / **cancelStage** APIs,i thinks Spark is unable to do 

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4326

* Manually tested

image

### Questions:
* Add another ACCETTED state to tell the user that the task is in the ACCEPTED state instead of the running state?【PENDING -> ACCEPTED -> RUNNING】 
* For a spark task in the ACCEPTED state, the user can only restart the interpreter and cannot cancel with the cancel button.

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
@@ -142,9 +142,10 @@ public class SparkSqlInterpreter extends Interpreter {
   public void cancel(InterpreterContext context) throws InterpreterException {
     SparkInterpreter sparkInterpreter = getSparkInterpreter();
     SQLContext sqlc = sparkInterpreter.getSQLContext();
-    SparkContext sc = sqlc.sparkContext();
-
-    sc.cancelJobGroup(Utils.buildJobGroupId(context));
+    if (sqlc != null) {
+      SparkContext sc = sqlc.sparkContext();
+      sc.cancelJobGroup(Utils.buildJobGroupId(context));
+    }
   }
 
   @Override
@@ -156,7 +157,11 @@ public class SparkSqlInterpreter extends Interpreter {
   @Override
   public int getProgress(InterpreterContext context) throws InterpreterException {
     SparkInterpreter sparkInterpreter = getSparkInterpreter();
-    return sparkInterpreter.getProgress(context);
+    if (sparkInterpreter.getSparkContext() == null) {
+      return 0;
+    }else {
+      return sparkInterpreter.getProgress(context);
+    }
   }
 
   @Override
--- a/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
+++ b/spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
@@ -164,6 +164,8 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
     if (BaseSparkScalaInterpreter.sessionNum.decrementAndGet() == 0) {
       if (sc != null) {
         sc.stop()
+      }else {
+        System.exit(1)
       }
       if (sparkHttpServer != null) {
         sparkHttpServer.getClass.getMethod("stop").invoke(sparkHttpServer)
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/LazyOpenInterpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/LazyOpenInterpreter.java
@@ -20,6 +20,8 @@ package org.apache.zeppelin.interpreter;
 import java.net.URL;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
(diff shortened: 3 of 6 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -156,7 +157,11 @@ public FormType getFormType() {
   @Override
   public int getProgress(InterpreterContext context) throws InterpreterException {
     SparkInterpreter sparkInterpreter = getSparkInterpreter();
-    return sparkInterpreter.getProgress(context);
+    if (sparkInterpreter.getSparkContext() == null) {
+      return 0;
+    }else {
```
zjffdu: nit.  Add space after `}`

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -164,6 +164,8 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
     if (BaseSparkScalaInterpreter.sessionNum.decrementAndGet() == 0) {
       if (sc != null) {
         sc.stop()
+      }else {
```
zjffdu: nit. Add space after `}`

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
```diff
@@ -164,6 +164,8 @@ abstract class BaseSparkScalaInterpreter(val conf: SparkConf,
...
+        System.exit(1)
```
zjffdu: Could you add one line of logging to illustrate that spark interpreter is exited before `SparkContext` created

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/LazyOpenInterpreter.java
```diff
@@ -32,6 +34,7 @@
     implements WrappedInterpreter {
   private Interpreter intp;
   volatile boolean opened = false;
+  private Lock lock = new ReentrantLock();
```
zjffdu: Why use both lock and synchronized ?

MOBIN-F: I think it can be changed like this?
```
  public void open() throws InterpreterException {
    if (lock.tryLock()) {
      try {
        if (opened == true) {
          return;
        } else {
          intp.open();
          opened = true;
        }
      } finally {
        lock.unlock();
      }
    }
  }
```
But now I don't understand why the 0.8 version of LazyInterpreter.open needs two synchronized?

zjffdu: hmm, I think one synchronized is enough. 
It can be changed as 

```
  public void open() throws InterpreterException {
    if (opened == true) {
      return;
    }

    synchronized (intp) {
      if (opened == false) {
        intp.open();
        opened = true;
      }
    }
  }
```

Any reason using lock instead ?

MOBIN-F: [create sparkContext thread] and [restart sparkInterpreter thread] compete for the same synchronized lock. If the SparkContext is not created, the spark interpreter will restart.
1. The phenomenon that the restart intereperter interface is stuck
2. On the basis of step 1, the user may kill -9 SparkSubmit process. At this time, the synchronized lock is released. [restart sparkInterpreter thread] acquires the lock and re-creates the sparkContext. **The phenomenon of re-appearing a spark app after killied a spark app**
   Because getSparkInterpreter() and LazyOpenInterpreter.cancel will be called to LazyOpenInterpreter.open(),I used lock.tryLock() instead of synchronized

spark/spark-scala-parent/src/main/scala/org/apache/zeppelin/spark/BaseSparkScalaInterpreter.scala
MOBIN-F: I now think system.exit(1) is redundant. Because the sparkSubmit and remoteinterpreterServer processes are in the same JVM, only the remoteinterpreterServer can exit normally.

zjffdu: @MOBIN-F What do you mean ? Do you mean system.exit could not resolve this issue ?

PR #3542, opened 2019-12-06
[ZEPPELIN-4469]. Cron is disabled after zeppelin restart

zjffdu:
### What is this PR for?
The root cause is that zeppelin didn't init cron job setting after zeppelin restart. This PR add one thread to list all notes to init the cron job. The reason to do in a separated thread is that there may be many notes, loop all notes in the main thread may block the restarting of Zeppelin server.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4469

* Test it manually.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/scheduler/QuartzSchedulerService.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/scheduler/QuartzSchedulerService.java
@@ -55,15 +55,14 @@ public class QuartzSchedulerService implements SchedulerService {
     this.scheduler = new StdSchedulerFactory().getScheduler();
     this.scheduler.start();
 
-    Thread loadingNotesThread = new Thread(){
-      @Override
-      public void run() {
+    // Do in a separated thread because there may be many notes,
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/scheduler/QuartzSchedulerService.java
```diff
@@ -54,6 +54,19 @@ public QuartzSchedulerService(ZeppelinConfiguration zeppelinConfiguration, Noteb
     this.notebook = notebook;
     this.scheduler = new StdSchedulerFactory().getScheduler();
     this.scheduler.start();
+
+    Thread loadingNotesThread = new Thread(){
```
alexott: Wouldn't be easier to write it with lambda syntax `() -> ...` ?

zjffdu: Thanks for the review @alexott this comment is addressed

PR #3549, opened 2019-12-09
[ZEPPELIN-4476] Running Trash Restore will result in an error

eldiem:
### What is this PR for?
If you run a restore when there are two or more notes or folders, an error occurs.

Screen Shot 2019-12-09 at 10 19 15 PM

### What type of PR is it?
[Bug Fix]

### Todos
* [x] - Update note list after emptying the Trash
* [x] - Fixes an error when running Trash restore

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4476

(before)
Dec-09-2019 22-57-57
Dec-09-2019 23-13-49

(after)
Dec-09-2019 23-11-29
Dec-09-2019 22-55-39

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -17,7 +17,6 @@
 
 package org.apache.zeppelin.notebook;
 
-import com.google.common.collect.Sets;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -297,11 +298,13 @@ public void emptyTrash(AuthenticationInfo subject) throws IOException {
   public void restoreAll(AuthenticationInfo subject) throws IOException {
     NoteManager.Folder trash = noteManager.getTrashFolder();
     // restore notes under trash folder
-    for (NoteManager.NoteNode noteNode : trash.getNotes().values()) {
+    List<NoteNode> notes = trash.getNotes().values().stream().collect(Collectors.toList());
```
zjffdu: Could you add one line of comment why we need to do like this ? So that others won't refactor it in future

eldiem: I'll add a comment

zjffdu:
Thanks @eldiem LGTM, will merge if no more comment

PR #3552, opened 2019-12-14
[ZEPPELIN-4474]. Move r interpreter of spark into a separated interpreter

zjffdu:
### What is this PR for?

This PR move the r interpreter of spark into a separated interpreter `r`, that means user can use r in its native form rather than always need to in spark interpreter. 

### What type of PR is it?
[ Feature | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4474

* Ci pass

### Questions:


zjffdu:
Will merge if no more comments.

PR #3554, opened 2019-12-14
[ZEPPELIN-4489] remove zeppelin.anonymous.allowed configuration

prabhjyotsingh:
### What is this PR for?
This is a proposal for removing/disabling `zeppelin.anonymous.allowed` when Shiro is configured.
My Proposal in this JIRA is; let us remove this `zeppelin.anonymous.allowed`.
So, when Shiro is configured then Zeppelin server runs with all the auth configured, and if the user wants anonymous access then they should not configure Shiro.ini.

### What type of PR is it?
[Bug Fix | Improvement]

### What is the Jira issue?
* [ZEPPELIN-4489](https://issues.apache.org/jira/browse/ZEPPELIN-4489)

* CI should be green.

### Questions:
* Is there breaking changes for older versions? yes

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -427,12 +427,6 @@
 </property>
 
 <property>
-  <name>zeppelin.anonymous.allowed</name>
-  <value>true</value>
-  <description>Anonymous user allowed by default</description>
-</property>
-
-<property>
   <name>zeppelin.username.force.lowercase</name>
   <value>false</value>
   <description>Force convert username case to lower case, useful for Active Directory/LDAP. Default is not to change case</description>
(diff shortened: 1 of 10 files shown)


zjffdu:
Getting rid of `zeppelin.anonymous.allowed` make sense for me. This property is really confusing. 
+1

prabhjyotsingh:
Thanks for the review @zjffdu. Since this will change the way Zeppelin used to work before, where else should I change the doc? Or can that be handled at the time of release?

zjffdu:
Thanks for reminding, we should add document here http://zeppelin.apache.org/docs/0.8.2/setup/operation/upgrading.html

zjffdu:
LGTM, thanks @prabhjyotsingh , let's wait for the Ci pass

zjffdu:
CI is passed, thanks @prabhjyotsingh

prabhjyotsingh:
Sure, let go ahead and merge this.

PR #3556, opened 2019-12-17
[ZEPPELIN-4488]. Support Flink 1.10

zjffdu:
### What is this PR for?
This ticket is to support Flink 1.10, support blink planner and more rich sql function. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4488

* CI pass

### Questions:


bin/interpreter.sh
```diff
@@ -204,10 +204,13 @@ elif [[ "${INTERPRETER_ID}" == "pig" ]]; then
   fi
 elif [[ "${INTERPRETER_ID}" == "flink" ]]; then
```
alexott: #discussing Maybe we should refactor this into a separate functions loaded from an interpreter-specific scripts? It could be easier to understand the code of interpreter as we add more and more interpreters. Plus, this will make it easier to have external interpreters.

flink/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
```diff
@@ -176,13 +199,90 @@ private void callCommand(SqlCommandParser.SqlCommandCall cmdCall,
...
+    context.out.write("Table has been dropped.\n");
+  }
+
+  private void callUseCatalog(String catalog, InterpreterContext context) {
+    this.tbenv.useCatalog(catalog);
+  }
+
+  private void callShowModules() {
+
+  }
```
alexott: Forgotten implementation?

flink/src/main/java/org/apache/zeppelin/flink/FlinkStreamSqlInterpreter.java
```diff
@@ -37,51 +38,80 @@ public FlinkStreamSqlInterpreter(Properties properties) {
...
+        throw new IOException("type must be specified for stream sql");
+      }
+      if (streamType.equalsIgnoreCase("single")) {
+        SingleRowStreamSqlJob streamJob = new SingleRowStreamSqlJob(
+                flinkInterpreter.getStreamExecutionEnvironment(),
+                flinkInterpreter.getStreamTableEnvironment(),
+                flinkInterpreter.getJobManager(),
+                context,
+                flinkInterpreter.getDefaultParallelism());
+        streamJob.run(sql);
```
alexott: I'm thinking if we can generalize this using constructor references, etc.?

bin/interpreter.sh
zjffdu: I thought about it before, but haven't decided how to do that.

flink/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
zjffdu: Thanks, will fix it.

bin/interpreter.sh
alexott: I'll put it into my list of TODOs, to think about it...

zjffdu: Thanks @alexott  I believe we could do it in 0.10

zjffdu:
Will merge if no more comment

PR #3558, opened 2019-12-22
[ZEPPELIN-4497] Change zeppelin-jupyter-adapter to zeppelin-jupyter-interpreter

zjffdu:
### What is this PR for?
In ZEPPELIN-4480, we introduce zeppelin-jupyter-adapter which build a bridge between zeppelin and jupyter, zeppelin can extends this class to leverage any kernel of jupyter.
After a second thought, I think it is not enough. We can allow user to use jupyter kernel directly instead of implement one by one. So the purpose is to allow user use jupyter kernel like this

```
%jupyter(kernel=ipython)

python_code
...
```

Of course user can enhance the jupyter kernel in Zeppelin, e.g. IPythonInterpreter of Zeppelin can enhance the ipython kernel by adding support of ZeppelinContext

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4497#

* CI pass

### Questions:

image

image

image


zjffdu:
Will merge if no more comments since this is a refactoring PR, no anything new is introduced.

PR #3564, opened 2019-12-24
[ZEPPELIN-4504] pre-index notebooks in the background to fix search ability

amakaur:
### What is this PR for?
Before this PR zepp was unable to search for keywords in existing notebooks that have not been updated in any way after a restart.

This PR pre-indexes all of the notebooks at startup time on a separate thread so that zepp is able to find keywords in existing notebooks whenever user uses the search bar.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4504

- Create a notebook using Spark interpreter
- add small piece of code such as "println("hello")
- Restart zeppelin
- Search for a keyword "println"

### Questions:


zjffdu:
@amakaur We should first try to figure out why the indexing is lost after restart. Is it due to the index in memory or other bugs ? Reindexing works, but it is some kind of workaround, not an elegant solution for this issue.

amakaur:
@zjffdu in the previous versions pre-index was done at startup time so looks like in 0.9.0 pre-index ONLY happens when any kind of update happens to a note.

PR #3569, opened 2019-12-26
[ZEPPELIN-4512]. Move SqlSplitter from jdbc interpreter to zeppelin-interpreter

zjffdu:
### What is this PR for?

This PR move the sql split logic from jdbc interpreter to zeppelin-interpreter, so that other interpreter can reuse it. Besides that, it also fix some bugs of sql splitter, and add more unit test to verify it. 

### What type of PR is it?
[Bug Fix | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://jira.apache.org/jira/browse/ZEPPELIN-4512

* CI pass

### Questions:


zjffdu:
Thanks for the comment @alexott 
I have updated the PR to allow adding additional single line comment prefix. I am not sure whether's any other places needs to be customized for sql dialect.

zjffdu:
Thanks @alexott Will merge if no more comment

PR #3591, opened 2020-01-07
[ZEPPELIN-4536] Select name when creating a node

frankivo:
### What is this PR for?
Making life a little easier when creating notes into folders, by selecting only the filename rather than the whole path.

### What type of PR is it?
Improvement 

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4536

* Test if you can create a note on the rootnode
* Test if you can create a note on deeper levels

image

### Questions:
  No
  No
  No


zjffdu:
LGTM, thanks for the contribution @frankivo

PR #3593, opened 2020-01-09
[ZEPPELIN-4543]. Support Shiny in Spark Interpreter

zjffdu:
### What is this PR for?
We support shiny in R interpreter in ZEPPELIN-4525, this ticket is to extend it in SparkInterpreter where R is also supported. Writing shiny app in SparkInterpreter is almost the same as do it in R interpreter, but also with additional support of Spark. Main thing is in `SparkShinyInterpreter` which extends `ShinyInterpreter` 

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4543

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #3609, opened 2020-01-19
[ZEPPELIN-4555]. paragraph title is disappear after I run the paragraph

zjffdu:
### What is this PR for?
This PR is to fix the title disappear issue.  The main change is in method `mergeConfig` where I would always insert the config in `interpreter-setting.json`

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4555

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
@@ -1108,12 +1108,12 @@ public class NotebookTest extends AbstractInterpreterTest implements ParagraphJo
     // create paragraphs
     Paragraph p1 = note.addNewParagraph(anonymous);
     Map<String, Object> config = p1.getConfig();
-    assertTrue(config.containsKey(Paragraph.PARAGRAPH_CONFIG_RUNONSELECTIONCHANGE));
-    assertTrue(config.containsKey(Paragraph.PARAGRAPH_CONFIG_TITLE));
-    assertTrue(config.containsKey(Paragraph.PARAGRAPH_CONFIG_CHECK_EMTPY));
-    assertEquals(config.get(Paragraph.PARAGRAPH_CONFIG_RUNONSELECTIONCHANGE), false);
-    assertEquals(config.get(Paragraph.PARAGRAPH_CONFIG_TITLE), true);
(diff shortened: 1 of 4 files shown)


zjffdu:
@liuxunorg Could you help review it ? Thanks

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -1123,29 +1123,35 @@ public void testInterpreterSettingConfig() {
...
-    note.runAll(AuthenticationInfo.ANONYMOUS, false);
+    p1.execute(true);
 
-    // wait until first paragraph finishes and second paragraph starts
-    while (p1.getStatus() != Status.FINISHED) Thread.yield();
+    // mock1 has no config setting in interpreter-setting.json, so keep the previous config
+    LOGGER.info("changed intp p1.getConfig() =  " + p1.getConfig());
+    assertEquals(config.get(InterpreterSetting.PARAGRAPH_CONFIG_RUNONSELECTIONCHANGE), false);
+    assertEquals(config.get(InterpreterSetting.PARAGRAPH_CONFIG_TITLE), true);
+    assertEquals(config.get(InterpreterSetting.PARAGRAPH_CONFIG_CHECK_EMTPY), false);
```
liuxunorg: `config_test` interpreter's setting checkEmpty=false
```
      "config": {
        "fontSize": 9,
        "colWidth": 12,
        "runOnSelectionChange": false,
        "title": true,
        "checkEmpty": false
      },
```

Change paragraph test to `p1.setText("%mock1 sleep 1000");`, 
But `mock1` interpreter's unsetting checkEmpty.
`InterpreterSetting.PARAGRAPH_CONFIG_CHECK_EMTPY` Should the default value be equal to `true`??

zjffdu: @liuxunorg The problem here we could not override the config with the config in interpreter-setting. Basically config could from 3 sources when user change interpreter in one paragraph:
* the old interpreter's default config
* the new interpreter's default config
* config made by user manually

Actually we could not decide where the config come from, that's why I will keep the previous config which I think is the best approach to handle this https://github.com/bbonnin/zeppelin-mongodb-interpreter

zjffdu:
Thanks @liuxunorg

PR #3617, opened 2020-01-22
[ZEPPELIN-4578] Cleanup of LoginRestApi

lfrancke:
### What is this PR for?
I've had to debug a few Zeppelin issues and looked at the LoginRestApi code.

It's using raw parameterized classes all over the place and a few other weird things but what really made me look is that every log in attempt (succesful or not) is logged as WARN.

I decided to fix the warnings my IDE shows me and changed the Log statements from WARN to INFO

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4578

This does not change any actual functionality so I rely on existing tests. 
Travis CI ran and there were a bunch of failures but they all look unrelated (Livy, Timeouts, ...)

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
@@ -16,11 +16,9 @@
  */
 package org.apache.zeppelin.rest;
 
-import com.google.gson.Gson;
 import java.text.ParseException;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -100,43 +100,42 @@ public Response getLogin(@Context HttpHeaders headers) {
...
+        LOG.error("Error in Login", e);
       }
     }
-    return new JsonResponse(Status.METHOD_NOT_ALLOWED).build();
+    return new JsonResponse<>(Status.METHOD_NOT_ALLOWED).build();
   }
 
   private KerberosRealm getKerberosRealm() {
-    Collection realmsList = authenticationService.getRealmsList();
+    Collection<Realm> realmsList = (Collection<Realm>) authenticationService.getRealmsList();
```
zjffdu: It's better to change the signature of AuthenticationService instead of using type cast.

lfrancke: I'd be happy to do that but hesitated because that's a public API and I didn't know Zeppelins guarantees about changing those.

zjffdu: I think it is fine, just go ahead.

lfrancke:
I'm fairly certain that the build failure is unrelated.

zjffdu:
Thanks @lfrancke LGTM

lfrancke:
Thanks for taking a look and committing.

PR #3620, opened 2020-01-31
[ZEPPELIN-4583] Fix artifact resolving issues

rimolive:
### What is this PR for?
Fixes the problem with install-interpreter script when it tries to resolve Maven artifacts in sonatype, in which only accepts https connections recently.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4583

* Build a zeppelin distribution with no interpreters installed
* Run install-interpreter script to install any interpreter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,10 @@ Thumbs.db
 .idea/
 *.iml
 
+# VSCode project files
+.vscode/
+**/.factorypath
+
 # maven target files
 target/
 **/target/
--- a/zeppelin-interpreter/src/test/java/org/apache/zeppelin/dep/DependencyResolverTest.java
+++ b/zeppelin-interpreter/src/test/java/org/apache/zeppelin/dep/DependencyResolverTest.java
@@ -87,10 +87,9 @@ public class DependencyResolverTest {
     FileUtils.cleanDirectory(testCopyPath);
 
     // load from added repository
-    resolver.addRepo("sonatype",
-        "https://oss.sonatype.org/content/repositories/ksoap2-android-releases/", false);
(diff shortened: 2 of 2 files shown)


zjffdu:
Thanks for the fix @rimolive 
I think the changes in `zeppelin-web/package-lock.json` is not necessary ?

rimolive:
Yes, I think. The code is a bit messed in my local repo, but I'll remove this file.

gerlowskija:
Would love to see this get in.  I work with a 3rd party interpreter for Solr called [zeppelin-solr](https://github.com/LucidWorks/zeppelin-solr), and this bug blocks us from installing it anywhere.  Been a big pain for our team.

Actually, now that I think of it, even when this gets merged we would have to drop a jar into our Zeppelin install or wait for a whole new release since the URL here is hardcoded in Java-land.

Do you guys know of any workaround in the interim?

rimolive:
@gerlowskija The only workaround I found is to apply these changes in a local repo and build zeppelin-interpreter code. Unfortunately, this change affect Java classes that must be compiled.

If you are using zeppelin 0.8.2, don't forget that `conf/interpreter-list` file must be changed to download and install 0.8.2 artifacts (currently it is downloading 0.8.0).

zjffdu:
@rimolive No other changes besides `.gitignore` ?

zjffdu:
ping @rimolive

rimolive:
I'm still looking at it. Something strange has happened.

rimolive:
@zjffdu I think someone already made the changes in Booter.java to use https, which makes this PR useless. I'm going to close it but looking forward for other ways to contribute with the project.

zjffdu:
Thanks @rimolive for the investigation.

PR #3633, opened 2020-02-09
unable to connect presto and hive through presto jdbc interpreter

fred521:
### What is this PR for?
when executed a sql query through the zeppelin presto jdbc interpreter in notebook，couldn't get a correct result, there was always sqlException printed in the log even though the completely correct configuration of this inerpreter, hope some others have good iedas and can resolve this bug of the presto jdbc interpreter

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4601

Pass CI 
and functional testing

### Questions:


zjffdu:
@fred521 Could you explain more about how this fix the issue ?

fred521:
Hi @zjffdu 
I think zeppelin jdbc interpreter default 'url' in the properties, but is invalid in the presto jdbc driver
see the [PrestoDriverUri.java] (https://github.com/prestodb/presto/blob/303e640f5da27a8c3785aae2f4129fa2293d062f/presto-jdbc/src/main/java/com/facebook/presto/jdbc/PrestoDriverUri.java#L324) 
and [ConnectionProperties.java](https://github.com/prestodb/presto/blob/303e640f5da27a8c3785aae2f4129fa2293d062f/presto-jdbc/src/main/java/com/facebook/presto/jdbc/ConnectionProperties.java#L58) 
the same as the hive in the jdbc interpreter.
see the jira exception https://issues.apache.org/jira/browse/ZEPPELIN-4601

PR #3635, opened 2020-02-10
[ZEPPELIN-4568] Fix can not update output when the data is empty

hsuanxyz:
### What is this PR for?

Fix can not update output when the data is empty

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-4568

Try the following code:

```pythone
import sys
import time
from IPython.display import display, clear_output
for i in range(10):
    time.sleep(0.25)
    clear_output(wait=True)
    print(i)
    sys.stdout.flush()
```

Should be updated instead of append the output

Kapture 2020-02-10 at 17 07 46

### Questions:


Leemoonsoo:
Thanks @hsuanxyz. LGTM

Leemoonsoo:
Merge to master

PR #3636, opened 2020-02-11
[ZEPPELIN-4597]. Deprecate r module

zjffdu:
### What is this PR for?

This PR just deprecate the existing the r module, but keep the source code until someone pick it up. We may delete it later if no one would like to take over it.

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4597

* CI pass

### Questions:

### DIFF
--- a/dev/publish_release.sh
+++ b/dev/publish_release.sh
@@ -46,7 +46,7 @@ if [[ $RELEASE_VERSION == *"SNAPSHOT"* ]]; then
   DO_SNAPSHOT="yes"
 fi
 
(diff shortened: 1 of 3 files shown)


alexott:
Just general comment - should we have a document that will describe "bigger" changes, like, removal or addition of interpreters, etc.?

zjffdu:
Make sense, thanks for the suggestion. @alexott

PR #3639, opened 2020-02-11
[ZEPPELIN-4598]. upgrade-note.sh GitNotebookRepo cannot be cast to OldNotebookRepo

zjffdu:
### What is this PR for?

This PR fix the issue that `OldGitNotebookRepo` not loaded correctly because we still try to load `GitNotebookRepo`

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4598

* Manually tested

### Questions:


zjffdu:
Someone has verified it (see jira comment). Will merge it if no more comments.

PR #3643, opened 2020-02-13
[ZEPPELIN-4611]. Fetching rows with newline character (\n) breaks entire table

zjffdu:
### What is this PR for?
This PR would replace all the special characters(\t, \n, \r\n) in table content with white space. so that it won't break the table format display.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4611

* Unit test is added

### Questions:


nephtyws:
LGTM. Thanks for the lightning-fast fix!

zjffdu:
Thanks for verification @nephtyws

PR #3356, opened 2019-04-25
[ZEPPELIN-3994] Notebook serving

Leemoonsoo:
### What is this PR for?
This PR implements [notebook serving](https://issues.apache.org/jira/browse/ZEPPELIN-3994) explained [here](https://docs.google.com/document/d/1YA6q8W9yO8a88xzLDYs9zv_fKu2_cnB58rmQbakxi1I/edit?usp=sharing).

While this PR is already quite large, front-end implementation will have separate PR.
This PR is work in progress.

### How to run 

1. Prepare kubernetes cluster
2. Prepare persistentVolumeClaim with name 'task-context-volume-claim' in your kubernetes cluster
3. Build Zeppelin docker image (follow instructions [here](https://github.com/apache/zeppelin/blob/master/docs/quickstart/kubernetes.md)
4. Deploy api router by running `kubectl apply -f k8s/serving-api-router/serving-api-router.yaml`
5. Start Zeppelin on Kubernetes by running `kubectl apply -f k8s/zeppelin-server.yaml`
6. Port forward to the api router (e.g. `kubectl port-forward <zeppelin-serving-api-router pod name> 8080:80`) and browse `localhost:8080`

 *Note: This branch does not include front-end implementation. To try with front-end, checkout [this branch](https://github.com/Leemoonsoo/zeppelin/pull/9).

### Summary of changes
To be updated

### What type of PR is it?
Feature

### What is the Jira issue?
 https://issues.apache.org/jira/browse/ZEPPELIN-3994

### How should this be tested?
* First time? Setup Travis CI as described on https://zeppelin.apache.org/contribution/contributions.html#continuous-integration
* Strongly recommended: add automated unit tests for any new or changed behavior
* Outline any manual steps to test the PR here.

### Questions:
* Does the licenses files need update? yes
* Is there breaking changes for older versions? no
* Does this needs documentation? yes

PR #3654, opened 2020-02-17
[ZEPPELIN-4619] Run a note from the commandline

Leemoonsoo:
### What is this PR for?
This PR adds a command-line option to run a single note.
The code is pickled from https://github.com/apache/zeppelin/pull/3356.

Usage is

```
bin/zeppelin.sh --run <noteId>
```

The command will launch a Zeppelin server and run a given note, and terminate on after all paragraph's successful run (with exit 0) or terminate on any paragraph error (with exit 1).

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4619

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
@@ -16,9 +16,12 @@
  */
 package org.apache.zeppelin.server;
 
+import com.google.gson.Gson;
 import java.io.File;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
+import java.util.Base64;
+import java.util.HashSet;
 import java.util.List;
 import java.util.EnumSet;
 import java.util.Objects;
@@ -30,6 +33,7 @@ import javax.servlet.DispatcherType;
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 import org.apache.commons.lang.StringUtils;
+import org.apache.directory.api.util.Strings;
 import org.apache.shiro.web.env.EnvironmentLoaderListener;
 import org.apache.shiro.web.servlet.ShiroFilter;
 import org.apache.zeppelin.cluster.ClusterManagerServer;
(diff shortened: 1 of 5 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -326,6 +336,46 @@ private static void initServerConnector(Server server, int port, int sslPort) {
     server.addConnector(connector);
   }
 
+  private static void runNoteOnStart(ZeppelinConfiguration conf) throws IOException, InterruptedException {
+    String noteIdToRun = conf.getNotebookRunId();
```
zjffdu: It's better to add logging here to show zeppelin server is running note after started.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -344,6 +344,42 @@ public boolean runParagraph(String noteId,
     }
   }
 
+  public boolean runNote(String noteId,
```
zjffdu: You can use `runAllParagraphs` https://github.com/apache/zeppelin/blob/master/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java#L347
Or maybe rename it to `runNote`

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
Leemoonsoo: Log message added.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
Leemoonsoo: I modified existing `runAllParagraphs` to return boolean value depends (that will be translated to exit value of the commandline). And added overloaded method to make it run without providing paragraph data. What do you think?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -344,19 +349,42 @@ public boolean runParagraph(String noteId,
...
+
+    return runAllParagraphs(
+            noteId,
+            rawParagraphs,
+            context,
+            callback
+    );
+  }
+
+  public boolean runAllParagraphs(String noteId,
```
zjffdu: How about using the existing runAllParagraphs, and check whether `paragraphs`. If it is empty, then use the paragraph in backend, otherwise use the paragraphs from frontend.

Leemoonsoo: Addressed the comment. Please take a look updated code.

zjffdu:
Thanks @Leemoonsoo LGTM

Leemoonsoo:
I removed a useless test. Thanks, @zjffdu for the review. I'll merge to master.

PR #3655, opened 2020-02-18
[ZEPPELIN-4620]. Exported note json is not pretty print

zjffdu:
### What is this PR for?
The exported note (both zpln and ipynb) are not pretty print. This PR fix this is straightforward.
For zpln, it is done in browser side, for ipynb it is done in server side because the conversion happens in server side. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4620

* Tested it manually

### Questions:


Leemoonsoo:
LGTM

zjffdu:
Thanks @Leemoonsoo  will merge it if no more comment

PR #3658, opened 2020-02-21
[ZEPPELIN-4631]. Don't save path in note

zjffdu:
### What is this PR for?
This PR just make path in Note.java as `transient`, so it won't be saved in note content. And every time it can be constructed from NotebookRepo when reading it in NoteManager which is the only way to load Note.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4631

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

zjffdu:
CI pass, will merge if no more comment

PR #3659, opened 2020-02-25
[ZEPPELIN-4640]. Save note after saving interpreter binding setting

zjffdu:
### What is this PR for?

This is a trivial PR which just add one line of code to save note after saving interpreter binding setting.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4640

* Tested manually

### Questions:


zjffdu:
Will merge if no more comment

PR #3661, opened 2020-02-26
[ZEPPELIN-4643]. Fail to start zeppelin server because zeppelin-interpreter-shaded is under lib

zjffdu:
### What is this PR for?

This PR just exclude zeppelin-interpreter-shaded from `zeppelin-markdown` in `zeppelin-jupyter/pom.xml`. So that we won't include it under lib when building zeppelin distribution. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4643

* Tested manaully

### Questions:


zjffdu:
I will merge it soon, not only because it is simple but also it fix the travis CI fail issue

PR #3663, opened 2020-02-26
[ZEPPELIN-4645]. Only list the interpreters that the user has permission in creating note and interpreter binding

zjffdu:
### What is this PR for?

It doesn't make sense to list the interpreters that user don't have permission. This PR just filter the interpreters that user don't have permission in the creating note page and interpreter binding section. But I would always add the default interpreter in the interpreter binding section without checking its permission, because it would make user confused if the default interpreter is lost after interpreter permission is changed. 

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4645

* CI pass and tested manually

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -27,6 +27,7 @@ import com.google.gson.reflect.TypeToken;
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


weand:
@zjffdu 

According to the PR description it's not clear whats happens to the currently selected default interpreter in interpreter binding section, in case the user has no permission for the default interpreter.

From my point of view, the default interpreter should not be changed automatically due to missing permission of any user trying to run the note. Instead the default interpreter should remain visible and remain the default. In case the user has no access, the current error message should be presented, when the user will try to run the note.

What do you think?

zjffdu:
Thanks for the comment @weand Actually I would always add default interpreter in the interpreter binding without checking its permission. Sorry for the unclear PR description, I will update it.

zjffdu:
ping @weand

weand:
yeah thats fine.

zjffdu:
Thanks @weand will merge if no more comment

PR #3668, opened 2020-02-29
[ZEPPELIN-4635]. Save note permission info into notebook-authorization.json

zjffdu:
### What is this PR for?

This PR is to revert the changes of ZEPPELIN-3985. Because it would cause the issue of ZEPPELIN-4612. In this PR I will make Zeppelin still save note permission info into `notebook-authorization.json`. But I also do some code refactoring, and we could store the permission info into other storage such as database in future. Because storing them into one file also has potential issue. such as scale issue, now each time we have to write all notes' permission info into file instead of in the note level.  

### What type of PR is it?
[Bug Fix | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4635

* CI pass

### Questions:


zjffdu:
@Leemoonsoo You might want to know this change.

Leemoonsoo:
Thanks @zjffdu LGTM

zjffdu:
Thanks @Leemoonsoo will merge if no more comments

jolo-dev:
Hi @zjffdu,
I am aware of the loading issue but it causes troubles on our end.
We have Zeppelin running on ECS and whenever we redeploy, the `notebook-authorization.json` is gone.
I guess a compromise needs to be handled.

zjffdu:
@jolo-dev What version of zeppelin do you use ? It is a bug if the notebook-authorization.json is gone.

jolo-dev:
@zjffdu We upgraded to zeppelin-preview2.
But that's not a bug. It's just the `notebook-authorization.json` is not in the notebook storage but in a separate file.
So the best thing would be to put the permissions directly to note.

zjffdu:
I see, actually ConfigStorage has 2 implementation: 
* LocalConfigStorage  (store notebook-authorization.json in local filesystem)
* FileSystemConfigStorage. (store notebook-authorization.json in hadoop compatible file system)

I think you can use FileSystemConfigStorage to resolve your issue.

jolo-dev:
Well, we are not using hadoop. Is it possible to read this file from an S3 bucket?

PR #3672, opened 2020-03-01
[ZEPPELIN-4661]. Fail to run whole note

zjffdu:
### What is this PR for?

This is to fix the regression issue caused by ZEPPELIN-4661. The root cause is that in the frontend we use `paragraph` to represent the script text while in the backend we use `text`, so that converting between frontend data to backend class `Paragraph` will cause this kind of error. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4661

* CI pass and manually tested

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -86,7 +86,6 @@ public class NotebookService {
   private Notebook notebook;
   private AuthorizationService authorizationService;
   private SchedulerService schedulerService;
-  private Gson gson = new Gson();
 
   @Inject
   public NotebookService(
@@ -293,6 +292,7 @@ public class NotebookService {
                               ServiceContext context,
                               ServiceCallback<Paragraph> callback) throws IOException {
 
+    LOGGER.info("Start to run paragraph: " + paragraphId + " of note: " + noteId);
     if (!checkPermission(noteId, Permission.RUNNER, Message.OP.RUN_PARAGRAPH, context, callback)) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


zjffdu:
@Leemoonsoo Could you help review it ? Thanks

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -376,25 +375,33 @@ public boolean runAllParagraphs(String noteId,
...
+            continue;
+          }
+          String text = (String) raw.get("paragraph");
+          String title = (String) raw.get("title");
+          Map<String, Object> params = (Map<String, Object>) raw.get("params");
+          Map<String, Object> config = (Map<String, Object>) raw.get("config");
+
+          if (!runParagraph(noteId, paragraphId, title, text, params, config, false, true,
+                  context, callback)) {
+            // stop execution when one paragraph fails.
```
alexott: I maybe would add more logging around `noteId` and `paragraphId` into the `runParagraph`. WDYT?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -376,25 +375,33 @@ public boolean runAllParagraphs(String noteId,
...
     try {
-      for (Map<String, Object> raw : paragraphs) {
-        String paragraphId = (String) raw.get("id");
-        if (paragraphId == null) {
-          continue;
+      if (paragraphs != null) {
+        // run note via the data passed from frontend
+        for (Map<String, Object> raw : paragraphs) {
+          String paragraphId = (String) raw.get("id");
+          if (paragraphId == null) {
```
alexott: maybe add a debug message here?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -376,25 +375,33 @@ public boolean runAllParagraphs(String noteId,
...
+            continue;
+          }
+          String text = (String) raw.get("paragraph");
+          String title = (String) raw.get("title");
+          Map<String, Object> params = (Map<String, Object>) raw.get("params");
+          Map<String, Object> config = (Map<String, Object>) raw.get("config");
```
alexott: Should we wrap this into `try`, so if the input data have incorrect type, we handle this gracefully...

Leemoonsoo:
Thanks @zjffdu for addressing it. It looks good to me!

zjffdu:
Thanks @Leemoonsoo  And @alexott I have addressed your comment, do you have any other comment ? Thanks

zjffdu:
Thanks @alexott , I will merge it soon

PR #3678, opened 2020-03-06
[ZEPPELIN-4479]. Don't use helium url in s3 by default

zjffdu:
### What is this PR for?
Trivial PR which just remove helium s3 url by default. This could solve zeppelin starting issue when users don't have internet access.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4479

* CI pass

### Questions:

### DIFF
--- a/docs/development/helium/overview.md
+++ b/docs/development/helium/overview.md
@@ -37,3 +37,7 @@ Currently, Helium supports 4 types of package.
 - [Helium Interpreter](../writing_zeppelin_interpreter.html): Adding a new custom interpreter
 
 
+## Configuration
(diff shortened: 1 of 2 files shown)


zjffdu:
@alexott I am afraid they need to manually configure `zeppelin.helium.registry` in zeppelin-site.xml to point to the s3 helium registry file. Thank for your remind, I think I need to update document here as well.

zjffdu:
Helium document is updated as well.

PR #3683, opened 2020-03-09
[ZEPPELIN-4672]. Display interpreter property description in interpreter setting page

zjffdu:
# What is this PR for?

This PR is just to display property description in interpreter setting page, so that user can easily see what does each property mean when they configure the interpreter. It also handles the upgrade scenario where there's no property description in `interpreter.json`. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4672

* Manually tested

Before
image

After
image

### Questions:


zjffdu:
Thanks for the review @alexott  will merge it soon

PR #3684, opened 2020-03-11
ZEPPELIN-4679: Add API validation for creating interpreter

prabhjyotsingh:
### What is this PR for?
This is an extension of ZEPPELIN-4377, I believe we should have an API validation for the same.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4679

Calling directly API should fail, and if somehow this is by pass from UI, UI should also return proper error.
```
curl 'http://localhost:8080/api/interpreter/setting' -H 'Content-Type: application/json;charset=UTF-8' --data-binary '{"name":"test space","group":"angular","properties":{},"dependencies":[],"option":{"remote":true,"isExistingProcess":false,"setPermission":false,"session":false,"process":false,"perNote":"shared","perUser":"shared","owners":[]},"propertyValue":"","propertyKey":"","propertyType":"textarea"}' 
```

N/A

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
@@ -29,6 +29,8 @@ import com.google.gson.reflect.TypeToken;
 import java.util.Properties;
 import java.util.LinkedHashMap;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
@@ -767,9 +769,17 @@ public class InterpreterSettingManager implements NoteEventListener, ClusterEven
                                                     InterpreterOption option,
                                                     Map<String, InterpreterProperty> properties)
       throws IOException {
-    if (name.indexOf(".") >= 0) {
-      throw new IOException("'.' is invalid for InterpreterSetting name.");
+    if (name == null) {
+      throw new IOException("Interpreter name shouldn't be empty.");
     }
+
+    // check if name is valid
+    Pattern pattern = Pattern.compile("^[-_a-zA-Z0-9]+$");
+    Matcher matcher = pattern.matcher(name);
+    if(!matcher.find()){
+      throw new IOException("Interpreter name shouldn't be empty, and can consist only of: -_a-zA-Z0-9");
+    }
+
     // check if name is existed
     for (InterpreterSetting interpreterSetting : interpreterSettings.values()) {
       if (interpreterSetting.getName().equals(name)) {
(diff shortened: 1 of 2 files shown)


prabhjyotsingh:
@alexott,  @zjffdu  can you help review this?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -767,9 +769,17 @@ private InterpreterSetting inlineCreateNewSetting(String name, String group,
...
                                                     Map<String, InterpreterProperty> properties)
       throws IOException {
-    if (name.indexOf(".") >= 0) {
-      throw new IOException("'.' is invalid for InterpreterSetting name.");
+    if (name == null) {
+      throw new IOException("Interpreter name shouldn't be empty.");
     }
+
+    // check if name is valid
+    Pattern pattern = Pattern.compile("^[-_a-zA-Z0-9]+$");
```
alexott: I would move the pattern into static constant to avoid its compilation on every call....

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -767,9 +769,17 @@ private InterpreterSetting inlineCreateNewSetting(String name, String group,
...
+    Matcher matcher = pattern.matcher(name);
+    if(!matcher.find()){
+      throw new IOException("Interpreter name shouldn't be empty, and can consist only of: -_a-zA-Z0-9");
```
alexott: I'm not 100% sure that we should use `IOException` here, it's more like "Invalid parameters, or something like". But maybe it will require big changes along the call chain to use something else, so I'm ok with it...

prabhjyotsingh: Yes, lazy me, I looked up what other APIs are using, and have used the same.

alexott: I have a TODO list to go through all places & fix this :-)

zjffdu: Regarding this, we do need to revisit that, and could do it in other PR

zjffdu:
@prabhjyotsingh We should also add unit test for `InterpreterSettingManager`

prabhjyotsingh:
@zjffdu  sure that makes sense, have added it.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
prabhjyotsingh: Sure let me do that.

prabhjyotsingh:
@zjffdu @alexott  does these changes looks ok?

zjffdu:
LGTM

prabhjyotsingh:
Thank you for the review will merge this to master.

PR #3694, opened 2020-03-19
[ZEPPELIN-2891]. Impossible to use jdbc interface with presto-jdbc >=0.180

zjffdu:
### What is this PR for?

This PR is to fix the issue of use presto via jdbc interpreter. In this PR, I would only add properties that is valid for Presto jdbc driver. I tested it manully on the 2 versions of presto: prestodb, prestosql.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-2891

* CI pass and manually tested on prestodb and prestosql.

image

image

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -141,7 +141,16 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private static final String DBCP_STRING = "jdbc:apache:commons:dbcp:";
   private static final String MAX_ROWS_KEY = "zeppelin.jdbc.maxRows";
 
-  private final HashMap<String, Properties> basePropretiesMap;
+  private static final Set<String> PRESTO_PROPERTIES = new HashSet<>(Arrays.asList(
+          "user", "password",
+          "socksProxy", "httpProxy", "clientTags", "applicationNamePrefix", "accessToken",
+          "SSL", "SSLKeyStorePath", "SSLKeyStorePassword", "SSLTrustStorePath",
+          "SSLTrustStorePassword", "KerberosRemoteServiceName", "KerberosPrincipal",
+          "KerberosUseCanonicalHostname", "KerberosServicePrincipalPattern",
+          "KerberosConfigPath", "KerberosKeytabPath", "KerberosCredentialCachePath",
+          "extraCredentials", "roles", "sessionProperties"));
+
+  private final HashMap<String, Properties> basePropertiesMap;
   private final HashMap<String, JDBCUserConfigurations> jdbcUserConfigurationsMap;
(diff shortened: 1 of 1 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -407,7 +408,33 @@ private void setUserProperty(String propertyKey, InterpreterContext interpreterC
...
+      Properties newProperties = new Properties();
+      // Only add valid properties otherwise presto won't work.
+      try {
+        Class clazz = null;
+        try {
+          clazz = Class.forName("com.facebook.presto.jdbc.ConnectionProperties");
+        } catch (ClassNotFoundException e) {
+          clazz = Class.forName("io.prestosql.jdbc.ConnectionProperties");
+        }
+        Method lookUpKeyMethod = clazz.getMethod("forKey", String.class);
```
electrum: Please don’t access internals of the driver with reflection. We can change this at any time in the future which will cause this code to break.

You can find the list of supported properties here: https://prestosql.io/docs/current/installation/jdbc.html

The important thing is to not add random properties like “url”. Zeppelin should not be adding properties unless it knows they are valid or they come from an end user. (If the user adds an invalid property, that’s on the user, and we don’t need to protect them from themselves.)

zjffdu: Thanks for the comment , will address it

zjffdu: @elec

> Please don’t access internals of the driver with reflection. We can change this at any time in the future which will cause this code to break.
> 
> You can find the list of supported properties here: https://prestosql.io/docs/current/installation/jdbc.html
> 
> The important thing is to not add random properties like “url”. Zeppelin should not be adding properties unless it knows they are valid or they come from an end user. (If the user adds an invalid property, that’s on the user, and we don’t need to protect them from themselves.)

One concern I have is whether presto would change these valid properties. Because we would like to support multiple versions of presto. So if presto introduce new properties or remove some existing properties, that would be a problem for us.

electrum: I don’t foresee us removing properties, as we take backwards compatibility seriously. I believe the main issue here was not validation of arbitrary properties, but that Zeppelin was adding its own properties like “url”. If we can avoid that, then we shouldn’t need to worry about validation.

zjffdu: Thanks for your conform

qiguoxiaosheng:
A PR is already submitted. #3712

qiguoxiaosheng:
A PR is already submitted. #3712

PR #3698, opened 2020-03-26
[ZEPPELIN-4694]. Limit the number of running interpreter process

zjffdu:
### What is this PR for?
This is to limit the number of running interpreter process to protect zeppelin machine ran out of memory. By default, you can run at most 10 interpreter processes. To be noticed, this won't impact k8s mode. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4694

* CI pass and unit test is added

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
```diff
@@ -96,6 +100,11 @@ public int getPort() {
 
   @Override
   public void start(String userName) throws IOException {
+
+    if (TOTAL_RUNNING_PROCESS >= MAX_RUNNING_PROCESS) {
+      throw new IOException("You have reached the max number of running interpreter process: "
```
alexott: it would be nice to change `interpreter process` to `interpreter processes`

jongyoul:
I don't think it's a critical issue but how about using `AtomicInteger` to count the number of processes?

PR #3707, opened 2020-03-31
[ZEPPELIN-4700] Cleanup Maven warnings & unify plugin management

alexott:
### What is this PR for?

This commit declares all plugins in the `pluginManagement` section and unifies versions of
all plugins - they are now declared as properties in the top-level `pom.xml`.  All other
`pom.xml` files shouldn't specify version explicitly.

### What type of PR is it?
Improvement

### What is the Jira issue?

* ZEPPELIN-4700

* One build was clean: https://travis-ci.org/github/alexott/zeppelin/builds/668812389, but after removing the not necessary files from commit, some tests began to fail: https://travis-ci.org/github/alexott/zeppelin/builds/669129366 - not sure if it's a case of flaky tests (yes, it looks like it the case - after restart of failed builds - they are all green)

### Questions:

### DIFF
--- a/cassandra/pom.xml
+++ b/cassandra/pom.xml
@@ -43,10 +43,6 @@
         <achilles.version>3.2.4-Zeppelin</achilles.version>
         <jna.version>4.2.0</jna.version>
 
-        <!-- plugin versions -->
-        <plugin.scala.version>2.15.2</plugin.scala.version>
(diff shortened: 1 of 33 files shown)


alexott:
@Leemoonsoo can you review it if you have time?

Leemoonsoo:
LGTM. Thank @alexott for the improvement!

PR #3710, opened 2020-04-01
[ZEPPELIN-4712]remove paragraph failed due to Interpreter connection refused

xiejiajun:
### What is this PR for?
- an interpreter process refused to connect due to it abnormal stop , which eventually caused the remove paragraph operation to fail to complete when i try remove a paragraph.

### What type of PR is it?
- Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4712

### Questions:
Does the licenses files need update? NO
Is there breaking changes for older versions? NO
Does this needs documentation? NO

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -713,7 +713,7 @@
 
 <property>
   <name>zeppelin.jobmanager.enable</name>
-  <value>true</value>
+  <value>false</value>
   <description>The Job tab in zeppelin page seems not so useful instead it cost lots of memory and affect the performance.
   Disable it can save lots of memory</description>
 </property>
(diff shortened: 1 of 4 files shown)


zjffdu:
I think we should add throw Exception for method `removeResourcesBelongsToParagraph` and let the caller decide how to handle the exception. What do you think ?

xiejiajun:
> I think we should add throw Exception for method `removeResourcesBelongsToParagraph` and let the caller decide how to handle the exception. What do you think ?

@zjffdu , i have considered this. If we do this, removeNote, removeFolder, and removeParagraph will need to handle this exception separately.

xiejiajun:
> I think we should add throw Exception for method `removeResourcesBelongsToParagraph` and let the caller decide how to handle the exception. What do you think ?

another purpose of  caught this exception in the removeResourcesBelongsToParagraph method is  in order to make  other running interpreters can  release resources that related to the paragraph that  needs to be removed. @zjffdu

zjffdu:
Sorry for late response, @xiejiajun  Your explanation make sense to me.  I think we need to improve the exception handling where we don't have very well convention. Currently your PR LGTM for fixing this issue

PR #3719, opened 2020-04-07
[ZEPPELIN-4725]. Merge process env into interpreter process env in InterpreterLauncher

zjffdu:
### What is this PR for?

This PR is to just merge process env into interpreter process env in InterpreterLauncher so that env defined in zeppelin-env.sh also apply to interpreters. e.g. `FLINK_HOME` specified in `zeppelin-env.sh` will apply on flink interpreter. But flink interpreter can override it.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4725

* CI pass

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
+++ b/zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.interpreter.launcher;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.interpreter.recovery.RecoveryStorage;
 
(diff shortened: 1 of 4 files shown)


zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
```diff
@@ -34,16 +35,16 @@ public FlinkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
...
     }
     File flinkHomeFile = new File(flinkHome);
     if (!flinkHomeFile.exists()) {
-      throw new IOException(String.format("FLINK_HOME %s doesn't exist", flinkHome));
+      throw new IOException(String.format("FLINK_HOME '%s' doesn't exist", flinkHome));
     }
     if (flinkHomeFile.isFile()) {
-      throw new IOException(String.format("FLINK_HOME %s is a file, but should be directory",
+      throw new IOException(String.format("FLINK_HOME '%s' is a file, but should be directory",
               flinkHome));
```
alexott, line 48: Instead, I would check for `! flinkHomeFile.isDirectory`, because on Linux/Mac, `isFile` will return `false` for devices, like, `/dev/ttywf`:

```
user=> (.isFile (file "/dev/ttywf"))
false
user=> (.isDirectory (file "/dev/ttywf"))
false
```

zjffdu: Thanks for the careful review, I have fixed it.

Reamer:
I think you have broke both tests in `StandardInterpreterLauncherTest` with `Map<String, String> env = EnvironmentUtils.getProcEnvironment();`
The environment variable are now much higher then [`1`](https://github.com/apache/zeppelin/blob/ec2f03f2acc37174de38a59b7da3520ebc344e33/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/launcher/StandardInterpreterLauncherTest.java#L82) or [`2`](https://github.com/apache/zeppelin/blob/ec2f03f2acc37174de38a59b7da3520ebc344e33/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/launcher/StandardInterpreterLauncherTest.java#L59) depending on your build environment.

Take a look into Travis.
https://travis-ci.org/github/apache/zeppelin/jobs/672840688

alexott:
interesting, I don't see that test failed in my setup...

zjffdu:
@Reamer @alexott I have fixed it via a hotfix PR https://github.com/apache/zeppelin/commit/98b26a12f4ebe73e9ab9aa011e9a8a56731addbe
But I still see another test failure, seems NotebookTest hangs there, it may take more time to investigate it. Sorry I should left comment earlier about the failed test.

alexott:
@zjffdu it was my mistake - I missed the failed test in the build :-(

zjffdu:
@Reamer @alexott It looks like CI pass now after I clean the travis cache, seems some environment issue.

PR #3721, opened 2020-04-07
[Zeppelin 4731] Ability to exclude fields from being saved in Note files

Muhammad-ms:
### What is this PR for?
Users should be able to specify which fields to exclude (if any) from being saved in note json files. This will be useful if, for example, paragraph outputs contain confidential information and therefore cannot be shared with others, or if users want to exclude fields which are not useful to their use case to avoid creating unnecessary diffs when running a note.
Fields to be excluded can be defined as a comma separated list in System Properties. for example:
export JAVA_OPTS="-Dzeppelin.note.file.exclude.fields=results,dateStarted,dateFinished,status"

### What type of PR is it?
[ Feature ]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4731


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -70,9 +71,13 @@
 
   // serialize Paragraph#runtimeInfos and Note#path to frontend but not to note file
   private static final ExclusionStrategy strategy = new ExclusionStrategy() {
+    List<String> fieldsExcluded = Arrays.asList(
+            System.getProperty("zeppelin.note.file.exclude.fields", "")
```
zjffdu, line 75: We'd better to put it into zeppelin-site.xml ?

Muhammad-ms: Added the propery to zeppelin-site.xml template

conf/zeppelin-site.xml.template
```diff
@@ -61,6 +61,12 @@
   <description>path or URI for notebook persist</description>
 </property>
 
+<property>
+  <name>zeppelin.note.file.exclude.fields</name>
+  <value>results,status</value>
```
zjffdu, line 66: The default value should be empty, because users would copy `zeppelin-site.xml.template` to `zeppelin-site.xml`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
zjffdu: Then we should `ZeppelinConfiguration.getX` of `System.getProperty`

conf/zeppelin-site.xml.template
Muhammad-ms: Comment addressed

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
Muhammad-ms: Comment addressed

Reamer:
Hi @Muhammad-ms,
I think you should create a function in `ZeppelinConfiguration.java` ([list example](https://github.com/apache/zeppelin/blob/3e614e0ca47a05338e642d5cdc9e77532fc96efd/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L654-L661)) to access your values.

NOTE: `Arrays.asList(new String[0]);` should replace to `Collections.emptyList();`

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -60,6 +60,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Arrays;
```
zjffdu, line 63: Not used ?

zjffdu:
ping @Muhammad-ms

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
Muhammad-ms: Comment Addressed

Muhammad-ms:
> Hi @Muhammad-ms,
> I think you should create a function in `ZeppelinConfiguration.java` ([list example](https://github.com/apache/zeppelin/blob/3e614e0ca47a05338e642d5cdc9e77532fc96efd/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java#L654-L661)) to access your values.
> 
> NOTE: `Arrays.asList(new String[0]);` should replace to `Collections.emptyList();`

Comment Addressed

zjffdu:
Sorry @Muhammad-ms I merge this PR with #3989

Reamer:
Closed because #3989 was merged.

PR #3726, opened 2020-04-08
[ZEPPELIN-4735]. Fail to getCompletion when interpreter doesn't implement completion

zjffdu:
### What is this PR for?

Trivial PR which only return empty list instead of null. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4735

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
@@ -32,8 +32,8 @@ import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Field;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/Interpreter.java
```diff
@@ -156,7 +157,7 @@ public abstract InterpreterResult interpret(String st,
   @ZeppelinApi
   public List<InterpreterCompletion> completion(String buf, int cursor,
       InterpreterContext interpreterContext) throws InterpreterException {
-    return null;
+    return new ArrayList<>();
```
alexott, line 160: do we need mutable List here? If not, can we use `Collections.emptyList()` ?

zjffdu: Good point, will fix it

PR #3735, opened 2020-04-15
[ZEPPELIN-4757] Fix startup script errors/warnings

alexott:
### What is this PR for?

Errors & warnings are produced from the checks that are assuming that Zeppelin is running
on Linux (`getent` is present) & in the container environment (`/etc/passwd` is
writable).

This patch fixes following:
 - added check for presence of the `getent`
 - added explicit check that script is running in the container by checking
   `/proc/self/cgroup`
 - also removed `MaxPermSize` JVM flag that isn't supported in Java 8

### What type of PR is it?

Bug Fix

### What is the Jira issue?

* ZEPPELIN-4757

* https://travis-ci.org/github/alexott/zeppelin/builds/675356068

### DIFF
--- a/bin/common.cmd
+++ b/bin/common.cmd
@@ -58,11 +58,11 @@ if not defined ZEPPELIN_ENCODING (
 )
 
 if not defined ZEPPELIN_MEM (
-    set ZEPPELIN_MEM=-Xms1024m -Xmx1024m -XX:MaxPermSize=512m
+    set ZEPPELIN_MEM=-Xms1024m -Xmx1024m
 )
 
 if not defined ZEPPELIN_INTP_MEM (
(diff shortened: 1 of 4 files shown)


Reamer:
Hi @alexott,
thanks for improving the start within a container. Can you please add [the same lines](https://github.com/apache/zeppelin/pull/3735/files#diff-1724182f3ebaf54f5c9e202dcdf82415R22-R43) into `interpreter.sh`.

alexott:
@Reamer done...

Leemoonsoo:
LGTM!

Reamer:
LGTM
We should also cleanup MvnOpts in a follow-up PR with [permsize](https://github.com/apache/zeppelin/search?q=permsize&unscoped_q=permsize).

alexott:
Yes, I created a ZEPPELIN-4758 for it. I did here only because they were in the same shell scripts

PR #3737, opened 2020-04-18
Zeppelin 4582

redsk:
### What is this PR for?
Add scala 2.12 support for interpreters

### What type of PR is it?
Improvement

### Todos
* [ ] - `scalate-maven-plugin` used in `cassandra` does not exist for scala 2.12 and has not been in development since 2015. However, there's a [version](https://github.com/stephennancekivell/scalate-maven-plugin) compiled for scala 2.12 [here](https://search.maven.org/artifact/com.stephenn/scalate-maven-plugin_2.12/0.0.2/maven-plugin). But no single version compiled for the three scala version (which would be needed for Zeppelin's pom.xml). I've hacked my way [like this](
https://github.com/redsk/zeppelin/commit/cb0fa2c5af547974dd995d0db93d90f54a9437f2) just to compile Zeppelin but we probably need a better solution.
* [] - Beam. `beam-runners-flink_2.12` does not exist. There seem to be a [workaround](https://issues.apache.org/jira/browse/BEAM-7544?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel) but I've not tried it. I've just disabled Beam from the modules for testing Zeppelin.

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4582

Not done yet.

### Questions:
* Is there breaking changes for older versions? Unsure
* Does this needs documentation? Probably

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,6 @@ tramp
 
 # Git properties
 **/git.properties
+
+# To keep track of the scala version used in the build
+/dev/current_scala_version
--- a/dev/change_scala_version.sh
+++ b/dev/change_scala_version.sh
@@ -19,7 +19,7 @@
 
 set -e
 
-VALID_VERSIONS=( 2.10 2.11 )
+VALID_VERSIONS=( 2.10 2.11 2.12 )
 
 usage() {
   echo "Usage: $(basename $0) [-h|--help] <version>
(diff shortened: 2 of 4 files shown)


alexott:
My general question is - do we really need to support multiple versions of Scala globally? Interpreters are better isolated right now from each other & core, so each interpreter can use its own version of Scala. And for some interpreters, like, `beam` we don't have specific Scala version (I think that I had this problem with Ignite as well).

Maybe it makes sense to start a discussion in dev list about need of the switch between different Scala versions?

PR #3739, opened 2020-04-20
[ZEPPELIN-4755] Fix the result graphs cannot be set to the correct size

hsuanxyz:
### What is this PR for?

Render the graphs at the next tick to get the correct size

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-4755

Repeatedly switch graph types and check their size

### Questions:


zjffdu:
Thanks @hsuanxyz I tried this PR, it works pretty well.

PR #3741, opened 2020-04-21
[ZEPPELIN-4759] Paragraph refreshing make the other paragraph shaking

zjffdu:
### What is this PR for?
This PR just resolve the paragraph shaking issue. Before this PR, I will refresh paragraph output by 2 steps: step 1. clear previous output, step 2. update paragraph with new output.

After this PR, I will just make it as just one step: update paragraph with new output. The main change is o method  InterpreterOutput#clear. I add one flag to indicate whether it should send the update to frontend to refresh the output.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4759

* CI pass

* Before
before

* After
after

### Questions:

### DIFF
--- a/flink/src/main/java/org/apache/zeppelin/flink/JobManager.java
+++ b/flink/src/main/java/org/apache/zeppelin/flink/JobManager.java
@@ -199,10 +199,10 @@ public class JobManager {
           if (isStreamingInsertInto) {
             StringBuilder builder = new StringBuilder("%html ");
             builder.append("<h1>Duration: " +
-                    Integer.parseInt(rootNode.getObject().getString("duration")) / 1000 +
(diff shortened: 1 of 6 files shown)


epugh:
This is a great fix, I find it very disjointing when I edit a notebook with someone else, and the repainting of the screen when they make a change on their end is very jarring!

zjffdu:
@epugh This is only to fix the issue happens in flink interpreter where I refresh paragraph when new data is coming from flink stream. I don't think it would fix the collaboration scenario you mentioned. Maybe you can create a ticket, we can fix it in other PR.

PR #3742, opened 2020-04-21
[ZEPPELIN-4764]. Support real time dashboard for jdbc interpreter

zjffdu:
### What is this PR for?
This PR is to support build real time dashboard via jdbc interpreter.  User just need to specify refreshInterval to enable it.  e.g.

```
%mysql(refreshInterval=2000)

select * from flink_cdn.cdn_access_statistic;
```

### What type of PR is it?
[Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4764

* Unit test is added, also manually tested.

jdbc_time

jdbc_refresh

### Questions:

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -33,6 +33,144 @@
   <version>0.9.0-SNAPSHOT</version>
   <name>Zeppelin: JDBC interpreter</name>
 
+  <properties>
+    <!--library versions-->
+    <interpreter.name>jdbc</interpreter.name>
+    <postgresql.version>9.4-1201-jdbc41</postgresql.version>
+    <hadoop.common.version>${hadoop2.7.version}</hadoop.common.version>
+    <h2.version>1.4.190</h2.version>
+    <commons.dbcp2.version>2.0.1</commons.dbcp2.version>
+
+    <!--test library versions-->
+    <mockrunner.jdbc.version>1.0.8</mockrunner.jdbc.version>
+  </properties>
+
+  <dependencies>
+
+	<dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>${postgresql.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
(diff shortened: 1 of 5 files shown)


jdbc/pom.xml
```diff
@@ -33,6 +33,144 @@
   <version>0.9.0-SNAPSHOT</version>
   <name>Zeppelin: JDBC interpreter</name>
 
+  <properties>
+    <!--library versions-->
+    <interpreter.name>jdbc</interpreter.name>
+    <postgresql.version>9.4-1201-jdbc41</postgresql.version>
+    <hadoop.common.version>2.7.2</hadoop.common.version>
+    <h2.version>1.4.190</h2.version>
+    <commons.dbcp2.version>2.0.1</commons.dbcp2.version>
```
alexott, line 42: why do we use version that is released in 2014th? Maybe it makes sense to upgrade to some later version?

jdbc/pom.xml
```diff
@@ -33,6 +33,144 @@
...
+    <!--test library versions-->
+    <mockrunner.jdbc.version>1.0.8</mockrunner.jdbc.version>
+  </properties>
+
+  <dependencies>
+
+	<dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>${postgresql.version}</version>
```
alexott, line 53: should it have `test` scope as well? Why we need specific JDBC artifact in Zeppelin itself?

jdbc/pom.xml
zjffdu: Yes, agree.

jdbc/pom.xml
zjffdu: I believe it is because we use postgresql by default in jdbc interpreter. This is by design I think.

zjffdu:
@alexott Thanks for your comment. The issues you mentioned make sense to me. But I think it is not related to this ticket. We could do it in another ticket. I do think we need to refactor and revisit the design of jdbc interpreter.

alexott:
@zjffdu I agree with that - that's why I commented only :-)
Otherwise, LGTM for me

PR #3745, opened 2020-04-22
[ZEPPELIN-4695] Bump dependencies to fix CVEs

alexott:
### What is this PR for?

This patch bumps versions of some components that were reported as having (potential?) vulnerabilities.  It also unifies usage of different hadoop versions, etc., moving version configuration to a top-level pom.xml, instead of specification of them inside individual poms. This for example, also fixes #3702. 

### What type of PR is it?
Refactoring

### TODOs

We'll need to followup on the other items, like:
* Check dependencies in the Docker image

### What is the Jira issue?
* ZEPPELIN-4695

* https://travis-ci.org/github/alexott/zeppelin/builds/678099649

### Questions:

### DIFF
--- a/beam/pom.xml
+++ b/beam/pom.xml
@@ -44,7 +44,6 @@
     <!-- library versions -->
     <netty.version>4.1.42.Final</netty.version>
     <servlet.api.version>3.1.0</servlet.api.version>
-    <commons.exec.version>1.3</commons.exec.version>
     <avro.version>1.8.1</avro.version> <!-- should match beam dependency -->
   </properties>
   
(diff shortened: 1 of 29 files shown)


Reamer:
great pull request!

Only one small question. Why setting a global JVM system property to disable the quartz update check?

I think the following code is better, because:
 1) The compiler checks, if the property is still available.
 1) You have no side-effects, which other threads/tasks
```
    Properties schedulerProperties = new Properties();
    schedulerProperties.setProperty(StdSchedulerFactory.PROP_SCHED_SKIP_UPDATE_CHECK, "true");
    return new StdSchedulerFactory(schedulerProperties).getScheduler();
```

alexott:
I didn't change code much - previous version did use system property. 
But I also tried your variant, and tests were failing for some reason - I think that it makes sense to investigate why this happens, but maybe in follow up JIRA

alexott:
I've created ZEPPELIN-4772 to investigate why configuration via properties doesn't work

Leemoonsoo:
LGTM

PR #3748, opened 2020-04-24
[ZEPPELIN-4775]. Sometimes sql text is not highlighted

zjffdu:
### What is this PR for?

This PR is to fix the highlight issue. The root cause is in the magic parsing logic in frontend. This PR just move the magic logic from frontend to backend. So that we use the consistent parsing logic. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4775

* CI pass

Before
image

After

image

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -885,7 +885,7 @@ public class NotebookService {
   }
 
   public void getEditorSetting(String noteId,
-                               String magic,
+                               String paragraphText,
                                ServiceContext context,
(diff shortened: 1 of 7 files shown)


zjffdu:
@alexott No test for this, it is pretty hard to test the highlight feature in frontend.

alexott:
I meant for parsing on the server side, not for highlighting...

zjffdu:
The server side test is already in. https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/ParagraphTextParserTest.java

alexott:
great! thank you - I don't have any more comments

PR #3752, opened 2020-04-28
[Zeppelin 4779] Allow Notebook directory to be nested within Git Repo

Muhammad-ms:
### What is this PR for?
This PR allows Git repo directory and notebook directory to be separated so users have the option of having notebook directory nested within git repo. This would allow users to drill down to a specific sub-folder of the git repo on GUI. For e.g. if the git repo consists of the following data structure:
-- Folder A
  -- Folder B
  -- Folder C
    -- Folder D 
And the user wants to only view Folder D on the gui, they can achieve this by defining git repo dir as <parent directory of Folder A> and notebook dir as <parent directory of Folder A>/Folder A/Folder C/Folder D

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4779


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/GitNotebookRepo.java
```diff
@@ -20,6 +20,7 @@
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
+import org.apache.commons.lang.StringUtils;
```
Reamer, line 23: Please use `org.apache.commons.lang3.StringUtils;`

Muhammad-ms: Switched to lang3

Reamer:
Why don't you work with a subpath? I think it should make things easier.

Muhammad-ms:
> Why don't you work with a subpath (concat a subpath to nodebookdir)? I think it should make things easier.

Using subpath changes git commands in a way which breaks them. For e.g. the root of git repo is `<parent of Folder A>`, but using subpath means that the relative note path for note A in Folder D becomes simply `note A` instead of `Folder A/Folder C/ Folder D/note A`

Reamer:
I think I understand your approach.
You should change your view. `ZEPPELIN_NOTEBOOK_DIR` should handle the Git Root-Directory you should introduce a new variable like `ZEPPELIN_NOTEBOOK_GIT_SUB_VIEW`. Use this variable (if set) to concat to the folder with is actually viewed in ZeppelinUI. In your example this variable should contain `Folder C/ Folder D`

Btw. I'm not the right person to review your change. Maybe @alexott can help with this.

Some thoughts.
I doesn't like the introduced [regular expression](https://github.com/apache/zeppelin/pull/3752/files#diff-3d1ddb2389a3d1baf113c60ab996f722R259-R271). `/` as a separator is only valid on Unix, what is with Windows? Therefore my approach with the sub-path.

jongyoul:
It looks quite a useful approach but I believe that you'd better make another notebook repo to keep current one and to adopt your own git repository. :-)

PR #3757, opened 2020-05-01
[dev fix] make fix_version unique

alexott:
### What is this PR for?

Right now, if you merge PR into 2 branches, like, `master`, and `branch-0.9`, script
generates suggestion for "Fix Version" as `0.9.0,0.9.0`, because both master & 0.9 branch
are using the same base version.  This small fix leaves only unique values for suggested
version(s).

### What type of PR is it?
Development Improvement


zjffdu:
LGTM

PR #3760, opened 2020-05-04
Minor refactoring

KirillTurchinskii:
### What is this PR for?
Position literals first in comparisons, if the second argument is null then NullPointerExceptions can be avoided, they will just return false.
Then I will move to annotations and other issues that my static analyzer shows
(My previous branch was removed because of the multiple problems, so I decided to start over)

### What type of PR is it?
Refactoring

### What is the Jira issue?
This task is not on jira, but I'll move to this
https://issues.apache.org/jira/browse/ZEPPELIN-3225?filter=12348928

https://travis-ci.org/github/KirillTurchinskii/zeppelin/builds/682716706

### Questions:
* What else besides annotations (that I will add) I can fix/refactor using static analysis?


Reamer:
LGTM

Leemoonsoo:
LGTM

PR #3763, opened 2020-05-06
[ZEPPELIN-4790]. Throw exception when using flink for scala 2.12

zjffdu:
### What is this PR for?

This PR will throw exception when using flink for scala 2.12 which is not supported now. See the screenshot below.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4790

* Manually tested.

image

### Questions:

### DIFF
--- a/flink/src/main/java/org/apache/zeppelin/flink/FlinkInterpreter.java
+++ b/flink/src/main/java/org/apache/zeppelin/flink/FlinkInterpreter.java
@@ -17,13 +17,10 @@
 
 package org.apache.zeppelin.flink;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apache.flink.api.scala.ExecutionEnvironment;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.runtime.jobgraph.SavepointConfigOptions;
 import org.apache.flink.streaming.api.scala.StreamExecutionEnvironment;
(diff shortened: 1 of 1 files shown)


flink/src/main/java/org/apache/zeppelin/flink/FlinkInterpreter.java
```diff
@@ -48,18 +48,39 @@
...
+    LOGGER.info("Using Scala: " + scalaVersionString);
+    if (scalaVersionString.contains("version 2.10")) {
+      return "2.10";
+    } else if (scalaVersionString.contains("version 2.11")) {
+      return "2.11";
+    } else if (scalaVersionString.contains("version 2.12")) {
+      return "2.12";
+    } else {
+      throw new InterpreterException("Unsupported scala version: " + scalaVersionString);
+    }
```
alexott, line 64: Do we plan to support multiple Scala versions in the future, and have some functionality depending on it?
Otherwise I would maybe just implement something like "isSupportedScalaVersion" with one check for Scala 2.11 as of now.

zjffdu: addressed

PR #3764, opened 2020-05-06
[ZEPPELIN-4797]. Enable multiple sql in jdbc interpreter by default

zjffdu:
### What is this PR for?

Before this PR, user have to set property `default.splitQueries` to enable multiple queries in one paragraph, this PR remove this property, this PR remove this property, and make multiple sql query in support by default.

### What type of PR is it?
[Feature | Documentation]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4797

* Unit test is updated

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -112,7 +112,6 @@ public class JDBCInterpreter extends KerberosInterpreter {
   static final String COMPLETER_SCHEMA_FILTERS_KEY = "completer.schemaFilters";
   static final String COMPLETER_TTL_KEY = "completer.ttlInSeconds";
   static final String DEFAULT_COMPLETER_TTL = "120";
-  static final String SPLIT_QURIES_KEY = "splitQueries";
   static final String JDBC_JCEKS_FILE = "jceks.file";
   static final String JDBC_JCEKS_CREDENTIAL_KEY = "jceks.credentialKey";
   static final String PRECODE_KEY_TEMPLATE = "%s.precode";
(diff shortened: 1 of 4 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -695,17 +688,7 @@ private InterpreterResult executeSql(String propertyKey, String sql,
...
-        sqlArray = sqlSplitter.splitSql(sql);
-      } else {
-        if (sql.endsWith(";")) {
-          sql = sql.substring(0, sql.length() - 1);
-        }
-        sqlArray = Arrays.asList(sql);
-      }
-
+      List<String> sqlArray = sqlSplitter.splitSql(sql);
       for (int i = 0; i < sqlArray.size(); i++) {
```
Reamer, line 692: a [For-Each Loop](https://docs.oracle.com/javase/8/docs/technotes/guides/language/foreach.html) would be nicer

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -695,17 +688,7 @@ private InterpreterResult executeSql(String propertyKey, String sql,
...
         String sqlToExecute = sqlArray.get(i);
```
alexott, line 693: for-each loop would look nicer :-)

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
zjffdu: addressed

PR #3765, opened 2020-05-07
[ZEPPELIN-4763]. Use isolated interpreter for cron note job

zjffdu:
### What is this PR for?

This PR is to use isolated interpreter for cron note job. Cron job is for production job, so we should use dedicated interpreter for it instead of using shared interpreter with other notes or users.  Currently interpreter binding mode is determined by note or users (scoped or isolated), This PR introduce `ExecutionContext` which include other factors (isCronMode) to determine how we create interpreters.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4763

* Unit test is updated, and manually tested

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/ExecutionContext.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/ExecutionContext.java
@@ -14,20 +14,43 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.zeppelin.interpreter;
 
-/**
- * InterpreterFactory Interface
- * Provides the interface to the ClusterManagerServer
- * through the user, nodeId, replName query interpreter
- * Since the InterpreterFactory is in the zeppelin-zengine module,
- * the ClusterManagerServer in the zeppelin-interpreter module
- * cannot access InterpreterFactory#getInterpreter(...),
- * So access through the interface.
- */
-public interface InterpreterFactoryInterface {
-  Interpreter getInterpreter(String replName,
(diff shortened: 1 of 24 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/ExecutionContext.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.interpreter;
+
+public class ExecutionContext {
+
+  private String user;
+  private String noteId;
+  private boolean inCronMode;
```
Reamer, line 24: You should add final, because you have no setter.

zjffdu: Good catch，will fix it.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -422,16 +424,23 @@ public String getGroup() {
     return group;
   }
 
-  private String getInterpreterGroupId(String user, String noteId) {
+  private static DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
```
Reamer, line 427: Should also be an candidate for the `final` keyword.

Reamer:
LGTM. Can you solve the merge conflict in `QuartzSchedulerService.java`.

PR #3767, opened 2020-05-09
[ZEPPELIN-4806]. Throw proper exception when hadoop jar is missing in yarn mode for flink interpreter

zjffdu:
### What is this PR for?
This PR actually solve 2 things in flink yarn mode.
1. You don't need flink-hadoop-shaded jar when flink is not in yarn mode (before this PR, it is still needed even when it is not in yarn mode)
2. Throw proper error message when hadoop jar is missing in yarn mode.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4806

* CI pass

image

### Questions:

### DIFF
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -236,6 +236,10 @@ elif [[ "${INTERPRETER_ID}" == "flink" ]]; then
 
   if [[ -n "${HADOOP_CONF_DIR}" ]] && [[ -d "${HADOOP_CONF_DIR}" ]]; then
     ZEPPELIN_INTP_CLASSPATH+=":${HADOOP_CONF_DIR}"
+    if ! [ -x "$(command -v hadoop)" ]; then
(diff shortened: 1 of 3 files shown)


bin/interpreter.sh
```diff
@@ -236,6 +236,10 @@ elif [[ "${INTERPRETER_ID}" == "flink" ]]; then
 
   if [[ -n "${HADOOP_CONF_DIR}" ]] && [[ -d "${HADOOP_CONF_DIR}" ]]; then
     ZEPPELIN_INTP_CLASSPATH+=":${HADOOP_CONF_DIR}"
+    if ! [ -x "$(command -v hadoop)" ]; then
```
Reamer, line 239: Is this exception in your screenshot even possible? It seems that the interpreter does not start if `hadoop' is not in the PATH?

PR #3771, opened 2020-05-15
[ZEPPELIN-4726] Use entry.user & entry.password instead of user.entry and password.entry for credential injection

zjffdu:
### What is this PR for?

This is a refactoring PR which just use entry.user & entry.password instead of user.entry and password.entry for credential injection .

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4726

* CI pass

### Questions:


zjffdu:
Document is updated

PR #3772, opened 2020-05-15
[ZEPPELIN-4805]. Error Stack Trace not received on rest API

zjffdu:
### What is this PR for?

 This PR is to fix the issue that no error message returned of running paragraph. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4805

* CI pass

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -279,6 +279,23 @@ public class NotebookService {
     }
   }
 
+  /**
+   * Executes given paragraph with passed paragraph info like noteId, paragraphId, title, text and etc.
+   *
+   * @param noteId
+   * @param paragraphId
+   * @param title
+   * @param text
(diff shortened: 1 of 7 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -279,6 +279,22 @@ public Note importNote(String notePath,
...
+   * @param paragraphId
+   * @param title
+   * @param text
+   * @param params
+   * @param config
+   * @param failIfDisabled
+   * @param blocking
+   * @param context
+   * @param callback
+   * @return
```
alexott, line 295: should we move "return true only when ..." here and at beginning something like "executes given paragraph"...?

zjffdu: javadoc is updated

PR #3773, opened 2020-05-17
[ZEPPELIN-4821] Spark interpreter uses incorrect property name

alexott:
### What is this PR for?

Spark interpreter did use incorrect property name for Spark Master - `master`, although a lot of code was dependent on `spark.master` - it's better to use only one name everywhere

### What type of PR is it?
Bug Fix 

### What is the Jira issue?
* ZEPPELIN-4821

* https://travis-ci.org/github/alexott/zeppelin/builds/699582429


alexott:
@zjffdu can you look when you have a time?

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -93,7 +93,8 @@ public void open() throws InterpreterException {
         }
       }
       // use local mode for embedded spark mode when spark.master is not found
-      conf.setIfMissing("spark.master", "local");
```
zjffdu, line 96: I think this is the line of code that cause the issue in the mail-list

zjffdu:
@alexott Overall I agree with using `spark.master` instead of `master`. But I think we should keep compatibility because lots of users are still using `master` today. 
And we should update this function to take `spark.master` precedence over `master` (https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L326) What do you think ?

alexott:
I agree with that Jeff - I was just thinking about migration path, etc. Give me a couple of days to fix that...

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
alexott: yep, looks like...

zjffdu:
@alexott is this ready to be reviewed again ?

alexott:
@zjffdu not yet - I need to rebase it on top of the Travis-CI cleanup effort, and re-check again...

alexott:
@zjffdu now it's ready for review as well... tests are finally passing.

I modified Spark launcher to use `--conf spark.master` instead of `--master`...

conf/zeppelin-env.cmd.template
```diff
@@ -17,7 +17,7 @@ REM limitations under the License.
 REM
 
 REM set JAVA_HOME=
-REM set MASTER=                 		REM Spark master url. eg. spark://master_addr:7077. Leave empty if you want to use local mode.
+REM set SPARK_MASTER=                 		REM Spark master url. eg. spark://master_addr:7077. Leave empty if you want to use local mode.
```
zjffdu, line 20: Not aligned with other rows

conf/zeppelin-env.sh.template
```diff
@@ -17,7 +17,7 @@
 #
 
 # export JAVA_HOME=
-# export MASTER=                 		# Spark master url. eg. spark://master_addr:7077. Leave empty if you want to use local mode.
+# export SPARK_MASTER=           	# Spark master url. eg. spark://master_addr:7077. Leave empty if you want to use local mode.
```
zjffdu, line 20: Not aligned with other rows

zjffdu:
LGTM

conf/zeppelin-env.cmd.template
alexott: yeah, the both files contained a mix of tabs & spaces, why it wasn't catch up...
Fixed in both

PR #3775, opened 2020-05-20
[ZEPPELIN-4832]. run note in non-blocking way could not guarantee the paragraph execution order

zjffdu:
### What is this PR for?

Before this PR, if user run the note in non-blocking way via rest api, the execution order of paragraphs is not determined, paragraph 2 will start to run after paragraph 1 enter running state. But we should only start paragraph 2 when paragraph 1 is finished. 

This PR fix this issue by minor change the `RemoteScheduler` and introduce `.execution.mode` to tell `RemoteScheduler` that whether the job is running as individual paragraph or as part of note execution. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4832

* Unit test is added

### Questions:


Leemoonsoo:
LGTM

PR #3778, opened 2020-05-27
[ZEPPELIN-4802]. pyspark warnings with spark-3.0.0-preview - Sync of properties between JVM and PVM

zjffdu:
### What is this PR for?

This PR will first set env `PYSPARK_PIN_THREAD` to be `true` and then suppress the warning message in `pyspark.util`.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4802

* CI pass
* Manually tested

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/IPySparkInterpreter.java
@@ -148,6 +148,10 @@ public class IPySparkInterpreter extends IPythonInterpreter {
     return sparkInterpreter.getSparkVersion().getMajorVersion() == 1;
   }
 
+  public boolean isSpark3() {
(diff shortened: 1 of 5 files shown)


zjffdu:
@Reamer Could you help review it ?

Reamer:
> @Reamer Could you help review it ?

I can confirm that the warnings are gone, but it is a good code style to ignore all warnings from `pyspark.util'.
```
if intp.isSpark3():
    warnings.filterwarnings(action='ignore', module='pyspark.util')
```

zjffdu:
@Reamer  I looked at the source code of spark, no matter what user do, it would always print warning message, so I think suppressing the warning is the only way to do that.

https://github.com/apache/spark/blob/master/python/pyspark/util.py#L117

Reamer:
Thank you very much for your explanation. Always printing out a warning message is not a good solution. Therefore I also think that suppressing the warning is a good way.

PR #3781, opened 2020-06-01
[ZEPPELIN-4845]. Recover running paragraph when recovery is enabled

zjffdu:
### What is this PR for?

This PR is to recover the running paragraph when recovery is enabled. 
1. Get all the running Interpreter processes from RecoveryStorage.
2. Reconnect to the running interpreter process.
3. Resubmit the paragraph to interpreter process.

### What type of PR is it?
[ Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4845

* CI pass and manully tested the recovery scenairo.

ezgif com-video-to-gif (1)

### Questions:


zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sStandardInterpreterLauncher.java
```diff
@@ -155,7 +155,7 @@ public InterpreterClient launch(InterpreterLaunchContext context) throws IOExcep
             properties,
             buildEnvFromProperties(context),
             getZeppelinService(),
-            getZeppelinServiceRpcPort(),
+            Integer.parseInt(getZeppelinServiceRpcPort()),
```
Reamer, line 158: Maybe we should change the return value of [getZeppelinServiceRpcPort()](https://github.com/apache/zeppelin/blob/261eb395cdb3e94db5fc530ee08a89f1f37b51bf/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sStandardInterpreterLauncher.java#L119) to int.

zjffdu: Make sense, let me update it.

PR #3785, opened 2020-06-04
[ZEPPELIN-4852]. Add name to RemoteInterpreterProcess

zjffdu:
### What is this PR for?

This is a trivial PR which add `getInterpreterGroupId` to `InterpreterClient`, and use `interpreterGroupId` as the identifier of RemoteInterpreterProcess because `interpreterGroupId` is also unique identifier of InterpreterGroup   

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4852

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
@@ -141,7 +141,7 @@ public class RemoteInterpreterManagedProcess extends RemoteInterpreterProcess {
 
   public void stop() {
     if (isRunning()) {
-      LOGGER.info("Kill interpreter process for interpreter group: " + getInterpreterGroupId());
+      LOGGER.info("Kill interpreter process for interpreter group: {}", getInterpreterGroupId());
       try {
         callRemoteFunction(new RemoteFunction<Void>() {
           @Override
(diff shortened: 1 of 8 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterRunningProcess.java
```diff
@@ -80,6 +88,10 @@ public Void call(RemoteInterpreterService.Client client) throws Exception {
         } catch (Exception e) {
           logger.warn("ignore the exception when shutting down interpreter process.", e);
         }
+
+        // Shutdown connection
+        shutdown();
+        logger.info("Remote process of interpreter group: " + getInterpreterGroupId() + " is terminated");
```
Reamer, line 94: Could you please make the logger 'public static final' and avoid the creation of objects if the logger is disabled for the INFO level ( `LOGGER.info(String format, Object... arguments);`).

zjffdu: fixed

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterManagedProcess.java
```diff
@@ -141,7 +141,7 @@ public void start(String userName) throws IOException {
 
   public void stop() {
     if (isRunning()) {
-      LOGGER.info("Kill interpreter process");
+      LOGGER.info("Kill interpreter process for interpreter group: " + getInterpreterGroupId());
```
Reamer, line 144: We should avoid the creation of objects if a logging level is disabled 
This particular line should change to
```
LOGGER.info("Kill interpreter process for interpreter group: {}", getInterpreterGroupId());
```

PR #3786, opened 2020-06-04
[ZEPPELIN-4844]. Support yarn interpreter launch mode

zjffdu:
### What is this PR for?

This PR is to support yarn interpreter launch mode, that means to launch the interpreter process in yarn container. So that we can mitigate the memory pressure of zeppelin server machine. 
This PR add new launcher module `yarn` for that. I manually tested shell, python, jdbc and flink interpreted.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4844

* CI pass and manually tested

### Questions:


Reamer:
Hi @zjffdu 
I noticed missing hadoop jars in my zeppelin-server CLASSPATH after this commit.
After adding `-Pincludehadoop` to my build options, my Zeppelin server image size increased from 410.95MB to 2.13GB. I checked the filesystem and every Zeppelin plugin dir (e.g. `plugins/Launcher/K8sStandardInterpreterLauncher/`) includes now several other jars.

I think we should avoid this and the plugins should share the same dependencies, right?

Reamer:
> Hi @zjffdu I noticed missing hadoop jars in my zeppelin-server CLASSPATH after this commit. After adding `-Pincludehadoop` to my build options, my Zeppelin server image size increased from 410.95MB to 2.13GB. I checked the filesystem and every Zeppelin plugin dir (e.g. `plugins/Launcher/K8sStandardInterpreterLauncher/`) includes now several other jars.
> 
> I think we should avoid this and the plugins should share the same dependencies, right?

Followup: https://github.com/apache/zeppelin/pull/3817

PR #3790, opened 2020-06-07
[ZEPPELIN-4796] Better representation of string-like data in Cassandra interpreter

alexott:
### What is this PR for?

Refactoring of the Cassandra interpreter to a new driver changes the results representation to be strict CQL format, that is not always desirable.

This PR improves the situation by allow to control the output format (CQL vs human-readable), and also control such things as locale, timezone, format of timestamp/date/time columns, and precision for float & double types.

All settings could be set either on interpreter, or cell level.

### What type of PR is it?

Improvement

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-4796

* https://travis-ci.org/github/alexott/zeppelin/builds/698160753
* Tested manually (see screenshot)

Screenshot from 2020-06-14 11-36-09

### DIFF
--- a/cassandra/src/main/java/org/apache/zeppelin/cassandra/CassandraInterpreter.java
+++ b/cassandra/src/main/java/org/apache/zeppelin/cassandra/CassandraInterpreter.java
@@ -112,6 +112,22 @@ public class CassandraInterpreter extends Interpreter {
   public static final String CASSANDRA_TRUSTSTORE_PASSWORD =
           "cassandra.ssl.truststore.password";
 
+  public static final String CASSANDRA_FORMAT_FLOAT_PRECISION =
+          "cassandra.format.float_precision";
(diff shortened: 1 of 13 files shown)


alexott:
Good question is what better - should we use local properties of interpreter instead of the properties specified on the cell level & handled by parser? like,

```
%cassandra(locale=...., timezone=...)
```

zjffdu:
I think local properties is better, so that we can do it in zeppelin-server side and make this feature available to all the interpreters, otherwise each interpreter need to implement this feature separately.

alexott:
@zjffdu it's ready for review... It still needs to implement code interpreter-side, as we need to "know" data type information, but at least we can have some common properties, etc. I'm thinking about creating followup ticket for refactoring it for supporting most interpreters.

PR #3794, opened 2020-06-10
[ZEPPELIN-4873]. Display rich duration info for insert into flink job

zjffdu:
### What is this PR for?

Trivial PR which display rich duration info instead of just x seconds. See screenshot below.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4873

* CI pass

image

### Questions:

### DIFF
--- a/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
+++ b/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
@@ -111,7 +111,7 @@ public class JobManager {
   }
 
   public void cancelJob(InterpreterContext context) throws InterpreterException {
-    LOGGER.info("Canceling job associated of paragraph: " + context.getParagraphId());
+    LOGGER.info("Canceling job associated of paragraph: {}", context.getParagraphId());
     JobClient jobClient = this.jobs.get(context.getParagraphId());
     if (jobClient == null) {
       LOGGER.warn("Unable to remove Job from paragraph {} as no job associated to this paragraph",
--- a/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
+++ b/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
@@ -29,6 +29,7 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
```diff
@@ -110,7 +111,7 @@ public int getJobProgress(String paragraphId) {
   }
 
   public void cancelJob(InterpreterContext context) throws InterpreterException {
-    LOGGER.info("Canceling job associated of paragraph: "+ context.getParagraphId());
+    LOGGER.info("Canceling job associated of paragraph: " + context.getParagraphId());
```
alexott, line 114: It makes sense to use `{}` placeholder instead of concatenation - it won't lead to the call of the `context.getParagraphId()` if logger is not configured for INFO

zjffdu: fixed

flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
```diff
@@ -218,15 +222,39 @@ public void run() {
...
-      public int getProgress () {
-        return progress;
-      }
+    public int getProgress() {
+      return progress;
     }
   }
+
+  static String toRichTimeDuration(long duration) {
+    long days = TimeUnit.SECONDS.toDays(duration);
```
alexott, line 238: Maybe just use [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) data type, and then use corresponding functions, like, `toDays`, `toHours`, ...?

flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
```diff
@@ -218,15 +222,39 @@ public void run() {
...
 
-      public int getProgress () {
-        return progress;
-      }
+    public int getProgress() {
+      return progress;
     }
   }
+
+  static String toRichTimeDuration(long duration) {
```
alexott, line 237: would be useful to add a javadoc for it, saying what is parameter is - seconds, milliseconds, etc.?

flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
zjffdu: It seems using Duration doesn't' save work compared current approach, I still have to get days and minus it from duration, and the same for hours, minutes, seconds. So I didn't change it.

flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
zjffdu: javadoc is added

flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
Reamer: Is `String.format("%d:%02d:%02d", s / 3600, (s % 3600) / 60, (s % 60));` to simple?

PR #3795, opened 2020-06-10
[ZEPPELIN-4878]. Unable to run flink 1.10.1 in yarn mode due to FLINK-17788

zjffdu:
### What is this PR for?

Flink interpreter doesn't work with flink 1.10.1 in yarn mode due to FLINK-17788, this PR fix it in zeppelin side. Because in scala shell, flink only support yarn session mode, so when it is yarn mode, we can set deployment target to be yarn-session directly. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4878

* CI pass and manually tested 

### Questions:

### DIFF
--- a/flink/flink1.10-shims/pom.xml
+++ b/flink/flink1.10-shims/pom.xml
@@ -33,7 +33,7 @@
     <name>Zeppelin: Flink1.10 Shims</name>
 
     <properties>
-        <flink.version>1.10.1</flink.version>
+        <flink.version>${flink1.10.version}</flink.version>
         <scala.binary.version>2.11</scala.binary.version>
         <scala.version>2.11.12</scala.version>
     </properties>
--- a/flink/flink1.10-shims/pom.xml
+++ b/flink/flink1.10-shims/pom.xml
@@ -33,7 +33,7 @@
     <name>Zeppelin: Flink1.10 Shims</name>
 
     <properties>
-        <flink.version>1.10.0</flink.version>
+        <flink.version>1.10.1</flink.version>
         <scala.binary.version>2.11</scala.binary.version>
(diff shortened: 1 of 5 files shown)


flink/flink1.10-shims/pom.xml
```diff
@@ -33,7 +33,7 @@
     <name>Zeppelin: Flink1.10 Shims</name>
 
     <properties>
-        <flink.version>1.10.0</flink.version>
```
alexott, line 36: does it make sense to define a constant in the `flink-parent` so it will be shared?

zjffdu: I am afraid not. I plan to add profile flink-1.10 and flink-1.11 for flink interpreter after flink 1.11 is released. But `flink.version` in `flink1.10-shims` will not be affected by profile, it should always be flink 1.10, `flink.version` in `flink/interpreter` module will be affected by this profile, and will run test under `flink/interpreter` for different flink version.

alexott: I meant, that we can define property, like, `<flink.1.10.version>` and refer to it... I did this for Hadoop:

```
    <hadoop2.7.version>2.7.7</hadoop2.7.version>
    <hadoop2.6.version>2.6.5</hadoop2.6.version>
    <hadoop3.0.version>3.0.3</hadoop3.0.version>
    <hadoop3.1.version>3.1.3</hadoop3.1.version>
```

and then use it as `${hadoop3.1.version}`, etc.

zjffdu: Good idea, I have fixed it

PR #3796, opened 2020-06-11
[ZEPPELIN-4877]. text between 2 sql comment is mistaken as comment

zjffdu:
### What is this PR for?

This PR fix the bug in Splitter that text between 2 sql comment is mistaken as comment. Unit test is added in this PR as well. Besides that, this PR also remove the comment in the middle of sql. e.g. 
```
select a -- comment 1
from table_1
```  
will be converted to 
```
select a 
from table_1
```

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4877

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
@@ -106,8 +106,8 @@ public class SqlSplitter {
               && text.length() > (index + 1)) {
         if (isSingleLineComment(text.charAt(index), text.charAt(index + 1))) {
           singleLineComment = true;
-        } else if (text.length() > (index + 2) && text.charAt(index + 2) != '+' &&
-                text.charAt(index) == '/' && text.charAt(index + 1) == '*') {
+        } else if (text.charAt(index) == '/' && text.length() > (index + 2)
+                && text.charAt(index + 1) == '*' && text.charAt(index + 2) != '+') {
           multiLineComment = true;
         }
       }
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
@@ -106,8 +106,8 @@ public class SqlSplitter {
               && text.length() > (index + 1)) {
         if (isSingleLineComment(text.charAt(index), text.charAt(index + 1))) {
           singleLineComment = true;
-        } else if (text.charAt(index) == '/' && text.charAt(index + 1) == '*'
-                && text.length() > (index + 2) && text.charAt(index + 2) != '+') {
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
```diff
@@ -79,11 +81,9 @@ public SqlSplitter(String... additionalSingleCommentPrefixList) {
       }
 
       // end of multiple line comment
-      if (multiLineComment && character == '/' && text.charAt(index - 1) == '*') {
+      if (multiLineComment && (index - 1) >= 0 && text.charAt(index - 1) == '/'
```
alexott, line 84: Can we replace ` (index - 1) >= 0` with ` (index - 2) >= 0` and just skip it later?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
```diff
@@ -106,39 +106,30 @@ public SqlSplitter(String... additionalSingleCommentPrefixList) {
               && text.length() > (index + 1)) {
         if (isSingleLineComment(text.charAt(index), text.charAt(index + 1))) {
           singleLineComment = true;
-        } else if (text.charAt(index) == '/' && text.charAt(index + 1) == '*') {
+        } else if (text.charAt(index) == '/' && text.charAt(index + 1) == '*'
+                && text.length() > (index + 2) && text.charAt(index + 2) != '+') {
```
alexott, line 110: I would move `text.length() > (index + 2)` before `text.charAt(index + 1) == '*'`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
zjffdu: Sorry, I don't get it. Here I just check whether it is the end of multiple comment via checking whether (index-1) is `/` and (index-2) is `*`

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
zjffdu: fixed

alexott: sorry, I meant that we can write it as:

```
text.charAt(index) == '/' && text.length() > (index + 2) 
  && text.charAt(index + 1) == '*' && text.charAt(index + 2) != '+'
```

so we have range check for `index+1` covered, just in case if `index == lenght-1`...

zjffdu: I see, now it is fixed

PR #3797, opened 2020-06-11
[ZEPPELIN-4851]. Add property to flink interpreter to allow replace the yarn address

zjffdu:
### What is this PR for?

Trivial PR which add property `flink.webui.yarn.yarnAddress` that allow user to replace the yarn address. This is for some cases that cloud vender would map the internal resource manager address to other url. e.g. some cloud vender will map `http://resource-manager:8088` to `https://xxx-yarn.yy.cn/gateway/kkk/yarn`

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4851

* CI pass

### Questions:

### DIFF
--- a/flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
+++ b/flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
@@ -24,6 +24,7 @@ import java.nio.file.Files
 import java.util.Properties
 import java.util.concurrent.TimeUnit
 import java.util.jar.JarFile
+import java.util.regex.Pattern
 
 import org.apache.commons.lang3.StringUtils
 import org.apache.commons.lang3.exception.ExceptionUtils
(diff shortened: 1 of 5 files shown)


flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
```diff
@@ -829,6 +835,12 @@ class FlinkScalaInterpreter(val properties: Properties) {
       }
     }
   }
+
+  private def replaceYarnAddress(webURL: String, yarnAddress: String): String = {
+    val pattern = "(http://.*:\\d+)(/.*)".r
```
alexott, line 840: I maybe would change it to the to parsing of `webURL` into URL or URI, and construction of the new object, but with changed components. 
Also it won't handle `https` schema, etc.

zjffdu: @alexott  `URL` doesn't expose set method. So it is not easy to construct URL with components, however I fixed the https issue and also add unit tests

PR #3799, opened 2020-06-11
[ZEPPELIN-4870] Improve parsing of the paragraph properties

alexott:
### What is this PR for?

We can provide properties that are local to the paragraph, that could be used to pass an additional information for interpreter that could affect its behavior.  Unfortunately existing parsing functionality relies on the fact that key/value pairs need to be separated by `,` character, and doesn't handle values with special characters (`,`, `=`, ...) inside, like this:

```
%cassandra(locale=ruRU, timeFormat="E, d MMM yy", floatPrecision = 5, outputFormat=cql)
```

This PR changes the parsing logic to perform character-by-character parsing, and handling of the quoted keys & values, escaping of the special characters, etc.

### What type of PR is it?

Improvement 

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4870

* https://travis-ci.org/github/alexott/zeppelin/builds/697522260
* additional unit tests were added

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphTextParser.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphTextParser.java
@@ -17,13 +17,12 @@
 
 package org.apache.zeppelin.notebook;
 
-import org.apache.commons.lang3.StringUtils;
-
 import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Parser which is used for parsing the paragraph text.
@@ -42,6 +41,7 @@ import java.util.regex.Pattern;
  * localProperties: Map(pool->pool_1)
  */
 public class ParagraphTextParser {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ParagraphTextParser.class);
 
   public static class ParseResult {
     private String intpText;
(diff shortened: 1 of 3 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphTextParser.java
```diff
@@ -67,41 +67,105 @@ public String getScriptText() {
...
-            continue;
-          }
-          if (kv.length > 2) {
-            throw new RuntimeException("Invalid paragraph properties format: " + split);
-          }
-          if (kv.length == 1) {
-            localProperties.put(kv[0].trim(), kv[0].trim());
-          } else {
-            localProperties.put(kv[0].trim(), kv[1].trim());
+      int startPos = headingSpace.length() + intpText.length() + 1;
```
zjffdu, line 81: Put the local property extraction logics into a separate function for more readability.

alexott: @zjffdu done.  If you ok with overall change, I can push it to master, and rework my changes regarding formatting in Cassandra...

zjffdu:
Over LGTM, one last question about this PR, what happens in frontend if user specify invalid properties ?

alexott:
> Overall LGTM, one last question about this PR, what happens in frontend if user specify invalid properties ?

Interpreter should throw an exception... The parsing itself just do the general checks that syntax is correct etc., the rest is responsibility of the interpreter - we'll generalize common properties, like, time format, etc. over the time - I'll file a JIRA for it.

zjffdu:
Would the frontend display the proper error message ?

alexott:
Should be, at least I saw them in current implementation of Cassandra interpreter, when I specified incorrect time format, etc.

zjffdu:
Thanks for the confirmation @alexott , +1 from me

PR #3800, opened 2020-06-12
[ZEPPELIN-4870][hotfix] fix edge case, and remove duplicate parser

alexott:
### What is this PR for?

Tests didn't cover all of the edge cases, that lead to error when parsing specific message.

Also, was found that parsing is also duplicated in the `Paragraph` class - it was replaced with call to `ParagraphTextParser`.

The most visible change is that `%interpreter(some text` will know fail as there is no matching `)` character. But it's works the same if we change it to `%interpreter (some text`...

### What type of PR is it?
Hot Fix

### What is the Jira issue?
* ZEPPELIN-4870

* Added one more test
* https://travis-ci.org/github/alexott/zeppelin/builds/697944179


alexott:
@zjffdu can you look when you have a time...

zjffdu:
LGTM

PR #3806, opened 2020-06-18
[ZEPPELIN-4863]. User credential usage in jdbc interpreter doesn't work when the interpreter name is not jdbc

zjffdu:
### What is this PR for?
User credential only works in the default jdbc interpreter. If user create a new jdbc interpreter, e.g. hive, then credential won't work. This PR fix this and also do some code refactoring to make the jdbc interpreter module more readable. 

### What type of PR is it?
[Bug Fix  | Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4863

* CI pass

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -344,12 +344,12 @@ public class JDBCInterpreter extends KerberosInterpreter {
     }
   }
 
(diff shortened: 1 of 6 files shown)


zjffdu:
will merge if no more comment

Reamer:
Please avoid string concatenation when working with `LOGGER`.

zjffdu:
Thanks for the review @Reamer  I have fixed them

PR #3815, opened 2020-06-21
[ZEPPELIN-4818] Additional refactoring of Cassandra interpreter for new driver versions

alexott:
### What is this PR for?

New DataStax Java Driver allows to override configuration options by different ways - via config file, system properties, etc.  

This PR implements following:

* Allow to override any option of the Java driver by putting additional options into interpreter's configuration.  For example, this allows to use DataStax Astra (Cassandra as a Service) without need to make any changes in the code of the interpreter.  It also allows to configure any additional parameter not exposed directly by interpreter
* Upgrades driver to latest released version
* Configures embedded Cassandra for faster startup when executing unit tests

### What type of PR is it?

Improvement

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-4818

* https://travis-ci.org/github/alexott/zeppelin/builds/700969578
* tested manually


zjffdu:
LGTM

PR #3822, opened 2020-06-24
[ZEPPELIN-4907]. Add LocalRecoveryStorage

zjffdu:
### What is this PR for?

This PR add LocalRecoveryStorage based on java native local file system. Without this PR, user can only enable recovery by using FileSystemRecoveryStorage which depends on hadoop. This implementation doesn't depends on hadoop, user can enable recovery just by using this RecoveryStorage.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4907

* CI pass

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -648,8 +648,14 @@
 <!--
 <property>
   <name>zeppelin.recovery.storage.class</name>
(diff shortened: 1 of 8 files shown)


zjffdu:
Will merge if no more comment

PR #3833, opened 2020-06-29
[ZEPPELIN-4910]. Don‘t convert relative path to local absolute path when using hadoop filesystem

zjffdu:
### What is this PR for?

`ZeppelinConfiguration#getRelativeDir` actually get the absolute path intead of relative path, this PR just rename this method. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4910

* CI pass

### Questions:

### DIFF
--- a/zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
+++ b/zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
@@ -35,7 +35,8 @@ public class OldFileSystemNotebookRepo implements OldNotebookRepo {
   }
 
   public void init(ZeppelinConfiguration zConf) throws IOException {
-    this.fs = new FileSystemStorage(zConf, zConf.getNotebookDir());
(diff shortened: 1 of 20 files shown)


zeppelin-plugins/notebookrepo/filesystem/src/main/java/org/apache/zeppelin/notebook/repo/OldFileSystemNotebookRepo.java
```diff
@@ -35,7 +35,8 @@ public OldFileSystemNotebookRepo() {
   }
 
   public void init(ZeppelinConfiguration zConf) throws IOException {
-    this.fs = new FileSystemStorage(zConf, zConf.getNotebookDir());
+    this.fs = new FileSystemStorage(zConf,
```
Reamer, line 38: Why you change the path from `notebook` to `recovery`?

Reamer: Thanks for change

Reamer:
LGTM

PR #3847, opened 2020-07-10
[ZEPPELIN-4935]. IRInterpreter is broken due to missing of hadoop jars

zjffdu:
### What is this PR for?

Trivial PR to add hadoop jars to rlang module, otherwise IRInterpreter won't work because IRInterpreter depends on SparkBackend to communicate between R process and Java process. and SparkBackend depends on hadoop jar.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4935

* CI pass

### Questions:

### DIFF
--- a/rlang/pom.xml
+++ b/rlang/pom.xml
@@ -110,6 +110,13 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>2.7.7</version>
+            <scope>compile</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.jsoup</groupId>
             <artifactId>jsoup</artifactId>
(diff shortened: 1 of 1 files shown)


rlang/pom.xml
```diff
@@ -109,6 +109,13 @@
             <version>${spark.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>2.7.7</version>
```
Reamer, line 115: @zjffdu Why you are using a hard coded version? Maybe it's better use `hadoop.version`? (https://github.com/apache/zeppelin/blob/master/pom.xml#L146)

PR #3852, opened 2020-07-13
[ZEPPELIN-4950].  Support for manually specifying the Java version of Spark Interpreter Scala REPL

xiejiajun:
### What is this PR for?
- Support for manually specifying the Java version of Spark Interpreter Scala REPL.
- This feature can be used to resolve runtime errors caused by reference to third-party libraries that use high-version Java in Spark Interpreter.
 
### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
*  https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-4950

* manually tested
    *  Specify spark.repl.target through %spark.conf or when configuring Spark Interpreter
    *  Reference and call a Java method in a third-party library developed using Java 8 and using Java 8 features such as interface static methods through the Spark interpreter. If the call is successful, it means the test passed

- A runtime error will be triggered when  we reference a third-party library that uses the new features of Java8  if we using the default value `jvm-1.6`  of `settings.target.value` :
image

### Questions:


zjffdu:
@xiejiajun What is the default value of `settings.target.value` if we don't specify it ?
What does your screenshot mean ? It looks like it could not use library that use jdk 8

xiejiajun:
> @xiejiajun What is the default value of `settings.target.value` if we don't specify it ?
> What does your screenshot mean ? It looks like it could not use library that use jdk 8

@zjffdu 1.  The default value of `settings.target.value` is `jvm-1.6` when we don't specify it.
2.  This screenshot is  to explain  that  runtime error will be triggered when  we reference a third-party library that uses the new features of Java8  if we using the default value `settings.target.value` (jvm-1.6) .

zjffdu:
@xiejiajun I don't see spark scala shell specifying `settings.target.value`, does that mean spark scala shell unable to use any libraries depends on jdk 8 as well ?

xiejiajun:
> @xiejiajun I don't see spark scala shell specifying `settings.target.value`, does that mean spark scala shell unable to use any libraries depends on jdk 8 as well ?
@zjffdu Yes, I just tested, spark-shell does have this problem
image

zjffdu:
Thanks @xiejiajun, LGTM

PR #213, opened 2015-08-14
ZEPPELIN-223 fix: reconection in case of disconnection.

prabhjyotsingh:

PR #3864, opened 2020-07-22
avoiding retrieving and redrawing the note on websocket reconnect

jason-ogaard:
### What is this PR for?
Currently, when the websocket reconnects GET_NOTE is called. The purpose of this call on reconnect is to add the note back to the NoteSocketMap (see discussion here: https://github.com/apache/zeppelin/pull/213).

However, GET_NOTE also does two other things:
1. It retrieves the note from storage and
2. Sends that note back to the browser where it is redrawn.

Doing this causes issues for our users:
1. The retrieval of the note means that any work done since the note was last saved will be lost. Notes save often, but we've had users complain about this.
2. The redraw causes the cursor to jump back to the beginning of the line.
3. In our environment, the redraw causes the notebook to scroll down.

This PR creates a new request that will be called instead of GET_NOTE. The purpose of this new request is to simply put the note back into the NoteSocketMap so that we can continue working.

### What type of PR is it?
Improvement

### Todos
None

### What is the Jira issue?
None

* Open a notebook, wait for the websocket to reconnect, ensure that you can still do everything you should be able to do.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -308,6 +308,9 @@ public class NotebookServer extends WebSocketServlet
         case GET_NOTE:
           getNote(conn, messagereceived);
           break;
+        case REFRESH_NOTE_CONNECTION:
+          refreshNoteConnection(conn, messagereceived);
+          break;
         case NEW_NOTE:
           createNote(conn, messagereceived);
           break;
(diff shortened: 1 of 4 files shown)


alexott:
How it will be handled if note was modified on server side as well?

jason-ogaard:
> How it will be handled if note was modified on server side as well? Like, when several people are working on the same note

Thanks for the comment Alex. I'm sure there are scenarios that I've failed to consider and this is one of them. We've prevented users from modifying the same note at the same time for so long that I forgot you could do that. Perhaps I should withdraw this PR as that scenario wouldn't be handled at all.

alexott:
I think that it makes sense to start discussion on the development mailing list about best approach to this. I agree with your points about discarding changes, etc.

PR #3865, opened 2020-07-24
[ZEPPELIN-4968] Provide commented out config for helium registry

alexott:
### What is this PR for?

It adds commented-out configuration parameter for Helium registry, so people can simply uncomment it, instead of going to docs, and copy/paste from it.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4968

* tested manually

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -401,6 +401,14 @@
   <description>Remote Yarn package installer url for Helium dependency loader</description>
 </property>
 
+<!--
+<property>
+  <name>zeppelin.helium.registry</name>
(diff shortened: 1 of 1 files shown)


zjffdu:
LGTM

jolo-dev:
Question:
Why is that commented by default?
My problem is that if I click on Helium, there is no list as it was in 0.8.2. If I put it back into the `zeppelin-site.xml` it appears, so I am wondering what's the purpose of this?

alexott:
@jolo-dev many people use Zeppelin without direct connection to internet, and having remote helium registry enabled by default is causing delays during startup, or something like.  That's because it was disabled by default (in don't remember exact JIRA for that).

PR #3877, opened 2020-08-14
ZEPPELIN-4992 Editor settings are being set to the default interpreter regardless of the interpreter specified in the paragraph.

jason-ogaard:
### What is this PR for?
I wrote up this [defect](https://issues.apache.org/jira/browse/ZEPPELIN-4992) based on the behavior of the zeppelin 0.9.0 image found on the docker [hub](https://hub.docker.com/r/apache/zeppelin/tags). Currently, the editor settings aren't being applied to each interpreter, instead only the default interpreter editor settings get applied to every paragraph regardless of the interpreter specified.

This PR should fix that issue.

### What type of PR is it?
[Bug Fix]

### Todos
* None

### What is the Jira issue?
* [ZEPPELIN-4992](https://issues.apache.org/jira/browse/ZEPPELIN-4992)

* Ensure that the various niceties for each interpreter are applied. E.G. syntax highlighting, code auto-hiding.

### Questions:
No
No
No


zjffdu:
Thanks for the contribution @jason-ogaard LGTM

PR #3878, opened 2020-08-15
[ZEPPELIN-4996]. Improve table streaming display

zjffdu:
### What is this PR for?

This ticket improve the table streaming display by disabling the table append. Instead just refresh the whole output at once. This could make the frontend display more smooth without any shaking. (See screenshot below)

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4996

Manually tested

* Before
before

* Now

now

### Questions:


Reamer:
How can you enable/disable this option via the frontend?

zjffdu:
> How can you enable/disable this option via the frontend?

I didn't do it in frontend. I just change the backend. Before this PR, there will be several append operations for table output refresh, after this PR, there will be only one output update operation for each table output refresh. This is the reason with this PR streaming refresh is more smooth.

Leemoonsoo:
LGTM

PR #3884, opened 2020-08-19
[ZEPPELIN-5004] support running flink interpreter on kubernetes

zhoulii:
### What is this PR for?
add support to run flink interpreter on kubernetes

### What type of PR is it?
[Improvement]

### Todos
* [x] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5004

* manually tested in kubernetes cluster 
* https://travis-ci.org/github/zhoulii/zeppelin/builds/719203579

### Questions:
* Does the licenses files need update? **NO**
* Is there breaking changes for older versions? **NO**
* Does this needs documentation? **NO**

### DIFF
--- a/k8s/interpreter/100-interpreter-spec.yaml
+++ b/k8s/interpreter/100-interpreter-spec.yaml
@@ -81,7 +81,7 @@ spec:
   {% if zeppelin.k8s.interpreter.group.name == "flink" %}
     volumeMounts:
     - name: flink-home
-      mountPath: /flink
+      mountPath: /opt/flink
   initContainers:
   - name: flink-home-init
     image: {{zeppelin.k8s.flink.container.image}}
--- a/k8s/interpreter/100-interpreter-spec.yaml
+++ b/k8s/interpreter/100-interpreter-spec.yaml
@@ -78,6 +78,21 @@ spec:
   - name: spark-home
     emptyDir: {}
   {% endif %}
+  {% if zeppelin.k8s.interpreter.group.name == "flink" %}
+    volumeMounts:
+    - name: flink-home
+      mountPath: /flink
+  initContainers:
+  - name: flink-home-init
+    image: {{zeppelin.k8s.flink.container.image}}
+    command: ["sh", "-c", "cp -r /opt/flink/* /flink/"]
+    volumeMounts:
+    - name: flink-home
+      mountPath: /flink
+  volumes:
+  - name: flink-home
+    emptyDir: {}
+  {% endif %}
 ---
 kind: Service
 apiVersion: v1
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -100,6 +100,16 @@ function addEachJarInDirRecursiveForIntp(){
   fi
 }
 
+function addJarInDirRecursiveForIntpWithoutLog4j(){
+  if [[ -d "${1}" ]]; then
+    for jar in ${1}/*.jar; do
+      if [ "$(echo $jar | grep log4j)" = "" ]; then
+        ZEPPELIN_INTP_CLASSPATH="$jar:${ZEPPELIN_INTP_CLASSPATH}"
+      fi
+    done
+  fi
+}
+
 function addJarInDir(){
   if [[ -d "${1}" ]]; then
     ZEPPELIN_CLASSPATH="${1}/*:${ZEPPELIN_CLASSPATH}"
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -238,7 +238,7 @@ elif [[ "${INTERPRETER_ID}" == "pig" ]]; then
     echo "TEZ_CONF_DIR is not set, configuration might not be loaded"
   fi
 elif [[ "${INTERPRETER_ID}" == "flink" ]]; then
-  addEachJarInDirRecursiveForIntp "${FLINK_HOME}/lib"
+  addJarInDirRecursiveForIntpWithoutLog4j "${FLINK_HOME}/lib"
 
   FLINK_PYTHON_JAR=$(find "${FLINK_HOME}/opt" -name 'flink-python_*.jar')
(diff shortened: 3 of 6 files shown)


k8s/interpreter/100-interpreter-spec.yaml
```diff
@@ -78,6 +78,21 @@ spec:
...
     emptyDir: {}
   {% endif %}
+  {% if zeppelin.k8s.interpreter.group.name == "flink" %}
+    volumeMounts:
+    - name: flink-home
+      mountPath: /flink
+  initContainers:
+  - name: flink-home-init
+    image: {{zeppelin.k8s.flink.container.image}}
+    command: ["sh", "-c", "cp -r /opt/flink/* /flink/"]
```
Reamer, line 88: I think you should change your `ZEPPELIN_K8S_CONTAINER_IMAGE` and add flink to /opt/flink.

zjffdu:
Thanks for the contribution @zhoulii , I think this is only to run flink interpreter on k8s, what about the JobManager & TaskManager ? Do you have plan to run flink cluster in k8s in future ?

k8s/interpreter/100-interpreter-spec.yaml
zhoulii: Actually, Flink is alreadly in `/opt/flink` directory in `ZEPPELIN_K8S_CONTAINER_IMAGE `. command here aims to mv flink dist file from `ZEPPELIN_K8S_CONTAINER_IMAGE` to volume `flink-home` which is mounted to '/flink' directory in initContainers. 

In interpreter container, I also mounted `flink-home` to '/flink' directory, maybe '/opt/flink' directory would be better. So I changed flink home to ‘/opt/flink’ in the new commit. [100-interpreter-spec.yaml Lin84](https://github.com/apache/zeppelin/pull/3884/files#diff-1b67202451861591b920d5d4a9e31ce8R84) ,  [zeppelin-server.yaml Line41](https://github.com/apache/zeppelin/pull/3884/files#diff-56ccb2e2c2617b27dbaae866d9431e51R41)

zhoulii:
> Thanks for the contribution @zhoulii , I think this is only to run flink interpreter on k8s, what about the JobManager & TaskManager ? Do you have plan to run flink cluster in k8s in future ?

Hi @zjffdu , thanks for reviewing this pr. you are right, this pr only aims to run flink interpreter on k8s, and execution modes of flink job are still `local|remote|yarn`.

Indeed it would be more reasonable to run flink job in [flink-kubernetes-session](https://ci.apache.org/projects/flink/flink-docs-release-1.11/ops/deployment/native_kubernetes.html#flink-kubernetes-session) mode when flink interpreter is running on k8s. I will find out how to add this new execution mode to flink interpreter in next pr.

btw, is there any problem with the change in `common.sh` and `interpreter.sh`? I am not sure whether it is a good way to solve jar conflict of log4j.

k8s/interpreter/100-interpreter-spec.yaml
Reamer: > Actually, Flink is alreadly in /opt/flink directory in ZEPPELIN_K8S_CONTAINER_IMAGE

If Flink is already in your image (`ZEPPELIN_K8S_CONTAINER_IMAGE'), why do you need to copy Flink files that are in the initContainer into the interpreter container?

zhoulii: sorry for wrong description. Actually , I mean flink is in '/opt/flink'  directory in image `ZEPPELIN_K8S_FLINK_CONTAINER_IMAGE`.

Reamer: Okay, thanks for clarifying. Why don't you extend your `ZEPPELIN_K8S_CONTAINER_IMAGE` with Flink under `/opt/flink`? If you extend your image, copying should not be necessary.

zhoulii: that would need some manual work, including extend `ZEPPELIN_K8S_CONTAINER_IMAGE` and set flink home in interpreter, I want to do these work automatically.

Reamer: We can add Flink in PR #3859, just like we add Python and R packages.

zjffdu:
@zhoulii What is the log4j conflict issue ？

zhoulii:
@zjffdu I have only checked the second approach that filter out `$FLINK_HOME\lib\log4j*.jar`, and it makes logging correctly.

k8s/interpreter/100-interpreter-spec.yaml
Reamer: All in all you are right.
> add different kinds of interpreter related files together may make interpreter image too large.

Therefore we want to give the users the [responsibility](https://github.com/apache/zeppelin/pull/3859/files#diff-f55297559468e8c543ae5ab6ee356ec4R66) to build the best image for their use case. 
I think the copy process reduces the start time of the container and it makes the `interpreter-spec` more complicated. Today we have a special flink routine, tomorrow another one.

> it’s not flexible to change the interpreter related files，for example, if we need to change the flink version, we have to rebuild interpreter image

I think this is the biggest advantage and disadvantage of images. If you need to update any software version, including Flink, OpenJDK and operating system libraries, you have to rebuild your image. The advantage, you can make a nice rollback if your image contains all the application software. The disadvantage, of course, it that you have to rebuild your image, which takes quite a long time.

If you want to start certain interpreter images in your environment to reduce the image size, I recommend to create two or more interpreter images and inset a small if-then-else change [here](https://github.com/apache/zeppelin/blob/8ab7e9c473c5ce5b487bbab2a7b682f3fb165cab/k8s/interpreter/100-interpreter-spec.yaml#L45).

zhoulii: Maybe it's not a good idea to copy files from images, I will close this PR, Thanks for explanation.

PR #3892, opened 2020-08-27
[ZEPPELIN-5019] Enable configurable HTML addons

weand:
### What is this PR for?
Currently its hard to integrate certain plotting libraries in one single zeppelin notebook efficiently. efficiently here means in the sense of how often library code needs to be embedded into a single notebook. while few plotting libraries provide APIs to properly enable lazy loading of their JS library (e.g. bokeh), others do not (e.g. plotly), which results in large overhead in notebooks with couple of plots, where each plot embeds the library code again and again. attempts to only embed library code in the first paragraph is not feasible, as its not guaranteed, that this paragraph will be loaded at first by the browser. Any attempt to load libraries via paragraph code will end up in such synchronization issues when the library does not support synchronization on its own.

A general solution for that problem is to enable the loading of 3rd party JS libraries within the root HTML.

Hence this issue will introduce a new optional zeppelin configuration that allows administrators to add global addons to the zeppelin root HTML, so analysts can rely on that the JS library is already loaded for every paragraph.

As there exist two places in HTML where initial loading may be placed, the issue will introduce the following two configuration properties:

- `zeppelin.server.html.head.addon` > addon html code to be placed at the end of the html->head section 
- `zeppelin.server.html.body.addon` > addon html code to be placed at the end of the html->body section

Each property may contain any valid HTML code, e.g.

```
<script defer src="https://url/to/my/lib.min.js" /><script defer src="https://url/to/other/lib.min.js" />
```

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5019

* Unit Test added for index.html of `zeppelin-web`
* can also be manually tested by adding config properties mentioned above. Then check the resulting source of the zeppelin web application (old and new ui)

### Questions:

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -660,6 +660,21 @@
 
 <!--
 <property>
+  <name>zeppelin.server.html.body.addon</name>
+  <value><![CDATA[<script defer src="https://url/to/my/lib.min.js" /><script defer src="https://url/to/other/lib.min.js" />]]></value>
+  <description>Addon html code to be placed at the end of the html->body section in index.html delivered by zeppelin server.</description>
+</property>
+
+<property>
+  <name>zeppelin.server.html.head.addon</name>
+  <value></value>
+  <description>Addon html code to be placed at the end of the html->head section in index.html delivered by zeppelin server.</description>
+</property>
+-->
+
+
+<!--
+<property>
   <name>zeppelin.interpreter.callback.portRange</name>
   <value>10000:10010</value>
 </property>
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -717,6 +717,14 @@ public class ZeppelinConfiguration extends XMLConfiguration {
     return getString(ConfVars.ZEPPELIN_SERVER_STRICT_TRANSPORT);
   }
 
+  public String getHtmlHeadAddon() {
+    return getString(ConfVars.ZEPPELIN_SERVER_HTML_HEAD_ADDON);
+  }
+
+  public String getHtmlBodyAddon() {
+    return getString(ConfVars.ZEPPELIN_SERVER_HTML_BODY_ADDON);
+  }
+
   public String getLifecycleManagerClass() {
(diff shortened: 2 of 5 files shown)


weand:
@zjffdu, @Leemoonsoo can you have a look here, and review. 

I've tested it manually in development and normal (war) mode, and with old and new zeppelin UI successfully

thanks in advance.

Leemoonsoo:
Thank @weand for the contribution.
It looks good to me!

Just a side question, although [spell](http://zeppelin.apache.org/docs/0.9.0-SNAPSHOT/development/helium/writing_spell.html) is not designed for this, it basically installs & loads javascript packages in front-end. Do you think it makes sense to leverage this kind of plugin mechanism to install javascript in the future, or it is overkill?

weand:
> Just a side question, although [spell](http://zeppelin.apache.org/docs/0.9.0-SNAPSHOT/development/helium/writing_spell.html) is not designed for this, it basically installs & loads javascript packages in front-end. Do you think it makes sense to leverage this kind of plugin mechanism to install javascript in the future, or it is overkill?

I'm not sure about existing use cases, but for me a helium spell plugin seems to be very limited in general, because they can't be integrated with Server Side Interpreters (e.g. %pyspark). Just for installing additional javascript code / libraries they are really overkill. Especially because of additional maintenance / versioning effort when just delivering libraries.

If there are no other comments within next few days, could you then please merge.

weand:
@zjffdu @Leemoonsoo I also thought about adding unit test, which would rely on existence of `zeppelin-web/dist/index.html` and/or  `zeppelin-web-angular/dist/index.html`. Seems for [core-tests](https://github.com/apache/zeppelin/blob/master/.travis.yml#L84) travis only builds `zeppelin-web` and not `zeppelin-web-angular` yet. could that module be activated for core tests as well, or is that no good idea?

weand:
Tests added. 

Added `@Ignore` to Test for zeppelin-web-angular, as its not build yet for [core tests](https://github.com/apache/zeppelin/blob/master/.travis.yml#L73-L84))

Tests showed, that my manual tests with `zeppelin-web` were not that good. Additional logic was required for the case, when closing html tags do not exist (which is the case for build results of `zeppelin-web`).

weand:
@Leemoonsoo 

squashed and rebased the changes. could you have a quick look again. and merge if there are no more comments.

Leemoonsoo:
Thank @weand! looks good to me. Merging this to master and branch-0.9

PR #3910, opened 2020-09-15
[ZEPPELIN-5041]. Move interpreter lifecyclemanager from zeppelin server to interpreter process

zjffdu:
### What is this PR for?
Currently, interpreter lifecycle manager run in zeppelin server side. One short coming is that interpreter process may still leak if zeppelin server exist unexpectedly. This PR move lifecyclemanager from zeppelin server to interpreter process, so that if user use TimeoutLifecycleManager, interpreter process will exit after a idle time even zeppelin server is existed. 

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5041

* CI pass & manually tested

### Questions:


Reamer:
Overall looks really good.
Before merging we should solve the following things:
 - Transfer Zeppelin server configuration to a remote interpreter (not solved in K8s launcher)

zjffdu:
Sorry @Reamer I didn't notice your comment, let me do it another PR.

PR #3917, opened 2020-09-22
[ZEPPELIN-5053]. Transfer Zeppelin server configuration to a remote interpreter

zjffdu:
### What is this PR for?

This PR add method `init` to pass `ZeppelinConfiguration` in the thrift interface between zeppelin server and interpreter process. And put lifecycle manager initialization in the init method. Zeppelin server will call this init method after interpreter process is started (register with zeppelin server)

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5053

* https://travis-ci.org/github/zjffdu/zeppelin/builds/730513842

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
```diff
@@ -123,6 +123,7 @@ public void open() throws InterpreterException {
           try {
             if (!(interpreter instanceof ConfInterpreter)) {
               ((RemoteInterpreter) interpreter).internal_create();
+              ((RemoteInterpreter) interpreter).internal_create();
```
Reamer, line 126: Typo?

Reamer:
When this is ready for review, let me know.

zjffdu:
@Reamer CI is passed，could you help review and verify it on k8s ?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreter.java
zjffdu: What typo do you mean ?

Reamer: double function call:
`((RemoteInterpreter) interpreter).internal_create();` is called twice

zjffdu: Oops，

zjffdu: Fixed

Reamer:
> @Reamer CI is passed，could you help review and verify it on k8s ?

I will test this on K8s today.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -184,12 +185,18 @@ public static synchronized ZeppelinConfiguration create() {
...
+    return this.properties;
+  }
+
   public static void reset() {
     conf = null;
   }
 
   public void setProperty(String name, String value) {
-    this.properties.put(name, value);
+    if (name != null && value != null) {
```
Reamer, line 197: Maybe `StringUtils.isNoneBlank()`  is better in this place, because with this code you can add keys and/or values like `" "`.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -76,14 +76,15 @@ private void initProperties() {
     for (ConfigurationNode p : nodes) {
       String name = (String) p.getChildren("name").get(0).getValue();
       String value = (String) p.getChildren("value").get(0).getValue();
-      if (!StringUtils.isEmpty(name)) {
+      if (!StringUtils.isEmpty(name) && !StringUtils.isEmpty(value)) {
```
Reamer, line 79: Here I would also prefer `StringUtils.isNoneBlank()`.

Reamer (changes requested):
Overall really good, just two minor comments.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
zjffdu: Good point, fixed

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
zjffdu: Fixed

Reamer:
> > @Reamer CI is passed，could you help review and verify it on k8s ?
> 
> I will test this on K8s today.

Hi @zjffdu,
I have tested it on K8s. Unfortunately it does not work.
```
INFO [2020-09-28 15:51:07,059] ({pool-3-thread-2} RemoteInterpreterServer.java[createLifecycleManager]:374) - Creating interpreter lifecycle manager: org.apache.zeppelin.interpreter.lifecycle.NullLifecycleManager
```
I have found the reason for this. At the moment I have configured Zeppelin only via the environment variable, as described in [our example](https://github.com/apache/zeppelin/blob/536e66ce76aaefabeb3784bdc8013f7a2513d7f0/k8s/zeppelin-server.yaml#L17-L38). If you have no configuration file `properties` are empty.

I see two ways to solve this problem.
 - Load the same environment key value map into the interpreter pod -> very easy and secure. Just copy this [part](https://github.com/apache/zeppelin/blob/536e66ce76aaefabeb3784bdc8013f7a2513d7f0/k8s/zeppelin-server.yaml#L143-L145) into [interpreter-spec.yaml](https://github.com/apache/zeppelin/blob/master/k8s/interpreter/100-interpreter-spec.yaml)
 - Filling properties during startup in the Zeppelin server

zjffdu:
Thanks @Reamer for the testing, I would prefer the second approach. Because I think this not only happens in k8s, in other modes it also possible to define configuration via env.

zjffdu:
@Reamer I have fixed that in another commit, please help try that in k8s.

Reamer:
> Thanks @Reamer for the testing, I would prefer the second approach. Because I think this not only happens in k8s, in other modes it also possible to define configuration via env.

You are right, we should also find a solution for other modes.

> @Reamer I have fixed that in another commit, please help try that in k8s.

Now it works on K8s with an environment configuration setup.

Reamer (approved):

zjffdu:

PR #3918, opened 2020-09-22
[ZEPPELIN-5055]. Interpreter log is missing in zeppelin server docker container

zjffdu:
### What is this PR for?

The root cause in the log4j.properties in the zeppelin server image. For k8s, we should use stdout for logging, but for the single zeppelin server container, we should use dailyfile for logging (both zeppelin server and interpreter). This PR add a new env `ZEPPELIN_IN_DOCKER` to choose the right log4j properties file for the right scenario. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5055

* CI pass

### Questions:

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -140,11 +140,17 @@ if [[ ( -z "${ZEPPELIN_INTP_MEM}" ) && ( "${ZEPPELIN_INTERPRETER_LAUNCHER}" != "
 fi
 
 JAVA_OPTS+=" ${ZEPPELIN_JAVA_OPTS} -Dfile.encoding=${ZEPPELIN_ENCODING} ${ZEPPELIN_MEM}"
-JAVA_OPTS+=" -Dlog4j.configuration=file://${ZEPPELIN_CONF_DIR}/log4j.properties"
+if [[ -n "${ZEPPELIN_IN_DOCKER}" ]]; then
+    JAVA_OPTS+=" -Dlog4j.configuration=file://${ZEPPELIN_CONF_DIR}/log4j_docker.properties"
+else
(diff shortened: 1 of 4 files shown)


Reamer:
Hello @zjffdu,
Dockers should not write the logs to the OverlayFS. I think that our current [log4j.properties](https://github.com/apache/zeppelin/blob/master/scripts/docker/zeppelin/bin/log4j.properties) is sufficient.
You can view the logs with `docker logs xyz`.

If you write your logs only with 'ConsoleAppender'', then you can get all container logs with an agent like [Filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-autodiscover.html#_docker_2) running on the host system.

zjffdu:
But this command can only see the zeppelin server log, the issue here is that the interpreter log is missing because it is also written to stdout

Reamer:
The Interpreting process runs inside the Zeppelin server container?

zjffdu:
Right, the scenario is that user run zeppelin-server in a docker container, he doesn't need k8s cluster, just docker container runtime. This is especially useful for new users who want to try zeppelin in a docker container without any setup. Here's the doc https://github.com/apache/zeppelin/blob/master/docs/setup/deployment/docker.md
BTW, it looks like the doc is missing in zeppelin site, we should also fix this as well.

zjffdu:
Any more comment @Reamer

Reamer:
> Any more comment ? @Reamer

No.

zjffdu:

PR #3922, opened 2020-09-24
[ZEPPELIN-5049]. NullPointerException when running z.run(noteId, paragraphId)

zjffdu:
# What is this PR for?

The root cause is that the interpreter context of python thread is null. This PR fix this issue by calling method `setInterpreterContextInPython` in python side. It fix this issue in both PySparkInterpreter & IPySparkInterpreter. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5049

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

Reamer:
I tested this manual and it seems to work.
+1 for merge

PR #3927, opened 2020-09-27
[ZEPPELIN-4409]. Set spark.app.name to know which user is running which notebook

zjffdu:
### What is this PR for?

Use  the interpreter group id as the spark app name if it is not set. So that user can easily figure out which yarn app is what he is looking for no matter it is note isolated or user isolated. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4409

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -78,6 +78,15 @@ public class SparkInterpreterLauncher extends StandardInterpreterLauncher {
       }
     }
 
+    // set spark.app.name if it is not set or empty
+    if (!sparkProperties.containsKey("spark.app.name") ||
+            StringUtils.isBlank(sparkProperties.getProperty("spark.app.name"))) {
+      sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId());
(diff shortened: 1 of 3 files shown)


zjffdu:
Will merge if no more comment

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -78,6 +78,15 @@ public SparkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
       }
     }
 
+    // set spark.app.name if it is not set or empty
+    if (!sparkProperties.containsKey("spark.app.name") ||
+            StringUtils.isBlank(sparkProperties.getProperty("spark.app.name"))) {
+      sparkProperties.setProperty("spark.app.name", context.getInterpreterGroupId());
+    } else {
+      String appName = context.getInterpreterGroupId();
+      sparkProperties.setProperty("spark.app.name", appName);
```
Reamer, line 87: @zjffdu Why do you overwrite the value?

zjffdu: oops, I should remove this.

PR #3933, opened 2020-10-03
[ZEPPELIN-5077] Run all paragraphs does not stop on failing paragraph

olkkoti:
### What is this PR for?

If a paragraph fails on something other than exception in Zeppelin, notebook will continue running paragraphs after it. This is unintuitive and regression from previous version.

To fix it, return "false" from running a paragraph, if it the actual user code fails. Also, after a failing "run all" operation, we need to broadcast paragraph states to client, so they wont stay in pending state after a failed run.

### What type of PR is it?

Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-5077

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -1488,8 +1488,15 @@ public class NotebookServer extends WebSocketServlet
             new TypeToken<List<Map<String, Object>>>() {
             }.getType());
 
-    getNotebookService().runAllParagraphs(noteId, paragraphs, getServiceContext(fromMessage),
-        new WebSocketServiceCallback<Paragraph>(conn));
+    if (!getNotebookService().runAllParagraphs(noteId, paragraphs, getServiceContext(fromMessage),
+        new WebSocketServiceCallback<Paragraph>(conn))) {
+      Note note = getNotebookService().getNote(noteId, getServiceContext(fromMessage), new SimpleServiceCallback());
+      if (note != null) {
+        for (Paragraph p : note.getParagraphs()) {
+          broadcastParagraph(note, p, null);
+        }
+      }
+    }
   }
 
   private void broadcastSpellExecution(NotebookSocket conn,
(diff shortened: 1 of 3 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1488,8 +1488,15 @@ private void runAllParagraphs(NotebookSocket conn,
             new TypeToken<List<Map<String, Object>>>() {
             }.getType());
 
-    getNotebookService().runAllParagraphs(noteId, paragraphs, getServiceContext(fromMessage),
-        new WebSocketServiceCallback<Paragraph>(conn));
+    if (!getNotebookService().runAllParagraphs(noteId, paragraphs, getServiceContext(fromMessage),
+        new WebSocketServiceCallback<Paragraph>(conn))) {
+      Note note = getNotebookService().getNote(noteId, getServiceContext(fromMessage), new SimpleServiceCallback());
```
zjffdu, line 1493: Could you add one comment to explain why we need to broadcast paragraphs to client ?

zjffdu:
@olkkoti Thanks for the contribution, could you add unit for it or create a follow up ticket for unit test ?

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
olkkoti: Client sets all paragraph states to "pending". Now, if we stop running paragraphs when one fails, paragraphs after the failed one are still in state "pending" in client. Broadcasting them to client resets the states.

zjffdu: I mean to add comment in the code so that user can understand what this piece of code is doing.

olkkoti: Done.

olkkoti:
Test added.

zjffdu:
@olkkoti What is the latest travis CI link ?

olkkoti:
https://travis-ci.org/github/olkkoti/zeppelin/builds/734680686

zjffdu:
@olkkoti There're still test fails.
```
Failed tests: 
  NotebookRestApiTest.testRunParagraphSynchronously:181 expected:<200> but was:<500>
  NotebookServiceTest.testParagraphOperations:391 null
```

olkkoti:
Made a cleaner fix which does not seem to break existing functionality. Can't get Flink tests to pass though. :/

https://travis-ci.org/github/olkkoti/zeppelin/builds/736283820

zjffdu:
> Made a cleaner fix which does not seem to break existing functionality. Can't get Flink tests to pass though. :/
> 
> https://travis-ci.org/github/olkkoti/zeppelin/builds/736283820

Thanks @olkkoti The flink tests are unrelated, +1

PR #3935, opened 2020-10-06
ZEPPELIN-5081: Add loopPostInit to bind $intp

khud:
### What is this PR for?
There is a bug in the project: $intp variable is not defined for Spark/Scala 2.12 interpreter. It looks like this part of REPL in Spark 2.12 and in Spark 2.11 is the same, so I use the code from Spark/Scala 2.11 interpreter to fix it.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[ZEPPELIN-5081]

* Tested manually with master and Spark 2.4.2/Scala 2.12.8

### DIFF
--- a/spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java
+++ b/spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java
@@ -136,6 +136,9 @@ public class SparkInterpreterTest {
     result = interpreter.interpret("/*line 1 \n line 2*/print(\"hello world\")", getInterpreterContext());
     assertEquals(InterpreterResult.Code.SUCCESS, result.code());
 
+    // test $intp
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks for the PR, could you provide more details about how to reproduce this issue ? And it would be better to add unit test for it.

khud:
Default behavior: 0.9.0-preview2/spark-2.4.2/scala-2.12: 

image

Fixed behavior: master + my fix/spark-2.4.2/scala-2.12:

image

zjffdu:
Could you add unit test here ? so that we won't cause regression in future https://github.com/apache/zeppelin/blob/master/spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkInterpreterTest.java#L70

khud:
Yes, I will do

khud:
Done

zjffdu:
LGTM

PR #3936, opened 2020-10-09
[ZEPPELIN-5088]. Explain statement is broken in flink interpreter

zjffdu:
### What is this PR for?

This PR fix the explain statement issue in flink interpreter. For flink 1.10, 1.11, explain works in different approach, so we need to use shim to do that. This PR also add unit test to prevent regression in future. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5088

* CI pass
https://travis-ci.org/github/zjffdu/zeppelin/builds/734175742

### Questions:


zjffdu:
will merge if no more comment

PR #3950, opened 2020-10-22
Branch 0.9

WingsLong:
NOT FOUND CLASS IN zeppelin-0.9.0-preview2

org.apache.zeppelin.interpreter.jupyter.proto.CancelRequest

org.apache.zeppelin.interpreter.jupyter.proto.*   All Class is not found in source


Reamer:
I am closing this PR because it makes no sense to me.

PR #3951, opened 2020-10-23
[ZEPPELIN-5102]. Simplify spark sql error message

zjffdu:
### What is this PR for?

In this PR, we only print the error message when it is `AnalysisException`, this would make the error message more clear for users.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5102

* Manually tested

image

### Questions:


zjffdu:
Will merge if no more comment

PR #3952, opened 2020-10-23
[ZEPPELIN-5103]. Simply jdbc interpreter error message

zjffdu:
### What is this PR for?

In this PR, I would only print exception message when it is SQLException, otherwise would print the full stacktrace, because in this case it is most likely due to jdbc interpreter's internal error.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5103

* CI pass

image

### Questions:


zjffdu:
Will merge if no more comment

PR #3954, opened 2020-10-27
[ZEPPELIN-5106]. Only save note when it is loaded or newly created

zjffdu:
### What is this PR for?

This PR is to save note only in 2 cases:
1. It is loaded. 
2. It is newly created. 

Otherwise it will throw NPE because paragraphs is null when note is in unloaded state.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5106

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #3957, opened 2020-10-28
[ZEPPELIN-5114]. Add api to ZeppelinClient for import note

zjffdu:
### What is this PR for?

This PR just add api to ZeppelinClient for importing note. This api would return noteId if importing is successful, otherwise it would throw exception.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5114

* Unit test is added
https://travis-ci.org/github/zjffdu/zeppelin/builds/739823431

### Questions:


zjffdu:
Will merge if no more comment

PR #3958, opened 2020-10-29
[ZEPPELIN-5112]. upgrade-note.sh throws exception while upgrading notes on hdfs storage

zjffdu:
### What is this PR for?

The root cause is that we didn't put hadoop jars in the classpath in `upgrade-note.sh`. This PR update script `upgrade-note.sh`. Besides that this PR also fix an upgrading issue in `ParagraphRuntimeInfo.java`. The `values` is List<String> in old version of zeppelin, but it is changed to List<Map<String, String>> in new version of zeppelin.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5112

* CI pass
https://travis-ci.org/github/zjffdu/zeppelin/builds/739821895

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfo.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfo.java
@@ -18,7 +18,7 @@ public class ParagraphRuntimeInfo {
 
   // runtimeInfos job url or dropdown-menu key in
   // zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
-  private List<Map<String, String>> values;  // values for the key-value pair property
+  private List<Object> values;  // values for the key-value pair property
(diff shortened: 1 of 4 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/ParagraphRuntimeInfo.java
```diff
@@ -18,7 +18,7 @@
 
   // runtimeInfos job url or dropdown-menu key in
   // zeppelin-web/src/app/notebook/paragraph/paragraph-control.html
-  private List<Map<String, String>> values;  // values for the key-value pair property
+  private List<Object> values;  // values for the key-value pair property
```
Reamer, line 21: Why do you change the type?

zjffdu:
Will merge if no more comment

PR #3961, opened 2020-10-30
[ZEPPELIN-5116] Accessing zeppelin via knox after knox logout should be redirected to  knox login page

PrarthiJain:
### What is this PR for?
When zeppelin is running under kerberosRealm the logout doesn't work, it should be redirected to knox login page. With this PR, a fix for this is provided.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-5116

1. Created a cluster
2. Deploy the changes to the cluster
3. Added respective entries in shiro.ini
4. Performed logout and tested

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/kerberos/KerberosRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/kerberos/KerberosRealm.java
@@ -67,14 +67,15 @@ import java.util.regex.Pattern;
  * The Shiro configuration section should be configured as:
  * [main]
  * krbRealm = org.apache.zeppelin.realm.kerberos.KerberosRealm
- * krbRealm.principal=HTTP/zeppelin.fqdn.domain.com@EXAMPLE.COM
- * krbRealm.keytab=/etc/security/keytabs/spnego.service.keytab
- * krbRealm.nameRules=DEFAULT
- * krbRealm.signatureSecretFile=/etc/security/http_secret
- * krbRealm.tokenValidity=36000
- * krbRealm.cookieDomain=domain.com
- * krbRealm.cookiePath=/
- * krbRealm.logout=/logout
+ * krbRealm.principal = HTTP/zeppelin.fqdn.domain.com@EXAMPLE.COM
+ * krbRealm.keytab = /etc/security/keytabs/spnego.service.keytab
+ * krbRealm.nameRules = DEFAULT
+ * krbRealm.signatureSecretFile = /etc/security/http_secret
+ * krbRealm.tokenValidity = 36000
+ * krbRealm.cookieDomain = domain.com
+ * krbRealm.cookiePath = /
+ * krbRealm.logout = logout
+ * krbRealm.logoutAPI = true
  * krbRealm.providerUrl = https://domain.example.com/
  * krbRealm.redirectParam = originalUrl
  * authc = org.apache.zeppelin.realm.kerberos.KerberosAuthenticationFilter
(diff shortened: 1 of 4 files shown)


PrarthiJain:
@VipinRathor @prabhjyotsingh @zjffdu  can you please help in reviewing

zeppelin-server/src/main/java/org/apache/zeppelin/realm/kerberos/KerberosRealm.java
```diff
@@ -91,6 +94,9 @@
   private boolean isCookiePersistent = false;
   private String signatureSecretFile = null;
   private String signatureSecretProvider = "file";
+  private String logout = "/logout";
+  private String providerUrl = "https://domain.example.com/";
+  private String redirectParam = "originalUrl";
```
prabhjyotsingh, line 99: I think we will need `logoutAPI` as well, similar to the implementation of KnoxJwtRealm.

zeppelin-server/src/main/java/org/apache/zeppelin/realm/kerberos/KerberosRealm.java
```diff
@@ -91,6 +94,9 @@
   private boolean isCookiePersistent = false;
   private String signatureSecretFile = null;
   private String signatureSecretProvider = "file";
+  private String logout = "/logout";
```
prabhjyotsingh, line 97: this should be without slash `/`.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -251,19 +265,27 @@ public Response logout() {
     }
     if (isKnoxSSOEnabled()) {
       KnoxJwtRealm knoxJwtRealm = getJTWRealm();
-      data.put("redirectURL", constructKnoxUrl(knoxJwtRealm, knoxJwtRealm.getLogout()));
+      StringBuilder redirectURL = new StringBuilder(knoxJwtRealm.getProviderUrl());
```
prabhjyotsingh, line 268: Instead of passing StringBuilder can this be only String?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/LoginRestApi.java
```diff
@@ -251,19 +265,27 @@ public Response logout() {
...
+      String redirect = knoxJwtRealm.getRedirectParam();
+      data.put("redirectURL", constructUrl(redirectURL, redirect,
+          knoxJwtRealm.getLogout()));
       data.put("isLogoutAPI", knoxJwtRealm.getLogoutAPI().toString());
+    } else if (isKerberosRealmEnabled()) {
+      KerberosRealm kerberosRealm = getKerberosRealm();
+      StringBuilder redirectURL = new StringBuilder(kerberosRealm.getProviderUrl());
+      String redirect = kerberosRealm.getRedirectParam();
+      data.put("redirectURL", constructUrl(redirectURL, redirect, kerberosRealm.getLogout()));
```
prabhjyotsingh, line 277: If `isLogoutAPI` is implemented from the previous comment, then we will need to add the same here.

prabhjyotsingh:
@PrarthiJain  can you fix https://github.com/apache/zeppelin/blob/master/zeppelin-web/src/components/navbar/navbar.controller.js#L110 as well, right now it only refreshes when API returns a non-20x response.

prabhjyotsingh:
Looks good to me, @PrarthiJain can you paste you travis-build link as well?

PrarthiJain:
Thanks, @prabhjyotsingh for reviewing. Here is the Travis-build link: https://travis-ci.com/github/PrarthiJain/zeppelin/builds/197124228

prabhjyotsingh (approved):
LGTM, will merge this if no more discussion.

PR #3965, opened 2020-11-05
[ZEPPELIN-5121]. Improve the support of hive on kerbose

zjffdu:
### What is this PR for?

Several improvements of hive on kerbors support
1. Remove ·`zeppelin.jdbc.auth.kerberos.proxy.enable`, it is never documented, use `prefix.proxy.user.property` is enough.
2. Append principal to url automatically for users. 
3. Remove the doAs code block, it looks like is not necessary.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5121

* CI pass

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -100,7 +100,6 @@ import org.apache.zeppelin.user.UsernamePassword;
 public class JDBCInterpreter extends KerberosInterpreter {
   private static final Logger LOGGER = LoggerFactory.getLogger(JDBCInterpreter.class);
 
-  static final String INTERPRETER_NAME = "jdbc";
   static final String COMMON_KEY = "common";
   static final String MAX_LINE_KEY = "max_count";
   static final int MAX_LINE_DEFAULT = 1000;
@@ -243,7 +242,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
   protected boolean isKerboseEnabled() {
     if (!isEmpty(getProperty("zeppelin.jdbc.auth.type"))) {
-      UserGroupInformation.AuthenticationMethod authType = JDBCSecurityImpl.getAuthtype(properties);
+      UserGroupInformation.AuthenticationMethod authType = JDBCSecurityImpl.getAuthType(properties);
       if (authType.equals(KERBEROS)) {
         return true;
       }
@@ -447,6 +446,9 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private void createConnectionPool(String url, String user, String dbPrefix,
       Properties properties) throws SQLException, ClassNotFoundException {
 
+    LOGGER.info("Creating connection pool for url: {}, user: {}, dbPrefix: {}, properties: {}",
+            url, user, dbPrefix, properties);
+
     String driverClass = properties.getProperty(DRIVER_KEY);
     if (driverClass != null && (driverClass.equals("com.facebook.presto.jdbc.PrestoDriver")
             || driverClass.equals("io.prestosql.jdbc.PrestoDriver"))) {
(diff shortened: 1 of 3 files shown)


Reamer:
I know this has nothing to do with your change, but the HiveMonitor thread seems to run forever after the start. It also seems that one thread is created per statement.
Is that right?

zjffdu:
@Reamer It would finish after the hive sql is finished. 
```
while (hiveStmt.hasMoreLogs() && !Thread.interrupted()) {
```

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -500,54 +501,21 @@ public Connection getConnection(String dbPrefix, InterpreterContext context)
...
-                ugi = UserGroupInformation.createProxyUser(
-                    user, UserGroupInformation.getCurrentUser());
-              } catch (Exception e) {
-                LOGGER.error("Error in getCurrentUser", e);
-                throw new InterpreterException("Error in getCurrentUser", e);
-              }
-
-              final String poolKey = dbPrefix;
-              try {
-                connection = ugi.doAs(new PrivilegedExceptionAction<Connection>() {
```
prabhjyotsingh, line 534: This Kerberos case looks important? Will the "doas" case continue to work for hive/phoenix without this?

prabhjyotsingh: I see, line:534, sounds like it should work. let me test this latest patch. Will update this thread.

prabhjyotsingh:
@zjffdu  I'm using following JDBC config:
```
hive.driver org.apache.hive.jdbc.HiveDriver 
hive.user hive  
hive.password   
hive.proxy.user.property  hive.server2.proxy.user 
hive.url  jdbc:hive2://phivea-3.phivea.root.psk.site:2181/;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=hiveserver2;retries=5;hiveCreateAsExternalLegacy=true  
hive.splitQueries true


zeppelin.jdbc.keytab.location /var/run/cloudera-scm-agent/process/180-zeppelin-ZEPPELIN_SERVER/zeppelin.keytab
zeppelin.jdbc.principal zeppelin/phivea-1.phivea.root.hwx.site@ROOT.HWX.SITE
zeppelin.jdbc.auth.type KERBEROS
```

And it does not work, fails with:
```
Could not open client transport for any of the Server URI's in ZooKeeper: Peer indicated failure: GSS initiate failed

**Comment by zjffdu** [comment]
@prabhjyotsingh Do you have the full jdbc interpreter log ?

**Comment by prabhjyotsingh** [comment]
Hope this helps:

This is the query that was executed:
```
%jdbc(hive)
select current_user()
```
Log - https://pastebin.com/5NxLUCpY

**Comment by zjffdu** [comment]
@prabhjyotsingh This is zeppelin server log, could you attach the jdbc interpreter log ?

**Comment by prabhjyotsingh** [comment]
There are no jdbc log
```
zeppelin@phivea-1:~/zeppelin$ ls -alh /var/log/zeppelin/
total 2.7M
drwxrwxrwx  2 zeppelin zeppelin 4.0K Nov  8 04:19 .
drwxr-xr-x 23 root     root     4.0K Nov  8 03:27 ..
-rw-rw-r--  1 zeppelin zeppelin    0 Nov  8 04:19 zeppelin-interpreter-jdbc-shared_process-zeppelin-phivea-1.phivea.root.hwx.site.log
-rw-rw-r--  1 zeppelin zeppelin 4.5K Nov  8 04:19 zeppelin-interpreter-md-shared_process-zeppelin-phivea-1.phivea.root.hwx.site.log
-rw-rw-r--  1 zeppelin zeppelin 2.7M Nov  8 04:20 zeppelin-zeppelin-phivea-1.phivea.root.hwx.site.log
-rw-rw-r--  1 zeppelin zeppelin 4.3K Nov  8 04:14 zeppelin-zeppelin-phivea-1.phivea.root.hwx.site.out

```

**Comment by prabhjyotsingh** [comment]
I could not make jdbc log to work, looks like there is a conflicting jar/property that gets downloaded with org.apache.hive:hive-jdbc. But here are hive-server logs:
When it does not work (with patch applied): https://pastebin.com/GZPMw26q
When it works (without patch): https://pastebin.com/6CZkxQC0

**Comment by zjffdu** [comment]
It looks like you are using ranger, not sure whether it is related. And if there's no log file, this is an issue that we should fix as well. (I remember I can produce the jdbc log file, not sure how it doesn't work for you)

**Comment by prabhjyotsingh** [comment]
This only happens is I specify jdbc-hive (CDH) that is somehow bundlling log4j.property which is causing no logs getting generated for jdbc interpreter.

But like i said on the same cluster without this patch things work.

File: jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java [hunk 2]
```diff
@@ -500,53 +502,41 @@ public Connection getConnection(String dbPrefix, InterpreterContext context)
     setUserProperty(dbPrefix, context);
 
     final Properties properties = jdbcUserConfigurations.getPropertyMap(dbPrefix);
-    final String url = properties.getProperty(URL_KEY);
+    String url = properties.getProperty(URL_KEY);
+    String principal = getProperty("zeppelin.jdbc.principal");
```
prabhjyotsingh, line 506: I did not notice this earlier, if I remove these three lines, it starts working again as expected.

prabhjyotsingh: In short for the cluster that I've configured this 
```jdbc:hive2://phivea-3.phivea.root.hwx.site:2181/;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=hiveserver2;retries=5;hiveCreateAsExternalLegacy=true;hive.server2.proxy.user=admin;
``` 
works, but 
```jdbc:hive2://phivea-3.phivea.root.hwx.site:2181/;serviceDiscoveryMode=zooKeeper;zooKeeperNamespace=hiveserver2;retries=5;hiveCreateAsExternalLegacy=true;principal=zeppelin/phivea-1.phivea.root.hwx.site@ROOT.HWX.SITE;hive.server2.proxy.user=admin;
``` 
fails

zjffdu: It is weird it fails with principal specified, do you have the exact error message ? IIUC, the principal is required. It should fails if it is not specified.

prabhjyotsingh: Yes, it works. Not sure may be something changed in the recent version of hive. Should we make this a configurable?

zjffdu: Let me revert this change. Originally I just want to remove the redundant principal in `default.url`, because we already ask user to specify principal in `zeppelin.jdbc.principal`, if hive behaivor changes, it is better not to append principal for user

zjffdu: @prabhjyotsingh I have removed this change, please help test again.

prabhjyotsingh (approved):
LGTM

PR #3970, opened 2020-11-13
[ZEPPELIN-5131]. Code completion pop up error message in shell interpreter

zjffdu:
### What is this PR for?

Minor PR to fix the code completion pop up issue in shell interpreter. Besides that, I also did some code refactoring in this PR.

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5131

* CI pass

### Questions:


Reamer:
Can you please avoid String concatenation during logging, because Strings are immutable.

zjffdu:
@Reamer Thanks for the review, I have fixed string concatenation

PR #3974, opened 2020-11-18
[ZEPPELIN-5135]. Keep default.splitQueries in jdbc

zjffdu:
### What is this PR for?

This PR is to bring back `default.splitQueries` in jdbc, so that user can choose whether split sql. But by default, we still enable it.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5135

* CI pass

### Questions:


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -19,6 +19,7 @@
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 import static org.apache.hadoop.security.UserGroupInformation.AuthenticationMethod.KERBEROS;
 
+import com.beust.jcommander.internal.Lists;
```
Reamer, line 22: wrong import?

zjffdu: Thanks, fixed now.

PR #3983, opened 2020-11-26
[ZEPPELIN-5141]. Solve the incompatibility issue of hive below 2.3

zhengslei:
### What is this PR for?
The current hive interpreter (jdbc) cannot work with hive below 2.3.  The code "hiveStmt.setInPlaceUpdateStream(InPlaceUpdateStream stream)" in HiveUtils uses InPlaceUpdateStream class which is only available in version 2.3 or above, which leads to runtime NoClassDefFoundError.  Move the code into ProgressBar to delay NoClassDefFoundError of InPlaceUpdateStream until ProgressBar instanced. When hive < 2.3, ProgressBar will not be instanced, so it works well.

### What type of PR is it?
[Bug Fix | Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5141

* Test the hive interpreter(jdbc) with version both above 2.3 and below 2.3 to check whether it can work well.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/hive/HiveUtils.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/hive/HiveUtils.java
@@ -24,6 +24,7 @@ import org.apache.zeppelin.jdbc.JDBCInterpreter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks @zhengslei LGTM

PR #3989, opened 2020-12-08
[ZEPPELIN-5150]. Allow option to allow exclude paragraph result when saving note to NotebookRepo

zjffdu:
### What is this PR for?

When using zeppelin as job server instead of interactive notebook, it is not necessary to save paragraph results (sometimes it is pretty large) into NotebookRepo which may cause high pressure on IO when many jobs are running. It doesn't affect current behavior, because by default we still store paragraph results in NotebookRepo

### What type of PR is it?
[Improvement | Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5150

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/SessionManagerService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/SessionManagerService.java
@@ -173,13 +173,13 @@ public class SessionManagerService {
         } else {
           // if it is running before, but interpreterGroup is not running now, that means the session is stopped.
           // e.g. InterpreterProcess is terminated for whatever unexpected reason.
-          if (sessionInfo.getState().equals(SessionState.RUNNING.name())) {
+          if (SessionState.RUNNING.name().equalsIgnoreCase(sessionInfo.getState())) {
             sessionInfo.setState(SessionState.STOPPED.name());
           }
         }
       }
     } else {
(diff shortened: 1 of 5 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/SessionManagerService.java
```diff
@@ -173,7 +173,7 @@ public SessionInfo getSessionInfo(String sessionId) throws Exception {
         } else {
           // if it is running before, but interpreterGroup is not running now, that means the session is stopped.
           // e.g. InterpreterProcess is terminated for whatever unexpected reason.
-          if (sessionInfo.getState().equals(SessionState.RUNNING.name())) {
+          if (sessionInfo.getState() != null && sessionInfo.getState().equals(SessionState.RUNNING.name())) {
```
Reamer, line 176: You could swap the objects of the 'equals'. A null check should then no longer be necessary.
```
SessionState.RUNNING.name().equals(sessionInfo.getState())
```

zjffdu: Right, thanks for pointing this

Reamer:
LGTM

PR #3991, opened 2020-12-09
[ZEPPELIN-5154]. Add option to disable broadcast Paragraph status/progress to frontend

zjffdu:
### What is this PR for?

Broadcast paragraph status/progress will be a very heavy operation if there's many jobs running, so I add one option to disable it. When zeppelin is used as job server instead of interactive notebook, it is not necessary to broadcast paragraph status/progress in real time. 

### What type of PR is it?
[ Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5154

* CI pass

### Questions:


zjffdu:
CI pass, will merge if no more comments

PR #3992, opened 2020-12-09
[ZEPPELIN-5153]. Too many unnecessary indexing in LuceneSearch

zjffdu:
### What is this PR for?

There's many unnecessary indexing in LuceneSearch which cause performance issue when there' lots of jobs running in zeppelin. This PR is to fix it.
* Only index the current paragraph when paragraph is updated.( All the paragraphs will be indexed 
* Only index/update note name index when note is updated.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5153

* Manually tesed

### Questions: 

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/search/LuceneSearchTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/search/LuceneSearchTest.java
@@ -174,7 +174,7 @@ public class LuceneSearchTest {
     assertThat(results).isNotEmpty();
   }
 
-  //@Test
+  @Test
   public void canDeleteNull() throws IOException {
     // give
     // looks like a bug in web UI: it tries to delete a note twice (after it has just been deleted)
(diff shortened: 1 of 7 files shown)


zeppelin-zengine/src/test/java/org/apache/zeppelin/search/LuceneSearchTest.java
```diff
@@ -173,12 +174,12 @@ public void canIndexAndReIndex() throws IOException, InterruptedException {
     assertThat(results).isNotEmpty();
   }
 
-  @Test
+  //@Test
```
Reamer, line 177: A null check in `deleteNoteIndex` should be sufficient to re-enable this test.

zjffdu: Fixed

PR #3997, opened 2020-12-21
[ZEPPELIN-5162]. Zeppelin reload note from storage doesn't work

zjffdu:
### What is this PR for?

The root cause is that we didn't reload the notes from NotebookRepo. It is a straightforwrad fix with only 1 line code change.

### What type of PR is it?

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5162

* Manually tested.

### Questions:


Reamer:
@zjffdu 
The second commit in this PR should be separated into another PR.

zjffdu:
Oops, my fault, let me fix it.

zjffdu:
Will merge will if no more comment

PR #4003, opened 2020-12-26
[ZEPPELIN-5174]. NPE in SparkSqlInterpreter

zjffdu:
### What is this PR for?

The exception in SparkSqlInterpreter may not have cause, so this PR is to fix the NPE via checking whether the cause is null.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5174

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
@@ -115,7 +115,7 @@ public class SparkSqlInterpreter extends AbstractInterpreter {
           context.out.flush();
           return new InterpreterResult(Code.ERROR);
         } else {
-          LOGGER.error("Error happens in sql: " + curSql, e);
+          LOGGER.error("Error happens in sql: {}", curSql, e);
           context.out.write("\nError happens in sql: " + curSql + "\n");
           if (Boolean.parseBoolean(getProperty("zeppelin.spark.sql.stacktrace", "false"))) {
             if (e.getCause() != null) {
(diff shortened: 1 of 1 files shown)


zjffdu:
Will merge if no more comment

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
```diff
@@ -115,14 +115,23 @@ public InterpreterResult internalInterpret(String st, InterpreterContext context
           context.out.flush();
           return new InterpreterResult(Code.ERROR);
         } else {
+          LOGGER.error("Error happens in sql: " + curSql, e);
```
Reamer, line 118: Please use the parameterized method.

Reamer (changes requested):
Only one small thing.
Everything else LGTM.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
zjffdu: The last parameter is Exception, I think we can not use parameterized method here.

Reamer: You are right.

Reamer (approved):

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkSqlInterpreter.java
Reamer: Hi @zjffdu,
I was a little bit confused. It is possible to use parameterised methods for exceptions as well.
Checkout: http://www.slf4j.org/faq.html#paramException

zjffdu: You are right, it is supported, I verified it locally, learned new things :)

Reamer: I also learned a lot from the Zeppelin project.

zjffdu:
Will merge if no more comment

PR #4004, opened 2020-12-26
[ZEPPELIN-5175]. Should not set driver-class-path when it is yarn cluster mode

zjffdu:
### What is this PR for?

For yarn cluster mode, driver runs in node manager whether there's no zeppelin jars. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5175

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

Reamer:
LGTM

PR #4005, opened 2020-12-26
[ZEPPELIN-5169] Hive set statement doesn't work for some hive version when there's empty line ahead

zjffdu:
### What is this PR for?

The current sql split logic will append empty lines before sql statement, this would cause issues for the sql statement for some hive version. This is to fix it by checking whether it is set statement, if yes, then trim the sql statement.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5169

* Manually tested

### Questions:


prabhjyotsingh (approved):
LGTM

PR #4008, opened 2020-12-28
[ZEPPELIN-5173] Fix the entity name according to the convention

comma337:
### What is this PR for?
JDBC interpreter cannot get the username and password from the credential page.
The reason is that the org.apache.zeppelin.jdbc.JDBCInterpreter#getEntityName method returns "hive" instead of "jdbc.hive".
According to the convention of the entity([interpreter group].[Interpreter name]), fixed the function to return "jdbc.hive".

### What type of PR is it?
Hot Fix

### Todos
* [ ] - Task

### What is the Jira issue?
*  https://issues.apache.org/jira/browse/ZEPPELIN-5173

* Run unit tests.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -100,6 +100,7 @@ import org.apache.zeppelin.user.UsernamePassword;
 public class JDBCInterpreter extends KerberosInterpreter {
   private static final Logger LOGGER = LoggerFactory.getLogger(JDBCInterpreter.class);
 
+  static final String INTERPRETER_NAME = "jdbc";
   static final String COMMON_KEY = "common";
   static final String MAX_LINE_KEY = "max_count";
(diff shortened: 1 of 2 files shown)


zjffdu:
@comma337 I think using `hive` as entity name is more proper,  `jdbc.hive` is a design issue in my opinion. When you create an interpreter `hive`, then `hive` is the new interpreter group name, not `jdbc`.

comma337:
@zjffdu Thanks for the quick review.
I agree with you, `hive` is easier to use than `jdbc.hive`.
I understand that you intend to change convention.

Still documentation and web ui are using [interpreter group].[Interpreter name].
https://zeppelin.apache.org/docs/0.9.0-preview2/setup/security/datasource_authorization.html#how-to-save-the-credential-information

I will close this PR and make another PR for documentation and web UI.
appreciate your help.

zjffdu:
Thanks @comma337 , looking forward your next PR

comma337:
@zjffdu I created a PR for doc/web ui. Thank you.
https://github.com/apache/zeppelin/pull/4009

PR #4010, opened 2021-01-01
[ZEPPELIN-5185]. HiveMonitor thread is never finished sometimes

zjffdu:
### What is this PR for?

Sometimes, hive monitor thread will never exit when getting the following exception (see screenshot) 
image

This PR would exit the monitor thread when any exception happens and also introduce property `zeppelin.jdbc.hive.monitor.query_interval` to control the query interval. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5185

* CI pass

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/hive/HiveUtils.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/hive/HiveUtils.java
@@ -71,11 +71,14 @@ public class HiveUtils {
     final ProgressBar progressBar = progressBarTemp;
     final long timeoutThreshold = Long.parseLong(
             jdbcInterpreter.getProperty("zeppelin.jdbc.hive.timeout.threshold", "" + 60 * 1000));
+    final long queryInterval = Long.parseLong(jdbcInterpreter.getProperty("zeppelin.jdbc.hive.monitor.query_interval",
+            DEFAULT_QUERY_PROGRESS_INTERVAL + ""));
     Thread thread = new Thread(() -> {
       boolean jobLaunched = false;
       long jobLastActiveTime = System.currentTimeMillis();
       while (hiveStmt.hasMoreLogs() && !Thread.interrupted()) {
         try {
+          Thread.sleep(queryInterval);
           List<String> logs = hiveStmt.getQueryLog();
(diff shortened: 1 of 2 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/hive/HiveUtils.java
```diff
@@ -71,11 +71,14 @@ public static void startHiveMonitorThread(Statement stmt,
...
     final long timeoutThreshold = Long.parseLong(
             jdbcInterpreter.getProperty("zeppelin.jdbc.hive.timeout.threshold", "" + 60 * 1000));
+    final long queryInterval = Long.parseLong(jdbcInterpreter.getProperty("zeppelin.jdbc.hive.monitor.query_interval",
+            DEFAULT_QUERY_PROGRESS_INTERVAL + ""));
     Thread thread = new Thread(() -> {
       boolean jobLaunched = false;
       long jobLastActiveTime = System.currentTimeMillis();
       while (hiveStmt.hasMoreLogs() && !Thread.interrupted()) {
         try {
+          Thread.sleep(queryInterval);
```
Reamer, line 81: Maybe we should catch the Interruptexception in a separate 'catch' to interrupt the current thread and to log a better message.

zjffdu: Fixed

Reamer: For good practice we should restore the interrupted state.
Take a look at this rule:
https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2142

Maybe it makes sense to swap the ’while’ and ’try’ to reduce the amount of ’break’.

zjffdu: Good idea, fixed

Reamer: `hiveStmt.isClosed())` should be part of the while condition.

Reamer:
LGTM

PR #4029, opened 2021-01-19
[ZEPPELIN-5211] Fix authType in JDBCInterpreter

zenozeng:
### What is this PR for?

Bug Fix: get `authType` using `getProperty` instead of `jdbcUserConfigurations.getPropertyMap(dbPrefix).getProperty`.

According to the JDBC Interpreter document (https://github.com/apache/zeppelin/blob/master/docs/interpreter/jdbc.md), we should use `zeppelin.jdbc.auth.type`, but It is observed in the JDBC Interpreter that:
- Setting `zeppelin.jdbc.auth.type`  to `KERBEROS` didn't work
- Setting `default.zeppelin.jdbc.auth.type`  to `KERBEROS` works

In `protected boolean isKerboseEnabled()`, we use `getProperty("zeppelin.jdbc.auth.type")`, but in `public Connection getConnection(String dbPrefix, InterpreterContext context)`, `jdbcUserConfigurations.getPropertyMap(dbPrefix).getProperty("zeppelin.jdbc.auth.type", "SIMPLE")` was used. 

### What type of PR is it?
Bug Fix

### Todos
* [x] - Make jdbcIntegrationTest happy

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5211

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -513,7 +513,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
     String url = properties.getProperty(URL_KEY);
     String connectionUrl = appendProxyUserToURL(url, user, dbPrefix);
 
-    String authType = properties.getProperty("zeppelin.jdbc.auth.type", "SIMPLE")
+    String authType = getProperty("zeppelin.jdbc.auth.type", "SIMPLE")
             .trim().toUpperCase();
     switch (authType) {
       case "SIMPLE":
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -513,7 +513,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
     String url = properties.getProperty(URL_KEY);
     String connectionUrl = appendProxyUserToURL(url, user, dbPrefix);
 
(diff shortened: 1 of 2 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -513,7 +513,7 @@ public Connection getConnection(String dbPrefix, InterpreterContext context)
     String url = properties.getProperty(URL_KEY);
     String connectionUrl = appendProxyUserToURL(url, user, dbPrefix);
 
-    String authType = properties.getProperty("zeppelin.jdbc.auth.type", "SIMPLE")
+    String authType = StringUtils.defaultIfEmpty(getProperty("zeppelin.jdbc.auth.type"), "SIMPLE")
```
zjffdu, line 516: Is `StringUtils.defaultIfEmpty` necessary ? getProperty would get default value if it is not set.

zenozeng: Without `StringUtils.defaultIfEmpty`, jdbcIntegrationTest will fail, see also:
https://github.com/apache/zeppelin/runs/1728251503

If `zeppelin.jdbc.auth.type` is empty string (`""`), `properties.getProperty("zeppelin.jdbc.auth.type", "SIMPLE")` will return `""`.
But `defaultIfEmpty(T str, T defaultStr)` returns the value of defaultStr the CharSequence is empty or null.

zenozeng: @zjffdu DONE

zjffdu:
LGTM，thanks @zenozeng

PR #4031, opened 2021-01-22
[ZEPPELIN-5215]. Use user defined exception instead of TException to propagate exception to zeppelin server

zjffdu:
### What is this PR for?

Currently we use TException in RemoteInterpreterServer which make zeppelin server unknown what kind of error happens in interpreter process side. Using user defined exception can propagate the exception to zeppelin server side. 
See https://livebook.manning.com/book/programmers-guide-to-apache-thrift/chapter-9/87

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5215

* Manually tested

Before 
image

After

image

### Questions:


Reamer:
I think we should use custom exceptions.

> User code operates at a higher layer and doesn’t generate TApplicationExceptions, rather making use of user-defined exceptions, which we cover next. TApplicationExceptions involve problems such as calling a method that isn’t implemented or failing to provide the necessary arguments to a method.

> Like TApplicationExceptions, user-defined exceptions are derived from TException. The distinction is that TApplicationExceptions are raised by the Apache Thrift framework and user-defined exceptions are raised by user code.

Other link for the same book: https://freecontent.manning.com/apache-thrift-handling-exceptions/

zjffdu:
@Reamer That's make sense, I have updated the PR with user defined exception `InterpreterRPCException`

Reamer:
The user defined exception looks much better. Do you have any idea why each function also throws the parent `TException`?

zjffdu:
> The user defined exception looks much better. Do you have any idea why each function also throws the parent `TException`?

That is auto generated by thrift.

zjffdu:
Will merge if no more comment

Reamer:
Please wait, I want to take a look at the rendering.

zjffdu:
No problem, take your time @Reamer

Reamer:
@zjffdu Wo ist der Code für diese Änderung?

Reamer:
I have checked the rendering options and do not see any improvement at the moment.

zjffdu:
Reopen it, it looks like I accidentally clean the previous change. @Reamer Please help review it .

Reamer:
LGTM
I am not sure if it makes sense to distinguish between an exception thrown by the remote interpreter and an exception thrown by the Zeppelin server for the remote interpreter.
What do you think?

zjffdu:
> LGTM
> I am not sure if it makes sense to distinguish between an exception thrown by the remote interpreter and an exception thrown by the Zeppelin server for the remote interpreter.
> What do you think?

Do you mean the 2 exceptions here ?  https://github.com/apache/zeppelin/pull/4031/files#diff-3a436f559ddff9ba7280b2617547f5cc9d759e7f4bc6f04304c7e7fe7e01c061R98

We would treat them differently, if it is due to Interpreter exception, then no retry, otherwise it would retry (because it may due to network issue)

Reamer:
> Do you mean the 2 exceptions here ? https://github.com/apache/zeppelin/pull/4031/files#diff-3a436f559ddff9ba7280b2617547f5cc9d759e7f4bc6f04304c7e7fe7e01c061R98

Yes I thing, that's the right location. My thoughts went in the direction of 3 Exception:
 - a TException for network problems -> retry
 - an "InterpreterRPCException" to indicate exceptions thrown by the interpreter
 - a "ServerRPCException" to indicate exceptions thrown by the server.
 
 At the moment the handling is the same, but it might be helpful in the future to make a difference here.

zjffdu:
How about doing it later ? @Reamer  Personally, I prefer to introduce new things when we have clear purpose, otherwise it may lead over design.

Reamer:
Yes, of course it can be done later. It was just a thought.

zjffdu:
Thanks @Reamer will merge if no more comment

PR #4032, opened 2021-01-22
[ZEPPELIN-5213]. Move spark-submit as a separated interpreter module

zjffdu:
### What is this PR for?

This PR is to move spark-submit from spark module to another separated interpreter module. Because it doesn't make sense to run spark-submit in the same JVM of spark interpreter. It would cause some very confused things. e.g. If running it in yarn mode, the yarn app would never be registered unless SparkContext is created. But if you only `%spark.submit`, SparkContext would never be created.  

### What type of PR is it?
[ Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5213

* Manually tested

### Questions:

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -59,10 +59,10 @@ jobs:
           conda info
       - name: install plugins and interpreter
         run: |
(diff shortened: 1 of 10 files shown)


Reamer:
LGTM

zjffdu:
Will merge if no more comment

PR #4036, opened 2021-01-29
[ZEPPELIN-5159]. Add api to clone note in ZeppelinClient

zjffdu:
### What is this PR for?

Add one simple api for clone note.  Besides that, also fix a bug in cloneNote, we should use deep copy instead of shallow copy.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5159

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4037, opened 2021-01-31
[ZEPPELIN-5218]. Don't exit shell interpreter if there's more logging produced

zjffdu:
### What is this PR for?

This PR change the timeout behavior of shell interpreter. Shell command would only timeout after there're no output produced. If there's continue output, then the shell command would continue to execute.

### What type of PR is it?
[ Improvement | Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5218

* Test is added

### Questions:

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -24,6 +24,7 @@ import org.apache.commons.exec.ExecuteWatchdog;
 import org.apache.commons.exec.PumpStreamHandler;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.interpreter.ZeppelinContext;
+import org.apache.zeppelin.util.ExecutorUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


Reamer:
I think that the `shellOutputCheckExecutor` task will run forever. Where is there a stop mechanism?

zjffdu:
@Reamer I have add the stop in close method

shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -78,6 +106,10 @@ public void close() {
         }
       }
     }
+
+    if (shellOutputCheckExecutor != null) {
+      shellOutputCheckExecutor.shutdownNow();
```
Reamer, line 111: The utility class [`ExecutorUtil`](https://github.com/apache/zeppelin/blob/master/zeppelin-interpreter/src/main/java/org/apache/zeppelin/util/ExecutorUtil.java) provides a way to gently shut down this executor.

zjffdu: Fixed

Reamer:
The changes look good to me. But CI fails because of checkstyle.

zjffdu:
CI is passed, will merge if no more comment

PR #4039, opened 2021-01-31
[ZEPPELIN-5204]. NPE when runAsOne is true in flink interpreter

zjffdu:
### What is this PR for?

This PR is to fix the NPE when `runAsOne` is true in flink interpreter. The root cause is that runAsOne should only take effect when there's at least one insert statement.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5204

* Test is added

### Questions:


zjffdu:
Will merge if no more comment

PR #4040, opened 2021-02-01
[ZEPPELIN-5228]. IPySpark unsupported environment would cause other spark interpreter fail

zjffdu:
### What is this PR for?

The root cause is that if ipyspark interpreter fail to open due to environment issue, it would close SparkInterpreter which close SparkContext. This PR won't do that in ipyspark interpreter's close method. This is fine, because when user click restarting interpreter button, all the sub interpreters will be closed. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5228

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4041, opened 2021-02-02
ZEPPELIN-5231: Livy Interpreter doesn't support Japanese Character - Encoding Issue

prabhjyotsingh:
### What is this PR for?
Livy Interpreter doesn't support Japanese Character - Encoding Issue

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5231

On a paragraph use below to generate data set:
```
%livy.scala
spark.sql("CREATE TABLE  test ( id int, name String, destination String)")
spark.sql("INSERT INTO test VALUES ('1', 'みんく', 'みんく')")
spark.sql("INSERT INTO test VALUES ('2', 'シュレヤン', 'シュレヤン')")
spark.sql("INSERT INTO test VALUES ('3', 'サイ', 'サイ')")
spark.sql("INSERT INTO test VALUES ('4', 'ツェッペリン', 'ツェッペリン')")
spark.sql("INSERT INTO test VALUES ('5', 'テスト', 'テスト')")
spark.sql("INSERT INTO test VALUES ('6', 'チャラン', 'チャラン')")
spark.sql("select * from test").show()
```

Then in a separate paragraph use this to verify:
```
%livy.sql
select * from test
```

You would notice that on toggling (true/false) between the newly introduced `tableWithUTFCharacter` the same 2nd paragraph works.

N/A

### Questions:
* Does the licenses files need update?N/A
* Is there breaking changes for older versions?N/A
* Does this needs documentation?N/A

### DIFF
--- a/docs/interpreter/livy.md
+++ b/docs/interpreter/livy.md
@@ -175,6 +175,11 @@ Example: `spark.driver.memory` to `livy.spark.driver.memory`
     <td>key_1: value_1; key_2: value_2</td>
     <td>custom http headers when calling livy rest api. Each http header is separated by `;`, and each header is one key value pair where key value is separated by `:`</td>
   </tr>
+  <tr>
+    <td>zeppelin.livy.tableWithUTFCharacters</td>
+    <td>false</td>
+    <td>If database contains UTF characters then set this as true.</td>
+  </tr>
 </table>
 
 **We remove livy.spark.master in zeppelin-0.7. Because we sugguest user to use livy 0.3 in zeppelin-0.7. And livy 0.3 don't allow to specify livy.spark.master, it enfornce yarn-cluster mode.**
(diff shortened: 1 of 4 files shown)


prabhjyotsingh:
@zjffdu  can you help review this. 

What I have done in this PR is added a new `tableWithUTFCharacter` in the interpreter setting, so that the old mechanism can still work, was not too sure if this JSON conversion will have any performance effect.
Please let me know if there is a better way to do the same.

zjffdu:
@prabhjyotsingh Is it possible to add UT ?

prabhjyotsingh:
@zjffdu  sure that makes sense, have just added the same.

prabhjyotsingh:
@zjffdu can you help review this.

zjffdu:
LGTM @prabhjyotsingh Is the CI fail related ?

prabhjyotsingh:
@zjffdu  Yes, I've looked at the failures it looks unrelated. It fails for `Neo4j interpreter` and one of `test-selenium-with-spark`. 

Thank you for the review, will merge this if no more discussion.

PR #4042, opened 2021-02-03
[ZEPPELIN-5232]. Default value of zeppelin server memory & interpreter memory should be 1024m

zjffdu:
### What is this PR for?

Trivial PR to change the default memory size to be 1024m for both zeppelin server & interpreter. And also remove `-Xms`, because for some interpreters, memory usage is very small, so just remove `-Xms` to save memory.   

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5232

* No test

### Questions:

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -138,11 +138,11 @@ if [[ -z "${ZEPPELIN_ENCODING}" ]]; then
 fi
 
 if [[ -z "${ZEPPELIN_MEM}" ]]; then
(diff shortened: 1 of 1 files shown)


Reamer:
LGTM

zjffdu:
Will merge if no more comment

PR #4043, opened 2021-02-03
[ZEPPELIN-5233]. Clear output before running paragraph

zjffdu:
### What is this PR for?

Trivial PR to clean output before running paragraph. setStatus will broadcast paragraph to frontend. That means the output in frontend will be cleaned before running

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5233

* Ci pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4045, opened 2021-02-04
ZEPPELIN-4952: Markdown interpreter can be used to store XSS in notebooks

prabhjyotsingh:
### What is this PR for?
Markdown interpreter can be used to store XSS in notebooks
The %md interpreter can be used to store XSS in notebooks. These cells are automatically loaded by the user when opening the notebook, so, no manual user interaction is needed.
 
Also, it doesn't matter if the cell has already a result or not.
 
``` 
%md
foo 
<script>alert(String.fromCharCode(88,83,83))</script>
bar
<bold onclick='alert("a");'>qqq</bold>  
```

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4952

* Use the above paragraph, and there should not be any XSS.

### Questions:

### DIFF
--- a/markdown/pom.xml
+++ b/markdown/pom.xml
@@ -37,7 +37,7 @@
     <interpreter.name>md</interpreter.name>
     <markdown4j.version>2.2-cj-1.0</markdown4j.version>
     <pegdown.version>1.6.0</pegdown.version>
-    <flexmark.all.version>0.50.40</flexmark.all.version>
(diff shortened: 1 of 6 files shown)


prabhjyotsingh:
@zjffdu  can you help review this one.

zjffdu:
LGTM, thanks @prabhjyotsingh Let's wait for its CI pass

prabhjyotsingh:

PR #4047, opened 2021-02-05
[ZEPPELIN-5237]. ConcurrentModificationException in Notebook.getNotesInfo

zjffdu:
### What is this PR for?

Trivial PR to resolve the ConcurrentModificationException by using ConcurrentMap instead of Map

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5237

* No test

### Questions:


cuspymd (approved):
Looks good!!. But the `NoteManager::getNotesInfo()` function, which directly returns a reference to the private field `notesInfo`, still looks vulnerable by design.

PR #4053, opened 2021-02-11
[ZEPPELIN-5223] - K8s+docker document format is not correct

omrisk:
Added formatting (to avoid escaping each `_` occurrence, I couldn't find a style guide for documentation or other examples that would suggest another method.
I noticed this also happens in the `docker.md` so I fixed that as well

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,7 +26,7 @@ xcode-select --install
 ```
 
 ## Run website locally
-If you don't want to encounter uglily rendered pages, run the documentation site in your local first.
+If you don't want to encounter ugly rendered pages, run the documentation site in your local environment first.
 
 In `$ZEPPELIN_HOME/docs`,
 
--- a/docs/quickstart/docker.md
+++ b/docs/quickstart/docker.md
@@ -205,5 +205,5 @@ Zeppelin can run locally (such as inside your IDE in debug mode) and able to run
 
 | Configuration variable | Value | Description |
 | ----- | ----- | ----- |
-| ZEPPELIN_RUN_MODE | docker | Make Zeppelin run interpreter on Docker |
-| ZEPPELIN_DOCKER_CONTAINER_IMAGE | <image>:<version> | Zeppelin interpreter docker image to use |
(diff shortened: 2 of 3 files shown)


zjffdu:
Thanks @omrisk for the quick fix, LGTM.
BTW, have you tried to build the docs ? Here's the documentation guide.  https://github.com/apache/zeppelin/blob/master/docs/README.md

omrisk:
@zjffdu , I will try, I'm working on macos-bigsur and ruby is quite broken.
I will update once I get it working (hopefully)

omrisk:
@zjffdu , I was able to get this to work locally, but only after wrapping in docker due to conflicts on my local machine (and with help from @danielRicardo ).

Would you like me to add the needed `Dockerfile` + `docker-compose` and instructions ?
This would allow building the documentation locally without installing any dependencies. 

image
image

zjffdu:
What kind of conflict do you meet ? As I remember it only works with python2.7 and (I use rvm for ruby environment, usually I use ruby 2.3.7)

omrisk:
Mostly `gem install bundler` on Ruby 2.0, you have to be > 2.3 - but then other packages fail there, additionally issue with installing `iff`. 
`An error occurred while installing yajl-ruby (1.1.0), and Bundler cannot continue.
Make sure that `gem install yajl-ruby -v '1.1.0' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  github-pages was resolved to 10, which depends on
    jekyll was resolved to 1.3.0, which depends on
      pygments.rb was resolved to 0.5.4, which depends on
        yajl-ruby
`
I think adding a containerised method may make it easier for future contributors, WDYT?

zjffdu:
Sure, docker approach might be easier for users.

omrisk:
Cool, would you like me to add that here or in a separate PR ?

Reamer:
A separate PR would be better.

omrisk:
@Reamer , thanks opened #4054

Reamer:
LGTM

zjffdu:
LGTM

PR #4054, opened 2021-02-11
[ZEPPELIN-5248] - Add local docker-compose support and updated readme instructions

omrisk:
### What is this PR for?
To make it easier for new contributors to build and test documentation changes, will add support to build and serve locally using docker.

### What type of PR is it?
Improvement 

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5248

* Automated CI
* Can be tested by following updated README instructions to run `docker-compose up` locally 
* 
Screen Shot 2021-02-11 at 15 00 31

### Questions:


Leemoonsoo:
Thank @omrisk for the contribution!

I tested this on

image
(I had to change the 'version' field of 'docker-compose.yml' from '3.8' to '3.3' in this docker version. After the change, worked well without problem)

as well as with the latest Docker engine
image

@omrisk Looks like [rat plugin fails](https://github.com/apache/zeppelin/pull/4054/checks?check_run_id=1879669313) due to missing license header on docker-compose.yml file. Could you try to add a license header? (e.g. https://github.com/apache/zeppelin/blob/master/k8s/zeppelin-server.yaml#L1-L16)

Reamer:
I think docker-compose is unnecessary as we only have one container.
A simple docker command should be enough, no need for docker-compose.

omrisk:
@Reamer , I though the `docker-compose` may be simpler :)
I'll convert to a docker command (also one less file to maintain).

Reamer:
> Just consider that will mean that each launch of this will re-build the docker container (about a minute to start).

I think this is bearable as long as you don't have to restart the container for every change in the docs.

omrisk:
@Reamer , ok. The command I added mounts the `doc` directory to the container so it watches the files for changes (same as if you launch it using the current jekyll implementation)

Reamer:
LGTM

zjffdu:
LGTM, I tested it and it works very well

PR #4055, opened 2021-02-16
[ZEPPELIN-5078] Make it possible to cancel "run all paragraphs" operation

olkkoti:
### What is this PR for?

Make it possible to cancel "run all paragraphs" operation. Currently, Cancel is blocked by UI as well as blocking "run all" operation on server. This PR will unblock cancel from UI and do "run all" operation in separate thread and so avoid blocking the websocket.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5078

Two automatic tests in zeppelin-integration package.
I have problems with CI, but I am able to run them successfully locally.

### DIFF
--- a/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
+++ b/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
@@ -79,6 +79,10 @@ abstract public class AbstractZeppelinIT {
     clickAndWait(by);
   }
 
+  protected void cancelParagraph(int paragraphNo) {
+    By by = By.xpath(getParagraphXPath(paragraphNo) + "//span[@class='icon-control-pause']");
+    clickAndWait(by);
+  }
 
   protected String getParagraphXPath(int paragraphNo) {
     return "(//div[@ng-controller=\"ParagraphCtrl\"])[" + paragraphNo + "]";
--- a/zeppelin-integration/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
+++ b/zeppelin-integration/src/test/java/org/apache/zeppelin/integration/ParagraphActionsIT.java
@@ -276,6 +276,46 @@ public class ParagraphActionsIT extends AbstractZeppelinIT {
     }
   }
 
+  @Test
+  public void testRunAllCancel() throws Exception {
+    try {
+      createNewNote();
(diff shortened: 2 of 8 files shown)


flink-cmd/pom.xml
```diff
@@ -23,14 +23,14 @@
   <parent>
     <artifactId>zeppelin-interpreter-parent</artifactId>
     <groupId>org.apache.zeppelin</groupId>
-    <version>0.10.0-SNAPSHOT</version>
+    <version>0.9.1-SNAPSHOT</version>
```
zjffdu, line 26: This change is not necessary. We use 0.9.1-SNAPSHOT for branch-0.9, and 0.10.0-SNAPSHOT for master

olkkoti: Ah, should parents refer to version 0.10.0 in both flink-cmd and spark-submit? At least I am not able to build that way.

zjffdu: No, all the versions should be 0.9.1-SNAPSHOT, do you mean you unable to build branch-0.9 ?

olkkoti: Yes, I am building on branch-0.9. Should I target this bug fix to master instead?

zjffdu: Yes, usually we create PR based on master branch and will cherry-pick that into branch-0.9 when merging.

olkkoti: Ok, I changed to master.

zjffdu:
@olkkoti There're still some failed selenium tests. Although there're some flaky test in selenium tests, but not sure whether it is related with your PR. Could you help to take a look at that ?

olkkoti:
> @olkkoti There're still some failed selenium tests. Although there're some flaky test in selenium tests, but not sure whether it is related with your PR. Could you help to take a look at that ?

I fixed my cancel tests to have shorter sleep times in paragraphs, and made so that they actually complete even if the tests fail.
The latest run seems to be a lot better, however there is still one error.

zjffdu:
@olkkoti I think this is unrelated, because I see this error in master as well. But it would better to run CI again, because this is flaky test, sometimes fails and sometimes succeed.

zjffdu:
Thanks @olkkoti  CI is passed

zjffdu:
Will merge if no more comment

PR #4057, opened 2021-02-20
[ZEPPELIN-4843] - Fix search in Elasticsearch interpreter

smiklosovic:
@zjffdu would you review please? thank you

### What is this PR for?
This PR fixes issue for search command in Elasticsearch as it is not working now. This is rather a serious problem as right now, ES interpreter is more or less useless when one can not search in it so it would be awesome if it is merged as not doing so renders that plugin useless. 

The problem is that in the past, there was this in the response from a server on search:

```
{
    "took": 1,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": 1
     },
/// other fields omitted for brevity
```

but now it returns:

```
{
    "took": 1,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 1,
            "relation": "eq"
        },
/// other fields omitted for brevity
```
This changed behaviour seems to be introduced in recent ES versions 7.x, I do not have any capability to check how far into the past this issue is present, I have done it in such a way that it is backward-compatible with older versions too.

### What type of PR is it?
Bug Fix.

### What is the Jira issue?
ZEPPELIN-4843

* I have run unit tests in Java for ES interpreter, they all worked.

### Questions:
* Does the licenses files need update? - no
* Is there breaking changes for older versions? - no, the older version is still supported, the solution works as it was but it fallbacks to new way of doing things for more recent ES versions.

### DIFF
--- a/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/client/HttpBasedClient.java
+++ b/elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/client/HttpBasedClient.java
@@ -24,6 +24,7 @@ import com.google.gson.JsonParseException;
 
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONArray;
+import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.UnsupportedEncodingException;
@@ -286,8 +287,7 @@ public class HttpBasedClient implements ElasticsearchClient {
       final JSONObject body = result.getBody() != null ? result.getBody().getObject() : null;
 
       if (isSucceeded(result)) {
-        final long total = getFieldAsLong(result, "hits/total");
-
+        final long total = getTotal(result);
         response = new ActionResponse()
             .succeeded(true)
             .totalHits(total);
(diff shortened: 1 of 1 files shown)


elasticsearch/src/main/java/org/apache/zeppelin/elasticsearch/client/HttpBasedClient.java
```diff
@@ -254,6 +255,15 @@ public ActionResponse index(String index, String type, String id, String data) {
     return response;
   }
 
+  private long getTotal(HttpResponse<JsonNode> result) {
+    final JSONObject hitsObject = result.getBody().getObject().getJSONObject("hits");
+    try {
+      return hitsObject.getLong("total");
+    } catch (final JSONException ex) {
+      return hitsObject.getJSONObject("total").getLong("value");
```
zjffdu, line 263: Could you add one comment to explain the backward compatibility logic here ?

zjffdu:
Thanks @smiklosovic  for the contribution, PR LGTM, only a minor comment

Reamer:
Maybe it's time to update our Elasticsearch library.
We are currently using [2.4.3](https://github.com/apache/zeppelin/blob/master/elasticsearch/pom.xml#L37), which was release in December 2016.

smiklosovic:
@Reamer btw, the  transport client is becoming obsolete these days, if you check e.g. spring data for elasticsearch they are already marking it as obsolete, maybe a ticket which just deletes it wont hurt, I tried transport client in Zeppelin for 7.11.0 and it was throwing exceptions etc ...

smiklosovic:
@zjffdu I have added javadoc and squashed the changes into 1 commit. Please review and eventually merge.

zjffdu:
@smiklosovic CI is failed

smiklosovic:
hi @zjffdu , I have fixed my errors, it was about Javadoc being longer tha 120 chars (something like that). I do not know how to influence other modules. It seems that Zeppelin has just a flaky tests.

smiklosovic:
@zjffdu is there anything else I should do in this matter?

smiklosovic:
@zjffdu is there anything else I should do in this matter?

zjffdu:
LGTM， will merge if no more comment

PR #4058, opened 2021-02-21
[ZEPPELIN-5257] Refactoring of ExecutionContext

zjffdu:
### What is this PR for?

This PR is to refactoring ExecutionContext: move its creation to `Note`, because most of fields of ExecutionContext is from Note. Moving it to `Note` (Note#getExecutionContext) can help us to create consistent object (avoid missing fields) 

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5257

* CI pass

### Questions:


Reamer:
In general, I like the builder pattern for `ExecutionContext` and also final as a keyword for attributes in `ExecutionContext`.
I also think that binding the `ExecutionContext` to a `Note` is a good idea.

I see that you are changing attributes of the `ExecutionContext` such as the user. I think it would be better to create an additional method in the `ExecutionContextBuilder` that takes another `ExecutionContext` and makes a "copy".

zjffdu:
@Reamer  The reason of moving `ExecutionContext` to `Note` is that `Note` is the only place to create `ExecutionContext`, and it makes sense because `ExecutionContext` should always be associated with a `Note`.  So I don't think it is necessary to introduce `ExecutionContextBuilder`.

zjffdu:
Will merge if no more comment

PR #4059, opened 2021-02-21
[ZEPPELIN-5258] Add Knox realm group lookup to shiro auth service

Kimahriman:
### What is this PR for?
This PR aims to add support for Knox groups in ShiroAuthenticationService.getAssociatedRoles. This allows Knox realm groups (which uses hadoop groups under the hood) to be used for things like the Notebook API that doesn't use native Shiro group checks.

Additionally, I removed some unused parts of the Knox realm. I can undo this if desired, but I figured I would clean it up since it's not used.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5258

Added new UT and tested manually

### Questions:


conf/shiro.ini.template
```diff
@@ -66,8 +66,6 @@ user3 = password4, role2
 #knoxJwtRealm.cookieName = hadoop-jwt
 #knoxJwtRealm.publicKeyPath = /etc/zeppelin/conf/knox-sso.pem
 #
-#knoxJwtRealm.groupPrincipalMapping = group.principal.mapping
-#knoxJwtRealm.principalMapping = principal.mapping
```
Kimahriman, line 70: In addition to being unused, this default config throws an error, as these are expected to be strings of the format

`<user1>=<mappeduser1>;<user2>=<mappeduser2>`

zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
```diff
@@ -83,14 +78,6 @@
   @Override
   protected void onInit() {
     super.onInit();
-    if (principalMapping != null && !principalMapping.isEmpty()
-        || groupPrincipalMapping != null && !groupPrincipalMapping.isEmpty()) {
-      try {
-        mapper.loadMappingTable(principalMapping, groupPrincipalMapping);
```
Kimahriman, line 89: `mapper` is never used anywhere else, so I removed everything related to it. Seems like something left over from copying some code from Knox.

zeppelin-server/src/main/java/org/apache/zeppelin/realm/jwt/KnoxJwtRealm.java
```diff
@@ -241,7 +228,6 @@ protected AuthorizationInfo doGetAuthorizationInfo(PrincipalCollection principal
     /* return the groups as seen by Hadoop */
     Set<String> groups = null;
     try {
-      hadoopGroups.refresh();
```
Kimahriman, line 244: Removed this so we could take advantage of hadoop's native group caching mechanisms that should be controllable via hadoop configurations

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -138,8 +139,8 @@ private String extractPrincipal(Subject subject) {
   @Override
   public Collection<Realm> getRealmsList() {
     String key = ThreadContext.SECURITY_MANAGER_KEY;
-    DefaultWebSecurityManager defaultWebSecurityManager = (DefaultWebSecurityManager) ThreadContext.get(key);
-    return defaultWebSecurityManager.getRealms();
+    DefaultSecurityManager defaultSecurityManager = (DefaultSecurityManager) ThreadContext.get(key);
```
Kimahriman, line 142: Using a less specific class made it easier to test (hit a weird issue trying to create a DefaultWebSecurityManager and it seemed unnecessary)

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -247,6 +248,9 @@ public boolean isAuthenticated() {
         } else if (ACTIVE_DIRECTORY_GROUP_REALM.equals(name)) {
           allRoles = ((ActiveDirectoryGroupRealm) realm).getListRoles();
           break;
+        } else if (realm instanceof KnoxJwtRealm) {
```
Kimahriman, line 251: Not sure if the other realms need to do string checks or should be changed to instanceof checks, but had to do it here for the test using the spy

zjffdu:
Thanks for the contribution @Kimahriman Is doc here needs to be updated ? http://zeppelin.apache.org/docs/0.9.0/setup/security/shiro_authentication.html#knox-sso

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
Reamer: You have an advantage when you do a string comparison.
There is a fixed string in the configuration and you can change the underlying implementation independently of the configuration.

Kimahriman:
> Thanks for the contribution @Kimahriman Is doc here needs to be updated ? http://zeppelin.apache.org/docs/0.9.0/setup/security/shiro_authentication.html#knox-sso

Good catch, updated that doc as well

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
Kimahriman: I don't really follow. We have to cast it to get the class specific behavior, so we're kinda relying on the implementation. And it prevents subclassing

Reamer: Forget my previous stupid comment. You are right. I didn't look at all the lines of code.

zjffdu:
Thanks @Kimahriman LGTM

zjffdu:
Will merge if no more comment

PR #4065, opened 2021-03-02
[ZEPPELIN-5272] Row return type FunctionHint doesn't work in flink interpreter

zjffdu:
### What is this PR for?

The root cause is in flink side. And flink 1.11 fix this issue via introducing a new api `createTemporarySystemFunction`. In this PR, I will this method for `flink1.11-shim` and `flink1.12-shim`. Unit test is added as well.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5272

* UT is added

image

### Questions:


zjffdu:
Will merge if no more comment

PR #4071, opened 2021-03-07
[ZEPPELIN-5267]. Use environment variable FLINK_HOME if it is not specified

zjffdu:
### What is this PR for?

When `FLINK_HOME` is not in flink interpreter setting, we should use the current environment variable `FLINK_HOME`. 
2 possible use scenario that use `FLINK_HOME` environment variable:
* Specify `FLINK_HOME` in `zeppelin-env.sh`
* Mount flink distribution and pass `FLINK_HOME` via docker command

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5267

* Manually tested

### Questions:


zjffdu:
Will merge if no more comment

PR #4074, opened 2021-03-10
InterpreterSetting: don't crash when sorting settings.

vprus:
The existing comparator would return -1 for both (a, b) and (b, a),
if both of these settings keys are not present in the template,
and such behaviour violates required comparator properties.
    
The issue manifests with exception in the form
    
      java.lang.IllegalArgumentException: Comparison method violates its general contract!
      at java.util.TimSort.mergeLo(TimSort.java:777)
    
and prevents server from starting up.

### Questions:
* Does the licenses files need update? I think so, the change is trivial.


vprus:
Closing, as it was against wrong branch.

PR #4075, opened 2021-03-10
InterpreterSetting: don't crash when sorting settings.

vprus:
The existing comparator would return -1 for both (a, b) and (b, a),
if both of these settings keys are not present in the template,
and such behaviour violates required comparator properties.

The issue manifests with exception in the form

  java.lang.IllegalArgumentException: Comparison method violates its general contract!
  at java.util.TimSort.mergeLo(TimSort.java:777)

and prevents server from starting up.

### What type of PR is it?
Bug Fix

### Questions:
* Does the licenses files need update? I don't think, the change is trivial.

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -587,19 +587,13 @@ public class InterpreterSetting {
           int i1 = sortedKeys.indexOf(o1);
           int i2 = sortedKeys.indexOf(o2);
           if (i1 != -1 && i2 != -1) {
-            if (i1 < i2) {
-              return -1;
-            } else if (i1 > i2) {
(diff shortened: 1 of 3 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -587,19 +587,13 @@ public void sortPropertiesByTemplate(Object propertiesInTemplate) {
...
+            // If both are present in the template, use natural order of indexes
+            return (i1 - i2);
           } else {
-            if (i1 == -1) {
-              return 1;
-            } else {
-              return -1;
-            }
+            // If one, or both are not in the template, use reverse order, so that missing
+	    // elements are placedare at the end
```
zjffdu, line 594: One minor indentation issue.

zjffdu:
LGTM Thanks @vprus

PR #4076, opened 2021-03-11
[ZEPPELIN-5284]. savepoint & checkpoint don't work in flink 1.12

zjffdu:
### What is this PR for?

The root cause of this issue is flink 1.12 use copied version of Configuration, so that the previous method of updating configuration doesn't' work for flink 1.12. In this PR, I get the configuration via reflection and update it directly. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5284

* Manually tested

### Questions:


zjffdu:
Will merge if no more comment

PR #4078, opened 2021-03-12
[ZEPPELIN-5286]. Unable to run some tutorial notes in zeppelin docker container

zjffdu:
### What is this PR for?

2 notes needs to be updated:
* Python Tutorial/2. IPython Visualization Tutorial_2F1S9ZY8Z.zpln . Add the following 2 lines to ignore warning message. Otherwise there will be many warning message in frontend.
```
import logging
logging.getLogger("params").setLevel(logging.ERROR)
```
* Python Tutorial/3. Keras Binary Classification (IMDB)_2F2AVWJ77.zpln.  Add one paragraph to install keras & tensorflow, because we didn't install in it in docker container
* Specify number & holoviews version, otherwise ipython visualization won't work
* Add files log4j2.properties & log4j2_docker.properties, otherwise flink interpreter log will be missing because flink use log4j2

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5286

* Manually tested

### Questions:


scripts/docker/zeppelin/bin/env_python_3_with_R.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - jupyter_client=5
   - bokeh=1.3.4
   - panel
-  - holoviews
+  - holoviz=0.11.6
```
Reamer, line 20: Very interesting package. But maybe we should give users the possibility to build the Docker image with the latest version of holoviz without changing the version in the Dockerfile.
Since holoviz is a meta package, it might make sense to remove the [dependencies](https://github.com/holoviz/holoviz/blob/master/setup.cfg#L33-L50). e.g. numpy, bokeh, matplotlib
A small comment for this particular dependency should also be placed.

zjffdu: @Reamer  sorry for late response, it seems adding holoviz sometimes doesn't work, I have removed it and add holoviews

zjffdu:
Will merge if no more comment

scripts/docker/zeppelin/bin/env_python_3_with_R.yml
```diff
@@ -1,33 +1,29 @@
-name: base
+name: python_3_with_R
 channels:
   - conda-forge
   - defaults
+  - pyviz
```
Reamer, line 5: Why is this channel needed?

zjffdu: This is used for installing hvplot (see https://pypi.org/project/hvplot/)

Reamer: hvplot is also available via [conda-forge](https://anaconda.org/conda-forge/hvplot)

zjffdu: oh, thanks @Reamer Let me delete it

PR #4080, opened 2021-03-26
[ZEPPELIN-5300] Paragraph pending until timeout when process launcher has already failed

zjffdu:
### What is this PR for?

This PR is to exist the waitForReady method earlier when process launcher is failed. Currently it would only exit when launch timeout.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5300

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
@@ -149,6 +149,8 @@ public class ExecRemoteInterpreterProcess extends RemoteInterpreterManagedProces
 
   private class InterpreterProcessLauncher extends ProcessLauncher {
 
+    private Thread waitForThread;
+
     public InterpreterProcessLauncher(CommandLine commandLine, Map<String, String> envs) {
       super(commandLine, envs);
     }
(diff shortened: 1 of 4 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
```diff
@@ -175,6 +177,7 @@ public void waitForShutdown(int timeout) {
 
     @Override
     public void waitForReady(int timeout) {
+      this.waitForThread = Thread.currentThread();
       synchronized (this) {
```
cuspymd, line 181: Wouldn't it be better to be inside the synchronized block?
And is it okay to not assign `this.waitForThread` to null again?

zjffdu: Fixed

Reamer:
testConnectExistOnlineIntpProcess(org.apache.zeppelin.interpreter.launcher.ClusterInterpreterLauncherTest)`
fails twice in CI.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
```diff
@@ -248,6 +252,11 @@ public void onProcessFailed(ExecuteException e) {
...
         notifyAll();
```
Reamer, line 253: This `notifyAll()` should wake up the `wait(timetoTimeout);`.
Any idea why it doesn't work? Maybe you can explain your change a little more.

zjffdu:
@Reamer Sorry for response late, the root cause is in the yarn-cluster mode, the launcher process may transition to COMPLETED state.  Now I fixed it in this PR

Reamer (approved):
LGTM

PR #4081, opened 2021-03-26
[ZEPPELIN-5299] Comment at end of query causes query to be ignored

zjffdu:
### What is this PR for?

This is to fix the corner case that when the comment is at the end of query, the query will be skipped due to bug in SqlSplitter.
This PR fix the bug in SqlSplitter and also add UT

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5299

* UT is added

### Questions: 


cuspymd (approved):
Looks good!!

zjffdu:
Will merge if no more comment

PR #4083, opened 2021-04-02
[ZEPPELIN-5307] Ignore the single quote and double quote in sql comment

zjffdu:
### What is this PR for?

This PR is to fix the bug of SqlSplitter when single/double quote in sql comment. UT is added. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5307

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
@@ -102,7 +102,7 @@ public class SqlSplitter {
         multiLineComment = false;
       }
 
-      if (character == '\'') {
+      if (character == '\'' && (!singleLineComment && !multiLineComment)) {
         if (singleQuoteString) {
           singleQuoteString = false;
         } else if (!doubleQuoteString) {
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/util/SqlSplitter.java
```diff
@@ -102,15 +102,15 @@ public SqlSplitter(String... additionalSingleCommentPrefixList) {
         multiLineComment = false;
       }
 
-      if (character == '\'') {
+      if (character == '\'' && (!singleLineComment && !multiLineComment)) {
```
cuspymd, line 105: As shown below, moving "!" out of parentheses will make it easier to understand the conditions.
```
if (character == '\'' && !(singleLineComment || multiLineComment)) {
```

zjffdu: Addressed

cuspymd (approved):
Looks good!!

zjffdu:

PR #4088, opened 2021-04-05
[ZEPPELIN-5310] SPARK_HOME in zeppelin-env.sh doesn't take effect in yarn cluster mode

zjffdu:
### What is this PR for?

The root cause is that we didn't get SPARK_HOME from env first when detecting scala version. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5310

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -124,7 +124,7 @@ public class SparkInterpreterLauncher extends StandardInterpreterLauncher {
           }
         }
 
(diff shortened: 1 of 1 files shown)


cuspymd (approved):
Looks good!! But it seems that the handling of environment variables and properties in the program is not consistent.

zjffdu:
@cuspymd Could you point where it inconsistent ? Any place we take environment variable first ?

cuspymd:
> 
> 
> @cuspymd Could you point where it inconsistent ? Any place we take environment variable first ?

Oh, I'm not saying because there is an error in the code. You don't need to care about it, because I don't have a good understanding of the code yet, so it may feel that way.

zjffdu:
No problem  @cuspymd Thanks for your time on the code review. Very appreciated !

zjffdu:
Will merge if no more comment

PR #4093, opened 2021-04-13
[ZEPPELIN-55322] Add Feature 'delete paragraph' for ZeppelinClient

Bowen0729:
### What is this PR for?
Add Feature 'delete paragraph' for ZeppelinClient

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5322

CI
unit test

### Questions:

### DIFF
--- a/zeppelin-client/src/main/java/org/apache/zeppelin/client/ZeppelinClient.java
+++ b/zeppelin-client/src/main/java/org/apache/zeppelin/client/ZeppelinClient.java
@@ -560,6 +560,24 @@ public class ZeppelinClient {
   }
 
   /**
+   * delete paragraph with noteId and paragraphID
+   *
+   * @param noteId
+   * @param paragraphId
+   * @throws Exception
+   */
+  public void deleteParagraph(String noteId, String paragraphId) throws Exception {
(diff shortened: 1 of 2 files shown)


zeppelin-client/src/main/java/org/apache/zeppelin/client/ZeppelinClient.java
```diff
@@ -559,6 +559,24 @@ public String addParagraph(String noteId, String title, String text) throws Exce
     return jsonNode.getObject().getString("body");
   }
 
+  /**
+   * delete paragraph with noteId and paragraphID
+   *
+   * @param noteId
+   * @param paragraphId
+   * @throws Exception
```
cuspymd, line 567: It would be nice if there were more helpful descriptions. Of course you've referenced the existing code, but isn't there a need to copy the existing bad parts?

Bowen0729:
> It would be nice if there were more helpful descriptions. Of course you've referenced the existing code, but isn't there a need to copy the existing bad parts?

Sure, you're right.
And I found other method to solve my problem, this pr is not that necessary.

Bowen0729:
@zjffdu should I close this pr or merge it after review?

zjffdu:
@Bowen0729 If this PR is not needed, you can close it an its jira ticket as well.

PR #4096, opened 2021-04-18
[HOTFIX] CI failed due to dns reverse lookup issue

zjffdu:
### What is this PR for?

This is a hotfix to fix the ci failure due to dns reverse lookup issue. Some of the code is copied from pulsar project.
https://github.com/lhotari/pulsar/commit/b4300db307f542d2ece7ad9f487f078d00210a27

### What type of PR is it?
[ Hot Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* 

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkShinyInterpreterTest.java
+++ b/spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkShinyInterpreterTest.java
@@ -102,7 +102,7 @@ public class SparkShinyInterpreterTest extends ShinyInterpreterTest {
     Thread.sleep(5 * 1000);
     // extract shiny url
     List<InterpreterResultMessage> resultMessages = context2.out.toInterpreterResultMessage();
-    assertEquals(resultMessages + "", 1, resultMessages.size());
+    assertEquals(resultMessages.toString(), 1, resultMessages.size());
     assertEquals(InterpreterResult.Type.HTML, resultMessages.get(0).getType());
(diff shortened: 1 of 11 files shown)


spark/interpreter/src/test/java/org/apache/zeppelin/spark/SparkShinyInterpreterTest.java
```diff
@@ -102,7 +102,7 @@ public void testSparkShinyApp() throws IOException, InterpreterException, Interr
     Thread.sleep(5 * 1000);
     // extract shiny url
     List<InterpreterResultMessage> resultMessages = context2.out.toInterpreterResultMessage();
-    assertEquals(1, resultMessages.size());
+    assertEquals(resultMessages + "", 1, resultMessages.size());
```
cuspymd, line 105: The result seems be the same as `resultMessages.toString()`, but wouldn't it be better to use a consistent expression with other code?

zjffdu: This change is only for diagnose to display the content of resultMessages when this assertion fails.

zjffdu: > The result seems be the same as `resultMessages.toString()`, but wouldn't it be better to use a consistent expression with other code?

Right, let me update it

zjffdu:
Will merge it soon so that it won't block other PRs

PR #4097, opened 2021-04-19
[ZEPPELIN-5330]. Support conda env for python interpreter in yarn mode

zjffdu:
### What is this PR for?

This is to run python interpreter in yarn mode and you can also customize the python runtime via conda. Check the `python.md` for more details

### What type of PR is it?
[Feature | Documentation]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5330

* CI pass

### Questions:


zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -242,13 +244,15 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
...
+    LOGGER.info("Use staging directory: {}", this.stagingDir);
     Map<String, LocalResource> localResources = new HashMap<>();
 
     File interpreterZip = createInterpreterZip();
     Path srcPath = localFs.makeQualified(new Path(interpreterZip.toURI()));
     Path destPath = copyFileToRemote(stagingDir, srcPath, (short) 1);
     addResource(fs, destPath, localResources, LocalResourceType.ARCHIVE, "zeppelin");
-    FileUtils.forceDelete(interpreterZip);
+    LOGGER.info("Add zeppelin archive: {}", destPath);
+    //FileUtils.forceDelete(interpreterZip);
```
cuspymd, line 255: Does it mean that commenting without erasing is a temporary modification?

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -266,6 +270,30 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
         addResource(fs, destPath, localResources, LocalResourceType.ARCHIVE, "hive_conf");
       }
     }
+
+    String yarnDistArchives = launchContext.getProperties().getProperty("zeppelin.yarn.dist.archives");
+    if (org.apache.commons.lang3.StringUtils.isNotBlank(yarnDistArchives)) {
```
cuspymd, line 275: In the file, `org.apache.commons.lang3.StringUtils` is used 7 times, and `org.apache.hadoop.util.StringUtils` is used 1 time. I think it would be okay to use `org.apache.commons.lang3.StringUtils` in short form.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -316,6 +344,25 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
     return amContainer;
   }
 
+  public URI resolveURI(String path) {
```
cuspymd, line 347: public?

cuspymd (commented):
Is it okay without a test?

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: Right, I have uncommented it now.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: change it to private now

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: Good point, fixed it now

Reamer:
I think this is also a great feature for the K8s launcher. Unfortunately, I don't see any way to manipulate files before launching, like you did with YARN. Any ideas how we can make it more general so that it works with other launchers?

zjffdu:
@Reamer Regarding your experience, would building conda env into docker image works for users ? 
User can use a base docker image with lots of common used python packages for their python interpreter. If they want to use other packages, they can use `!pip install` in zeppelin to install packages

Reamer:
> Regarding your experience, would building conda env into docker image works for users ?

It works, but it's not as flexible as your approach with YARN. At the moment it's just an image with a Conda environment with a fixed set of Python libraries and a fixed Python version.
Updates of the image with a changed Python version and libraries are not really possible because you don't know which other notebooks might be broken afterwards.

The same problem exists in pyspark, so I have been thinking for quite a long time how we can activate a similar functionality there.
I found this blog post which is very interesting.
https://databricks.com/de/blog/2020/12/22/how-to-manage-python-dependencies-in-pyspark.html

zjffdu:
@Reamer Actually the approach here also apply for pyspark interpreter, I have verified that

zjffdu:
> Updates of the image with a changed Python version and libraries are not really possible because you don't know which other notebooks might be broken afterwards.

I mean allowing user to choose different image for their notes. Updating image won't be a frequent operation IIUC

Reamer:
> I mean allowing user to choose different image for their notes. Customizing image won't be a frequent operation IIUC

Maintaining different Zeppelin interpreter images is quite difficult. I would like to see the user (download and) activate a specific conda environment in his note.
The main advantage is that the image of the Zeppelin interpreter would be much smaller.

zjffdu:
> > I mean allowing user to choose different image for their notes. Customizing image won't be a frequent operation IIUC
> 
> Maintaining different Zeppelin interpreter images is quite difficult. I would like to see the user (download and) activate a specific conda environment in his note.
> The main advantage is that the image of the Zeppelin interpreter would be much smaller.

But these python interpreter images can share them same base image and add another layer with different python environment. Although the image size may look a little big, the physical size or downloading size would be the same of using the current python interpreter image and then download the python environment via whatever approach.

Reamer:
> Although the image size may look a little big, the physical size or downloading size would be the same of using the current python interpreter image and then download the python environment via whatever approach.

You are right, the final disk size would be the same. It is even possible that it is lower, because if the same Python interpreter is started twice, they both use the same container layer.
But the image approach has big disadvantages in my eyes.
1) If we but the python environment (conda environment) into the image, then the user must select the image. I don't currently know of any way to check whether the image is present, which makes error handling in K8s very difficult. I think the error handling inside the zeppelin interpreter is much easier.
2) For pyspark you need to create two images. A Zeppelin interpreter image (with the Python environment, Spark and the Zeppelin interpreter) and a Spark executor image (with the Python environment and Spark).

Do you know if it is possible to set a URL for `spark.archives`? If not, then there is no other way to put the specific Python environment into the image.

zjffdu:
Thanks @Reamer `spark.archives` seems to be able to work both in yarn and k8s, let me try this.

zjffdu:
@Reamer `spark.archives` works, but it is only available after spark 3.1, and I think it is better to put conda env in cloud storage and then specify it via `spark.archives` (because we spark-submit runs in pod so we can not specify local file as `spark.archives`. Do you think whether it work for users ?

And I think this PR is orthogonal to `spark.archives`, because `spark.archives` will control both the python environment of driver and executor. The executor is out of control of zeppelin.  That's why this approach here only works for python interpreter which zeppelin can control its python environment. But of course we should make the configuration between spark and python interpreter as similar as possible.

Reamer:
You are right, putting the Conda environment in a cloud storage will be the best option. Do you have any idea what possibilities for integration `spark.archives` supports? Local mounting via filesystem is not an option in Kubernetes. I am hoping for an HTTP endpoint, which is very flexible and should work for most users.
YARN should also be fine with an HTTP endpoint, so that the Conda environment can be dynamically loaded by the Zeppelin interpreter when the Python or PySpark interpreter is started.

zjffdu:
For spark interpreter, we can leverage `spark.archives` to download and setup conda environment in both driver(spark interpreter) and executor. But for python interpreter, I don't think there's unified approach for that for now. Buy we can introduce unified configuration for that. e.g. We can introduce `python.archive` which will be translated to yarn/k8s specific configuration.

Reamer:
You are right, for the pyspark zeppelin interpreter we should use `spark.archives` to enable the python (conda) environment.
For the python zeppelin interpreter we should use a configuration parameter that does almost the same as the spark equilant.

> But for python interpreter, I don't think there's unified approach for that for now. Buy we can introduce unified configuration for that. e.g. We can introduce `python.archive` which will be translated to yarn/k8s specific configuration.

The current approach seems to load the conda environment into the HDFS, which seems to be quite effective as the Zeppelin server and the Zeppelin interpreter process within YARN share the same content.
A common way for docker, K8s and YARN could be a dynamic download starting from the Zeppelin interpreter just before it starts the Python process.
At the moment I don't know if `spark.archives` supports a download via HTTP. I will find out as soon as possible. If this is possible, `python.archive` should also do the download so that you don't have to pack python (conda) environments several times.

Reamer:
> At the moment I don't know if `spark.archives` supports a download via HTTP. I will find out as soon as possible. If this is possible, `python.archive` should also do the download so that you don't have to pack python (conda) environments several times.

I can confirm, that `spark.archives` supports HTTP as a source.
The following Spark command was successfully executed in my K8s cluster.
```
PYSPARK_PYTHON=./environment/bin/python ./bin/spark-submit --deploy-mode cluster --archives https://leandi.my_company.de/nexus/service/local/repositories/de.my_company.it.bes.3rdparty/content/de/my_company/conda/my_company-test/1.0/my_company-test-1.0.tar.gz#environment examples/src/main/python/pi.py
```

zjffdu:
Thanks @Reamer for the investigation, so how about using `python.archives` for specifying conda env archives in both yarn and k8s, in this PR, I will make it work in yarn and we can create another ticket to support k8s. What do you think ?

zjffdu:
@Reamer Actually `YarnRemoteInterpreterProcess` doesn't do the downloading, it just upload the conda to hdfs as yarn app resource, and yarn will download it from hdfs before starting yarn container. 
There're 2 benefits of using yarn app resource:
* Don't need to update conda archives to hdfs, just use the local file system to store the conda env. This would make the development much smooth, user use the local conda env to verify it and then move it to yarn environment in production environment.
* We can leverage yarn's resource cache mechanism. That means the same conda env downloaded by yarn_app_1 can be reused by yarn_app_2. If we download it in `JupyterKernelInterpreter.java`, it may cause network congestion if many python interpreters runs at the same time.

Reamer:
Hi @zjffdu,

> We can leverage yarn's resource cache mechanism. That means the same conda env downloaded by yarn_app_1 can be reused by yarn_app_2. If we download it in JupyterKernelInterpreter.java, it may cause network congestion if many python interpreters runs at the same time.

I was aware of this, but it seems that downloading dependencies several times is the way of `spark.archives`. It is clear that this is not optimal.

> Don't need to update conda archives to hdfs, just use the local file system to store the conda env. This would make the development much smooth, user use the local conda env to verify it in local environment and then move it to yarn environment in production environment.

By local, do you mean the local file system of the Zeppelin server? In my environment, the Zeppelin user does not have access to the local file system of the Zeppelin server. Therefore, I would prefer a remote endpoint that is under the control of the Zeppelin user.
I understand your development approach and it sounds great, but I think this is not suitable for a production environment.

zjffdu:
> I was aware of this, but it seems that downloading dependencies several times is the way of `spark.archives`. It is clear that this is not optimal.

If I read spark code correctly, spark only download `spark.archives` in driver side and distribute to executors via its internal mechanism between driver and executor. 

> By local, do you mean the local file system of the Zeppelin server? In my environment, the Zeppelin user does not have access to the local file system of the Zeppelin server. Therefore, I would prefer a remote endpoint that is under the control of the Zeppelin user.
> I understand your development approach and it sounds great, but I think this is not suitable for a production environment.

Let me clarify it, actually it is not only local file system, it could be any hadoop compatible file system, such as hdfs, s3. 

> Maybe we can support the download in `JupyterKernelInterpreter.java` with an additional property. Then it should not matter whether the files were provided by YARN or the download.

Actually for spark yarn mode, it would still use yarn cache mechanism to distribute archives [1]. Of course, for k8s mode, we can use other property to download the archive in `JupyterKernelInterpreter.java` for pure python interpreter, but for pyspark, it is not necessary, because it is already done by SparkSubmit [2]

* [1] https://github.com/apache/spark/blob/master/resource-managers/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala#L1663
* [2] https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L391

Reamer:
Thank you for the clarification. I have created a [Jira ticket](https://issues.apache.org/jira/browse/ZEPPELIN-5384).

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -316,6 +347,25 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
     return amContainer;
   }
 
+  private URI resolveURI(String path) {
```
Reamer, line 350: It would be good if we could add tests with high coverage for this function, as the process is a bit special

zeppelin-jupyter-interpreter/src/main/java/org/apache/zeppelin/jupyter/JupyterKernelInterpreter.java
```diff
@@ -183,8 +195,43 @@ public String checkKernelPrerequisite(String pythonExec) {
...
+      executor.setStreamHandler(psh);
+      int exitCode = executor.execute(cmd);
+      if (exitCode != 0) {
+        throw new IOException("Fail to activate conda env, " + stdout.toString());
+      } else {
+        LOGGER.info("Activate conda env successfully");
+        this.condaEnv = envName;
+        this.pythonExecutable = envName + "/bin/python";
+      }
+    } catch (Exception e) {
```
Reamer, line 226: This should be changed to specific exceptions. More at [RSPEC-2221](https://rules.sonarsource.com/java/tag/error-handling/RSPEC-2221)

Reamer (changes requested):
junit tests should be added.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: Addressed

zeppelin-jupyter-interpreter/src/main/java/org/apache/zeppelin/jupyter/JupyterKernelInterpreter.java
zjffdu: Addressed

zjffdu:
CI is failed, I am looking at that

zjffdu:
CI is fixed now, ready for review @Reamer

zeppelin-jupyter-interpreter/src/main/java/org/apache/zeppelin/jupyter/JupyterKernelInterpreter.java
```diff
@@ -216,6 +259,23 @@ private void launchJupyterKernel(int kernelPort)
     }
   }
 
+  private File buildBootstrapScriptFile(int kernelPort) throws IOException {
+    StringBuilder builder = new StringBuilder();
+    if (condaEnv != null) {
+      builder.append("source " + condaEnv + "/bin/activate\n");
+    }
+    builder.append(pythonExecutable);
+    builder.append(" " + kernelWorkDir.getAbsolutePath() + "/kernel_server.py");
```
Reamer, line 268: If you use the StringBuilder, you should use the String Builder for the entire string.
```
builder.append(" ").append(kernelWorkDir.getAbsolutePath()).append("/kernel_server.py");
...
```

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnLauncherUtil.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class YarnLauncherUtil {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(YarnLauncherUtil.class);
+
+  public static URI resolveURI(String path) {
```
Reamer, line 31: Some javadoc would be nice.
Please describe why `URI uri = new URI(path);` is not sufficient.

zeppelin-plugins/launcher/yarn/src/test/java/org/apache/zeppelin/interpreter/launcher/YarnLauncherUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+public class YarnLauncherUtilTest {
+
+  @Test
+  public void testURIUtil() {
+    URI uri = YarnLauncherUtil.resolveURI("/tmp/env_1");
+    assertEquals("file", uri.getScheme());
+    assertEquals("/tmp/env_1", uri.getPath());
+    assertNull(uri.getFragment());
+
+    uri = YarnLauncherUtil.resolveURI("/tmp/env_1#env");
```
Reamer, line 36: Thank you for the tests. I think we should add some more tests.
1) a string with a scheme
2) a relative path

Reamer (changes requested):

zeppelin-plugins/launcher/yarn/src/test/java/org/apache/zeppelin/interpreter/launcher/YarnLauncherUtilTest.java
zjffdu: Fixed

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnLauncherUtil.java
zjffdu: @Reamer You are right, `URI uri = new URI(path)` is sufficient, I have updated it.

zjffdu:
@Reamer Do you have any more comment ? Otherwise I will merge it.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -259,13 +266,48 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
...
+    String yarnDistArchives = launchContext.getProperties().getProperty("zeppelin.yarn.dist.archives");
+    if (StringUtils.isNotBlank(yarnDistArchives)) {
+      for (String distArchive : yarnDistArchives.split(",")) {
+        URI distArchiveURI = null;
+        try {
+          distArchiveURI = new URI(distArchive);
+        } catch (URISyntaxException e) {
+          throw new IOException("Invalid uri: " + distArchive, e);
+        }
+        if (distArchiveURI.getScheme() == null || "file".equals(distArchiveURI.getScheme())) {
```
Reamer, line 286: A null check is not required at this point.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
```diff
@@ -259,13 +266,48 @@ private ContainerLaunchContext setUpAMLaunchContext() throws IOException {
...
+          // zeppelin.yarn.dist.archives is local file
+          srcPath = localFs.makeQualified(new Path(distArchiveURI));
+          destPath = copyFileToRemote(stagingDir, srcPath, (short) 1);
+        } else {
+          // zeppelin.yarn.dist.archives is files on any hadoop compatible file system
+          destPath = new Path(removeLink(distArchive));
```
Reamer, line 292: `removeLink` sounds wrong at this point. I think `removeFragment` sounds better.
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
```diff
@@ -67,7 +67,7 @@ public String getProtocol() {
   }
 
   public synchronized void send(String serializeMessage) throws IOException {
-    connection.getRemote().sendStringByFuture(serializeMessage);
+    connection.getRemote().sendString(serializeMessage);
```
Reamer, line 70: This line should not be part of this PR.

Reamer (changes requested):
Just a few minor comments.

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: Fixed

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
zjffdu: Fixed

zeppelin-plugins/launcher/yarn/src/main/java/org/apache/zeppelin/interpreter/launcher/YarnRemoteInterpreterProcess.java
zjffdu: fixed

Reamer:
LGTM

Reamer (approved):
LGTM

PR #4099, opened 2021-04-20
Add configuration of flink.udf.packages and flink.udf.packages.packages

Jiabao-Sun:
### What is this PR for?
Add flink.udf.packages configuration to support udf jars in maven repository.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
[ZEPPELIN-5332]

* Set flink.udf.packages and flink.udf.packages.packages configuration and verify the udf functions is loaded correctly.


docs/interpreter/flink.md
```diff
@@ -155,6 +155,16 @@ You can also add and set other flink properties which are not listed in the tabl
     <td></td>
     <td>Packages (comma separated) that would be searched for the udf defined in `flink.udf.jars`.</td>
   </tr>
+  <tr>
+    <td>flink.udf.packages</td>
+    <td></td>
+    <td>Flink udf packages (comma separated), zeppelin will register udf in this jar automatically for user. These udf packages could be found in the maven central or private repository. The udf name is the class name. e.g. org.apache.flink:flink-udfs-demo0:1.10.0,org.apache.flink:flink-udfs-demo1:1.10.0</td>
+  </tr>
+  <tr>
+    <td>flink.udf.packages.packages</td>
```
cuspymd, line 164: There are several properties with similar names, so it seems difficult to understand the difference between them without checking how they are used in the actual code from the user's point of view.
- flink.udf.jars
- flink.udf.jars.packages
- flink.udf.packages
- flink.udf.packages.packages

Jiabao-Sun: This is true, but the basis for naming is to refer to fink.udf.packages.
It may cause misunderstandings and compatibility issues with the old code, if we change this naming style.

If there is a more suitable name, I hope to get your suggestions.

cuspymd: Maybe I feel difficult because I don't have domain knowledge. For those familiar with flink, if these property names are familiar, it doesn't matter. @zjffdu What do you think?

zjffdu: Yes，the naming is a little confusing. I would suggest to use :
* flink.udf.jars
* flink.udf.packages

And them use `flink.udf.search.packages` to replace `flink.udf.jars.packages` for which packages to search for udf classes. But this may introduce incompatibility.

Jiabao-Sun: Thanks @cuspymd and @zjffdu.
I renamed flink.udf.jars.packages to flink.udf.search.packages in a new commit.
Please help to review it again.

PR #4102, opened 2021-04-27
ZEPPELIN-5342: Fixes infinite loop in zeppelin.sh

asm0dey:
### What is this PR for?
If a user is passing an invalid argument to zeppelin.sh it freezes in an infinite loop. This small change changes this behavior and outputs usage information instead of looping forever in the `while` cycle.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5342

Just call `zeppelin.sh` with any type of unsupported arguments

### DIFF
--- a/bin/zeppelin.sh
+++ b/bin/zeppelin.sh
@@ -44,7 +44,6 @@ fi
 
 function usage() {
     echo "Usage: bin/zeppelin.sh [--config <conf-dir>] [--run <noteId>]"
-    exit 0
 }
 
 POSITIONAL=()
@@ -62,11 +61,14 @@ do
     shift # past argument
     shift # past value
     ;;
-    --help)
+    -h|--help)
         usage
+        exit 0
         ;;
-    -h)
+    *)
+        echo "Unsupported argument."
         usage
+        exit 1
         ;;
   esac
 done


bin/zeppelin.sh
```diff
@@ -68,6 +68,9 @@ do
     -h)
         usage
         ;;
+    *)
+        usage
+        ;;
```
cuspymd, line 73: It could be an unintended mistake by the user, so it would be better if the script exited with a non-zero value.

asm0dey: Agreed.

What do you think about 

```bash
        echo "Unsupported argument. Please, call with -h/--help for usage information"
        exit 1
        ;;
```

cuspymd: It would be nice if the following items could be modified as well. What do you think?
- Combine `--help` and `-h` processing into one
- Separate the `exit` call part from the `usage` function.
- `*)` In the processing, call the `usage` function after `echo "Unsupported argument."`

asm0dey: Done!

cuspymd (approved):
Looks good!!

Reamer:
LGTM

asm0dey:
@Reamer @cuspymd what else should I do for this to be merged?

Reamer:
> @Reamer @cuspymd what else should I do for this to be merged?

We should test your changes by CI. I think your changes won't break anything.

asm0dey:
Perfect, just checking :)

zjffdu:
ping @asm0dey Could rebase it ?

asm0dey:
@zjffdu I would be happy to, but how should I recreate PR on GitHub?

asm0dey:
Oh, orry, my bad, I've updated the PR

PR #4104, opened 2021-05-01
[ZEPPELIN-5338]. Add logging prefix in interpreter.sh for easy diagnose

zjffdu:
### What is this PR for?

Currently, we have to change log4j.properties to get more output from script `interpreter.sh` in log4j output for diagnose. This PR is to add one trick: as long as the output is start with `[INFO]` in `interpreter.sh`, it will be redirected to log4j output.  

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5338

* CI pass

### Questions:

### DIFF
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -310,5 +310,6 @@ else
 fi
 
 # Don't remove this echo, it is for diagnose, this line of output will be redirected to java log4j output.
-echo "Interpreter launch command: ${INTERPRETER_RUN_COMMAND[@]}"
+# Output that starts with `[INFO]` will be redirected to log4j output.
+echo "[INFO] Interpreter launch command: ${INTERPRETER_RUN_COMMAND[@]}"
 exec "${INTERPRETER_RUN_COMMAND[@]}"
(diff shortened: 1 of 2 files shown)


bin/interpreter.sh
```diff
@@ -296,5 +296,6 @@ else
 fi
 
 # Don't remove this echo, it is for diagnose, this line of output will be redirected to java log4j output.
-echo "Interpreter launch command: ${INTERPRETER_RUN_COMMAND[@]}"
+# Output that starts with `[INFO]` will be redirected to log4j output.
```
cuspymd, line 299: Isn't it better to express "log4j INFO output" or "log4j INFO message" more precisely than "log4j output"? Depending on the log level setting, debug messages can also be displayed in the output.

zjffdu: Thanks for the comment @cuspymd It is fixed now.

zjffdu:
@cuspymd Any more comment ?

cuspymd (approved):
Looks good!!

zjffdu:
Will merge if no more comment

PR #4105, opened 2021-05-02
[ZEPPELIN-5337] Spark scope mode is broken

zjffdu:
### What is this PR for?

Spark scope mode is broken because we use different scala shell output dir for each spark scala shell. This PR fix this issue by using a shared output folder for all the scala shell instances.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5337

* UT is added

### Questions:


Leemoonsoo:
Thank @zjffdu for the fix!

cuspymd (approved):
Looks good!! 
But are there any side effects that can occur by sharing the output folder in scope mode?

zjffdu:
@cuspymd There's no side effect based on my current testing.

zjffdu:
Will merge if no more comment

PR #4111, opened 2021-05-08
[ZEPPELIN-5356] Support flink 1.13

zjffdu:
### What is this PR for?

Add new module `flink-1.13-shims` to support flink 1.3. The main change is due to the api change of flink 1.3, we have to use different `TableConfig` for each `TableEnvironment` instances, otherwise it will cause conflicts.

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5356

* Ci pass and manually tested 1.13.0

### Questions:


flink/interpreter/src/main/java/org/apache/zeppelin/flink/TableEnvFactory.java
```diff
@@ -55,30 +55,50 @@
   private Executor executor;
   private org.apache.flink.api.scala.ExecutionEnvironment benv;
   private org.apache.flink.streaming.api.scala.StreamExecutionEnvironment senv;
-  private TableConfig tblConfig;
+  private TableConfig sTableConfig;
+  private TableConfig bTableConfig;
```
cuspymd, line 59: It will be easy to understand if there is an explanation about the difference between "stable" and "btable".

zjffdu:
Will merge if no more comment

PR #4113, opened 2021-05-11
[ZEPPELIN-5343] Zeppelin Interface becomes unresponsive

zjffdu:
### What is this PR for?

This is experimental PR for fixing the zeppelin unresponsive issue. Regarding the jstack, it seems blocked in `NotebookSocket#send` method, this PR change it from sync call to async call. 

### What type of PR is it?
[Bug Fix ] 

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5343

* CI 

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookSocket.java
@@ -67,7 +67,7 @@ public class NotebookSocket extends WebSocketAdapter {
   }
 
   public synchronized void send(String serializeMessage) throws IOException {
-    connection.getRemote().sendString(serializeMessage);
+    connection.getRemote().sendStringByFuture(serializeMessage);
   }
 
   public String getUser() {


cuspymd:
Are you sure this change fix a client hang? Blocking of `send` might be the result of a client hang not the cause of it.

zjffdu:
@cuspymd  I am not sure whether this is the right fix, as I mentioned in the PR, it is an experimental PR. As it is hard to reproduce this issue, it is not easy to verify the fix. Do you have any idea on what might be the root cause ?

cuspymd:
> @cuspymd I am not sure whether this is the right fix, as I mentioned in the PR, it is an experimental PR. As it is hard to reproduce this issue, it is not easy to verify the fix. Do you have any idea on what might be the root cause ?

Sorry, I haven't seen this issue yet.

zjffdu:
Will merge if no more comment

Reamer:
Does it make a difference in your environment?

chillinger:
@zjffdu based on your MR we built the zeppelin-server jar for our environment and will test it - I will share the results as soon as we have solid feedback.

One question to the org.apache.zeppelin.socket.NotebookSocket.send() method: we could not find any reason for the "synchronized" marking of the method - is there a reason for this?

zjffdu:
Thank @chillinger for the help. Regarding `synchronized`, it is done in https://issues.apache.org/jira/browse/ZEPPELIN-1480.

chillinger:
@zjffdu I can verify that the proposed Fix completely eliminated the deadlocks of zeppelin experienced before.

We built a patched version of Zeppelin 0.9 with the change in this MR and deployed in on all stages of our environment, 100+ Users on PROD. With vanilla 0.9 we had about one restart per day. Since the deployment of the patched version not a single restart.

+1 for merging

zjffdu:
Thanks @chillinger for the verification, will merge it soon

PR #4089, opened 2021-04-08
[ZEPPELIN-5249]. Update to thrift 0.14.2

PrarthiJain:
### What is this PR for?
• This PR is to upgrade thrift to 0.14.2

### What type of PR is it?
• [Improvement]

### Todos
• [ ] - Task

### What is the Jira issue?
• https://issues.apache.org/jira/browse/ZEPPELIN-5249

### How should this be tested?
• CI pass

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #4116, opened 2021-05-13
[ZEPPELIN-5352] Support flink in k8s mode

zjffdu:
### What is this PR for?

Flink support k8s native mode from flink 1.11, this PR is to support k8s-application mode. The implementation is very similar as yarn-application mode.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5352

* Ci pass and manually tested

### Questions:

### DIFF
--- a/flink/interpreter/pom.xml
+++ b/flink/interpreter/pom.xml
@@ -42,6 +42,7 @@
     <hive.version>2.3.4</hive.version>
     <hiverunner.version>4.0.0</hiverunner.version>
     <grpc.version>1.15.0</grpc.version>
+    <kubernetes.client.version>5.3.1</kubernetes.client.version>
 
     <scala.macros.version>2.0.1</scala.macros.version>
     <scala.binary.version>2.11</scala.binary.version>
@@ -93,6 +94,13 @@
       <groupId>org.apache.zeppelin</groupId>
       <artifactId>flink1.13-shims</artifactId>
       <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+      <version>${kubernetes.client.version}</version>
+      <scope>compile</scope>
     </dependency>
 
     <dependency>
--- a/flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
+++ b/flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
@@ -146,8 +146,8 @@ class FlinkScalaInterpreter(val properties: Properties) {
     // load udf jar
     this.userUdfJars.foreach(jar => loadUDFJar(jar))
 
-    if (mode == ExecutionMode.YARN_APPLICATION) {
-      // have to call senv.execute method before running any user code, otherwise yarn application mode
+    if (ExecutionMode.isApplicationMode(mode)) {
+      // have to call senv.execute method before running any user code, otherwise application mode
       // will cause ClassNotFound issue. Needs to do more investigation. TODO(zjffdu)
       val initCode =
         """
@@ -182,16 +182,27 @@ class FlinkScalaInterpreter(val properties: Properties) {
       properties.getProperty("flink.execution.mode", "LOCAL")
         .replace("-", "_")
         .toUpperCase)
-    if (mode == ExecutionMode.YARN_APPLICATION) {
+
+    if (ExecutionMode.isYarnAppicationMode(mode)) {
       if (flinkVersion.isFlink110) {
-        throw new Exception("yarn-application mode is only supported after Flink 1.11")
+        throw new Exception("application mode is only supported after Flink 1.11")
       }
       // use current yarn container working directory as FLINK_HOME, FLINK_CONF_DIR and HIVE_CONF_DIR
       val workingDirectory = new File(".").getAbsolutePath
       flinkHome = workingDirectory
       flinkConfDir = workingDirectory
       hiveConfDir = workingDirectory
+    } else if (ExecutionMode.isK8sApplicationMode(mode)) {
+      if (flinkVersion.isFlink110) {
+        throw new Exception("application mode is only supported after Flink 1.11")
+      }
+      // use current k8s working directory as FLINK_HOME
+      val workingDirectory = new File(".").getAbsolutePath
+      flinkHome = workingDirectory
+      flinkConfDir = workingDirectory + "/conf"
+      hiveConfDir = workingDirectory + "/conf"
     }
+
     LOGGER.info("FLINK_HOME: " + flinkHome)
     LOGGER.info("FLINK_CONF_DIR: " + flinkConfDir)
     LOGGER.info("HADOOP_CONF_DIR: " + hadoopConfDir)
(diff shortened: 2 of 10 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
```diff
@@ -142,6 +141,7 @@
   private ScheduledExecutorService resultCleanService = Executors.newSingleThreadScheduledExecutor();
 
   private boolean isTest;
+  private boolean isFlinkK8sApplicationMode = false;
```
cuspymd, line 144: `RemoteInterpreterServer` is a high-level object that handles `Interpreter`s identically. It seems a little odd that this object needs to know the operation mode of a particular interpreter.

flink/interpreter/pom.xml
```diff
@@ -95,6 +96,13 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+      <version>${kubernetes.client.version}</version>
+      <scope>compile</scope>
```
Reamer, line 103: Where have you used this dependency?

zjffdu: Right，I didn't use it yet, Just plan to add ingress support in next PR. Let me remove it in this PR.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
zjffdu: That's right, but I think this is the only way for flink k8s mode. For flink k8s mode, the `RemoteInterpreterServer#main` is the not the actual main entry point, `RemoteInterpreterServer#main` will be called by flink framework via reflection. So if we call `System.exit()` method in flink's k8s mode, it would make flink framework think that the flink job is exited abnormally, and will restart it again which is not what we expect. Let me add some comment to explain it in this class

flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
```diff
@@ -312,9 +323,12 @@ class FlinkScalaInterpreter(val properties: Properties) {
...
           if (mode == ExecutionMode.YARN_APPLICATION) {
             val yarnAppId = System.getenv("_APP_ID");
-            LOGGER.info("Use FlinkCluster in yarn application mode, appId: {}", yarnAppId)
+            LOGGER.info("Use FlinkCluster in yarn-application mode, appId: {}", yarnAppId)
             this.jmWebUrl = "http://localhost:" + HadoopUtils.getFlinkRestPort(yarnAppId)
             this.displayedJMWebUrl = HadoopUtils.getYarnAppTrackingUrl(yarnAppId)
+          } else if (mode == ExecutionMode.KUBERNETES_APPLICATION) {
+            LOGGER.info("Use FlinkCluster in kubernetes-application mode")
+            this.jmWebUrl = "http://localhost:8083"
           } else {
```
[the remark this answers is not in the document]
zjffdu: I split it into 2 methods: createFlinkILoop & initFlinkILoop

zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
```diff
@@ -45,25 +48,44 @@ public FlinkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
...
     }
     envs.put("FLINK_LIB_DIR", flinkHome + "/lib");
     envs.put("FLINK_PLUGINS_DIR", flinkHome + "/plugins");
 
-    // yarn application mode specific logic
-    if ("yarn-application".equalsIgnoreCase(
-            context.getProperties().getProperty("flink.execution.mode"))) {
-      updateEnvsForYarnApplicationMode(envs, context);
+    String mode = context.getProperties().getProperty("flink.execution.mode");
+    String a = FLINK_EXECUTION_MODES.stream().collect(Collectors.joining(", "));
```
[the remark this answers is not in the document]
zjffdu: Good catch, let me remove it

zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
```diff
@@ -45,25 +48,44 @@ public FlinkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
...
     }
 
     return envs;
   }
 
-  private String updateEnvsForFlinkHome(Map<String, String> envs,
-                                      InterpreterLaunchContext context) throws IOException {
+  private void verifyExecutionMode(String mode) {
+
+  }
```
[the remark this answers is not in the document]
zjffdu: removed

zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
```diff
@@ -45,25 +48,44 @@ public FlinkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
...
+
+  private boolean isApplicationMode(String mode) {
+    return "yarn-application".equals(mode) || "kubernetes-application".equals(mode);
```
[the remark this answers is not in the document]
zjffdu: Fixed

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
Reamer: I agree with @cuspymd, changing `RemoteInterpreterServer` for a specific interpreter is not good.
In general, calling `System.exit()` is not good style, so we should wait for #4089 and help @PrarthiJain if necessary. This PR removes a `System.exit(0)`.

Maybe we are also able to remove the second [`System.exit(1)`](https://github.com/apache/zeppelin/blob/97ce203a7fb72fd0300ff84ea5fa8e52f6125ebc/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java#L705)

If we cannot remove the `System.exit()`, then we should work with an interpreter-independent variable that should be set via the start command or evaluated at start-up.

flink/interpreter/pom.xml
Reamer: I also have the requirement to add additional ingress k8s resources during the runtime of the zeppelin interpreter. See [ZEPPELIN-5145](https://issues.apache.org/jira/browse/ZEPPELIN-5145).

In my opinion, the interpreter pod should **not** have the right to create ingress resources, but the Zeppelin server service user should have the right.

So my plan is to add an additional interface to the Zeppelin interpreter project. Something like `ExternalURLSupported`.
This interface should be implemented by the `InterpreterProcess` e.g. `K8sRemoteInterpreterProcess.java` on the ZeppelinServer side and any specific interpreter can query the function via `instanceof`.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/remote/RemoteInterpreterServer.java
zjffdu: No problem, let's wait for  #4089

flink/interpreter/src/main/scala/org/apache/zeppelin/flink/FlinkScalaInterpreter.scala
```diff
@@ -182,16 +182,27 @@ class FlinkScalaInterpreter(val properties: Properties) {
...
       flinkHome = workingDirectory
       flinkConfDir = workingDirectory
       hiveConfDir = workingDirectory
+    } else if (ExecutionMode.isK8sApplicationMode(mode)) {
+      if (flinkVersion.isFlink110) {
+        throw new Exception("application mode is only supported after Flink 1.11")
+      }
+      // use current k8s working directory as FLINK_HOME
+      val workingDirectory = new File(".").getAbsolutePath
+      flinkHome = workingDirectory
```
cuspymd, line 201: The repeated codes might be moved outside the if statement(186 line).

zjffdu:
> LGTM, but keep in mind that I don't use Flink at all.

Although we don't have integration test, I have tested it manually.

PR #4118, opened 2021-05-14
[ZEPPELIN-5370] Polish pom.xml files

cuspymd:
### What is this PR for?
Polish pom.xml files
- Delete unnecessary tags
- Delete duplicated dependencies 

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5370

* CI

### Questions:

### DIFF
--- a/zeppelin-plugins/pom.xml
+++ b/zeppelin-plugins/pom.xml
@@ -29,7 +29,6 @@
 
     <artifactId>zengine-plugins-parent</artifactId>
     <packaging>pom</packaging>
-    <version>0.10.0-SNAPSHOT</version>
     <name>Zeppelin: Plugins Parent</name>
     <description>Zeppelin Plugins Parent</description>
     <properties>
--- a/zeppelin-plugins/pom.xml
+++ b/zeppelin-plugins/pom.xml
@@ -25,10 +25,8 @@
         <artifactId>zeppelin</artifactId>
         <groupId>org.apache.zeppelin</groupId>
         <version>0.10.0-SNAPSHOT</version>
-        <relativePath>..</relativePath>
     </parent>
 
-    <groupId>org.apache.zeppelin</groupId>
     <artifactId>zengine-plugins-parent</artifactId>
     <packaging>pom</packaging>
     <version>0.10.0-SNAPSHOT</version>
--- a/alluxio/pom.xml
+++ b/alluxio/pom.xml
@@ -29,7 +29,6 @@
 
     <artifactId>zeppelin-alluxio</artifactId>
     <packaging>jar</packaging>
-    <version>0.10.0-SNAPSHOT</version>
     <name>Zeppelin: Alluxio interpreter</name>
 
     <properties>
--- a/alluxio/pom.xml
+++ b/alluxio/pom.xml
@@ -27,7 +27,6 @@
         <relativePath>../zeppelin-interpreter-parent/pom.xml</relativePath>
     </parent>
 
-    <groupId>org.apache.zeppelin</groupId>
     <artifactId>zeppelin-alluxio</artifactId>
     <packaging>jar</packaging>
     <version>0.10.0-SNAPSHOT</version>
@@ -52,12 +51,6 @@
         </dependency>
 
         <!-- TEST -->
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
-
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
--- a/angular/pom.xml
+++ b/angular/pom.xml
@@ -27,7 +27,6 @@
     <relativePath>../zeppelin-interpreter-parent/pom.xml</relativePath>
   </parent>
 
-  <groupId>org.apache.zeppelin</groupId>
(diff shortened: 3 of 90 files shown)


zjffdu:
Seems 2 CI fails @cuspymd

zeppelin-plugins/pom.xml
```diff
@@ -25,10 +25,8 @@
         <artifactId>zeppelin</artifactId>
         <groupId>org.apache.zeppelin</groupId>
         <version>0.10.0-SNAPSHOT</version>
-        <relativePath>..</relativePath>
```
Reamer, line 28: I think `relativePath` should still be used at this point, but it should point to the pom.xml.
```
<relativePath>../pom.xml</relativePath>
```
I know that the parent directory is the default and can be omitted, but sometimes you don't run Maven over the whole project, like in our [tests](https://github.com/apache/zeppelin/blob/97ce203a7fb72fd0300ff84ea5fa8e52f6125ebc/.github/workflows/core.yml#L67), then we should reference the parent project with a relative path.

> However, that would work if the parent project was already installed in our local repository or was in that specific directory structure (parent pom.xml is one directory higher than that of the module's pom.xml).
https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance

zeppelin-plugins/pom.xml
```diff
@@ -25,10 +25,8 @@
...
     </parent>
 
-    <groupId>org.apache.zeppelin</groupId>
```
Reamer, line 31: When you clean up the pom files, you should also delete the versions.

>Alternatively, if you want the groupId or the version of your modules to be the same as their parents, you can remove the groupId or the version identity of your module in its POM.
https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance

zeppelin-plugins/pom.xml
cuspymd: The default value of the tag `relativePath` is "../pom.xml". Nevertheless, do you think it should be declared explicitly?
> The relative path of the parent <code>pom.xml</code> file within the check out. If not specified, it defaults to <code>../pom.xml</code>.
> https://maven.apache.org/xsd/maven-4.0.0.xsd

zeppelin-plugins/pom.xml
cuspymd: Upon checking, only the version value of "r/pom.xml" is "0.9.0-SNAPSHOT". Can I understand that it is missing from the version update changes?

zeppelin-plugins/pom.xml
Reamer: You are right, I changed my mind because of the good xsd.
Defaults should not be declared, so you should remove this.

zeppelin-plugins/pom.xml
Reamer: The r module is deprecated.
https://github.com/apache/zeppelin/commit/b62c6078d1d6dee30f9e35e10bf7e6d0f1e1ffc5

Reamer: Maybe it's time to remove the module.
https://github.com/Reamer/zeppelin/commit/1298e2619d168647806c09a913a69a5bc8883c8b
@zjffdu what are your thinking?

cuspymd: I have deleted duplicated version tags of sub modules

Reamer (commented):
`zeppelin-interpreter-shaded` is a fixed dependency in [zeppelin-interpreter-parent](https://github.com/apache/zeppelin/blob/df69b18fb2d915b4268304ae769478efd88b619a/zeppelin-interpreter-parent/pom.xml#L37-L41).

EDIT: Forget this comment. I was not able to read the commit history....

Reamer:
@cuspymd Please rebase to current master to fix CI.

cuspymd:
> @cuspymd Please rebase to current master to fix CI.

I have done.

Reamer:
@cuspymd Please rebase to current master to fix CI.

zjffdu:
@cuspymd Could you check the failed spark integration test ?

zjffdu:
ping @cuspymd

zeppelin-plugins/pom.xml
Reamer: PR #4145 deletes the R module.

Reamer: @cuspymd I removed the deprecated r module, please rebase your PR.

cuspymd:
> 
> 
> ping @cuspymd

I fixed it and rebased codes

Reamer:
I am very interested in this PR. @cuspymd Thank you for your work.
Should we also backport this to branch-0.9?

zjffdu:
ping @cuspymd

cuspymd:
> 
> 
> I am very interested in this PR. @cuspymd Thank you for your work.
> Should we also backport this to branch-0.9?

Are there any benefits to backporting it to branch-0.9?
And, at the time of rebase, there were conflicts with the modifications of @zjffdu, 
so it was difficult to merge. Backporting only these fixes would not be easy.

Reamer:
You are right, I have merged this into Master.

PR #4120, opened 2021-05-17
[ZEPPELIN-5374] Don't update paragraph config when latest checkpoint of flink is unchanged

zjffdu:
### What is this PR for?

Simple PR to check whether the current latest checkpoint is changed, only update paragraph config when it is changed. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5374

* CI pass

### Questions:

### DIFF
--- a/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
+++ b/flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
@@ -188,6 +188,7 @@ public class JobManager {
     private AtomicBoolean running = new AtomicBoolean(true);
     private boolean isFirstPoll = true;
     private long checkInterval;
+    private String latestCheckpointPath;
 
     FlinkJobProgressPoller(String flinkWebUrl, JobID jobId, InterpreterContext context, long checkInterval) {
       this.flinkWebUrl = flinkWebUrl;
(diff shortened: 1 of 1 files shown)


flink/interpreter/src/main/java/org/apache/zeppelin/flink/JobManager.java
```diff
@@ -254,11 +255,12 @@ public void run() {
               if (completedObject.has("external_path")) {
                 String checkpointPath = completedObject.getString("external_path");
                 LOGGER.debug("Latest checkpoint path: {}", checkpointPath);
-                if (!StringUtils.isBlank(checkpointPath)) {
+                if (!StringUtils.isBlank(checkpointPath) && !checkpointPath.equals(latestCheckpointPath)) {
```
cuspymd, line 258: It seems that using the operator "not (!)" only once is good for readability.
```suggestion
                if (!(StringUtils.isBlank(checkpointPath) || checkpointPath.equals(latestCheckpointPath))) {
```

zjffdu: Actually here 2 not is more readable. The if condition is when checkpoint is not null and it is not the same as latestCheckpointPath. 
If I change it to 1 not, it is a little harder to understand what the if condition means

cuspymd (approved):
Looks good!!

zjffdu:
Will merge if no more comment

PR #4121, opened 2021-05-18
[ZEPPELIN-5376] Polish server and service package codes of zeppelin-server

cuspymd:
### What is this PR for?
Polish server and service package codes of zeppelin-server
- Delete unused import
- Delete and optimize redundant codes
- Add `final` keyword to some variables that have unchanged reference.

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
* [ZEPPELIN-5376] Polish server and service package codes of zeppelin-server

* CI

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
@@ -41,9 +41,9 @@ public class JobManagerService {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(JobManagerService.class);
 
-  private Notebook notebook;
-  private AuthorizationService authorizationService;
-  private ZeppelinConfiguration conf;
+  private final Notebook notebook;
+  private final AuthorizationService authorizationService;
+  private final ZeppelinConfiguration conf;
 
   @Inject
   public JobManagerService(Notebook notebook,
@@ -65,9 +65,10 @@ public class JobManagerService {
     Note jobNote = notebook.getNote(noteId);
     if (jobNote == null) {
       callback.onFailure(new IOException("Note " + noteId + " not found"), context);
+    } else {
+      notesJobInfo.add(new NoteJobInfo(jobNote));
+      callback.onSuccess(notesJobInfo, context);
     }
-    notesJobInfo.add(new NoteJobInfo(jobNote));
-    callback.onSuccess(notesJobInfo, context);
     return notesJobInfo;
   }
 
@@ -84,7 +85,7 @@ public class JobManagerService {
 
     List<NoteJobInfo> notesJobInfo = notebook.getNoteStream()
             .filter(note -> authorizationService.isOwner(context.getUserAndRoles(), note.getId()))
-            .map(note -> new NoteJobInfo(note))
+            .map(NoteJobInfo::new)
             .filter(noteJobInfo -> noteJobInfo.unixTimeLastRun > lastUpdateServerUnixTime)
             .collect(Collectors.toList());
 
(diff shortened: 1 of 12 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/NoAuthenticationService.java
```diff
@@ -30,7 +30,7 @@
 import org.slf4j.LoggerFactory;
 
 public class NoAuthenticationService implements AuthenticationService {
-  private static Logger logger = LoggerFactory.getLogger(NoAuthenticationService.class);
+  private static final Logger logger = LoggerFactory.getLogger(NoAuthenticationService.class);
```
zjffdu, line 33: Rename it to be `LOGGER`, this is our convention for static final variables.

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -407,8 +407,8 @@ public boolean isAuthenticated() {
     Connection con = null;
     PreparedStatement ps = null;
     ResultSet rs = null;
-    DataSource dataSource = null;
-    String authQuery = "";
+    DataSource dataSource;
+    String authQuery;
```
zjffdu, line 411: Only change these 2 variables ?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NoAuthenticationService.java
cuspymd: In the example in the [Constant names](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) section of the Google java style document, it says that `Logger` is not a constant. Could you check it out again?
> // Not constants
...
static final Logger logger = Logger.getLogger(MyClass.getName());

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
cuspymd: In the case of `con`, `ps`, and `rs` variables, they are used in the `finally` block, so if not initialized, a compilation error occurs.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NoAuthenticationService.java
zjffdu: I think the logger in google java style document refer to `java.util.logging.Logger` which is a class and do has method to change its state (e.g. Logger#config). But what zeppelin use is `org.slf4j.Logger` is an interface which has no method to change its state.

zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
```diff
@@ -65,9 +65,10 @@ public JobManagerService(Notebook notebook,
     Note jobNote = notebook.getNote(noteId);
     if (jobNote == null) {
       callback.onFailure(new IOException("Note " + noteId + " not found"), context);
+    } else {
+      notesJobInfo.add(new NoteJobInfo(jobNote));
+      callback.onSuccess(notesJobInfo, context);
```
Reamer, line 70: This is a change of programme. But it looks good.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NoAuthenticationService.java
cuspymd: I changed it.

zjffdu:
LGTM

PR #4122, opened 2021-05-20
[ZEPPELIN-5379] LIST_CONFIGURATIONS command cannot return the default…

mnmhouse:
### What is this PR for?

when download distirbution package ,  run zeppelin server ,do not config zeppelin-site.xml ,i found LIST_CONFIGURATIONS command cannot return the default configuration in the ZeppelinConfiguration class code,so the web  lack some configuration 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
*  [ZEPPELIN-5379] https://issues.apache.org/jira/browse/ZEPPELIN-5379

* Ci pass and manually tested

image

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -871,6 +871,10 @@ public class ZeppelinConfiguration {
         completeConfiguration.put(c.getVarName(), sysConfig.getString(c.getVarName()));
       } else if (envConfig.containsKey(c.name())) {
         completeConfiguration.put(c.getVarName(), envConfig.getString(c.name()));
+      }else {
+        if (getString(c)!=null){
+          completeConfiguration.put(c.getVarName(),getString(c));
+        }
       }
     }
     return completeConfiguration;
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -864,17 +864,10 @@ public class ZeppelinConfiguration {
    * @return
    */
   public Map<String, String> getCompleteConfiguration() {
-    Map<String, String> completeConfiguration = new HashMap<>(properties);
-    // Is it possible that we overwrite properties
+    Map<String, String> completeConfiguration = new HashMap<>();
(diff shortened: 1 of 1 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -871,6 +871,10 @@ public boolean isPrometheusMetricEnabled() {
         completeConfiguration.put(c.getVarName(), sysConfig.getString(c.getVarName()));
       } else if (envConfig.containsKey(c.name())) {
         completeConfiguration.put(c.getVarName(), envConfig.getString(c.name()));
+      }else {
+        if (getString(c)!=null){
+          completeConfiguration.put(c.getVarName(),getString(c));
+        }
```
cuspymd, line 877: Since `envConfig` and `sysConfig` are checked in `getString()`, would it be okay to delete lines 870~873?

However, it seems that the check order of `envConfig` and `sysConfig` is different only in `getCompleteConfiguration()`.
@zjffdu @Reamer  Which of `envConfig` and `sysConfig` takes precedence?

Reamer: @cuspymd You can find the order in the [current docs](https://github.com/apache/zeppelin/blob/master/docs/setup/operation/configuration.md).
Comment by @zjffdu: https://github.com/apache/zeppelin/pull/4028#discussion_r560194922

I noticed a small typo. PR #4123 should fix this.

Reamer: > Since `envConfig` and `sysConfig` are checked in `getString()`, would it be okay to delete lines 870~873?

You should be right. I think we can also start with an empty Hashmap.
```java
  public Map<String, String> getCompleteConfiguration() {
    Map<String, String> completeConfiguration = new HashMap<>();
    for (ConfVars c : ConfVars.values()) {
        if (getString(c) != null){
            completeConfiguration.put(c.getVarName(), getString(c));
        }
    }
    return completeConfiguration;
  }
```

mnmhouse: @cuspymd   You look  very carefully，i did not notice  the getString() method  
@Reamer  thank you ,I will modify according to the code  you provided

cuspymd (approved):
Looks good!!

zjffdu:
LGTM, will merge if no more comment

PR #4128, opened 2021-05-30
[ZEPPELIN-5339] Support scala 2.12 for flink

zjffdu:
### What is this PR for?

This is to support scala 2.12 of flink interpreter. Here's the main changes:

1. Copy `FlinkILoop`, `FlinkShell`, `ScalaShellEnvironment`, `ScalaShellStreamEnvironment` & `JarHelper` from flink project. The reason is that currently flink scala shell only support scala 2.11, we need to copy it here and do some hack work to support scala 2.12
2. Introduce 2 modules: `flink-scala-2.11`  & `flink-scala-2.12` , each of them will build one interpreter jar, and FlinkInterpreterLauncher will choose the right interpreter jar based on the current scala version of flink.

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5339

* CI pass

### Questions:


Reamer:
I don't like the symlinks and the Maven build helper steps because this construct breaks some IDEs. Is it not possible to create only one submodule with two Scala compiler profiles?

I know that we also have this symlink construct in our Spark submodule. That's how I know it's not good.

Example of Source-Root:
/src/main/scala/...
/src/main/scala_2.11/... (only included with Scala profile 2.11)
/src/main/scala_2.12/... (only included with Scala profile 2.11)

zjffdu:
@Reamer  Using profiles won't work here, because our goal is to make one zeppelin instance to work with multiple scala version. Building different zeppelin distribution for each scala version is impractical. Regarding the code structure in IDE, I do have some tips to make it work in intellij, let me add one contribution doc to clarify that.

flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/FlinkInterpreter.java
```diff
@@ -25,45 +26,74 @@
...
-    
-    this.innerIntp = new FlinkScalaInterpreter(getProperties());
-    this.innerIntp.open();
-    this.z = this.innerIntp.getZeppelinContext();
+    try {
+      this.innerIntp = loadFlinkScalaInterpreter();
+      this.innerIntp.open();
+      this.z = this.innerIntp.getZeppelinContext();
+    } catch (Exception e) {
+      throw new InterpreterException("Fail to open FlinkInterpreter", e);
```
cuspymd, line 78: It seems be strange to run again `new InterpreterException()` when `loadFlinkScalaInterpreter()` throws an exception of the type `InterpreterException`.

Reamer:
> Regarding the code structure in IDE, I do have some tips to make it work in intellij, let me add one contribution doc to clarify that

That would be great. Thank you

zjffdu:
@Reamer  `README` is added

zjffdu:
Will merge if no more comment

PR #4130, opened 2021-06-04
[ZEPPELIN-5266]: Enable extension of flexmark in markdown interpreter

EricGao888:
### What is this PR for?
* Enable extension of flexmark in markdown interpreter

### What type of PR is it?
Improvement

### Todos
* [ ] - Make emoji auto-scale according to the size of text 

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5266

* Create a cell of markdown type, set the interpreter to be flexmark and type in an emoji code, e.g. :joy:

image
image

### Questions:
* None


Reamer:
Where did you find the emoji images? I think it's better to download the emojis during the build and not put them all in our git repository.

EricGao888:
> Where did you find the emoji images? I think it's better to download the emojis during the build and not put them all in our git repository.

Good suggestion, may I ask how I can download the emojis to the web assets folder during the build process? BTW, I found a link to https://www.webfx.com/tools/emoji-cheat-sheet/ in flexmark docs where I got those images. : )

zjffdu:
@EricGao888 You can use `download-maven-plugin` to download these images and put them in the zeppelin-web module. 
See the example here https://github.com/apache/zeppelin/blob/master/spark/interpreter/pom.xml#L361

EricGao888:
> Where did you find the emoji images? I think it's better to download the emojis during the build and not put them all in our git repository.

I found that unicode can be used instead of images for emojis. I think it's a better way.

Reamer:
> > Where did you find the emoji images? I think it's better to download the emojis during the build and not put them all in our git repository.
> 
> I found that unicode can be used instead of images for emojis. I think it's a better way.

Unicode should be sufficient. It would be nice if you would update your image.

LGTM

EricGao888:
> > > Where did you find the emoji images? I think it's better to download the emojis during the build and not put them all in our git repository.
> > 
> > 
> > I found that unicode can be used instead of images for emojis. I think it's a better way.
> 
> Unicode should be sufficient. It would be nice if you would update your image.
> 
> LGTM

Sure, will update the images just in case the unicode stuff breaks.

cuspymd (approved):
Looks good!!

zjffdu:
LGTM

zjffdu:
Will merge if no more comment

PR #4135, opened 2021-06-12
[ZEPPELIN-5290] NPE on empty dynamic form input

zjffdu:
### What is this PR for?

Simple PR to address NPE on empty dynamic form input. Instead of throwing NPE, just make the form value as empty string. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5290

* CI pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/display/Input.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/display/Input.java
@@ -373,7 +373,11 @@ public class Input<T> implements Serializable {
         }
       } else {
         // single-selection
-        expanded = value.toString();
+        if (value == null) {
+          expanded = "";
+        } else {
+          expanded = value.toString();
+        }
       }
       replaced = match.replaceFirst(expanded);
       match = pattern.matcher(replaced);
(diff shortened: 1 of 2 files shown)


cuspymd (approved):
Looks good!!

zjffdu:
Will merge if no more comment

zeppelin-interpreter/src/main/java/org/apache/zeppelin/display/Input.java
```diff
@@ -373,7 +373,11 @@ public static String getSimpleQuery(Map<String, Object> params, String script, b
         }
       } else {
         // single-selection
-        expanded = value.toString();
+        if (value == null) {
```
Reamer, line 376: [StringUtils.defaultString](http://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringUtils.html#defaultString%28java.lang.String%29) looks more elegant.

zjffdu: Thanks for the review @Reamer  I have updated it

Reamer:
LGTM

PR #4137, opened 2021-06-13
[ZEPPELIN-5392] Timezone is not applied in the timestamp field in flink interpreter

zjffdu:
### What is this PR for?

Add timezone support for flink interpreter, It is only available for flink 1.13, previous version are not supported.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5392

* Manually tested.

image

### Questions:

### DIFF
--- a/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/sql/AbstractStreamSqlJob.java
+++ b/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/sql/AbstractStreamSqlJob.java
@@ -105,7 +105,8 @@ public abstract class AbstractStreamSqlJob {
 
   public String run(String st) throws IOException {
     this.table = stenv.sqlQuery(st);
-    String tableName = "UnnamedTable_" + SQL_INDEX.getAndIncrement();
+    String tableName = "UnnamedTable_" +
+            "_" + SQL_INDEX.getAndIncrement();
     return run(table, tableName);
   }
 
@@ -201,12 +202,13 @@ public abstract class AbstractStreamSqlJob {
 
   protected abstract String buildResult();
 
-  protected String table2String(List<Row> rows) {
+  protected String tableToString(List<Row> rows) {
     StringBuilder builder = new StringBuilder();
     for (Row row : rows) {
-      String[] fields = flinkShims.row2String(row, table, stenv.getConfig());
-      String rowString =
-              Arrays.stream(fields).map(TableDataUtils::normalizeColumn).collect(Collectors.joining("\t"));
+      String[] fields = flinkShims.rowToString(row, table, stenv.getConfig());
+      String rowString = Arrays.stream(fields)
+              .map(TableDataUtils::normalizeColumn)
+              .collect(Collectors.joining("\t"));
       builder.append(rowString);
       builder.append("\n");
     }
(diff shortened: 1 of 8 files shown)


flink/flink1.10-shims/src/main/java/org/apache/zeppelin/flink/Flink110Shims.java
```diff
@@ -319,4 +319,22 @@ public Map extractTableConfigOptions() {
     }
     return configOptions;
   }
+
+  @Override
+  public String[] row2String(Object row, Object table, Object tableConfig) {
+    return rowToString((Row) row);
+  }
```
cuspymd, line 326: Wouldn't it be better to declare the parameter's type as a more specific type than `Object`?
And it would be better to use either `row2String` or `rowToString` consistently.

zjffdu: Thanks for the review @cuspymd I have rename method name to be rowToString. Regarding the parameter type, it is not possible to use the specific type because the api inconsistent across different flink versions, so I have to use type Object and let each version of `FlinkShims` to do the type casting

flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/sql/AbstractStreamSqlJob.java
```diff
@@ -197,6 +201,18 @@ protected void processRecord(Tuple2<Boolean, Row> change) {
 
   protected abstract String buildResult();
 
+  protected String table2String(List<Row> rows) {
+    StringBuilder builder = new StringBuilder();
+    for (Row row : rows) {
+      String[] fields = flinkShims.row2String(row, table, stenv.getConfig());
```
cuspymd, line 207: `rowToString`?

zjffdu: Fixed

flink/flink1.10-shims/src/main/java/org/apache/zeppelin/flink/Flink110Shims.java
cuspymd: Looking at the current implementation, the `rowToString()` function is only called with specific parameter types in the `Abstract StreamSqlJob::table2String()` function. I don't quite understand how different types can be passed.

zjffdu: @cuspymd The reason is that module `flink-shims` doesn't depend on any specific version of flink api,  the flink dependency is specified in each version of flink-shims, so we can not use flink api in `flink-shims`

flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/sql/AbstractStreamSqlJob.java
```diff
@@ -99,7 +104,7 @@ private static TableSchema removeTimeAttributes(TableSchema schema) {
   protected abstract String getType();
 
   public String run(String st) throws IOException {
-    Table table = stenv.sqlQuery(st);
+    this.table = stenv.sqlQuery(st);
     String tableName = "UnnamedTable_" +
             "_" + SQL_INDEX.getAndIncrement();
     return run(table, tableName);
```
cuspymd, line 110: `this.table` member variable can not be set normally if `run(Table table, String tableName)` method be called directly.
Is it OK in this scenario?

zjffdu: Good point @cuspymd I have set table in this run method as well.

zjffdu:
Will merge if no more comment

PR #4139, opened 2021-06-15
[ZEPPELIN-5412] Unable to cancel flink job due to fail to take savepoint

zjffdu:
### What is this PR for?

Sometimes flink job is unable to be cancelled when savepoint is enabled. This PR would fall back to cancel job without savepoint if cancelling with savepoint is failed. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5412

* Manually tested

### Questions:

### DIFF
--- a/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/JobManager.java
+++ b/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/JobManager.java
@@ -151,7 +151,7 @@ public class JobManager {
                   jobClient.getJobID(), context.getParagraphId(), savePointPath);
         } catch (Exception e) {
           LOGGER.warn("Fail to cancel job of paragraph {} with savepoint, try to cancel it without savepoint",
-                  context.getParagraphId());
+                  context.getParagraphId(), e);
           jobClient.cancel();
         }
       }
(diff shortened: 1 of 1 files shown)


flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/JobManager.java
```diff
@@ -141,13 +141,19 @@ public void cancelJob(InterpreterContext context) throws InterpreterException {
...
+          Map<String, String> config = new HashMap<>();
+          config.put(SAVEPOINT_PATH, savePointPath);
+          context.getIntpEventClient().updateParagraphConfig(
+                  context.getNoteId(), context.getParagraphId(), config);
+          LOGGER.info("Job {} of paragraph {} is stopped with save point path: {}",
+                  jobClient.getJobID(), context.getParagraphId(), savePointPath);
+        } catch (Exception e) {
+          LOGGER.warn("Fail to cancel job of paragraph {} with savepoint, try to cancel it without savepoint",
+                  context.getParagraphId());
+          jobClient.cancel();
```
cuspymd, line 155: It may be good to log the exception for future debugging.

zjffdu: Fixed

cuspymd (approved):
Looks good!!

PR #4140, opened 2021-06-16
[ZEPPELIN-5413] Throw proper error message when user set execution.runtime-mode in flink sql

zjffdu:
### What is this PR for?

`execution.runtime-mode` is available in flink sql-client, but it is not available in zeppelin flink interpreter. So we need to throw proper error message when user set execution.runtime-mode in flink sql. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5413

* Manually tested

### Questions:

### DIFF
--- a/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
+++ b/flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
@@ -502,9 +502,9 @@ public abstract class FlinkSqlInterrpeter extends AbstractInterpreter {
 
   public abstract void callInnerSelect(String sql, InterpreterContext context) throws IOException;
 
-  public void callSet(String key, String value, InterpreterContext context) throws IOException {
+  public void callSet(String key, String value, InterpreterContext context) throws Exception {
     if (key.equals("execution.runtime-mode")) {
-      throw new IOException("execution.runtime-mode is not supported to set, " +
+      throw new UnsupportedOperationException("execution.runtime-mode is not supported to set, " +
               "you can use %flink.ssql & %flink.bsql to switch between streaming mode and batch mode");
     }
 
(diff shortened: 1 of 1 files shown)


flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
```diff
@@ -503,10 +503,16 @@ public void callSelect(String sql, InterpreterContext context) throws IOExceptio
   public abstract void callInnerSelect(String sql, InterpreterContext context) throws IOException;
 
   public void callSet(String key, String value, InterpreterContext context) throws IOException {
+    if (key.equals("execution.runtime-mode")) {
+      throw new IOException("execution.runtime-mode is not supported to set, " +
+              "you can use %flink.ssql & %flink.bsql to switch between streaming mode and batch mode");
+    }
```
cuspymd, line 509: Only `IOException` is available here? `UnsupportedOperationException` seems be more suitable.

zjffdu: That's right, thanks for the review, Fixed it

cuspymd (approved):
Looks good!!

flink/flink-scala-parent/src/main/java/org/apache/zeppelin/flink/FlinkSqlInterrpeter.java
```diff
@@ -502,11 +502,17 @@ public void callSelect(String sql, InterpreterContext context) throws IOExceptio
 
   public abstract void callInnerSelect(String sql, InterpreterContext context) throws IOException;
 
-  public void callSet(String key, String value, InterpreterContext context) throws IOException {
+  public void callSet(String key, String value, InterpreterContext context) throws Exception {
+    if (key.equals("execution.runtime-mode")) {
```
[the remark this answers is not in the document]
zjffdu: Fixed

zjffdu:
Will merge if no more comment

PR #4141, opened 2021-06-16
[ZEPPELIN-5405] ClassNotFoundException in YarnAppMonitor when hadoop client is not installed

zjffdu:
### What is this PR for?

Minor PR to only use YarnAppMonitor when hadoop client is installed, otherwise ClassNotFoundException will be thrown

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5405

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
@@ -100,9 +100,20 @@ public class ExecRemoteInterpreterProcess extends RemoteInterpreterManagedProces
       Matcher m = YARN_APP_PATTER.matcher(launchOutput);
       if (m.find()) {
         String appId = m.group(1);
-        LOGGER.info("Detected yarn app: {}, add it to YarnAppMonitor", appId);
-        YarnAppMonitor.get().addYarnApp(ConverterUtils.toApplicationId(appId), this);
+        if (isHadoopClientAvailable()) {
+          LOGGER.info("Detected yarn app: {}, add it to YarnAppMonitor", appId);
+          YarnAppMonitor.get().addYarnApp(ConverterUtils.toApplicationId(appId), this);
+        }
       }
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/remote/ExecRemoteInterpreterProcess.java
```diff
@@ -100,12 +100,23 @@ public void start(String userName) throws IOException {
       Matcher m = YARN_APP_PATTER.matcher(launchOutput);
       if (m.find()) {
         String appId = m.group(1);
-        LOGGER.info("Detected yarn app: {}, add it to YarnAppMonitor", appId);
-        YarnAppMonitor.get().addYarnApp(ConverterUtils.toApplicationId(appId), this);
+        if (isHadoopClientAvailable()) {
+          LOGGER.info("Detected yarn app: {}, add it to YarnAppMonitor", appId);
+          YarnAppMonitor.get().addYarnApp(ConverterUtils.toApplicationId(appId), this);
+        }
```
cuspymd, line 106: No pattern matching is needed if hadoop client is not available.
It would be good to move this condition to line 98.

cuspymd (approved):
Looks good!!

Reamer:
@zjffdu 
I found this [nice code](https://github.com/apache/shiro/blob/1c97815737fee587729b1d3827814f3f58d2b6ec/lang/src/main/java/org/apache/shiro/lang/util/ClassUtils.java#L149-L190) in apache shiro, maybe we could use this.

Generally, I also plan to use such a function, as we have a dependency on [log4j in the interpreter module](https://github.com/apache/zeppelin/blob/628f064fc4ff1d307c24e7c83758c7b952bc644e/zeppelin-interpreter/pom.xml#L144-L147).

zjffdu:
Thanks @Reamer for the suggestion, but it would be a little weird for zeppelin to depend on shiro's utility class.

zjffdu:
any more comment ? @Reamer

Reamer:
> any more comment ? @Reamer

No. LGTM

PR #4150, opened 2021-06-23
[ZEPPELIN-5425] Polish Interpreter class

cuspymd:
### What is this PR for?
- Move `interpolate()` function into `AbstractInterpreter` class which is only one caller of this function
- Fix IDE warning messages

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5425

* Unit tests and CI

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/AbstractInterpreter.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/AbstractInterpreter.java
@@ -18,10 +18,14 @@
 package org.apache.zeppelin.interpreter;
 
 import org.apache.zeppelin.interpreter.thrift.InterpreterCompletion;
+import org.apache.zeppelin.resource.Resource;
+import org.apache.zeppelin.resource.ResourcePool;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public abstract class AbstractInterpreter extends Interpreter {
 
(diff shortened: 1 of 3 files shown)


Reamer:
I restarted the CI because many tests failed, which does not seem to be related to these changes.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/AbstractInterpreter.java
```diff
@@ -47,6 +51,35 @@ public InterpreterResult interpret(String st,
     return internalInterpret(st, context);
   }
 
+  static String interpolate(String cmd, ResourcePool resourcePool) {
+    Pattern zVariablePattern = Pattern.compile("([^{}]*)([{]+[^{}]*[}]+)(.*)", Pattern.DOTALL);
```
zjffdu, line 55: Move it outside as static variable ?

Reamer: With `static` and `final` :grin: 
https://rules.sonarsource.com/java/tag/performance/RSPEC-4248

cuspymd: I fixed. But No difference was found from the execution time of `ZeppCtxtVariableTest`.
Below is the execution time of 18 tests of `ZeppCtxtVariableTest`. (measured 10 times each)
statistics | before (ms) | after (ms) 
-------|--------|-------
average | 235.6| 234.6
std. dev. | 12.6 | 26.5

zjffdu:
LGTM

PR #4152, opened 2021-06-23
[ZEPPELIN-5426] mongodb interpreter: support mongosh

cnoelle:
### What is this PR for?
The mongodb interpreter works by calling into the deprecated `mongo` shell (https://docs.mongodb.com/manual/reference/program/mongo/), which is going to be replaced by `mongosh` (https://docs.mongodb.com/mongodb-shell/). This is a largely backwards compatible migration, so that the interpreter works almost unchanged. Some internals have changed nevertheless, this pull request ensures the interpreter works with the new shell without breaking backwards compatibility.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5426

### Questions:
* Does this needs documentation? At some point...

### DIFF
--- a/mongodb/src/main/resources/shell_extension.js
+++ b/mongodb/src/main/resources/shell_extension.js
@@ -97,14 +97,15 @@ function printTable(dbquery, fields, flattenArray) {
     });
 }
 
-DBQuery.prototype.table = function (fields, flattenArray) {
+(DBQuery.prototype || DBQuery).table = function (fields, flattenArray) {
     if (this._limit > tableLimit) {
         this.limit(tableLimit);
     }
     printTable(this, fields, flattenArray);
 };
 
-DBCommandCursor.prototype.table = DBQuery.prototype.table;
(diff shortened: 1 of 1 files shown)


cuspymd (approved):
Looks good!!

zjffdu:
LGTM

coachaac:
Does not work with mongosh 1.10.1 from mongodb 6.0.3

MyqueWooMiddo:
> Does not work with mongosh 1.10.1 from mongodb 6.0.3
> 
> i got this error : TypeError: db.users.aggr ... ]).table is not a function
> 
> it seems that the end of shell_extention.js is no more doing is job :
> 
> globalThis.DBCommandCursor seems to be undefined.

I could not access mongo client via Zeppelin , but I can access it via cli

I use mongo rather than mongosh

" E QUERY    [js] ReferenceError: globalThis is not defined :" is the Exception

PR #4154, opened 2021-06-24
[ZEPPELIN-5423] Running %spark.conf cell seem to break placeholder #{user} from config

zjffdu:
### What is this PR for?

The root cause is that we do context parameter replacement in zeppelin server side (RemoteInterpreter). Actually we should do it in interpreter process side. Also add tests for this. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5423

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4155, opened 2021-06-24
[ZEPPELIN-5428] Unify flink configuration between different execution modes

zjffdu:
### What is this PR for?

This PR is to unify the configuration of flink interpreter. Due to historical reason, some properties are only available for yarn mode, but not for yarn-application mode, this PR is to unify them. Here's the properties we handle in this PR.
* flink.tm.memory
* flink.jm.memory
* flink.yarn.appName
* flink.yarn.queue
* flink.tm.slot

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5423

* CI pass
* 
### Questions:

### DIFF
--- a/docs/interpreter/flink.md
+++ b/docs/interpreter/flink.md
@@ -106,17 +106,17 @@ You can also add and set other flink properties which are not listed in the tabl
     <td>Port of running JobManager. Only used for remote mode</td>
   </tr>
   <tr>
-    <td>flink.jm.memory</td>
+    <td>jobmanager.memory.process.size</td>
     <td>1024</td>
-    <td>Total number of memory(mb) of JobManager</td>
+    <td>Total number of memory of JobManager, default unit: MB</td>
   </tr>
   <tr>
-    <td>flink.tm.memory</td>
+    <td>taskmanager.memory.process.size</td>
     <td>1024</td>
-    <td>Total number of memory(mb) of TaskManager</td>
+    <td>Total number of memory of TaskManager, default unit: MB</td>
   </tr>
   <tr>
-    <td>flink.tm.slot</td>
+    <td>taskmanager.numberOfTaskSlots</td>
     <td>1</td>
     <td>Number of slot per TaskManager</td>
   </tr>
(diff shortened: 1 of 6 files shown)


zeppelin-plugins/launcher/flink/src/main/java/org/apache/zeppelin/interpreter/launcher/FlinkInterpreterLauncher.java
```diff
@@ -66,6 +69,30 @@ public FlinkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
...
+                              String newKey,
+                              boolean isMemoryProperty) {
+    String value = properties.getProperty(oldKey);
+    if (StringUtils.isNotBlank(value) && !properties.containsKey(newKey)) {
+      if (isMemoryProperty) {
+        properties.put(newKey, value + "mb");
+      } else {
+        properties.put(newKey, value);
+      }
+    }
```
cuspymd, line 93: If the value of `newKey` exists, is it not necessary to append "mb" to it?

zjffdu: The conversion here only happens when there's no new key exists. 
So the order is newKey > oldKey

cuspymd: > 
> 
> The conversion here only happens when there's no new key exists.
> So the order is newKey > oldKey

I know what you said. But if the user sets the `taskmanager.memory.process.size` value referring to the manual (flink.md line 114~116), there is a high probability that "mb" is not attached to the set value.

In this case, I ask if it is necessary to check the newKey setting value and add "mb".

zjffdu: Oops, you are right, I have updated the doc. It is official [flink property](https://ci.apache.org/projects/flink/flink-docs-release-1.13/docs/deployment/config/), user should not add unit in this property.

docs/interpreter/flink.md
```diff
@@ -106,17 +106,17 @@ You can also add and set other flink properties which are not listed in the tabl
     <td>Port of running JobManager. Only used for remote mode</td>
   </tr>
   <tr>
-    <td>flink.jm.memory</td>
+    <td>jobmanager.memory.process.size</td>
     <td>1024</td>
```
cuspymd, line 110: Shouldn't it be fixed here as well?

zjffdu: Fixed

zjffdu:
Will merge if no more comment

PR #4161, opened 2021-07-05
[ZEPPELIN-5446] SparkInterpreterLauncher#detectSparkScalaVersion may return wrong scala version

zjffdu:
### What is this PR for?

This issue happens when user set `SPARK_HOME` in zeppelin spark interpreter setting, but zeppelin server host has `SPARK_CONF_DIR` which point to another version of spark.  In this case, even when user set `SPARK_CONF_DIR` explicitly, `SparkInterpreterLauncher#detectSparkScalaVersion` would still return the scala version of the spark version of `SPARK_CONF_DIR` point to.

This PR would pass the envs built from interpreter setting to the process of `detectSparkScalaVersion`

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5446

* Manually tested

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -64,7 +64,7 @@ public class SparkInterpreterLauncher extends StandardInterpreterLauncher {
 
   @Override
   public Map<String, String> buildEnvFromProperties(InterpreterLaunchContext context) throws IOException {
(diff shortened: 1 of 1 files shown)


zjffdu:
Will merge if no more comment

Reamer:
In my eyes, an env (environment) contains multiple values. Why do you change the variable name to envs?

zjffdu:
> In my eyes, an env (environment) contains multiple values. Why do you change the variable name to envs?

You are right, let me revert that.

Reamer (approved):
LGTM

PR #4162, opened 2021-07-06
[ZEPPELIN-5441] Check no jobs running before kill interpreter job in TimeoutLifecycleManager

zjffdu:
### What is this PR for?

Minor PR to check no jobs running before kill interpreter job in `TimeoutLifecycleManager`, otherwise it is possible to kill interpreter process that still has jobs running.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5441

* Ci pass

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
@@ -57,6 +57,9 @@ public class TimeoutLifecycleManager extends LifecycleManager {
     ScheduledExecutorService checkScheduler = ExecutorFactory.singleton()
         .createOrGetScheduled("TimeoutLifecycleManager", 1);
     checkScheduler.scheduleAtFixedRate(() -> {
+      // check both lastBusyTimeInMillis & noJobsRunnings
+      // In one corner case, lastBusyTimeInMilli won't be updated, that is when zeppelin server is dead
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/interpreter/lifecycle/TimeoutLifecycleManager.java
```diff
@@ -53,7 +57,8 @@ public TimeoutLifecycleManager(ZeppelinConfiguration zConf,
     ScheduledExecutorService checkScheduler = ExecutorFactory.singleton()
         .createOrGetScheduled("TimeoutLifecycleManager", 1);
     checkScheduler.scheduleAtFixedRate(() -> {
-      if ((System.currentTimeMillis() - lastBusyTimeInMillis) > timeoutThreshold) {
+      if ((System.currentTimeMillis() - lastBusyTimeInMillis) > timeoutThreshold &&
+              noJobsRunning()) {
```
cuspymd, line 61: According to the original design, if a job is running, shouldn't `lastBusyTimeInMilli` be updated?

zjffdu: In one corner case, `lastBusyTimeInMilli` won't be updated, that is when zeppelin server is dead then `RemoteInterpreterServer#getProgress`  and `RemoteInterpreterServer#getStatus` won't be called. Thanks for the comment, let me update this PR with more comment to explain that.

PR #4163, opened 2021-07-06
[ZEPPELIN-5448] Hive job runs under anonymous user instead of default.user

zjffdu:
### What is this PR for?

Hive job should run under `default.user` instead of `anonymous`.  The root cause of this issue is that we didn't verify it is `anonymous` before passing it to jdbc.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5448

* Manually tested

image

### Questions:


cuspymd (approved):
Looks good!!

zjffdu:
Thanks for review @cuspymd  will merge it if no more comment

PR #4166, opened 2021-07-07
[ZEPPELIN-5439] Improve the logic of extract hive job url in JdbcInterpreter

zjffdu:
### What is this PR for?

Previously we extract hive job url from hive jdbc job log, but it seems the in the latest hive, there's no job url info in log. 
This PR is use another approach to do that, we use the paragraphId as the hive job yarn tag, after hive sql execution is started, we look at the yarn app that has this tag. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5439

* CI pass

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -529,7 +529,8 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
     final Properties properties = jdbcUserConfigurations.getPropertyMap(dbPrefix);
     String url = properties.getProperty(URL_KEY);
-    String connectionUrl = appendProxyUserToURL(url, user, dbPrefix);
+    url = appendProxyUserToURL(url, user, dbPrefix);
+    String connectionUrl = appendTagsToURL(url, context);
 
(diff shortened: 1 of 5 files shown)


jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -596,6 +598,24 @@ private String appendProxyUserToURL(String url, String user, String propertyKey)
     return connectionUrl.toString();
   }
 
+  // only add tags for hive jdbc
+  private String appendTagsToURL(String url, InterpreterContext context) {
+    StringBuilder builder = new StringBuilder(url);
+    if (url.startsWith("jdbc:hive2:")) {
+      Integer lastIndexOfQMark = builder.indexOf("?");
+      if (lastIndexOfQMark == -1) {
+        builder.append("?");
```
cuspymd, line 607: There seems be no appending "?" mark if it doesn't exit at `appendProxyUserToURL()` line 585~586.

zjffdu: Not sure what you mean. User may specify '?' in `default.url`

zjffdu:
Will merge if no more comment

PR #4169, opened 2021-07-09
[ZEPPELIN-5454] Explain output should be in text format

zjffdu:
### What is this PR for?

Use text format instead of table format for the spark sql explain statement. Because the table format is not readable. See the following screenshot. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5454

* UT is added and Manually tested

Before
image

After
image

### Questions:

### DIFF
--- a/spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
+++ b/spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
@@ -28,6 +28,7 @@ import org.apache.spark.sql.catalyst.expressions.GenericRow;
 import org.apache.spark.sql.types.StructType;
 import org.apache.spark.ui.jobs.JobProgressListener;
 import org.apache.zeppelin.interpreter.InterpreterContext;
+import org.apache.zeppelin.interpreter.InterpreterResult;
 import org.apache.zeppelin.interpreter.ResultMessages;
 import org.apache.zeppelin.interpreter.SingleRowInterpreterResult;
 import org.apache.zeppelin.tabledata.TableDataUtils;
@@ -81,9 +82,16 @@ public class Spark1Shims extends SparkShims {
         }
       }
 
+      InterpreterResult.Type outputType = InterpreterResult.Type.TABLE;
+      String sql = context.getLocalProperties().get("sql");
+      if (StringUtils.isNotBlank(sql) && sql.trim().toLowerCase().startsWith("explain")) {
+        outputType = InterpreterResult.Type.TEXT;
+      }
+
       StringBuilder msg = new StringBuilder();
-      msg.append("\n%table ");
+      msg.append("\n%" + outputType.toString().toLowerCase() + " ");
       msg.append(StringUtils.join(TableDataUtils.normalizeColumns(columns), "\t"));
+
       msg.append("\n");
       boolean isLargerThanMaxResult = rows.size() > maxResult;
       if (isLargerThanMaxResult) {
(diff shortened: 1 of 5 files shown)


spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
```diff
@@ -81,17 +82,28 @@ public String showDataFrame(Object obj, int maxResult, InterpreterContext contex
...
 
+      InterpreterResult.Type outputType = InterpreterResult.Type.TABLE;
+      String sql = context.getLocalProperties().get("sql");
+      if (StringUtils.isNotBlank(sql) && sql.trim().toLowerCase().startsWith("explain")) {
+        outputType = InterpreterResult.Type.TEXT;
+      }
+
       StringBuilder msg = new StringBuilder();
-      msg.append("\n%table ");
+      msg.append("\n%" + outputType.toString().toLowerCase() + " ");
```
Reamer, line 92: Use the StringBuilder to concatenate the string.

spark/spark2-shims/src/main/scala/org/apache/zeppelin/spark/Spark2Shims.java
```diff
@@ -81,17 +82,28 @@ public String showDataFrame(Object obj, int maxResult, InterpreterContext contex
...
 
+      InterpreterResult.Type outputType = InterpreterResult.Type.TABLE;
+      String sql = context.getLocalProperties().get("sql");
+      if (StringUtils.isNotBlank(sql) && sql.trim().toLowerCase().startsWith("explain")) {
+        outputType = InterpreterResult.Type.TEXT;
+      }
+
       StringBuilder msg = new StringBuilder();
-      msg.append("\n%table ");
+      msg.append("\n%" + outputType.toString().toLowerCase() + " ");
```
Reamer, line 92: Same as above.

spark/spark3-shims/src/main/scala/org/apache/zeppelin/spark/Spark3Shims.java
```diff
@@ -80,18 +82,28 @@ public String showDataFrame(Object obj, int maxResult, InterpreterContext contex
...
       }
+      InterpreterResult.Type outputType = InterpreterResult.Type.TABLE;
+      String sql = context.getLocalProperties().get("sql");
+      if (StringUtils.isNotBlank(sql) && sql.trim().toLowerCase().startsWith("explain")) {
+        outputType = InterpreterResult.Type.TEXT;
+      }
 
       StringBuilder msg = new StringBuilder();
-      msg.append("%table ");
+      msg.append("\n%" + outputType.toString().toLowerCase() + " ");
```
Reamer, line 92: Same as above.

Reamer (changes requested):
Just a few little things.

spark/spark1-shims/src/main/scala/org/apache/zeppelin/spark/Spark1Shims.java
zjffdu: I think JVM will optimize it via StringBuilder, I write in this way just for readability purpose. 
https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

Reamer: You may be right, but I get the following suggestion in intellij when using StringBuilder and normal string concatenation.
String_concatenation

zjffdu: I check the bytecode, jvm indeed do optimization here.

Reamer: I thought you were also using Intellij. Don't you get the comment in your IDE?

PR #4170, opened 2021-07-12
[ZEPPELIN-5360] K8S interpreter create spark pods, should support the 'imagePullSecrets' config

rickchengx:
### What is this PR for?
When Zeppelin is creating the interpreter pod, it may need to pull the image from the **private repository**. However, the current `k8s/interpreter/100-interpreter-spec.yaml` file does not support setting the config `imagePullSecrets`. 

So this PR adds a property named: `zeppelin.k8s.interpreter.imagePullSecrets`, which allows users to set the **comma-separated list of Kubernetes secrets**. So that the interpreter pod can pull the image from the private repository.

For example:
```
%spark.conf
zeppelin.k8s.interpreter.imagePullSecrets mysecret1,mysecret2,mysecret3
```

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5360>

* CI pass and manually tested

### Questions:

### DIFF
--- a/docs/quickstart/kubernetes.md
+++ b/docs/quickstart/kubernetes.md
@@ -245,15 +245,15 @@ to customize,
   4. Run a paragraph will create an interpreter using modified yaml files.
 
 The interpreter pod can also be customized through the interpreter settings. Here are some of the properties:
-| Property | Value | Description |
+| Property Name | Default Value | Description |
 | ----- | ----- | ----- |
-| `zeppelin.k8s.namespace` | `<k8s namespace>` | The Kubernetes namespace to use. |
-| `zeppelin.k8s.interpreter.container.image` | `<image>:<version>` | The interpreter image to use. |
-| `zeppelin.k8s.interpreter.cores` | `<cpu cores>` | The number of cpu cores to use. |
-| `zeppelin.k8s.interpreter.memory` | `<memory>` | The memory to use, e.g., `1g`. |
-| `zeppelin.k8s.interpreter.gpu.type` | `<gpu type>` | Set the type of gpu to request when the interpreter pod is required to schedule gpu resources, e.g., `nvidia.com/gpu`. |
-| `zeppelin.k8s.interpreter.gpu.nums` | `<gpu nums>` | Tne number of gpu to use. |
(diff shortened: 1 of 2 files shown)


Reamer:
Same as in #4171

> I don't think we need to include all the configuration options in our example 100-interpreter-spec.yaml. This file should be configured by the user.
> Maybe we should write more documentation on how to change the file to suit use cases.

rickchengx:
> Same as in #4171
> 
> > I don't think we need to include all the configuration options in our example 100-interpreter-spec.yaml. This file should be configured by the user.
> > Maybe we should write more documentation on how to change the file to suit use cases.

Hi, @Reamer 
Thanks for your comment. In order to support more k8s features, the `100-interpreter-spec.yaml` file may indeed become more complicated. But if users are required to modify `100-interpreter-spec.yaml`, it may lead to a poor user experience. Especially when the user directly runs the zeppelin server in the k8s cluster, every time the `100-interpreter-spec.yaml` file is modified, the docker image needs to be rebuilt and deployed.

Therefore, I think that common features should be placed in the `100-interpreter-spec.yaml` file as much as possible so that users can directly configure their desired properties in the interpreter settings. When users need to configure some more customized properties, they can modify the `100-interpreter-spec.yaml` file.

Users may often need to pull the image from a private repository, so I think this property is relatively important.

Reamer:
Please expand the documentation on this function so that new users know about it and how to use it.

Reamer:
> Especially when the user directly runs the zeppelin server in the k8s cluster, every time the `100-interpreter-spec.yaml` file is modified, the docker image needs to be rebuilt and deployed.

You can include the yaml file via a ConfigMap, then no restart is necessary.

rickchengx:
> Please expand the documentation on this function so that new users know about it and how to use it.

The documentation has been updated.

Reamer:
> The documentation has been updated.

Maybe it makes sense to describe the default values. What do you think?

rickchengx:
> > The documentation has been updated.
> 
> Maybe it makes sense to describe the default values. What do you think?

I agree. And I update the default values in the documentation. Note that some of the properties are optional (e.g., `zeppelin.k8s.interpreter.cores`).

PR #4171, opened 2021-07-12
[ZEPPELIN-5456] Allow users to set the config 'imagePullPolicy' under k8s mode

rickchengx:
### What is this PR for?
Currently, the interpreter pod created by `k8s/interpreter/100-interpreter-spec.yaml` cannot set the `imagePullPolicy` for both the zeppelin image and the spark image (When using the spark interpreter, the spark image needs to be set for the init container).

So this PR adds two properties:
* `zeppelin.k8s.interpreter.container.imagePullPolicy`, to set the pull policy of the zeppelin image.
* `zeppelin.k8s.spark.container.imagePullPolicy`, to set the pull policy of the spark image.

Users can set these two properties directly in the interpreter settings, such as:
```
%spark.conf
zeppelin.k8s.interpreter.container.imagePullPolicy Always
zeppelin.k8s.spark.container.imagePullPolicy Always
```

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5456>

* CI pass and manually tested

### Questions:

### DIFF
--- a/docs/quickstart/kubernetes.md
+++ b/docs/quickstart/kubernetes.md
@@ -255,6 +255,7 @@ The interpreter pod can also be customized through the interpreter settings. Her
 | `zeppelin.k8s.interpreter.gpu.nums` | (optional) | Tne number of gpu to use. |
 | `zeppelin.k8s.interpreter.imagePullSecrets` | (optional) | Set the comma-separated list of Kubernetes secrets while pulling images, e.g., `mysecret1,mysecret2` |
 | `zeppelin.k8s.interpreter.container.imagePullPolicy` | (optional) | Set the pull policy of the interpreter image, e.g., `Always` |
+| `zeppelin.k8s.spark.container.imagePullPolicy` | (optional) | Set the pull policy of the spark image, e.g., `Always` |
 
 ## Future work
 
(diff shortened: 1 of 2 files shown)


Reamer:
I don't think we need to include all the configuration options in our example `100-interpreter-spec.yaml`. This file should be configured by the user.
Maybe we should write more documentation on how to change the file to suit use cases.

rickchengx:
> I don't think we need to include all the configuration options in our example `100-interpreter-spec.yaml`. This file should be configured by the user.
> Maybe we should write more documentation on how to change the file to suit use cases.

Hi, @Reamer 
Thanks for your comment. In order to support more k8s features, the `100-interpreter-spec.yaml` file may indeed become more complicated. But if users are required to modify `100-interpreter-spec.yaml`, it may lead to a poor user experience. Especially when the user directly runs the zeppelin server in the k8s cluster, every time the `100-interpreter-spec.yaml` file is modified, the docker image needs to be rebuilt and deployed.

Therefore, I think that common features should be placed in the `100-interpreter-spec.yaml` file as much as possible so that users can directly configure their desired properties in the interpreter settings. When users need to configure some more customized properties, they can modify the `100-interpreter-spec.yaml` file.

I think `imagePullpolicy` is a property that users may often need to configure. For example, when users update images with the same tag in the repository, they can set the `imagePullpolicy` to be `Always`. Or when users need to shorten the time of image pull, they can set the `imagePullpolicy` to be `IfNotPresent`.

Reamer:
Please expand the documentation on this function so that new users know about it and how to use it.

Reamer:
I have merged #4170. Please rebase and add the documentation part.

rickchengx:
> I have merged #4170. Please rebase and add the documentation part.

Thanks, I have rebased and added the documentation part.

Reamer (approved):
LGTM

PR #4172, opened 2021-07-12
[ZEPPELIN-5444] Allow users to set service account for interpreter

rickchengx:
### What is this PR for?
Currently, users cannot set the service account for the interpreter pod under k8s mode. And there is no service account defined in the pod `spec` in `k8s/interpreter/100-interpreter-spec.yaml`. According to the documentation of Kubernetes (<https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/>):
> When you create a pod, if you do not specify a service account, it is automatically assigned the **default** service account in the same namespace.

, which means that currently the interpreter pod can only use the default service account.

In order to allow users to directly set the service account for the interpreter pod in the interpreter settings, this PR adds a k8s template property, named `zeppelin.k8s.interpreter.serviceAccount`.

Example usage:
```
%spark.conf
zeppelin.k8s.interpreter.serviceAccount yourServiceAccount
```

Note that the `k8s/interpreter/100-interpreter-spec.yaml` creates a **role binding** when using the spark interpreter. It was originally directly bound to the default service account. This PR makes it bind to the service account set by user.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5444>

* CI pass and manually tested

### Questions:

### DIFF
--- a/docs/quickstart/kubernetes.md
+++ b/docs/quickstart/kubernetes.md
@@ -248,6 +248,7 @@ The interpreter pod can also be customized through the interpreter settings. Her
 | Property Name | Default Value | Description |
 | ----- | ----- | ----- |
 | `zeppelin.k8s.namespace` | `default` | The Kubernetes namespace to use. |
+| `zeppelin.k8s.interpreter.serviceAccount` | `default` | The Kubernetes service account to use. |
 | `zeppelin.k8s.interpreter.container.image` | `apache/zeppelin:<ZEPPELIN_VERSION>` | The interpreter image to use. |
 | `zeppelin.k8s.interpreter.cores` | (optional)  | The number of cpu cores to use. |
 | `zeppelin.k8s.interpreter.memory` | (optional) | The memory to use, e.g., `1g`. |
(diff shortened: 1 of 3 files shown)


Reamer:
Please a documentation.
Missing parts:
 - How to create the service account

rickchengx:
> Please add documentation.
> Missing parts:
> 
> * How to create the service account

If user needs to set a specific service account (rather than default), he needs to create the account in advance.
For example, user can create a service account through `kubectl`:
```
kubectl create serviceaccount my-service-account
```

Reamer:
Please expand the [documentation](https://github.com/apache/zeppelin/blob/master/docs/quickstart/kubernetes.md) about this feature.

Reamer:
I have merged #4170. Please rebase and add the documentation part.

rickchengx:
> I have merged #4170. Please rebase and add the documentation part.

Thanks, I have rebased and added the documentation part.

Reamer:
Thanks @rickchengx for your good code contribution.

PR #4173, opened 2021-07-12
[ZEPPELIN-5455] Spark conf can not contain white space

zjffdu:
### What is this PR for?

This PR would use `|` as parameter delimiter, although it can not resolve the issue fundamentally, because parameter may still contain `|`, but it would mitigate the issue a lot. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5455

* Manually tested

image

### Questions:


Reamer:
@zjffdu 
Please take a look at #4127 and my [comment](https://github.com/apache/zeppelin/pull/4127#issuecomment-864825992). I tried to remove IFS, which is not correct for word splitting.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -106,7 +106,7 @@ public SparkInterpreterLauncher(ZeppelinConfiguration zConf, RecoveryStorage rec
               " to false if you want to use other modes.");
     }
 
-    StringBuilder sparkConfBuilder = new StringBuilder();
+    List<String> sparkConfList = new ArrayList<>();
```
Reamer, line 109: A StringJoiner is more elegant

https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html

zjffdu: Thanks for the review, PR is updated, `SparkInterpreterLauncher` & `FlinkInterpreterLauncher` both use StringJoiner now.

zjffdu:
Will merge if no more comment

PR #4174, opened 2021-07-12
[ZEPPELIN-5442] Allow users to set a separate namespace for the interpreter

rickchengx:
### What is this PR for?
Currently, if zeppelin is running under k8s mode and the zeppelin server is running inside the k8s cluster, it will start the interpreter pod under the namespace specified in `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. But if zeppelin-server is running **outside of kubernetes cluster**, it will use the config `zeppelin.k8s.namespace` to start the interpreter pod under the given namespace. However, this approach will cause users to only set a **global namespace**.

Hence, this PR allows user to set a set **separate namespace** for the interpreter in the interpreter settings:

Example usage:
```
%spark.conf
zeppelin.k8s.interpreter.namespace spark
```
The spark interpreter pod will be created under the namespace: `spark`.

```
%python.conf
zeppelin.k8s.interpreter.namespace python
```
The python interpreter pod will be created under the namespace: `python`.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5442>

* CI pass and manually tested

### Questions:

### DIFF
--- a/k8s/zeppelin-server.yaml
+++ b/k8s/zeppelin-server.yaml
@@ -201,7 +201,7 @@ kind: ServiceAccount
 metadata:
   name: zeppelin-server
 ---
-kind: Role
+kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: zeppelin-server-role
@@ -217,10 +217,11 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: zeppelin-server-role-binding
+  namespace: default
 subjects:
 - kind: ServiceAccount
   name: zeppelin-server
 roleRef:
-  kind: Role
+  kind: ClusterRole
   name: zeppelin-server-role
   apiGroup: rbac.authorization.k8s.io
--- a/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
@@ -55,7 +55,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
   private static final Logger LOGGER = LoggerFactory.getLogger(K8sRemoteInterpreterProcess.class);
   private static final int K8S_INTERPRETER_SERVICE_PORT = 12321;
   private final KubernetesClient client;
-  private final String namespace;
+  private final String interpreterNamespace;
   private final String interpreterGroupName;
   private final File specTemplates;
   private final String containerImage;
@@ -81,7 +81,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
 
   public K8sRemoteInterpreterProcess(
           KubernetesClient client,
-          String namespace,
+          String interpreterNamespace,
           File specTemplates,
           String containerImage,
           String interpreterGroupId,
@@ -110,7 +110,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
           interpreterGroupId,
           isUserImpersonatedForSpark);
     this.client = client;
-    this.namespace = namespace;
+    this.interpreterNamespace = interpreterNamespace;
     this.specTemplates = specTemplates;
     this.containerImage = containerImage;
     this.interpreterGroupName = interpreterGroupName;
@@ -132,8 +132,8 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
   /**
    * @return Get namespace
    */
-  public String getNamespace() {
-    return namespace;
+  public String getInterpreterNamespace() {
+    return interpreterNamespace;
   }
 
   /**
@@ -161,7 +161,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
       // WATCH
       PodPhaseWatcher podWatcher = new PodPhaseWatcher(
           phase -> StringUtils.equalsAnyIgnoreCase(phase, "Succeeded", "Failed", "Running"));
-      try (Watch watch = client.pods().inNamespace(namespace).withName(podName).watch(podWatcher)) {
+      try (Watch watch = client.pods().inNamespace(interpreterNamespace).withName(podName).watch(podWatcher)) {
         podWatcher.getCountDownLatch().await();
       } catch (InterruptedException e) {
         LOGGER.error("Interrupt received during waiting for Running phase. Try to stop the interpreter and interrupt the current thread.", e);
@@ -200,7 +200,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
     super.stop();
     // WATCH for soft shutdown
     PodPhaseWatcher podWatcher = new PodPhaseWatcher(phase -> StringUtils.equalsAny(phase, "Succeeded", "Failed"));
-    try (Watch watch = client.pods().inNamespace(namespace).withName(podName).watch(podWatcher)) {
+    try (Watch watch = client.pods().inNamespace(interpreterNamespace).withName(podName).watch(podWatcher)) {
       if (!podWatcher.getCountDownLatch().await(RemoteInterpreterServer.DEFAULT_SHUTDOWN_TIMEOUT + 500L,
           TimeUnit.MILLISECONDS)) {
         LOGGER.warn("Pod {} doesn't terminate in time", podName);
@@ -234,7 +234,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
 
   public String getPodPhase() {
     try {
-      Pod pod = client.pods().inNamespace(namespace).withName(podName).get();
+      Pod pod = client.pods().inNamespace(interpreterNamespace).withName(podName).get();
       if (pod != null) {
         PodStatus status = pod.getStatus();
         if (status != null) {
@@ -276,9 +276,9 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
       LOGGER.info("Apply {} with {} K8s Objects", path.getAbsolutePath(), k8sObjects.get().size());
       LOGGER.debug(template);
       if (delete) {
-        k8sObjects.inNamespace(namespace).delete();
+        k8sObjects.inNamespace(interpreterNamespace).delete();
       } else {
-        k8sObjects.inNamespace(namespace).createOrReplace();
+        k8sObjects.inNamespace(interpreterNamespace).createOrReplace();
       }
     } else {
       LOGGER.error("Can't apply {}", path.getAbsolutePath());
@@ -290,7 +290,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
     Properties k8sProperties = new Properties();
 
     // k8s template properties
-    k8sProperties.put("zeppelin.k8s.namespace", getNamespace());
+    k8sProperties.put("zeppelin.k8s.interpreter.namespace", getInterpreterNamespace());
     k8sProperties.put("zeppelin.k8s.interpreter.pod.name", getPodName());
     k8sProperties.put("zeppelin.k8s.interpreter.serviceAccount", getServiceAccount());
     k8sProperties.put("zeppelin.k8s.interpreter.container.name", interpreterGroupName.toLowerCase());
@@ -302,10 +302,15 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
     k8sProperties.put("zeppelin.k8s.interpreter.rpc.portRange", getInterpreterPortRange());
     k8sProperties.put("zeppelin.k8s.server.rpc.service", intpEventServerHost);
     k8sProperties.put("zeppelin.k8s.server.rpc.portRange", intpEventServerPort);
-    if (ownerUID() != null && ownerName() != null) {
+
+    String serverNamespace = K8sUtils.getCurrentK8sNamespace();
+    String interpreterNamespace = getInterpreterNamespace();
+    //Set the owner reference (zeppelin-server pod) for garbage collection when zeppelin server and the zeppelin interpreter is in the same namespace (Kubernetes cannot specify an owner in different namespace).
+    if (ownerUID() != null && ownerName() != null && StringUtils.equals(serverNamespace, interpreterNamespace)) {
       k8sProperties.put("zeppelin.k8s.server.uid", ownerUID());
       k8sProperties.put("zeppelin.k8s.server.pod.name", ownerName());
     }
+    
     Map<String, String> k8sEnv = new HashMap<>(getEnv());
     // environment variables
     k8sEnv.put(ENV_SERVICE_DOMAIN, getEnv().getOrDefault(ENV_SERVICE_DOMAIN, System.getenv(ENV_SERVICE_DOMAIN) == null ? "local.zeppelin-project.org" : System.getenv(ENV_SERVICE_DOMAIN)));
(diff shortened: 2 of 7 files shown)


zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sStandardInterpreterLauncher.java
```diff
@@ -73,6 +73,11 @@ String getNamespace() throws IOException {
     if (isRunningOnKubernetes()) {
       return readFile(Config.KUBERNETES_NAMESPACE_PATH, Charset.defaultCharset()).trim();
     } else {
+      //namespace may be set from the interpreter properties
```
Reamer, line 76: This feature should also be available with a Zeppelin server running on Kubernetes.

Reamer (changes requested):
Feature should also be available for Zeppelin server running in K8s + CI failures.

rickchengx:
> Feature should also be available for Zeppelin server running in K8s + CI failures.

Hi, @Reamer 
Thanks for your comment. I have fixed the CI failure.
After my test, when zeppelin server is also running inside k8s, there'll be a few more problems:
1. the zeppelin server needs to obtain the **cluster role** instead of role to create pods and services in other namespaces.
2. There're 2 namespaces: the namespace of zeppelin server and the namespace of zeppelin interpreter. So the namespace return by method `getNamespace` may be wrong if we return the `zeppelin.k8s.interpreter.namespace` while zeppelin server is also running in k8s.
3. When zeppelin server pod and interpreter pod are in different namespaces, the `ownerReference` can not be set. Since Cross-namespace owner references are disallowed by design <https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/>.

But if zeppelin server is running outside k8s, it can correctly create the interpreter pod in different namespaces. There won't be a problem of cross-namespace communication between zeppelin server and interpreter.

So how about support the feature of separate namespace while zeppelin server is running in k8s in another PR.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -275,7 +275,7 @@ Properties getTemplateBindings(String userName) {
     Properties k8sProperties = new Properties();
 
     // k8s template properties
-    k8sProperties.put("zeppelin.k8s.namespace", getNamespace());
+    k8sProperties.put("zeppelin.k8s.interpreter.namespace", getNamespace());
```
Reamer, line 278: This line is a breaking change. Please correct your PR description. People need to correct their own `interpreter-spec.yaml`.

Reamer (changes requested):
Thank you very much for your analysis. I understand the problems when Zeppelin server is running in K8s and they are not easy to solve.
Please add the documentation which explain the flag. Please make it clear that this feature only works when Zeppelin is running outside of K8s.

rickchengx:
> Thank you very much for your analysis. I understand the problems when Zeppelin server is running in K8s and they are not easy to solve.
> Please add the documentation which explain the flag. Please make it clear that this feature only works when Zeppelin is running outside of K8s.

Hi, @Reamer 
Thanks for the comment. Due to potential code conflicts, I think it's better to achieve the feature of setting the separate namespace for the interpereter when Zeppelin server is running inside k8s. 

So I added several commits to address the above three problems. Specifically,
1. Change `role` and `rolebinding` to `clusterrole` and `clusterrolebinding`
2. Update the conditions for setting `ownerReference`
3. Return the correct namespace of Zeppelin Server


Hence, it can set the separate namespace for each interpreter no matter Zeppelin server is running inside k8s or not.

For example, user can start a spark interpreter in the namespace of `spark`:
截屏2021-07-21 下午5 05 45
截屏2021-07-21 下午5 39 38

Meanwhile, user can start a shell interpreter in the namespace of `sh`:
截屏2021-07-21 下午5 02 55
截屏2021-07-21 下午5 03 04

k8s/zeppelin-server.yaml
```diff
@@ -213,14 +213,15 @@ rules:
...
 ---
-kind: RoleBinding
+kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: zeppelin-server-role-binding
 subjects:
 - kind: ServiceAccount
   name: zeppelin-server
+  namespace: default
```
Reamer, line 223: It is not sure that the `zeppelin-server` ServiceAccount is part of the `default` namespace.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sStandardInterpreterLauncher.java
```diff
@@ -70,6 +70,11 @@ boolean isRunningOnKubernetes() {
    * @throws IOException if namespace file could not be read
    */
   String getNamespace() throws IOException {
+    //namespace may be set from the interpreter properties
```
Reamer, line 73: The `K8sStandardInterpreterLauncher` is executed only on the Zeppelin server side.
Please rename this function to `getInterpreterNamespace()`.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -135,6 +139,43 @@ public String getNamespace() {
     return namespace;
```
Reamer, line 139: Rename this variable to `serverNamespace` and change the function name to `getServerNamespace()`.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
Reamer: I think this should be the place where you need to call another method, something like `getInterpreterNamespace()`, where the default value is the output of `getServerNamespace()`.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sStandardInterpreterLauncher.java
```diff
@@ -95,9 +100,11 @@ String getHostname() {
    */
   private String getZeppelinService(InterpreterLaunchContext context) throws IOException {
     if (isRunningOnKubernetes()) {
+      //The namespace of zeppelin server can only be read from Config.KUBERNETES_NAMESPACE_PATH while it runs in k8s cluster, it may be different from the namespace of interpreter
+      String serverNamespace = readFile(Config.KUBERNETES_NAMESPACE_PATH, Charset.defaultCharset()).trim();
```
Reamer, line 104: Please move this part to a separate function. Something like `getServerNamespace()`.

Reamer (commented):
Some comments to improve readability.
In most cases it is not clear which namespace (server, interpreter) is used.
We should also avoid duplicate code. I think it makes sense to outsource code to the utility class.

k8s/zeppelin-server.yaml
rickchengx: Yes, the namespace of zeppelin server is not sure. But the `clusterrolebinding` needs to specify the namespace of the service account, which is applied by user when user is creating the zeppelin server in k8s.

So if user does not apply the `zeppelin-server` in the default namespace, he also needs to modify this `namespace` field in the `clusterrolebinding`.

rickchengx:
Hi, @Reamer 
Thanks for the comments. I have moved the common functions (`isRunningOnKubernetes()`, `readFile()`, `getInterpreterNamespace()` and `getCurrentK8sNamespace()`) into `org.apache.zeppelin.interpreter.launcher.K8sUtils` to avoid duplicate code.

And I also make it more clear to distinguish the namespace of server and interpreter.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sUtils.java
```diff
@@ -74,4 +82,61 @@ private static long convertToBytes(String memory) {
...
+   * to use K8sStandardInterpreterLauncher. This is useful for development. It allows Zeppelin server running on your
+   * IDE and creates your interpreters in Kubernetes. So any code changes on Zeppelin server or kubernetes yaml spec
+   * can be applied without re-building docker image.
+   * @return true, if running on K8s
+   */
+  public static boolean isRunningOnKubernetes() {
+    return new File(Config.KUBERNETES_NAMESPACE_PATH).exists();
+  }
+
+  public static String readFile(String path, Charset encoding) throws IOException {
```
Reamer, line 138: Please change this to `private`.

k8s/zeppelin-server.yaml
Reamer: Instead of creating a `ClusterRole` and a `ClusterRoleBinding`, I would prefer an additional blueprint with a `Role` and a `RoleBinding`. Inside the `RoleBinding` the user has to replace the `Namespace`.

Reamer (changes requested):

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sUtils.java
rickchengx: Sure.

rickchengx:
> Instead of creating a ClusterRole and a ClusterRoleBinding, I would prefer an additional blueprint with a Role and a RoleBinding. Inside the RoleBinding the user has to replace the Namespace.

But if `zeppelin server` needs to create interpreter pod in another namespace, its service account needs to be granted a `clusterrolebinding` because a `rolebinding` only grants permissions within a specific namespace. What do you think?

Reamer:
In my opinion an additional step should be needed to give the Zeppelin server more rights on a namespace via a RoleBinding. Creating a ClusterRoleBinding gives too many rights to the zeppelin-server. With a ClusterRoleBinding the Zeppelin server has rights in namespaces that are not used by the Zeppelin server at all.

rickchengx:
> In my opinion an additional step should be needed to give the Zeppelin server more rights on a namespace via a RoleBinding. Creating a ClusterRoleBinding gives too many rights to the zeppelin-server. With a ClusterRoleBinding the Zeppelin server has rights in namespaces that are not used by the Zeppelin server at all.

Sorry for the late response. I agree that `clusterrole` may gives too many rights to the zeppelin server. But the zeppelin server  pod uses the service account created in <https://github.com/apache/zeppelin/blob/master/k8s/zeppelin-server.yaml#L199>, which is a namespaced resource.

Assuming that the zeppelin server and its service account are created in `default` namespace. And if the `zeppelin-server.yaml` uses the `rolebinding` to obtain the permissions in another namespace (such as `spark`), it cannot find the created service account in `default` namespace.

Or is there other ways to set the separate namespace with `role` and `rolebinding`?

Reamer:
To avoid creating the same `Role` in every namespace, a `ClusterRole` would be good.
You can then create a `RoleBinding` in each namespace, which has the Zeppelin server `ServiceAccount` as subject.
Example `RoleBinding` in `spark` namespace.

```
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: zeppelin-server-role-binding
  namespace: spark
subjects:
  - kind: ServiceAccount
    name: zeppelin-server
    namespace: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: zeppelin-server-role
```

I think that should be sufficient. This `RoleBinding` should be created manuell by the user for each namespace, which should be used be zepperlin.

rickchengx:
> To avoid creating the same `Role` in every namespace, a `ClusterRole` would be good.
> You can then create a `RoleBinding` in each namespace, which has the Zeppelin server `ServiceAccount` as subject.
> Example `RoleBinding` in `spark` namespace.
> 
> ```
> kind: RoleBinding
> apiVersion: rbac.authorization.k8s.io/v1
> metadata:
>   name: zeppelin-server-role-binding
>   namespace: spark
> subjects:
>   - kind: ServiceAccount
>     name: zeppelin-server
>     namespace: default
> roleRef:
>   apiGroup: rbac.authorization.k8s.io
>   kind: ClusterRole
>   name: zeppelin-server-role
> ```
> 
> I think that should be sufficient. This `RoleBinding` should be created manuell by the user for each namespace, which should be used be zepperlin.

Thanks for the comments. I have changed the `clusterrolebinding` to `rolebinding` and updated the docs. Now by default, user can create the interpreter pod in the `default` namespace. If the user wants to create and run the interpreter in another namespace, he also needs to add a `rolebinding` in the `zeppelin-server.yaml`

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -287,10 +288,15 @@ Properties getTemplateBindings(String userName) {
     k8sProperties.put("zeppelin.k8s.interpreter.rpc.portRange", getInterpreterPortRange());
     k8sProperties.put("zeppelin.k8s.server.rpc.service", intpEventServerHost);
     k8sProperties.put("zeppelin.k8s.server.rpc.portRange", intpEventServerPort);
-    if (ownerUID() != null && ownerName() != null) {
+
+    String serverNamespace = K8sUtils.getCurrentK8sNamespace();
+    String interpreterNamespace = getInterpreterNamespace();
+    //Set the owner reference (zeppelin-server pod) for garbage collection when zeppelin server and the zeppelin interpreter is in the same namespace (Kubernetes cannot specify an owner in different namespace).
+    if (ownerUID() != null && ownerName() != null && serverNamespace != null && serverNamespace.equals(interpreterNamespace)) {
```
Reamer, line 295: `K8sUtils.getCurrentK8sNamespace()` can return `null`. Please use [`StringUtils.equals`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#equals-java.lang.CharSequence-java.lang.CharSequence-)

Reamer (changes requested):

Reamer:
> Now by default, user can create the interpreter pod in the `default` namespace.

Is it necessary to add a namespace?

k8s/zeppelin-server.yaml
```diff
@@ -217,10 +217,12 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: zeppelin-server-role-binding
+  namespace: default
 subjects:
 - kind: ServiceAccount
   name: zeppelin-server
+  namespace: default
```
Reamer, line 224: Please remove the namespace parameter. Afterwards, the RoleBinding is created in the namespace in which the other resources are also created.

Reamer (changes requested):
Sorry for my late reply. Just a small change request.

rickchengx:
> Please remove the namespace parameter. Afterwards, the RoleBinding is created in the namespace in which the other resources are also created.

Sorry for the late response since I am busy with school stuff.
I have removed the namespace parameter and rebased this PR to avoid conflicts.

k8s/zeppelin-server.yaml
rickchengx: Sure.

Reamer (approved):
LGTM
This PR introduces a breaking change. Please update your custom interpreter-spec.yaml.
`zeppelin.k8s.namespace` -> `zeppelin.k8s.interpreter.namespace`

PR #4175, opened 2021-07-14
[ZEPPELIN-5459] Unable to change graph setting when paragraph is in RUNNING

zjffdu:
### What is this PR for?

The issue happens in Flink streaming visualization. If I change the graph setting when paragraph is in RUNNING, it would be revert back to its original setting when backend send data to frontend. The root cause is that zeppelin didn't save paragraph setting change when paragraph is in RUNNING.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5459

*  manually tested

Before

before

After
after

### Questions:


zjffdu:
Will merge if no more comment

PR #4176, opened 2021-07-15
[ZEPPELIN-5445] Allow users to access the spark UI outside the kubernetes cluster

rickchengx:
### What is this PR for?
Currently, when zeppelin server is running outside of k8s cluster, users cannot access the spark UI directly from the web page of Zeppelin. So this PR creates ingress resource in `k8s/interpreter/100-interpreter-spec.yaml` and enables the `SPARK JOB` button to jump directly to the URL defined by ingress.

If user does not specify the property `zeppelin.spark.uiWebUrl`, it will be automatically generated according to the template: `{{PORT}}-{{SERVICE_NAME}}.{{SERVICE_DOMAIN}}`, which is also the URL that the `SPARK JOB` button points to. So this PR also configured the ingress.host by this URL.

This PR also adds a property `zeppelin.k8s.spark.useIngress` to let user choose whether to create ingress. 
Example usage:
```
zeppelin.k8s.spark.useIngress true
```

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5445>

* CI pass and manually tested

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
@@ -317,6 +317,20 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
               getPodName(),
               k8sEnv.get(ENV_SERVICE_DOMAIN)
           ));
+
+      // configure interpreter property "zeppelin.k8s.spark.ingress.host" if not defined, to enable spark ui through ingress
+      String ingressHost = (String) properties.get("zeppelin.k8s.spark.ingress.host");
+      if (StringUtils.isBlank(ingressHost)) {
+        ingressHost = "{{PORT}}-{{SERVICE_NAME}}.{{SERVICE_DOMAIN}}";
+      }
+      properties.put("zeppelin.k8s.spark.ingress.host",
+          sparkUiWebUrlFromTemplate(
+              ingressHost,
+              webUiPort,
+              getPodName(),
+              k8sEnv.get(ENV_SERVICE_DOMAIN)
+          ));
+      
       // Resources of Interpreter Pod
       if (properties.containsKey(SPARK_DRIVER_MEMORY)) {
         String memory;
--- a/docs/quickstart/kubernetes.md
+++ b/docs/quickstart/kubernetes.md
@@ -245,6 +245,7 @@ to customize,
   4. Run a paragraph will create an interpreter using modified yaml files.
 
 The interpreter pod can also be customized through the interpreter settings. Here are some of the properties:
+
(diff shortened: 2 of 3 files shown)


zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -307,15 +307,19 @@ Properties getTemplateBindings(String userName) {
...
+                webUiPort,
+                getPodName(),
+                k8sEnv.get(ENV_SERVICE_DOMAIN)
+        );
+        properties.put("zeppelin.spark.uiWebUrl", "//" + bindedWebUrl);
+
+        //configure the ingress.host to binded web url, to enable spark ui through ingress
+        k8sProperties.put("zeppelin.k8s.spark.ingress.host", bindedWebUrl);
       }
-      properties.put("zeppelin.spark.uiWebUrl",
```
Reamer, line 312: Why are you moving this part to the `if`?
In my environment I have defined a custom `webUrl` and I need the function `SparkUiWebUrlFromTemplate`.

rickchengx: > Why are you moving this part to the `if`?
> In my environment I have defined a custom `webUrl` and I need the function `SparkUiWebUrlFromTemplate`.

Thanks for the comment. I realized that this is a mistake since the customed `zeppelin.spark.uiWebUrl` also needs to be bound by the function `SparkUiWebUrlFromTemplate`.

rickchengx:
When user does not specify the property `zeppelin.spark.uiWebUrl`, it will be automatically `//{{PORT}}-{{SERVICE_NAME}}.{{SERVICE_DOMAIN}}`. And the default value of `zeppelin.k8s.spark.ingress.host` is `{{PORT}}-{{SERVICE_NAME}}.{{SERVICE_DOMAIN}}`.

So if user does not customize the URL, he just needs to set the property `zeppelin.k8s.spark.useIngress true` to create the ingress. And he can directly click the button `SPARK JOB` to jump to the spark UI through ingress.

But if user customizes the URL, he can also set the property `zeppelin.k8s.spark.ingress.host` to enable the spark UI through ingress :
```
zeppelin.spark.uiWebUrl //{{SERVICE_NAME}}.{{SERVICE_DOMAIN}}
zeppelin.k8s.spark.useIngress true
zeppelin.k8s.spark.ingress.host {{SERVICE_NAME}}.{{SERVICE_DOMAIN}}
```

Reamer (changes requested):
Please add documentation.

rickchengx:
> Please add documentation. Everything else looks good.

Thanks for the comment. The documentation has been added.

Reamer (approved):
LGTM

PR #4177, opened 2021-07-15
[ZEPPELIN-5460] Handling of hidden files while applying files under k8s mode

rickchengx:
### What is this PR for?
Currently, if there are some hidden files in the directory `k8s/interpreter`, zeppelin just outputs some logs `Skip {...}` but does not actually skip those hidden files. So if there are some odd files (e.g., `.DS_Store`), there will be some errors while applying them to k8s.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5460>

* CI pass and manually tested

### Questions:


Reamer (approved):
LGTM

PR #4179, opened 2021-07-16
[ZEPPELIN-5458] fix that zeppelin can not parse columns  which contains chinese character

zhxiaoping:
### What is this PR for?
fix that zeppelin can not parse columns  which contains chinese character

this PR do two things

the first one thing insert a special character (/u0001) which nerver use after every chinese character,
so zeppeline can split string correctly, replace /u0001 with empty string, before add record to rows

the second thing avoid that chinese character is escaped.
### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5458

already added test case in test units

image

### Questions:

### DIFF
--- a/livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
+++ b/livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
@@ -19,9 +19,18 @@ package org.apache.zeppelin.livy;
 
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
-import java.util.Arrays;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Properties;
+import java.util.Arrays;
 import java.util.Map;
+import java.util.Locale;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.interpreter.InterpreterContext;
 import org.apache.zeppelin.interpreter.InterpreterException;
@@ -33,11 +42,6 @@ import org.apache.zeppelin.interpreter.ResultMessages;
 import org.apache.zeppelin.scheduler.Scheduler;
 import org.apache.zeppelin.scheduler.SchedulerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
-
-import static org.apache.commons.lang3.StringEscapeUtils.escapeEcmaScript;
 
 /**
  * Livy SparkSQL Interpreter for Zeppelin.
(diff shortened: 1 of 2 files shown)


livy/src/main/java/org/apache/zeppelin/livy/LivySparkSQLInterpreter.java
```diff
@@ -197,7 +200,18 @@ public FormType getFormType() {
...
+            "\u1100-\u115F" +
+            "\u2E80-\uA4CF" +
+            "\uAC00-\uD7A3" +
+            "\uF900-\uFAFF" +
+            "\uFE10-\uFE19" +
+            "\uFE30-\uFE6F" +
+            "\uFF00-\uFF60" +
+            "\uFFE0-\uFFE6" +
+            "])";
+    String output = str.replaceAll(fullWidthRegex, "$1\u0001");
```
zjffdu, line 214: Use reg expression doesn't seems an elegant approach to me, could you find thrid party library that has such ability ?

zhxiaoping: sorry,  I have no idea neither, this regx is copied from spark source.

zjffdu: Could you link where these spark source code ?

zhxiaoping: org.apache.spark.util.Utils#fullWidthRegex
image
this picture above is code where the chinese character has two placeholders

zjffdu: Thanks @zhxiaoping Could you add more comment ? Especially please comment where the spark code you refer ?

zhxiaoping: the regex is refered to org.apache.spark.util.Utils#fullWidthRegex

image

for spark chinese character has two placeholder, one placeholder is one char.
for zeppelin chinese has only one placeholder.  
they have different standards.
so zeppelin can not parse columns based on column size.
just because zeppelin take chinese character as one placeholder, but actually it is two placeholder.

this pr do two things
the first one thing  insert a special character (/u0001)  which nerver use  after every chinese character,
so zeppline can split string correctly,  replace /u0001 with empty string, before add record to rows

the second thing  avoid that chinese character is escaped.

zjffdu: I mean to add comment in your PR so that the next contributor can understand what this change is for

zhxiaoping: added comment in code

zjffdu: Thanks @zhxiaoping Let's wait for the CI

zjffdu:
CI passed， LGTM， thanks @zhxiaoping

PR #4180, opened 2021-07-19
[ZEPPELIN-5463] Introduce flink jm url template configuration

zjffdu:
### What is this PR for?

Introduce configuration `zeppelin.flink.uiWebUrl` as user specified job manager url, it could be used in 2 cases:
1. remote mode: where flink cluster is started already, so that job manager url is know
2. yarn/yarn-application mode: user can only access knox url or other vendor provided url. 

The usage of `zeppelin.flink.uiWebUrl` is consistent with `zeppelin.spark.uiWebUrl`

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5463

* Manually tested

### Questions:


zjffdu:
Will merge if no more comment

PR #4184, opened 2021-07-21
[ZEPPELIN-5465] SERVICE_DOMAIN may be null while Zeppelin server is running outside k8s

rickchengx:
### What is this PR for?
The function `sparkUiWebUrlFromTemplate` needs to use the variable `SERVICE_DOMAIN` to bind the `jinjava` template, which sets the environment variable `SERVICE_DOMAIN` as the default value.
However, user may not set the environment variable when Zeppelin server is running outside of k8s, and it will get a NullPointerExcepetion:
```
NullPointerException: null value in entry: SERVICE_DOMAIN=null
```

When the user does not care about `zeppelin.spark.uiWebUrl`, he does not need to set the `SERVICE_DOMAIN` environment.

So this PR change the default value of `SERVICE_DOMAIN` to `local.zeppelin-project.org` if `System.getenv(ENV_SERVICE_DOMAIN) == null` to avoid the potential NullPointerExcepetion.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5465>

* CI pass and manually tested

### Questions:


zjffdu:
LGTM

PR #4185, opened 2021-07-23
[ZEPPELIN-5466] Can not specify the spark image from the interpreter settings

rickchengx:
### What is this PR for?
When Zeppelin launches a spark interpreter pod under k8s mode, it will build the **spark submit options** through [buildSparkSubmitOptions()](https://github.com/apache/zeppelin/blob/master/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java#L372-L393)

But the property `--conf spark.kubernetes.container.image=` will use the value from the `zConf.getK8sSparkContainerImage()`. Note that although the interpreter properties overrides the values through

```
// interpreter properties overrides the values
k8sProperties.putAll(Maps.fromProperties(properties));
```
 
But the spark submit options has already been built. So if user sets the `spark.kubernetes.container.image` from the interpreter settings, it will not correctly set the spark image in the `spark submit options`.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* <https://issues.apache.org/jira/browse/ZEPPELIN-5466> 

* CI pass and manually tested

### Questions:


zjffdu:
LGTM

Reamer (approved):
LGTM

PR #4188, opened 2021-07-28
[ZEPPELIN-5471] spark yarn app is not stopped when cancel button is clicked

zjffdu:
### What is this PR for?

Currently, SparkSubmitInterpreter only kill the client process, but leave the spark yarn app alive when cancel method is call.
This PR would detect yarn app id from logs and associate it with the current paragraph. And in the cancel method, it would kill the associated yarn app. 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5471

* CI pass

### Questions:


zjffdu:

PR #4190, opened 2021-07-28
[ZEPPELIN-5470] set statement may not work when key value around quotes

zjffdu:
### What is this PR for?

Set statement will fail when key value is around quotes, this PR fix this issue by removing the quote, and unit test is added.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5470

* Ci pass

image

### Questions:


zjffdu:
Will merge if no more comment

PR #4191, opened 2021-07-31
[ZEPPELIN-5474] Add configuration to specify the port range of shiny app

zjffdu:
### What is this PR for?

Add configuration `zeppelin.R.shiny.portRange` for both R interpreter and SparkR, so that user can specify the portRange of shiny app. This would be useful when user has port usage constraint due to security reason. 

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5474

* Manually tested

### Questions:


cuspymd (approved):
Looks good!!

zjffdu:

PR #4192, opened 2021-07-31
[ZEPPELIN-5475] fix zep k8s service hostname

ebuildy:
### What is this PR for?

According doc at https://kubernetes.io/docs/concepts/services-networking/service/ , service lookup must end by svc.cluster.local. This PR fix it.

### What type of PR is it?
Bug fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5475

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
@@ -393,7 +393,7 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
   }
 
   private String getInterpreterPodDnsName() {
-    return String.format("%s.%s.svc",
+    return String.format("%s.%s.svc.cluster.local",
         getPodName(), // service name and pod name is the same
         getNamespace());
   }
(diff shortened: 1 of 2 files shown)


zjffdu:
Thanks @ebuildy for the contribution, I don't find that service lookup must end by svc.cluster.local in the doc, could you point me the exact place ?

ebuildy:
Hello @zjffdu , I found DNS spec at https://github.com/kubernetes/dns/blob/master/docs/specification.md , also GKE doc is more explicit https://cloud.google.com/kubernetes-engine/docs/concepts/service-discovery?hl=en , or in the classic doc at https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/.

The ``.cluster.local`` is the "cluster domain", in 99% of k8s cluster I have worked on, this is always the same value. But this can change, in a multiple clusters, this become ``.clusterset.local``. 

How do you test zeppelin on k8s?

zjffdu:
@ebuildy Thanks for the clarification, this sounds like convention, right ? Anyway, this PR LGTM

Reamer:
"Cluster.local" as you mentioned is not always correct. I would prefer either a domain value that can be configured or add documentation that you need to configure your cluster DNS service with a search domain as described in the [DNS K8s documentation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#namespaces-of-services).

ebuildy:
Absolutely right @Reamer , in practice, I have never seen a cluster with a different search domain.

Do you want me to create a config entry ? Something like ``k8s_cluster_search_domain``

Reamer:
> Do you want me to create a config entry ? Something like k8s_cluster_search_domain

Yes, please.
This is then a possible breaking change (for 1%). Please update your PR description.
You also need to update the documentation.
The default for the configuration value should of course be "cluster.local".

zjffdu:
ping @ebuildy

ebuildy:
sorry for the big late, I will do it this week!

PR #4193, opened 2021-08-01
[ZEPPELIN-5473] Enable z.show for R

zjffdu:
### What is this PR for?

Enable `z.show` in `%r.ir` to make it consistent across languages. Unit test and docs are added.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5473

* Ci pass and manually tested

image

image

### Questions:


cuspymd (approved):
Looks good!!

zjffdu:

tecgie:
Can you add a logic to check if the max result size is reached?  If the max is reached, print out a warning message.  Essentially, the same behavior as found in the python interpreter (zeppelin_context.py).

        if exceed_limit:
            print("\n%html <font color=red>Results are limited by {}.</font>".format(self.max_result))

zjffdu:
@tecgie Good catch, let me update it

zjffdu:
@tecgie PR is updated

tecgie:
Thanks for the fix.  Question.  In zeppelin_context.py, it gets the default max_result value from the getMaxResult method of the zeppelin context.  Here we have a hardcoded max (1000).  It seems the behavior is inconsistent, it would be good that if we can get the same default max result.

zjffdu:
@tecgie Thanks for the careful review, PR is updated, now I use `zeppelin.R.maxResult` to configure the max rows to display.

tecgie:
Looks good.  Thank you.

zjffdu:
Thanks for the review @tecgie will merge if no more comment

PR #4195, opened 2021-08-02
save

zjffdu:
### What is this PR for?

### What type of PR is it?

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:


flink/flink1.10-shims/src/main/java/org/apache/zeppelin/flink/Flink110Shims.java
```diff
@@ -337,4 +344,27 @@ public Map extractTableConfigOptions() {
     }
     return fields;
   }
+
+  public boolean isTimeIndicatorType(Object type) {
+    return FlinkTypeFactory.isTimeIndicatorType((TypeInformation<?>) type);
+  }
+
+  @Override
+  public Object lookupExecutor(ClassLoader classLoader, Object environmentSettings, Object sEnv) {
```
cuspymd, line 353: Every implementation of `isTimeIndicatorType()`, `lookupExecutor()` seems be the same. Is it better move it to `FlinkShims` class?

zjffdu: flink-shims don't have flink dependencies, so can not use these flink classes. Only sub-modules has flink dependency. This do cause code duplication, but I don't have better way to avoid that.

cuspymd: As bellow, the class `Flink110Shims` inherits the class `FlinkShims`.
Do you mean the method `Flink110Shims::lookupExecutor()` can't be moved to `FlinkShims::lookupExecutor()` due to a dependency issue?
```
public class Flink110Shims extends FlinkShims
```

zjffdu: Yes, if you take a look at the pom.xml under `flink-shims` module, there's no flink dependency, because we can not specify flink dependency here, it leave its child module to decide which flink version to use.

cuspymd: Thanks for the detailed explanation.

zjffdu:
CI is passed, ready for review

Reamer:
Maybe it is also time to remove old Flink versions.
Based on https://flink.apache.org/downloads.html only 1.12.x, 1.13.x and 1.14.x are the stable versions.

zjffdu:
> Maybe it is also time to remove old Flink versions. Based on https://flink.apache.org/downloads.html only 1.12.x, 1.13.x and 1.14.x are the stable versions.

That's right, this PR is for the next minor release of Zeppelin: 0.10.1, so I will keep compatibility, all versions of flink will be supported. I plan to remove the old version support in the next major release: 0.11.0. I will do it in another ticket.

zjffdu:
Will merge if no more comment

PR #4199, opened 2021-08-08
[ZEPPELIN-5481] Support to run R interpreter in yarn mode with customized conda env

zjffdu:
### What is this PR for?

Simple PR to make R interpreter run in yarn mode with customized conda env. 

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5481

* Manually tested

image

### Questions:


rlang/src/main/java/org/apache/zeppelin/r/IRInterpreter.java
```diff
@@ -152,6 +153,17 @@ protected void initIRKernel() throws IOException, InterpreterException {
     }
   }
 
+  @Override
+  protected Map<String, String> setupKernelEnv() throws IOException {
+    Map<String, String> envs = super.setupKernelEnv();
+    String pathEnv = envs.getOrDefault("PATH", "");
+    if (condaEnv != null) {
+      pathEnv = new File(".").getAbsolutePath() + "/" + condaEnv + "/bin:" + pathEnv;
```
Reamer, line 161: Please use `File.separator` and `File.pathSeparator` to build the `PATH` environment variable.

Reamer (changes requested):

rlang/src/main/java/org/apache/zeppelin/r/IRInterpreter.java
zjffdu: Fixed

Reamer: Looks weirder than I thought, but I think that should be correct.

Reamer (approved):
LGTM

zjffdu:
Will merge if no more comment

PR #4207, opened 2021-08-24
[ZEPPELIN-5494] Make shell interpreter concurrency configurable

zjffdu:
### What is this PR for?

Introduce property `zeppelin.concurrency.max` to make shell concurrency configurable, by default it is 10.

### What type of PR is it?
[Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5494

* CI pass

### Questions:

### DIFF
--- a/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
+++ b/shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
@@ -57,6 +57,7 @@ public class ShellInterpreter extends KerberosInterpreter {
   private static final String DEFAULT_TIMEOUT = "60000";
   private static final String DEFAULT_CHECK_INTERVAL = "10000";
   private static final String DIRECTORY_USER_HOME = "shell.working.directory.user.home";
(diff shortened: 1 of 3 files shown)


shell/src/main/java/org/apache/zeppelin/shell/ShellInterpreter.java
```diff
@@ -213,7 +214,17 @@ public int getProgress(InterpreterContext context) {
...
   public Scheduler getScheduler() {
     return SchedulerFactory.singleton().createOrGetParallelScheduler(
-        ShellInterpreter.class.getName() + this.hashCode(), 10);
+        ShellInterpreter.class.getName() + this.hashCode(), getMaxConcurrent());
+  }
+
+  private int getMaxConcurrent() {
+    try {
+      return Integer.valueOf(getProperty(MAX_CONCURRENCY));
+    } catch (Exception e) {
```
Reamer, line 223: Please catch only the `NumberFormatException`.

PR #4208, opened 2021-08-25
[ZEPPELIN-5497] Memory leak of GenericObjectPool

zjffdu:
### What is this PR for?

Should close `clientPool` explicitly, otherwise it won't be garbage collected. 

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5497?filter=-2

* CI pass

### Questions:


Reamer:
We should implement the [`AutoCloseable` interface](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) in this class.
Maybe we can use the `try-with-resources` statement in the class that uses the class.
Of course we have to rename the function name from `shutdown` to `close`.

PR #4214, opened 2021-09-02
[ZEPPELIN-5499] NPE in IPySparkInterpreter

zjffdu:
### What is this PR for?

The root cause is the `SparkConf` in PySparkInterpreter is null when IPySparkInterpreter#open is called because SparkInterpreter is not opened yet at that time. This PR would pass `SparkConf` explicitly to avoid this issue. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5499

* CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
@@ -172,6 +172,10 @@ public class PySparkInterpreter extends PythonInterpreter {
   @Override
   protected String getPythonExec() {
     SparkConf sparkConf = getSparkConf();
+    return getPythonExec(sparkConf);
+  }
+
+  String getPythonExec(SparkConf sparkConf) {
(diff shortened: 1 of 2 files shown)


spark/interpreter/src/main/java/org/apache/zeppelin/spark/PySparkInterpreter.java
```diff
@@ -172,6 +172,10 @@ public void setInterpreterContextInPython() {
   @Override
   protected String getPythonExec() {
     SparkConf sparkConf = getSparkConf();
+    return getPythonExec(sparkConf);
+  }
+
+  String getPythonExec(SparkConf sparkConf) {
```
Reamer, line 178: Please add a access modifier (e.g. `public`, `private`)

zjffdu: Only package level is needed, so I didn't specify access modifier

Reamer:
@zjffdu I have tested the PR in my environment and it works without NPE.
It would be nice if we could merge this quickly.

PR #4218, opened 2021-09-13
[ZEPPELIN-5472] Disable dynamic form for shell interpreter by default

zjffdu:
### What is this PR for?

This PR change the form type of shell interpreter from SIMPLE to NATIVE, so that ${var} won't be translated into dynamic form. Because ${var} is often used as variable reference, so we should not confuse user in Zeppelin. If user do want to use dynamic form, they can still use local property to override that. e.g.

```
%sh(form=simple)

echo ${name}

``` 

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5472

* Unit test is added.

### Questions:


zjffdu:
Will merge if no more comment

Reamer (approved):
LGTM

PR #4227, opened 2021-09-18
[ZEPPELIN-5526] interpreter setting still in error status when remove the invalid dependency

zjffdu:
### What is this PR for?

The root cause is that the status and errorReason is not reset after invalid dependency is removed. This PR just reset them and also add unit test for that. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5526

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4230, opened 2021-09-22
[ZEPPELIN-5525] Python vanillar interpreter doesn't' work in Python 3.8

zjffdu:
### What is this PR for?

The root cause is due to ast api is changed after python 3.8, see https://github.com/ipython/ipython/pull/11593.
The PR fixed this issue in `zeppelin_python.py`.  Besides that, this PR add tests for Python 3.8 for both python interpreter & spark interpreter (spark 2.4 doesn't support python 3.8, so only there's no python 3.8 test for spark 2.4).  `jupyter_client` 5 is used because jupyter_client 7 doesn't work (tracked in ZEPPELIN-5533)

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5525

* Ci pass

### Questions:


zjffdu:
CI is passed, ready for review

Reamer (approved):
LGTM
I saw that you removed most of the pinned package versions in conda :+1:

zjffdu:
Will merge if no more comment

PR #4232, opened 2021-09-24
[ZEPPELIN-5385] Wrong src folder for spark-shims modules

huage1994:
### What is this PR for?
Wrong src folder for spark-shims modules. We should put it in java folder instead of scala folder.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5385

* Ci pass

### Questions:


Reamer:
Good find. Let's wait for the CI.

zjffdu:
LGTM

PR #4233, opened 2021-09-24
ZEPPELIN-5536 Update protoc and grpc for Linux ARM64

martin-g:
### What is this PR for?

Update protoc and grpc to their first versions that support Linux ARM64 architecture.
With this the build on Linux ARM64 passes!

### What type of PR is it?
Bug Fix

### What is the Jira issue?

[ZEPPELIN-5536](https://issues.apache.org/jira/browse/ZEPPELIN-5536)

Run the build on Linux ARM64, e.g. at TravisCI or even better CircleCI on ARM64 hardware! 
Let me know if you want me to setup either of those!

### Questions:
* Is there breaking changes for older versions? - Hopefully not! There is one API break in Grpc APIs (`usePlaintext()`), but there might be problems in reading serialized messages with the old versions 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,6 +157,8 @@
     <quartz.scheduler.version>2.3.2</quartz.scheduler.version>
     <jettison.version>1.4.0</jettison.version>
     <jsoup.version>1.13.1</jsoup.version>
+    <protoc.version>3.5.0</protoc.version>
+    <grpc.version>1.26.0</grpc.version>
 
     <!-- test library versions -->
     <junit.version>4.12</junit.version>
--- a/rlang/pom.xml
+++ b/rlang/pom.xml
@@ -35,7 +35,6 @@
     <properties>
         <interpreter.name>r</interpreter.name>
         <spark.version>2.4.5</spark.version>
-        <grpc.version>1.15.0</grpc.version>
 
         <spark.archive>spark-${spark.version}</spark.archive>
         <spark.src.download.url>
--- a/zeppelin-jupyter-interpreter/pom.xml
+++ b/zeppelin-jupyter-interpreter/pom.xml
@@ -34,7 +34,6 @@
   <properties>
     <interpreter.name>jupyter</interpreter.name>
     <python.py4j.version>0.10.7</python.py4j.version>
-    <grpc.version>1.15.0</grpc.version>
   </properties>
 
   <dependencies>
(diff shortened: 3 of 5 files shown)


martin-g:
Actually the build passes only if I disable the test for the web app module, because they use PhantomJS which has no binary for Linux aarch64, and since PhantomJS is a dead project it will never have such officially.

jongyoul:
@martin-g Hello, I might not understand what you are saying but could you please explain more about PhantomJS and the result of this CI? It looks like all passed without disabling anything.

martin-g:
The CI passed because GitHub Actions run on x86_64 machines. 
This PR is about improving the situation on Linux aarch64 (ARM64).
Officially PhantomJS provides only x86_64 binaries. Trying to `npm install phantomjs` on non-x86_64 would fail.
Since PhantomJS is not maintained since few years some projects have replaced it with headless chrome/chromium.

jongyoul:
@martin-g Thank you for explaining it. I also think PhantomJS is quite old but we don't have many resources to fix/improve it. Do you have any idea to improve it?

zjffdu:
@martin-g Do you know whether GitHub Actions support ARM64 ? If so, we should enable it to avoid regression.

martin-g:
> Thank you for explaining it. I also think `PhantomJS` is quite old but we don't have many resources to fix/improve it. Do you have any idea to improve it?

I could try to replace it with headless Chromium! I'll keep you posted!

jongyoul:
> I could try to replace it with headless Chromium! I'll keep you posted!

Sounds like a great idea!! Thank you!!

martin-g:
> Do you know whether GitHub Actions support ARM64 ? If so, we should enable it to avoid regression.

GitHub Actions does not provide Linux ARM64 VMs at the moment (https://github.com/actions/virtual-environments/issues/2536).
But there are few ways to do it:
* self-hosted runner - There are security concerns with it!  GHA does not recommend using self-hosted runners for public repositories. Apache AirFlow uses custom GHA runner with better security. If you go this way then using [Oracle Cloud ARM64](https://blogs.oracle.com/cloud-infrastructure/post/announcing-github-actions-arm-runners-for-the-arm-compute-platform-on-oracle-cloud-infrastructure) for the runner is very easy! 
* Remote Kubernetes - I have blogged about it [here](https://martin-grigorov.medium.com/githubactions-build-and-test-on-huaweicloud-arm64-af9d5c97b766). The issue here is that the setup becomes a bit more complex

Other projects use TravisCI or CircleCI (my recommendation) for testing on Linux ARM64/PPC64/s390x. The issue is - one more CI to maintain and watch over.
Yet another option is to use Apache Jenkins or Buildbot.

zjffdu:
Thanks @martin-g for the investigation, I checked CircleCI. But it seems not free like github action. We have to think about other options. It would be done in another PR. Regarding this PR, LGTM +1

martin-g:
> But it seems not free like github action.

CircleCI is free for public repos.
And resource type `arm64.medium` is free. `arm4.large` is paid though!

martin-g:
> Sounds like a great idea!! Thank you!!

@jongyoul I will work on it in a separate PR!

zjffdu:
> CircleCI is free for public repos

ha, right, CircleCI has free plan for open source project.

martin-g:
@zjffdu I will prepare a PR for testing on Linux ARM64 at CircleCI!

zjffdu:
Thanks @martin-g I will merge this PR if no more comment

PR #4236, opened 2021-09-27
[ZEPPELIN-5540] fix the unexpected NPE when use hive with Flink

legendtkl:
### What is this PR for?
When we enable hive in Flink interpreter, we encounter the unexpected NPE. The root cause is the value `properties.getOrDefault("HIVE_CONF_DIR", System.getenv("HIVE_CONF_DIR"))` is null. It will throw NPE when apply the `toString()` method.

### What type of PR is it?
[Bug Fix ]

### Todos
None

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5540

CI pass

### Questions:


legendtkl:
hi @zjffdu , would you help review this 'tiny' PR? 

Thanks,
Kelu.

Reamer (approved):
LGTM

zjffdu:
Thanks for the contribution @legendtkl  LGTM

PR #4237, opened 2021-09-27
ZEPPELIN-5542 Use headless Firefox instead of PhantomJS

martin-g:
### What is this PR for?

Replaces the usage of PhantomJS for the JavaScript tests with headless Firefox.
PhantomJS is not maintained since few years.
Using a real browser will make the tests more close to the real usage.

### What type of PR is it?

Improvement

### What is the Jira issue?
* Jira https://issues.apache.org/jira/browse/ZEPPELIN-5542

* `mvn test -Pweb-e2e` should pass for `zeppenlin-web` module

### Questions:


epugh:
Any thoughts on using Puppeteer instead?   I migrated from PhantomJS to that project.

martin-g:
I've tried to use Puppeteer in another project about an year ago and I wasn't able to make it work for some reason :-/

epugh:
I might just try and copy what you did here for testing on https://github.com/o19s/quepid ;-)

zjffdu:
@martin-g & @EricGao888 What's the difference between `Puppeteer` and `PhantomJS`, I am not frontend expert. If there's no major difference, I am fine with any of them.

jongyoul:
> One advantage of Puppeteer is that it can download the browser for you. With Karma.js you have to make sure the browser is pre-installed.

It looks attractive when we test on the local machine. BTW, I'm also not an expert for frontend so I'm not sure how it takes cost when we change it but I, sometimes, hope we could change to Puppeteer.

PR #4240, opened 2021-10-02
[ZEPPELIN-5547] Rename note doesn't change the name field in file json

zjffdu:
### What is this PR for?

This PR would update note file if the file name is changed. Otherwise the `name` field in note json would be inconsistent with note file name. They would become consistent when user do save note operation afterwards. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5547

* CI pass

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
@@ -246,6 +246,13 @@ public class NoteManager {
 
     // update notebookrepo
     this.notebookRepo.move(noteId, notePath, newNotePath, subject);
+
+    // save note if note name is changed, because we need to update the note field in note json.
+    String oldNoteName = getNoteName(notePath);
+    String newNoteName = getNoteName(newNotePath);
+    if (!oldNoteName.equalsIgnoreCase(newNoteName)) {
+      this.notebookRepo.save(noteNode.note, subject);
+    }
   }
 
   public void moveFolder(String folderPath,
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.zeppelin.notebook;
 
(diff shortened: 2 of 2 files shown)


zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
```diff
@@ -78,7 +80,7 @@
 
   private Notebook notebook;
   private NoteManager noteManager;
-  private NotebookRepo notebookRepo;
+  private VFSNotebookRepo notebookRepo;
```
Reamer, line 83: Why is this necessary? I think we should work with the interface.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
```diff
@@ -246,6 +246,13 @@ public void moveNote(String noteId,
 
     // update notebookrepo
     this.notebookRepo.move(noteId, notePath, newNotePath, subject);
+
+    // save note if note name is changed, because we need to update the note field in note json.
+    String oldNoteName = getNoteName(notePath);
+    String newNoteName = getNoteName(newNotePath);
+    if (!oldNoteName.equalsIgnoreCase(newNoteName)) {
```
Reamer, line 253: I think it would be good to use StringUtils.equalsIgnoreCase(..) to be null safe.

Reamer (commented):
Just two small comments.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
zjffdu: Good point, Fixed

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NotebookTest.java
zjffdu: Fixed

Reamer (approved):
LGTM, but let's wait for CI

zjffdu:
Will merge if no more comment

PR #4242, opened 2021-10-05
[MINOR] Adopt maven wrapper to use the same version of Maven

jongyoul:
### What is this PR for?
Using maven wrapper when building and testing

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Change `mvn` to `./mvnw`

### What is the Jira issue?
N/A

Pass CI

### Questions:


Reamer:
I've heard of the Maven wrapper, but didn't know what the project was for until today. I read the following article about it.
https://www.baeldung.com/maven-wrapper
I am against using the maven wrapper for the following reasons.
1) A Jar is checked in the SCM.
2) I see it as an advantage to have multiple versions in the wild to build the zeppelin project.

jongyoul:
> A Jar is stored in the SCM.
This is a very similar mechanism with the Gradle wrapper which stores only one jar is to get the actual Gradle version only. As you already knew, the jar is not a maven itself. It's just a wrapper program to get a specific version of Maven when running maven wrapper for the first time.
> I see it as an advantage to have multiple versions in the wild to build the zeppelin project.
I don't think Maven is not unstable at all but the main advantage of using it is to keep the same version for CI and the development environment. If we want to multiple version of maven as we have multiple project using maven, it's quite complicated. If we use a maven wrapper, we don't care about the maven version for Zeppelin because maven wrapper download and use the specific version for Zeppelin project. Moreover, we don't have to install maven in my local machine to build Zeppelin - yes, I also think most Java developers already installed Maven :-) -

The most important thing is that it's not mandatory to use a maven wrapper. It's just a helper feature. Everyone doesn't have to use a maven wrapper. If someone wants to use local maven, it's ok to use it as it has the same behaviors. Nonetheless, the reason why I thought I would like to adopt it is that I just want to use the same version of CI. (Recently, I got different warning messages between CI and my local machine.)

PR #4255, opened 2021-10-14
[ZEPPELIN-5560] spark yarn app end with failed status in yarn-cluster mode

zjffdu:
### What is this PR for?

The root cause is that `RemoteInterpreterServer` would call System.exit to forceShutdown spark driver in yarn-cluster mode.
This PR would disable forceShutdown in spark's yarn-cluster mode.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5560

* CI

### Questions:


zjffdu:
Will merge if no more comment

PR #4263, opened 2021-10-29
[ZEPPELIN-5574] ConcurrentModificationException in JsonResponse.toString() when getNote with restAPI

zjffdu:
### What is this PR for?

The root cause of this issue is that when Note is serialized to json, it is being modified by another thread. This PR use thread-safe class for Note/Paragraph. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5574

* Tested it for 2 days in customer's environment and no issue happen again. (Before this PR, this issue happens everyday)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
@@ -64,6 +64,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
@@ -122,15 +123,19 @@ public class Note implements JsonSerializable {
   private static final DateTimeFormatter DATE_TIME_FORMATTER =
           DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
 
+  // Use CopyOnWriteArrayList to make Note thread-safe which is required by Note serialization
+  // (saving note to NotebookRepo or broadcast to frontend), see ZEPPELIN-5530.
   private CopyOnWriteArrayList<Paragraph> paragraphs = new CopyOnWriteArrayList<>();
   private String name = "";
   private String id;
   private String defaultInterpreterGroup;
   private String version;
 
-  private Map<String, Object> noteParams = new LinkedHashMap<>();
-  private Map<String, Input> noteForms = new LinkedHashMap<>();
-  private Map<String, List<AngularObject>> angularObjects = new HashMap<>();
+  // Use ConcurrentHashMap to make Note thread-safe which is required by Note serialization
+  // (saving note to NotebookRepo or broadcast to frontend), see ZEPPELIN-5530.
+  private Map<String, Object> noteParams = new ConcurrentHashMap<>();
+  private Map<String, Input> noteForms = new ConcurrentHashMap<>();
+  private Map<String, List<AngularObject>> angularObjects = new ConcurrentHashMap<>();
 
   /*
    * note configurations.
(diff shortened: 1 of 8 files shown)


Reamer:
@zjffdu 
I want to try to solve the concurrent problem in pull request #4252 with a ReentrantReadWriteLock. This solution should also catch this pull request.

Give me some time for the implementation.

zjffdu:
@Reamer Sure, let me know whey you are ready

Reamer:
I merged #4252 into master. Please rebase to the current master. I think this PR is still needed.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -327,10 +327,9 @@ public String cloneNote(String sourceNoteId, String newNotePath, AuthenticationI
             newNote.setConfig(new HashMap<>(sourceNote.getConfig()));
             newNote.setInfo(new HashMap<>(sourceNote.getInfo()));
             newNote.setDefaultInterpreterGroup(sourceNote.getDefaultInterpreterGroup());
-            newNote.setNoteForms(new HashMap<>(sourceNote.getNoteForms()));
-            newNote.setNoteParams(new HashMap<>(sourceNote.getNoteParams()));
```
Reamer, line 331: I like this approach because it makes it clear that we are creating a flat copy of the HashMap. Your new approach does not show this.
Perhaps we should have `getNoteForms()` and `getNoteParams()` return an unmodifiable HashMap.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -862,6 +881,9 @@ public boolean isTrash() {
   }
 
   public CopyOnWriteArrayList<Paragraph> getParagraphs() {
+    if (this.paragraphs == null) {
```
Reamer, line 884: After #4252, I see no way that `paragraphs` can be null.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
zjffdu: You are right, I have reverted this change

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
```diff
@@ -78,12 +81,18 @@
...
-  private Map<String, Object> config = new HashMap<>();
+  // Use ConcurrentHashMap to make Note thread-safe which is required by Note serialization
+  // (saving note to NotebookRepo or broadcast to frontend), see ZEPPELIN-5530.
+  private Map<String, Object> config = new ConcurrentHashMap<>();
   // form and parameter settings
   public GUI settings = new GUI();
   private InterpreterResult results;
   // Application states in this paragraph
-  private final List<ApplicationState> apps = new LinkedList<>();
+  private final Queue<ApplicationState> apps = new ConcurrentLinkedQueue<>();
```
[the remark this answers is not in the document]
zjffdu: Good catch, I don't think it is necessary, I have removed it.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -279,22 +284,36 @@ public void setDefaultInterpreterGroup(String defaultInterpreterGroup) {
     this.defaultInterpreterGroup = defaultInterpreterGroup;
   }
 
-  public Map<String, Object> getNoteParams() {
-    return noteParams;
+  public ConcurrentHashMap<String, Object> getNoteParams() {
+    if (noteParams == null) {
+      return new ConcurrentHashMap<>();
+    }
+    return new ConcurrentHashMap<>(noteParams);
```
Reamer, line 291: Why return a new ConcurrentHashMap?
I don't know if returning an Collections.unmodifiableMap might be a good approach here.
This [code part ](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L450-L455) should be the only place where the map is edited from another class. Here qw could force the class to create a new map to work on and then communicate it to the Note.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Note.java
```diff
@@ -279,22 +284,36 @@ public void setDefaultInterpreterGroup(String defaultInterpreterGroup) {
...
-    return noteForms;
+  public void removeNoteParam(String key) {
+    noteParams.remove(key);
+  }
+
+  public ConcurrentHashMap<String, Input> getNoteForms() {
+    if (noteForms == null) {
+      return new ConcurrentHashMap<>();
+    }
+    return new ConcurrentHashMap<>(noteForms);
```
Reamer, line 306: Why return a new ConcurrentHashMap?
I don't know if returning an Collections.unmodifiableMap might be a good approach here.
This [code part ](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L450-L455) should be the only place where the map is edited from another class. Here qw could force the class to create a new map to work on and then communicate it to the Note.

PR #4267, opened 2021-11-22
[ZEPPELIN-5580] Pass scala version from SparkInterpreterLauncher instead of detect it at runtime

huage1994:
### What is this PR for?
Currently we detect scala version via scala.util.Properties.versionString(); 
but it depends on the resource file library.version on classpath, sometimes user may package this resource of scala-2.11 into his jar which cause we detect the wrong scala version.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5580

CI pass

### Questions:

### DIFF
--- a/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+++ b/spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
@@ -74,6 +74,7 @@ public class SparkInterpreter extends AbstractInterpreter {
   private Object sparkSession;
 
   private SparkVersion sparkVersion;
+  private String scalaVersion;
   private boolean enableSupportedVersionCheck;
 
   public SparkInterpreter(Properties properties) {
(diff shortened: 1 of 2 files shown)


zjffdu:
@huage1994 Spark related CI is failed.

spark/interpreter/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
```diff
@@ -252,20 +253,28 @@ public SparkVersion getSparkVersion() {
     return sparkVersion;
   }
 
-  private String extractScalaVersion() throws InterpreterException {
-    String scalaVersionString = scala.util.Properties.versionString();
+  private String extractScalaVersion(SparkConf conf) throws InterpreterException {
+    String scalaVersionString;
+    if (conf.contains("zeppelin.spark.scala.version")) {
```
zjffdu, line 258: It's better to add comment on when `zeppelin.spark.scala.version` is used and when `scala.util.Properties.versionString()` is used

huage1994: Thanks for the review @zjffdu.  I've added comment.

zjffdu:
Thanks @huage1994 PR LGTM

zjffdu:
LGTM, failed tests are not related

PR #4268, opened 2021-11-24
[ZEPPELIN-5590] Support flink per job mode

zjffdu:
### What is this PR for?

This PR is to support flink per job mode when user click `Run all paragraphs` button.  It introduce new property `zeppelin.notebook.run_all.isolated` to control whether run the whole note in isolated mode (use isolated mode for interpreter and close interpreter after running whole note).

### What type of PR is it?
[ Feature | Documentation ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5590

* Manually testd

### Questions:

### DIFF
--- a/zeppelin-server/conf/notebook-authorization.json
+++ b/zeppelin-server/conf/notebook-authorization.json
@@ -1,0 +1,10 @@
+{
+  "authInfo": {
+    "2GMP1TW4Y": {
+      "readers": [],
+      "owners": [],
+      "writers": [],
+      "runners": []
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


zeppelin-server/conf/notebook-authorization.json
```diff
@@ -0,0 +1,10 @@
+{
```
Reamer, line 1: Why you include this file?

Reamer (changes requested):
Only a small question.

PR #4271, opened 2021-12-03
[ZEPPELIN-3493]

skyer9:
### What is this PR for?
First of all, thanks to https://github.com/babokim/zeppelin/tree/presto_acl_20160809
My work is heavily based on his code & idea.

Second, this is for JDBCInterpreter only.
But, you can find it is easy porting to other Interpreter.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-3493

* execute select sql responsing more then 1000 rows.
* check CSV(ALL) menu works.

### Questions:
* Does this needs documentation? not necessary, but good for convenience

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -19,6 +19,8 @@ package org.apache.zeppelin.rest;
 
 import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
+
+import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -36,6 +38,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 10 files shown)


skyer9:
# detailed action

- when execute sql, all data is saved to file
- when requested, download saved file.
- after four hour, saved file is deleted automatically.
- when there is no saved file, displayed data is downloaded.

zjffdu:
@skyer9 Thanks for the contribution, could you paste a screenshot about how this feature is used by users?

skyer9:
i added screenshot.
but it can take 10 seconds or more.
it's image is served on nano aws server

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -614,6 +619,51 @@ public Response updateParagraphConfig(@PathParam("noteId") String noteId,
...
+
+    Paragraph p = note.getParagraph(paragraphId);
+    if (p == null) {
+      return new JsonResponse(Status.NOT_FOUND, "paragraph not found.").build();
+    }
+
+    String resultFileParent = conf.getString(ZeppelinConfiguration.ConfVars.ZEPPELIN_RESULT_DATA_DIR);
+
+    String fileName = noteId + "_" + paragraphId;
+    final File file = new File(resultFileParent + "/" + fileName);
```
Reamer, line 651: Please use the following constructor, which takes the OS path separator into account. https://docs.oracle.com/javase/7/docs/api/java/io/File.html#File(java.lang.String,%20java.lang.String)

skyer9: thank you for review my code.
changed to OS path separator.
and replaced "/tmp" to System.getProperty("java.io.tmpdir")

Reamer (changes requested):
Please also add frontend integration to the newer zeppelin-web-angular.

PR #4274, opened 2021-12-19
[ZEPPELIN-5611] getNextSessionParagraph may remove finished paragraph before client get the final status

zjffdu:
### What is this PR for?

This issue happens when users use Zeppelin SDK to run jobs. The root cause is that before the client gets the final job status, the associated paragraph is deleted by Zeppelin server because another job is submitted and there's no available paragraphs in this session note(we have a limitation on the number of paragraphs, default is 500). 

This PR introduces another property `zeppelin.session.paragraph.keep.threshold`, so that we only delete the paragraph when it is finished before this threshold time gap. Besides that, we would also delete the earliest finished paragraph. 

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5611

* Unit tests are added

### Questions:


zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -346,7 +344,7 @@ public String importNote(String notePath,
   /**
    * Executes given paragraph with passed paragraph info like noteId, paragraphId, title, text and etc.
    *
-   * @param noteId
+   * @param note
```
Reamer, line 347: Typo or merge failure.

PR #4275, opened 2021-12-19
[ZEPPELIN-5582] Remove old NotebookRepo

zjffdu:
### What is this PR for?

This PR is to remove the old NotebookRepo which is not supported after 0.9, it is only there for upgrading from versions before 0.9. It is time to remove this to reduce maintenance effort.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5582)

* CI pass

### Questions:


Reamer (approved):
LGTM

jongyoul (approved):

PR #4282, opened 2022-01-06
[ZEPPELIN-5624] Check if the path directory is compliant.

iiiusky:
### What is this PR for?
Check the legitimacy of the settingId in requests from /api/interpreter/setting/{settingId}

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5624
CI
                
### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -1035,6 +1035,9 @@ private boolean inlineRemove(String id, boolean initiator) throws IOException {
     }
 
     File localRepoDir = new File(conf.getInterpreterLocalRepoPath() + "/" + id);
+    if (localRepoDir.getAbsolutePath().contains("..")) {
```
Reamer, line 1038: Any idea why deleting the localRepoDir is not part of the if block?

jongyoul: I also wonder which APIs or methods are affected by this logic? I think we'd better have tests for this change.

iiiusky: Normal access should not allow him to be able to use... / to access unintended directories

jongyoul: @iiiusky Ah, do you mean normal users should not call with the id including `..` and so on?

iiiusky: @jongyoul Yes, or it will filter .. all internally, and then go to a specific directory to start concatenating the final path, instead of completely trusting the string path passed in by the user

Reamer: > Any idea why deleting the localRepoDir is not part of the if block?

Let me clarify my statement.
Any idea why deleting the localRepoDir is not part of the if block `if (interpreterSettings.containsKey(id)) {` which is directly above this change?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -1035,6 +1035,9 @@ private boolean inlineRemove(String id, boolean initiator) throws IOException {
...
+      return removed;
+    }
     FileUtils.deleteDirectory(localRepoDir);
```
iiiusky, line 1041: @Reamer Because the following will be deleted whether the if statement is executed or not

Reamer: But is it logical to delete a localRepoDir without an interpreter existing?

iiiusky: Normally it is unreasonable, maybe it can be changed to this?

Reamer: At first I would try to move the deletion to the if block. After that, a path check should be unnecessary.
If this solution doesn't work I would prefer such code.
```java
File localRepoDir = new File(conf.getInterpreterLocalRepoPath())
if (Arrays.stream(localRepoDir.list()).anyMatch(id::equals)) {
    FileUtils.deleteDirectory(new File(localRepoDir, id));
}
```
`..` is not a good solution, as this does not protect against attacks that delete files inside the `localRepoDir`.

xufengnian:
Are you sure this way can really delete zeppelin application directory?
I just try same way in zeppelin 0.9.0 with docker,but it can not delete any directory.
As this log，any  id from the REST request will be deleted.but Function 'FileUtils.deleteDirectory' seem can not delete directory in used
By the way，if only check the id contains("..")，attacker maybe try to use "%2e%2e"
image

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
zjffdu: Agree with @Reamer , put it into if block will fix this issue as well.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -1038,8 +1038,10 @@ private boolean inlineRemove(String id, boolean initiator) throws IOException {
       removed = true;
     }
 
-    File localRepoDir = new File(conf.getInterpreterLocalRepoPath() + "/" + id);
-    FileUtils.deleteDirectory(localRepoDir);
+    File localRepoPath = new File(conf.getInterpreterLocalRepoPath());
```
zjffdu, line 1041: It is not necessary to compare interpreter setting id again here, just move this to if loop should fix it.

iiiusky:
```
if (interpreterSettings.containsKey(id)) {
      InterpreterSetting intp = interpreterSettings.get(id);
      intp.close();
      removeInterpreterSetting(id);
      if (initiator) {
        // Event initiator saves the file
        // Cluster event accepting nodes do not need to save files repeatedly
        saveToFile();
      }
      FileUtils.deleteDirectory(new File(localRepoPath, id));
      removed = true;
    }

```

Do you mean to do it this way?

iiiusky: @zjffdu

zjffdu: That's right

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -1038,8 +1038,10 @@ private boolean inlineRemove(String id, boolean initiator) throws IOException {
...
+    if (Arrays.stream(localRepoPath.list()).anyMatch(id::equals)) {
```
zjffdu, line 1042: It is not necessary to do a comparison again, because Zeppelin doesn't allow interpreter setting id to contain dot

zjffdu: ping @iiiusky

iiiusky: I'm sorry, I've been busy these two days, can you give me some time to deal with my own affairs first? Submit the code by next week at the latest

zjffdu: Sure, take it easy

zjffdu:
LGTM

PR #4286, opened 2022-01-13
[ZEPPELIN-5613] zeppelin-interpreter-parent: Add dependencyManagement for log4j2 in pom.xml

woowahan-jaehoon:
### What is this PR for?
I found that geode-interpreter using log4j2 so I added dependencyManagement for log4j2.

I think that this PR prevent potential log4jShell through future interpreter.

### What type of PR is it?
[Bug Fix]

### Todos
* Nothing

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5613

```
mvn dependency:tree
```

### Questions:
    * No
    * No
    * No

### DIFF
--- a/zeppelin-interpreter-parent/pom.xml
+++ b/zeppelin-interpreter-parent/pom.xml
@@ -33,6 +33,35 @@
   <version>0.9.0-SNAPSHOT</version>
   <name>Zeppelin: Interpreter Parent</name>
 
+  <properties>
+    <log4j2.version>2.17.1</log4j2.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-1.2-api</artifactId>
+        <version>${log4j2.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api</artifactId>
+        <version>${log4j2.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-core</artifactId>
+        <version>${log4j2.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-web</artifactId>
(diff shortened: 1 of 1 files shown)


zeppelin-interpreter-parent/pom.xml
```diff
@@ -33,6 +33,35 @@
...
 
+  <properties>
+    <log4j2.version>2.17.1</log4j2.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-1.2-api</artifactId>
```
zjffdu, line 44: This is for log4j 1 ?

Reamer: I think the library does almost the same as `slf4j-log4j12`. Redirect all output made against the old log4j 1.2 API to log4j2. https://logging.apache.org/log4j/2.x/log4j-1.2-api/index.html

I think we should drop this managed dependency in favor of `slf4j-log4j12` and exclude `log4j-1.2-api` in other dependencies.

woowahan-jaehoon: Zeppelin can be drop log4j 1.x but I think that dependencies cannot drop log4j 1.x. 

I found log4j2 dependencies in other dependencies for interpreter, not zeppelin own code.

Reamer: I also searched for log4j1 library usages in Zeppelin and only found the following class which still uses log4j1 methods. Additionally, here even the implementation is used and not only the API.
https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java

0Ams: @woowahan-jaehoon, @Reamer 
Is there no further progress on the content?
I think that excluding log4j1, applying log4j2 is necessary for security. I hope there will be an additional response.

jongyoul: In my understanding, Log4j1 has a potential security issue but Zeppelin doesn't use the methods which are known issues. Of course, I agree with you but we might need to investigate other components including old versions of Spark and Flink. So this PR looks fine still and let's discuss the security issues in another channel

PR #4287, opened 2022-01-13
[ZEPPELIN-5631] block notebook listed when user didn't have owner/reader/admin

woowahan-jaehoon:
### What is this PR for?
Block notebook listed when user didn't have owner/reader/admin.

When I upgraded zeppelin 0.7 to 0.9, all notebooks are listed all users even if public option was disabled.

`Runner` permission is added after 0.7 so when upgrade 0.7 to 0.9, runner permission is set blank(all user allowed). user who has runner permission is allowed to read notebook, current logic.

### What type of PR is it?
[Bug Fix]

### Todos
* Nothing

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5631

* Set black to runner permission in notebook

### Questions:
** No
** No
** No


zjffdu:
Should we fix it in the upgrading tool via setting the runner as the reader?

woowahan-jaehoon:
> 

I think that it is not enough that only apply upgrading tool because user can remove runner permission.

Possible to read notebook for runner without read permission is allowed?

PR #4288, opened 2022-01-13
[ZEPPELIN-5632] zeppelin-zengine: Skip load notebook when it is damaged.

woowahan-jaehoon:
### What is this PR for?
Skip load notebook when it is damaged.

When I upgraded zeppelin 0.7 to 0.9 and I converted notebook, some notebooks are damaged. Caused damaged notebooks, zeppelin cannot start completely. But damaged notebooks can find very very hard.

### What type of PR is it?
[Bug Fix]

### Todos
* Nothing

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5632

* Maybe create damaged notebook and start zeppelin...?

### Questions:
    * No
    * No
    * No

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
@@ -85,8 +85,8 @@ public class NoteManager {
         {
           try {
             addOrUpdateNoteNode(new NoteInfo(entry.getKey(), entry.getValue()));
-          } catch (IOException e) {
-            LOGGER.warn(e.getMessage());
+          } catch (Throwable e) {
+            LOGGER.warn("addOrUpdateNoteNode Fail: {}", entry.getKey() + " - " + e.getMessage(), e);
           }
         });
   }


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
```diff
@@ -78,8 +78,8 @@ private void init() throws IOException {
         {
           try {
             addOrUpdateNoteNode(new Note(new NoteInfo(entry.getKey(), entry.getValue())));
-          } catch (IOException e) {
-            LOGGER.warn(e.getMessage());
+          } catch (Throwable e) {
+            LOGGER.warn("addOrUpdateNoteNode Fail: {}", entry.getKey() + " - " + e.getMessage(), e);
```
zjffdu, line 82: I think it is better to catch the exception and log the warning when invoking NotebookRepos#get

zjffdu: And please rebase your PR first.

zjffdu:
ping @woowahan-jaehoon

zjffdu:
@woowahan-jaehoon Could you rebase it?

woowahan-jaehoon:
> @woowahan-jaehoon Could you rebase it?

I did.

zjffdu:
Thanks @woowahan-jaehoon , could you take a look at this comment?https://github.com/apache/zeppelin/pull/4288#discussion_r793226376

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
woowahan-jaehoon: > I think it is better to catch the exception and log the warning when invoking NotebookRepos#get

In my case, Throwed exception that line when I started zeppelin, not open notebook.

I checked `addOrUpdateNoteNode` method, It isn't use `NotebookRepos#get`.

Why do you think that?

zjffdu: Could you paste the stacktrace? It is weird how damaged notebook will happen if no reading is invoked.

Mabin-J: Unfortunately, I cannot get stacktrace, anymore because I moved company. 

I think that maybe filenames or folder names  were weird.

When I saw that exception, I tried to upgrade 0.7 to 0.9. So I ran script for upgrade notebooks. Many staff of company used zeppelin (not only data team), many notebook's name could be bad. (even use `*`, `(`, `)`,`[`, `]`, etc)

zjffdu: I see, could you add one comment to explain that the invalid file name would cause exception here?

zjffdu:
@woowahan-jaehoon Can you add unit test for it?

jongyoul:
@zjffdu I think this issue introduces more potential problems but it's not related to this PR directly. How about merging it first? It looks like a useful log. WDYT?

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
Mabin-J: I’m sorry, but I’m not familiar with this anymore since I moved to a different job.

However, I think Zeppelin should skip reading notebooks that contain errors and log those errors for faster operation and troubleshooting.

PR #4298, opened 2022-02-25
[ZEPPELIN-5660] Implement move gcsnotebookrepo folder

BagusThanatos:
### What is this PR for?
This is to implement move function that is used for renaming folders and moving it to trash.
Currently it would result in a no op, making users unable to rename folders if using GCS as the notebook repository.

### What type of PR is it?
Feature

### What is the Jira issue?
[* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/
](https://issues.apache.org/jira/browse/ZEPPELIN-5660)

- CI
- A custom jar using this patch is currently being used in our Zeppelin instance

### Questions:


zeppelin-plugins/notebookrepo/gcs/src/test/java/org/apache/zeppelin/notebook/repo/GCSNotebookRepoTest.java
```diff
@@ -187,6 +187,28 @@ public void testRemove() throws Exception {
     assertThat(storage.get(makeBlobId(runningNote.getId(), runningNote.getPath()))).isNull();
   }
 
+  @Test
+  public void testRemoveFolder_nonexistent() throws Exception {
+    try {
+      notebookRepo.remove("id", "/name", AUTH_INFO);
+      fail();
+    } catch (IOException e) {}
```
Reamer, line 195: try catch block is not necessary.

zeppelin-plugins/notebookrepo/gcs/src/test/java/org/apache/zeppelin/notebook/repo/GCSNotebookRepoTest.java
```diff
@@ -202,6 +224,29 @@ public void testMove() throws Exception {
     assertThat(storage.get(makeBlobId(runningNote.getId(), runningNote.getPath()))).isNull();
   }
 
+  @Test
+  public void testMoveFolder_nonexistent() throws Exception {
+    try {
+      notebookRepo.move("/name", "/name_new", AUTH_INFO);
+      fail();
+    } catch (IOException e) {}
```
Reamer, line 232: try catch block is not necessary.

Reamer (approved):
Just little comments, everything else LGTM

zeppelin-plugins/notebookrepo/gcs/src/test/java/org/apache/zeppelin/notebook/repo/GCSNotebookRepoTest.java
BagusThanatos: If I removed the try catch block, then the test would fail.

The purpose of this test is to make sure that the function would throw an exception if it cannot remove the folder, resulting in the `fail()` part of the test to never be reached.

Reamer: Perhaps the following article will help you. This shows you how your test expects an exception.
https://www.baeldung.com/junit-assert-exception

I would prefer 
`@Test(expected = IOException.class)`

BagusThanatos: I see

That can do

I was following the previous tests written in the class

Changing it.

BagusThanatos: @Reamer 
Updated
Please check again
Thanks!

zjffdu:
LGTM

PR #4300, opened 2022-03-02
[ZEPPELIN-5663] Provide REST API about notebook version control

huage1994:
### What is this PR for?
This PR is to provide REST API about notebook version control

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5663

* CI passed

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -371,7 +371,7 @@ public class NotebookService {
   /**
    * Executes given paragraph with passed paragraph info like noteId, paragraphId, title, text and etc.
    *
-   * @param noteId
+   * @param note
    * @param paragraphId
    * @param title
    * @param text
@@ -1012,12 +1012,14 @@ public class NotebookService {
 
   }
 
-  public void getNotebyRevision(String noteId,
+  // notebook.getNoteByRevision(...) does not use the NoteCache,
+  // so we can return a Note object here.
+  public Note getNotebyRevision(String noteId,
                                 String revisionId,
                                 ServiceContext context,
                                 ServiceCallback<Note> callback) throws IOException {
 
-    notebook.processNote(noteId ,
+    return notebook.processNote(noteId,
       note -> {
         if (note == null) {
           callback.onFailure(new NoteNotFoundException(noteId), context);
@@ -1031,7 +1033,7 @@ public class NotebookService {
         Note revisionNote = notebook.getNoteByRevision(noteId, note.getPath(), revisionId,
             context.getAutheInfo());
         callback.onSuccess(revisionNote, context);
-        return null;
+        return revisionNote;
       });
   }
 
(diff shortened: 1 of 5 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -1012,12 +1012,12 @@ public void setNoteRevision(String noteId,
 
   }
 
-  public void getNotebyRevision(String noteId,
+  public Note getNotebyRevision(String noteId,
```
Reamer, line 1015: Could you please put the following comment to this method.
`notebook.getNoteByRevision(...) does not use the NoteCache, so we can return a Note object here.`

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
```diff
@@ -164,6 +165,112 @@ public void testGetNoteByPathWithPathNotExist() throws IOException {
...
   }
 
+  @Test
+  public void testGetNoteRevisionHistory() throws IOException {
+    LOG.info("Running testGetNoteRevisionHistory");
+    String note1Id = null;
+    Notebook notebook = TestUtils.getInstance(Notebook.class);
+    try {
+      String notePath = "note1";
+      note1Id = TestUtils.getInstance(Notebook.class).createNote(notePath, anonymous);
```
Reamer, line 175: Use `notebook` here.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
```diff
@@ -164,6 +165,112 @@ public void testGetNoteByPathWithPathNotExist() throws IOException {
...
+  }
+
+  @Test
+  public void testGetNoteByRevision() throws IOException {
+    LOG.info("Running testGetNoteByRevision");
+    String note1Id = null;
+    Notebook notebook = TestUtils.getInstance(Notebook.class);
+    try {
+      String notePath = "note1";
+      note1Id = TestUtils.getInstance(Notebook.class).createNote(notePath, anonymous);
```
Reamer, line 236: use `notebook` here.

Reamer (approved):
PullRequest LGTM, just a couple of really small comments.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
huage1994: > Could you please put the following comment to this method. `notebook.getNoteByRevision(...) does not use the NoteCache, so we can return a Note object here.`

Sure! Thanks a lot for your review!

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
huage1994: Thanks for giving me the good advice!  I've changed all the code like this in this PR.

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
huage1994: Thanks for giving me the good advice!  I've changed all the code like this in this PR.

PR #4305, opened 2022-03-11
[ZEPPELIN-5648] log4j.properties is not configured properly for spark interpreter

zjffdu:
### What is this PR for?

For spark interpreter, the class name and method are not printed (It is always Logging.scala).
This PR fixed this issue by updating the log4j.properties.

### What type of PR is it?
[Bug Fix ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5648
*
* Manually tested

### Questions:


Reamer:
Please provide an example (after, before).
It seems that the Zeppelin server log4j configuration is also adjusted. There should not be the `Logging.scala` problem.

zjffdu:
Thanks for the review @Reamer  I added the screenshot

Reamer:
Thanks for the screenshots.
I think that the line number `:57` is very confusing. We should remove the line number of `Logging.scala` as well.
In addition, it seems that you also adjust the log4j configuration of the Zeppelin server. The Zeppelin server is not affected by this issue. Here, of course, the previous output is much more helpful.

zjffdu:
Sorry for late response @Reamer `:57 is very confusing` This is in the first screenshot (before this PR), the second screenshot is the logs after this PR. 
The reason I update log4j.properties of zeppelin-server is just to make the logs of zeppelin-server & interpreter consistent.

Reamer:
I would prefer the Spark log format without `:%L`. Otherwise we have the confusing log line number.
Take a look at the Spark project.
https://github.com/apache/spark/blob/v3.2.0/conf/log4j.properties.template#L23

PR #4307, opened 2022-03-11
[ZEPPELIN-5604] Interpreter scheduler may not be shutdown properly

zjffdu:
### What is this PR for?

The issue will cause the thread used by the interpreter scheduler is not released, so overtime the thread pool may be exhausted. The root cause is that some uncaught exception may happen when closing interpreter, which causes the interpreter scheduler shutdown is not invoked. This PR put the scheduler shutdown in the finally block, so that it is always called.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5604

* CI pass

### Questions:


zjffdu:
Will merge if no more comment

PR #4308, opened 2022-03-13
[ZEPPELIN-5655] OSSNotebookRepo support version control.

huage1994:
### What is this PR for?
This PR is to make OSSNotebookRepo support version control.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5655

* CI 
* automated unit tests on aliyun oss:  Run `mvn test -pl zeppelin-plugins/notebookrepo/oss/`

### Questions:


huage1994:
Hi all, 
CI failed because that I add add a new test file `OSSNotebookRepoTest` for OSSNotebook and I can't  provide my private aliyun OSS accessKey and secret in test code.

Actually `OSSNotebookRepoTest` has been tested and  passed in my local environment.

Should I add `@ignore` for  `OSSNotebookRepoTest.class` ?  
Or do we have aliyun OSS accessKey and secret  in our CI  environment?

Reamer:
Do you see any way to mock the service? 
External service connections should be avoided due to availability and latency.

huage1994:
> Do you see any way to mock the service? External service connections should be avoided due to availability and latency.

Thx a lot for your advice @Reamer . I'm not sure how to mock but I will do some research.

huage1994:
Hi @Reamer and @zjffdu !  
In `MockStorageOperator`,   I use the local file system to mock the operation on OSS Service.  
So the test case. can be run and it has passed.

Please help review the code if you have time. 😄 😄

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepo.java
```diff
@@ -214,4 +190,89 @@ public void updateSettings(Map<String, String> settings, AuthenticationInfo subj
     LOGGER.warn("Method not implemented");
   }
 
+
+  String buildRevisionsDirName(String noteId, String notePath) throws IOException {
```
Reamer, line 194: Please add access specifiers and possible `static` at this point.

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/storage/MockStorageOperator.java
```diff
@@ -0,0 +1,99 @@
+package org.apache.zeppelin.notebook.repo.storage;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.filefilter.TrueFileFilter;
+
```
Reamer, line 5: `MockStorageOperator.java` should be moved to the test folder.

zeppelin-plugins/notebookrepo/oss/src/test/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepoTest.java
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  @After
+  public void tearDown() throws InterruptedException, IOException {
+    if (notebookRepo != null) {
+      notebookRepo.close();
+    }
+    ossOperator.deleteDir(bucket, "");
+    ossOperator.deleteBucket(bucket);
+    Thread.sleep(1000);
```
Reamer, line 81: Why a `sleep`?

Reamer (changes requested):
I have made a small review. Thanks for the mocking of the storage. I think that a mock with the local file system is sufficient.

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepo.java
```diff
@@ -63,144 +53,130 @@ public void init(ZeppelinConfiguration conf) throws IOException {
     String endpoint = conf.getOSSEndpoint();
     bucketName = conf.getOSSBucketName();
     rootFolder = conf.getNotebookDir();
-    // rootFolder is part of OSS key which doesn't start with '/'
-    if (rootFolder.startsWith("/")) {
-      rootFolder = rootFolder.substring(1);
-    }
+    NOTE_MAX_VERSION_NUM = conf.getOSSNoteMaxVersionNum();
+    // rootFolder is part of OSS key which doesn't start with '/' or './' or './/'
```
zjffdu, line 57: It's better to move this comment to function `formatPath`

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/storage/OSSOperator.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class OSSOperator implements RemoteStorageOperator {
```
zjffdu, line 32: Add a comment to explain the difference between OSSOperator & OSSClient.

zeppelin-plugins/notebookrepo/oss/src/test/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepoTest.java
huage1994: > Why a `sleep`?

The delete operations on OSS Service above has a delay.
It's ok when test with `MockStorageOperator` but it would affect setup of next test case if we do not wait for them to end when test with `OSSOperator`.

huage1994:
Hi @Reamer and @zjffdu .
Your reviews is so valuable to me. 👍 👍 

And I have corrected my code.

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/storage/OSSOperator.java
Reamer: Please use Javadoc for your class description.
https://www.oracle.com/de/technical-resources/articles/java/javadoc-tool.html#descriptions

zeppelin-plugins/notebookrepo/oss/src/test/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepoTest.java
Reamer: I don't understand. With the TempDir approach, each test should get its own TempDir. Also, the deletion of the files is a synchronous operation.
Which component deletes asynchronously here?

Reamer:
Please insert the license header into `MockStorageOperator.java` and `RemoteStorageOperator.java`.

zeppelin-plugins/notebookrepo/oss/src/test/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepoTest.java
huage1994: Yes, you are right, actually `sleep` is not needed in  in zeppelin Github CI. 
Currently I use MockStorageOperator to mock OSSOperator to run `OSSNotebookRepoTest` in zeppelin Github CI.

But `sleep` is needed when someone use `OSSOperator` instead of `MockStorageOperator`  to run the test `OSSNotebookRepoTest` in local environment.  It is the reason I want to reserve it.

> I don't understand. With the TempDir approach, each test should get its own TempDir. Also, the deletion of the files is a synchronous operation. Which component deletes asynchronously here?

TempDir approach is used in  `MockStorageOperator` but not in real OSSOperator.  OSSOperator use only one bucket.
However, when i use `OSSOperator` instead of `MockStorageOperator`  to run the test `OSSNotebookRepoTest` in my local environment. It would throw `OSSException` `the bucket already exists` if there is no `Thread.sleep(1000);`  when the setup of second test case.

huage1994:
> Please insert the license header into `MockStorageOperator.java` and `RemoteStorageOperator.java`.:
👍 

I did forget. License header has been added.

docs/setup/storage/storage.md
```diff
@@ -454,6 +454,12 @@ And you should configure oss related properties in file **zeppelin-site.xml**.
   <description>Access key secret for your OSS account</description>
 </property>
 
+<property>
+  <name>zeppelin.notebook.oss.version.max</name>
+  <value></value>
```
Reamer, line 459: Please insert "30" as default value at this point.

Reamer (approved):
Just one last little comment. Everything else LGTM. I am not able to test this function, but I think @huage1994 has done a lot of testing.

huage1994:
> Just one last little comment. Everything else LGTM. I am not able to test this function, but I think @huage1994 has done a lot of testing.

Done. Thanks you for the reviews!!

zjffdu (approved):

zjffdu:
Will merge if no more comment

Reamer:
ping @zjffdu

jongyoul:
I just wonder if it's possible to enable/disable this feature without any side effect? Of course, versions will be deleted/ignored by disabling this feature.

Reamer:
> I just wonder if it's possible to enable/disable this feature without any side effect? Of course, versions will be deleted/ignored by disabling this feature.

I don't know of a configuration value to turn the history version on or off.

zjffdu:
Good point @jongyoul , is it possible to turn off this feature by setting `zeppelin.notebook.oss.version.max` to 1? @huage1994

huage1994:
> Good point @jongyoul , is it possible to turn off this feature by setting `zeppelin.notebook.oss.version.max` to 1? @huage1994

HI all,
I think if we enable/disable version control,  the webUI should be changed too.    
It is judged by `notebookRepo instanceof NotebookRepoWithVersionControl`  in `Notebook.java` at line  764.

Actually, we just need to change webUI to turn off version control, and it is applicable to all notebookRepo. 

So my idea is to add a configuration value to enable/disable version control globally, and enable/disable version control  widget in webUI. We just need to add a `if`  statement in `Notebook.java` .

jongyoul:
@huage1994 Agreed that a mew global configuration should handle enabling/disabling versioning feature and it could be handled in another PR. So I wondered if the versioning PR could handle it internally or not.

huage1994:
> @huage1994 Agreed that a mew global configuration should handle enabling/disabling versioning feature and it could be handled in another PR. So I wondered if the versioning PR could handle it internally or not.

👍 Good point. 
As @zjffdu said, I think it is ok that turn off this feature by setting zeppelin.notebook.oss.version.max to 0.
I would do more test to verify.

jongyoul:
@huage1994 Sounds great.

>I would do more tests to verify.

After testing it, please let me know. I'll merge it.

huage1994:
> @huage1994 Sounds great.
> 
> > I would do more tests to verify.
> 
> After testing it, please let me know. I'll merge it.

Hi @jongyoul , I made a small code change in the latest commit. 
I modified the checkpoint method and revisionHistory method for  zeppelin.notebook.oss.version.max setting to 0. 

I have tested it on Aliyun OSS.   
Could you help review it ？  Thanks a lot for your advice ! 😄

huage1994:
The error log of the only failed CI task seems to be not related.

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepo.java
```diff
@@ -17,43 +17,33 @@
...
  * NotebookRepo for Aliyun OSS (https://cn.aliyun.com/product/oss)
  */
-public class OSSNotebookRepo implements NotebookRepo {
+public class OSSNotebookRepo implements NotebookRepoWithVersionControl {
   private static final Logger LOGGER = LoggerFactory.getLogger(OSSNotebookRepo.class);
 
-  private OSS ossClient;
   private String bucketName;
   private String rootFolder;
+  private static int NOTE_MAX_VERSION_NUM;
```
jongyoul, line 43: Do you have any special reason that you make it `static` variable?

jongyoul:
@huage1994 Sorry for the late review but I wondered why you introduced `RemoteStorageOpertaor`. It looks natural to make an interface and its implementation in Java but I feel like it's only for the test object of `MockRemoteStorageOperator` If it does, how about using mockito instead of making MockRemoteOperator? Rest of them, I'm reading and will finish reviewing them by tomorrow.

zeppelin-plugins/notebookrepo/oss/src/test/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepoTest.java
```diff
@@ -0,0 +1,243 @@
...
+      assertEquals(notebookRepo.get(note1.getId(), note1.getPath(), anonymous).getParagraphs().size(), i);
+    }
+
+    // finally revert note to revision 31
+    notebookRepo.setNoteRevision(note1.getId(), note1.getPath(), revisionList.get(OSS_VERSION_MAX).id, anonymous);
+    assertEquals(notebookRepo.get(note1.getId(), note1.getPath(), anonymous).getParagraphs().size(), OSS_VERSION_MAX + 1);
+
+    notebookRepo.remove("/version_control", anonymous);
+  }
+}
```
jongyoul, line 243: nit: Add a newline at the end of the file.

huage1994:
> @huage1994 Sorry for the late review but I wondered why you introduced `RemoteStorageOpertor`. It looks natural to make an interface and its implementation in Java but I feel like it's only for the test object of `MockRemoteStorageOperator` If it does, how about using mockito instead of making MockRemoteOperator? Rest of them, I'm reading and will finish reviewing them by tomorrow.

Hi @jongyoul , thanks a lot for your reviews!
`MockRemoteStorageOperator` is a mock with the local file system to do real test for OSSNotebookRepo.  I think it is more sufficient than using mockito  and it did help me find some problem when running Github CI task.

Additionally  I introduce `RemoteStorageOpertor` because  I think the code in `NotebookRepo` should avoid underlying complex logic of storage operation.  By the way I think `RemoteStorageOpertor`  and `MockRemoteStorageOperator` can be also reused by other Notebook Storage in the future.  Ideally,  we only need to implement `RemoteStorageOpertor`  instead of NotebookRepo.

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepo.java
huage1994: I have deleted `static`.
Firstly I wanted to use final `private final static int NOTE_MAX_VERSION_NUM` , but found `final` cannot be used. Then I delete `final` but didn't delete `static`. 😬

docs/setup/storage/storage.md
```diff
@@ -454,6 +454,16 @@ And you should configure oss related properties in file **zeppelin-site.xml**.
   <description>Access key secret for your OSS account</description>
 </property>
 
+<property>
+  <name>zeppelin.notebook.oss.version.max</name>
+  <value>30</value>
```
jongyoul, line 459: It's just an opinion but how about disabling this feature at first? Otherwise, if we set it to 30 in a non-history folder structure, it would be handled automatically?

zeppelin-plugins/notebookrepo/oss/src/main/java/org/apache/zeppelin/notebook/repo/OSSNotebookRepo.java
```diff
@@ -17,43 +17,33 @@
...
  * NotebookRepo for Aliyun OSS (https://cn.aliyun.com/product/oss)
  */
-public class OSSNotebookRepo implements NotebookRepo {
+public class OSSNotebookRepo implements NotebookRepoWithVersionControl {
   private static final Logger LOGGER = LoggerFactory.getLogger(OSSNotebookRepo.class);
 
-  private OSS ossClient;
   private String bucketName;
   private String rootFolder;
+  private int NOTE_MAX_VERSION_NUM;
```
jongyoul, line 43: nit.
```suggestion
  private int maxVersionNumber;
```

jongyoul (approved):
LGTM except for the minor comments

docs/setup/storage/storage.md
huage1994: Hi @jongyoul , It's ok if we set it to 30 in a non-history folder structure.  Actually non-history folder is a just a special case of history folder because I have handled that in the code.  The new OSSNotebookRepo is completely compatible with this situation

huage1994:
Thanks @jongyoul for the comments, they has been resolved. 
The only failed CI seems to be not related.

jongyoul (approved):

PR #4313, opened 2022-03-15
[ZEPPELIN-5678] Remove the wordcount init job in Flink yarn-application mode

zjffdu:
### What is this PR for?

The cause is that in the `yarn-application`, `StreamExecutionEnvironment#execute(StreamGrpah)` is invoked when calling `TableEnvironment#execute.` But we didn't override it before this PR, so the scala shell jar is not added into the flink job. 
This PR fix this issue by overriding the method `StreamExecutionEnvironment#execute(StreamGrpah)`

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5678

* Ci pass

### Questions:


jongyoul (approved):
LGTM

zjffdu:
Thanks for the review @jongyoul , will merge if no more comment

PR #4279, opened 2021-12-26
[ZEPPELIN-5612] Upgrade junit 4 to 5

huage1994:
### What is this PR for?
1. Upgrade junit 4 to 5, then we can write junit 5 tests and migrate junit 4 tests to 5.
2. Run the existing junit 4 tests on Junit 5 engine.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5612

### How should this be tested?
CI pass

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #4321, opened 2022-03-20
[ZEPPELIN-5612] Upgrade junit 4 to 5

huage1994:
### What is this PR for?
1. Upgrade junit 4 to 5, then we can write junit 5 tests and migrate junit 4 tests to 5.
2. Run the existing junit 4 tests on Junit 5 engine.

3. Refactor `zeppelin-zengine/src/test/java/org/apache/zeppelin/util/UtilTest.java`  with jUnit 5 as an example.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5612

CI pass

### Questions:

### DIFF
--- a/hbase/pom.xml
+++ b/hbase/pom.xml
@@ -75,11 +75,23 @@
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-client</artifactId>
       <version>${hbase.hbase.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-annotations</artifactId>
       <version>${hbase.hbase.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.google.protobuf</groupId>
--- a/hbase/pom.xml
+++ b/hbase/pom.xml
@@ -75,23 +75,11 @@
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-client</artifactId>
       <version>${hbase.hbase.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>junit</groupId>
-          <artifactId>junit</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-annotations</artifactId>
       <version>${hbase.hbase.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>junit</groupId>
-          <artifactId>junit</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.google.protobuf</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -154,7 +154,7 @@
     <grpc.version>1.26.0</grpc.version>
 
     <!-- test library versions -->
-    <junit.version>4.12</junit.version>
+    <junit.jupiter.version>5.7.1</junit.jupiter.version>
     <mockito.version>3.12.4</mockito.version>
     <assertj.version>1.7.0</assertj.version>
     <powermock.version>2.0.9</powermock.version>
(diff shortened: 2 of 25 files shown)


pom.xml
```diff
@@ -204,7 +204,7 @@
     <plugin.scm.version>1.11.2</plugin.scm.version>
     <plugin.shade.version>3.2.2</plugin.shade.version>
     <plugin.source.version>3.2.1</plugin.source.version>
-    <plugin.surefire.version>2.22.2</plugin.surefire.version>
+    <plugin.surefire.version>3.0.0-M5</plugin.surefire.version>
```
zjffdu, line 207: Is this upgrade required by junit 5?

huage1994: Actually 2.22.2 is ok for junit 5 but I found that 3.0.0-M3 fix some bug about junit 5. 
So I upgrade surefire to the latest version.

> [JUnit Runner that writes to System.out corrupts Surefire's STDOUT when using JUnit's Vintage Engine](https://issues.apache.org/jira/browse/SUREFIRE-1614)

zjffdu: Thanks for clarification.

jongyoul:
I think it's good. Is there any pain point I should check?

huage1994:
When I upgrade junit 4 to junit 5 #4279 , Frontend CI job would block at downloading dependency when  `mvn install` the module `k8s-standard`, but it is ok to `mvn package` this module. 
It 's really weird that both `mvn install` and `mvn package` is OK in my local environment, this problem only happens in Github CI jobs . 
It has costed me very much time, but hard to find the real reason.I would really appreciate if someone tell me the reason.

zjffdu:
Not sure what is the root cause, maybe you can try upgrade surefire plugin to 3.x @huage1994

Reamer:
I suspect the checkstyle plugin. This plugin produces a lot of the Maven warnings at the beginning.
I've been wanting to fix this for a long time, but haven't gotten around to it yet.

huage1994:
@zjffdu @Reamer @jongyoul , really thanks for  helping out~ 
I'm going to test your guesses and continue to address this issue

Reamer:
> > I suspect the checkstyle plugin. This plugin produces a lot of the Maven warnings at the beginning.
> > I've been wanting to fix this for a long time, but haven't gotten around to it yet.
> 
> Agreed. We should have better idea to enforce to keep it. I'll think of it. :-)

PR #4466 should improve the CheckStyle integration.

jongyoul:
Thank you. I'm checking the PR now

Reamer:
@huage1994 
My checkstyle pull request https://github.com/apache/zeppelin/pull/4466 was merged with the master. Can you please check if you still get the error when you run `mvn install`.

huage1994:
> @huage1994 My checkstyle pull request #4466 was merged with the master. Can you please check if you still get the error when you run `mvn install`.

Thanks @Reamer !  I'll try that.

huage1994:
Hi @jongyoul  @zjffdu  @Reamer .
I tried your suggestion but the problem still exists.

**I think there are some dependencies in `launcher-k8s-standard` conflict with junit 5**, because Frontend CI runs well if I skip this module [here](https://github.com/huage1994/zeppelin/actions/runs/3218076640).

It's really weird that Frontend CI job would block at downloading dependency only when mvn install the module `launcher-k8s-standard`, but it is ok to mvn package this module `launcher-k8s-standard`. And there is no special plugin goals bound to install phrase. 

Now I'm trying to verify the dependencies in `launcher-k8s-standard` one by one.

hbase/pom.xml
```diff
@@ -75,11 +75,23 @@
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-client</artifactId>
       <version>${hbase.hbase.version}</version>
+      <exclusions>
```
Reamer, line 78: This is not a test dependency, so the exclude should be unnecessary. Normal dependencies do not contain test libraries.

hbase/pom.xml
```diff
@@ -75,11 +75,23 @@
...
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-annotations</artifactId>
       <version>${hbase.hbase.version}</version>
+      <exclusions>
```
Reamer, line 89: This is not a test dependency, so the exclude should be unnecessary. Normal dependencies do not contain test libraries.

pom.xml
```diff
@@ -1898,6 +1911,7 @@
             <plugin>
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
+              <version>${plugin.surefire.version}</version>
```
Reamer, line 1914: The version is already stored here, so this is unnecessary.
https://github.com/apache/zeppelin/blob/aa4731e69916ed20a3aad56a89544d7a08d57d24/pom.xml#L1605-L1609

pom.xml
```diff
@@ -1929,6 +1943,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
+            <version>${plugin.surefire.version}</version>
```
Reamer, line 1946: The version is already stored here, so this is unnecessary.
https://github.com/apache/zeppelin/blob/aa4731e69916ed20a3aad56a89544d7a08d57d24/pom.xml#L1605-L1609

Reamer (changes requested):
I think it looks good overall. Just a few minor issues.

huage1994:
> I think it looks good overall. Just a few minor issues.

Hi @Reamer, really thanks for your useful advice.
I've update my code and waiting for the CI  result.

Reamer (approved):
LGTM

jongyoul (approved):

PR #4323, opened 2022-03-21
[ZEPPELIN-5680] Remove spark-dependencies module

zjffdu:
### What is this PR for?

This PR removed module `spark-dependencies` and also update `interpreter.sh`
 
### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5680

* CI pass

### Questions:


jongyoul:
I'm just curious that this means we need to download spark manually if we would like to use Spark Interpreter?

jongyoul:
@zjffdu agreed. In an early stage, it was needed for expanding our community by providing easiest way to use Zeppelin itself but thing's are changed. BTW, we also need to change our guide if we remove it, don't we?

jongyoul:
LGTM especially, it will help us to reduce whole build time as well.

By the way, if we forgive embedded mode, can we remove downloading python and R as well? We will replace them by setting paths based on SPARK_HOME. WDYT?

zjffdu:
Good point @jongyoul  let me update that.

zjffdu:
@jongyoul After looking at the code, I realise that downloading `pyspark` and `sparkr` can not be removed, they are required for unit test.

jongyoul:
> @jongyoul After looking at the code, I realise that downloading pyspark and sparkr can not be removed, they are required for unit test.

I have an idea about it. In my understanding, we are downloading Spark binary while running CI. Can we replace it? I'll also check it :-)

zjffdu:
I don't think so, we should enable user to run unit test locally.

jongyoul:
@zjffdu Yes, correct. Agreed. Then, does the download work only for testing? I just wonder if it's possible to reduce our whole binary or not. If it can, we'd better consider it sincerely.

zjffdu:
@jongyoul I mean the downloading in this maven pom.xml (https://github.com/apache/zeppelin/blob/master/spark/interpreter/pom.xml#L355), it is only for unit test. It won't increase the zeppelin binary distribution as long as we remove spark-dependencies module

jongyoul:
@zjffdu Thank you for the clarification. LGTM

jongyoul (approved):

jongyoul:
BTW, I've checked the CI but the error is ...
```
ImportError: cannot import name 'Markup' from 'jinja2' (/usr/share/miniconda/envs/python_3_with_R/lib/python3.7/site-packages/jinja2/__init__.py) expected:<SUCCESS> but was:<ERROR>
```
Is it good?

jongyoul:
The error above might be related to #4331

zjffdu:
That's right @jongyoul Now, CI is passed

PR #4332, opened 2022-03-25
[ZEPPELIN-5668] Support statement set in flink

zjffdu:
### What is this PR for?

This PR is to support statement set with the same syntax of flink sql-client:
```
begin statement set;
...
end
```
as an alternative of `runAsOne`

### What type of PR is it?
[Feature | Documentation ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5668

### Questions:

### DIFF
--- a/flink/flink1.12-shims/pom.xml
+++ b/flink/flink1.12-shims/pom.xml
@@ -176,7 +176,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-<!--                        <arg>-target:jvm-1.8</arg>-->
+                        <arg>-target:jvm-1.8</arg>
                     </args>
                     <jvmArgs>
                         <jvmArg>-Xms1024m</jvmArg>
--- a/flink/flink1.12-shims/pom.xml
+++ b/flink/flink1.12-shims/pom.xml
@@ -176,7 +176,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-                        <arg>-target:jvm-1.8</arg>
+<!--                        <arg>-target:jvm-1.8</arg>-->
                     </args>
                     <jvmArgs>
                         <jvmArg>-Xms1024m</jvmArg>
--- a/flink/flink1.13-shims/pom.xml
+++ b/flink/flink1.13-shims/pom.xml
@@ -178,7 +178,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-<!--                        <arg>-target:jvm-1.8</arg>-->
+                        <arg>-target:jvm-1.8</arg>
                     </args>
                     <jvmArgs>
                         <jvmArg>-Xms1024m</jvmArg>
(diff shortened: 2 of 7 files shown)


flink/flink1.13-shims/pom.xml
```diff
@@ -178,7 +178,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-                        <arg>-target:jvm-1.8</arg>
+<!--                        <arg>-target:jvm-1.8</arg>-->
```
jongyoul, line 181: Same here

flink/flink1.14-shims/pom.xml
```diff
@@ -171,7 +171,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-                        <arg>-target:jvm-1.8</arg>
+<!--                        <arg>-target:jvm-1.8</arg>-->
```
jongyoul, line 174: Same here

jongyoul (approved):
LGTM but I have an opinion. First of all, I'm not an expert on Flink so you can freely tell me where I might misunderstand it. I saw the `Flink113SqlInterpreter` and felt like `statementModeMap` is acting like a pair of `statementOperationMap`. I think we can change the logic like check if `statementOperationMap.get(context.getParagraphId()) == null` or not. If it's null, it look same as `statementModeMap.get(context.getParagraphId()) == false`. If it's true, you'd better remove `statementModeMap` and add logic to check nullability of `statmentOperationsMap.get(...) == null`. Am I correct? :-)

Regardless of the opinion above, It looks good to me :-)

flink/flink1.12-shims/pom.xml
```diff
@@ -176,7 +176,7 @@
                         <arg>-unchecked</arg>
                         <arg>-deprecation</arg>
                         <arg>-feature</arg>
-                        <arg>-target:jvm-1.8</arg>
+<!--                        <arg>-target:jvm-1.8</arg>-->
```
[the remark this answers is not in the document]
jongyoul:
```suggestion
```

flink/flink1.13-shims/pom.xml
jongyoul:
```suggestion
```

flink/flink1.14-shims/pom.xml
jongyoul:
```suggestion
```

flink/flink1.12-shims/pom.xml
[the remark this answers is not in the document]
zjffdu: It is about another issue, let me revert it and do it in another PR.

zjffdu:
> LGTM but I have an opinion. First of all, I'm not an expert on Flink so you can freely tell me where I might misunderstand it. I saw the `Flink113SqlInterpreter` and felt like `statementModeMap` is acting like a pair of `statementOperationMap`. I think we can change the logic like check if `statementOperationMap.get(context.getParagraphId()) == null` or not. If it's null, it look same as `statementModeMap.get(context.getParagraphId()) == false`. If it's true, we'd better remove `statementModeMap` and add logic to check nullability of `statmentOperationsMap.get(...) == null`. Am I correct? :-)
> 
> Regardless of the opinion above, It looks good to me :-)

You are right @jongyoul  I have updated it.

PR #4339, opened 2022-04-02
[DRAFR] [ZEPPELIN-5703] Support to run another notebook in notebook inline

zjffdu:
### What is this PR for?

### What type of PR is it?

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSettingManager.java
```diff
@@ -975,6 +975,9 @@ public void restart(String settingId, String user, String noteId) throws Interpr
 
   // restart in note page
   public void restart(String settingId, ExecutionContext executionContext) throws InterpreterException {
+    if (settingId == null || settingId.startsWith("__internal__")) {
```
jongyoul, line 978: How about adding comment to describe `__internal__`?
```suggestion
    // __internal__ is set by RunNotebookInterpreter
    if (settingId == null || settingId.startsWith("__internal__")) {
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Notebook.java
```diff
@@ -37,13 +37,7 @@
...
 import org.apache.zeppelin.display.AngularObject;
 import org.apache.zeppelin.display.AngularObjectRegistry;
-import org.apache.zeppelin.interpreter.Interpreter;
-import org.apache.zeppelin.interpreter.InterpreterFactory;
-import org.apache.zeppelin.interpreter.InterpreterGroup;
-import org.apache.zeppelin.interpreter.InterpreterNotFoundException;
-import org.apache.zeppelin.interpreter.InterpreterSetting;
-import org.apache.zeppelin.interpreter.InterpreterSettingManager;
-import org.apache.zeppelin.interpreter.ManagedInterpreterGroup;
+import org.apache.zeppelin.interpreter.*;
```
jongyoul, line 40: I'm not sure about the current style rule but might be replaced by IDE automatically?

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/RunNotebookInterpreter.java
```diff
@@ -0,0 +1,294 @@
...
+      } else {
+        return new InterpreterResult(InterpreterResult.Code.ERROR, allResultMessages);
+      }
+    } catch (IOException e) {
+      return new InterpreterResult(InterpreterResult.Code.ERROR, ExceptionUtils.getStackTrace(e));
+    }
+  }
+
+  @VisibleForTesting
+  public static Map<String, String> parseParams(String text) throws InvalidParameterFormatException {
```
jongyoul, line 171: Just curious. Why don't we use regex to handle it? It also handle some exceptional cases but I wonder if we need to handle that case or not.

jongyoul (approved):
LGTM except for some trivial comments.

jongyoul:
@zjffdu PR itself looks good to me but I have a question about this idea itself. We had a plan to provide job manager which handle DAG for paragraphs but it's not proceeded well. You PR looks a different way to handle a similar request. Then, can we remote job manager on Zeppelin? It shows some status but it's not that helpful even the quest is too complicated. WDYT?

PR #4345, opened 2022-04-07
[ZEPPELIN-5697] Remove old planner support of flink

zjffdu:
### What is this PR for?

The old planner is removed starting from Flink 1.14. It is only used by a few users in some uncommon scenarios. This PR is to remove it from Flink interpreter to reduce maintenance effort.

### What type of PR is it?
[ Improvement ]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5697

* CI 

### Questions:


zjffdu:
CI is passed

PR #4347, opened 2022-04-11
[ZEPPELIN-4627] Fix class loader inconsistency between driver and executor when running on Spark 3.x

Kontinuation:
### What is this PR for?
Setting thread context classloader of interpreter thread before creating SparkContext object in order to preserve correct classloader hierarchy.

We worked out this patch for fixing ZEPPELIN-4627, and it also seems to fix ZEPPELIN-5688.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-4627
* https://issues.apache.org/jira/browse/ZEPPELIN-5688

* Manually tested cases described in ZEPPELIN-4627
* Added unit test for cases described in ZEPPELIN-5688

### Questions:


zjffdu:
LGTM，thanks @Kontinuation

jongyoul (approved):

PR #4360, opened 2022-04-17
[ZEPPELIN-5614] A JDBC interpreter property for disable setting tag for application of hive engines

huage1994:
### What is this PR for?
This PR introduce a JDBC interpreter property for disable setting tag for application of hive engines.

Some users disabled tez.* property in hive  and hence unable to use hive in zeppelin. The reason is Zeppelin jdbc interpreter always set both tez and MR tags property when connect to hive.   This PR provide a property to disable setting tag for tez and MR application in hive.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5614

* CI passed

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -604,6 +604,16 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
   // only add tags for hive jdbc
   private String appendTagsToURL(String url, InterpreterContext context) {
+    // To deal engines which don't need to set application tag
+    String disabledAppTypeStr = getProperty("zeppelin.jdbc.hive.tagged.engines.exclude");
+    HashSet<String> disabledAppType = new HashSet<>();
+    if (StringUtils.isNotEmpty(disabledAppTypeStr)) {
+      disabledAppType = new HashSet<>(Arrays.asList(disabledAppTypeStr.split(",")));
+      if ((disabledAppType.contains("MAPREDUCE") && disabledAppType.contains("TEZ"))) {
+        return url;
+      }
+    }
+
     StringBuilder builder = new StringBuilder(url);
     if (url.startsWith("jdbc:hive2:")) {
       Integer lastIndexOfQMark = builder.indexOf("?");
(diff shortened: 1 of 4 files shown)


jdbc/src/main/resources/interpreter-setting.json
```diff
@@ -143,6 +143,13 @@
         "defaultValue": "1000",
         "description": "Query interval for hive statement",
         "type": "number"
+      },
+      "zeppelin.jdbc.hive.tagged.engines.exclude": {
```
zjffdu, line 147: I think this could be a boolean property, it is not necessary to differentiate mr or tez, because the user can switch engine easily.

huage1994: Get it ! 😀

huage1994: Done. 
By the way I fixed a problem about the type of  default value of   `"zeppelin.jdbc.auth.kerberos.proxy.enable"`.
If it is string `"true"`  instead of  boolean `true`,  the checkbox in UI would not ✅ .

zjffdu: Also need to update `jdbc.md` for this new property.

huage1994: > Also need to update `jdbc.md` for this new property.

Hi @zjffdu , This new property is only related to hive, should I put it into `hive.md` ?   
I didn't find a good place for this property in `jdbc.md`.

zjffdu: Both places should update. There's one section about hive in `jdbc.md`, https://zeppelin.apache.org/docs/0.10.1/interpreter/jdbc.html#apache-hive. BTW, it looks like to we need to converge the hive related doc in these 2 places. But it could be done in another PR.

huage1994: Hi @zjffdu , I have put all hive-related properties into `hive.md` and add a link to `hive.md` in `jdbc.md`.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -604,6 +604,10 @@ private String appendProxyUserToURL(String url, String user, String propertyKey)
 
   // only add tags for hive jdbc
   private String appendTagsToURL(String url, InterpreterContext context) {
+    if (!Boolean.parseBoolean(getProperty("zeppelin.jdbc.hive.engines.tag.enable"))) {
```
zjffdu, line 607: set default value as `true` in case `zeppelin.jdbc.hive.engines.tag.enable` is not set when user upgrade zeppelin from old versions

huage1994: > set default value as `true` in case `zeppelin.jdbc.hive.engines.tag.enable` is not set when user upgrade zeppelin from old versions

Good point!

zjffdu (approved):

zjffdu:
Will merge if no more comment

PR #4374, opened 2022-05-28
[ZEPPELIN-5744] Allow NoteManager for concurrent operation

saLeox:
### What is this PR for?
When multi users perform some notebook/folder operation, like rename or move, some notebooks become not accessible from zeppelin page, while they are still existing on the HDFS, and only resort to restart Zeppelin daemon to solve the problem.

In this PR, unit test is created to simulate the concurrent request scenario and reproduce the issue successfully, including the operations of saveNote, moveNote, moveFolder and removeNote.

Accordingly, some changes are made to ensure thread safety of note manager operation, so that the note or folder info will not be lost, after which the unit test should get passed.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5744

* Automated unit tests for the change is added.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
@@ -79,7 +80,7 @@ public class NoteManager {
   // build the tree structure of notes
   private void init() throws IOException {
     this.notesInfo = notebookRepo.list(AuthenticationInfo.ANONYMOUS).values().stream()
-        .collect(Collectors.toMap(NoteInfo::getId, NoteInfo::getPath));
+        .collect(Collectors.toConcurrentMap(NoteInfo::getId, NoteInfo::getPath));
     this.notesInfo.entrySet().stream()
         .forEach(entry ->
         {
@@ -181,7 +182,11 @@ public class NoteManager {
     } else {
       addOrUpdateNoteNode(new NoteInfo(note));
       noteCache.putNote(note);
-      this.notebookRepo.save(note, subject);
+      // Make sure to execute `notebookRepo.save()` successfully in concurrent context
+      // Otherwise, the NullPointerException will be thrown when invoking notebookRepo.get() in the following operations.
+      synchronized (this) {
+        this.notebookRepo.save(note, subject);
+      }
     }
   }
 
@@ -218,7 +223,6 @@ public class NoteManager {
   public void moveNote(String noteId,
                        String newNotePath,
                        AuthenticationInfo subject) throws IOException {
-    String notePath = this.notesInfo.get(noteId);
     if (noteId == null) {
       throw new IOException("No metadata found for this note: " + noteId);
     }
@@ -228,6 +232,7 @@ public class NoteManager {
     }
 
     // move the old NoteNode from notePath to newNotePath
+    String notePath = this.notesInfo.get(noteId);
     NoteNode noteNode = getNoteNode(notePath);
     noteNode.getParent().removeNote(getNoteName(notePath));
     noteNode.setNotePath(newNotePath);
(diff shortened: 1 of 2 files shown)


jongyoul (approved):

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NoteManagerTest.java
```diff
@@ -158,4 +166,117 @@ public void testLruCache() throws IOException {
...
+
+    @Override
+    public void run(int index) throws IOException {
+      String curPath = "/dev/project_" + index, tarPath = "/staging/note_" + index;
+      noteManager.moveFolder(curPath, tarPath, AuthenticationInfo.ANONYMOUS);
+    }
+  }
+
+  class ConcurrentTaskRemoveNote extends ConcurrentTask {
+    public ConcurrentTaskRemoveNote(ExecutorService threadPool, int noteNum, Map notes, String pathPattern) {
```
Reamer, line 272:
```suggestion
    public ConcurrentTaskRemoveNote(ExecutorService threadPool, int noteNum, Map<Integer, String> notes, String pathPattern) {
```

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NoteManagerTest.java
```diff
@@ -158,4 +166,117 @@ public void testLruCache() throws IOException {
...
+
+    @Override
+    public void run(int index) throws IOException {
+      String tarPath = String.format(super.pathPattern, index, index);
+      noteManager.moveNote(super.notes.get(index), tarPath, AuthenticationInfo.ANONYMOUS);
+    }
+  }
+
+  class ConcurrentTaskMoveFolder extends ConcurrentTask {
+    public ConcurrentTaskMoveFolder(ExecutorService threadPool, int noteNum, Map notes, String pathPattern) {
```
Reamer, line 260:
```suggestion
   public ConcurrentTaskMoveFolder(ExecutorService threadPool, int noteNum, Map<Integer, String> notes, String pathPattern) {
```

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NoteManagerTest.java
```diff
@@ -158,4 +166,117 @@ public void testLruCache() throws IOException {
...
+        threadPool.execute(() -> {
+          try {
+            this.run(index);
+            latch.countDown();
+          } catch (IOException e) {
+            e.printStackTrace();
+          }
+        });
+      }
+      latch.await(); //wait till all tasks are completed
```
Reamer, line 219:
```suggestion
      assertTrue(latch.await(5, TimeUnit.SECONDS)); //wait till all tasks are completed
```
An additional comment at this point would be good.

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/NoteManagerTest.java
```diff
@@ -158,4 +166,117 @@ public void testLruCache() throws IOException {
...
+    threadPool.shutdown();
+  }
+
+  abstract class ConcurrentTask {
+    private ExecutorService threadPool;
+    private int noteNum;
+    private Map<Integer, String> notes;
+    private String pathPattern;
+
+    public ConcurrentTask(ExecutorService threadPool, int noteNum, Map notes, String pathPattern) {
```
Reamer, line 196:
```suggestion
    public ConcurrentTask(ExecutorService threadPool, int noteNum, Map<Integer, String> notes, String pathPattern) {
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
```diff
@@ -181,7 +182,9 @@ public void saveNote(Note note, AuthenticationInfo subject) throws IOException {
     } else {
       addOrUpdateNoteNode(new NoteInfo(note));
       noteCache.putNote(note);
-      this.notebookRepo.save(note, subject);
+      synchronized (this) {
```
Reamer, line 185: A comment at this point why a `synchronized` has to be used here would be very helpful.

Reamer (changes requested):
Overall, it looks really good. Just a few small suggestions. Please adjust the imports.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
saLeox: The reason makes it `synchronized` is that the `note` object will be empty in the case of concurrent saveNote, which comes from following code:
```note = notebookRepo.get(noteInfo.getId(), noteInfo.getPath(), AuthenticationInfo.ANONYMOUS);```
And since it's just an abstract function, I just add `synchronized` to make it return non-empty object.

Reamer: Please add this as a comment in the code so it is clear to other contributors.

PR #4375, opened 2022-05-31
ZEPPELIN-5746: Add support for escaping html in markdown

PeterAlfredLee:
This is implemented by flexmark. A new configuration is also added, and
this configuration is turned on by default.

### What is this PR for?
Zeppelin should support escaping html in markdown.

### What type of PR is it?
* Bug Fix

### What is the Jira issue?
* [ZEPPELIN-5746](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-5746)

* CI

### Questions:

I also noticed there is another call of flexmark in `org.apache.zeppelin.jupyter.parser.MarkdownParser`. But it seems the package `jupyter` can not read from configuration. I am not sure if this occurrence should be fixed or not. Feel free to tell me if this need to be fixed. :)


jongyoul (approved):

Reamer (approved):
LGTM

PR #4390, opened 2022-06-30
[ZEPPELIN-5757] Assume role in NotebookRepo S3

zlosim:
### What is this PR for?
Currently NotebookRepo S3 is not able to assume role from env variables due to missing aws sts sdk
- added missing dependency
- fixed documentation

### What type of PR is it?
Bug Fix
Improvement

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5757

### Questions:
* Does this needs documentation? just minor fix, docs updated with this pull request

### DIFF
--- a/zeppelin-plugins/notebookrepo/s3/pom.xml
+++ b/zeppelin-plugins/notebookrepo/s3/pom.xml
@@ -34,7 +34,7 @@
     <description>NotebookRepo implementation based on S3</description>
 
     <properties>
-        <aws.sdk.s3.version>1.11.736</aws.sdk.s3.version>
+        <aws.sdk.version>1.11.736</aws.sdk.version>
         <plugin.name>NotebookRepo/S3NotebookRepo</plugin.name>
     </properties>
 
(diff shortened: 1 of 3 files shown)


zeppelin-plugins/notebookrepo/s3/pom.xml
```diff
@@ -34,15 +34,28 @@
...
+            <version>${aws.sdk.version}</version>
+            <exclusions>
+                <!-- jcl-over-slf4j is provided by zeppelin-interprerter -->
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
```
Reamer, line 54: Since you are not using the library in your code, you should leave a comment why it is needed.

Reamer (changes requested):
Is the functionality given by adding the library?
Do you see a way to validate the function with a unit test?

zeppelin-plugins/notebookrepo/s3/pom.xml
zlosim: sure, explanation added

zlosim:
@Reamer thanks for the suggestions, as for the questions:
- Yes, without the lib WebIdentityTokenCredentialsProvider [will fail](https://github.com/aws/aws-sdk-java/blob/9cb8d4b7d7d65e42f8a8a1f1b277ad9d6744ef76/aws-java-sdk-core/src/main/java/com/amazonaws/auth/WebIdentityTokenCredentialsProvider.java#L33) so DefaultAWSCredentialsProviderChain will skip it and move to the next provider 
- Not really sure, one thing comes to my mind is to try creating S3 client with WebIdentityTokenCredentialsProvider and check if given provider fails with class not found exception - you think this could be enough?
- license file already contains info for [AWS SDKs ](https://github.com/apache/zeppelin/blob/bd951ee53d696a4d21a69f09f5cd7ba76f173a28/zeppelin-distribution/src/bin_license/LICENSE#L5)

jongyoul (approved):

Reamer:
> Not really sure, one thing comes to my mind is to try creating S3 client with WebIdentityTokenCredentialsProvider and check if given provider fails with class not found exception - you think this could be enough?

Yes such a small unit test should be sufficient.

zlosim:
> Yes such a small unit test should be sufficient.

Thanks @Reamer 
I found checking directly for the class from STS lib less error prone, if that is good enough :)

Reamer (approved):
LGTM

PR #4398, opened 2022-07-07
[ZEPPELIN-5760] fix passing configs to spark in k8s

zlosim:
### What is this PR for?
Some important spark configs can`t be passed via SparkConf in k8s and in client mode including spark.jars.packages and spark.driver.extraJavaOptions
This fix checks if these are set in interpreter configuration and passes them to spark-submit command 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5760

updated test to test this fix

### Questions:
* Is there breaking changes for older versions?  No

### DIFF
--- a/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
@@ -21,16 +21,12 @@ package org.apache.zeppelin.interpreter.launcher;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ArrayUtils;
-import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreterManagedProcess;
 import org.apache.zeppelin.interpreter.remote.RemoteInterpreterServer;
@@ -78,16 +74,8 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
   private static final String SPARK_CONTAINER_IMAGE = "zeppelin.k8s.spark.container.image";
   private static final String ENV_SERVICE_DOMAIN = "SERVICE_DOMAIN";
   private static final String ENV_ZEPPELIN_HOME = "ZEPPELIN_HOME";
-  private static final String SPARK_DRIVER_CLASSPATH = "spark.driver.extraClassPath";
   private static final String SPARK_DRIVER_DEFAULTJAVAOPTS = "spark.driver.defaultJavaOptions";
-  private static final String SPARK_DRIVER_EXTRALIBPATH = "spark.driver.extraLibraryPath";
   private static final String SPARK_DRIVER_EXTRAJAVAOPTS = "spark.driver.extraJavaOptions";
-  private static final String SPARK_JARS = "spark.jars";
-  private static final String SPARK_JARS_PACKAGES = "spark.jars.packages";
-  private static final String SPARK_JARS_EXCLUDES = "spark.jars.excludes";
-  private static final String SPARK_JARS_IVY = "spark.jars.ivy";
-  private static final String SPARK_JARS_IVYSETTINGS ="spark.jars.ivySettings" ;
-  private static final String SPARK_JARS_REPOSITORIES = "spark.jars.repositories";
 
   public K8sRemoteInterpreterProcess(
           KubernetesClient client,
@@ -329,26 +317,28 @@ public class K8sRemoteInterpreterProcess extends RemoteInterpreterManagedProcess
     if (isSpark()) {
       int webUiPort = 4040;
       k8sProperties.put("zeppelin.k8s.spark.container.image", sparkImage);
-      if (isSparkOnKubernetes(properties)) {
-        k8sEnv.put("SPARK_SUBMIT_OPTIONS",
-            getEnv().getOrDefault("SPARK_SUBMIT_OPTIONS", "") + buildSparkSubmitOptions(userName));
-      }
-      k8sEnv.put("SPARK_HOME", getEnv().getOrDefault("SPARK_HOME", "/spark"));
 
       // There is already initial value following --driver-java-options added in interpreter.sh
       // so we need to pass spark.driver.defaultJavaOptions and spark.driver.extraJavaOptions
       // as SPARK_DRIVER_EXTRAJAVAOPTIONS_CONF env variable to build spark-submit command correctly.
       StringBuilder driverExtraJavaOpts = new StringBuilder();
       if (properties.containsKey(SPARK_DRIVER_DEFAULTJAVAOPTS)) {
-        driverExtraJavaOpts.append((String) properties.get(SPARK_DRIVER_DEFAULTJAVAOPTS));
+        driverExtraJavaOpts.append((String) properties.remove(SPARK_DRIVER_DEFAULTJAVAOPTS));
       }
       if (properties.containsKey(SPARK_DRIVER_EXTRAJAVAOPTS)) {
         driverExtraJavaOpts.append(" ");
-        driverExtraJavaOpts.append((String) properties.get(SPARK_DRIVER_EXTRAJAVAOPTS));
+        driverExtraJavaOpts.append((String) properties.remove(SPARK_DRIVER_EXTRAJAVAOPTS));
       }
       if (driverExtraJavaOpts.length() > 0) {
         k8sEnv.put("SPARK_DRIVER_EXTRAJAVAOPTIONS_CONF", driverExtraJavaOpts.toString());
       }
+
+      if (isSparkOnKubernetes(properties)) {
+        this.addSparkK8sProperties();
+        k8sEnv.put("ZEPPELIN_SPARK_CONF", prepareZeppelinSparkConf(userName));
+      }
+      k8sEnv.put("SPARK_HOME", getEnv().getOrDefault("SPARK_HOME", "/spark"));
+
       // configure interpreter property "zeppelin.spark.uiWebUrl" if not defined, to enable spark ui through reverse proxy
       String webUrl = (String) properties.get("zeppelin.spark.uiWebUrl");
       if (StringUtils.isBlank(webUrl)) {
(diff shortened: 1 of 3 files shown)


Reamer:
I think most Spark configuration values can be included in the environment variable `SPARK_SUBMIT_OPTIONS`. Please note that the K8sProcess works for all Zeppelin interpreters.

zlosim:
Hi @Reamer  I'm sorry but I am not sure if I understand you correctly. 
As you stated, spark configs can be included in the `SPARK_SUBMIT_OPTIONS ` but they were NOT. This fix specifically adds these configs to `SPARK_SUBMIT_OPTIONS`
there are also two spark configs that CAN'T be passed in `SPARK_SUBMIT_OPTIONS` because of some functionality happening in 'interpreter.sh' thus they are passed in 'SPARK_DRIVER_EXTRAJAVAOPTIONS_CONF' 

also this fix only changes behavior for spark only

Reamer:
I understand your motivation. I just don't like the list of Spark configuration at this point, because it can become very long. Spark has very many configuration options and we can't map them all again here in Zeppelin. I am looking for a more generic approach.
@zjffdu  If `SPARK_SUBMIT_OPTIONS` is deprecated, how can we fill in the function?

zjffdu:
@zlosim [SparkInterpreterLauncher](https://github.com/apache/zeppelin/blob/c4c580a37fde649553d336984a94bcb1b2821201/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L80) supports all the spark configurations, I think we can do similar thing in `K8sRemoteInterpreterProcess`

zjffdu:
@zlosim [SparkInterpreterLauncher](https://github.com/apache/zeppelin/blob/c4c580a37fde649553d336984a94bcb1b2821201/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L80) supports all the spark configurations. It detects all the spark configuration and concatenate them via env variable `ZEPPELIN_SPARK_CONF`.
I think we can do similar thing in `K8sRemoteInterpreterProcess`.

The reason why we deprecate `SPARK_SUBMIT_OPTIONS` is we would like to encourage user to configure spark in Zeppelin via a unified approach like spark (just configure each property one by one), otherwise it may cause inconsistency (e.g. Zeppelin admin may configure `spark.driver.memory` via `SPARK_SUBMIT_OPTIONS` in `zeppelin-env.sh`,  but another user also define it via inline configuration, It is hard for Zeppelin to detect this kind of conflicts).  So even for `ZEPPELIN_SPARK_CONF` in `SparkInterpreterLauncher`, it is for internal usage, it is not exposed to zeppelin users.

zlosim:
Hi @zjffdu thanks for the response.
To be honest, I think this fix is doing exactly what you are saying you want to achieve. Right now setting e.g. [spark dependencies via zeppelin interpreter settings](https://i.imgur.com/VeSvIFa.png) is not working in K8s because we are trying to set it in runtime which is too late as driver is already started. This PR picks settings from interpreter settings passed by user that have to be passed directly to spark submit command when creating k8s pod with driver and puts them in env var so interpreter.sh script will start driver correctly. For end user this is what we are trying to achieve - he sets his settings via zeppelin interepreter settings and everything works as expected. 
As for `SPARK_SUBMIT_OPTIONS` and possibility of zeppelin admin alter configs - [we can alter the append and ignore value](https://github.com/apache/zeppelin/blob/ed251f4b48215f372f437f4490d7f07680793b7b/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java#L324) effectively leaving `SPARK_SUBMIT_OPTIONS` unexposed to zeppelin users - same as `ZEPPELIN_SPARK_CONF`.

zjffdu:
@zlosim How about updating [buildSparkSubmitOptions](https://github.com/apache/zeppelin/blob/master/zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java#L406)
to what we did in [SparkInterpreterLauncher](https://github.com/apache/zeppelin/blob/master/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L203). Just collect all the spark properties (start with prefix `spark.`) and concatenate them into `ZEPPELIN_SPARK_CONF` and pass it to `interpreter.sh`, instead of collecting each individual spark driver property explicitly.

zlosim:
Hi @zjffdu , sorry for the late response. I rewrote it so we are now using ZEPPELIN_SPARK_CONF. Please let me know if there is anything I can do better.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -21,16 +21,12 @@
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
```
jongyoul, line 24: nit. Could you please change your IDE setting not to use `*`?

zlosim: good point :) fixed

Reamer (changes requested):
I think I have found a bug. Have you tested the changes locally?

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -403,28 +421,35 @@ boolean isSparkOnKubernetes(Properties interpreterProperties) {
...
-  String buildSparkSubmitOptions(String userName) {
-    StringBuilder options = new StringBuilder();
-
-    options.append(" --master k8s://https://kubernetes.default.svc");
-    options.append(" --deploy-mode client");
-    if (properties.containsKey(SPARK_DRIVER_MEMORY)) {
-      options.append(" --driver-memory ").append(properties.get(SPARK_DRIVER_MEMORY));
-    }
+  String prepareZeppelinSparkConf(String userName) {
+    StringJoiner sparkConfSJ = new StringJoiner(" ");
```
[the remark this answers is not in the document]
zlosim: Hi @Reamer, you are right, got my builds messed up when testing, sorry about that. I have just rebuild everything from scratch and tested everything to confirm everything works as expected

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -403,28 +421,35 @@ boolean isSparkOnKubernetes(Properties interpreterProperties) {
...
-    options.append(" --conf spark.driver.port=").append(getSparkDriverPort());
-    options.append(" --conf spark.blockManager.port=").append(getSparkBlockManagerPort());
-
-    return options.toString();
+
+    for (String key : properties.stringPropertyNames()) {
+      String propValue = properties.getProperty(key);
+      if (isSparkConf(key, propValue)) {
+        sparkConfSJ.add("--conf");
+        sparkConfSJ.add(key + "=" + propValue + "");
```
Reamer, line 435: Why `""` at the end?

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -319,9 +321,25 @@ Properties getTemplateBindings(String userName) {
...
+
+      // There is already initial value following --driver-java-options added in interpreter.sh
+      // so we need to pass spark.driver.defaultJavaOptions and spark.driver.extraJavaOptions
+      // as SPARK_DRIVER_EXTRAJAVAOPTIONS_CONF env variable to build spark-submit command correctly.
+      StringBuilder driverExtraJavaOpts = new StringBuilder();
+      if (properties.containsKey(SPARK_DRIVER_DEFAULTJAVAOPTS)) {
+        driverExtraJavaOpts.append((String) properties.remove(SPARK_DRIVER_DEFAULTJAVAOPTS));
+      }
+      if (properties.containsKey(SPARK_DRIVER_EXTRAJAVAOPTS)) {
+        driverExtraJavaOpts.append(" ");
```
jongyoul, line 333: nit. You can use `StringJoiner(" ")` to reduce `.append(" ");`

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -403,28 +421,35 @@ boolean isSparkOnKubernetes(Properties interpreterProperties) {
...
-  String buildSparkSubmitOptions(String userName) {
-    StringBuilder options = new StringBuilder();
-
-    options.append(" --master k8s://https://kubernetes.default.svc");
-    options.append(" --deploy-mode client");
-    if (properties.containsKey(SPARK_DRIVER_MEMORY)) {
-      options.append(" --driver-memory ").append(properties.get(SPARK_DRIVER_MEMORY));
-    }
+  String prepareZeppelinSparkConf(String userName) {
+    StringJoiner sparkConfSJ = new StringJoiner("|");
```
jongyoul, line 425: Sorry for the fundamental question, but I wonder if `|` is correct?

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -319,9 +321,25 @@ Properties getTemplateBindings(String userName) {
...
+        driverExtraJavaOpts.append((String) properties.remove(SPARK_DRIVER_EXTRAJAVAOPTS));
+      }
+      if (driverExtraJavaOpts.length() > 0) {
+        k8sEnv.put("SPARK_DRIVER_EXTRAJAVAOPTIONS_CONF", driverExtraJavaOpts.toString());
+      }
+
       if (isSparkOnKubernetes(properties)) {
-        k8sEnv.put("SPARK_SUBMIT_OPTIONS",
-            getEnv().getOrDefault("SPARK_SUBMIT_OPTIONS", "") + buildSparkSubmitOptions(userName));
+        this.addSparkK8sProperties();
```
jongyoul, line 341:
```suggestion
        addSparkK8sProperties();
```

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
Reamer: Yes `|` is correct. Take a look into https://github.com/apache/zeppelin/blob/4e9cab538e75faa9f4168143f1fdd25c27abedb2/bin/interpreter.sh#L259
and the discussion https://github.com/apache/zeppelin/pull/4127.

jongyoul: Aha ~ Thank you!

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
zlosim: fixed

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
zlosim: fixed

zlosim:
Hi @zjffdu , I think this PR is ready to merge unless there are any other requests

Reamer:
I would like to test this beforehand in my local K8s cluster, but I am still on vacation right now. Please have a little patience.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -403,28 +420,35 @@ boolean isSparkOnKubernetes(Properties interpreterProperties) {
...
-    options.append(" --deploy-mode client");
-    if (properties.containsKey(SPARK_DRIVER_MEMORY)) {
-      options.append(" --driver-memory ").append(properties.get(SPARK_DRIVER_MEMORY));
-    }
+  String prepareZeppelinSparkConf(String userName) {
+    StringJoiner sparkConfSJ = new StringJoiner("|");
     if (isUserImpersonated() && !StringUtils.containsIgnoreCase(userName, "anonymous")) {
-      options.append(" --proxy-user ").append(userName);
+      sparkConfSJ.add("--proxy-user");
+      sparkConfSJ.add(userName);
```
Reamer, line 427: Possible error.
This creates "--proxy-user|user". desired is "--proxy-user|user"

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
```diff
@@ -403,28 +420,35 @@ boolean isSparkOnKubernetes(Properties interpreterProperties) {
...
-    options.append(" --conf spark.driver.host=").append(getInterpreterPodDnsName());
-    options.append(" --conf spark.driver.port=").append(getSparkDriverPort());
-    options.append(" --conf spark.blockManager.port=").append(getSparkBlockManagerPort());
-
-    return options.toString();
+
+    for (String key : properties.stringPropertyNames()) {
+      String propValue = properties.getProperty(key);
+      if (isSparkConf(key, propValue)) {
+        sparkConfSJ.add("--conf");
```
Reamer, line 433: Here the same error.
desired is "--conf spark=123|--conf spark.1=123"

Reamer (commented):
I think I have found a mistake. Let me check this. The purpose of the pipe is that the individual config values can be cleanly separated from other config values. A separation in the config values itself is not desired.

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
zlosim: Hi @Reamer I`m not sure I can see the bug here, right now the output of this snippet is `--proxy-user|user` and the pipe would be replaced in interpreter.sh with space
as far as i can see same code snippet can be found in [SparkInterpreterLauncher](https://github.com/apache/zeppelin/blob/c4c580a37fde649553d336984a94bcb1b2821201/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L196)

zeppelin-plugins/launcher/k8s-standard/src/main/java/org/apache/zeppelin/interpreter/launcher/K8sRemoteInterpreterProcess.java
zlosim: similar to the previous one, in [SparkInterpreterLauncher](https://github.com/apache/zeppelin/blob/c4c580a37fde649553d336984a94bcb1b2821201/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L203) desired output is `--conf|spark=123|--conf|spark.1=123` but can fix it to 
`--conf spark=123|--conf spark.1=123` if that is really what we want, please let me know

Reamer:
Hi @zlosim 
Sorry for the late reply. I tried to test your branch. Currently, I am getting a startup error because I have already fully transitioned my environment to Spark-3.3. Could you please do a rebase to the master.

zlosim:
Hi @Reamer 
no prob, should be rebased

Reamer (approved):
LGTM.
Thank you for your pull request, with this it is possible to remove the deprecated SPARK_SUBMIT option in the K8s launcher.
I have successfully tested your patch in my cluster.
As indicated, it might cause problems in the future if the pipe is also between "--conf" and the attachment. But since this is also done in the other start command, this should not block the merge.

PR #4402, opened 2022-07-10
[ZEPPELIN-5762] Fix default profile activation option to build succes…

dntjr8096:
### What is this PR for?
When building from source code, if the Profile option is not given to the maven command by default, an error like Screenshot occurs. So I changed the default profile activation option in the "zeppelin-server module" to hadoop2.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Documentation?

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5762

I tested with Travis CI, which is set by default.

* ```./mvnw clean package -DskipTests ``` before modifying the profile with 'hadoop2'
image

* ```./mvnw clean package -DskipTests ``` after modifying the profile with 'hadoop2'
image

### Questions:

### DIFF
--- a/zeppelin-server/pom.xml
+++ b/zeppelin-server/pom.xml
@@ -458,7 +458,7 @@
     <profile>
       <id>using-source-tree</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
+        <activeByDefault>false</activeByDefault>
       </activation>
       <properties>
         <zeppelin.daemon.package.base>
@@ -481,7 +481,9 @@
 
     <profile>
       <id>hadoop2</id>
-
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
       <properties>
         <hadoop.version>${hadoop2.7.version}</hadoop.version>
       </properties>


zeppelin-server/pom.xml
```diff
@@ -457,9 +457,6 @@
   <profiles>
     <profile>
       <id>using-source-tree</id>
-      <activation>
```
jongyoul, line 460: This is also needed. Please revert this change.

jongyoul:
Overall, LGTM except for the minor comment.

zeppelin-server/pom.xml
dntjr8096: Yes. Instead of deleting the line, I changed the activeByDefulat value to false.

dntjr8096: Yes. Instead of deleting the line, I changed the activeByDefulat value to false.

jongyoul: Yes, that should be reasonable but I think it's better not to change an unrelated thing for the PR so how about making another PR to handle this activation change?

dntjr8096: ok i'll do that

YeonghyeonKO (commented):
LGTM

jongyoul:
@YeonghyeonKO FYI, It's already been merged by #4403

PR #4403, opened 2022-07-10
[ZEPPELIN-5762] Modify activeByDefault value of hadoop3 profile to true for successful build.

dntjr8096:
### What is this PR for?
When building from the source code, the default profile activation option of the zeppelin-server module is "using-source-tree", so I got the error as in the screenshot. So I modified activeByDefault value of hadoop3 profile in the "zeppelin-server module" to true for successful build.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Documentation?

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5762

I tested with Travis CI, which is set by default. And after modifying the source code, I tried to build again.

* ```./mvnw clean package -DskipTests ``` before modifying the profile with 'hadoop2'
image

* ```./mvnw clean package -DskipTests ``` after modifying the profile with 'hadoop2'
image

### Questions:

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -448,3 +448,30 @@ jobs:
           R -e "IRkernel::installspec()"
       - name: run tests
         run: ./mvnw verify -DskipRat -pl livy -am  -B
+
+  default-build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Tune Runner VM
+        uses: ./.github/actions/tune-runner-vm
+      - name: Set up JDK 8
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: 8
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.m2/repository
+            !~/.m2/repository/org/apache/zeppelin/
+            ~/.spark-dist
+            ~/.cache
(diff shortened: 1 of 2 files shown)


zjffdu:
It's better to update `.github/workflows/core.yml` to verify this change.

jongyoul:
> It's better to update .github/workflows/core.yml to verify this change.

Do you mean to add a new task to confirm the compilation itself?

Reamer:
Instead of Hadoop 2, we should prefer Hadoop 3.

dntjr8096:
ok. I ran a git-rebase and added a 'default-build' workflow.

.github/workflows/core.yml
```diff
@@ -448,3 +448,29 @@ jobs:
...
+          path: |
+            ~/.m2/repository
+            !~/.m2/repository/org/apache/zeppelin/
+            ~/.spark-dist
+            ~/.cache
+          key: ${{ runner.os }}-zeppelin-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-zeppelin-
+      - name: build without any profiles
+        run: ./mvnw clean package -DskipTests
```
jongyoul, line 476: Please enter a new line at the end of the file :-)

jongyoul (approved):
I checked the new profile was passed

image

LGTM except for the one minor thing

.github/workflows/core.yml
dntjr8096: ok :) I'll be doing that

Reamer (approved):
LGTM

PR #4410, opened 2022-07-14
[ZEPPELIN-5771]Zeppelin Interpreter Support Alluxio From 1.0.0 To 2.5.0

ganlute:
### What is this PR for?
Zeppelin Interpreter Support Alluxio From 1.0.0 To 2.5.0

### What type of PR is it?
Improvement

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5771

CI passed

### Questions:


jongyoul:
@ganlute Hello ~ First of all, thank you for the contribution. BTW, could you please share your experiences with Alluxio in Zeppelin? I like this kind of contribution but we are trying to minimize interpreters managed in order to focus on our core architecture so I think it would be helpful for us to keep the interpreter.

ganlute:
We provide customers with big data cluster services, including HDFS, YARN, Hive, Hbase, Flink, Alluxio, etc. In order to use or manager cluster services more easily, we also provide services like Hue and Zeppelin as one of operation enters. In particular, on the one hand, we use Zeppelin to directly use or manage alluxio instead of logging into the server node, on the other hand, we can also better protect the privacy of our server node. @jongyoul

ganlute:
image
hi @jongyoul , I pass CI loaclly, can you help to approve running workflows? thank you.

jongyoul:
Thank you for the update!!! BTW, we are checking the status of interpreters so could you please update the wiki of https://wiki.apache.org/confluence/display/ZEPPELIN/Interpreter+Maintenance?

ganlute:
> Thank you for the update!!! BTW, we are checking the status of interpreters so could you please update the wiki of https://wiki.apache.org/confluence/display/ZEPPELIN/Interpreter+Maintenance?

Since I do not have permission to edit the wiki, I did not modify it. 
However, I have finished the following actions:
1. add the confluence page to issue links
2. add label of interpreter-maintanence to the issue

jongyoul:
Yes, you need a permission. Could you please give me your confluence ID? I'll give it to you.

ganlute:
> Yes, you need a permission. ~~Could you please give me your confluence ID? I'll give it to you. ~~ I gave you the permission baed on the ID which likes the wiki page similar to your github ID.

Yes, my confluence ID is ganlute which is the same as github ID.

I have finished updating status of alluxio just now.

NoorImann (approved):

NoorImann:
> ### What is this PR for?
> Zeppelin Interpreter Support Alluxio From 1.0.0 To 2.5.0
> 
> ### What type of PR is it?
> Improvement
> 
> ### Todos
> * [ ]  - Task
> 
> ### What is the Jira issue?
> https://issues.apache.org/jira/browse/ZEPPELIN-5771
> 
> ### How should this be tested?
> CI passed
> 
> ### Screenshots (if appropriate)
> ### Questions:
> * Does the licenses files need to update? No
> * Is there breaking changes for older versions? No
> * Does this needs documentation? XME

NoorImann:
> 1,2,3,4,5,6

ganlute:
hi @jongyoul , It is unbelievable that the CI runs failed again. 

I run action workflows twice in my repository(https://github.com/ganlute/zeppelin) and they work successfully(https://github.com/ganlute/zeppelin/actions). 

The failed case seems not to be related to this PR, should I pay attention to it?

jongyoul (approved):
LGTM but is it fine to remove some tests? We encourage adding tests so I recommend keeping tests if it's available. WDYT?

ganlute:
> LGTM but is it fine to remove some tests? We encourage adding tests so I recommend keeping tests if it's available. WDYT?

Yes, I agree with you. I only keep the basic test cases like cat/ls/load/copy/mkdir/count now and I have tested and pushed the code.

ganlute:
> Yes, I just want to know the reason why you delete some - complex - tests. Could you please tell me the reason why you delete them instead of modifying them?

In fact, the implementation of the Alluxio interpreter is `fs.run(args)` and it is contained in the module "alluxio-shell".

Without considering importing dependencies conflict between "alluxio-shell" and the other dependencies, maybe we just need add one "hello world" case like `ls` in test cases. Since the "hello world" case passed, if other case does not work well, it may be a bug of Alluxio not Zeppelin. In fact, at this time, I only import alluxio-shell as dependencies so it similar to above case. However, it is inevitable that dependencies will be imported and conflict may happened. So, the better way is to add all the Alluxio Cli test cases in Alluxio interpreter test case so that we are able to catch the conflict at the first time. However, it is a big project and need lots of works. As a compromise, I choose the only basic and daily used command test  to keep in the test cases.

jongyoul:
@ganlute Thank you for the detailed explanation. It makes sense to me. :-)

jongyoul:
Will merge it soon.

ganlute:
> Will merge it soon.

oh that is great, thank you~ 😄

PR #4416, opened 2022-07-19
[ZEPPELIN-5773]Fix isWindowsPath Function in ZeppelinConfiguration Class

JoWonHyeung:
### What is this PR for?
Even though the user's operating system is windows, 'isWindowsPath' function returns False and enters the else.
Therefore, I think that I need to change isWindowsPath function.

### What type of PR is it?
Bug Fix

### Todos
I replace isWindowsPath Function with isWindowsCheck function. isWindowCheck function uses SystemUtils.IS_OS_WINDOWS
to determine whether the OS is Windows or not. If the user's operating system is windows, It returns true.

### What is the Jira issue?
 https://issues.apache.org/jira/browse/ZEPPELIN-5773

CI

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -25,11 +25,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -327,7 +324,7 @@ public String getServerContextPath() {
 
   public String getKeyStorePath() {
     String path = getString(ConfVars.ZEPPELIN_SSL_KEYSTORE_PATH);
-    if (path != null && path.startsWith("/") || isWindowsPath(path)) {
+    if (path != null && path.startsWith("/") || isWindowsCheck()) {
```
zjffdu, line 327: Does just call `SystemUtils.IS_OS_WINDOWS` enough?

zjffdu:
@JoWonHyeung Do you plan to create another PR?

JoWonHyeung:
Yes, An error occurred in the test and  I am looking into it. I will upload it again later. Thank you!

PR #4419, opened 2022-07-22
[ZEPPELIN-5773] Fix isWindowsPath Function in ZeppelinConfiguration c…

JoWonHyeung:
What is this PR for?
Even though the user's operating system is windows, 'isWindowsPath' function returns False and enters the else.
Therefore, I think that I need to change isWindowsPath function.

What type of PR is it?
Bug Fix

Todos
I replace isWindowsPath Function with isWindowsCheck function. isWindowCheck function uses SystemUtils.IS_OS_WINDOWS
to determine whether the OS is Windows or not. If the user's operating system is windows, It returns true.

What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5773

How should this be tested?
CI

Questions:
Does the licenses files need to update? No
Is there breaking changes for older versions? Yes


zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -25,11 +25,7 @@
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
jongyoul, line 28: I think you might change your IDE setting not to use `*`. You can change the value to `200` for instance. :-)

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -636,8 +633,12 @@ public String getInterpreterPortRange() {
     return getString(ConfVars.ZEPPELIN_INTERPRETER_RPC_PORTRANGE);
   }
 
-  public boolean isWindowsPath(String path){
-    return path.matches("^[A-Za-z]:\\\\.*");
+  public boolean isWindowsCheck(){
```
jongyoul, line 636: I'm not sure but how about using `isWindows` or `checkWindows` as a method name?

JoWonHyeung: OK, I will do what you say.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -636,8 +633,12 @@ public String getInterpreterPortRange() {
...
+    Boolean wc = SystemUtils.IS_OS_WINDOWS;
+    if(Objects.isNull(wc)) {
```
jongyoul, line 638: It returns a primitive type. I guess you don't have to check if it's null or not. 

https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/SystemUtils.html#IS_OS_WINDOWS

JoWonHyeung: OK, Thank you.  I will edit and re-upload.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -636,8 +637,9 @@ public String getInterpreterPortRange() {
     return getString(ConfVars.ZEPPELIN_INTERPRETER_RPC_PORTRANGE);
   }
 
-  public boolean isWindowsPath(String path){
-    return path.matches("^[A-Za-z]:\\\\.*");
+  public boolean checkWindows(){
+    Boolean wc = SystemUtils.IS_OS_WINDOWS;
+    return wc;
```
jongyoul, line 642: How about?
```suggestion
    return SystemUtils.IS_OS_WINDOWS;
```

zeppelin-zengine/src/test/java/org/apache/zeppelin/conf/ZeppelinConfigurationTest.java
```diff
@@ -69,18 +69,18 @@ public void getAllowedOriginsNoneTest() throws MalformedURLException {
...
 
   @Test
-  public void isWindowsPathTestFalse() {
+  public void checkWindowsTestFalse() {
 
     ZeppelinConfiguration conf = ZeppelinConfiguration.create("zeppelin-test-site.xml");
-    Boolean isIt = conf.isWindowsPath("~/test/file.xml");
+    Boolean isIt = conf.checkWindows();
     Assert.assertFalse(isIt);
   }
```
jongyoul, line 85: How do the tests work?

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
Reamer: much better

zeppelin-zengine/src/test/java/org/apache/zeppelin/conf/ZeppelinConfigurationTest.java
JoWonHyeung: I will question it.

I only changed the function name in the previous version for the time being. The internal code is the same as before.

_SystemUtils.IS_OS_WINDOWS_ will have different results for each operating system, so how can I create test code?

Do you have any good ideas?

Thank you.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
yaini: nitpick) How about `isWindowsOS`? It would be better to specify what to check.

zeppelin-zengine/src/test/java/org/apache/zeppelin/conf/ZeppelinConfigurationTest.java
yaini: Previously written tests were checked using file paths, but as this function becomes system-dependent, the meaning of the previous tests seems to have changed.
If you write a test, it would be better to mock `checkWindows` and test the result value of a function that uses `checkWindows`, such as `getAbsoluteDir` and `getKeyStorePath` .

jongyoul: We can define a new CI env like `CI_LINUX=TRUE` in Github action and use it when testing like
```
    if(System.getenv("CI_LINUX").equals("TRUE")) {
      // Github action
      Assert.assertFalse(conf.checkWindows());
    } else if (System.getenv("CI_WINDOW").equals("TRUE")) {
      // Appveyor
      Assert.assertTrue(conf.checkWindows());
    } else {
      // Local
      Assert.assertFalse(conf.checkWindows());
    }
```

jongyoul (approved):

PR #4427, opened 2022-07-25
[ZEPPELIN-5638] Add flink job interval check

yaini:
### What is this PR for?
zeppelin-interpreter-flink-yarn produce a lot of logs, caused disk space to run out

### What type of PR is it?
Improvement

### Todos
* [x] - Add validation the value of zeppelin.flink.job.check_interval

### Need Help
I think negative value is invalid input, so I make it just return and not to add job
But, the issuer suggest that set default value 1000. Is it better?

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5638

* CI

### Questions:


jongyoul:
BTW, if it returns without a progress checker, doesn't it occur infinitive loop?

jongyoul (approved):
Oh, I completely misunderstood your change. I checked the change again. LGTM

PR #4431, opened 2022-07-27
[ZEPPELIN-4794] Add comment about changing localhost when running Zeppelin on WSL or Windows

JoWonHyeung:
### What is this PR for?
When trying to run Zeppelin on Windows WSL linux, server property in Zeppelin-site.xml should be set to 0.0.0.0.
But, we don't change from 127.0.0.0 to 0.0.0.0 because of security issue. I think that we should at least write a comment.

### What type of PR is it?
Documentation

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-4794

CI

### Questions:
* Does the licenses files need to update? No.

### DIFF
--- a/conf/zeppelin-site.xml.template
+++ b/conf/zeppelin-site.xml.template
@@ -22,7 +22,7 @@
 <property>
   <name>zeppelin.server.addr</name>
   <value>127.0.0.1</value>
-  <description>Server binding address</description>
+  <description>Server binding address. If you cannot connect to your web browser on WSL or Windows, change 127.0.0.1 to 0.0.0.0.</description>
 </property>
 
 <property>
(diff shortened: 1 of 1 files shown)


jongyoul:
@JoWonHyeung Could you please move your comment in the description field?

JoWonHyeung:
@jongyoul Ok. I will re-update.

conf/zeppelin-site.xml.template
```diff
@@ -22,7 +22,7 @@
 <property>
   <name>zeppelin.server.addr</name>
   <value>127.0.0.1</value>
-  <description>Server binding address</description>
+  <description>Server binding address. If you cannot connect to your web browser on WSL or Windows, change 127.0.0.1 to 0.0.0.0.</description>
```
jongyoul, line 25:
```suggestion
  <description>Server binding address. If you cannot connect to your web browser on WSL or Windows, change 127.0.0.1 to 0.0.0.0. It, however, causes security issues when you open your machine to the public</description>
```

JoWonHyeung: Thank you.

PR #4436, opened 2022-07-27
[ZEPPELIN-5789] Update Dockerfile to use JDK instead of JRE

vjr:
### What is this PR for?
Fixes issues such as trying to add a new interpreter belonging to "jdbc" group having hive-jdbc conf results in missing dependencies (tools.jar) and such.

### What type of PR is it?
Bug Fix

### Todos
(none)

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5789

Build, publish and run the image via Dockerfile `docker run -d -p 80:8080 --name test imagename`

### Questions:
* Does the licenses files need to update? NO

### DIFF
--- a/scripts/docker/zeppelin/bin/Dockerfile
+++ b/scripts/docker/zeppelin/bin/Dockerfile
@@ -29,7 +29,7 @@ ENV LOG_TAG="[ZEPPELIN_${Z_VERSION}]:" \
 
 RUN echo "$LOG_TAG install basic packages" && \
     apt-get -y update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y locales language-pack-en tini openjdk-8-jre-headless wget unzip && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y locales language-pack-en tini openjdk-8-jdk-headless wget unzip && \
     # Cleanup
     rm -rf /var/lib/apt/lists/* && \
     apt-get autoclean && \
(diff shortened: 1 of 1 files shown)


jongyoul (approved):

jongyoul:
Thank you for your contribution. :-)

Reamer:
`tools.jar` should not be used any further. In JDK 9 `tools.jar` and `rt.jar` was even removed from the JDK.
We should better raise the dependency that `tools.jar` needs during runtime. Take a look at https://openjdk.org/jeps/220

vjr:
Agreed , although I propose to have this pr merged as interim fix until Zeppelin itself moves to newer JDK version, is this okay?

Reamer:
Please add a comment. When we upgrade to > jdk8, we should switch back to jre.

vjr:
> Please add a comment. When we upgrade to > jdk8, we should switch back to jre.

Done.

I was only able to set the target version to 0.11 in Jira... Is there a way to request inclusion in any further 0.10.x release if it happens? Or I presume that will happen by default if this PR gets merged?

Reamer (approved):
LGTM

Reamer:
> I was only able to set the target version to 0.11 in Jira... Is there a way to request inclusion in any further 0.10.x release if it happens? Or I presume that will happen by default if this PR gets merged?

The merger can backport this small change to branch-0.10.

jongyoul (approved):

PR #4453, opened 2022-08-28
[ZEPPELIN-5804] support python 3.10

PrimWILL:
### What is this PR for?
Change py4j version (0.10.7 -> 0.10.9.7) in python interpreter directory to support python 3.10

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5804

CI and manual (3.9 and 3.10)

python3 10

### Questions:


jongyoul (approved):

Reamer:
I think this should be tested via CI.

PrimWILL:
> I think this should be tested via CI.

In CI, this already passed python 3.7 and python 3.8 tests . Do I have to add more python tests for 3.10 in CI?

jongyoul:
I'll merge it if there's no further discussion

Reamer:
> Do I have to add more python tests for 3.10 in CI?

Sooner or later we should also test with Python 3.10 in the CI setup. But this can be done in a separate pull request.

PrimWILL:
> Sooner or later we should also test with Python 3.10 in the CI setup. But this can be done in a separate pull request.

Got it. Then I'll register related issue in jira.

PR #4441, opened 2022-08-07
[ZEPPELIN-4871] Fix livy interpreter 0.7.1 donwload link 

soicem:
### What is this PR for?
During a previous [PR](https://github.com/apache/zeppelin/pull/4433) related to livy interpreter version upgrade, CI is not working because of unvalid download link of livy interpreter 0.7.1.  then, I create a new PR for fixing a download link of livy interpreter zip file.

@jongyoul 
[previous comment](https://github.com/apache/zeppelin/pull/4433#issuecomment-1200735324)

### What type of PR is it?
bug

### What is the Jira issue?
(https://issues.apache.org/jira/browse/ZEPPELIN-4871)

### How should this be tested?
* Strongly recommended: add automated unit tests for any new or changed behavior
* Outline any manual steps to test the PR here.

AS IT
https://dist.apache.org/repos/dist/release/incubator/livy/0.7.1-incubating/livy-0.7.1-incubating-bin.zip
TO BE
https://dist.apache.org/repos/dist/release/incubator/livy/0.7.1-incubating/apache-livy-0.7.1-incubating-bin.zip

### Questions:
* Does the licenses files need to update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #4470, opened 2022-09-27
update livy archive name

raboof:
### What is this PR for?

Fix the download script so it can correctly download with livy 0.6.0 and later

### What type of PR is it?
Bug Fix

The `livy-0-7-with-spark-2-2-0-under-python3` GitHub Actions workflow uses this script, so we should be able to see that this doesn't fail to download livy anymore.

### DIFF
--- a/testing/downloadLivy.sh
+++ b/testing/downloadLivy.sh
@@ -49,7 +49,7 @@ download_with_retry() {
 }
 
 LIVY_CACHE=".livy-dist"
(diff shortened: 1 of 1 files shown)


Reamer:
Please also take a look at #4441

raboof:
> Please also take a look at #4441

sorry, totally missed that, that is indeed more complete :+1:

Reamer (approved):
LGTM

PR #4476, opened 2022-10-11
Upgrade gson to 2.8.9 due to CVE-2022-25647

huage1994:
### What is this PR for?
Upgrade gson to 2.8.9 due to CVE-2022-25647
The version of gson  is 2.8.6 in parent POM. And the LICENESE should be updated too.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-5829

CI passed

### Questions:


Reamer (approved):
LGTM

PR #4482, opened 2022-10-13
fix(sec): upgrade org.scala-lang:scala-compiler to 2.11.12

Zhfuln:
### What happened？
There are 1 security vulnerabilities found in org.scala-lang:scala-compiler 2.11.8
- [CVE-2017-15288](https://www.oscs1024.com/hd/CVE-2017-15288)

### What did I do？
Upgrade org.scala-lang:scala-compiler from 2.11.8 to 2.11.12 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/spark/interpreter/pom.xml
+++ b/spark/interpreter/pom.xml
@@ -14,10 +14,7 @@
   ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
-  -->
-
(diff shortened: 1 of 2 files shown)


Reamer (changes requested):
Please change the formatting of the file to the original. A newline at the end of the file is also wanted and must not be removed.
You are downgrading scala, why?

PR #4512, opened 2022-11-21
[MINOR] Polish maven plugins for root pom

jongyoul:
### What is this PR for?
* Deactivate apache-rat by default and change some CI settings
* Move git-commit-id-plugin under `zeppelin-interpreter`

### What type of PR is it?
Improvement

### Todos
* [x] - Polish pom.xml

### What is the Jira issue?
N/A

CI passed

We cannot see two plugins anymore in root pom
image

### Questions:

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -59,9 +59,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-zeppelin-
       - name: install application with some interpreter
-        run: ./mvnw install -Pbuild-distr -DskipRat -DskipTests -pl zeppelin-server,zeppelin-web,spark-submit,spark/scala-2.11,spark/scala-2.12,spark/scala-2.13,markdown,angular,shell -am -Phelium-dev -Pexamples -P${{ matrix.hadoop }} -B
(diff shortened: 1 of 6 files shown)


Reamer:
I find the change regarding rat very good. Why the change with the git plugin? It can be useful that the Git information is in each artifact of Zeppelin.

jongyoul:
> Why the change with the git plugin?

Basically, I would like not to create `target` directory of the root directory. And I thought the `git.properties` is being only in some API only. Concerning the information, I think we already have a version number in our jars. Do you still think we'd better include those information in all jars?

jongyoul:
>I have not seen the use of the file for the API yet. I'll have a look right now to see if I can find the post.

Here is the code. https://github.com/apache/zeppelin/blob/85ed8e2e51e1ea10df38d4710216343efe218d60/zeppelin-server/src/main/java/org/apache/zeppelin/rest/ZeppelinRestApi.java#L63-L64

The PR related to `git-commit-id-plugin` are https://issues.apache.org/jira/browse/ZEPPELIN-2622 and https://github.com/apache/zeppelin/pull/2398. I couldn't find motivations for keeping build information for all jars in the PR. So I think the contributor wanted to make a new API including versions. I, however, think we could keep the information for debugging purposes.

jongyoul:
> Currently I see for zeppelin-common only the following information inside MANIFEST.MF. Is this information sufficient?

I meant the name of jars like `zeppelin-server-0.11.0-SNAPSHOT.jar`. :-) I thought general users would use the pre-built version like `0.10.x`. In the case of developers like us, we will know our branch so I don't think that we need them.

Reamer (approved):
LGTM
All questions were answer and I agree with the changes.

PR #4513, opened 2022-11-23
[MINOR] Remove warnning

jongyoul:
### What is this PR for?
Remove outdated maven plugin setting.

I saw some warnings when I check some maven logs.
```
[WARNING] The POM for org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved: org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced
```

I tried to find the package in maven central but couldn't find it and reached a description of https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html. This link shows that the setting is only for maven plugin developers. I don't know the initial motivation for the plugin because it's added in 2013 and I didn't contribute to Zeppelin at that time. I, however, feel like we don't need it for now because the description indicates it's only for plugin developers.

### What type of PR is it?
Improvement

### Todos
* [x] - Remvoe plugin

### What is the Jira issue?
N/A

Nothing should change

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1697,60 +1697,7 @@
           <version>${plugin.antrun.version}</version>
         </plugin>
 
-        <!--This plugin's configuration is used to store Eclipse m2e settings
-          only. It has no influence on the Maven build itself. -->
-        <plugin>
-          <groupId>org.eclipse.m2e</groupId>
-          <artifactId>lifecycle-mapping</artifactId>
-          <version>${plugin.lifecycle.mapping.version}</version>
-          <configuration>
-            <lifecycleMappingMetadata>
-              <pluginExecutions>
-                <pluginExecution>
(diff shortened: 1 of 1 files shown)


Reamer (approved):
I am using Eclipse. After I removed the change in the root pom.xml, everything continued to run properly.

PR #4520, opened 2022-11-30
[MINOR] Fix Appveyor error

jongyoul:
### What is this PR for?
Fix the issue for appveyor which has an error result.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Fix .appveyor.yml

### What is the Jira issue?
N/A

Windows CI passes

### Questions:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,4 +31,4 @@ install:
 
 build_script:
   - echo "Build"
-  - ps: ./mvnw clean package -DskipTests -B -q
+  - ps: ./mvnw clean package -DskipTests -B
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,7 +24,7 @@ shallow_clone: true
 build: Script
 
 os:
-  - Visual Studio 2015
+  - Visual Studio 2022
 
 environment:
   APPVEYOR_SAVE_CACHE_ON_ERROR: True
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,12 +36,7 @@ cache:
   - '%USERPROFILE%/.m2'
 
 build_script:
-  - cmd: echo %PATH%
   - cmd: set JAVA_HOME=C:\Program Files\Java\jdk1.8.0
-  - cmd: set PATH=%JAVA_HOME%;%PATH%
-  - cmd: echo %PATH%
-  - cmd: echo %USERPROFILE%
-  - cmd: java -version
   - cmd: >-
       ./mvnw.cmd -v clean package -DskipTests ^
       -Dorg.slf4j.simpleLogger.log.org.apache.maven.plugins.shade.mojo.ShadeMojo=warn ^
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,4 +31,4 @@ install:
 
 build_script:
   - echo "Build"
-  - ps: ./mvnw clean package -DskipTests -B
+  - ps: ./mvnw clean package -DskipTests -Dorg.slf4j.simpleLogger.log.org.apache.maven.plugins.shade.mojo.ShadeMojo=warn -Dorg.slf4j.simpleLogger.log.com.googlecode.download.maven.plugin.internal.WGet=warn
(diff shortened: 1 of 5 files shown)


Reamer:
Nice Maven option that I didn't know about until now.
It would be very elegant if we also changed `-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn` to `--no-transfer-progress` in the GitHub CI.

https://github.com/apache/zeppelin/blob/364c8554a66d83446025235c2f7d0e3b223db121/.github/workflows/core.yml#L14

I'll prepare a PR.

jongyoul:
Yes, sometimes quite annoying when we see the logs :-)

.appveyor.yml
```diff
@@ -31,4 +31,4 @@ install:
 
 build_script:
   - echo "Build"
-  - ps: ./mvnw clean package -DskipTests
+  - ps: ./mvnw clean package -DskipTests -B
```
jongyoul, line 34:
```suggestion
  - ps: ./mvnw clean package -DskipTests -B -q
```

jongyoul:
@Reamer @zjffdu This PR makes appveyor back to normal. It was a problem that scala compiler prints warning messages directly and it's judged as error in appveyor. I disabled to print warnings as it's general warning messages. Could you please check kindly check it?

pom.xml
```diff
@@ -1585,6 +1585,14 @@
           <groupId>net.alchim31.maven</groupId>
           <artifactId>scala-maven-plugin</artifactId>
           <version>${plugin.scala.alchim31.version}</version>
+          <configuration>
+            <recompileMode>all</recompileMode>
```
pan3793, line 1589: is it necessary? incremental mode should be faster

jongyoul: Agreed with the performance. But it causes failure as it shows an error when zinc is not downloaded for the first time. So I checked it. Moreover, we are developing scala code but it's not a heavy job so I think it should be fine. Otherwise, we can introduce some CI profile and enable it on it but I don't know if it's really necessary

pan3793 (approved):

pom.xml
```diff
@@ -1585,6 +1585,14 @@
           <groupId>net.alchim31.maven</groupId>
           <artifactId>scala-maven-plugin</artifactId>
           <version>${plugin.scala.alchim31.version}</version>
+          <configuration>
```
Reamer, line 1588: Thank you for solving the CI problem.

I would do the configuration of this plugin in the sub-projects. In most Scala subprojects there is already a custom compiler configuration and the settings from the parent pom.xml should have no effect there. It can be confusing if there is compiler configuration in two places.

jongyoul: Oh! I honestly merged this setting after adding several places at first. But I also think it's good. I'll add similar settings in the proper places then.

Reamer: If we create a uniform configuration across all sub-modules, then I would definitely prefer a configuration in the parent pom. But I don't know if this is possible.

jongyoul: I also don't think so. we should override parent configuration sometimes so I think it could be fine to have separate configuration for every sub-project.

zjffdu: I think it is possible to put it in root pom's `pluginManagement` as default configuration, and sub module can customize its configuration.

jongyoul: I removed redundant configuration only in this PR. I, however, feel like we'd better clean up scala configuration. @Reamer @zjffdu Maven option is only for building Zeppelin itself, not running it, so I think we could simplify the scala compile option. I created https://issues.apache.org/jira/browse/ZEPPELIN-5864

Reamer:
Please rebase your branch to the current master, PR #4534 has corrected core / interpreter-test-non-core.

jongyoul:
@Reamer BTW, in the case of the core module, does the PR fix them as well?

Reamer:
> @Reamer BTW, in the case of the core module, does the PR fix them as well?

unfortunately not

jongyoul:
@Reamer BTW, I have checked and found some workarounds, fortunately. Let me make a PR soon.

jongyoul:
@Reamer @zjffdu I think it archives its own purpose to make appveyor healthy again :-)

Reamer (approved):
LGTM

PR #4529, opened 2022-12-03
[MINOR] Specify branch names for Github action

jongyoul:
### What is this PR for?
Specify the condition to trigger GitHub action when pushing commits. E.g. dependabot's PRs run twice for GitHub action as we allow to run it for all branches in the repository.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Add `branches:` information for `push:`

### What is the Jira issue?
N/A

* Check dependabot's PR so that it runs only once for each Github actions

### Questions:

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -2,9 +2,8 @@ name: core
 
 on:
   push:
-    branches:
-      - master
-      - branch-*
+    ignore-branches:
+      - dependabot/**
   pull_request:
     branches:
       - master
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -1,11 +1,16 @@
 name: core
+
 on:
   push:
+    branches:
+      - master
+      - branch-*
   pull_request:
     branches:
       - master
       - branch-*
-    types: [opened, synchronize]
+    paths:
+      - !**.md # Ignore documentation contribution
 
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/core.yml
```diff
@@ -1,11 +1,14 @@
...
 on:
   push:
+    branches:
+      - master
+      - branch-*
   pull_request:
     branches:
       - master
       - branch-*
-    types: [opened, synchronize]
```
jongyoul, line 8: > By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

pan3793 (dismissed):

.github/workflows/core.yml
```diff
@@ -1,11 +1,14 @@
 name: core
+
 on:
   push:
+    branches:
```
Reamer, line 5: The GitHub CI pipeline is turned off with this configuration in developer branches within forks.

jongyoul: I also thought so. Let me finding a solution so that it should be adopted in apache repository :-)

jongyoul: I changed it to use `branches-ignores` instead. :-)

Reamer (approved):
LGTM

PR #4523, opened 2022-11-30
[ZEPPELIN-5857] Bump gRPC and protobuf to support build on macOS M1 chips

pan3793:
### What is this PR for?

I suppose this upgrade will not introduce compatibility issue, see details discussion about version requirements and compatibility https://github.com/apache/incubator-uniffle/pull/13

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5857

### How should this be tested?

Build on my M1 mac.

### Screenshots (if appropriate)

before
image

after
image

### Questions:
* Does the licenses files need to update? Yes, and updated.
* Is there breaking changes for older versions? No.
* Does this needs documentation? No.

PR #4541, opened 2022-12-29
[MINOR] Support aarch for frontend plugin

jongyoul:
### What is this PR for?
Fixing build error when using aarch jdk for mac OS. our version of node doesn't support amd64 arch but the old plugin tries to download arm version.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Update `frontend-maven-plugin`

### What is the Jira issue?
N/A

* Pass to build with arm jdk in macOS

image

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
     <!-- frontend maven plugin related versions-->
     <node.version>v12.3.1</node.version>
     <npm.version>6.9.0</npm.version>
-    <plugin.frontend.version>1.6</plugin.frontend.version>
+    <plugin.frontend.version>1.12.1</plugin.frontend.version>
 
     <!-- common library versions -->
     <slf4j.version>1.7.30</slf4j.version>
--- a/zeppelin-plugins/notebookrepo/s3/pom.xml
+++ b/zeppelin-plugins/notebookrepo/s3/pom.xml
@@ -34,7 +34,7 @@
     <description>NotebookRepo implementation based on S3</description>
 
     <properties>
-        <aws.sdk.version>1.12.373</aws.sdk.version>
+        <aws.sdk.version>1.12.261</aws.sdk.version>
         <plugin.name>NotebookRepo/S3NotebookRepo</plugin.name>
     </properties>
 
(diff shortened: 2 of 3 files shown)


jongyoul:
BTW, as mentioned in #4523, The compilation time becomes faster if we use arm jdk in m1 mac

huage1994 (dismissed):

huage1994:
It looks good to me 👍

jongyoul:
@huage1994 Thank you. BTW, it has a dependency issue. It's related to Helium feature because it uses frontend-core to download npm package for custom visualization feature. I think we'd better reinvent the Helium with minimum but it's a problem anyway. I'll fix it and let you know. Please review it again after fixing the issue.

```
Dependency convergence error for com.fasterxml.jackson.core:jackson-core:2.13.0 paths to dependency are:
+-org.apache.zeppelin:zeppelin-interpreter-integration:0.11.0-SNAPSHOT
  +-org.apache.zeppelin:zeppelin-zengine:0.11.0-SNAPSHOT
    +-com.github.eirslett:frontend-plugin-core:1.12.1
      +-com.fasterxml.jackson.core:jackson-core:2.13.0
and
+-org.apache.zeppelin:zeppelin-interpreter-integration:0.11.0-SNAPSHOT
  +-org.apache.zeppelin:zeppelin-server:0.11.0-SNAPSHOT
    +-com.fasterxml.jackson.core:jackson-databind:2.12.6.1
      +-com.fasterxml.jackson.core:jackson-core:2.12.6
```

Reamer (changes requested):
CI is broken, because of dependency-convergence. I added the review so no one would accidentally click merge.

pan3793:
I roughly remember that NodeJS officially supported Apple Silicon in v16(actually in v15, but it's not an LTS, and was EOL), and eventually, they decided to port this feature to v14 but does not publish the binary artifacts.

So I suppose upgrading NodeJS v16 is necessary.

Sorry I can not take this because I don't have much knowledge in frontend stacks.

jongyoul:
Thank you for the description. I also checked it and found the latest version is supporting that way. In our current version, the plugin tries to add suffix for arm64 to download proper node binary but in the latest version, it checks if the our node version is below 16 or not.

jongyoul:
@Reamer The CI becomes back as the previous so it's ok. Could you please recheck it?

Reamer (approved):
LGTM

PR #4568, opened 2023-02-22
[ZEPPELIN-5883] Upgrade influxDB client version to 2.3.0

akishichinibu:
### What is this PR for?
Zeppelin Interpreter Support InfluxDB From 1.7.0 To 2.3.0

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5883

* CI passed

### Questions:

### DIFF
--- a/influxdb/pom.xml
+++ b/influxdb/pom.xml
@@ -35,8 +35,8 @@
     <properties>
         <interpreter.name>influxdb</interpreter.name>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <influxdb.client.version>1.7.0</influxdb.client.version>
-        <dependency.okhttp3.version>3.13.1</dependency.okhttp3.version>
+        <influxdb.client.version>2.3.0</influxdb.client.version>
+        <dependency.okhttp3.version>4.7.2</dependency.okhttp3.version>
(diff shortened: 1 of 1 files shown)


influxdb/pom.xml
```diff
@@ -35,8 +35,8 @@
     <properties>
         <interpreter.name>influxdb</interpreter.name>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <influxdb.client.version>1.7.0</influxdb.client.version>
-        <dependency.okhttp3.version>3.13.1</dependency.okhttp3.version>
+        <influxdb.client.version>2.3.0</influxdb.client.version>
+        <dependency.okhttp3.version>4.7.2</dependency.okhttp3.version>
```
akishichinibu, line 39: The test dependency [okhttp3](https://mvnrepository.com/artifact/com.squareup.okhttp3/mockwebserver) should also be upgrade to `4.7.2` or the tests will fail.

Reamer (approved):
I don't have this interpreter in use, but I think an update of the client library is not wrong.
Are there perhaps new client features that could be covered with a simple unit test?

PR #4573, opened 2023-03-07
[ZEPPELIN-5888] Upgrade sparql version to 3.17.0

akishichinibu:
### What is this PR for?
Zeppelin Interpreter Support SPARQL From 3.12.0 To 3.17.0

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5888

* CI passed

### Questions:

### DIFF
--- a/sparql/pom.xml
+++ b/sparql/pom.xml
@@ -33,7 +33,7 @@
 
     <properties>
         <interpreter.name>sparql</interpreter.name>
-        <jena.version>3.12.0</jena.version>
+        <jena.version>3.17.0</jena.version>
     </properties>
 
     <dependencies>


sparql/pom.xml
```diff
@@ -33,7 +33,7 @@
 
     <properties>
         <interpreter.name>sparql</interpreter.name>
-        <jena.version>3.12.0</jena.version>
+        <jena.version>3.17.0</jena.version>
```
akishichinibu, line 36: The Java 11 is [required since Jena 4](https://jena.apache.org/download/index.cgi#:~:text=Jena4%20requires%20Java%2011.).

jongyoul:
BTW, is it necessary to keep this interpreter? I don't know about the detail but sometimes I feel like we should think we could maintain this interpreter continuously or not. WDYT?

PR #4578, opened 2023-03-31
[ZEPPELIN-5891] Method contains useless variables

NicolasGong:
### What is this PR for?
Method contains useless variables

### What type of PR is it?
Improvement

### What is the Jira issue?
* [ZEPPELIN-5891]

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
@@ -94,11 +94,6 @@ public class AuthorizationService implements ClusterEventListener {
     this.notesAuth.put(noteId, noteAuth);
   }
 
-  public void cloneNoteMeta(String noteId, String sourceNoteId, AuthenticationInfo subject) {
-    NoteAuth noteAuth = new NoteAuth(noteId, subject, conf);
-    this.notesAuth.put(noteId, noteAuth);
-  }
-
   /**
    * Persistent NoteAuth
    *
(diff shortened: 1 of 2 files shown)


NicolasGong:
> Very strange, we want to clone the metadata. How does the code work without a source NoteId?

I think this method is to put the new NoteId and AuthenticationInfo into the cache. This method does not require sourceNoteId.

Reamer:
I think we can delete the `cloneNoteMeta` method completely and replace it with `createNoteAuth`.

In my opinion, the original rights should not be transferred from the old Note to the new Note. It makes sense to store the roles of the user who is currently cloning the note.

NicolasGong:
I agree with you and resubmitted the code.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
```diff
@@ -93,12 +93,7 @@ public void createNoteAuth(String noteId, AuthenticationInfo subject) {
     NoteAuth noteAuth = new NoteAuth(noteId, subject, conf);
     this.notesAuth.put(noteId, noteAuth);
   }
-
-  public void cloneNoteMeta(String noteId, String sourceNoteId, AuthenticationInfo subject) {
-    NoteAuth noteAuth = new NoteAuth(noteId, subject, conf);
-    this.notesAuth.put(noteId, noteAuth);
-  }
-
+
```
Reamer, line 96: I suspect spaces here.

Reamer (changes requested):
Please check whether the blanks are present.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/AuthorizationService.java
NicolasGong: I removed extra spaces.

Reamer (approved):
LGTM

NicolasGong:
Some checks were not successful,but i don't know why. These errors do not appear to be related to this commit。

PR #4591, opened 2023-04-26
[ZEPPELIN-5898]fix:down csv data error

zhugezifang:
### What is this PR for?

This pull request fix the bug of download csv data error

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5898

* CI

### Questions:

### DIFF
--- a/docs/interpreter/flink.md
+++ b/docs/interpreter/flink.md
@@ -144,12 +144,14 @@ docker run -u $(id -u) -p 8080:8080 --rm -v /mnt/disk1/flink-sql-cookbook-on-zep
 
 Download Flink 1.10 or afterwards (Scala 2.11 & 2.12 are both supported)
 
-### Specific for Flink 1.15
+### Specific for Flink 1.15 and above
 
 Flink 1.15 is scala free and has changed its binary distribution. If you would like to make Zeppelin work with Flink 1.15, you need to do the following extra steps.
 * Move FLINK_HOME/opt/flink-table-planner_2.12-1.15.0.jar to FLINK_HOME/lib
 * Move FLINK_HOME/lib/flink-table-planner-loader-1.15.0.jar to FLINK_HOME/opt
 * Download flink-table-api-scala-bridge_2.12-1.15.0.jar and flink-table-api-scala_2.12-1.15.0.jar to FLINK_HOME/lib
+
+Flink 1.16 introduces new `ClientResourceManager` for sql client, you need to move `FLINK_HOME/opt/flink-sql-client-1.16.0.jar` to `FLINK_HOME/lib`
 
 ## Flink on Zeppelin Architecture
 


Reamer:
Still

> Thanks for the contribution @zhugezifang , could you update the PR description?

zhugezifang:
> Still
> 
> > Thanks for the contribution @zhugezifang , could you update the PR description?

ok，updated it

Reamer:
I do not understand the fix. Could you please explain it.

zhugezifang:
> I do not understand the fix. Could you please explain it.

let we look the example of https://issues.apache.org/jira/browse/ZEPPELIN-5898 

when i query the result in zeppelin，the result is

image

and click download
image


but the data of downloaded has problems

image


the column of params_value is json，but  open  in  csv, it is not ，so it has a bug

this pr is to fix the bug

zhugezifang:
>

zhugezifang:
> I do not understand the fix. Could you please explain it.

excuce me，the is a webchat group or other group of zeppelin，we can discuss it more timely。

and i want to add the feature to zeppelin ，it is a brief description in https://issues.apache.org/jira/browse/ZEPPELIN-5896

i discuss the feature with @zjffdu ,he say i can try it

Reamer:
I don't think it's a specific JSON problem, but rather a delimiter problem.
The CSV is broken because in the JSON example there is a `,` in the text field and this is also used as a delimiter.
What do you think about it?

zhugezifang:
> I don't think it's a specific JSON problem, but rather a delimiter problem. The CSV is broken because in the JSON example there is a `,` in the text field and this is also used as a delimiter. What do you think about it?

yes ,the code solve the  ',' of delimiter, but it solved not perfect，it also exist bug

image

zhugezifang:
> > I do not understand the fix. Could you please explain it.
> 
> excuce me，there is a webchat group or other group of zeppelin，we can discuss it more timely。
> 
> and i want to add the feature to zeppelin ，it is a brief description in https://issues.apache.org/jira/browse/ZEPPELIN-5896
> 
> i discuss the feature with @zjffdu ,he say i can try it

@Reamer  hi，i also want to add this feature into zeppelin，and could I get your support？

zhugezifang:
> > @Reamer hi，i also want to add this feature into zeppelin，and could I get your support？
> 
> Of course. I was able to reproduce the problem and understood the problem. Your change definitely solves the problem. It's just not a JSON specific problem, but a general masking problem.
> 
> I would remove the if statement. It is not necessary because replaceAll also looks for the quotes.
> 
> ```
> // Mask all quotes
> stringValue = stringValue.replaceAll('"', '""');
> ```

ok,i update it later and then add a new pr

PR #4609, opened 2023-05-30
[ZEPPELIN-5920] Fix deprecation warnings

matthias-koch:
### What is this PR for?
If z.show is used with a Panda DataFrame that has a timestamp as index, then a DeprecationWarning appears for every row in the data frame.
This issue was fixed.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5920

* A new unit test was added

### Questions:

### DIFF
--- a/python/src/main/resources/python/zeppelin_context.py
+++ b/python/src/main/resources/python/zeppelin_context.py
@@ -221,7 +221,7 @@ class PyZeppelinContext(object):
                 body_buf.write(self.normalizeColumn(str(cell)))
             # don't print '\n' after the last row
             with warnings.catch_warnings():
-                warnings.simplefilter("ignore")
+                warnings.simplefilter("ignore", category=DeprecationWarning)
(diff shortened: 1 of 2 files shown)


Reamer:
Do you think it would be smart to ignore all the warnings?

matthias-koch:
I could limit it to DeprecationWarnings.
Another option would be to check the type of index. However to do this I would have to import numpy in zeppelin_context.py, which I wanted to avoid.

Reamer:
It seems that the data types are different. What do you think about simply counting down?
```
rowNumber = min(len(rows), len(df.index.values)) 
for idx, row in zip(index, rows):
    rowNumber = rowNumber - 1;
    if (rowNumber == 0)
       body_buf.write("\n")
```

A screenshot of my reproduction
grafik

Reamer:
Let's wait for the tests.

PR #4611, opened 2023-05-31
[ZEPPELIN-5915]improve query efficiency

zhugezifang:
### What is this PR for?
improve query efficiency:deal complex sql

Design Document: https://docs.google.com/document/d/1wruK0ZZ0XiriYOraFa5WYSz531pcsCpJeIBmne57fJY/edit?usp=sharing

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5915

* CI

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -31,11 +31,15 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.alias.CredentialProvider;
 import org.apache.hadoop.security.alias.CredentialProviderFactory;
+import org.apache.zeppelin.antlr.ComplexSqlSplitVisitor;
+import org.apache.zeppelin.antlr.ComplexSqlVisitor;
+import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.interpreter.SingleRowInterpreterResult;
 import org.apache.zeppelin.interpreter.ZeppelinContext;
 import org.apache.zeppelin.interpreter.util.SqlSplitter;
 import org.apache.zeppelin.jdbc.hive.HiveUtils;
 import org.apache.zeppelin.tabledata.TableDataUtils;
+import org.apache.zeppelin.util.DateUtil;
 import org.apache.zeppelin.util.PropertiesUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -56,6 +60,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.LinkedHashMap;
+import java.util.Collections;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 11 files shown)


zhugezifang:
hi @Reamer @jongyoul @huage1994 ,could you help me to review this pr?

Reamer:
A few thoughts about your work:
 - All functionality is located in the JDBC interpreter. All new dependencies and classes should also be located there.
 - It seems that in case of a complex query you modify the dataset with the JDBC interpreter at the source. With your current implementation, the user is not aware of this right away.
 - You want to cache complex queries with an LRUCache. Depending on the size of the cache, queries may not be removed from the cache and the user may see stale data. I think a time based cache would be more appropriate here.
 - Have you ever heard of PreparedStatements?
 - At first view your configuration options are only JDBC interpreter configuration, therefore the configuration should be stored only in the JDBC interpreter and not as global ZeppelinConfiguration.

zhugezifang:
@Reamer  hi，firstly thanks for your advice

"All functionality is located in the JDBC interpreter. All new dependencies and classes should also be located there"---- it contains antlr4 ,and it also can be used in the feature in sql debug https://github.com/apache/zeppelin/pull/4598
but i am very sorry , the feature cost you  a lot of time ,but it can not be merged

if this feature of sql debug can be merge, the use of antlr can also be used in this feature.

'queries may not be removed from the cache and the user may see stale data' --- this is really a good question，it is i want to do in next step，i create temp table as the data, but if the origin data change or temp table very much ,it need to fix ,so i write the Design Document, the last step is to clean the temp table 

https://docs.google.com/document/d/1wruK0ZZ0XiriYOraFa5WYSz531pcsCpJeIBmne57fJY/edit

zhugezifang:
> A few thoughts about your work:
> 
> * All functionality is located in the JDBC interpreter. All new dependencies and classes should also be located there.
> * It seems that in case of a complex query you modify the dataset with the JDBC interpreter at the source. With your current implementation, the user is not aware of this right away.
> * You want to cache complex queries with an LRUCache. Depending on the size of the cache, queries may not be removed from the cache and the user may see stale data. I think a time based cache would be more appropriate here.
> * Have you ever heard of PreparedStatements?
> * At first view your configuration options are only JDBC interpreter configuration, therefore the configuration should be stored only in the JDBC interpreter and not as global ZeppelinConfiguration.

and could you help me to continue complete the feature of sql debug  https://github.com/apache/zeppelin/pull/4598

zhugezifang:
> A few thoughts about your work:
> 
> * All functionality is located in the JDBC interpreter. All new dependencies and classes should also be located there.
> * It seems that in case of a complex query you modify the dataset with the JDBC interpreter at the source. With your current implementation, the user is not aware of this right away.
> * You want to cache complex queries with an LRUCache. Depending on the size of the cache, queries may not be removed from the cache and the user may see stale data. I think a time based cache would be more appropriate here.
> * Have you ever heard of PreparedStatements?
> * At first view your configuration options are only JDBC interpreter configuration, therefore the configuration should be stored only in the JDBC interpreter and not as global ZeppelinConfiguration.

and the data cache is really for hive like this large data database (not  for mysql), as we know ,complex hive sql have shuffle and reducer ,it cost a lot of time

jongyoul:
Sorry for the late and short comment but how about making debug paragraph as a new interpreter? I mean, for instance, `%jdbc.debug`. Debugging jdbc paragraph itself is reasonable but I'm worried that this feature would be included in ZeppelinServer because this is only for jdbc interpreter and it won't work for other interpereters. WDYT?

jongyoul:
> and the data cache is really for hive like this large data database (not for mysql), as we know , hive sql will tranlate to mapreduce job， complex hive sql have shuffle and reducer ,it cost a lot of time

I agree that Zeppelin could help to reduce resource usage but I, as one of the maintainers, believe that we shouldn't change any user behaviors and mechanisms working as it is. Your approach might be better for some cases but it could have a kind of hidden steps and it could have another problem. If you would like to do it really, I believe that you can make your/your company's own interpreter. It can be merged into this repository but we, as reviewers, should consider not to make significant side effects with the highest priority.

zhugezifang:
> > and the data cache is really for hive like this large data database (not for mysql), as we know , hive sql will tranlate to mapreduce job， complex hive sql have shuffle and reducer ,it cost a lot of time
> 
> I agree that Zeppelin could help to reduce resource usage but I, as one of the maintainers, believe that we shouldn't change any user behaviors and mechanisms working as it is. Your approach might be better for some cases but it could have a kind of hidden steps and it could have another problem. If you would like to do it really, I believe that you can make your/your company's own interpreter. It can be merged into this repository but we, as reviewers, should consider not to make significant side effects with the highest priority.

ok, thanks for your advice, and could you help to review this pr of sql debug, https://github.com/apache/zeppelin/pull/4598
is it suitable?

PR #4620, opened 2023-06-18
Fixed minor formatting issues preventing the full build

zinal:
### What is this PR for?
Fixed minor issues in Java files formatting to make maven-checkstyle-plugin happy

### What type of PR is it?
Improvement

* Regular regressions tests

### Questions:

No license updates, no breaking changes, no need for extra docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,6 +97,9 @@ Thumbs.db
 .idea/
 *.iml
 
+# vscode project files
+.vscode/
+
 # maven target files
 target/
 **/target/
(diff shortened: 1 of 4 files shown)


Reamer:
I think it would be good if with this fix the checkstyle plugin is enabled.

zinal:
> I think it would be good if with this fix the checkstyle plugin is enabled.

It seems to be enabled, at least it was complaining without the fixes when I followed the general build instructions from the [docs](https://zeppelin.apache.org/docs/0.10.1/setup/basics/how_to_build.html)

Reamer:
I see. Which Maven command did you run?

zinal:
> I see. Which Maven command did you run?

mvn clean package -DskipTests

Reamer:
I am surprised that there are no errors in the CI. 
Do you see errors in the CI system? Maybe we should adjust the default build.

Reamer:
Thank you, I can reproduce the issue. 
Can you change the Goal `package` to `verify` in the following line? Then the CI system should run checkstyle. Unfortunately the phase `package` is not enough.
https://github.com/apache/zeppelin/blob/1e9c7fd0496f13b73c26dc3eef254c90c4a70ee7/.github/workflows/core.yml#L486

Reamer (approved):

PR #4624, opened 2023-06-22
[ZEPPELIN-5934] Check notebook folder permissions before allowing to rename, remove or restore it

idzikovsky:
### What is this PR for?
Users who are able to see notes in some directory can rename, move to trash and remove from trash that directory without being owner or having write permissions for notes in that directory.
This is resolved in first commit.

Secondly, after renaming directory to the name of already existing directory, old notes in target directory become unaccessible (in fact those notes are removed in file system, but still visible in UI as they are present in NoteManager registry).
This addresses 2nd point in ZEPPELIN-5333.
Fixed in 2nd commit.

### What type of PR is it?
Improvement

### What is the Jira issue?
* [ZEPPELIN-5934](https://issues.apache.org/jira/browse/ZEPPELIN-5934)
* [ZEPPELIN-5333.](https://issues.apache.org/jira/browse/ZEPPELIN-5333.)

* I've tested this manually by creating some note in directory, giving read permission to it for some other user, and checking whether that user can move to trash, remove, rename or restore directory with that note.
* I've checked and it does not seem like there are available infrastructure in `NotebookServiceTest` or in `NoteManagerTest` to cover multi-user scenarios.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -694,6 +694,11 @@ public class NotebookService {
       return;
     }
     try {
+      NoteManager.Folder folder = notebook.getFolder(folderPath);
+      if (!checkFolderPermission(folder, Permission.OWNER, Message.OP.RESTORE_FOLDER, context, callback)) {
+        return;
+      }
+
       String destFolderPath = folderPath.replace("/" + NoteManager.TRASH_FOLDER, "");
       notebook.moveFolder(folderPath, destFolderPath, context.getAutheInfo());
       callback.onSuccess(null, context);
@@ -708,6 +713,12 @@ public class NotebookService {
                          ServiceCallback<Void> callback) throws IOException {
 
     try {
+      String folderPath = "/" + NoteManager.TRASH_FOLDER;
+      NoteManager.Folder folder = notebook.getFolder(folderPath);
+      if (!checkFolderPermission(folder, Permission.OWNER, Message.OP.RESTORE_ALL, context, callback)) {
+        return;
+      }
+
       notebook.restoreAll(context.getAutheInfo());
       callback.onSuccess(null, context);
     } catch (IOException e) {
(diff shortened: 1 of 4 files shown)


idzikovsky:
Sorry, I made this PR on top of 0.10.1 version, and it seems like something has been changed in this filed in master branch since then.
Now I have local fix of this issue for master branch. I'll update the PR as soon as I have successful run of tests.

idzikovsky:
Issues in my last comment were resolved in the latest commit.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -1192,24 +1203,38 @@ public void moveFolderToTrash(String folderPath,
...
 
-    String destFolderPath = "/" + NoteManager.TRASH_FOLDER + "/" + folderPath;
-    if (notebook.containsNote(destFolderPath)) {
-      destFolderPath = destFolderPath + " " +
-          TRASH_CONFLICT_TIMESTAMP_FORMATTER.format(Instant.now());
-    }
+    try {
+      NoteManager.Folder folder = notebook.getFolder("/" + folderPath);
+      if (!checkFolderPermission(folder, Permission.OWNER, Message.OP.MOVE_FOLDER_TO_TRASH, context, callback)) {
+        return;
```
jongyoul, line 1212: before returning, shouldn't we call `callback.onFailure`?

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -1497,4 +1530,78 @@ private <T> boolean checkPermission(String noteId,
...
+      return true;
+    } else {
+      String errorMsg = String.format(
+              "Insufficient %s privileges to '%s' note.\n" +
+              "Allowed users or roles: %s\n" +
+              "But the user %s belongs to: %s",
+              permission, noteInfo.getNoteName(),
+              allowed,
+              context.getAutheInfo().getUser(), context.getUserAndRoles());
+      callback.onFailure(new ForbiddenException(errorMsg), context);
```
jongyoul, line 1579: Oh, I see this code but we'd better move this code outside of this method because it's only adopted when failure. I think it would be good to make it clear the purpose of this method and `callbak.onFailure` doesn't seem to be the purpose of this method.

jongyoul (approved):
Thank you for your contribution. It looks good to me except for the minor comments. Could you please address the comment?

jongyoul:
BTW, there are no tests for the features. Could you please add tests for your feature?

Reamer (changes requested):
I think it is not a good idea to add the logic regarding the rights in the service. Furthermore, information on Notes are leaked, where a user may not have access at all.

PR #4652, opened 2023-09-11
[ZEPPELIN-5946][FOLLOWUP] Use Spark 3.4.1 in default

pan3793:
### What is this PR for?

Followup ZEPPELIN-5946, Use Spark 3.4.1 in default

### What type of PR is it?

Improvement
Documentation

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5946

Pass GA

### Questions:

### DIFF
--- a/docs/setup/basics/how_to_build.md
+++ b/docs/setup/basics/how_to_build.md
@@ -93,7 +93,7 @@ Set scala version (default 2.12). Available profiles are
 
 #### Spark Interpreter
 
-To be noticed, the spark profiles here only affect the uni test (no need to specify `SPARK_HOME`) of spark interpreter. 
+To be noticed, the spark profiles here only affect the unit test (no need to specify `SPARK_HOME`) of spark interpreter. 
 Zeppelin doesn't require you to build with different spark to make different versions of spark work in Zeppelin.
 You can run different versions of Spark in Zeppelin as long as you specify `SPARK_HOME`. Actually Zeppelin supports all the versions of Spark from 3.2 to 3.4.
 
--- a/docs/setup/basics/how_to_build.md
+++ b/docs/setup/basics/how_to_build.md
@@ -84,7 +84,7 @@ You can directly start Zeppelin by running the following command after successfu
 
 To be noticed, this scala profile affect the modules (e.g. cassandra) that use scala except Spark interpreter (Spark interpreter use other profiles to control its scala version, see the doc below).
 
-Set scala version (default 2.12). Available profiles are
+Set scala version (default 2.11). Available profiles are
 
 ```
 -Pscala-2.11
--- a/docs/setup/basics/how_to_build.md
+++ b/docs/setup/basics/how_to_build.md
@@ -84,7 +84,7 @@ You can directly start Zeppelin by running the following command after successfu
 
 To be noticed, this scala profile affect the modules (e.g. cassandra) that use scala except Spark interpreter (Spark interpreter use other profiles to control its scala version, see the doc below).
 
-Set scala version (default 2.10). Available profiles are
+Set scala version (default 2.12). Available profiles are
 
 ```
 -Pscala-2.11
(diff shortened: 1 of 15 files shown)


spark/scala-2.13/src/main/scala/org/apache/zeppelin/spark/SparkScala213Interpreter.scala
```diff
@@ -129,7 +129,7 @@ class SparkScala213Interpreter(conf: SparkConf,
                           context: InterpreterContext): java.util.List[InterpreterCompletion] = {
     scalaCompletion.complete(buf.substring(0, cursor), cursor)
       .candidates
-      .map(e => new InterpreterCompletion(e.defString, e.defString, null))
+      .map(e => new InterpreterCompletion(e.name, e.name, null))
```
pan3793, line 132: changed in https://github.com/scala/scala/pull/9510

Reamer:
Thx for the follow-up.

pan3793:
The failed Kotlin test seems to be https://youtrack.jetbrains.com/issue/KT-38325

I suppose we must upgrade Kotlin first to address it. Unfortunately, the Kotlin interpreter uses a very old version (1.3.x, the latest is 1.9.x), and couple with lots of unstable APIs.

Reamer:
@pan3793 
Is this PR ready for review?

pan3793:
@Reamer not yet, I need to revert some changes.

Sorry, I'm quite busy these days, will find the time to finish it ASAP.

docs/setup/basics/how_to_build.md
```diff
@@ -95,7 +95,7 @@ Set scala version (default 2.10). Available profiles are
 
 To be noticed, the spark profiles here only affect the uni test (no need to specify `SPARK_HOME`) of spark interpreter.
```
matthias-koch, line 96:
```suggestion
To be noticed, the spark profiles here only affect the unit test (no need to specify `SPARK_HOME`) of spark interpreter. 
```
Found a little typo :-)

spark/scala-2.13/src/main/scala/org/apache/zeppelin/spark/SparkScala213Interpreter.scala
pan3793: Reconsidered, the change is not safe. 

https://github.com/scala/scala/pull/9510 changed in 2.13.6, but since Zeppelin supports Spark 3.2 and above, we should use refection here to make it work on all versions

- Spark 3.2.4 uses Scala 2.13.5
- Spark 3.3.3 uses Scala 2.13.8
- Spark 3.4.1 uses Scala 2.13.8
- Spark 3.5.0 uses Scala 2.13.8

pan3793:
The profiles `scala-2.11` and `scala-2.12` defined at root `pom.xml` seems only affects the livy module's test, if so, do we really need to maintain such profiles? Why not hardcode in `livy/pom.xml`(seems livy is not much active, the latest version only support Scala 2.11)

pan3793:
cc @Reamer

Reamer:
> The profiles `scala-2.11` and `scala-2.12` defined at root `pom.xml` seems only affects the livy module's test, if so, do we really need to maintain such profiles? Why not hardcode in `livy/pom.xml`(seems livy is not much active, the latest version only support Scala 2.11)

The profiles `scala-2.11` and `scala-2.12` set the properties `scala.version` and `scala.binary.version`.

I see accesses to these properties in the Zeppelin Display Module and Zeppelin Distribution Module. Unfortunately, I don't currently know why Scala dependencies are needed in the Zeppelin Distribution Module.

https://github.com/apache/zeppelin/blob/85bb3b4ca2d4a5c79194f8c12a90b732d777ccc5/zeppelin-display/pom.xml#L34-L55

https://github.com/apache/zeppelin/blob/85bb3b4ca2d4a5c79194f8c12a90b732d777ccc5/zeppelin-distribution/pom.xml#L49-L73

pan3793:
Oh, I missed the Zeppelin Display Module(it does not appear when I do global search using IDEA), it leverages the scala-xml to build `AngularElem`, that's why scala is required

pan3793:
Emm, is "Zeppelin Display Module" only used by "Zeppelin: Spark Interpreter"? If so, I suppose the Zeppelin core becomes scala free if we can merge this module into "Zeppelin: Spark Interpreter"

Reamer:
As far as I know, the Zeppelin display module is only used by the Spark interpreter.

If it is possible we can merge the Display Module to the Spark Interpreter Module.

pan3793:
> As far as I know, the Zeppelin display module is only used by the Spark interpreter.
> 
> If it is possible we can merge the Display Module to the Spark Interpreter Module.

Thanks for information, would like a try in another PR.

I limit this PR' scope to the spark related stuffs, CI already passed, and [the last commit](https://github.com/apache/zeppelin/pull/4652/commits/c6bcc863771eb53356fb800821bb9ef5ebd5ca26) only touch the docs

Reamer (approved):
LGTM

Reamer:
I will merge this pull request on Friday as long as no other comment is received.

jongyoul (approved):
I missed this PR. LGTM

PR #4658, opened 2023-09-14
[ZEPPELIN-5964] Fix renaming bug

matthias-koch:
### What is this PR for?
If a note is renamed by changing only the capitalization, the note is copied at the file system level when it is executed again.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-5964

* New unit test were added

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
@@ -247,7 +247,7 @@ public class NoteManager {
     this.notebookRepo.move(noteId, notePath, newNotePath, subject);
 
     // Update path of the note
-    if (!notePath.equals(newNotePath)) {
+    if (!StringUtils.equals(notePath, newNotePath)) {
       processNote(noteId,
         note -> {
           note.setPath(newNotePath);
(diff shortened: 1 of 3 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
```diff
@@ -247,7 +247,7 @@ public void moveNote(String noteId,
     this.notebookRepo.move(noteId, notePath, newNotePath, subject);
 
     // Update path of the note
-    if (!StringUtils.equalsIgnoreCase(notePath, newNotePath)) {
+    if (!notePath.equals(newNotePath)) {
```
Reamer, line 250: `StringUtils.equals(x, y)` is null safe and should be preferred.

Reamer (commented):
Except for one small comment, I think the PR is very good.
I don't know if all notebook storage systems are case sensitive. This could lead to problems. Could you check this?

jongyoul (approved):
I agreed with @Reamer's concern but basically, it's good

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
matthias-koch: Git: case sensitive
Hadoop: case sensitive
S3: case sensitive
Azure: case sensitive
Google Cloud Storage: case sensitive
Aliyun OSS: case sensitive
MongoDB: case sensitive
Github: case sensitive

It seems we are safe.

Reamer: awesome

Reamer (approved):
LGTM

PR #4667, opened 2023-10-08
[MINOR] Update deb maintainer information

pan3793:
### What is this PR for?

Update deb maintainer information from the certain developer of this project to Apache Zeppelin PMC, this is a common practice for Apache projects.

### What type of PR is it?

Documentation

### Todos

N/A

### What is the Jira issue?

MINOR

Review

### Questions:

### DIFF
--- a/zeppelin-distribution/src/deb/control/control
+++ b/zeppelin-distribution/src/deb/control/control
@@ -21,6 +21,6 @@ Version: [[version]]-[[buildNumber]]
 Section: misc
 Priority: optional
 Architecture: all
-Maintainer: Lee moon soo <leemoonsoo@gmail.com>
+Maintainer: Apache Zeppelin PMC <dev@zeppelin.apache.org>
 Description: [[name]]
 Distribution: development


pan3793:
cc @Reamer

Reamer:
I have never used the Debian package. I do not know if this change will cause problems. Maybe it is also better if we remove deb support. ping @Leemoonsoo

jongyoul:
This is what I made and it was used in Apache Bigtop before. I checked it now simply and it's not used anymore. I think it could be removed.

pan3793:
@jongyoul do you mean remove whole `zeppelin-distribution/src/deb`?

jongyoul:
@pan3793 Yes, I think so. BTW, to do that, it would be better to make a ticket to handle it.

pan3793:
@jongyoul sure, I'm going to create a JIRA

pan3793:
Close and move to ZEPPELIN-5967 (https://github.com/apache/zeppelin/pull/4670)

PR #4665, opened 2023-10-08
[ZEPPELIN-5966] Support Spark 3.5

pan3793:
### What is this PR for?

Spark 3.5.0 was released a few weeks ago, this PR tries to add Spark 3.5 support in Zeppelin.
https://spark.apache.org/releases/spark-release-3-5-0.html

### What type of PR is it?

Improvement

### What is the Jira issue?

ZEPPELIN-5966

### How should this be tested?

Update GitHub workflow to cover Spark 3.5 tests

### Questions:
* Does the license files need to update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? Yes, updated correspondingly.

PR #4668, opened 2023-10-08
[ZEPPELIN-5946][FOLLOWUP] Clean up code for legacy Spark versions

pan3793:
### What is this PR for?

Clean up unreachable code for legacy Spark 1.x and 2.x versions, since Zeppelin only supports Spark 3.2 ~ 3.4 now.

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5946

Pass CI

### Questions:

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
@@ -70,7 +70,6 @@ public class PythonInterpreter extends Interpreter {
   private ZeppelinContext zeppelinContext;
   // set by PythonCondaInterpreter
   private String condaPythonExec;
-  private boolean usePy4jAuth = false;
 
   public PythonInterpreter(Properties property) {
     super(property);
@@ -116,7 +115,6 @@ public class PythonInterpreter extends Interpreter {
     }
 
     try {
-      this.usePy4jAuth = Boolean.parseBoolean(getProperty("zeppelin.py4j.useAuth", "true"));
       createGatewayServerAndStartScript();
     } catch (IOException e) {
       LOGGER.error("Fail to open PythonInterpreter", e);
@@ -132,8 +130,7 @@ public class PythonInterpreter extends Interpreter {
     // container can also connect to this gateway server.
     String serverAddress = PythonUtils.getLocalIP(properties);
     String secret = PythonUtils.createSecret(256);
-    this.gatewayServer = PythonUtils.createGatewayServer(this, serverAddress, port, secret,
-        usePy4jAuth);
+    this.gatewayServer = PythonUtils.createGatewayServer(this, serverAddress, port, secret);
     gatewayServer.start();
 
     // launch python process to connect to the gateway server in JVM side
@@ -149,9 +146,7 @@ public class PythonInterpreter extends Interpreter {
 
     outputStream = new InterpreterOutputStream(LOGGER);
     Map<String, String> env = setupPythonEnv();
-    if (usePy4jAuth) {
-      env.put("PY4J_GATEWAY_SECRET", secret);
-    }
+    env.put("PY4J_GATEWAY_SECRET", secret);
     if (LOGGER.isInfoEnabled()) {
       LOGGER.info("Launching Python Process Command: {} {}",
           cmd.getExecutable(), StringUtils.join(cmd.getArguments(), " "));
--- a/python/src/main/java/org/apache/zeppelin/python/PythonUtils.java
+++ b/python/src/main/java/org/apache/zeppelin/python/PythonUtils.java
@@ -39,17 +39,12 @@ public class PythonUtils {
                                                   String secretKey) throws IOException {
     LOGGER.info("Launching GatewayServer at {}:{}", serverAddress, port);
     try {
-      Class clz = Class.forName("py4j.GatewayServer$GatewayServerBuilder", true,
-          Thread.currentThread().getContextClassLoader());
-      Object builder = clz.getConstructor(Object.class).newInstance(entryPoint);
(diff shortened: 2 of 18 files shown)


python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -116,7 +115,6 @@ public void open() throws InterpreterException {
     }
 
     try {
-      this.usePy4jAuth = Boolean.parseBoolean(getProperty("zeppelin.py4j.useAuth", "true"));
```
pan3793, line 119: This change is based on the fact that `isSecretSupported` is always true for Spark 3.2 and above.

property `zeppelin.py4j.useAuth` is only used internally, and is not exposed to the end user, the removal should be safe.

pan3793:
cc @jongyoul @Reamer

python/src/main/java/org/apache/zeppelin/python/PythonUtils.java
```diff
@@ -37,35 +36,22 @@ public class PythonUtils {
...
-          GatewayServer.DEFAULT_PYTHON_PORT,
-          InetAddress.getByName(serverAddress),
-          InetAddress.getByName(serverAddress),
-          GatewayServer.DEFAULT_CONNECT_TIMEOUT,
-          GatewayServer.DEFAULT_READ_TIMEOUT,
-          (List) null);
+                                                  String secretKey) throws IOException {
+    LOGGER.info("Launching GatewayServer at {}:{}", serverAddress, port);
+    try {
+      Class clz = Class.forName("py4j.GatewayServer$GatewayServerBuilder", true,
```
Reamer, line 42: Can you document here why reflection is used here.

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
Reamer: :heavy_check_mark:

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -294,37 +292,17 @@ private String detectSparkScalaVersion(String sparkHome, Map<String, String> env
...
-        return "2.10";
-      } catch (ClassNotFoundException e) {
-        return "2.11";
-      }
+    File sparkJarsFolder = new File(sparkHome + "/jars");
+    boolean sparkRepl212Exists =
+            Stream.of(sparkJarsFolder.listFiles()).anyMatch(file -> file.getName().contains("spark-repl_2.12"));
+    boolean sparkRepl213Exists =
+            Stream.of(sparkJarsFolder.listFiles()).anyMatch(file -> file.getName().contains("spark-repl_2.13"));
+    if (sparkRepl212Exists) {
```
Reamer, line 300: I like this approach. Do we want to throw an exception as soon as multiple versions are found?

Reamer (approved):
I have looked through the PR but have not validated that `py4j.useAuth` is also used in other sections.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
pan3793: I think it's fine to suppose that the user provides a valid SPARK_HOME, otherwise, things could be much more complicated. And, a valid Spark installation won't contain both `spark-repl_2.12` and `spark-repl_2.13`

python/src/main/java/org/apache/zeppelin/python/PythonUtils.java
pan3793: I'm not very sure, it existed here without documents.

Maybe we can eliminate the reflection now, let me try

pan3793: After eliminating the reflection, CI result is good

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
Reamer: In the old code, an exception is thrown for unclean installation. I would still like to see such error handling in the code.
```
      if (sparkAssemblyJars.length > 1) {
        throw new Exception("Multiple spark assembly file found in SPARK_HOME: " + sparkHome);
      }
```

python/src/main/java/org/apache/zeppelin/python/PythonUtils.java
Reamer: Great.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
pan3793: OK, let me add such check too

pan3793: added check code in https://github.com/apache/zeppelin/pull/4668/commits/0c30040a2766221542b9363abbd53974666bb2c0

Reamer (approved):
LGTM

jongyoul (approved):

pan3793:
@jongyoul the failed CI is irrelevant, could you merge this one? I'm going to rebase the #4665 once this gets in

PR #4690, opened 2023-11-09
[ZEPPELIN-5979] Support and use Python 3.9 in default

pan3793:
### What is this PR for?

Currently, Zeppelin only tests with Python 3.7(default) and 3.8.

According to https://endoflife.date/, Python 3.7 is EOL. This PR adds the Python 3.9 test with minimal tunes. Although Python 3.12 is available, I conservatively just moved to Python 3.9 for safety (I'm not a Python expert), also Python 3.9 is the first version officially supporting Apple Sillicon.

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5979

Pass GA.

### Questions:

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -68,12 +68,12 @@ jobs:
         run: ./mvnw install -Pbuild-distr -DskipTests -pl zeppelin-server,zeppelin-web,spark-submit,spark/scala-2.12,spark/scala-2.13,markdown,angular,shell -am -Phelium-dev -Pexamples -P${{ matrix.hadoop }} ${MAVEN_ARGS}
       - name: install and test plugins
         run: ./mvnw package -pl zeppelin-plugins -amd ${MAVEN_ARGS}
-      - name: Setup conda environment with python 3.7 and R
+      - name: Setup conda environment with python 3.9 and R
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: python_3_with_R
-          environment-file: testing/env_python_3.7_with_R.yml
-          python-version: 3.7
+          environment-file: testing/env_python_3.9_with_R.yml
+          python-version: 3.9
           miniforge-variant: Mambaforge
           channels: conda-forge,defaults
           channel-priority: true
@@ -119,12 +119,12 @@ jobs:
             ${{ runner.os }}-zeppelin-
       - name: install environment
         run: ./mvnw install -DskipTests -am -pl ${INTERPRETERS} -Pscala-2.11 ${MAVEN_ARGS}
-      - name: Setup conda environment with python 3.7 and R
+      - name: Setup conda environment with python 3.9 and R
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: python_3_with_R_and_tensorflow
           environment-file: testing/env_python_3_with_R_and_tensorflow.yml
-          python-version: 3.7
+          python-version: 3.9
           miniforge-variant: Mambaforge
           channels: conda-forge,defaults
           channel-priority: true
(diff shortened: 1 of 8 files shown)


testing/env_python_3_with_R_and_tensorflow.yml
```diff
@@ -35,4 +36,4 @@ dependencies:
   - r-irkernel
   - r-shiny
   - r-googlevis
-  - tensorflow=1.13
+  - tensorflow
```
pan3793, line 39: It's only used in submarine testing.

BTW, does the community consider removing submarine support? 

- submarine was listed at [here](https://cwiki.apache.org/confluence/display/ZEPPELIN/Interpreter+Maintenance) with no clear decision
- there is no functionality change / bugfix in submarine module since 2020
- submarine already provided the built [notebook](https://submarine.apache.org/docs/designDocs/notebook-implementation) in the new version, and submarine docs does not mention zeppelin

thus, I suppose we can remove it?

pan3793: cc @jongyoul @Reamer @xunliu

pan3793:
also cc @zjffdu and @huage1994

.github/workflows/core.yml
```diff
@@ -342,8 +353,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: [ 3.7, 3.8 ]
+        python: [ 3.9 ]
```
huage1994, line 356: why not modify to  [ 3.7, 3.8, 3.9 ] ?

pan3793: Because Flink 1.15 does not support Python 3.9

.github/workflows/core.yml
```diff
@@ -243,7 +248,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        flink: [115, 116, 117]
+        python: [ 3.9 ]
+        flink: [116, 117]
+        include:
+          # Flink 1.15 supports Python 3.6, 3.7, and 3.8
```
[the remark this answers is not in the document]
pan3793: why? it's true

.github/workflows/core.yml
pan3793: I think we don't need to test all combinations of Java and Python.

`all_java_version * default_python_version` plus `default_java_version * all_python_version_except_default_one` is sufficient.

.github/workflows/core.yml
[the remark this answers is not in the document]
huage1994: Umm,  maybe I am nitpicking.   I thought you were going to test all the three python versions, following the comments.

.github/workflows/core.yml
huage1994: > I think we don't need to test all combinations of Java and Python.
> 
> `all_java_version * default_python_version` plus `default_java_version * all_python_version_except_default_one` is sufficient.

Thank you very much for explaining.

.github/workflows/core.yml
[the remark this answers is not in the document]
pan3793: Flink 1.15 is an exception, it does not work with Python 3.9, that's why I leave comments to explain that.

huage1994 (approved):

testing/env_python_3_with_R_and_tensorflow.yml
Reamer: I do not use submarine. As long as it doesn't cause any problems, we should keep it.

Reamer (approved):

testing/env_python_3_with_R_and_tensorflow.yml
pan3793: @Reamer OK, thanks for your advice.

PR #4691, opened 2023-11-13
[ZEPPELIN-5969] Remove Hadoop2 and move to Hadoop3 shaded client

pan3793:
### What is this PR for?

This PR is based on https://github.com/apache/zeppelin/pull/4674.

This PR aligns Hadoop dependencies version (3.3.6) in all Zeppelin modules, additionally, it switches to Hadoop shaded client (introduced in HADOOP-11804 since Hadoop 3.0.0) which is recommended by Hadoop, Spark also switched to Hadoop shaded client in SPARK-33212 (3.2.0), which dramatically reduce the Hadoop dependency management efforts, basically, only 3 hadoop jars are required now:
- `hadoop-client-api`
- `hadoop-client-runtime`
- `hadoop-client-minicluster` (only for testing)

There is a JDK 10 compatible issue fixed in HIVE-21508 (Hive 2.3.7), so this PR also upgrades Flink's Hive version from 2.3.4 to 2.3.7. Hive 2.3.7 is chosen because Flink supports Hive 2.3.8 and 2.3.9 in FLINK-26739 (1.16.0) while Zeppelin supports Flink 1.15~1.17.

### What type of PR is it?

Improvement

### What is the Jira issue?

ZEPPELIN-5969

The test cases are modified corresponding, and all GHA tests are passed

### Questions:


rlang/pom.xml
```diff
@@ -116,18 +116,10 @@
...
         </dependency>
-
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
-            <scope>compile</scope>
-        </dependency>
-
         <dependency>
```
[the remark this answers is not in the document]
pan3793: Thanks for checking. I think `hadoop-client-runtime` will pull `hadoop-client-api`, thus there should be no problem.

This PR is not ready (I'm a little busy these days, I may can continue this work next week)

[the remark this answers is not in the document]
matthias-koch: `hadoop-client-api` is a runtime dependency of `hadoop-client-runtime`, so `maven-shaded-plugin` should include it. However, I have created the R interpreter package on my local machine and without adding the dependency explicitly, `hadoop-client-api` is not present in the shaded jar. (But maybe I am doing somthing wrong.)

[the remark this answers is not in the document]
Reamer: How do you know that the maven-shade plugin also adds runtime dependencies? After a first overview I think that the dependency `hadoop-client-api` has to be added manually.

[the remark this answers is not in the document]
matthias-koch: The description at https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html says:

> Requires dependency resolution of artifacts in scope: runtime

But I am really not sure, how all dependencies are resolved.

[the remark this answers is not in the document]
pan3793: @matthias-koch the `maven-shaded-plugin` will explicitly print the logs when you perform the `mvn package` command, something like

```
[INFO] --- maven-shade-plugin:3.4.1:shade (default) @ zeppelin-flink-cmd ---
[INFO] Including org.apache.zeppelin:zeppelin-shell:jar:0.11.0-SNAPSHOT in the shaded jar.
[INFO] Including com.google.code.gson:gson:jar:2.8.9 in the shaded jar.
[INFO] Including org.apache.commons:commons-lang3:jar:3.12.0 in the shaded jar.
[INFO] Excluding org.apache.zeppelin:zeppelin-interpreter-shaded:jar:0.11.0-SNAPSHOT from the shaded jar.
...
```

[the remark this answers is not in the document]
Reamer: I do not see the dependency `hadoop-client-api` in the output.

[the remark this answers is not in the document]
pan3793: alright, I know where the issue comes from. 

Do we want to respect the `hadoop.deps.scope` if we switch to the Hadoop Shaded client?

Note: Spark drops such profiles and always ships Hadoop Shaded client jars into the binary artifacts.

```
<hadoop.deps.scope>provided</hadoop.deps.scope>
...
<profile>
  <id>include-hadoop</id>
  <properties>
    <hadoop.deps.scope>compile</hadoop.deps.scope>
  </properties>
</profile>
```

pan3793:
I basically have fixed all compile and test issues, the next step is to split it into several small PRs to speed up the review process.

I think we should start with the interpreter modules one by one, and then zengine, server and other modules, eventually dropping the `hadoop2` profile and updating docs.

@Reamer could you give some advice?

rlang/pom.xml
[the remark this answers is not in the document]
pan3793: > As far as I understand it correctly, YARN has already loaded the Hadoop dependencies and therefore they don't need to be in Zeppelin additionally.

There is a switch in YARN to enable/disable Hadoop class population in the containers.

> we should make sure that we do not deliver the library more than once

QQ, I understand we should not include Hadoop classes in plugins, because they will be loaded into the same JVM with Zeppelin server, so that they can share the Hadoop classes. What about the interpreteres? I assume the interpreters are always run in dedicated JVMs, so Hadoop classes seem always necessary (except for those runtimes who already provided Hadoop classes, e.g. Spark, Flink)?

Reamer:
I would prefer a larger PR. Where individual tasks are contained in commits. It was clear that the drop of Hadoop2 is very large. Thank you for your work so far.

I think it's great that you have deleted all the excludes in the parent `pom.xml`, that makes the file much more readable.

Btw. I do not insist on co-authorship.

pan3793:
unfortunately, I found this IT didn't run properly now, see https://github.com/apache/zeppelin/pull/4699, we may need to postpone this PR after recovering IT

flink/flink-scala-2.12/pom.xml
```diff
@@ -314,73 +300,163 @@
...
     <dependency>
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-metastore</artifactId>
       <version>${hive.version}</version>
       <scope>provided</scope>
       <exclusions>
         <exclusion>
-          <artifactId>hadoop-auth</artifactId>
-          <groupId>org.apache.hadoop</groupId>
+          <groupId>org.apache.hive</groupId>
```
pan3793, line 310: the exclusion list is borrowed from Apache Flink's pom.xml

flink/flink-scala-2.12/pom.xml
```diff
@@ -578,6 +999,13 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>com.lmax</groupId>
+      <artifactId>disruptor</artifactId>
```
pan3793, line 1004: hive enabled log4j2 async log by default, which requires disruptor available on classpath

flink/flink-scala-2.12/pom.xml
```diff
@@ -904,31 +1332,5 @@
...
-      <activation>
-        <activeByDefault>true</activeByDefault>
-      </activation>
-      <properties>
-        <hive.version>2.3.4</hive.version>
-        <hiverunner.version>4.0.0</hiverunner.version>
-      </properties>
-    </profile>
-    <profile>
-      <id>hive1</id>
```
pan3793, line 919: Hive 1.x is officially dead https://lists.apache.org/thread/c7lyv2c2phzyspov48yxjnvks944tmkd, nor support Hadoop 3 neither

hbase/pom.xml
```diff
@@ -33,8 +33,6 @@
   <properties>
     <!--library versions-->
     <interpreter.name>hbase</interpreter.name>
-    <hbase.hbase.version>2.4.12</hbase.hbase.version>
```
pan3793, line 36: the HBase interpreter is based on shell command, neither HBase nor Hadoop deps are required

rlang/pom.xml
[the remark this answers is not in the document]
pan3793: I revised this question, the rlang interpreter can work both independently and combine with SparkR, for independent case, Hadoop deps are not required, for SparkR case, the Hadoop jars are handled by Spark runtime, either ship Hadoop jars in SPARK_HOME/jars, or setup HADOOP_CLASSPATH properly, so the rlang interpreter don't need to ship Hadoop deps

pan3793:
@Reamer it's ready for review, please take a look when you have time

pom.xml
```diff
@@ -139,17 +139,12 @@
...
     <micrometer.version>1.6.0</micrometer.version>
+    <findbugs.jsr305.version>3.0.2</findbugs.jsr305.version>
 
-    <hadoop2.7.version>2.7.7</hadoop2.7.version>
-    <hadoop3.0.version>3.0.3</hadoop3.0.version>
-    <hadoop3.1.version>3.1.3</hadoop3.1.version>
     <hadoop3.2.version>3.2.4</hadoop3.2.version>
     <hadoop3.3.version>3.3.6</hadoop3.3.version>
-    <hadoop.version>${hadoop2.7.version}</hadoop.version>
+    <hadoop.version>${hadoop3.3.version}</hadoop.version>
```
pan3793, line 146: Hadoop 3.2 is EOL, from the community perspective, we'd better choose a supported version.
https://lists.apache.org/thread/ybygrhvqok0f44s7yzm6c24n0b0s727s

pan3793: another reason is, as we use the hadoop-minicluster to launch Mini YARN cluster for testing in some modules, while since Spark 3.2.0, the pre-built Spark tgz ships Hadoop 3.3.x jars, Hadoop requires that client version <= server version, so we must use Hadoop 3.3 to launch Mini YARN cluster to allow Spark works properly.

flink/flink-scala-2.12/pom.xml
Reamer: A code comment would be useful here.

alluxio/pom.xml
```diff
@@ -68,15 +68,27 @@
...
+                    <artifactId>hadoop-client</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <version>3.2.4</version>
+            <artifactId>hadoop-client-api</artifactId>
```
Reamer, line 81: The `hadoop-client-api` is a transitive dependency of the `hadoop-client-runtime.` We should remove the explicit definition of `hadoop-client-api`.
grafik

hbase/pom.xml
Reamer: Quite strange, but I agree.

hbase/pom.xml
```diff
@@ -53,26 +51,6 @@
...
-      <version>${hbase.hbase.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.hbase</groupId>
-      <artifactId>hbase-annotations</artifactId>
-      <version>${hbase.hbase.version}</version>
-    </dependency>
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
```
Reamer, line 56: I think that `protobuf` and `jline` are also unnecessary.

pom.xml
```diff
@@ -402,612 +396,45 @@
...
-          <exclusion>
-            <groupId>log4j</groupId>
-            <artifactId>log4j</artifactId>
-          </exclusion>
-        </exclusions>
       </dependency>
 
       <dependency>
         <groupId>org.apache.hadoop</groupId>
         <artifactId>hadoop-yarn-api</artifactId>
```
Reamer, line 420: Is this dependency used anywhere?
Same for `hadoop-yarn-client`, `hadoop-yarn-common`, `hadoop-yarn-server-tests`, `hadoop-common` and `hadoop-common` (test-classifier)

rlang/pom.xml
```diff
@@ -38,7 +38,7 @@
 
         <spark.archive>spark-${spark.version}</spark.archive>
         <spark.bin.download.url>
-            https://archive.apache.org/dist/spark/${spark.archive}/${spark.archive}-bin-without-hadoop.tgz
+            https://archive.apache.org/dist/spark/${spark.archive}/${spark.archive}-bin-hadoop3.tgz
```
Reamer, line 41: Why this change? As I see it, only R libraries are needed and extracted from the download.

Reamer (changes requested):
I really like this PR. It removes a lot of unnecessary legacy from the Zeppelin code. I have set a few comments that I would like to have clarified before the merge.

alluxio/pom.xml
pan3793: I'm not sure if it's a good idea.
Technically, your suggestion is feasible and makes POM cleaner, but typically, we should explicitly declare the dependency that we consume directly.

flink/flink-scala-2.12/pom.xml
pan3793: added

hbase/pom.xml
pan3793: yes, actually, only `jruby-complete` is required here

pom.xml
pan3793: they're useless, removed

rlang/pom.xml
pan3793: we have `SparkRBackend` and `RInterpreterTest#testSparkRInterpreter`.

Spark core requires Hadoop client classes to run, previously, we forcibly shipped the Hadoop classes into the Rlang interpreter, IMO this is not the right approach, the Spark runtime should handle Hadoop classes.

alluxio/pom.xml
Reamer: Yes, I understand the approach and can understand it. Thank you very much

rlang/pom.xml
Reamer: I think you have not understood my comment or I have overlooked a place where the files are used.
In my opinion, the downloaded files are only used in the Maven execution maven-resources-plugin:copy-sparkr-files.
Only a small fraction of the files are copied here. Each file under `/R/lib`. This folder is contained in both `spark-bin-hadoop3.tgz` and `spark-bin-without-hadoop.tgz`.

I don't see the hadoop libraries from this download being used anywhere.

pan3793: @Reamer You are right, the CI result looks good.

pan3793:
@Reamer all failed tests are known flaky tests, this patch should be good to go :)

Reamer (approved):

Reamer:
I will merge the pull request on Wednesday as long as no further comments are received.

Armadik:
Could this change break the build? I try to collect and get an error

` WARN [2024-05-23 17:04:06,762] ({main} WebAppContext.java[doStart]:533) - Failed startup of context o.e.j.w.WebAppContext@4816278d{/,jar:file:///opt/zeppelin/zeppelin-web-0.12.0-SNAPSHOT.war!/,STOPPED}{/opt/zeppelin/zeppelin-web-0.12.0-SNAPSHOT.war}
java.io.FileNotFoundException: JAR entry WEB-INF/lib/hadoop-client-api-3.3.6.jar!/ not found in /opt/zeppelin/zeppelin-web-0.12.0-SNAPSHOT.war
`

pan3793:
@Armadik mind providing a reproducible step? e.g. build command, start command, OS platform, JDK version, etc.

Armadik:
I see an error when running the **zeppelin.sh** script

**Ubuntu 22.04.4 LTS**
`apt update
apt install -y curl git maven openjdk-11-jdk
wget https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.tar.gz
sudo tar -zxf apache-maven-3.6.3-bin.tar.gz -C /usr/local/
sudo ln -s /usr/local/apache-maven-3.6.3/bin/mvn /usr/local/bin/mvn
cd Documents/
git clone https://github.com/apache/zeppelin.git
cd zeppelin/
ls
export MAVEN_OPTS="-Xms1024M -Xmx4096M -XX:MaxMetaspaceSize=1024m -XX:-UseGCOverheadLimit -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=war"
./mvnw -B package -DskipTests -Pbuild-distr -Pspark-3.3 -Pinclude-hadoop -Phadoop3 -Pspark-scala-2.12 -Pweb-angular -Pweb-dist -pl '!groovy,!submarine,!flink,!cassandra,!jdbc,!bigquery,!alluxio,!mongodb,!neo4j' -am --no-transfer-progress
`

Armadik:
```
] Copying webapp resources [/home/micha/Documents/zeppelin/zeppelin-web/dist]
[INFO] deleting outdated resource WEB-INF/lib/hadoop-client-api-3.3.6.jar
[INFO] deleting outdated resource WEB-INF/lib/hadoop-client-runtime-3.3.6.jar
[INFO] Building war: /zeppelin/zeppelin-web/target/zeppelin-web-0.12.0-SNAPSHOT.war``

This one doesn't seem to work

The team helped me
zip -d /opt/zeppelin/zeppelin-web-0.12.0-SNAPSHOT.war WEB-INF/lib/*

pan3793:
@Armadik sorry, can not reproduce, both classic and new UI are good on my side.

Armadik:
I tried pure surka. It seems the problem was in my environment(

PR #4692, opened 2023-11-13
[ZEPPELIN-5980] Merge zeppelin-display into spark-interperter

pan3793:
### What is this PR for?

This PR aims to merge `zeppelin-display` into `spark-interperter`, as suggested in https://github.com/apache/zeppelin/pull/4652#issuecomment-1739360154

It also upgrades the scalatest to 3.2.15 (aligning with Spark 3.4.1) to support Scala 2.13

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5980

Pass GA.

### Questions:


pan3793:
cc @Reamer, this is ready for review

Reamer (approved):
I don't like the mixing of changes in Spark and the Cassandra interpreter. Do you see a way to split this up?
I'm not very good at Scala myself, but the changes in this pull request are manageable.
What IDE do you use where there is such good Scala and Java support?

pan3793:
> I don't like the mixing of changes in Spark and the Cassandra interpreter. Do you see a way to split this up?

Let me send a PR to upgrade the Cassandra part first.

> What IDE do you use where there is such good Scala and Java support?

I have used JetBrains IntelliJ IDEA Community Edition for a long time, I don't have much experience with other IDEs.

jongyoul:
BTW, with the same reason, we can merge zeppelin-zengine to zeppelin-server. :-)

pan3793:
@jongyoul thanks for the additional information. 

One of the motivations for merging this module is to reduce the dependence of Scala in Zeppelin core modules (I mean modules other than interpreter and plugins), after this patch,
the global `scala-2.11` and `scala-2.12` does not take effect and we can remove it. Now,

- Cassandra interpreter is hardcoded with scala 2.12
- Spark interpreter is controlled by `spark-scala-2.12` and `spark-scala-2.13`
- Flink interpreter is going to be hardcoded with scala 2.12
- Livy interpreter itself does not depend on Scala, but the test deps depends on Scala 2.11(because Livy only published the Scala 2.11 jars to Maven Central)

pan3793:
> BTW, with the same reason, we can merge zeppelin-zengine to zeppelin-server. :-)

Thanks for the information, I have not read through those codes, will take a look after the cleanup work of the interpreters is done.

Reamer:
> > BTW, with the same reason, we can merge zeppelin-zengine to zeppelin-server. :-)
> 
> Thanks for the information, I have not read through those codes, will take a look after the cleanup work of the interpreters is done.

I think it's actually pretty good that we have the two modules (zeppelin-zengine and zeppelin-server). Theoretically, it is possible to replace the application server (currently Jetty). But I don't know anyone who really does that. Nevertheless, it is important to me that we only use standardized configurations and do not rely on deep Jetty functions.

But I've already learned about some of the disadvantages. Maybe it makes sense to merge the two modules here. Even if this will certainly not be easy.

In principle, I am also thinking of a zeppelin-api module, where there are only interfaces so that you cannot access the class directly in "plugins".

pan3793:
actually, I think `zeppelin-interpreter` should be kind of a lightweight API module, but I found packaging `zeppelin-interpreter-shaded` is extremely slow, and the shading process includes bunches of third-party jars, looks unreasonable ...

Reamer:
zeppelin-interpreter contains, among other things, the code to execute a Zeppelin interpreter remotely.

The configuration of Zeppelin is also included here. In my opinion, it belongs in the zeppelin-zengine module. The interpreter should only have a simple hashmap, which is transferred when the remote interpreter is started.

PR #4693, opened 2023-11-17
[ZEPPELIN-5981] Check binded interpreter inside Paragraph.jobAbort()

elbakramer:
### What is this PR for?
Inside an open zeppelin notebook, after running a paragraph by clicking the play icon button, canceling the running paragraph by clicking the pause icon button (or the Ctrl+Option+C shortcut) does not interrupt the running process and cancel the paragraph.

I've checked that `CANCEL_PARAGRAPH` message gets well passed through the websocket connection but server was somehow completely ignoring the message.

After some more investigation, I've found that the reason for that was the `Paragraph`'s `jobAbort()` method, which does nothing due to the null interpreter member.

https://github.com/apache/zeppelin/blob/73e1aa37d403626d23138b3b346aec5ef2c12813/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L505

I've checked some other methods of `Paragraph` such as `completion()`, `execute()`, `jobRun()`, `recover()` and they were initializing interpreter before doing something by calling `getBindedInterpreter()`, so I've changed the `jobAbort()` method to do the same.

https://github.com/apache/zeppelin/blob/73e1aa37d403626d23138b3b346aec5ef2c12813/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L258

https://github.com/apache/zeppelin/blob/73e1aa37d403626d23138b3b346aec5ef2c12813/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L333

https://github.com/apache/zeppelin/blob/73e1aa37d403626d23138b3b346aec5ef2c12813/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L399

https://github.com/apache/zeppelin/blob/73e1aa37d403626d23138b3b346aec5ef2c12813/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java#L810

Other than that, I have some doubts on whether it's right thing to return true on failing what was requested.

But after all, I just sticked to the original behavior (returning true for every case) to prevent some unexpected side effects.

### What type of PR is it?
Bug Fix

### Todos
N/A

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5981

* Run a paragraph that can be canceled, like a python interpreter paragraph with `for i in range(10): time.sleep(1)`
* Click the cancel button and check if it gets canceled .

N/A

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/Paragraph.java
@@ -502,12 +502,19 @@ public class Paragraph extends JobWithProgressPoller<InterpreterResult> implemen
 
   @Override
   protected boolean jobAbort() {
-    try {
-      this.interpreter = getBindedInterpreter();
-    } catch (InterpreterNotFoundException e) {
-      LOGGER.debug("Unable to abort because there's no interpreter bind to it", e);
-      setInterpreterNotFound(e);
+    if (isAborted()) {
+      LOGGER.debug("Paragraph is already aborted.");
       return true;
+    }
+    if (isRunning() && this.interpreter == null) {
+      try {
+        LOGGER.debug("Trying to get the binded interpreter to cancel.");
+        this.interpreter = getBindedInterpreter();
+      } catch (InterpreterNotFoundException e) {
+        LOGGER.debug("Unable to abort because there's no interpreter bind to it", e);
+        setInterpreterNotFound(e);
+        return true;
+      }
     }
     if (interpreter == null) {
       LOGGER.debug("Unable to abort because interpreter is null");


jongyoul:
@elbakramer Thank you for the contribution. BTW, could you please check the CI? All tests for core-modules failed. I hope you investigate and fix them before starting reviews.

elbakramer:
@jongyoul I think the failures are due to CI actions taking too long, hopefully not by the code change. Could you please check if it's the CI issue first? Or is there any way to investigate the CI or just rerun by myself?
스크린샷 2023-11-21 오후 2 33 18

jongyoul:
@elbakramer I don't think we have any critical issues in CI now but let me restart it. FYI, I checked the log briefly and it looks like hanging for a long time in some tests.

jongyoul:
Hello, I re-ran the test but there are still some error with the same situation before. I checked the current CI and it didn't have a problem. Could you please check why the CI stops and hangs?

jongyoul:
Thank you for checking and fixing it. I think your analysis looks good. Let's see how the CI goes. I ran it again.

jongyoul:
Could you please rebase this PR to trigger the latest CI again? one of them still fails.

PR #4698, opened 2023-11-20
[ZEPPELIN-5985] Remove global scala-* profiles

pan3793:
### What is this PR for?

the global `scala-2.11` and `scala-2.12` does not take effect and we can remove it, now

- Cassandra interpreter is hardcoded with scala 2.12
- Spark interpreter is controlled by spark-scala-2.12 and spark-scala-2.13
- Flink interpreter is hardcoded with scala 2.12
- Livy interpreter itself does not depend on Scala, but the test deps depends on Scala 2.11(because Livy only published the Scala 2.11 jars to Maven Central)

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-5985

Review, and manually test.

### Questions:
* Does this needs documentation? Yes, updated


dev/create_release.sh
```diff
@@ -98,9 +97,9 @@ function make_binary_release() {
 git_clone
 make_source_package
 
-make_binary_release netinst "-Pweb-angular -Phadoop-2.6 -pl !hbase,!jdbc,!file,!flink,!cassandra,!elasticsearch,!bigquery,!alluxio,!livy,!groovy,!java,!neo4j,!submarine,!sparql,!mongodb -am"
```
pan3793, line 101: `hadoop-2.6` also does not take effect. currently, `hadoop-2.7` is the default one

jongyoul (approved):
LGTM. BTW, we'd better run the script after merging this PR.

Reamer (approved):

dev/create_release.sh
pan3793: an additional question which beyond this PR's scope.

`!flink` was listed here, but it does not work. what's the original intention here? do we want to include flink stuff into `netinst` artifacts?

PR #4707, opened 2024-02-07
[HOTFIX] Add warning for sh interpreter

jongyoul:
### What is this PR for?
Adding warning message when using sh interpreter

### What type of PR is it?
Documentation

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/security.md
+++ b/security.md
@@ -36,6 +36,12 @@ available to trusted users, and the server on which Zeppelin is
 installed does not contain any secrets or have privileges beyond
 those the users are trusted with.
 
+Especially, `sh` interpreter can access the local shell and execute
+arbitrary commands with the privileges of the user running Zeppelin 
+server. This is also intentional behaviors of Zeppelin, but it's too 
+dangerous to allow by default. This will be disabled from the release 
+of 0.11.1.
+
 ### Zeppelin on Docker
 
 An exception to the above is when the Zeppelin interpreter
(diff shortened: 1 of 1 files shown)


raboof (approved):
Good to make this more explicit!

Of course many other interpreters that have equivalent power: the language-based ones (Groovy, Java, Python) trivially so. Are you planning on disabling those by default as well?

My only concern with this wording is that by saying "`sh` interpreter can access the local shell and execute arbitrary commands", you might give people the incorrect impression that the other interpreters cannot provide this access. Perhaps you could try something like: 

"All interpreters should be assumed to be able to access the local shell and execute arbitrary commands with the privileges of the user running the Zeppelin server. As generic interpreters such as `sh`, `Groovy`, `Java` and `Python` make this especially trivial, we plan to disable the `sh` interpreter by default from version 0.11.1 onward.

Reamer (approved):

pan3793 (approved):

pan3793:
Maybe we should add a DISClAIMER like another notebook [polynote](https://polynote.org/latest/docs/installation/) does?

jongyoul:
@raboof Thank you for the detailed suggestions. I'll change it on the repository of zeppelin-site.

@pan3793 It looks good idea :-)

jongyoul:
BTW, I added it on https://github.com/apache/zeppelin-site/commit/da2424181cb7e907eb2f34afd166db59fcf0f543. I forgot to change the way to deploy the site. I'll also remove gh-pages soon because https://github.com/apache/zeppelin-site already has all the histories.

PR #4709, opened 2024-02-08
[ZEPPELIN-5990] Disable sensitive configuration for JDBC url

jongyoul:
### What is this PR for?
Disabling some sensitive configuration for JDBC URL when running queries.

### What type of PR is it?
Feature

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-5990

* Test should be passed
* Sensitive configuration shouldn't be accepted

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -153,6 +153,14 @@ public class JDBCInterpreter extends KerberosInterpreter {
           "KerberosConfigPath", "KerberosKeytabPath", "KerberosCredentialCachePath",
           "extraCredentials", "roles", "sessionProperties"));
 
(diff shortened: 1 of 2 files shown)


jongyoul:
ping @zjffdu @Reamer @huage1994

zjffdu (approved):

pjfanning:
This is a fix for CVE-2024-31864
I don't believe a CVE should have been issued for this and it has opened up a long list of requests for similar CVEs.
I see no problem with hardening the Zeppelin code, I just disagree with assigning CVEs for things like this.
If you read https://zeppelin.apache.org/security.html
* Zeppelin is all about letting you run code
* You are expected to run Zeppelin on a secure network or to enable Shiro auth if you need to open up access
* In short, Zeppelin is not expected to be run in an adversarial environment
* MySQL jars are not shipped with Zeppelin - we should just say that users who add JDBC driver jars to Zeppelin are doing so at their own risk

fyi @jongyoul @zjffdu @Reamer @raboof

PR #4719, opened 2024-02-19
[NO-ISSUE] Use reload4j instead of log4j

pjfanning:
### What is this PR for?

* reload4j does not prevent us moving to log4jv2 later
* reload4j fixes some security issues in log4j
* it doesn't feel great for an ASF project to still be using insecure logging frameworks like log4j v1

### What type of PR is it?
Bug Fix
Improvement
Feature
Documentation
Hot Fix
Refactoring
*Please leave your type of PR only*

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,7 @@
     <plugin.frontend.version>1.12.1</plugin.frontend.version>
 
     <!-- common library versions -->
-    <slf4j.version>1.7.36</slf4j.version>
+    <slf4j.version>1.7.35</slf4j.version>
     <reload4j.version>1.2.25</reload4j.version>
     <libthrift.version>0.13.0</libthrift.version>
     <flexmark.all.version>0.62.2</flexmark.all.version>
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,8 @@
     <plugin.frontend.version>1.12.1</plugin.frontend.version>
 
     <!-- common library versions -->
-    <slf4j.version>1.7.35</slf4j.version>
+    <slf4j.version>1.7.36</slf4j.version>
+    <reload4j.version>1.2.25</reload4j.version>
     <libthrift.version>0.13.0</libthrift.version>
     <flexmark.all.version>0.62.2</flexmark.all.version>
     <gson.version>2.8.9</gson.version>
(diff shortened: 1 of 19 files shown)


pjfanning:
@jongyoul is this something that could be merged?

jongyoul (dismissed):

pom.xml
```diff
@@ -114,8 +114,7 @@
     <plugin.frontend.version>1.12.1</plugin.frontend.version>
 
     <!-- common library versions -->
-    <slf4j.version>1.7.30</slf4j.version>
-    <log4j.version>1.2.17</log4j.version>
+    <slf4j.version>1.7.35</slf4j.version>
```
pan3793, line 117: why not [1.7.36](https://mvnrepository.com/artifact/org.slf4j/slf4j-reload4j/1.7.36)?

pjfanning: causes a maven convergence issue - due to other dependencies of zeppelin using 1.7.35

I can try further changes but it could take a few days.

The key thing is to get agreement that keeping log4jv1 is not a good thing and that reload4j is a reasonable short term change to avoid using a completely insecure logging framework.

pan3793: thanks for the elaboration, that makes sense.

pom.xml
```diff
@@ -240,12 +239,6 @@
         <version>${slf4j.version}</version>
       </dependency>
 
-      <dependency>
-        <groupId>log4j</groupId>
-        <artifactId>log4j</artifactId>
```
pan3793, line 245: removing the definition from `dependencyManagement` does not mean excluding the jar from the final binary tarball, there is still a chance that log4j is collected as the transitive dependency into the final tarball, additional manual checks are required.

zeppelin-server/pom.xml
```diff
@@ -88,6 +88,14 @@
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-log4j12</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-reload4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
```
Reamer, line 97: Double exclusion see line 88

pom.xml
pjfanning: I'll have a better look over the next day or 2 to see what transitive dependencies could leaking log4j dependencies into zeppelin. There already appears to be issues (pre-existing issues unrelated to this change) where log4j and log4j-1.2-api are conflicting. log4j-1.2-api is a log4jv2 jar that pretends to be log4jv1 and you shouldn't have this on the classpath at the same time as you have the legacy log4j jar.

zeppelin-server/pom.xml
pjfanning: thanks - pushed a fix

pjfanning:
I built zeppelin-distribution/target/zeppelin-0.11.0-SNAPSHOT.tar.gz locally with these changes and log4j jar was not in the dist.

pan3793 (approved):

pjfanning:
@jongyoul @pan3793 @Reamer could this PR be merged? The ASF is suffering damage to our reputation by having project's who won't use secure libraries - in particular old versions of log4j attract make us look really bad.

jongyoul (approved):

PR #4722, opened 2024-02-20
[NO-ISSUE] Upgrade org.json jar

pjfanning:
### What is this PR for?

See ZEPPELIN-5996

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-5996](https://issues.apache.org/jira/browse/ZEPPELIN-5996)

### Questions:

### DIFF
--- a/flink/flink-scala-2.12/pom.xml
+++ b/flink/flink-scala-2.12/pom.xml
@@ -254,6 +254,17 @@
       <groupId>com.mashape.unirest</groupId>
       <artifactId>unirest-java</artifactId>
       <version>1.4.9</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.json</groupId>
+          <artifactId>json</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
     </dependency>
 
     <dependency>
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -84,6 +84,11 @@
     </dependency>
 
     <dependency>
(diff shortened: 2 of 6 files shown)


jongyoul:
Don't we have to update the LICENSE file?

flink/flink-scala-2.12/pom.xml
```diff
@@ -254,6 +254,17 @@
       <groupId>com.mashape.unirest</groupId>
       <artifactId>unirest-java</artifactId>
```
Reamer, line 255: I think it makes more sense to raise this dependency and not play around with the transitive dependency.
From version 3.0 this library uses  `com.google.code.gson » gson` and no longer `org.json » json`.
We also use GSON in many other places in Zeppelin.

Reamer (commented):
I have also checked the other dependencies. The JSON dependency has been dropped and has mostly been replaced by GSON.

flink/flink-scala-2.12/pom.xml
pjfanning: I am not a Zeppelin committer. I never use Zeppellin. I'm just trying to help out but have very limited time to be experimenting.

pjfanning:
> Don't we have to update the LICENSE file?

I've just updated [zeppelin-distribution/src/bin_license/LICENSE](https://github.com/apache/zeppelin/pull/4722/files#diff-73d534459c803c1cd56450d7b0386aeff4b1f0ecffb3a80a0bc00c0d053a57c6)

pjfanning:
@Reamer @jongyoul could this PR be merged? Zeppelin is breaking ASF rules by using a Category X dependency.

jongyoul (approved):

PR #4730, opened 2024-03-13
[ZEPPELIN-6002] Fix completer NPE

kevinjmh:
### What is this PR for?
NPE occur when adding the driver specific SQL completions. Some driver(hive-jdbc-uber-2.6.5.0-292.jar in my case) does not implement method `DatabaseMetaData.getDriverName()` well and cause exception , then fail to initialize `keywordCompleter`.

Since it is optional, we should avoid such driver interfere  the default one, and log it for the profession

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6002

image

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
@@ -207,8 +207,7 @@ public class SqlCompleter {
     }
   }
 
-  public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throws IOException,
-          SQLException {
+  public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throws IOException {
     // Add the default SQL completions
     String keywords =
             new BufferedReader(new InputStreamReader(
@@ -217,11 +216,11 @@ public class SqlCompleter {
     Set<String> completions = new TreeSet<>();
 
     if (null != meta) {
-      // Add the driver specific SQL completions
-      String driverSpecificKeywords =
-              "/" + meta.getDriverName().replace(" ", "-").toLowerCase() + "-sql.keywords";
-      logger.info("JDBC DriverName:" + driverSpecificKeywords);
       try {
+        // Add the driver specific SQL completions
+        String driverSpecificKeywords =
+                "/" + meta.getDriverName().replace(" ", "-").toLowerCase() + "-sql.keywords";
+        logger.info("JDBC DriverName: {}", driverSpecificKeywords);
         if (SqlCompleter.class.getResource(driverSpecificKeywords) != null) {
           String driverKeywords =
                   new BufferedReader(new InputStreamReader(
@@ -229,35 +228,35 @@ public class SqlCompleter {
                           .readLine();
           keywords += "," + driverKeywords.toUpperCase();
         }
-      } catch (Exception e) {
+      } catch (IOException | SQLException e) {
         logger.debug("fail to get driver specific SQL completions for " +
-                driverSpecificKeywords + " : " + e, e);
+            meta.getClass().getName() + " : " + e, e);
       }
 
       // Add the keywords from the current JDBC connection
       try {
         keywords += "," + meta.getSQLKeywords();
-      } catch (Exception e) {
+      } catch (SQLException e) {
         logger.debug("fail to get SQL key words from database metadata: " + e, e);
       }
       try {
         keywords += "," + meta.getStringFunctions();
-      } catch (Exception e) {
+      } catch (SQLException e) {
         logger.debug("fail to get string function names from database metadata: " + e, e);
       }
       try {
         keywords += "," + meta.getNumericFunctions();
-      } catch (Exception e) {
+      } catch (SQLException e) {
         logger.debug("fail to get numeric function names from database metadata: " + e, e);
       }
       try {
         keywords += "," + meta.getSystemFunctions();
-      } catch (Exception e) {
+      } catch (SQLException e) {
         logger.debug("fail to get system function names from database metadata: " + e, e);
       }
       try {
         keywords += "," + meta.getTimeDateFunctions();
-      } catch (Exception e) {
+      } catch (SQLException e) {
         logger.debug("fail to get time date function names from database metadata: " + e, e);
       }
 


jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
```diff
@@ -217,11 +217,11 @@ public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throw
     Set<String> completions = new TreeSet<>();
 
     if (null != meta) {
```
Reamer, line 219:
```suggestion
    if (meta != null &&  meta.getDriverName() != null) {
```

jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
```diff
@@ -217,11 +217,11 @@ public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throw
...
-      // Add the driver specific SQL completions
-      String driverSpecificKeywords =
-              "/" + meta.getDriverName().replace(" ", "-").toLowerCase() + "-sql.keywords";
-      logger.info("JDBC DriverName:" + driverSpecificKeywords);
       try {
+        // Add the driver specific SQL completions
+        String driverSpecificKeywords =
+                "/" + meta.getDriverName().replace(" ", "-").toLowerCase() + "-sql.keywords";
+        logger.info("JDBC DriverName:" + driverSpecificKeywords);
```
Reamer, line 224:
```suggestion
        logger.info("JDBC DriverName: {}", driverSpecificKeywords);
```

Reamer (changes requested):
I would just check for null in the if statement. If simple string operations in the try throw exceptions, it doesn't look good. The RuntimeExection NullPointerException should not occur.
Perhaps it is also possible to convert the general exception into an IOException.

jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
Reamer: I see, therefore the additional null check `meta.getDriverName() != null`.
So the method should take the default keywords
```
String keywords =
            new BufferedReader(new InputStreamReader(
                    SqlCompleter.class.getResourceAsStream("/ansi.sql.keywords"))).readLine();
```
and return this set.

kevinjmh: However, exception is thrown once calling `getDriverName` , then `keywords` mentioned above cannot return and completer cannot work.  

Methods in the driver classes are implemented by other developers. We cannot ensure their actions, so put them in try-catch is more code-safe.

Reamer: Are you sure that the call `getDriverName()` throws an NPE? Looking at the code, I think that the call `getDriverName().replace('...')` throws the NPE because `getDriverName()` returns null.

I have looked at the implementation of the Hive driver. I hope I have found the right place. https://github.com/apache/hive/blob/489f6a20a5eedc1836854fd40255eb78aa967d99/jdbc/src/java/org/apache/hive/jdbc/HiveDriver.java#L264-L303

I don't see any possibility that an NPE is thrown. But an SQL exception can be thrown. To do this, `loadManifestAttributes();` would have to throw an IOException -> then the `SQL-Exception` is thrown.

Reamer: I didn't realize that `getDriverName()` can throw an `SQLException`, so I'm closing this suggestion.

jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
```diff
@@ -230,8 +230,7 @@ public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throw
           keywords += "," + driverKeywords.toUpperCase();
         }
       } catch (Exception e) {
```
Reamer, line 232: We should catch the explicit exceptions. Also, `SQL-Exception` should be removed as a throw declaration.
```suggestion
      } catch (IOException | SQLException e) {
```

Reamer (changes requested):
The `SQLException` can no longer be thrown by the `getSqlKeywordsCompletions` method, so the method signature must still be adjusted.

jdbc/src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java
```diff
@@ -217,21 +217,20 @@ public static Set<String> getSqlKeywordsCompletions(DatabaseMetaData meta) throw
...
                   new BufferedReader(new InputStreamReader(
                           SqlCompleter.class.getResourceAsStream(driverSpecificKeywords)))
                           .readLine();
           keywords += "," + driverKeywords.toUpperCase();
         }
-      } catch (Exception e) {
-        logger.debug("fail to get driver specific SQL completions for " +
-                driverSpecificKeywords + " : " + e, e);
+      } catch (IOException | SQLException e) {
+        logger.debug("fail to get driver specific SQL completions for " + meta.getClass().getName() + " : " + e, e);
```
Reamer, line 233: CI Error:
```
 [INFO] There is 1 error reported by Checkstyle 6.11.2 with zeppelin/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/zeppelin/jdbc/SqlCompleter.java:[233] (sizes) LineLength: Line is longer than 100 characters (found 116).
```

Reamer:
```suggestion
        logger.debug("fail to get driver specific SQL completions for " +
            meta.getClass().getName() + " : " + e, e);
```

Reamer (changes requested):
One checkstyle error

kevinjmh:
@Reamer thanks for your help!

Reamer:
> The `SQLException` can no longer be thrown by the `getSqlKeywordsCompletions` method, so the method signature must be adjusted.

Reamer (approved):
Thank you for your contribution, the changes look good. I hope my quality requirements have not scared you off for future contributions.

PR #4732, opened 2024-03-14
[ZEPPELIN-6003] Log source info of SQL in JDBCInterpreter

kevinjmh:
### What is this PR for?
Currently we can monitor running SQL in JDBCInterpreter  but unable to know who/which note fire the SQL, especially a couple of users using zeppelin at the same time. 

Bond the SQL with additional info,  we can inform the user with user/noteid/paragragh info  if needed

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6003

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -763,6 +763,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
     Statement statement;
     ResultSet resultSet = null;
     String paragraphId = context.getParagraphId();
+    String noteId = context.getNoteId();
     String user = getUser(context);
 
     try {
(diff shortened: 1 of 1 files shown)


jongyoul (approved):

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -799,7 +800,8 @@ private InterpreterResult executeSql(String sql,
           // so we need to trim it first in this case.
           sqlToExecute = sqlToExecute.trim();
         }
-        LOGGER.info("Execute sql: " + sqlToExecute);
+        LOGGER.info(String.format("[%s|%s|%s] Execute sql: %s",
+                user, noteId, paragraphId, sqlToExecute));
```
Reamer, line 804:
```suggestion
        LOGGER.info("[{}|{}|{}] Execute sql: {}", user, noteId, paragraphId, sqlToExecute);
```
See: https://rules.sonarsource.com/java/RSPEC-2629/

Reamer (changes requested):
String.format is not necessary for our logging facade

Reamer (approved):
LGTM

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
jongyoul: Yes, I also think it's better.

jongyoul (approved):

PR #4748, opened 2024-04-01
[ZEPPELIN-6008] Fix parameter usage of bokeh in test case

pan3793:
### What is this PR for?

https://github.com/bokeh/bokeh/wiki/Migration-Guides#300

`legend` keyword argument to glyph methods has been removed. Use more specific `legend_label`, `legend_group`, and `legend_field` instead.

### What type of PR is it?

Hot Fix

### Todos
* [ ] - Task

### What is the Jira issue?

ZEPPELIN-6008

GHA

### Questions:


pan3793:
cc @jongyoul, this may help

pan3793:
also cc @zjffdu @Reamer @huage1994

Reamer (approved):
I also saw the error in the CI. Without looking into the code, I thought it was being triggered incorrectly by another Python library.
Thank you for analysing the problem.

PR #4749, opened 2024-04-03
[MINOR] Fix deployment issues

jongyoul:
### What is this PR for?
Fixing some issues that happened when releasing.

### What type of PR is it?
Hot Fix

### Todos
* [x] - Remove some plugins from the shell interpreter instead of removing them from the distribution as some interpreters depend on it
* [x] - Upgrade python version for docker image to 3.9

### What is the Jira issue?
N/A

Distribution and building docker image should work fine.

### Questions:

### DIFF
--- a/shell/pom.xml
+++ b/shell/pom.xml
@@ -125,9 +125,9 @@
       </plugin>
       <plugin>
         <artifactId>maven-resources-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <artifactId>maven-shade-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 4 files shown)


shell/pom.xml
```diff
@@ -125,9 +125,9 @@
       </plugin>
       <plugin>
         <artifactId>maven-resources-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <artifactId>maven-shade-plugin</artifactId>
+        <configuration>
```
pan3793, line 128: what's the effect of this change?

pan3793: It seems the whole `sh` dir disappears in the `interpreter` folder, not only "configurations", is it expected?

jongyoul: Yes, correct. Shell interpreter was and will be removed from the official binary because it lead a lot of security issues.

pan3793: LGTM if it's the expected behavior. Test may be need to adjusted to adapt this change.

jongyoul:
There are a lot of test failures. Let me fix it one by one soon

PR #4765, opened 2024-05-30
[ZEPPELIN-6022] Skip decryption of credentials.json when file is empty

zeotuan:
### What is this PR for?
This PR skip credentials.json decryption if the file is empty. 
This allow using of type=bind on conf/credentials.json when running docker image especially for initial startup where credentials.json does not exist

### What type of PR is it?
Improvement

*Please leave your type of PR only*

### Todos
* [ ] - Task

### What is the Jira issue?
(https://issues.apache.org/jira/browse/ZEPPELIN-6022)

* CI with automated test

### Questions:
* Does this needs documentation? Maybe

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/user/Credentials.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/user/Credentials.java
@@ -139,7 +139,7 @@ public class Credentials {
   private void loadFromFile() throws IOException {
     try {
       String json = storage.loadCredentials();
-      if (json != null && encryptor != null && !json.isEmpty()) {
+      if (encryptor != null && StringUtils.isNotBlank(json)) {
         json = encryptor.decrypt(json);
       }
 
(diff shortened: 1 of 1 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/user/Credentials.java
```diff
@@ -139,7 +139,7 @@ private void loadCredentials() throws IOException {
   private void loadFromFile() throws IOException {
     try {
       String json = storage.loadCredentials();
-      if (json != null && encryptor != null) {
+      if (json != null && encryptor != null && !json.isEmpty()) {
```
Reamer, line 142:
```suggestion
      if (encryptor != null && StringUtils.isNotBlank(json)) {
```

Reamer (commented):
Thank you for your contribution. I think the change should not be a problem.

The result is almost the same, the code is just a bit more elegant. A blank string does not need to be encrypted either.

zeppelin-zengine/src/main/java/org/apache/zeppelin/user/Credentials.java
zeotuan: Thank you. Updated to use StringUtils

Reamer (approved):
LGTM, let's wait for CI

PR #4775, opened 2024-07-22
[ZEPPELIN-6031] Clean up test directories even after unexpected exit of `AbstractInterpreterTest`

JooHyukKim:
### What is this PR for?

Resolves issue [ZEPPELIN-6031](https://issues.apache.org/jira/browse/ZEPPELIN-6031).

Currently, if someone runs tests locally and stop tests there will be left over config files, notebooks, etc.
It seems to happen more frequently with interpreter tests extending `AbstractInterpreterTest`.

This PR resolves by adding `forceDeleteOnExit()` hook so the files are deleted after JVM exit (even after test interruption)

### What type of PR is it?
Improvement

### Todos
* [x] Add `forceDeleteOnExit()` hook so the files are deleted after JVM exit (even after test interruption)

### What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-6031

* Just Github Actions will do.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java
@@ -112,6 +112,7 @@ public abstract class AbstractInterpreterTest {
     notebook = new Notebook(conf, authorizationService, notebookRepo, noteManager, interpreterFactory, interpreterSettingManager, credentials);
     interpreterSettingManager.setNotebook(notebook);
 
+    // cleanup the test directories on exit
     FileUtils.forceDeleteOnExit(interpreterDir);
     FileUtils.forceDeleteOnExit(confDir);
     FileUtils.forceDeleteOnExit(notebookDir);
--- a/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java
+++ b/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java
@@ -111,6 +111,10 @@ public abstract class AbstractInterpreterTest {
     Credentials credentials = new Credentials(conf, storage);
     notebook = new Notebook(conf, authorizationService, notebookRepo, noteManager, interpreterFactory, interpreterSettingManager, credentials);
     interpreterSettingManager.setNotebook(notebook);
+
+    FileUtils.forceDeleteOnExit(interpreterDir);
+    FileUtils.forceDeleteOnExit(confDir);
+    FileUtils.forceDeleteOnExit(notebookDir);
   }
 
   @AfterEach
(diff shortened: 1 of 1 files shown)


jongyoul:
Thank you for the contribution. Let's see how CI works.

JooHyukKim:
Below are three Github Actions failing on CI, none of them seems to be related to change here.
Could it be that all the tests are flakey? If so, can we try re-running the CI?

### spark-test 3.8

```log
Error:  Errors: 
Error:    IPySparkInterpreterTest>IPythonInterpreterTest.setUp:86->startInterpreter:98 » Interpreter org.apache.zeppelin.interpreter.InterpreterException: java.io.IOException: Fail to setup JVMGateway
```

### **flink-test-and-flink-integration-test (3.8, 115)** fails to even set 

```log
    × python setup.py egg_info did not run successfully.
    │ exit code: 1
    ╰─> [1 lines of output]
        ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
        [end of output]
```

### interpreter-test-jupyter-python-rlang (3.8, 8)

```log
Error:    IPythonKernelTest.testUpdateOutput:290 » Interpreter Fail to open JupyterKernelInterpreter:
```

Reamer:
I understand your problem. In general, we should move away from the following line and run the tests within the project tree.
https://github.com/apache/zeppelin/blob/289e954db5a2cd4577a8ecd0437ef2f7353cbf55/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java#L72

However, this requires bigger modifications.

Your lines modify the shutdown hook, so it would be better to place your lines after creating the folders.
https://github.com/apache/zeppelin/blob/289e954db5a2cd4577a8ecd0437ef2f7353cbf55/zeppelin-zengine/src/test/java/org/apache/zeppelin/interpreter/AbstractInterpreterTest.java#L79-L81

JooHyukKim:
> Your lines modify the shutdown hook, so it would be better to place your lines after creating the folders.

You mean "right after" creating directories, @Reamer ? If so, what would be the difference? The shutdown hooks are already placed after creating-directories.

Reamer:
> If so, what would be the difference?

This only serves to keep things together that belong together.

JooHyukKim:
I see. Not that I don't agree with you, I was wondering if there was any other reason behind the suggestion :).

Thank you for the review and I will play the changes as soon as I get back home.

Reamer (approved):
LGTM

Reamer:
I think that the CI problem will be solved with the following fix.
https://github.com/conda-forge/backports-feedstock/pull/7

I always come across the following error.
```
  File "/home/runner/miniconda3/envs/python_3_with_R/lib/python3.8/site-packages/ipykernel/eventloops.py", line 12, in <module>
    from pkg_resources import parse_version as V
  File "/home/runner/miniconda3/envs/python_3_with_R/lib/python3.8/site-packages/pkg_resources/__init__.py", line 95, in <module>
    from jaraco.text import (
  File "/home/runner/miniconda3/envs/python_3_with_R/lib/python3.8/site-packages/setuptools/_vendor/jaraco/text/__init__.py", line 12, in <module>
    from jaraco.context import ExceptionTrap
  File "/home/runner/miniconda3/envs/python_3_with_R/lib/python3.8/site-packages/setuptools/_vendor/jaraco/context.py", line 17, in <module>
    from backports import tarfile
ImportError: cannot import name 'tarfile' from 'backports' (/home/runner/miniconda3/envs/python_3_with_R/lib/python3.8/site-packages/backports/__init__.py)
```

PR #4779, opened 2024-07-30
[ZEPPELIN-6039] Exclude dependencies in plugin docker launcher

proceane:
### What is this PR for?
Exclude dependencies that use a different version than the global version.

### What type of PR is it?
Improvement

### Todos
* [x] - Exclude jsoup:1.10.3, guava:25.0-jre in jinjava dependency
* [x] - Exclude commons-compress:1.18 in docker-client

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6039

CI

### Questions:


jongyoul:
LGTM. Let's see how CI goes

jongyoul:
By the way, the failed tests are not common. Could you please check the test and it works well in your local?

PR #4780, opened 2024-08-02
[ZEPPELIN-6040] Run mode "docker" not working properly

tbonelee:
### What is this PR for?

Several issues have been identified with the "docker" run mode that are causing it to function improperly. Below are the specific problems encountered:
 #### Base Image Version Issue in `/scripts/docker/interpreter/Dockerfile`:

- The base image version is outdated, currently at `0.8.0`, which is significantly behind the latest version.
- The URL for downloading Spark is not working, causing failures in the build process.
 
#### Host Environment Variables Overriding Container Environment Variables:

-  The `JAVA_HOME` variable from the host is unnecessarily being passed to the container, potentially causing conflicts with the intended Java environment inside the container.
- The `PATH` from the host is being propagated into the container, which may lead to unexpected behavior due to differing paths and binaries.
- The `ZEPPELIN_HOME` variable from the host is also being passed, which can interfere with the Zeppelin environment within the container.

 #### Miscellaneous Issues:

- Additional issues have been identified testing Spark interpreter, I will check for it in another PR

### What type of PR is it?
Bug Fix

### Todos

* [x] - Fix dockerfile
* [ ] - Check for major interpreters
  * [x] md
  * [x] python
  * [x] shell
  * [x] R
  * [ ] spark

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/6040

1. Build Zeppelin using the following command:
    
    ```bash
    $ ./mvnw clean package -DskipTests -Pinclude-hadoop
    ```
    
2. Create the `conf/zeppelin-site.xml` file with the following content:
    
    ```xml
    <configuration>
      <property>
        <name>zeppelin.run.mode</name>
        <value>docker</value>
        <description>Zeppelin run mode</description>
      </property>
      <property>
        <name>zeppelin.docker.container.image</name>
        <value>chanho0325/interpreter:0.11.2</value>
        <description>'Docker image I've built with scripts/docker/interpreter/Dockerfile'</description>
      </property>
    </configuration>
    ```
    
4. Start Zeppelin using the command below:
    
    ```bash
    $ ./bin/zeppelin-daemon.sh start
    ```

### Questions:

### DIFF
--- a/scripts/docker/interpreter/Dockerfile
+++ b/scripts/docker/interpreter/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 FROM apache/zeppelin:0.11.2-SNAPSHOT
-MAINTAINER Apache Software Foundation <dev@zeppelin.apache.org>
+MAINTAINER Apache Zeppelin Community <dev@zeppelin.apache.org>
 
 ENV SPARK_VERSION=3.5.1
 ENV HADOOP_VERSION=3
--- a/scripts/docker/interpreter/Dockerfile
+++ b/scripts/docker/interpreter/Dockerfile
@@ -16,8 +16,8 @@
 FROM apache/zeppelin:0.11.2-SNAPSHOT
 MAINTAINER Apache Software Foundation <dev@zeppelin.apache.org>
 
-ENV SPARK_VERSION=2.3.3
-ENV HADOOP_VERSION=2.7
+ENV SPARK_VERSION=3.5.1
+ENV HADOOP_VERSION=3
 
 USER root
 
--- a/scripts/docker/interpreter/Dockerfile
+++ b/scripts/docker/interpreter/Dockerfile
@@ -13,11 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM apache/zeppelin:0.8.0
+FROM apache/zeppelin:0.11.2
 MAINTAINER Apache Software Foundation <dev@zeppelin.apache.org>
 
 ENV SPARK_VERSION=2.3.3
 ENV HADOOP_VERSION=2.7
+
+USER root
 
 # support Kerberos certification
 RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -yq krb5-user libpam-krb5 && apt-get clean
@@ -25,10 +27,10 @@ RUN export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -
 RUN apt-get update && apt-get install -y curl unzip wget grep sed vim tzdata && apt-get clean
 
 # auto upload zeppelin interpreter lib
-RUN rm -rf /zeppelin
+RUN rm -rf /opt/zeppelin
 
 RUN rm -rf /spark
-RUN wget https://www-us.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
+RUN wget https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
 RUN tar zxvf spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
-RUN mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} spark
+RUN mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} /spark
 RUN rm spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
--- a/scripts/docker/interpreter/Dockerfile
+++ b/scripts/docker/interpreter/Dockerfile
@@ -32,5 +32,5 @@ RUN rm -rf /opt/zeppelin
 RUN rm -rf /spark
 RUN wget https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
 RUN tar zxvf spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz
-RUN mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} /spark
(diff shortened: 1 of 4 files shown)


tbonelee:
I am currently checking the functionality of the Spark interpreter

scripts/docker/interpreter/Dockerfile
```diff
@@ -13,22 +13,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM apache/zeppelin:0.8.0
+FROM apache/zeppelin:0.11.2
```
jongyoul, line 16:
```suggestion
FROM apache/zeppelin:0.11.2-SNAPSHOT
```

zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -190,7 +192,7 @@ public void start(String userName) throws IOException {
     // check if the interpreter process exit script
     // if interpreter process exit, then container need exit
     StringBuilder sbStartCmd = new StringBuilder();
-    sbStartCmd.append("sleep 10; ");
+    sbStartCmd.append("sleep 20; ");
```
jongyoul, line 195: By the way, I always wonder if we could have a better way to wait until the process is ready, not just praying :-)

tbonelee: For Docker mode, there might be a way to actively execute the process instead of just waiting. 
However, in this case, we may have chosen this structure to upload interpreter files by using commands like `rm`, `mkdir`, while the container is running, and then execute the `interpreter.sh`. 

I will consider whether we can change this structure.

pan3793:
the shipped spark version is too old, maybe you can bump it to 3.5

tbonelee:
ping @jongyoul @pan3793

jongyoul:
@tbonelee To make sure, do we assume the zeppelin home for containers is /opt/zeppelin by default?

tbonelee:
> @tbonelee To make sure, do we assume the zeppelin home for containers is /opt/zeppelin by default?

Yes, I have matched the zeppelin home in the `scripts/docker/interpreter/Dockerfile` (which is based on `scripts/docker/zeppelin/bin/Dockerfile`, so it refers to the zeppelin home in that Dockerfile) with the default `zeppelin.docker.container.home` configuration value.

In practice, when the run mode is 'docker', the Zeppelin server uploads the interpreter files to the container's zeppelin home path at runtime within the interpreter Docker container, so it might not be strictly necessary to match them. However, to prevent potential conflicts, I determined it would be better to align it with the zeppelin home path of the base image(of dockerfile `scripts/docker/zeppelin/bin/Dockerfile`).

pan3793:
Actually, Spark hardcoding `SPARK_HOME` to `/opt/spark` and `HADOOP_CONF_DIR` to `/opt/hadoop/conf` in docker image and Spark on K8s cases, so far it works well.

I don't see any potential issues on hardcoding the `ZEPPELIN_HOME` inside the docker image, since docker already provides good isolation, to make life easier, I think we can do similar things too.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -1106,6 +1106,7 @@ public enum ConfVars {
 
     // Used by K8s and Docker plugin
     ZEPPELIN_DOCKER_CONTAINER_IMAGE("zeppelin.docker.container.image", "apache/zeppelin:" + Util.getVersion()),
+    ZEPPELIN_DOCKER_CONTAINER_HOME("zeppelin.docker.container.home", "/opt/zeppelin"),
 
     ZEPPELIN_DOCKER_CONTAINER_SPARK_HOME("zeppelin.docker.container.spark.home", "/spark"),
```
pan3793, line 1111: I also suggest using `/opt/spark` as SPARK_HOME, which matches users who experienced Spark on K8s

scripts/docker/interpreter/Dockerfile
```diff
@@ -13,22 +13,24 @@
...
 # limitations under the License.
 
-FROM apache/zeppelin:0.8.0
+FROM apache/zeppelin:0.11.2-SNAPSHOT
 MAINTAINER Apache Software Foundation <dev@zeppelin.apache.org>
 
-ENV SPARK_VERSION=2.3.3
-ENV HADOOP_VERSION=2.7
+ENV SPARK_VERSION=3.5.1
+ENV HADOOP_VERSION=3
```
pan3793, line 20: is this Dockerfile supposed to be used by user? if so, we should use ARG instead of ENV.

additionally, it's better to change it to SPARK_NAME or something else because the official spark binary artifacts have the below suffix 
```
-hadoop2 (removed since 3.4.0)
-hadoop3
-without-hadoop
```
users are also allowed to `dev/make-distribution.sh --name custom` to create spark bin tarball with `-custom` suffix.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
tbonelee: Then I'll standardize both the Spark installation directory in `scripts/docker/interpreter/Dockerfile` and `ZEPPELIN_DOCKER_CONTAINER_SPARK_HOME` to `/opt/spark`.

scripts/docker/interpreter/Dockerfile
tbonelee: Now that I think about it, I agree with your comment. It does seem more appropriate for users to control the Spark bin name.
I'd reflect your suggestion in a new commit.

tbonelee:
@pan3793
Thank you for the detailed review.
I've reflected your review and also made some adjustments to the "How should this be tested?" section.
Does this align with the direction you requested?

jongyoul:
@pan3793 kindly ping :-)

pan3793 (approved):

tbonelee:
@jongyoul Just rebased and pushed!

jongyoul (approved):
LGTM. Thank you for your first contribution to Apache Zeppelin!

PR #4781, opened 2024-08-03
[ZEPPELIN-6034] Write a Dockerfile for shell interpreter

seung-00:
### What is this PR for?
This PR adds a Dockerfile to build a shell interpreter. 
Although the interpreters are currently running as separate JVM processes, the existing structure confines them to a single machine, resulting in low availability. As a first step to address this, I've created a docker image for the shell interpreter using independent computing resources.

Since the current system wasn't designed with modular operation in mind, this PR involves some complexities in terms of configuration and testing. I'm planning to dockerize the other interpreters too, and I’ll try to iron out these kinks as I go.

### What type of PR is it?
Feature

### What is the Jira issue?
* [ZEPPELIN-6034](https://issues.apache.org/jira/browse/ZEPPELIN-6034)

* Build and run the docker container
* Verify that the shell interpreter container communicates with the zeppelin server

### Questions:
* Does the license files need to update? Yes

### DIFF
--- a/shell/Dockerfile
+++ b/shell/Dockerfile
@@ -23,12 +23,16 @@ WORKDIR /zeppelin
 
 RUN chmod +x ./mvnw
 
-RUN ./mvnw install -am -pl zeppelin-interpreter,shell -DskipTests
+RUN ./mvnw package -am -pl zeppelin-interpreter-shaded,zeppelin-interpreter,shell -DskipTests
 
 
 FROM openjdk:11
 
-COPY --from=builder /zeppelin /zeppelin
+COPY --from=builder /zeppelin/bin /zeppelin/bin/
+COPY --from=builder /zeppelin/conf /zeppelin/conf
+
+COPY --from=builder /zeppelin/interpreter/sh /zeppelin/interpreter/sh
+COPY --from=builder /zeppelin/zeppelin-interpreter-shaded/target /zeppelin/zeppelin-interpreter-shaded/target
 
 WORKDIR /zeppelin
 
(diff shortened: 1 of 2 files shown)


shell/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+WORKDIR /zeppelin
+
+ENV SHELL_INTERPRETER_PORT=8081
+
+RUN chmod +x ./mvnw && chmod +x ./bin/interpreter.sh
+
+RUN ./mvnw install -am -pl zeppelin-interpreter,shell -DskipTests
+
+CMD ./bin/interpreter.sh -d interpreter/sh -c host.docker.internal -p ${INTERPRETER_EVENT_SERVER_PORT} -r ${SHELL_INTERPRETER_PORT}:${SHELL_INTERPRETER_PORT} -i sh-shared_process -l ./local-repo -g sh
```
jongyoul, line 30: Could you please divide the builder and runner for this docker container? It reduces the size of the docker container as well.

seung-00: @jongyoul 

Sure. Thanks for the suggestion.
[9ea65b9](https://github.com/apache/zeppelin/pull/4781/commits/9ea65b990097572b870fdeebc6b13eb30be31690)

jongyoul (dismissed):
LGTM except for we need to verify if JRE should be fine to run interpreter.  I also wonder local-repo issue but let's talk about it with separate PR.

seung-00:
> LGTM except for we need to verify if JRE should be fine to run interpreter. I also wonder local-repo issue but let's talk about it with separate PR.

@jongyoul 
Thanks for the comments. I pushed a new commit so that only the necessary files go into the runner, which reset the PR approval. The image size has been reduced to about one-third.

jongyoul (approved):

PR #4784, opened 2024-08-07
[ZEPPELIN-6026] Zeppelin Websocket code 1006

Armadik:
### What is this PR for?

When testing version 0.11.1 I found an unpleasant bug. The system is deployed in the kubernetes environment in conjunction with ha-proxy. Some time after starting work, error 1006 occurs - after which the socket is not created again - only refreshing the page helps. There is no such problem in version 0.10.1. 

### What type of PR is it?
Bug Fix

### Todos
[ZEPPELIN-6026]

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6026

I tested the problem in the k8s environment. With debug mode enabled in the browser. As soon as an error with code 1006 occurs, a new Websocket should be created without refreshing the page.

### Questions:
no
no
no


Reamer:
I think your IDE is making some automatic changes here. The IDE changes need to be avoided as these can differ between developers.
Only make small changes.

zeppelin-web/package.json
```diff
@@ -40,7 +40,7 @@
     "angular-sanitize": "1.5.7",
     "angular-touch": "1.5.7",
     "angular-ui-ace": "0.2.3",
-    "angular-websocket": "1.1.0",
+    "angular-websocket": "^1.0.13",
```
jongyoul, line 43: @Armadik Hello, Could you please leave come comments that we should keep the version with some issues? I recommend upgrading versions recently so it can be upgraded with others without any notice.

jongyoul:
LGTM except for the minor comment

Armadik:
> I think your IDE is making some automatic changes here. The IDE changes need to be avoided as these can differ between developers. Only make small changes.

zeppelin-web/package.json
Armadik: @jongyoul 
Hi, do you need to leave a comment in package.json file with a link to the task?

Armadik:
> I think your IDE is making some automatic changes here. The IDE changes need to be avoided as these can differ between developers. Only make small changes.

I don't have much experience with npm.
I deleted the package-lock.json file, then changed the version of the package "angular-websocket": "^1.0.13" in package.json, and then ran npm install. There were some warnings about package compatibility, but I noticed that they are also in the versions from the master branch. I ran all the commands in the console, maybe there is another way?

Reamer:
> I don't have much experience with npm.

My knowledge is also not really good.

I have modified the package.json as follows.

```
diff --git a/zeppelin-web/package.json b/zeppelin-web/package.json
index c0044146e..d3de12d17 100644
--- a/zeppelin-web/package.json
+++ b/zeppelin-web/package.json
@@ -40,7 +40,7 @@
     "angular-sanitize": "1.5.7",
     "angular-touch": "1.5.7",
     "angular-ui-ace": "0.2.3",
-    "angular-websocket": "1.1.0",
+    "angular-websocket": "1.0.13",
     "angular-xeditable": "0.1.12",
     "bootstrap": "~3.4.1",
     "bootstrap3-dialog": "~1.35.4",
```

`~1.0.13` means `>=1.0.13 <1.1.0-0`, so basically version "1.0.14" can be used. Version `1.0.14` looks also good to me. https://github.com/PatrickJS/angular-websocket/compare/v1.0.13...v1.0.14
`^1.0.13` means `>=1.0.13 < 2.0.0-0`, so basically version "1.1.0" can be used, which has the bug.

To generate an updated package-lock.json use `npm install` inside `zeppelin-web` module.

It is up to you whether you want to use "~" or enter the exact version.

Reamer:
Before using `npm install` you should reset the package-lock.json and delete the `node_modules` folder. This should result in minimal changes in the package-lock.json.

Armadik:
> Before using `npm install` you should reset the package-lock.json and delete the `node_modules` folder. This should result in minimal changes in the package-lock.json.

in the last commit, i did so

Reamer:
My changes are smaller. I have no idea where this comes from.
https://gist.github.com/Reamer/c076906b65667ed2fa6b4fd5a7039508

@jongyoul 
Do we want to merge the pull request like this? I have roughly checked the changes in package-lock.json and have not noticed any major changes.

Reamer (approved):

jongyoul:
I'm also not a FE expert so I just know that the package-lock.json is generated by npm and it's based on package.json. 

@cherrie-k @Henry-Hong Can you help to verify the file?

Henry-Hong:
> It is up to you whether you want to use "~" or enter the exact version.

You can choose to use either “~” or an exact version as @Reamer mentioned, since NPM follows [semver](https://semver.org/).

From what I understand, we need to avoid version 1.1.0 due to the bug, so it doesn’t really matter whether we use “~” or specify the exact version. 

However, in my opinion, it might be better to use the exact version like "angular-websocket": "1.0.14".

Most projects in web front-end development typically use ~ or ^, so specifying the exact version could help highlight the issue and provide context for why this particular version was chosen. (If they trying to find why)

So, LGTM 👍

jongyoul:
Thank you for checking it @Henry-Hong

jongyoul (approved):

PR #4789, opened 2024-08-12
[ZEPPELIN-6051] Skip the hidden files in notebook listing

seung-00:
### What is this PR for?
This PR modifies the notebook listing logic to exclude hidden files that start with `.`. 
This is because showing non-executable files, such as meta files, could confuse users, as seen in this issue [ZEPPELIN-6051](https://issues.apache.org/jira/browse/ZEPPELIN-6051).

Additionally I have enhanced the error popup message to be more specific, which appears when a user opens a notebook with parsing errors. I think it would be more helpful for users to recognize and resolve the issues on their own.

### What type of PR is it?
* Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6051

* Create a notebook that can't be parsed and try to open it.

as-is
Screenshot 2024-08-12 at 01 15 41

~~to-be~~
Screenshot 2024-08-12 at 01 14 15

to-be
image

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java
@@ -58,16 +58,16 @@ public class VFSNotebookRepo extends AbstractNotebookRepo {
   }
 
   @Override
-  public void init(ZeppelinConfiguration conf, NoteParser noteParser) throws IOException {
-    super.init(conf, noteParser);
-    setNotebookDirectory(conf.getNotebookDir());
+  public void init(ZeppelinConfiguration zConf, NoteParser noteParser) throws IOException {
+    super.init(zConf, noteParser);
+    setNotebookDirectory(zConf.getNotebookDir());
   }
 
   protected void setNotebookDirectory(String notebookDirPath) throws IOException {
     URI filesystemRoot = null;
     try {
       LOGGER.info("Using notebookDir: {}", notebookDirPath);
-      if (conf.isWindowsPath(notebookDirPath)) {
+      if (zConf.isWindowsPath(notebookDirPath)) {
         filesystemRoot = new File(notebookDirPath).toURI();
       } else {
         filesystemRoot = new URI(notebookDirPath);
@@ -77,7 +77,7 @@ public class VFSNotebookRepo extends AbstractNotebookRepo {
     }
 
     if (filesystemRoot.getScheme() == null) { // it is local path
-      File f = new File(conf.getAbsoluteDir(filesystemRoot.getPath()));
+      File f = new File(zConf.getAbsoluteDir(filesystemRoot.getPath()));
       filesystemRoot = f.toURI();
     }
     this.fsManager = VFS.getManager();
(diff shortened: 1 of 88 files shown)


zeppelin-web/src/components/websocket/websocket-event.factory.js
```diff
@@ -172,7 +172,7 @@ function WebsocketEventFactory($rootScope, $websocket, $location, baseUrlSrv, sa
         closeByBackdrop: false,
         closeByKeyboard: false,
         title: 'Details',
-        message: _.escape(data.info.toString()),
+        message: _.escape(data.info.toString()).replace(/\n/g, '<br>'),
```
seung-00, line 175: I added newline processing to the error popup message. It could help users better understand the message.

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepoTest.java
```diff
@@ -45,22 +45,21 @@ class VFSNotebookRepoTest {
   private ZeppelinConfiguration zConf;
   private NoteParser noteParser;
   private VFSNotebookRepo notebookRepo;
-  private File notebookDir;
```
seung-00, line 48: I removed this field. I thought the `notebookRepo.rootNotebookFolder` would be enough to check notebook directory.
It is also related to the issue that a directory is created every time the test is run.

seung-00, line 48: I removed this field. I think the notebookRepo.rootNotebookFolder would be enough to check notebook directory.
This is also related to the issue that a temp directory was created every time [this test](https://github.com/apache/zeppelin/pull/4789/files#diff-3dde8f55d9baf47750cc307761685baf82d85588e042984b181baa32893e9552R144) is run. I think it was caused by the directory path data being managed by multiple variables.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java
```diff
@@ -134,10 +137,23 @@ public Note get(String noteId, String notePath, AuthenticationInfo subject)
...
-    note.setPath(notePath);
-    return note;
+
+    try {
+      Note note = noteParser.fromJson(noteId, json);
+      // setPath here just for testing, because actually NoteManager will setPath
+      note.setPath(notePath);
+      return note;
+    } catch (CorruptedNoteException e) {
+      String errorMessage = String.format(
```
Reamer, line 147: Please do not use newlines in short messages. If the error is logged, this makes parsing unnecessarily difficult.

zeppelin-zengine/src/test/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepoTest.java
Reamer: Yes, I agree

zeppelin-web/src/components/websocket/websocket-event.factory.js
Reamer: I don't know what message is being displayed here. If it is the `CorruptedNoteException`, then I think it is unnecessary.
See comment below on the multi-line error message.

zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/VFSNotebookRepo.java
seung-00: @Reamer 
I've removed the newlines. I didn't consider the logging. To me, the popup seems a bit harder to read now. But ensuring operational convenience could be more important.

Reamer (approved):
LGTM

Reamer:
If no further comments are received, I will merge this pull request on Thursday.

PR #4790, opened 2024-08-13
[ZEPPELIN-6033] Add tests for de/serialization

JooHyukKim:
### What type of PR is it?
Improvement

### Todos
* [x] - Write tests

### What is the Jira issue?

[JIRA-ISSUE](https://issues.apache.org/jira/browse/ZEPPELIN-6033)

This PR adds test coverage to existing classes that are part of de/serialization.
Note that this PR is pure in sense that `Gson vs Jackson` is not covered here, but only data processing part.

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
+++ b/zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.zeppelin.cluster.event;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class ClusterMessageSerDeserTest {
+
+    private ClusterMessage clusterMessage;
+
+    @BeforeEach
+    void setUp() {
+        clusterMessage = new ClusterMessage(ClusterEvent.BROADCAST_NEW_PARAGRAPH);
+    }
+
+    @Test
+    void toJson() {
+        // Given
+        clusterMessage.put("key1", "value1");
+        clusterMessage.put("key2", "value2");
+        clusterMessage.setMsgId(UUID.randomUUID().toString());
+
+        // First, let's check the serialized JSON serialization
+        String serializedJSON = ClusterMessage.serializeMessage(clusterMessage);
+        assertEquals(a2q("{" +
+                "'clusterEvent':'JOIN'," +
+                "'data':{'key1':'value1','key2':'value2'}," +
+                "'msgId':'" + clusterMessage.getMsgId() + "'" +
+                "}"), serializedJSON);
+
+        // Now, let's check the serialized JSON deserialization
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage(serializedJSON);
+        assertEquals(clusterMessage.clusterEvent, deserialized.clusterEvent);
+        assertEquals(clusterMessage.getData(), deserialized.getData());
+        assertEquals(clusterMessage.getMsgId(), deserialized.getMsgId());
+    }
+
+    @Test
+    void fromJson_withInvalidField() {
+        String invalidJson = a2q("{" +
+                "'clusterEvent':'JOIN'," +
+                "'data':{'key1':'value1','key2':'value2'}," +
+                "'invalidField':'invalidValue'" +
+                "}");
+
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage(invalidJson);
+        assertEquals(ClusterEvent.SET_RUNNERS_PERMISSIONS, deserialized.clusterEvent);
+        assertEquals(mapOf("key1", "value1", "key2", "value2"), deserialized.getData());
+        assertNull(deserialized.getMsgId());
+    }
+
+    @Test
+    void fromJson_withNothing() {
+        // When only the empty JSON object is provided,
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage("{}");
+
+        // Then, all fields should be null or empty
+        assertNull(deserialized.clusterEvent);
+        assertTrue(deserialized.getData().isEmpty());
+        assertNull(deserialized.getMsgId());
+    }
+
+    @Test
+    void fromJson_withNullFields() {
+        String jsonWithNullFields = a2q("{" +
+                "'clusterEvent':null," +
+                "'data':null," +
+                "'msgId':null" +
+                "}");
+
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage(jsonWithNullFields);
+        assertNull(deserialized.clusterEvent);
+        assertNull(deserialized.getData());
+        assertNull(deserialized.getMsgId());
+    }
+
+    @Test
+    void toJson_withEmptyData() {
+        ClusterMessage original = new ClusterMessage(ClusterEvent.CLEAR_PERMISSION);
+        original.setMsgId("");
+        original.getData().clear();
+
+        String serializedJSON = ClusterMessage.serializeMessage(original);
+        assertEquals(a2q("{" +
+                "'clusterEvent':'JOIN'," +
+                "'data':{}," +
+                "'msgId':''" +
+                "}"), serializedJSON);
+
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage(serializedJSON);
+        assertEquals(original.clusterEvent, deserialized.clusterEvent);
+        assertEquals(original.getData(), deserialized.getData());
+        assertEquals(original.getMsgId(), deserialized.getMsgId());
+    }
+
+    @Test
+    void fromJson_withLargeDataSet() {
+        Map<String, String> largeData = new HashMap<>();
+        for (int i = 0; i < 1000; i++) {
+            largeData.put("key" + i, "value" + i);
+        }
+
+        ClusterMessage original = new ClusterMessage(ClusterEvent.SET_READERS_PERMISSIONS);
+        original.setMsgId(UUID.randomUUID().toString());
+        original.put(largeData);
+
+        String serializedJSON = ClusterMessage.serializeMessage(original);
+        ClusterMessage deserialized = ClusterMessage.deserializeMessage(serializedJSON);
+
+        assertEquals(original.clusterEvent, deserialized.clusterEvent);
+        assertEquals(original.getData(), deserialized.getData());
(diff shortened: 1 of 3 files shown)


JooHyukKim:
### Closing (will re-open after fix)

The tests are failling. Seems like I didn't commit everything upstream.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
```diff
@@ -0,0 +1,156 @@
...
+    ) {
+        Map<String, String> map = new HashMap<>();
+        map.put(key, value);
+        map.put(key2, value2);
+        return map;
+    }
+
+    private static String a2q(String json) {
+        return json.replace("'", "\"");
+    }
```
seung-00, line 155: Is a2q really necessary? If not, I think using a pure JSON format without a2q could make the test clearer. Keeping the test focused on the target, which in this case is JSON serialization/deserialization, could help keep the test simple. 
Also, in this PR HeliumRegistrySerializerTest does not use a2q.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+            "  \"className\" : \"your.package.name.YourApplicationClass-A\",\n" +
+            "  \"resources\" : [\n" +
+            "    [\"resource.name\", \":resource.class.name\"],\n" +
+            "    [\"alternative.resource.name\", \":alternative.class.name\"]\n" +
+            "  ],\n" +
+            "  \"icon\" : \"<i class='icon'></i>\"\n" +
+            "}";
+
+    @Test
+    void testDeserialization() throws IOException {
```
seung-00, line 48: How about unifying the method names? 
toJson/fromJson in RemoteZeppelinServerResourceTest vs. deserialization/serialization in HeliumRegistrySerializerTest.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/RemoteZeppelinServerResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+                "}"), serializedJSON);
+
+        RemoteZeppelinServerResource deserialized = RemoteZeppelinServerResource.fromJson(serializedJSON);
+        assertEquals(original.getOwnerKey(), deserialized.getOwnerKey());
+        assertEquals(original.getResourceType(), deserialized.getResourceType());
+        assertEquals(original.getData(), deserialized.getData());
+    }
+
+    @Test
+    void fromJson_withLargeDataSet() {
```
seung-00, line 116: It seems like a performance test. In my opinion, it's better to avoid performance testing in unit tests. I think unit tests and performance tests have different purposes.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class HeliumRegistrySerializerTest {
+
+    private String HELIUM_PACKAGE_JSON_1 = "{\n" +
```
seung-00, line 35: Why is there a "1" at the end of the variable? 👀

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/RemoteZeppelinServerResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertEquals(original.getOwnerKey(), deserialized.getOwnerKey());
+        assertEquals(original.getResourceType(), deserialized.getResourceType());
+        assertEquals(original.getData(), deserialized.getData());
+    }
+
+    private Map<String, String> mapOf(
+            String key, String value,
+            String key2, String value2
+    ) {
+        // TreeMap to ensure serialization order
```
seung-00, line 139: The ClusterMessageSerDeserTest uses a HashMap. Are there no issues like this

zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class ClusterMessageSerDeserTest {
```
seung-00, line 28: The name `SerDeser` seems a bit implicit to me. In my opinion, a more clear name, even if longer, would be better.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+            "}";
+
+    @Test
+    void testDeserialization() throws IOException {
+        // Given
+
+        // When
+        // Define the output file
+        Path dir = Files.createDirectory(Paths.get("./" + UUID.randomUUID().toString()));
+        File newFile = Files.createTempFile(dir, UUID.randomUUID().toString(), ".json").toFile();
```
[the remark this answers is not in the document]
JooHyukKim: Clean up is already implemented. 
There is a delete- hook at line 56.
What I think we need is moving the delete-hook line up to right below this line.
WDYT, @seung-00 ?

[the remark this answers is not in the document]
seung-00: @JooHyukKim 
You're right. But for me, the temporary directory (path: /zeppelin-zengine) isn't being deleted. Could you check please?

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(2, heliumPackage.getResources().length);
+        assertEquals("resource.name", heliumPackage.getResources()[0][0]);
+        assertEquals(":resource.class.name", heliumPackage.getResources()[0][1]);
+        assertEquals("alternative.resource.name", heliumPackage.getResources()[1][0]);
+        assertEquals(":alternative.class.name", heliumPackage.getResources()[1][1]);
+        assertEquals("<i class='icon'></i>", heliumPackage.getIcon());
+
+    }
+
+    private File createHeliumPackage(Path testDir, String JSON) throws IOException {
```
seung-00, line 80: This method isn't used. Is it necessary? 🤔

seung-00, line 80: This method isn't used. If it's not necessary, how about removing it?

zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
JooHyukKim: > Is a2q really necessary? If not, I think using a pure JSON format without a2q could make the test clearer. Keeping the test focused on the target, which in this case is JSON serialization/deserialization, could help keep the test simple. 

Hmmm..  You are right method `a2q()` sure is not necessary. But from a different point of view, `a2q()` is just a little helper method to make it easier to read the json input so we can focus more on JSON serialization/deserialization. Some people do use helper methods to actually focus on the SUT in this case the JSON serialization/deserialization. 

What I would like to see from you is some guidelines or references on how helper methods should be written so tests are more on focus.

> Also, in this PR HeliumRegistrySerializerTest does not use a2q.

Agreed 👏🏽👏🏽 I will use `a2q` in all tests introduced in this PR

JooHyukKim: Sure I will remove a2q methods.
I disagree with you and a2q methods don't hinder us from focusing on the SUT.
I just don't have much preference over `a2q` here.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        // Write JSON string to file using Apache Commons FileUtils
+        HeliumRegistry registry = new HeliumLocalRegistry("my-registry", dir.toString());
+
+        // Then
+        assertEquals("my-registry", registry.name());
+        assertNotNull(registry.getAll());
+        assertEquals(1, registry.getAll().size());
+
+        HeliumPackage heliumPackage = registry.getAll().stream().findFirst().orElseThrow(RuntimeException::new);
```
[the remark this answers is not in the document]
JooHyukKim: This is not an action at all, just another value assertion following...

```java
assertEquals(1, registry.getAll().size());
```
But I do think it could be simplified.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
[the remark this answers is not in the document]
JooHyukKim: Strange, I added delete-on-exit hook on the directory too.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/RemoteZeppelinServerResourceTest.java
JooHyukKim: Hmmmm🤔 , I don't think so. 
Or do you have anything in mind?

zeppelin-interpreter/src/test/java/org/apache/zeppelin/cluster/event/ClusterMessageSerDeserTest.java
JooHyukKim: Makes sense 👍🏼Will change.

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/RemoteZeppelinServerResourceTest.java
JooHyukKim: Oh my bad. I think the test name made you get the wrong idea 😅.
The test is intended to test (sort of) long data as input.
There is a whole lot of Zeppelin integration tests that run wayyyy over a minute 😆
This test won't even leave a dent on the performance of Zeppelin test suite.

I changed the name from `LargeDataSet` to `LongDataset`.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
JooHyukKim: Done!

JooHyukKim:
Thank you @seung-00 for review!

zeppelin-interpreter/src/test/java/org/apache/zeppelin/interpreter/RemoteZeppelinServerResourceTest.java
seung-00: @JooHyukKim 
Does using long data actually change the functionality here? If it's just a unit test, the `toJson` test in this class already seems to cover the same function.

zeppelin-zengine/src/test/java/org/apache/zeppelin/helium/HeliumRegistrySerializerTest.java
seung-00: @JooHyukKim 
If the '1' doesn't have a specific meaning, how about removing it?

PR #4791, opened 2024-08-13
[ZEPPELIN-6049] Change hadoop-client-api scope in R module to fix class not found error

tbonelee:
### What is this PR for?
Fixes an issue where IR interpreter fails to initialize `SparkRBackend` due to the inability to find `org.apache.hadoop.fs.FSDataInputStream`.

The dependency scope of the `hadoop-client-api` module, which includes `FSDataInputStream`, was set to `provided`(inherited from the root).
The method that failed was invoked using Java reflection, which may have prevented the class from being found at runtime.
Therefore, I changed the dependency scope of the `hadoop-client-api` module to `compile`.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-6049/

Build Zeppelin and run some code with `%r.ir` interpreter.

#### Before fix
image

#### After fix
image

### Questions:

### DIFF
--- a/rlang/pom.xml
+++ b/rlang/pom.xml
@@ -116,6 +116,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client-api</artifactId>
+            <scope>compile</scope>
         </dependency>
 
         <dependency>


Reamer:
How do you build Zeppelin? Some platforms e.g. YARN already have the Hadoop classes in the classpath as far as I remember, which is why the value is set to provided by default.
However, if you run Zeppelin in Kubernetes or Docker, for example, this is not the case. Here you have to set the profile `include-hadoop` when building. Take a look at the [Dockerfile](https://github.com/apache/zeppelin/blob/afc4141b97827918de4465a708b59ba39692398d/Dockerfile#L24).

tbonelee:
@Reamer 
Understood. It works when I set the `include-hadoop` profile as you suggested. I was running it in local mode.
So, does this mean that I should always set the `include-hadoop` profile when building, if I am not using Yarn launcher?

Reamer:
If you are not using YARN, I would recommend this. There is also the possibility to make the Hadoop library available with the environment variable `HADOOP_CONF_DIR` at startup.
See https://github.com/apache/zeppelin/blob/fb64356e57c5f3e9280d6eb9cea96858efcb356a/bin/zeppelin-daemon.sh#L86-L98

Or corresponding positions in https://github.com/apache/zeppelin/blob/master/bin/interpreter.sh
But I don't see anything suitable for R here.

tbonelee:
Thanks for checking this issue! I'll try what you've suggested.

PR #3449, opened 2019-09-17
[ZEPPELIN-4333] escape dialog title and message

Leemoonsoo:
### What is this PR for?
There're couple of places in the code that dialog title and message is not being escaped

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4333

### Questions:
* Does the licenses files need update? no
* Is there breaking changes for older versions? no
* Does this needs documentation? no

PR #4795, opened 2024-08-17
[HOTFIX] Escaping vulnerable text  when modal opens

Henry-Hong:
#### Description
- Relate : https://github.com/apache/zeppelin/pull/3449
- Thanks to above PR, using Helium got much safer. But there're still remaning vulnerability issues with Hellium package.
- If somebody makes malicious Hellium package, it can lead to serious consequences.

#### Solution
- `zeppelin-web` is dependent on `lodash@3.9.3` which supports escaping method `_.escape`
- I applied `_.escape(vulnerable-text)` to vulnerable text.

### AS-IS
화면 기록 2024-08-17 오후 3 57 43
image
_You can see the HTML is not escaped and `` tag has rendered_

### TO-BE
image

image
_HTML is escaped and `<`, `>` has converted into HTML Entity which is safer_

### What type of PR is it?
Hot Fix

### Todos
* [x] - Escape HTML message before renders

### What is the Jira issue?
- N/A

- N/A

### Questions:
* Does the license files need to update? NO

### DIFF
--- a/zeppelin-web/src/app/helium/helium.controller.js
+++ b/zeppelin-web/src/app/helium/helium.controller.js
@@ -240,10 +240,10 @@ export default function HeliumCtrl($scope, $rootScope, $sce,
           `<div style="color:gray">${getHeliumTypeText(type)}</div>` +
           '<hr style="margin-top: 10px; margin-bottom: 10px;" />' +
           '<div style="font-size: 14px;">Description</div>' +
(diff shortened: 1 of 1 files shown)


jongyoul:
Thank you for the contribution! By the way, is there any possible field with the same behaviors? I think it looks good but just in case.

jongyoul (approved):

PR #4797, opened 2024-08-17
[ZEPPELIN-6060] Integrate Thrift code generation with Maven build and enhance genthrift.sh

jhl8109:
### What is this PR for?
This PR aims to enhance the build process in Zeppelin by integrating Thrift Java file generation into the Maven build workflow. It also updates the `genthrift.sh` script to automatically handle Thrift installation. These changes are intended to reduce manual errors and ensure consistent build environments.

### What type of PR is it?
Improvement

### Todos
* [ ] - Add `exec-maven-plugin` to `pom.xml` to automate the generation of Java files from Thrift IDL during the Maven build process.
* [ ] - Enhance `genthrift.sh` to support automatic Thrift installation, including improved checks for Thrift being installed.

### What is the Jira issue?
* https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6060

1. Verify Thrift installation
2. Confirm Thrift .java file generation through Maven

### Questions:
* Does the license files need to update? NO

### DIFF
--- a/zeppelin-interpreter/pom.xml
+++ b/zeppelin-interpreter/pom.xml
@@ -112,7 +112,7 @@
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
       <version>1.9.4</version>
-        <exclusions>
+      <exclusions>
         <!-- using jcl-over-slf4j instead -->
         <exclusion>
           <groupId>commons-logging</groupId>
@@ -240,6 +240,42 @@
           <failOnNoGitDirectory>false</failOnNoGitDirectory>
           <dateFormat>yyyy-MM-dd HH:mm:ss</dateFormat>
         </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>3.0.0</version>
+        <executions>
+          <execution>
+            <!-- This execution runs during the compile phase to generate Java files from Thrift files -->
+            <id>generate-thrift-sources</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>bash</executable>
+              <arguments>
+                <argument>${basedir}/src/main/thrift/genthrift.sh</argument>
+              </arguments>
+            </configuration>
+          </execution>
+          <execution>
+            <!-- This execution can be triggered manually to install Thrift by running the following command: -->
+            <!-- mvn exec:exec@install-thrift -->
+            <id>install-thrift</id>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>bash</executable>
+              <arguments>
+                <argument>${basedir}/src/main/thrift/genthrift.sh</argument>
+                <argument>install</argument>
+              </arguments>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
 
--- a/zeppelin-interpreter/src/main/thrift/genthrift.sh
+++ b/zeppelin-interpreter/src/main/thrift/genthrift.sh
@@ -17,13 +17,176 @@
 # * limitations under the License.
 # */
 
-rm -rf gen-java
-rm -rf ../java/org/apache/zeppelin/interpreter/thrift
-thrift --gen java RemoteInterpreterService.thrift
-thrift --gen java RemoteInterpreterEventService.thrift
-for file in gen-java/org/apache/zeppelin/interpreter/thrift/* ; do
-  cat java_license_header.txt ${file} > ${file}.tmp
-  mv -f ${file}.tmp ${file}
-done
-mv gen-java/org/apache/zeppelin/interpreter/thrift ../java/org/apache/zeppelin/interpreter/thrift
-rm -rf gen-java
+
+THRIFT_VERSION="0.13.0"
+THRIFT_URL="https://archive.apache.org/dist/thrift/${THRIFT_VERSION}/thrift-${THRIFT_VERSION}.tar.gz"
+
+check_thrift_installed() {
+    if command -v thrift &> /dev/null; then
+        INSTALLED_VERSION=$(thrift -version 2>&1 | awk '{print $3}')
+        if [ "$INSTALLED_VERSION" == "$THRIFT_VERSION" ]; then
+            return 0  # Return 0 if the same version is installed
+        else
+            echo "Installed Thrift version is $INSTALLED_VERSION, but required version is $THRIFT_VERSION."
+            return 1  # Return 1 if a different version is installed
+        fi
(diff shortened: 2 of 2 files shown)


Reamer:
Installation of packages via `yum` or `apt-get` requires `root` rights.
Maven runs without `root` rights.

I also looked at the problem with `thrift` some time ago and hoped that the Java classes could be generated with a Maven plugin. Similar to GRPC. Unfortunately, I have not found anything.


My goal was just to have the thrift files in git and generate the thrift classes each time. But due to the installation process I don't see a way to do this. How do other Thrift projects implement it?

JooHyukKim:
Did you come up with `genthrift.sh` script all by yourself, or is there reference(or similar script) you referred to and share with us? Like @Reamer says, it would be good to refer to how other projects handle situations like this.

zeppelin-interpreter/src/main/thrift/genthrift.sh
```diff
@@ -17,13 +17,176 @@
...
+    make
+    make install
+
+    echo "Thrift ${THRIFT_VERSION} installed successfully."
+}
+
+install_thrift_macos() {
+  echo "Installing Thrift ${THRIFT_VERSION} on macOS..."
+  if command -v brew &> /dev/null; then
+    brew install thrift@0.13
```
pan3793, line 112: this does not work. and formulae with the suffix `name@<version>` won't be added into PATH automaticlly.

```
$ brew install thrift@0.13
Warning: No available formula with the name "thrift@0.13". Did you mean thrift@0.9?
==> Searching for similarly named formulae and casks...
==> Formulae
thrift@0.9

To install thrift@0.9, run:
  brew install thrift@0.9
```

pan3793:
integrating thrift codegen into the regular build process makes the developer hard to set up the environment, because thrift community does not publish the convenient `thrift` binary on releases, the downstream developers have to build thrift from source or get from the platform package managers like apt/brew, neither ways are convenient due it additional native deps requirements or only a few of versions providing.

given the generated thrift code won't change frequently, tracking the generated thrift code and providing a script to regenerate it is a common way. I think we can make some improvements in `genthrift.sh`, for example, make it working dir agnostic, allow passing env `THRIFT_EXEC` to use a custom `thrift`

jongyoul:
Thank you for the discussion. My first goal is to ask him to divide thrift definition and generated files with zeppelin-intepreter module to minimize itself and clean up the module. During this job, I thought it would be better to generate thrift files easier. By the way, in my past experience, I provided all binaries for thrift compilers into a repository. Can we choose a similar way? You also can use docker to build the thrift files. Current implementation looks good personally but as others said it might be a quite difficult. WDYT?

zeppelin-interpreter/src/main/thrift/genthrift.sh
```diff
@@ -17,176 +17,25 @@
 # * limitations under the License.
 # */
 
+# Determine the base directory dynamically (directory where the script is located)
+THRIFT_EXEC="$(dirname "$(realpath "$0")")"
```
jhl8109, line 21: @pan3793 It seems feasible to implement the configuration to operate independently of the working directory using THRIFT_EXEC="$(dirname "$(realpath "$0")")". This approach allows for determining the location of the thrift script file itself and executing the thrift operations from that location.

jhl8109:
@jongyoul Installing thrift and helping with its setup could be quite complicated and challenging. Therefore, I considered using Docker, but the current Docker thrift image has been deprecated since [version 0.12.0](https://hub.docker.com/_/thrift), and there is no support for the current version, 0.13.0. 

However, through @Reamer's review, I found an alternative approach by exploring the thrift-related code here: https://github.com/apache/hbase/blob/master/hbase-thrift/pom.xml

Looking at this code, instead of supporting the installation directly, we could consider an approach where we simply check if thrift is installed and proceed only if it is. This method could reduce the build burden for users who don't have thrift installed while maintaining build consistency for those who do. 
However, this approach might only be feasible because, in the example above, the thrift code is separated into its own module.

zeppelin-interpreter/src/main/thrift/genthrift.sh
Reamer: I understand your approach. However, the path is not the Thrift.exe, but the Zeppelin Thrift folder. Please change the variable name, it has just confused me a lot.

zeppelin-interpreter/pom.xml
```diff
@@ -241,6 +241,42 @@
           <dateFormat>yyyy-MM-dd HH:mm:ss</dateFormat>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>3.0.0</version>
+        <executions>
+          <execution>
+            <!-- This execution runs during the compile phase to generate Java files from Thrift files -->
```
Reamer, line 250: Thrift is not installed in the CI pipeline, so these steps can only be executed using profiles similar to hbase-thrift.

zeppelin-interpreter/pom.xml
```diff
@@ -241,6 +241,42 @@
...
+            <!-- mvn exec:exec@install-thrift -->
+            <id>install-thrift</id>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>bash</executable>
+              <arguments>
+                <argument>${basedir}/src/main/thrift/genthrift.sh</argument>
+                <argument>install</argument>
```
Reamer, line 274: genthrift.sh does not support arguments, so the execution can be removed.

Reamer (changes requested):
I like your changes in `genthrift.sh`, especially the comments. I think the changes in pom.xml are unnecessary.

zeppelin-interpreter/src/main/thrift/genthrift.sh
pan3793: `THRIFT_EXEC` should point to the thrift command directly.

pan3793: why I prefer `THRIFT_EXEC` instead of `THRIFT_HOME`?

the `thrift` binary has no extra dependency and can be copied to any folder directly, for example, I put the thrift binaries under `$HOME/bin`

```
$ ll bin | grep thrift
lrwxr-xr-x@ 1 chengpan  staff    13B Aug 19 19:35 thrift -> thrift-0.16.0
-rwxr-xr-x@ 1 chengpan  staff   3.4M Aug 19 11:13 thrift-0.14.1
-r-xr-xr-x@ 1 chengpan  staff   3.3M Aug 19 18:35 thrift-0.16.0
-r-xr-xr-x@ 1 chengpan  staff   3.6M Aug  1 10:15 thrift-0.20.0
-rwxr-xr-x@ 1 chengpan  staff   2.9M Aug 19 11:12 thrift-0.9.3
```

PR #4801, opened 2024-08-23
[ZEPPELIN-6063] Fix delay in paragraph results rendering in new UI

tbonelee:
### What is this PR for?

In new UI, the results are rendered only after some interactions, such as a click on the other paragraph.

This PR addresses the issue by ensuring immediate rendering of the results when they are set in the `ngAfterViewInit` lifecycle hook. Previously, `markForCheck()` was used after setting `this.innerHTML`, which scheduled the component to be checked in the next change detection cycle but did not trigger an immediate update.
This has been changed to `detectChanges()` to force an immediate change detection, ensuring that the results are rendered right after they are set, without needing further user interaction.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/6063

- Run any paragraph in new UI

#### Before

Aug-23-2024 19-59-47

#### After

Aug-23-2024 20-27-01

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
@@ -224,10 +224,7 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit, 
         this.renderAngular();
         break;
     }
-    this.cdr.markForCheck();
-    if (this.published) {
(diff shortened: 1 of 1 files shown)


jongyoul (approved):
LGTM!

pan3793:
Thanks, merged to master

PR #4802, opened 2024-08-25
[ZEPPELIN-6064] Change default web UI to new UI

tbonelee:
### What is this PR for?

For several years, Zeppelin has supported both the old and new UIs, with the old UI remaining as the default.
However, as the new UI offers a more modern design, I believe it might be beneficial to consider making it the default interface.

#### Rationale
The new UI presents a more contemporary user experience, which could align better with users' expectations today.
Transitioning to the new UI as the default may help in keeping Zeppelin more up-to-date and user-friendly.
Of course, I am aware that this change might impact various parts of the project, so I propose we carefully review this before proceeding.

### What type of PR is it?
Improvement

### Todos
* [x] - Update all references to the web app context paths.

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/6064

- Verify that the new UI is accessible via root path.
- Ensure the old UI can still be accessed via `/classic`.

### Questions:
* Is there breaking changes for older versions? No, but it may affect users' accustomed access paths.
* Does this needs documentation? Maybe it would be helpful to provide guidance on how users could access the old UI under the new path, especially for those accustomed to the previous default. However, I'm not entirely sure if this is the best approach, so I haven't added any documentation for now.


tbonelee:
@jongyoul I've made a PR to set the new UI as the default. Please review it when you have some time.

kevinjmh:
I would like to switch to the new UI too, but tests and bugs fixed should be ahead.

kevinjmh:
image
I tried several options above, short-cut does not work, clone only create an empty paragraph without content

jongyoul:
By the way, I assume we can deploy two versions of Apache Zeppelin with current and new UI in some ways like setting parameters when packaging them and so on. However, your change supports only running Zeppelin with the new UI. Do you have any idea? :-)

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -226,8 +226,8 @@ protected void configure() {
         });
 
     // Multiple Web UI
-    final WebAppContext defaultWebApp = setupWebAppContext(contexts, zConf, zConf.getString(ConfVars.ZEPPELIN_WAR), zConf.getServerContextPath());
-    final WebAppContext nextWebApp = setupWebAppContext(contexts, zConf, zConf.getString(ConfVars.ZEPPELIN_ANGULAR_WAR), WEB_APP_CONTEXT_NEXT);
+    final WebAppContext defaultWebApp = setupWebAppContext(contexts, zConf, zConf.getString(ConfVars.ZEPPELIN_WAR), WEB_APP_CONTEXT_CLASSIC);
```
Reamer, line 229: I think it makes sense to rename the Java variables as well.

zeppelin-web-angular/src/app/share/header/header.component.html
```diff
@@ -67,7 +67,7 @@
           <li nz-menu-item (click)="logout()">Logout</li>
         </ng-container>
         <li nz-menu-divider></li>
-        <li nz-menu-item><a href="/">Old Version</a></li>
+        <li nz-menu-item><a href="/classic">Old Version</a></li>
```
Reamer, line 70: Perhaps “Classic Version” should be used here instead of “Old Version”.
```suggestion
        <li nz-menu-item><a href="/classic">Classic Version</a></li>
```

jongyoul:
>We have the 'web-angular' profile, which builds the new user interface. This should definitely be switched on by default when the default interface is changed.
As far as I can see, this change does not result in running Zeppelin with the new UI.

Yes, correct. Moreover, even if we include the profile, we still need to change some server code as ZeppelinServer has a hardcoded configuration for the current and new UI.

By the way, for the improvement of the new UI, deploying the new UI version would be helpful.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
tbonelee: I missed that part. Thanks for catching it! 🙏

zeppelin-web/src/components/navbar/navbar.html
```diff
@@ -106,7 +106,7 @@
               <li ng-if="ticket.principal && ticket.principal !== 'anonymous'" role="separator" style="margin: 5px 0;" class="divider"></li>
               <li ng-if="ticket.principal && ticket.principal !== 'anonymous'"><a ng-click="navbar.logout()">Logout</a></li>
               <li role="separator" style="margin: 5px 0;" class="divider"></li>
-              <li><a href="/next">Try the new Zeppelin</a></li>
+              <li><a href="/">Try the new Zeppelin</a></li>
```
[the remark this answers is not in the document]
tbonelee: Are you suggesting that we should change the "Try the new Zeppelin" text?
I initially decided to keep it the same even after switching the default UI, thinking it would be natural for old UI users to see the same wording guiding them to the new UI.
Do you have any alternative wordking in mind?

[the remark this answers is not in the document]
pan3793: might be "Classic UI" vs. "Default UI"

pan3793:
Strongly +1 for making the Angular UI as default even though there are some bugs need to be fixed, and renaming the existing one to "Classic UI".

zeppelin-web/src/components/navbar/navbar.html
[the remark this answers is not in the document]
tbonelee: As you mentioned, using the term "Default UI" seems natural.
How about this phrase: "Switch to the default UI"

zeppelin-web-angular/src/app/share/header/header.component.html
```diff
@@ -67,7 +67,7 @@
           <li nz-menu-item (click)="logout()">Logout</li>
         </ng-container>
         <li nz-menu-divider></li>
-        <li nz-menu-item><a href="/">Old Version</a></li>
+        <li nz-menu-item><a href="/classic">Classic Version</a></li>
```
pan3793, line 70: should we use the same words? I suggest using either
1. "Classic UI" vs. "Default UI" or
2. "Switch to Classic UI" vs. "Switch to Default UI"

zeppelin-web/src/components/navbar/navbar.html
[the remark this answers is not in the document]
pan3793: @tbonelee I replied here https://github.com/apache/zeppelin/pull/4802/files#r1731289824

zeppelin-web-angular/src/app/share/header/header.component.html
tbonelee: I prefer a more specific phrasing, so I'll try 'Switch to Classic UI' vs. 'Switch to Default UI'.

Reamer (changes requested):
Please revert the change regarding renaming. This leads to an unnecessary number of changes. I think the profile change in pom.xml makes sense.

tbonelee:
@Reamer 
I have reverted the commits related to 2. and 3.
Is there anything else that needs to be reverted?

pan3793:
code changes lgtm, I have not tested it locally though.

Reamer:
The code changes are now much clearer. Please perform a rebase to the current master branch, as there are merge conflicts. Furthermore, I think the Selenium tests for the “Classic” UI need to be adjusted.

pan3793:
@jongyoul The current release bundles both classic and new UI, and allows users to switch to each other by clicking a button, the proposed approach "switch new UI to default UI" looks straightforward to me, and users still have the option to switch back to the classic UI by just clicking a button if they find bugs in the new UI.

jongyoul:
> @jongyoul The current release bundles both classic and new UI, and allows users to switch to each other by clicking a button, the proposed approach "change new UI to default UI" looks straightforward to me, and users still have the option to switch back to the classic UI by just clicking a button if they find bugs in the new UI.

Yes, right. we can switch for now but I think it would be better to change gradually so my idea is to provide a new version with a separate package like zeppelin-bin-new-ui.tgz and so on for the next release. Then, we can check the community's feedback and change the default permanently.

Reamer:
I would have liked more feedback from the community and improvements with the current solution. Unfortunately, there was only sporadic feedback.
I see the change of user interfaces as uncritical, especially as it is easy to switch between the two.
In my opinion, a separate release makes the release process and implementation unnecessarily complicated.

jongyoul:
Yes, I hope so. For the release project, we can edit the release script to release two versions but we still need to change some code in ZeppelinServer and so on.

tbonelee:
After rebasing, I updated the URLs for the Selenium tests, and I also updated the test URLs in `DirAccessTest`

Reamer:
@tbonelee Please check the CI

tbonelee:
The Selenium test passed in my personal repository's GitHub Actions. Therefore I assumed it was a temporary issue and did not make any change.
I've tried to fix the remaining broken tests. Please review and confirm.

pan3793 (approved):
I have tested this patch locally
- Switch to Classic/Default UI button works well
- Successfully run simple Spark jobs with both classic and default UIs


image image

jongyoul:
Thank you for checking it. By the way, I'm still worried about providing the new UI as a default for the next release but let's see how's going.

I suggest that we provide an option to store the type of UI, which means we should use the same type of UI after we restart the Zeppelin. Could you please add the configuration? Otherwise, could you please create a ticket for it?

tbonelee:
@jongyoul 
Are you suggesting that when someone enters the top-level URL of Zeppelin, they should be redirected to the most recently accessed UI? Or did you have sth else in mind?
In any case, I think there might be additional aspects we need to discuss, so it might be a good idea to create a new issue ticket.
I'll go ahead and register a new issue.

pan3793:
> when someone enters the top-level URL of Zeppelin, they should be redirected to the most recently accessed UI?

Hmm... maybe we just need a configuration to allow the administrator to decide the default UI?

tbonelee:
I've made a new issue regarding UI type persistence feature.
I believe there are some points that need further discussion. Please feel free to comment on the issue.
https://issues.apache.org/jira/browse/ZEPPELIN-6074

jongyoul:
As Cheng said, I thought we would have a configuration storing the type of UI and the Zeppelin server uses it. But, we don't have that kind of configuration yet so we'd better discuss it.

Reamer (approved):
LGTM
Thank you for your contribution.
As long as no further comments are received, I will merge this change into the master on Wednesday.

jongyoul (approved):

pan3793:
As all comments are addressed, merging to master

PR #4824, opened 2024-09-11
[ZEPPELIN-6085] Add configuration option for setting the default UI

tbonelee:
### What is this PR for?

This PR introduces a new configuration option `zeppelin.default.ui` to set the default UI for Zeppelin.
The available options for this setting are `new` and `classic`.
By default, the value is set to `new`, meaning the new UI will be served at the root path (`"/"`), while the classic UI will be available at `"/classic"`.
If the configuration is set to `classic`, the classic UI will be served at `"/"`, and the new UI at `"/new"`.

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/6085

- Testing the `new` option :
  - Set the `zeppelin.default.ui` config to `new` (the default).
  - Verify that the new UI is served at the root (`"/"`) and the classic UI is available at `"/classic"`.
  - Verify that navigating between pages works without any issue.
  - Verify that the UI switching button works properly in both UIs.
- Testing `classic` option :
  - Similar steps as the above, except the new UI is served at `"/classic"` and the classic UI is available at the root (`"/"`)

### Questions:
* Does this needs documentation? Yes, documentation for the new configuration option has been added.

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -80,6 +80,11 @@ public class ZeppelinConfiguration {
     DOCKER
   }
 
+  public enum DEFAULT_UI {
+    NEW,
+    CLASSIC
+  }
+
   // private constructor, so that it is singleton.
   private ZeppelinConfiguration(@Nullable String filename) {
      try {
@@ -903,6 +908,10 @@ public class ZeppelinConfiguration {
 
   public boolean isPrometheusMetricEnabled() {
     return getBoolean(ConfVars.ZEPPELIN_METRIC_ENABLE_PROMETHEUS);
+  }
+
+  public DEFAULT_UI getDefaultUi() {
+    return DEFAULT_UI.valueOf(getString(ConfVars.ZEPPELIN_DEFAULT_UI).toUpperCase());
   }
 
   /**
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
@@ -965,6 +965,8 @@ public class ZeppelinConfiguration {
     ZEPPELIN_WAR("zeppelin.war", "zeppelin-web/dist"),
     ZEPPELIN_ANGULAR_WAR("zeppelin.angular.war", "zeppelin-web-angular/dist/zeppelin"),
     ZEPPELIN_WAR_TEMPDIR("zeppelin.war.tempdir", "webapps"),
+    // "new" or "classic"
+    ZEPPELIN_DEFAULT_UI("zeppelin.default.ui", "new"),
     ZEPPELIN_JMX_ENABLE("zeppelin.jmx.enable", false),
     ZEPPELIN_JMX_PORT("zeppelin.jmx.port", 9996),
 
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
@@ -158,7 +158,6 @@ public class ZeppelinServer implements AutoCloseable {
     jettyWebServer = setupJettyServer();
     sharedServiceLocator = ServiceLocatorFactory.getInstance().create(serviceLocatorName);
     storage = ConfigStorage.createConfigStorage(zConf);
-
   }
 
   public void startZeppelin() {
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
@@ -63,6 +63,7 @@ import org.apache.shiro.web.servlet.ShiroFilter;
 import org.apache.zeppelin.cluster.ClusterManagerServer;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.conf.ZeppelinConfiguration.ConfVars;
+import org.apache.zeppelin.conf.ZeppelinConfiguration.DEFAULT_UI;
 import org.apache.zeppelin.display.AngularObjectRegistryListener;
 import org.apache.zeppelin.healthcheck.HealthChecks;
 import org.apache.zeppelin.helium.ApplicationEventListener;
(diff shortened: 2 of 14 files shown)


Reamer (changes requested):
I have made some comments on the Java class. I don't know if the changes are possible in the Java script frontends. Aren't the URLs to other elements (images, CSS etc.) fixed during the building of the frontends?

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -333,6 +345,7 @@ private void initJMX() {
       }
     }
   }
+
```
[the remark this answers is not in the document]
tbonelee: Okay I removed the line

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -141,6 +142,8 @@ public class ZeppelinServer implements AutoCloseable {
   private final Server jettyWebServer;
   private final ServiceLocator sharedServiceLocator;
   private final ConfigStorage storage;
+  private final String classicUiWebAppContextPath;
```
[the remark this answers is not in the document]
tbonelee: I moved them closer to where they are used inside the `startZeppelin()`.

tbonelee:
@Reamer 
It seems like it could depend on the situation.
In my opinion, if the resource URLs are set as fixed absolute paths, there could be an issue, but if they are relative paths, there shouldn't be a problem.

In the initial code I wrote, I added code on the server side to insert a `base` tag into the `index.html` at runtime.
Howver, when I ran the app and checked the network tab, resources were being fetched from the correct paths even without the `base` tag, so I removed it.

From reviewing the frontend code, my assumption is that most of the resources are set with relative paths, or they dynamically generate absolute paths based on each app's path, so there didn't seem to be an issue.

However, upon reconsideration, I thought explicitly specifying the `base` tag would be a more reliable soltuion, so I added the code back in.

tbonelee:
Additionally, I missed this part earlier, but for the `MathJax` library, the library root path for resolving the module loading URL needs to be changed.
However, it was set to always fetch from the root regardless of the app's path. (In fact, this caused the new UI app, which was served from the `/next` path, to fetch the MathJax code from the classic UI app's resource path.)

So, I made additional modifications to ensure this part changes according to the app's path.

jongyoul:
It includes various types of code :-) Thank you for your work. I'll test it sooner than later.

tbonelee:
@Reamer Just pinging to review :)

tbonelee:
I've rebased this branch and added documentation for configuring the default Ui on the `explore_ui` page.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -157,6 +158,7 @@ public ZeppelinServer(ZeppelinConfiguration zConf, String serviceLocatorName) th
     jettyWebServer = setupJettyServer();
     sharedServiceLocator = ServiceLocatorFactory.getInstance().create(serviceLocatorName);
     storage = ConfigStorage.createConfigStorage(zConf);
+
```
pan3793, line 161:
```suggestion
```

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -348,7 +360,6 @@ private void initMetrics() {
     new UptimeMetrics().bindTo(Metrics.globalRegistry);
     new JVMInfoBinder().bindTo(Metrics.globalRegistry);
   }
-
   public void shutdown(int exitCode) {
```
pan3793, line 363:
```suggestion

  public void shutdown(int exitCode) {
```

jongyoul (approved):

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
```diff
@@ -905,6 +910,10 @@ public boolean isPrometheusMetricEnabled() {
     return getBoolean(ConfVars.ZEPPELIN_METRIC_ENABLE_PROMETHEUS);
   }
 
+  public DEFAULT_UI getDefaultUi() {
+    return DEFAULT_UI.valueOf(getString(ConfVars.ZEPPELIN_DEFAULT_UI).toUpperCase());
```
Reamer, line 914: What happens if I pass neither `new` nor `classic` as a configuration value at this point?

zeppelin-web-angular/src/index.html
```diff
@@ -44,6 +43,14 @@ <h1 class="brand-title" style="display: none">Zeppelin</h1>
         "HTML-CSS": { availableFonts: ["TeX"] },
         messageStyle: "none"
       }
+      const origin = window.location.origin;
```
Reamer, line 46: Why is this needed?

zeppelin-web-angular/src/index.html
```diff
@@ -17,7 +17,6 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" href="./assets/images/zeppelin.png" type="image/x-icon">
   <title>Zeppelin</title>
-  <base href="/">
```
Reamer, line 20: You take this out in the new UI, which I understand. Where is the place in the old UI?

zeppelin-web-angular/src/index.html
tbonelee: MathJax resolves its dependency URLs based on its configuration.
Previously, it was loading its dependency files from the root URL (the default UI app URL), regardless of whether it was being served from the default UI or the `/next` UI.
Since both apps included MathJax as a dependency, this didn't cause any issues, but I felt it was more appropriate for MathJax to load its dependencies from the correct app-specific paths.

So, I updated th MathJax configurations to ensure it resolves URLs relative to the correct app.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
tbonelee: If neither new nor classic is provided as a configuration value, the valueOf() method will throw an IllegalArgumentException during server startup.
I based this behavior on the getRunMode() method, which similarly doesn't handle exceptions for invalid input.

Should we add more graceful error handling here?

zeppelin-web-angular/src/index.html
tbonelee: Both the old and new UIs are served from their respective sub-paths (depending on the configuration, either `/`, `/new`, or `/classic`), and as I understand it, the trailing part of the URL after `#/...` is treated as an anchor and ignored.
Therefore, without the base tag, static resources should be correctly fetched from the appropriate web app serving path, since the `src` attributes for those resources are set using relative paths.
I verified this in the browser, and everything worked as expected.

Given this, I believe neither app requires a base tag. In fact the old UI didn't have one to begin with.

I understand your question as asking where the base tag is in the old UI, so I've responded accordingly.
However, if I misunderstood your point, please feel free to clarify.

zeppelin-interpreter/src/main/java/org/apache/zeppelin/conf/ZeppelinConfiguration.java
tbonelee: @jongyoul If the value is empty, `getString()` will return default value (`new`) and if an invalid value (neither `new` nor `classic`) is provided, it will throw an exception.

Do you think it is not appropriate to provide a default value for an empty value in this configuration?

Reamer: In my opinion, the exception and the termination of the start in the event of incorrect configuration is acceptable.

Reamer (approved):
I looked at the code and it might work. Tests would of course be nice, but are difficult to implement at such a high level to test this small feature.
I have not tested manually myself and take your word for it.

PR #4826, opened 2024-09-13
[ZEPPELIN-6088] Fix some flaky tests in zeppelin-integration

tbonelee:
### What is this PR for?

It seems that some integration tests are randomly failing due to timing discrepancies between rendering and verification.

To ensure that the values are validated after rendering is fully completed, I added a slight sleep time.

Before the fix, 7 out of 20 runs failed in my personal repository's Github Actions due to one or both the two tests.
However, after the fix, there were no failures from those two tests in 30 runs.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/6088

- CI

### Questions:


pan3793 (approved):

pan3793:
Thanks, merged to master

PR #4832, opened 2024-09-15
[ZEPPELIN-6072] Unify tooltips across action bar buttons, and add tooltips to paragraph controller

cherrie-k:
### What is this PR for?
I noticed that only some buttons in the first group of the action bar (highlighted in the red box in the image below) had tooltips, while others did not. To improve consistency and enhance user experience, I ensured that all buttons in the first group of the action bar display tooltips on hover.
image
I also considered adding tooltips to the other groups of the action bar (outside the red box), but due to the complexity of implementing this with the ng-zorro UI library, more time and investigation are needed. I aim to address this in a future update.

In addition, the paragraph controller currently lacks tooltips on its icons, which may be confusing for new users unfamiliar with Zeppelin (like myself). Without tooltips, it's unclear what each button does, leading to potential usability issues. I added tooltips to all icons in the paragraph controller to improve clarity and guide users of the functionality of each button.
image

These changes aim to enhance the overall usability of the interface, making it more intuitive and consistent for all users.

**Before**
Some buttons show no tooltips.
before

**After**
Missing tooltips are now displayed.
after

### What type of PR is it?
Improvement

### Todos
* [x] - Add tooltips to action bar
* [x] - Add tooltips to paragraph controller

### What is the Jira issue?
[ZEPPELIN-6072](https://issues.apache.org/jira/browse/ZEPPELIN-6072)

* Build and test - check if tooltips are visible

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
@@ -47,7 +47,7 @@
     <i nz-icon [nzType]="tableHide ? 'book' : 'read'" nzTheme="outline"></i>
   </a>
   <a nz-dropdown [nzDropdownMenu]="menu" [nzTrigger]="'click'" [nzClickHide]="false" [(nzVisible)]="dropdownVisible">
-    <span nz-tooltip nzTitle="More">
+    <span nz-tooltip nzTooltipTitle="More">
       <i nz-icon nzType="setting" nzTheme="outline"></i>
     </span>
   </a>
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
@@ -28,22 +28,28 @@
        (click)="runParagraph.emit()"
        nzType="play-circle"
        class="run-para"
-       nzTheme="outline"></i>
+       nzTheme="outline"
+       nz-tooltip
+       nzTooltipTitle="Run paragraph"></i>
     <i nz-icon
        *ngIf="status=='RUNNING' || status=='PENDING'"
        (click)="cancelParagraph.emit()"
        class="cancel-para"
        nzType="pause-circle"
-       nzTheme="outline"></i>
+       nzTheme="outline"
+       nz-tooltip
+       nzTooltipTitle="Cancel paragraph"></i>
   </a>
-  <a (click)="toggleEditor()">
+  <a (click)="toggleEditor()" nz-tooltip nzTooltipTitle="Show/hide the code">
     <i nz-icon [nzType]="editorHide ? 'fullscreen' : 'fullscreen-exit'" nzTheme="outline"></i>
   </a>
-  <a (click)="toggleOutput()">
+  <a (click)="toggleOutput()" nz-tooltip nzTooltipTitle="Show/hide the output">
     <i nz-icon [nzType]="tableHide ? 'book' : 'read'" nzTheme="outline"></i>
   </a>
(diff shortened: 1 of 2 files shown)


zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
```diff
@@ -28,22 +28,28 @@
...
+  <a (click)="toggleEditor()" nz-tooltip nzTooltipTitle="Show/hide the code">
     <i nz-icon [nzType]="editorHide ? 'fullscreen' : 'fullscreen-exit'" nzTheme="outline"></i>
   </a>
-  <a (click)="toggleOutput()">
+  <a (click)="toggleOutput()" nz-tooltip nzTooltipTitle="Show/hide the output">
     <i nz-icon [nzType]="tableHide ? 'book' : 'read'" nzTheme="outline"></i>
   </a>
   <a nz-dropdown [nzDropdownMenu]="menu" [nzTrigger]="'click'" [nzClickHide]="false" [(nzVisible)]="dropdownVisible">
-    <i nz-icon nzType="setting" nzTheme="outline"></i>
+    <span nz-tooltip nzTitle="More">
```
tbonelee, line 50: Wouldn't it be better to change `nzTitle` to `nzTooltipTitle` for consistency with the code changes above and to clarify the meaning of the attribute?

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
```diff
@@ -28,22 +28,28 @@
...
        (click)="cancelParagraph.emit()"
        class="cancel-para"
        nzType="pause-circle"
-       nzTheme="outline"></i>
+       nzTheme="outline"
+       nz-tooltip
+       nzTooltipTitle="Cancel paragraph"></i>
   </a>
-  <a (click)="toggleEditor()">
+  <a (click)="toggleEditor()" nz-tooltip nzTooltipTitle="Show/hide the code">
```
tbonelee, line 43: It seems that this tooltip is covering the buttons located to the left of its target, making it difficult to click those buttons once the tooltip appears.
It might be better to fix the tooltip's position, perhaps above the target.
I think you could use a property like `nzTooltipPlacement`.

tbonelee (changes requested):

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.html
pan3793: the current behavior is ok to me given the tooltip disappears immediately after moving the mouse away

cherrie-k: I tried using nzTooltipPlacement="top", but it didn’t work as expected, likely due to limited space around the target element, causing the tooltip to shift to the left.

I also looked into other features like nzTooltipArrowPointAtCenter and nzTooltipOverlayStyle, but those aren't available in our current Ant Design version due to the Angular version we’re using.

As @pan3793 mentioned, I think the current behavior is acceptable. Alternatively, nzTooltipPlacement="topRight" places the tooltip above the button as in the image below, but it’s not centered, so I’d appreciate more feedback on this.
image

Open to other suggestions. Thank you a lot @tbonelee

tbonelee: @cherrie-k 
If the option doesn't work as expected, it's not a big deal.
The issue I mentioned was more of a secondary concern, so I'm okay with merging things as they are.
Thanks a lot for exploring alternative solutions anyway!

pan3793 (approved):

pan3793:
Thanks, merged to master/0.12

PR #4831, opened 2024-09-13
[ZEPPELIN-6089][INFRA] Improve the merge PR script

pan3793:
### What is this PR for?

Zeppelin has `dev/merge_zeppelin_pr.py` that was borrowed from Spark, I would recommend committers use this script over the GitHub button to merge PR, which has some benefits:

1. Simplify the backport process

the tools will ask you to backport the commit to lower maintained branches after you merge a PR to master, if there are no conflicts, all things you need to do are just type a "branch name" that you want to backport.

2. Automatically update JIRA information

the script uses the python jira client to update JIRA ticket, for example, automatically closes the JIRA ticket after PR is merged, fills in the fixed versions, which is important to users to know the features/bug fixes applied to versions.

3. Better PR title, body, and "Signed-off-by" info

Before
image

After 
image

This PR syncs the change from the Spark upstream (around 4.0.0-preview2), and has several improvements recently, e.g. support using tokens instead of passwords for GitHub and JIRA authentication. Additionally, this PR switches to the GitHub open API @jongyoul suggested to merge the PR, which fixed the merged PR status from "Closed" to "Merged"

### What type of PR is it?

Improvement

### Todos

* [x] - verify this script by merging at least 3 PRs

### What is the Jira issue?

ZEPPELIN-6089

### How should this be tested?

Manually test. Currently not work due to permission issues.
```
$ dev/merge_zeppelin_pr.py 
git rev-parse --abbrev-ref HEAD
Which pull request would you like to merge? (e.g. 34): 4837

=== Pull Request #4837 ===
title   [MINOR] Remove duplicate entry in .gitignore
source  MyLanPangzi/patch-2
target  master
url     https://api.github.com/repos/apache/zeppelin/pulls/4837
Proceed with merging pull request #4837? (y/N): y
git config --get user.name
git config --get user.email
git fetch apache master
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 3 (delta 2), reused 0 (delta 0), pack-reused 0 (from 0)
Unpacking objects: 100% (3/3), 1.53 KiB | 260.00 KiB/s, done.
From github.com:apache/zeppelin
 * branch                master     -> FETCH_HEAD
   ad79848a9..35e129912  master     -> apache/master
Pull request #4837 merged!
Merge hash: 35e12991

Would you like to pick 35e12991 into another branch? (y/N): y
[...]

PR #4452, opened 2022-08-25
[MINOR] Add some GitHub features for PR

jongyoul:
### What is this PR for?
This provide dismiss feature when changing code after getting approval.

### What type of PR is it?
Improvement

### Questions:
* Does the licenses' files need to update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #4833, opened 2024-09-17
Revert "[MINOR] Add some GitHub features for PR (#4452)"

pan3793:
This reverts commit d045c3b2d8c6892128d25ec164598b0c0317e2e5.

### What is this PR for?

As discussed in https://github.com/apache/zeppelin/pull/4831, we need to revert d045c3b2d8c6892128d25ec164598b0c0317e2e5 to make the `dev/merge_zeppelin_pr.py` work. 

### What type of PR is it?

Bug Fix

### What is the Jira issue?

N/A

This is a dev only change, I will verify `dev/merge_zeppelin_pr.py` after merging this.

### Questions:

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -34,11 +34,6 @@ github:
     merge: false
     squash: true
     rebase: false
-  protected_branches:
-    master:
-      required_pull_request_reviews:
-        dismiss_stale_reviews: true
-        required_approving_review_count: 1
   autolink_jira:
     - ZEPPELIN
 


pan3793:
@jongyoul I'm not against exploring other approaches for better committer experience, but can we get this in and recover the `dev/merge_zeppelin_pr.py` first?

pan3793:
we don't need to revert this change after switching to GitHub open API to merge PR, see https://github.com/apache/zeppelin/pull/4831

PR #4835, opened 2024-09-18
[ZEPPELIN-6094] Fix broken impersonation

ruslanfialkovskii:
### What is this PR for?

The issue stems from the way the interpreter.sh script handles the ZEPPELIN_IMPERSONATE_CMD and constructs the command to be executed. Specifically, the command string passed to bash -c is not being properly formed as a single string, causing exec to fail when it tries to execute the impersonation command.

1. Command Construction: When ZEPPELIN_IMPERSONATE_CMD is set, the script attempts to build an array INTERPRETER_RUN_COMMAND that includes this command. However, if ZEPPELIN_IMPERSONATE_CMD is a string, it gets added as a single element to the array, leading to incorrect command parsing.
2. bash -c Behavior: The -c option in bash expects a single string argument that contains the command to execute. If this command is split into multiple array elements, bash misinterprets it, causing failures when it encounters shell built-ins like source.
3. exec Execution: The exec command replaces the shell with the specified command. If the command is not correctly formatted, exec cannot find it, resulting in errors like exec: sudo -H -u rfialkovskii bash -c: not found.

To fix the issue needs to adjust how the command is constructed in interpreter.sh when ZEPPELIN_IMPERSONATE_CMD is used. Specifically, the command string passed to bash -c is a single, properly formatted string.

Needs:
Define ZEPPELIN_IMPERSONATE_CMD as an Array.
Modify interpreter.sh to Build a Single Command String:

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN/
(https://issues.apache.org/jira/browse/ZEPPELIN-6094)

Enable shiro auth, configure impersonation and try to run any interpreter under impersonated user

Screenshot 2024-09-19 at 01 46 38

### Questions:
No
Yes, needs to change env from:
export ZEPPELIN_IMPERSONATE_CMD='sudo -H -u "${ZEPPELIN_IMPERSONATE_USER}" bash -c'
to:
export ZEPPELIN_IMPERSONATE_CMD=(sudo -H -u "${ZEPPELIN_IMPERSONATE_USER}" bash -c)

Yes

### DIFF
--- a/bin/interpreter.sh
+++ b/bin/interpreter.sh
@@ -247,13 +247,19 @@ addJarInDirForIntp "${LOCAL_INTERPRETER_REPO}"
 
 if [[ -n "$ZEPPELIN_IMPERSONATE_USER" ]]; then
   if [[ "${INTERPRETER_ID}" != "spark" || "$ZEPPELIN_IMPERSONATE_SPARK_PROXY_USER" == "false" ]]; then
-    suid="$(id -u "${ZEPPELIN_IMPERSONATE_USER}")"
-    if [[ -n  "${suid}" || -z "${SPARK_SUBMIT}" ]]; then
-       INTERPRETER_RUN_COMMAND+=("${ZEPPELIN_IMPERSONATE_RUN_CMD[@]}")
-       if [[ -f "${ZEPPELIN_CONF_DIR}/zeppelin-env.sh" ]]; then
-           INTERPRETER_RUN_COMMAND+=("source" "${ZEPPELIN_CONF_DIR}/zeppelin-env.sh;")
-       fi
+    # Build the command string
+    COMMAND_STRING=""
+    if [[ -f "${ZEPPELIN_CONF_DIR}/zeppelin-env.sh" ]]; then
+        COMMAND_STRING+="source ${ZEPPELIN_CONF_DIR}/zeppelin-env.sh; "
     fi
+    
+    # Add interpreter command to the command string
+    IFS=' ' read -r -a JAVA_INTP_OPTS_ARRAY <<< "${JAVA_INTP_OPTS}"
+    IFS=' ' read -r -a ZEPPELIN_INTP_MEM_ARRAY <<< "${ZEPPELIN_INTP_MEM}"
+    COMMAND_STRING+="${ZEPPELIN_RUNNER} ${JAVA_INTP_OPTS_ARRAY[@]} ${ZEPPELIN_INTP_MEM_ARRAY[@]} -cp '${ZEPPELIN_INTP_CLASSPATH_OVERRIDES}:${ZEPPELIN_INTP_CLASSPATH}' ${ZEPPELIN_SERVER} ${CALLBACK_HOST} ${PORT} ${INTP_GROUP_ID} ${INTP_PORT}"
+    
+    # Set INTERPRETER_RUN_COMMAND with the impersonation command and command string
+    INTERPRETER_RUN_COMMAND=("${ZEPPELIN_IMPERSONATE_CMD[@]}" "${COMMAND_STRING}")
   fi
 fi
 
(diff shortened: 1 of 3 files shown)


pan3793:
will it create subprocesses when impersonation is enabled? if yes, what's the hierarchy?

ruslanfialkovskii:
@pan3793 Yes, when impersonation is enabled in Apache Zeppelin, it creates subprocesses to execute the interpreter under the impersonated user’s identity. 
Zeppelin Server Process (Parent)
    └── interpreter.sh (Interpreter Launch Script)
        └── sudo (or ssh) as ZEPPELIN_IMPERSONATE_CMD (Switches User Context)
            └── Interpreter Process (e.g., Spark, Flink, etc.)
                └── [Optional] Interpreter Subprocesses (e.g., Spark driver, executors)
                
Nothing new, this is a fix for what worked before.
It doesn't interfere into process work, it just fixed the command format.

pan3793:
I think ZEPPELIN-5070 changed the process's hierarchy, at least for impersonation-disabled cases.

ruslanfialkovskii:
Also, this fix enables impersonation permanently if ZEPPELIN_IMPERSONATE_CMD exists in env file. I appreciate to point on the clue on how to enable it only if the option "Set impersonation" is active.

pan3793:
@ruslanfialkovskii I tend to accept this change as it recovers a broken feature, but the differences in the process's hierarchy between impersonate enabled and disabled might break some assumptions.

pan3793:
cc @Reamer, you may have the context to evaluate the impact of different process's hierarchy

Reamer:
In general, I am not a friend of user impersonation. In my opinion, both the attempt with `sudo` and `ssh` are hacked solutions.
In general, there are such approaches that the running program changes the user. This can be seen, for example, with the Apache web server, which changes from `root` to `www-data`. As far as I know, this is done because Apache opens privileged ports (<1024). In the case of Apache, this is usually 80 or 443.
Zeppelin is already running on a high port e.g. 8080, so it is not necessary to change the user here, but you can start directly with an unprivileged user.

Both ssh and sudo may have problems passing signals to the JVM. This can lead to the Zeppelin interpreter not shutting down properly.

I had the signal problem in K8s, which is why I switched to `exec`.

Reamer:
If the user impersonation works again with these corrections, that's fine by me.

pan3793:
If I understand correctly, after ZEPPELIN-5070, with impersonation disabled, the process's hierarchy looks like below?
```
Zeppelin Server Process (Parent)
    └── Interpreter Process (e.g., Spark, Flink, etc.)
        └── [Optional] Interpreter Subprocesses (e.g., Spark driver, executors)
```

ruslanfialkovskii:
Hi @Reamer, the point of using impersonation is running interpreters under specific authenticated users (in my case by ldap)
It is a security team request. Zeppelin server of course is already running under Zeppelin system user (zeppelin) but I need to configure it to run a specific interpreter under a specific user in a separate isolated process. 
Screenshot 2024-09-19 at 13 44 30
because running interpreter under system user gives too wide permissions in my case.

pan3793:
Hmm... I believe I understand why [gosu](https://github.com/tianon/gosu) present and is adopted widely now, maybe we can explore leverage it to avoid creating subporcess, though this might be out of the scope of this PR.

pan3793 (approved):

ruslanfialkovskii:
@Reamer yes, I set up  a user for each LDAP user on the system, which is then addressed via sudo.

pan3793:
@ruslanfialkovskii thanks, merged to master for 0.12. Given it involves a breaking change - user should update the default value, I only merge this to the master for the next feature version

PR #4838, opened 2024-09-19
[ZEPPELIN-6095] validate decoded url in jdbc interpreter

s2moon98:
### What is this PR for?

Add some validation check conditions to existing url validator in jdbc interpreter. So now it can check URLs with the conditions below if it has an unallowable configuration.
- UTF-8 encoded

### What type of PR is it?

Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

[ZEPPELIN-6095](https://issues.apache.org/jira/browse/ZEPPELIN-6095)

Input the url with unallowable configurations in UTF-8 encoded in JDBC type interpreter.  Then run the command in notebook and see if the command is blocked from running.

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -591,13 +591,9 @@ public class JDBCInterpreter extends KerberosInterpreter {
 
   private void validateConnectionUrl(String url) {
     String decodedUrl;
-    try {
-      decodedUrl = URLDecoder.decode(url, StandardCharsets.UTF_8.toString());
-    } catch (UnsupportedEncodingException e) {
-      throw new IllegalArgumentException("Connection URL decode failed");
-    }
+    decodedUrl = URLDecoder.decode(url, StandardCharsets.UTF_8);
 
-    if (containsIgnoreCase(decodedUrl, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
+      if (containsIgnoreCase(decodedUrl, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
             containsIgnoreCase(decodedUrl, AUTO_DESERIALIZE) ||
             containsIgnoreCase(decodedUrl, ALLOW_LOCAL_IN_FILE_NAME) ||
             containsIgnoreCase(decodedUrl, ALLOW_URL_IN_LOCAL_IN_FILE_NAME)) {
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -42,6 +42,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
 import java.security.PrivilegedExceptionAction;
 import java.sql.Connection;
 import java.sql.DriverManager;
@@ -587,10 +589,18 @@ public class JDBCInterpreter extends KerberosInterpreter {
   }
 
   private void validateConnectionUrl(String url) {
-    if (containsIgnoreCase(url, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, AUTO_DESERIALIZE) ||
-        containsIgnoreCase(url, ALLOW_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, ALLOW_URL_IN_LOCAL_IN_FILE_NAME)) {
+    String decodedUrl;
+    try {
+      decodedUrl = URLDecoder.decode(url, "UTF-8").replaceAll("\\s", "");
+    } catch (UnsupportedEncodingException e) {
+      LOGGER.info("fail to decode url : {}. continue to get connection.", url, e);
+      return;
+    }
+
+    if (containsIgnoreCase(decodedUrl, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
+            containsIgnoreCase(decodedUrl, AUTO_DESERIALIZE) ||
+            containsIgnoreCase(decodedUrl, ALLOW_LOCAL_IN_FILE_NAME) ||
+            containsIgnoreCase(decodedUrl, ALLOW_URL_IN_LOCAL_IN_FILE_NAME)) {
       throw new IllegalArgumentException("Connection URL contains sensitive configuration");
     }
   }
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -591,7 +591,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private void validateConnectionUrl(String url) {
     String decodedUrl;
     try {
-      decodedUrl = URLDecoder.decode(url, "UTF-8").replaceAll("\\s", "");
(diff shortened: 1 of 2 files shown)


tbonelee (commented):
The code looks good and works as intended!
I’ve checked that URLs with unallowable configurations, like being URL-encoded or containing white spaces, are properly blocked.
That said, I noticed that the leading and trailing white spaces are already being trimmed when reading the property. So I’m curious—do we really need to trim white spaces again here? Do we actually need to handle white spaces in the middle of URLs at this point?

pan3793:
> do we really need to remove white spaces here?

I have the same question too. Is there any potential risk if we have encoded while spaces in the JDBC URL?

In [Apache Kyuubi cases](https://kyuubi.readthedocs.io/en/master/client/jdbc/kyuubi_jdbc.html), it allows the user to pass the `spark.app.name` as part of JDBC URL, so the encoded URL might be like
```
jdbc:kyuubi://host:port/db;#spark.app.name=Hello%20World;spark.driver.memory=4g
```

For example,
```
$  bin/beeline -u 'jdbc:hive2://localhost:10009/#spark.app.name=Hello%20World'
```

```
0: jdbc:hive2://localhost:10009/#spark.app.na> set spark.app.name;

24/09/20 00:30:25 INFO ExecuteStatement:
           Spark application name: Hello World
                 application ID: local-1726763393107
                 application web UI: http://192.168.8.150:53761
                 master: local[*]
                 deploy mode: client
                 version: 3.3.1
           Start time: 2024-09-20T00:29:52.154
           User: anonymous
...
2024-09-20 00:30:26.123 INFO org.apache.kyuubi.operation.ExecuteStatement: Processing anonymous's query[a8771241-5233-44fe-8c69-ad94a8f0217d]: RUNNING_STATE -> FINISHED_STATE, time taken: 0.128 seconds
+-----------------+--------------+
|       key       |    value     |
+-----------------+--------------+
| spark.app.name  | Hello World  |
+-----------------+--------------+
1 row selected (0.401 seconds)
```

s2moon98:
@tbonelee @pan3793 
Thank you for mentioning the white space issue. I searched for some cases that need white space in JDBC url, and realized that the cases exist. (such as Apache kyuubi cases you mentioned or database names, etc)

So, I will modify the code to validate the url with UTF-8 deocded and not eliminating the white spaces.

s2moon98:
@tbonelee @pan3793 
Thank you for mentioning the white space issue. 

I put the eliminating white spaces logic to block the unallowable configurations with white space included such as :
`allow%20LoadLocalInfile=true`

The URL with eliminating white spaces is used only when validating, not when creating connection with the URL. So I think it does not cause [risky situation](https://github.com/apache/zeppelin/pull/4838#issuecomment-2361501953) like a URL containing white space is not connected.

However I confirmed that if user put url with white space contained, the configuration does not apply and [JDBC throws an error.](https://github.com/apache/zeppelin/pull/4838#pullrequestreview-2315892923) So I will remove the eliminating white spaces logic.

tbonelee (approved):

tbonelee (approved):

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -587,10 +589,18 @@ public Connection getConnection(InterpreterContext context)
   }
 
   private void validateConnectionUrl(String url) {
-    if (containsIgnoreCase(url, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, AUTO_DESERIALIZE) ||
-        containsIgnoreCase(url, ALLOW_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, ALLOW_URL_IN_LOCAL_IN_FILE_NAME)) {
+    String decodedUrl;
+    try {
+      decodedUrl = URLDecoder.decode(url, "UTF-8");
```
Reamer, line 594: If we wait for #4882 we can use `decode(String s, Charset charset)`
[URLDecoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html)

s2moon98: oh, I didn't noticed that `decode(String s, Charser charset)` doesn't exist in JAVA8. Thank you for letting me know.
Then, if https://github.com/apache/zeppelin/pull/4882 merged into master, I will request approval again

s2moon98:
https://github.com/apache/zeppelin/pull/4882 (Java release to 11) is merged to master branch, 
so can @pan3793 review this pr, so I can merge this branch?

pan3793:
some thoughts:
- PR title and desc are misleading, the key change of this PR is, validating the decoded URL instead of the encoded URL
- if the decoded URL is an illegal UTF-8 sequence, the new code just skips validation, I'm not sure this is the correct behavior. from the security perspective, we should fail immediately.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
Reamer: #4882 was merged into master. Please use [StandardCharsets](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html) as second input Parameter.

Reamer (changes requested):

s2moon98:
- changed the PR title to show the key change.
- changed code to fail immediately if url decoding throws error. Previously, I thought if decoding step fails, validating logic has a fault. However as @pan3793 says above, fail immediately is more close behavior to secure, I changed it.
- use StandardCharsets as second input parameter as @Reamer says above

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -587,10 +590,17 @@ public Connection getConnection(InterpreterContext context)
   }
 
   private void validateConnectionUrl(String url) {
-    if (containsIgnoreCase(url, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, AUTO_DESERIALIZE) ||
-        containsIgnoreCase(url, ALLOW_LOCAL_IN_FILE_NAME) ||
-        containsIgnoreCase(url, ALLOW_URL_IN_LOCAL_IN_FILE_NAME)) {
+    String decodedUrl;
+    try {
+      decodedUrl = URLDecoder.decode(url, StandardCharsets.UTF_8.toString());
```
Reamer, line 595:
```suggestion
      decodedUrl = URLDecoder.decode(url, StandardCharsets.UTF_8);
```
should be sufficient. Check https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html

Reamer (changes requested):
Minor adjustment necessary.

pan3793:
LGTM, I have no further concerns.

s2moon98:
Thanks for @Reamer, I fix the `URLDecoder.decode` method to fit for java 11. 
Additionally, in[ java 11 URLDecoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html), it never throws [UnsupportedEncodingException](https://docs.oracle.com/javase/8/docs/api/java/io/UnsupportedEncodingException.html) so I removed the relevant try-catch statement.

Reamer (changes requested):
I have checked out your pull request, please perform a git rebase on the current master so that your pull request benefits from the JDK 11 transition.

s2moon98:
@Reamer 
I rebased my pr after the master branch, and applied your reviews ([fix indent & remove unused import](https://github.com/apache/zeppelin/pull/4838/commits/b22eafe7fe56b806fe53e64b6a3ab374d803504a)). 
Please check for it!

Reamer (approved):
LGTM

Reamer:

PR #4844, opened 2024-09-28
[ZEPPELIN-6104] Fix NPE when some MongoDB Interpreter configs are missing

seung-00:
### What is this PR for?
This PR avoids an NPE when some MongoDB Interpreter configs are missing by replacing them with empty strings.

If certain MongoDB Interpreter configs, such as the username, password, or authenticationDatabase, are not entered, it throws an NPE. MongoDB allows logins without authentication when authentication is not enabled, so these configs are not always required. Since authentication is off by default in MongoDB, and these configs are also empty by default in Apache Zeppelin, this could be confusing for users.
And as you can see in [shell_extension.js](https://github.com/apache/zeppelin/blob/ac48cc2db1b55257230cae2c382961da2a9cf80d/mongodb/src/main/resources/shell_extension.js#L115), the MongoDB interpreter appears to be designed to skip authorization when the username is null.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-6104]

* run the MongoDB notebook

### Questions:


pan3793 (approved):

pan3793:
Thanks, merged to master/0.12

PR #4848, opened 2024-09-29
[ZEPPELIN-6110] Specify the Node.js engine version below 17 in the `web-angular`

seung-00:
### What is this PR for?
Specify the Node.js engine version below 17 in the `zeppelin-web-angular` package.json file to avoid errors related to Node.js 17+ and OpenSSL in Webpack.

If we upgrade the Angular version to 13+ in the future, we will be able to use Node.js 17+ as well.

[Reference: nodejs 17: digital envelope routines::unsupported · Issue #14532 · webpack/webpack](https://github.com/webpack/webpack/issues/14532)

### What type of PR is it?
Improvement

### What is the Jira issue?
* [ZEPPELIN-6110]

1. Set your Node.js version to 17+
2. Execute `npm run postinstall` in `zeppelin-web-angular`

### Questions:

### DIFF
--- a/zeppelin-web-angular/package.json
+++ b/zeppelin-web-angular/package.json
@@ -15,6 +15,9 @@
     "lint": "ng lint",
     "e2e": "ng e2e"
   },
+  "engines": {
+    "node": "<17.0.0"
+  },
   "private": true,
   "dependencies": {
     "@angular/animations": "~8.2.10",


pan3793:
it's a pity that the new UI also uses an ancient angular version, it would be nice if someone can upgrade it to a modern version

seung-00:
@pan3793 
You're right. That said, I think specifying the Node.js engine is needed, not just to avoid errors but also to reduce confusion.

pan3793 (approved):

pan3793:
Merged to master/0.12

PR #4865, opened 2024-10-05
[ZEPPELIN-6121] Write a Dockerfile for python interpreter image build

seung-00:
### What is this PR for?
This PR adds a Dockerfile to build a Python interpreter container. I think it could be helpful to use distributed computing resources.

### What type of PR is it?
Improvement

### What is the Jira issue?
* [ZEPPELIN-6121](https://issues.apache.org/jira/browse/ZEPPELIN-6121)

* Build and run the docker container
* Verify that the interpreter container communicates with the zeppelin server

### Questions:
* Does the license files need to update? Y

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -28,6 +28,11 @@ RUN ./mvnw clean package -am -pl zeppelin-interpreter-shaded,zeppelin-interprete
 
 FROM openjdk:11
 
+RUN apt-get update && \
+    apt-get install -y python3 python3-pip && \
+    ln -s /usr/bin/python3 /usr/bin/python && \
+    rm -rf /var/lib/apt/lists/*
+
 COPY --from=builder /zeppelin/bin /zeppelin/bin/
 COPY --from=builder /zeppelin/conf /zeppelin/conf
 
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -30,6 +30,7 @@ FROM openjdk:11
 
 RUN apt-get update && \
     apt-get install -y python3 python3-pip && \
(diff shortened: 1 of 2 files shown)


python/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+FROM openjdk:11 as builder
+
+COPY . /zeppelin/
+
+WORKDIR /zeppelin
+
+RUN chmod +x ./mvnw
+
+RUN ./mvnw clean package -am -pl zeppelin-interpreter-shaded,zeppelin-interpreter,python -DskipTests
+
```
tbonelee, line 27: I was wondering if it might be better to ensure that `%python`, `%python.ipython`, and `%python.sql` are all supported by default.
My thinking comes from the fact that these interpreters are listed in the overview section of the Zeppelin Python interpreter documentation, and IPython is also recommended there.

For IPython, it seems that adding the following command here can be useful for installing the necessary packages:
```suggestion
RUN pip install jupyter-client grpcio protobuf~=3.20 ipython ipykernel
```

seung-00: @tbonelee 
Thank you for your comment. I have applied your suggestion and agree with you. That said, there are additional libraries, like pandas, that wouldn't be installed. In my opinion, we need an alternative solution to inject these libraries without having to rewrite/deploy the Dockerfile. I think it'd be better to address this in a separate task.

tbonelee: I agree with your point.
I believe that handling additional packages, like pandas, would require a more flexible approach. Such as using a conda-like package management system. It might be better to address this in a separate PR later.

tbonelee (approved):

PR #4879, opened 2024-10-18
[ZEPPELIN-6132] Bump Maven 3.9.9 and Maven Wrapper 3.3.2

pan3793:
### What is this PR for?

Install the latest Maven 3.9.9 locally, and run `mvn wrapper:wrapper -Dtype=source`

https://maven.apache.org/wrapper/index.html#usage-with-or-without-binary-jar

Note: Apache project does not recommend shipping binary files in source release, so this PR replaces `.mvn/wrapper/maven-wrapper.jar` with `.mvn/wrapper/MavenWrapperDownloader.java`

### What type of PR is it?

Improvement

### What is the Jira issue?

ZEPPELIN-6132

Pass GHA.

### Questions:


Reamer (approved):
I really like the fact that you are removing the binary from the SCM. Java as the only requirement is great.

Many thanks for the change.

PR #4881, opened 2024-10-21
[ZEPPELIN-6135] Fix clean goal not removing interpreter/ output(Spark,Flink)

tbonelee:
### What is this PR for?
The Maven clean goal does not remove some interpreter files in `interpreter/` directory as expected.
Specifically, the `spark` and `flink` interpreters are not cleaned.

This happens because when extending the `pom.xml` in the `pom.xml` of these interpreters, they lack the `interpreter.name` property.
As a result, the clean target directory configured in `zeppelion-interpreter-parent` was not properly resolved before.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-6135

- Build Zeppelin
- Clean Zeppelin and check if all interpreters are removed in `interpreter/`

### Questions:

### DIFF
--- a/flink/flink-scala-2.12/pom.xml
+++ b/flink/flink-scala-2.12/pom.xml
@@ -33,7 +33,6 @@
 
   <properties>
     <!--library versions-->
-    <interpreter.name>flink</interpreter.name>
     <flink.version>${flink1.17.version}</flink.version>
     <flink.scala.version>2.12.7</flink.scala.version>
     <flink.scala.binary.version>2.12</flink.scala.binary.version>
--- a/flink/pom.xml
+++ b/flink/pom.xml
@@ -41,6 +41,8 @@
     </modules>
 
     <properties>
+        <interpreter.name>flink</interpreter.name>
+
         <flink1.15.version>1.15.1</flink1.15.version>
         <flink1.16.version>1.16.0</flink1.16.version>
         <flink1.17.version>1.17.1</flink1.17.version>
--- a/spark/interpreter/pom.xml
+++ b/spark/interpreter/pom.xml
@@ -265,18 +265,6 @@
     <plugins>
 
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-clean-plugin</artifactId>
-        <configuration>
-          <filesets>
-            <fileset>
-              <directory>${project.basedir}/../../interpreter/spark/</directory>
-            </fileset>
-          </filesets>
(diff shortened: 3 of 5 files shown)


tbonelee:
@Reamer 
Thanks for pointing that out. Just to confirm, you're saying that it makes sense for the `clean` goal to run in the module where the output is created, right? I updated it to run from `spark/interpreter` based on that.

Also, quick question, is it okay to delete the entire `interpreter/spark` directory, including the `spark-scala` interpreter JARs? I used a wildcard for simplicity, but if those files need to stay, I'll adjust it to only delete specific directories.

Reamer:
In general, this interpreter jar cleanup should already work.
The special clean goal in the interpreter parent should be triggered by the Flink submodule `flink-scala-2.12` and the Spark submodule `spark-interpreter`
You can find the cleanup goal here.
https://github.com/apache/zeppelin/blob/ea0802d24b5fd46a6113481cad3a5f53cecec879/zeppelin-interpreter-parent/pom.xml#L163-L174

Maybe it doesn't work because of the special submodule structure. I will check this out.

Reamer:
The path is not correct due to the submodule structure in Flink and Spark. Not to make it too complicated, I would be in favor of your first contribution suggestion. Put the interpreter name in the parent pom.xml of Spark and Flink. Theoretically you could then remove it in the corresponding submodules.
It is desirable that the complete folder is removed.

In general, I don't think it's good that all interpreters store files outside. But this can be adjusted in the future.

Thank you for finding the bug.

tbonelee:
@Reamer Thanks for checking things out!
I've reverted the second commit to go with the simpler solution you suggested. I also removed the unnecessary clean tasks in the Spark interpreter submodules.
Could you please check if these changes align with your suggestions?

Reamer:
Just two minor lines.
I think we can delete the following lines
https://github.com/apache/zeppelin/blob/3575a3cf8834cdc0100f0a554846829584df0c30/flink/flink-scala-2.12/pom.xml#L36
https://github.com/apache/zeppelin/blob/3575a3cf8834cdc0100f0a554846829584df0c30/spark/interpreter/pom.xml#L36

What do you think?

tbonelee:
@Reamer I agree with you; it's better to keep a single source of truth. I've updated these as you suggested.

Reamer (approved):
LGTM

PR #4891, opened 2024-11-03
[ZEPPELIN-6190] Prevent directory escape bypass through repeated URL decoding

tbonelee:
### What is this PR for?
This PR addresses an issue in `NotebookService` where the notebook path validation only performs a single decoding pass.
This allowed a malicious user to bypass validation by double-encoding the `".."` token.
By implementing the repeated decoding, we can prevent this bypass.
Additionally, to prevent excessive decoding attempts, a maximum limit on the number of decoding attempts has been added.

### What type of PR is it?
Hot Fix

### What is the Jira issue?
https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6190

* CI

### Questions:
  * There may be minor compatibility issues if a user relies on multiple encoded paths, but this is unlikely in realistic scenarios.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
@@ -24,6 +24,7 @@ import static org.apache.zeppelin.interpreter.InterpreterResult.Code.ERROR;
 import static org.apache.zeppelin.scheduler.Job.Status.ABORT;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
@@ -239,7 +240,7 @@ public class NotebookService {
 
     notePath = notePath.replace("\r", " ").replace("\n", " ");
 
-    notePath = URLDecoder.decode(notePath, StandardCharsets.UTF_8.toString());
+    notePath = decodeRepeatedly(notePath);
     if (notePath.endsWith("/")) {
       throw new IOException("Note name shouldn't end with '/'");
     }
@@ -1552,5 +1553,22 @@ public class NotebookService {
       callback.onFailure(new ForbiddenException(errorMsg), context);
       return false;
     }
+  }
+
+  private String decodeRepeatedly(final String encoded) throws IOException {
+    String previous = encoded;
+    int maxDecodeAttempts = 5;
+    int attempts = 0;
+
+    while (attempts < maxDecodeAttempts) {
+      String decoded = URLDecoder.decode(previous, StandardCharsets.UTF_8.toString());
+      attempts++;
+      if (decoded.equals(previous)) {
+        return decoded;
+      }
+      previous = decoded;
+    }
+
+    throw new IOException("Exceeded maximum decode attempts. Possible malicious input.");
   }
 }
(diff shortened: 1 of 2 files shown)


tbonelee:
cc @jongyoul

tbonelee:
@pan3793 @Reamer 
When you have a moment, could you check this PR? Appreciate it!

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -1553,4 +1554,21 @@ private <T> boolean checkPermission(String noteId,
...
     }
   }
+
+  private String decodeRepeatedly(final String encoded) throws IOException {
+    String previous = encoded;
+    int maxDecodeAttempts = 5;
+    int attempts = 0;
+
+    while (attempts < maxDecodeAttempts) {
+      String decoded = URLDecoder.decode(previous, StandardCharsets.UTF_8.toString());
```
Reamer, line 1564:
```suggestion
      String decoded = URLDecoder.decode(previous, StandardCharsets.UTF_8);
```
After https://github.com/apache/zeppelin/commit/414f4d1c470b5046628cfc378e6602fb78240e8f, we can now use JDK 11 functions.

zeppelin-server/src/main/java/org/apache/zeppelin/service/NotebookService.java
```diff
@@ -1553,4 +1554,21 @@ private <T> boolean checkPermission(String noteId,
       return false;
     }
   }
+
+  private String decodeRepeatedly(final String encoded) throws IOException {
```
Reamer, line 1558: This method can be static.
```suggestion
  private static String decodeRepeatedly(final String encoded) throws IOException {
```

Reamer (changes requested):
Just a few minor comments. The five attempts seem quite random. But I think it's good that an error message is thrown afterwards.

tbonelee:
@Reamer Thank you for the feedback! I've added a commit based on your feedback.

I agree that using five attempts feels somewhat arbitrary. However I couldn't think of a better alternative, so I went with this value.
If you have any other suggestions, I'd love to hear them!

Reamer (approved):
LGTM, let's wait for CI

pan3793:
I read the change and understand what it does, but I don't understand why should this change.
And please create a JIRA ticket for it, if you think this involves CVE which is not suitable to discuss publicly, you can write a detailed message to `private@zeppelin.apache.org`

Reamer:
I agree with @pan3793 . Please create a Jira ticket and adjust the pull request description accordingly.

tbonelee:
For context: this PR was based on a security issue I received from @jongyoul, but I didn't have a Jira ticket at the time, which may have caused some confusion.

Apologies for the earlier lack of clarity - I've now created [ZEPPELIN-6190](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6190) and linked this PR to it.

jongyoul:
@tbonelee Thank you for sharing it. 

@pan3793 @Reamer I missed sharing context with others as well. Sorry. This change relates to the security issue, so I thought it would be good to omit creating a ticket. Feel free to reach me if something is not clear. 🙏

Reamer:

PR #4893, opened 2024-11-03
[ZEPPELIN-6143] Add Interpreter Event Server Port configuration option

tbonelee:
### What is this PR for?
This PR introduces the `ZEPPELIN_EVENT_SERVER_PORT` configuration option.
When this option is set, the Event server that communicates with interpreters will listen on the specified port.

The `ZEPPELIN_EVENT_SERVER_PORT` option takes precedence over `ZEPPELIN_SERVER_RPC_PORTRANGE` option.
If `ZEPPELIN_EVENT_SERVER_PORT` is set, the `ZEPPELIN_SERVER_RPC_PORTRANGE` range will be ignored.

This option is particularly useful when running individual Zeppelin components in container orchestration environments like Kubernetes.
By explicitly specifying the port, it allows for a more declarative and clearer configuration of service resources compared to `ZEPPELIN_SERVER_RPC_PORTRANGE`.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-6143

* Run Zeppelin with and without this configuration and check zeppelin-server logs for the message `"InterpreterEventServer is starting at "`. Verify that the specified port is used as intended(If the option is set).

### Questions:

### DIFF
--- a/conf/zeppelin-env.sh.template
+++ b/conf/zeppelin-env.sh.template
@@ -29,7 +29,7 @@
 # export ZEPPELIN_SSL_PORT                      # ssl port (used when ssl environment variable is set to true)
 # export ZEPPELIN_JMX_ENABLE                    # Enable JMX feature by defining "true"
 # export ZEPPELIN_JMX_PORT                      # Port number which JMX uses. If not set, JMX won't be enabled
-# export ZEPPELIN_EVENT_SERVER_PORT             # Port for the Zeppelin Interpreter Event Server. If not set, an available port will be used automatically.
+# export ZEPPELIN_SERVER_RPC_PORT               # Port for the Zeppelin Interpreter Event Server. If not set, an available port will be used automatically.
 
 # export ZEPPELIN_LOG_DIR                       # Where log files are stored.  PWD by default.
 # export ZEPPELIN_PID_DIR                       # The pid files are stored. ${ZEPPELIN_HOME}/run by default.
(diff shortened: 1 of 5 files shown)


Reamer:
I understand your approach. Currently I use the `ZEPPELIN_SERVER_RPC_PORTRANGE` `12320:12320` in my K8s landscape to configure a K8s-Service cleanly.
Why did you decide on the name `ZEPPELIN_EVENT_SERVER_PORT`, wouldn't `ZEPPELIN_SERVER_RPC_PORT` be better?

tbonelee:
@Reamer 
I considered both `ZEPPELIN_EVENT_SERVER_PORT` and `ZEPPELIN_SERVER_RPC_PORT`. I chose the former because the term "SERVER" in the latter could be too general, making it unclear which server it refers to, and `ZEPPELIN_SERVER_RPC_PORT` may be easily confused with `ZEPPELIN_SERVER_RPC_PORTRANGE` due to the similarity in names.

However, aligning with previous variable names might improve predictability as well, which is why I didn't have a strong preference for the former option from the beginning. What do you think?

Reamer:
I favor `ZEPPELIN_SERVER_RPC_PORT`. Can you remove `portRange` as a class variable of `RemoteInterpreterEventServer` and perform the evaluation of the port or portrange in the `start()`-method.
Furthermore, I do not see any documentation for `ZEPPELIN_SERVER_RPC_PORTRANGE` you can also add it near the documentation for `ZEPPELIN_SERVER_RPC_PORT` .

tbonelee:
@Reamer Thanks for the feedback. I've added some commits based on your suggestions:
- Renamed the `ZEPPELIN_EVENT_SERVER_PORT` variable to `ZEPPELIN_SERVER_RPC_PORT`.
- Removed `portRange` variable from `RemoteInterpreterEventServer`.
- Added documentation for the `ZEPPELIN_SERVER_RPC_PORTRANGE` and `ZEPPELIN_INTERPRETER_RPC_PORTRANGE` variables.

Reamer (approved):
LGTM

PR #4897, opened 2024-11-05
[ZEPPELIN-6147] Make interpreter searching case-insensitive in new ui

seung-00:
### What is this PR for?
The interpreter search in the new UI doesn’t ignore case sensitivity, unlike the old UI. I updated it to be case-insensitive in the new UI as well for better UX.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-6147](https://issues.apache.org/jira/browse/ZEPPELIN-6147/)

* Ensure interpreters can be searched case-insensitively

**old ui**

**new ui (before update)**

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/interpreter/interpreter.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/interpreter/interpreter.component.ts
@@ -46,9 +46,8 @@ export class InterpreterComponent implements OnInit, OnDestroy {
   }
 
   filterInterpreters(value: string) {
-    const lowerCaseValue = value.toLowerCase();
-    this.filteredInterpreterSettings = this.interpreterSettings.filter(e =>
-      e.name.toLowerCase().includes(lowerCaseValue)
+    this.filteredInterpreterSettings = this.interpreterSettings.filter(
+      e => e.name.search(new RegExp(value, 'i')) !== -1
     );
 
     this.cdr.markForCheck();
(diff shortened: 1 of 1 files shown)


zeppelin-web-angular/src/app/pages/workspace/interpreter/interpreter.component.ts
```diff
@@ -46,7 +46,11 @@ export class InterpreterComponent implements OnInit, OnDestroy {
   }
 
   filterInterpreters(value: string) {
-    this.filteredInterpreterSettings = this.interpreterSettings.filter(e => e.name.search(value) !== -1);
+    const lowerCaseValue = value.toLowerCase();
+    this.filteredInterpreterSettings = this.interpreterSettings.filter(e =>
+      e.name.toLowerCase().includes(lowerCaseValue)
+    );
+
```
tbonelee, line 53:
```suggestion
    this.filteredInterpreterSettings = this.interpreterSettings.filter(e => e.name.search(new RegExp(value, 'i')) !== -1);

```
Consider using the 'i' flag with RegExp here. This way, I think we can support both Regex patterns and regular string searches while keeping it case-insensitive, which would closely match the original functionality.

seung-00: @tbonelee Makes sense. I've adapted it. [0727f10](https://github.com/apache/zeppelin/pull/4897/commits/0727f10d96060c83f6a68686ed3eba62821e4738)

tbonelee (approved):
It works as expected. LGTM

pan3793 (approved):

Reamer:

PR #4900, opened 2024-11-15
[ZEPPELIN-6151] Remove AngularJS -> Angular Code Migration Suggestion Feature

tbonelee:
### What is this PR for?
I propose removing the feature in Zeppelin's new UI that converts AngularJS code to Angular code.

This feature currently prompts users when they run code with the `%angular` prefix, asking whether to migrate AngularJS code to Angular, and performs the migration if confirmed.

The reasons for suggesting its removal are as follows:

- The library supporting this feature is not compatible with the latest Angular versions due to peer dependency issues, making it difficult to upgrade Angular.
- It is likely that this feature is not widely used by the community.

Removing this feature will reduce unnecessary library dependencies, which will help streamline future updates to the latest Angular versions.

If you have any other opinions, please feel free to share them.

### What type of PR is it?
Removing Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6151

### Questions:


pan3793 (approved):
+1, and will get it in if no further comments/objections until next Monday

pan3793:
Thanks, merged to master/0.12

PR #4911, opened 2025-03-11
[ZEPPELIN-6168] Fix k8s interpreter service name removing all dots from pod name

venom13k:
### What is this PR for?
Fix the problem with k8s interpreter service creating

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6168

Compiled and tested manually.

### Questions:
* No
* No
* No


pan3793:
@venom13k could you please add a UT for this method, with some positive and negative cases?

pan3793 (approved):

pan3793:
Thanks, merged to master/0.12.1

PR #4676, opened 2023-10-15
[ZEPPELIN-3725] Fix SQL injection

vtkhanh:
### What type of PR is it?
Security - SQL Injection

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-3725](https://issues.apache.org/jira/browse/ZEPPELIN-3725)

### How should this be tested?
* No behavior change
* Existing tests passed.

### Questions:
* Does the license files need to update? NO
* Is there breaking changes for older versions? NO
* Does this needs documentation? NO

PR #4926, opened 2025-04-29
[ZEPPELIN-6186] fix searching user using JdbcRealm

sh1nj1:
### What is this PR for?

This PR fixes a broken feature where getUserList fails when using JdbcRealm for Shiro authentication.

The previous PR addressing SQL injection introduced an error that prevents getUserList from working correctly with JdbcRealm. As a result, users are unable to set notebook permissions.

Ref: #4676

### What type of PR is it?

Bug Fix

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-3725
* https://issues.apache.org/jira/browse/ZEPPELIN-6186

Added new unit test should be passed

* `./mvnw clean package -pl zeppelin-server -am`

Verify in UI

* Go to notebook permission settings
* Type username and see if username listed

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
@@ -466,10 +466,8 @@ public class ShiroAuthenticationService implements AuthenticationService {
       ps = con.prepareStatement(userquery);
       ps.setString(1, "%" + searchText + "%");
       rs = ps.executeQuery();
-      int count = 0;
-      while (rs.next() && count < numUsersToFetch) {
-        userlist.add(rs.getString(1).trim());
-        count++;
+      while (rs.next() && userlist.size() < numUsersToFetch) {
+        userlist.add(rs.getString(1));
       }
     } catch (Exception e) {
       LOGGER.error("Error retrieving User list from JDBC Realm", e);
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
@@ -26,7 +26,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import jakarta.inject.Inject;
+import java.util.regex.Pattern;
+
 import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
 import javax.naming.directory.Attributes;
@@ -34,6 +35,8 @@ import javax.naming.directory.SearchControls;
 import javax.naming.directory.SearchResult;
 import javax.naming.ldap.LdapContext;
 import javax.sql.DataSource;
+
+import jakarta.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.shiro.UnavailableSecurityManagerException;
@@ -67,6 +70,12 @@ public class ShiroAuthenticationService implements AuthenticationService {
   private static final String LDAP_GROUP_REALM = "org.apache.zeppelin.realm.LdapGroupRealm";
   private static final String ACTIVE_DIRECTORY_GROUP_REALM = "org.apache.zeppelin.realm.ActiveDirectoryGroupRealm";
   private static final String JDBC_REALM = "org.apache.shiro.realm.jdbc.JdbcRealm";
+
+  private static final Pattern VALID_SQL_NAME_IDENTIFIER_PATTERN = Pattern.compile("^[a-zA-Z0-9_]+$");
+
+  private static boolean isValidSqlIdentifier(String name) {
+    return name != null && VALID_SQL_NAME_IDENTIFIER_PATTERN.matcher(name).matches();
+  }
 
   private final ZeppelinConfiguration zConf;
 
@@ -174,7 +183,7 @@ public class ShiroAuthenticationService implements AuthenticationService {
             usersList.addAll(
                 getUserList((ActiveDirectoryGroupRealm) realm, searchText, numUsersToFetch));
           } else if (JDBC_REALM.equals(realClassName)) {
-            usersList.addAll(getUserList((JdbcRealm) realm));
+            usersList.addAll(getUserList((JdbcRealm) realm, searchText, numUsersToFetch));
           }
         }
       }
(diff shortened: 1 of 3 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -415,26 +424,39 @@ private List<String> getUserList(JdbcRealm obj) {
...
+	  }
+	  if (!isValidSqlIdentifier(tablename)) {
+	    throw new IllegalArgumentException(
+	      "Invalid table name in authenticationQuery to build userlist query: "
+				  + authQuery + ", allowed pattern: " + VALID_SQL_NAME_IDENTIFIER_PATTERN
+				  + ", name identifier: [" + tablename + "]");
+	  }
 
-      userquery = "SELECT ? FROM ?";
+	  userquery = String.format("SELECT %s FROM %s WHERE %s LIKE ?", username, tablename, username);
```
jongyoul, line 458: I think it would be better to use `PreparedStatement` instead of making a query using `String.format`.

jongyoul:
Thank you for your contribution! I left one comment. Please check it.

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
sh1nj1: That's the last PR's issue for this, you can use parameter replacement in where clause, not the table name or projections. That's why I make this PR.
`SELECT ? FROM ?` makes query syntax error because it create invalid query by the PreparedStatement because you can not use '?' in select clause or from clause.

jongyoul: Can you please explain more about it? If PreparedStatement disallows it, it shouldn't be passed as a query. The test case doesn't cover it as well.

jongyoul: I'm not 100% sure but it could be `SELECT ? FROM ? WHERE ? LIKE ?` and we can pass it as `ps.setString(1, "%" + username + "%")`?

sh1nj1: https://stackoverflow.com/a/2917509/1395707 this explains it better than me.

I use `PreparedStatement` only set `WHERE username LIKE ?` but remove from select clause and from clause.
If I use old implementation, it creates an error in test case that I added here.

jongyoul: Ah .. I got the point. You meant we shouldn't use an identifier as a literal in the preparedStatement.

jongyoul (approved):
LGTM. By the way, we can refator the whole logic to parse a tablename and username, and get them from separate variables and pass them directly. Can you consider refactoring the logic as well?

jongyoul:
@sh1nj1 Can you please create a new ticket so that I can give you credit as the assignee for this issue?

sh1nj1:
@jongyoul updated the ticket ID

> Can you consider refactoring the logic as well?

I can do in a separate PR

sh1nj1:
@Reamer fixed indentation

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -443,11 +464,12 @@ private List<String> getUserList(JdbcRealm obj) {
     try {
       con = dataSource.getConnection();
       ps = con.prepareStatement(userquery);
-      ps.setString(1, username);
-      ps.setString(2, tablename);
+      ps.setString(1, "%" + searchText + "%");
       rs = ps.executeQuery();
-      while (rs.next()) {
+      int count = 0;
```
Reamer, line 469: Why the `count`?

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -415,26 +423,39 @@ private List<String> getUserList(JdbcRealm obj) {
...
       dataSource = (DataSource) FieldUtils.readField(obj, "dataSource", true);
       authQuery = (String) FieldUtils.readField(obj, "authenticationQuery", true);
-      LOGGER.info(authQuery);
+      LOGGER.debug("authenticationQuery={}", authQuery);
       String authQueryLowerCase = authQuery.toLowerCase();
       retval = authQueryLowerCase.split("from", 2);
       if (retval.length >= 2) {
         retval = retval[1].split("with|where", 2);
-        tablename = retval[0];
+        tablename = retval[0].trim();
```
Reamer, line 431: Just a little thing.
Meanwhile, I would use `strip()` because this method is Unicode aware. What do you think?

```suggestion
        tablename = retval[0].strip();
```

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
```diff
@@ -415,26 +423,39 @@ private List<String> getUserList(JdbcRealm obj) {
...
         retval = retval[1].split("where", 2);
         if (retval.length >= 2) {
           retval = retval[1].split("=", 2);
         } else {
           retval = retval[0].split("=", 2);
         }
-        username = retval[0];
+        username = retval[0].trim();
```
Reamer, line 438:
```suggestion
        username = retval[0].strip();
```

Reamer (changes requested):
The indentation looks much better. I have two minor points to mention.

zeppelin-server/src/main/java/org/apache/zeppelin/service/ShiroAuthenticationService.java
sh1nj1: is it naming issue, no need issue or anything else?
BTW, I removed the variable.

Reamer (approved):
LGTM
If no further comments are received, I will merge the whole thing next Monday.

Reamer:
Sorry for the delay. For some unknown reason the pipeline `interpreter-test-non-core (11)` always fails.

Reamer:
@sh1nj1 Please perform a rebase to the current master. There have been corrections to the CI pipeline.

sh1nj1:
@Reamer pushed rebased codes

Reamer:

PR #4931, opened 2025-05-17
[ZEPPELIN-6191] Remove loading external resources in Zeppelin documentations

tbonelee:
### What is this PR for?
Due to stricter Content Security Policy (CSP) settings enforced by Apache, the Zeppelin website fails to load some externally hosted resources such as CSS and jQuery scripts. As a result, fonts are broken and some interactive components no longer function properly.

This PR removes external resource dependencies from the documentation site templates to ensure all necessary assets are bundled and served locally. These changes will be applied to future documentation pages generated during version releases.

### What type of PR is it?
Bug Fix
Documentation

### What is the Jira issue?

- Ensure all necessary assets are available locally in the documentation build
- Validate styling and interactivity work without external dependencies

### Questions:


Reamer:
Wouldn't it be better to download the files from the original sources when building?

tbonelee:
I couldn' find a reliable way to include external CDN assets automatically with Jekyll. Since the documentation source code doesn't change often, I thought it would be fine to include the asset files directly in the repository.

But I'm open to suggestions, do you have any ideas how we could handle this better?

Reamer:
I have not found a way with jekyll either. I think we can continue with a copy.

Reamer:
What do you think of the version in the folder or file name? You have realized it for jQuery, but not for the other files.

Reamer (approved):
LGTM
@jongyoul Can you take over and relaunch the homepage?

jongyoul:
@Reamer If we cherry-pick this into branch-0.12, we can publish a new homepage.

Reamer:
@jongyoul Merged into master and branch-0.12.
@tbonelee Thanks for your contribution.

PR #4941, opened 2025-06-07
[ZEPPELIN-6195] Fix Selenium tests by switching from Chrome to Firefox and upgrading outdated Spark

ParkGyeongTae:
### What is this PR for?
 The `test-selenium-with-spark-module-for-spark-3-4` job in GitHub Actions was consistently failing with the error:
  ```bash
 from unknown command: 'Runtime.evaluate' wasn't found
 ```
 
 This issue was caused by ChromeDriver not recognizing the `Runtime.evaluate` command.
To resolve it, the browser used in the test was switched from Chrome to Firefox.

During the migration to Firefox, another issue occurred:
```bash
Unable to download from https://dlcdn.apache.org/spark/spark-3.4.2/spark-3.4.2-bin-hadoop3.tgz
java.io.FileNotFoundException: https://dlcdn.apache.org/spark/spark-3.4.2/spark-3.4.2-bin-hadoop3.tgz
```
This was due to the Spark 3.4.2 binaries no longer being available from the Apache mirror.
The test environment was therefore updated to use Spark 3.5.6, which resolved the download problem.

Summary of changes:
- Switched Selenium tests from Chrome to Firefox (non-snap) using geckodriver.
- Updated Spark version in the test environment from 3.4.2 to 3.5.6 to fix missing binary downloads.

Before: test-selenium-with-spark-module-for-spark-3-4 (Chrome, Spark 3.4.2)
After: test-selenium-with-spark-module-for-spark-3-5 (Firefox, Spark 3.5.6)

### What type of PR is it?
Bug Fix

### Todos
* [x] - Verified that other integration tests (if any) are not affected
* [x] - Confirmed long-term compatibility with the Chrome for Testing approach

### What is the Jira issue?
* Jira: https://issues.apache.org/jira/browse/ZEPPELIN-6195

- Run the `test-selenium-with-spark-module-for-spark-3-4` job in GitHub Actions.
- Verify that `AuthenticationIT` and any other affected Selenium tests pass without encountering `Runtime.evaluate` errors.

N/A

### Questions:

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -24,6 +24,7 @@ env:
   SPARK_LOCAL_IP: 127.0.0.1
   ZEPPELIN_LOCAL_IP: 127.0.0.1
   INTERPRETERS: '!hbase,!jdbc,!file,!flink,!cassandra,!elasticsearch,!bigquery,!alluxio,!livy,!groovy,!java,!neo4j,!sparql,!mongodb'
+  ZEPPELIN_SELENIUM_BROWSER: firefox
 
 permissions:
   contents: read # to fetch code (actions/checkout)
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -24,7 +24,6 @@ env:
   SPARK_LOCAL_IP: 127.0.0.1
   ZEPPELIN_LOCAL_IP: 127.0.0.1
   INTERPRETERS: '!hbase,!jdbc,!file,!flink,!cassandra,!elasticsearch,!bigquery,!alluxio,!livy,!groovy,!java,!neo4j,!sparql,!mongodb'
-  ZEPPELIN_SELENIUM_BROWSER: firefox
 
 permissions:
   contents: read # to fetch code (actions/checkout)
@@ -35,16 +34,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-
       - name: Tune Runner VM
         uses: ./.github/actions/tune-runner-vm
-
       - name: Set up JDK 11
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: 11
-
       - name: Cache local Maven repository
         uses: actions/cache@v4
         with:
@@ -56,38 +52,10 @@ jobs:
           key: ${{ runner.os }}-zeppelin-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-zeppelin-
-
-      - name: Install Firefox (non-snap) and geckodriver
-        run: |
-          sudo snap remove firefox || true
-
-          sudo add-apt-repository -y ppa:mozillateam/ppa
-          sudo tee -a /etc/apt/preferences.d/mozilla-firefox <<EOF
-          Package: *
-          Pin: release o=LP-PPA-mozillateam
-          Pin-Priority: 1001
-          EOF
-
-          sudo apt-get update
-          sudo apt-get install -y firefox
-
-          GECKODRIVER_VERSION=$(curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | grep tag_name | cut -d '"' -f 4)
-          wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          tar -xzf geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          chmod +x geckodriver
-          sudo mv geckodriver /usr/local/bin/
-          geckodriver --version
-          firefox --version
-
       - name: Install application
-        run: ./mvnw clean install -DskipTests -am -pl zeppelin-web -Pweb-classic -Pspark-scala-2.12 -Pspark-3.5 -Pweb-dist ${MAVEN_ARGS}
-
+        run: ./mvnw clean install -DskipTests -am -pl zeppelin-web -Pweb-classic -Pspark-scala-2.12 -Pspark-3.4 -Pweb-dist ${MAVEN_ARGS}
       - name: Run headless test
-        run: |
-          xvfb-run --auto-servernum --server-args="-screen 0 1024x768x24" \
-            ./mvnw verify -pl zeppelin-web \
-            -Pweb-classic -Pspark-scala-2.12 -Pspark-3.5 -Pweb-dist -Pweb-e2e ${MAVEN_ARGS}
-
+        run: xvfb-run --auto-servernum --server-args="-screen 0 1024x768x24" ./mvnw verify -pl zeppelin-web -Pweb-classic -Pspark-scala-2.12 -Pspark-3.4 -Pweb-dist -Pweb-e2e ${MAVEN_ARGS}
       - name: Print zeppelin logs
         if: always()
         run: if [ -d "logs" ]; then cat logs/*; fi
@@ -123,24 +91,22 @@ jobs:
     defaults:
       run:
         shell: bash -l {0}
+    env:
+      ZEPPELIN_SELENIUM_BROWSER: firefox
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-
       - name: Install Firefox (non-snap) and geckodriver
         run: |
           sudo snap remove firefox || true
-
           sudo add-apt-repository -y ppa:mozillateam/ppa
           sudo tee -a /etc/apt/preferences.d/mozilla-firefox <<EOF
           Package: *
           Pin: release o=LP-PPA-mozillateam
           Pin-Priority: 1001
           EOF
-
           sudo apt-get update
           sudo apt-get install -y firefox
-
           GECKODRIVER_VERSION=$(curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | grep tag_name | cut -d '"' -f 4)
           wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
           tar -xzf geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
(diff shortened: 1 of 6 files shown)


Reamer:
Thank you for wanting to correct the Selenium tests. I see that your solution works. However, I find the downgrade very problematic, as it is not future-proof.

I had also looked at the problem some time ago and found a problem with the maximize method.
https://github.com/apache/zeppelin/blob/94ae845cd4a7fd6058e936d82f817e1ef8618a64/zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java#L159

The problem with Selenium is analyzed in more detail here.
https://github.com/SeleniumHQ/selenium/issues/15358

I would suggest a fixed height and width. Unfortunately, my initial tests were not successful. Perhaps you can find suitable values.

I like the update of the Selenium version, but not the downgrade of Chrome.

ParkGyeongTae:
@Reamer 
I always appreciate your feedback. I understand your point, and I’ve reviewed the discussion in the link you shared.
I’ll look into a way to keep Chrome up to date while ensuring the tests run without errors.

pan3793:
The issue looks to be there for a long time, if it is specific to Chrome, how about switching the test to Firefox?

Reamer:
Firefox has other bugs as far as I remember

ParkGyeongTae:
@pan3793 @Reamer 
Thanks for the suggestion! I switched the tests to Firefox, and they are now passing successfully. Please take a look and let me know if you have any further concerns.

.github/workflows/frontend.yml
```diff
@@ -86,32 +86,37 @@ jobs:
...
-          path: |
-            ~/.m2/repository
-            !~/.m2/repository/org/apache/zeppelin/
-            ~/.spark-dist
-            ~/.cache
-          key: ${{ runner.os }}-zeppelin-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-zeppelin-
+
+      - name: Install Firefox (non-snap) and geckodriver
```
Reamer, line 99: What's wrong with the Snap version?

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -80,83 +80,44 @@ public WebDriver getWebDriver() {
...
-        return new SafariDriver();
-      } catch (Exception e) {
-        LOG.error("Exception in WebDriverManager while SafariDriver ", e);
-        return null;
-      }
-    };
+    WebDriver driver = getFirefoxDriver();
 
-    WebDriver driver;
-    switch (SystemUtils.getEnvironmentVariable("ZEPPELIN_SELENIUM_BROWSER", "").toLowerCase(Locale.ROOT)) {
```
Reamer: I think this switch with the environment variable is quite good; we should use this behavior in the tests to start a Firefox browser.

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -80,83 +80,44 @@ public WebDriver getWebDriver() {
...
       }
     }
 
     assertTrue(loaded);
 
-    driver.manage().window().maximize();
+    try {
+      driver.manage().window().maximize();
+    } catch (Exception e) {
+      LOG.warn("Failed to maximize Firefox window. Consider using setSize instead.", e);
```
Reamer: The error message should be generic with regard to the browser used.

Reamer (changes requested):
Thank you for looking at the issue again. In my opinion, too much useful code for other platforms has been removed.

.github/workflows/frontend.yml
ParkGyeongTae: @Reamer 
I took action because I encountered the following error when running the tests:
```bash
Error:  Tests run: 21, Failures: 0, Errors: 19, Skipped: 2, Time elapsed: 53.89 s <<< FAILURE! -- in org.apache.zeppelin.integration.ParagraphActionsIT
Error:  org.apache.zeppelin.integration.ParagraphActionsIT.testSingleDynamicFormTextInput -- Time elapsed: 0.903 s <<< ERROR!
org.openqa.selenium.SessionNotCreatedException: 
Could not start a new session. Response code 400. Message: binary is not a Firefox executable 
...
```
It seems that the Snap version of Firefox caused issues because the WebDriver couldn’t recognize it as a valid Firefox binary.

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
ParkGyeongTae: Thanks for the suggestion! I updated the message from "Failed to maximize Firefox window. Consider using setSize instead." to "Failed to maximize browser window. Consider using setSize() instead." to make it more generic.

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
ParkGyeongTae: I agree with you, and I've reverted the code to use this behavior accordingly.

pan3793:
I see the failed `run-e2e-tests-in-zeppelin-web` still prints

```
[INFO] [09:56:05] I/launcher - Running 1 instances of WebDriver
[INFO] [09:56:05] I/direct - Using ChromeDriver directly...
```

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -133,17 +133,17 @@ private WebDriver constructWebDriver(int port) {
...
     while (System.currentTimeMillis() - start < 60 * 1000) {
       // wait for page load
       try {
-        (new WebDriverWait(driver, Duration.ofSeconds(30))).until(new ExpectedCondition<Boolean>() {
+        (new WebDriverWait(driver, Duration.ofSeconds(60))).until(new ExpectedCondition<Boolean>() {
           @Override
           public Boolean apply(WebDriver d) {
             return d.findElement(By.xpath("//i[@uib-tooltip='WebSocket Connected']"))
-                .isDisplayed();
+                    .isDisplayed();
```
pan3793: please restore the unnecessary indention changes

ParkGyeongTae: I’ve restored the indentation to how it was originally.

pan3793: there are other 2 places, please check the diff

ParkGyeongTae: I’ve reverted the other two places as well.

.github/workflows/frontend.yml
```diff
@@ -24,6 +24,7 @@ env:
   SPARK_LOCAL_IP: 127.0.0.1
   ZEPPELIN_LOCAL_IP: 127.0.0.1
   INTERPRETERS: '!hbase,!jdbc,!file,!flink,!cassandra,!elasticsearch,!bigquery,!alluxio,!livy,!groovy,!java,!neo4j,!sparql,!mongodb'
+  ZEPPELIN_SELENIUM_BROWSER: firefox
```
Reamer: Please only add the env to the Selenium tests.

Reamer (changes requested):
Please completely revert your changes to the `run-e2e-tests-in-zeppelin-web` test. We should stick with one frontend test for this pull request and not touch both.

.github/workflows/frontend.yml
ParkGyeongTae: I’ve updated the workflow to add the environment variable only to the `test-selenium-with-spark-module-for-spark-3-5` job. As for the error in the `run-e2e-tests-in-zeppelin-web` job, I’ll address it in a separate issue.

zeppelin-web/package-lock.json
```diff
@@ -11800,7 +11800,7 @@
       "dev": true,
       "dependencies": {
         "concat-stream": "^1.5.0",
-        "duplexify": "^3.4.2",
+        "duplexify": "^3.5.6",
```
Reamer: Was there any unwanted search and replace behavior here?
Same version change as with the Spark version.

zeppelin-web/package-lock.json
```diff
@@ -17107,7 +17107,7 @@
...
```
Reamer: Same as above

Reamer (changes requested):
The question still needs to be clarified.

zeppelin-web/package-lock.json
ParkGyeongTae: I updated the `DEFAULT_SPARK_VERSION` in `zeppelin-test/src/main/java/org/apache/zeppelin/test/DownloadUtils.java` from `3.4.2` to `3.5.6`.

During debugging, I encountered the following error, which occurred because the `3.4.2` distribution file is no longer available. Upgrading the Spark version resolved the issue.

```bash
13:42:26,840  INFO org.apache.zeppelin.test.DownloadUtils:248 - Unable to download from https://dlcdn.apache.org/spark/spark-3.4.2/spark-3.4.2-bin-hadoop3.tgz
java.io.FileNotFoundException: https://dlcdn.apache.org/spark/spark-3.4.2/spark-3.4.2-bin-hadoop3.tgz
```

zeppelin-web/package-lock.json
ParkGyeongTae: Same as the above comment.

zeppelin-web/package-lock.json
Reamer: What does duplexify (a npm package) have to do with Spark?

ParkGyeongTae: I accidentally changed it and have reverted it back to 3.4.2.

zeppelin-web/package-lock.json
ParkGyeongTae: I changed it by mistake and reverted it back to 3.4.2.

pan3793:
@ParkGyeongTae thanks for fixing this, could you please update the PR title and description to reflect the final solution?

ParkGyeongTae:
@pan3793 
I’ve updated the PR title and description, as well as the Jira issue title and description.

pan3793 (approved):

Reamer:

PR #4949, opened 2025-06-25
[ZEPPELIN-6204] do not allow init params in JDBC URLs for H2

pjfanning:
### What is this PR for?
ZEPPELIN-6204

Slight tidy of the existing disallow list for strings in JDBC urls so that they are checked against just the query params and not the hostname in the URL. 

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:
* Does the license files need to update? no

### DIFF
--- a/jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
+++ b/jdbc/src/test/java/org/apache/zeppelin/jdbc/JDBCInterpreterTest.java
@@ -58,7 +58,11 @@ import static org.apache.zeppelin.jdbc.JDBCInterpreter.DEFAULT_STATEMENT_PRECODE
 import static org.apache.zeppelin.jdbc.JDBCInterpreter.DEFAULT_URL;
 import static org.apache.zeppelin.jdbc.JDBCInterpreter.DEFAULT_USER;
 import static org.apache.zeppelin.jdbc.JDBCInterpreter.PRECODE_KEY_TEMPLATE;
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 
 /**


Reamer:
I don't know whether we are preventing features that others want to use with these changes. 

Doesn't it make more sense to secure the H2 server according to the requirements? After all, Zeppelin is just one JDBC client of many.

pjfanning:
@Reamer @jongyoul is it ok to consider this for merge now? There are probably a few more query params that could be added to the disallow list but this covers some of the more commonly mentioned ones.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -589,17 +597,123 @@ public Connection getConnection(InterpreterContext context)
...
+
+    if (containsKeyIgnoreCase(params, ALLOW_LOAD_LOCAL) ||
+            containsKeyIgnoreCase(params, ALLOW_LOAD_LOCAL_IN_FILE_NAME) ||
+            containsKeyIgnoreCase(params, ALLOW_LOCAL_IN_FILE_NAME) ||
+            containsKeyIgnoreCase(params, ALLOW_URL_IN_LOCAL_IN_FILE_NAME) ||
+            containsKeyIgnoreCase(params, ALLOW_LOAD_LOCAL_IN_FILE_IN_PATH) ||
+            containsKeyIgnoreCase(params, AUTO_DESERIALIZE) ||
+            containsKeyIgnoreCase(params, SOCKET_FACTORY)) {
+      throw new IllegalArgumentException("Connection URL contains sensitive configuration");
+    }
```
seung-00, line 612:
```suggestion
       Stream.of(ALLOW_LOAD_LOCAL, ALLOW_LOCAL_IN_FILE_NAME, ...).forEach(rule -> {
            if (containsKeyIgnoreCase(params, rule)) throw new IllegalArgumentException("Connection URL contains sensitive configuration: " + rule);
        });
```

I find this version easier to read.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -589,17 +597,123 @@ public Connection getConnection(InterpreterContext context)
   }
 
   private void validateConnectionUrl(String url) {
-    String decodedUrl;
-    decodedUrl = URLDecoder.decode(url, StandardCharsets.UTF_8);
+    final String decodedUrl = urlDecode(url, url, 0);
+    final Map<String, String> params = parseUrlParameters(decodedUrl);
```
tbonelee, line 601: Just a small thought. I was wondering if decoding the full URL before parsing could lead to unexpected behavior in some edge cases. For example, if a parameter value contains encoded characters like `%26` (&) or `%3D` (=), decoding them too early might interfere with how the parameters are parsed.

I’m not sure if values like passwords are actually allowed or commonly used in this context, so this might not be relevant. But I thought it was worth mentioning just in case. 🙏

tbonelee:
Just a small thought. I was wondering if decoding the full URL before parsing could lead to unexpected behavior in some edge cases. For example, if a parameter value contains encoded characters like `%26` (&) or `%3D` (=), decoding them too early might interfere with how the parameters are parsed.

I’m not sure if values like passwords are actually allowed or commonly used in this context, so this might not be relevant. But I thought it was worth mentioning just in case. 🙏

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -589,17 +597,123 @@ public Connection getConnection(InterpreterContext context)
...
 
+  /**
+   * Decode the URL encoded string recursively until no more decoding is needed.
+   * This is to handle cases where the URL might be double-encoded.
+   *
+   * @param url the original URL (for logging purposes)
+   * @param encoded the URL encoded string
+   * @param recurseCount the current recursion depth
+   * @return the decoded string
+   * @throws IllegalArgumentException if the recursion depth exceeds 100
```
seung-00: The function seems to recurse up to 10 times.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -156,14 +156,22 @@ public class JDBCInterpreter extends KerberosInterpreter {
...
   private static final String ALLOW_LOCAL_IN_FILE_NAME = "allowLocalInfile";
 
   private static final String ALLOW_URL_IN_LOCAL_IN_FILE_NAME = "allowUrlInLocalInfile";
 
+  private static final String AUTO_DESERIALIZE = "autoDeserialize";
+
+  private static final String SOCKET_FACTORY = "socketFactory";
+
+  private static final String INIT = "INIT";
+
```
seung-00, line 174: could be nice to group these into a Set.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -589,17 +597,123 @@ public Connection getConnection(InterpreterContext context)
...
+    // MySQL supports parentheses in the URL
+    // https://dev.mysql.com/doc/connectors/en/connector-j-reference-jdbc-url-format.html
+    // eg jdbc:mysql://(host=myhost,port=1111,allowLoadLocalInfile=true)/db
+    int parensIndex = extractFromParens(url, 0, parameters);
+    while (parensIndex > 0) {
+      parensIndex = extractFromParens(url, parensIndex, parameters);
+    }
+
+    // Split the URL into the base part and the parameters part
+    String[] parts = url.split("[?&;]");
```
seung-00, line 655: could be a bit faster if the regex was cached to avoid rebuilding every time

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
```diff
@@ -589,17 +597,123 @@ public Connection getConnection(InterpreterContext context)
...
+      return -1;
+    }
+    final int endIndex = input.indexOf(')', startIndex);
+    if (startIndex != -1 && endIndex != -1) {
+      String params = input.substring(startIndex + 1, endIndex);
+      String[] keyValuePairs = params.split(",");
+      for (String pair : keyValuePairs) {
+        splitNameValue(pair, parameters);
+      }
+    }
```
seung-00, line 697: Does it work well if parentheses are in the URL query parameters?
`jdbc:mysql://.../db?password=abc(def)`

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
pjfanning: I can't commit this unless it compiles so please don't use the suggestion option to suggest incomplete code.
I also disagree that this is more readable. The existing code in this PR is closer to the existing style in the existing codebase.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
pjfanning: I'm happy to stick with the existing code style.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
pjfanning: thanks - I fixed the comment. I recently reduced the recurse count from 100

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
pjfanning: the cost of this regex will be tiny compared to the I/O cost of using JDBC commands so I'm not very interested in optimising every line of code here

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
pjfanning: the code doesn't care about what the values are, it only cares about the param names - write some test cases if you are suspicious

pjfanning:
> Just a small thought. I was wondering if decoding the full URL before parsing could lead to unexpected behavior in some edge cases. For example, if a parameter value contains encoded characters like `%26` (&) or `%3D` (=), decoding them too early might interfere with how the parameters are parsed.
> 
> I’m not sure if values like passwords are actually allowed or commonly used in this context, so this might not be relevant. But I thought it was worth mentioning just in case. 🙏

The code in this PR ignores the param values. It only worries about the param names.
The original URL as entered by the user is passed to the JDBC layer - my code splits up the URL but only for my checks for suspicious param names.

jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
seung-00: I tested the following cases:
```java
testBannedMySQLQueryParamWithValidParam("password=allowLoadLocalInfile"); // 1. fail
testBannedMySQLQueryParamWithValidParam("password=(allowLoadLocalInfile)"); // 2. pass
```
I might be missing something, but test case 2 seems unintended. 
Part of the value `allowLoadLocalInfile` appears to be treated as a key due to the parentheses.

pjfanning: thanks - I'll look into the issue - but in the end of the day, it's not a very likely password
and the existing checks before this PR, they just do a string match on the whole URL so that code wouldn't handle this either

tbonelee:
@pjfanning Thanks for the clarification. I took a closer look and you're right. Decoding encoded special characters in values before validation would at most introduce an extra key, but it does not interfere with the parsing of subsequent parameters. I had initially worried about false negatives (i.e. cases where disallowed parameters might go undetected), but that doesn't seem to be the case. Thanks again, and sorry for the confusion.

pjfanning:
@jongyoul @Reamer would you be able to review this change?

pjfanning:
@jongyoul @Reamer would you be able to review this change? I think this fixes a couple of issues in the existing implementation.

pjfanning:
@jongyoul do you think this could be merged?

pjfanning:
@jongyoul @Reamer I am not a committer so can't merge this. Would either of you be in a position to do so?

tbonelee:
I’ve merged this into master.
Since I considered it a breaking change, I didn’t include it in branch-0.12.
Do you think it needs to be backported?

pjfanning:
Thanks @tbonelee - I would regard it more as a bug fix or fixes. It shouldn't change the behaviour for any use cases that we want to support.

PR #4951, opened 2025-06-26
[ZEPPELIN-6206] Fix notebook paths in job manager page

tbonelee:
### What is this PR for?
URLs to notebook page are set wrong in job manager page

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6206

- Enable cron and job manager configurations and check a links in the page
  - `zeppelin.jobmanager.enable` : `true`
  - `zeppelin.notebook.cron.enable`: `true`
  - `zeppelin.notebook.cron.folders` : `/`
image

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/job-manager/job/job.component.html
+++ b/zeppelin-web-angular/src/app/pages/workspace/job-manager/job/job.component.html
@@ -17,7 +17,7 @@
       [nzType]="icon"
       nzTheme="outline">
     </i>
-    <a [routerLink]="['notebook', note.noteId]" [innerHTML]="note.noteName | nzHighlight: highlight: 'gi': 'mark-highlight'"></a> -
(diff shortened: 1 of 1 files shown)


Reamer:
Perhaps this is related to https://github.com/apache/zeppelin/pull/4947
What do you think?

tbonelee:
@Reamer 
Thanks Reamer! I had a look at #4947, and while both involve notebook paths, I believe the issues are different.

That one addresses URI-encoded names, whereas this PR fixes a routing issue in Angular, where links were generated as relative paths instead of root-based, leading to incorrect URLs under nested routes like `/#/jobmanager`.

So while they seem related in topic, I think the underlying causes and fixes are different.

ParkGyeongTae:
👍 LGTM

Reamer:

PR #4952, opened 2025-06-26
[ZEPPELIN-6207] Bump Angular to 9 in zeppelin-web-angular

tbonelee:
### What is this PR for?
Upgrade Angular from version 8 to 9 in zeppelin-web-angular.

The current Zeppelin web(new) application uses Angular 8 which is incompatible with several modern development tools and libraries.

This upgrade is a transitional step toward supporting newer Angular versions in the future.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6207

- Check that all features work as expected.

### Questions:


Reamer:
Thank you for the great migration. I don't feel able to do a review here. @jongyoul Can you do the review?
I could build the version manually locally and click through the frontend a bit.

smileyboy2019:
[PolyNote](https://github.com/polynote/polynote)I hope the development interface is similar to IDE, with code prompts. Can we integrate PolyNote products or refer to it

jongyoul:
@Henry-Hong Could you please help review it?

jongyoul:
@smileyboy2019 I like to improve the frontend part. We had some trials, but failed to replace it. I think we need to have a plan on how we can replace it gradually. I like to discuss and hope to start a discussion in dev@

tbonelee:
Additionally, I upgraded the currently used editor library, `monaco-editor`, to a version that supports inline suggestion features. While we may eventually switch to a different editor library, I thought updating the one we're using now should take priority for the time being.

Further upgrades of the editor library are currently blocked by compatibility issues between its JavaScript syntax and the version of Webpack used by Angular 9. Once we upgrade to a higher version of Angular, we should be able to revisit and continue the upgrade.

zeppelin-web-angular/src/app/app.module.ts
```diff
@@ -71,7 +71,8 @@ registerLocaleData(en);
       useValue: {
         defaultEditorOption: {
           scrollBeyondLastLine: false
-        },
+          // TODO: Change 'as' to 'satisfies' when typescript version is over 4.9 to detect unsupported editor options at compile time.
```
Henry-Hong, line 74: 👍 👍 👍

zeppelin-web-angular/src/app/services/completion.service.ts
```diff
@@ -76,12 +78,19 @@ export class CompletionService extends MessageListenersManager {
...
+                      label: i.name,
+                      insertText: i.name,
+                      range: {
+                        startLineNumber: position.lineNumber,
+                        endLineNumber: position.lineNumber,
+                        startColumn: word.startColumn,
+                        endColumn: word.endColumn
+                      }
+                    })
+                  )
```
Henry-Hong, line 93: Such a nice work!!

Since these are UI changes, it would be really helpful to see an as-is/to-be demo on PR Readme.

Henry-Hong (approved):

Henry-Hong:
LGTM.

zeppelin-web-angular/src/app/app.module.ts
Henry-Hong: Let's make `ts version up` as an issue. I would do it.

Henry-Hong: https://issues.apache.org/jira/browse/ZEPPELIN-6236

zeppelin-web-angular/src/app/services/completion.service.ts
tbonelee: Thanks for the review!
This change is actually just adding the default values for the `range` property, which has has been required since v0.16.0.
You can find more details in the [CHANGELOG.md](https://github.com/microsoft/monaco-editor/blob/f420968fc91d2d10d9d3e47dc52e2f2bda0778dd/CHANGELOG.md?plain=1#L1019) or in [this VSCode issue](https://github.com/microsoft/vscode/issues/63454).

pan3793 (approved):
LGTM, verified locally by building and running some basic spark code

zeppelin-web-angular/src/app/services/completion.service.ts
Henry-Hong: Oh, so it wasn’t about the UI changes after all.

I assumed the updated version supported inline suggestions and that you had implemented that.

But from what I see, it looks like you just filled in the required values while keeping the previous UI, right?

Sorry for the interruption! and Thank you for the detailed reply.

pan3793:
Thanks, merged to master.

PR #4954, opened 2025-07-01
[MINOR] postgresql 42.7.7

pjfanning:
Our current version does not have a CVE but a new CVE applies to newer versions but latest version fixes that.
https://mvnrepository.com/artifact/org.postgresql/postgresql

Seems best to upgrade to latest version.

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -46,7 +46,7 @@
 
   <dependencies>
 
-	  <dependency>
+    <dependency>
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
       <version>${postgresql.version}</version>


jdbc/pom.xml
```diff
@@ -46,7 +46,7 @@
 
   <dependencies>
 
-	<dependency>
+	  <dependency>
```
Reamer: Can you fix the indentation problem here?

pjfanning: this line is badly aligned - I didn't spot the tab so I have a new commit that fixes the indent by removing the tab

- or do you want to not worry about this indent at all?

Reamer (approved):
LGTM

jdbc/pom.xml
Reamer: Your correction of the line is the best way.

Reamer:
Merged into master

PR #4963, opened 2025-07-12
[ZEPPELIN-6203] Connect shortcut for info button

dididy:
### What is this PR for?
image

Nothing happened when I click above info button inside of Zeppelin Notebook. So I connect this function work properly. 

This button related in [here](https://github.com/apache/zeppelin/blob/526eb15b8732909ab994adba76e0c3e33bfa001c/zeppelin-web-angular/src/app/pages/workspace/notebook/action-bar/action-bar.component.ts#L239) so I made shared shortcut component and connect it. Now it works well.

스크린샷 2025-07-12 오후 6 10 05

I refer to [zeppelin-web's shortcut file](https://github.com/apache/zeppelin/blob/526eb15b8732909ab994adba76e0c3e33bfa001c/zeppelin-web/src/app/notebook/shortcut.html)

### What type of PR is it?
Bug Fix
Feature
Refactoring

### Todos
* [ ] TBD

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6203

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/action-bar/action-bar.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/action-bar/action-bar.component.ts
@@ -30,6 +30,7 @@ import { Note, OP, RevisionListItem } from '@zeppelin/sdk';
 import { MessageService, NoteActionService, NoteStatusService, SaveAsService, TicketService } from '@zeppelin/services';
 
 import { NoteCreateComponent } from '@zeppelin/share/note-create/note-create.component';
+import { ShortcutComponent } from '@zeppelin/share/shortcut/shortcut.component';
 
 @Component({
   selector: 'zeppelin-notebook-action-bar',
@@ -237,7 +238,11 @@ export class NotebookActionBarComponent extends MessageListenersManager implemen
   }
 
   showShortCut() {
-    // TODO(hsuanxyz)
+    this.nzModalService.info({
+      nzTitle: `Shortcut Info`,
+      nzWidth: '600px',
+      nzContent: ShortcutComponent
+    });
   }
 
   togglePermissions() {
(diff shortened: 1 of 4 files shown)


jongyoul:
@dididy Can we add some tests for it?

voidmatcha:
> @dididy Can we add some tests for it?

Are you referring to E2E test?

jongyoul:
Yes, it would be great if we have.

voidmatcha:
87c4d19
I added the missing `isMac` variable to detect macOS. This makes the modifier key label display as `Option` on macOS and `Alt` on other platforms.

6e75ed6
In the Classic UI, auto-completion triggered by `Control + .` had to be manually activated, but in the Monaco editor used in the NewUI, it is always enabled by default. Therefore, the shortcut is unnecessary and has been removed.

https://issues.apache.org/jira/browse/ZEPPELIN-6314

I’m currently looking into the E2E environment setup, and the Jira ticket is as above. It will likely use Playwright.

#5044 / #5064
Once the subsequent PRs are merged, I’ll review again, and this PR can be merged afterward.

zeppelin-web/src/app/notebook/shortcut.html
tbonelee, line 1: Are you intending to remove the auto-completion shortcut info only from the new UI? If so, this should be reverted, as it removes the shortcut information from the classic UI.

voidmatcha: 79ca7fa / 2dd2a6c
I accidentally deleted the Classic UI version instead of the New UI one. I’ve reverted that and modified it to delete the New UI version instead. Thanks for checking it.

tbonelee (approved):
LGTM

tbonelee:

PR #4965, opened 2025-07-13
[ZEPPELIN-6237] Add serviceLocator shutdown when server is close

proceane:
### What is this PR for?
In the `zeppelin-server/rest` module, it was observed that when running multiple tests together, configuration values of notebooks used in previous tests remain.
To address this, code has been added to shutdown `serviceLocator` when the test finishes and the server is stopped.

### What type of PR is it?
Bug Fix

### Todos
* Add serviceLocatior shutdown

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6237

* In the `zeppelin-server/rest` module, run test.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
@@ -364,6 +364,7 @@ public class ZeppelinServer implements AutoCloseable {
         if (sharedServiceLocator != null) {
           if (!zConf.isRecoveryEnabled()) {
             sharedServiceLocator.getService(InterpreterSettingManager.class).close();
+            sharedServiceLocator.shutdown();
           }
           sharedServiceLocator.getService(Notebook.class).close();
         }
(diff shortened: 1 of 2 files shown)


Reamer:
This pull request is a workaround for me, because the ServiceLocator should actually be executed when Zeppelin is shut down.
https://github.com/apache/zeppelin/blob/d7740440d1992606c577e1e0d5f884d3d4bb3286/zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java#L357-L377

Somewhere here should be `sharedServiceLocator.shutdown();`.

Unfortunately, my efforts failed because of the recovery tests.
Would you like to try a real shutdown here?

In my opinion, the `close()` method of the “services” must take the recovery mechanism into account and not at the top of the Zeppelin server.

zeppelin-server/src/main/java/org/apache/zeppelin/server/ZeppelinServer.java
```diff
@@ -364,6 +364,7 @@ public void shutdown(int exitCode) {
         if (sharedServiceLocator != null) {
           if (!zConf.isRecoveryEnabled()) {
             sharedServiceLocator.getService(InterpreterSettingManager.class).close();
+            sharedServiceLocator.shutdown();
```
Reamer, line 367: Yes, this is the simple variant, but it actually belongs outside the ‘if’ statement.

PR #4976, opened 2025-07-15
[MINOR] Fix flag parsing bug in FileInterpreter

renechoi:
# [MINOR] Fix flag parsing bug in FileInterpreter

## What is this PR for?
This PR fixes a bug in the `FileInterpreter` class where the dash character (`-`) was incorrectly included as a flag when parsing command arguments.

## What type of PR is it?
Bug Fix

## What is the Jira issue?
N/A (Minor fix)

## How should this be tested?
1. Run the new unit tests: `FileInterpreterTest.java`
2. Verify that flag parsing works correctly:
   - Command: `ls -l` should only have flag `'l'`, not `'-'` and `'l'`
   - Command: `ls -lah` should have flags `'l'`, `'a'`, `'h'`, without `'-'`

## Screenshots (if appropriate)
N/A

## Questions:
* Does the license files need update? No

### Problem
The current implementation of `parseArg()` method in `FileInterpreter.java` incorrectly includes the dash character as a flag:

```java
// Before (line 73)
for (int i = 0; i < arg.length(); i++) {
```

When parsing `-l`, this adds both `'-'` and `'l'` to the flags HashSet.

### Solution
Start the loop from index 1 to skip the dash character:

```java
// After
for (int i = 1; i < arg.length(); i++) {
```

### Impact Analysis
- **Current behavior is not affected** because the existing code (`HDFSFileInterpreter`) only checks for specific flag characters like `'l'` and `'h'`
- However, this is a **correctness issue** that could cause problems in future implementations
- The fix ensures that only actual flag characters are stored in the flags set

### Test Coverage
Added comprehensive unit tests in `FileInterpreterTest.java` to verify:
- Single flags (`-l`)
- Multiple flags (`-la`, `-lah`)
- Separate flags (`-l -h`)
- Edge cases (empty flag `-`)
- Verification that dash is not included in the flags set

### DIFF
--- a/file/src/main/java/org/apache/zeppelin/file/FileInterpreter.java
+++ b/file/src/main/java/org/apache/zeppelin/file/FileInterpreter.java
@@ -70,7 +70,7 @@ public abstract class FileInterpreter extends Interpreter {
 
     private void parseArg(String arg) {
       if (arg.charAt(0) == '-') {                   // handle flags
-        for (int i = 0; i < arg.length(); i++) {
+        for (int i = 1; i < arg.length(); i++) {
           Character c = arg.charAt(i);
           flags.add(c);
         }
--- a/file/src/test/java/org/apache/zeppelin/file/FileInterpreterTest.java
+++ b/file/src/test/java/org/apache/zeppelin/file/FileInterpreterTest.java
@@ -1,0 +1,181 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


Reamer (changes requested):
LGTM, but please create a JIRA ticket for this.

renechoi:
@Reamer  Thank you for the review! I've created a JIRA ticket as requested: [ZEPPELIN-6247](https://issues.apache.org/jira/browse/ZEPPELIN-6247)

I'll update the PR title to include the JIRA issue number.

Reamer:
@renechoi Two Checkstyle errors
```
[INFO] --- checkstyle:3.4.0:check (checkstyle-fail-build) @ zeppelin-file ---
[INFO] There are 2 errors reported by Checkstyle 9.3 with zeppelin/checkstyle.xml ruleset.
Error:  src/test/java/org/apache/zeppelin/file/FileInterpreterTest.java:[1] (misc) NewlineAtEndOfFile: File does not end with a newline.
Error:  src/test/java/org/apache/zeppelin/file/FileInterpreterTest.java:[39,5] (modifier) RedundantModifier: Redundant 'public' modifier.
```

renechoi:
@Reamer  Thank you for catching those checkstyle errors! I've fixed both issues:

1. Added the missing newline at the end of FileInterpreterTest.java
2. Removed the redundant 'public' modifier from the TestFileInterpreter constructor

The changes have been applied and the code should now pass all checkstyle checks.

Reamer:
Let's wait for CI.

Reamer:

PR #4979, opened 2025-07-16
[ZEPPELIN-6246] Replace deprecated Character constructor

hyunw9:
### What is this PR for?

In the following file [zeppelin/file/src/main/java/org/apache/zeppelin/file/HDFSFileInterpreter.java](https://github.com/apache/zeppelin/blob/91f091e287b94295ec7f10254b4f4ac800209899/file/src/main/java/org/apache/zeppelin/file/HDFSFileInterpreter.java#L180)

The Character constructor is deprecated in Java 9 and later, recommending the use of the static factory method Character.valueOf() instead.

This change's objective is to update the code to use the modern recommended approach.

As-is
```
 if (args.flags.contains(new Character('l'))) {
```
To-be
```
if (args.flags.contains(Character.valueOf('l'))) {
```

Also, Changed single word's type from **""** -> **''**

### What type of PR is it?
Refactoring

### Todos
* [ ] - Replace deprecated Character constructor
* [ ] - Change single word's data type 

### What is the Jira issue?
* Open an issue on [Jira](https://issues.apache.org/jira/browse/ZEPPELIN-6246)

### Questions:

### DIFF
--- a/file/src/main/java/org/apache/zeppelin/file/HDFSFileInterpreter.java
+++ b/file/src/main/java/org/apache/zeppelin/file/HDFSFileInterpreter.java
@@ -177,13 +177,13 @@ public class HDFSFileInterpreter extends FileInterpreter {
   }
 
   private String listOne(String path, OneFileStatus fs) {
-    if (args.flags.contains(new Character('l'))) {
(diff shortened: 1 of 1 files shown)


Reamer:

PR #4980, opened 2025-07-16
ZEPPELIN-6200 Update docs URL to /docs/latest

gurioh:
### What is this PR for?

The current official documentation link is set as:

http://zeppelin.apache.org/docs/{{zeppelinVersion}}/index.html

However, the current version is 0.12.0-SNAPSHOT, which is still under development and not an official release. As a result, the documentation link for this version is not available.

Since we only need to provide a link to the latest stable release, how about updating the link to:

http://zeppelin.apache.org/docs/latest/index.html

This would ensure users are always directed to the latest available documentation.

### What type of PR is it?
Improvement

### Todos
* [o] - Verified that the updated URL

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/ZEPPELIN-6200

### Questions:
* Does the license files need to update? no

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/home/home.component.html
+++ b/zeppelin-web-angular/src/app/pages/workspace/home/home.component.html
@@ -26,7 +26,7 @@
     <div nz-col [nzSm]="24" [nzLg]="12">
       <h3>Help</h3>
       Get started with <a style="text-decoration: none;" target="_blank"
-                          href="http://zeppelin.apache.org/docs/{{ticketService.version}}/index.html">Zeppelin
+                          href="http://zeppelin.apache.org/docs/latest/index.html">Zeppelin
       documentation</a><br/>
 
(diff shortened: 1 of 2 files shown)


tbonelee:
Thanks for the update!

I just wanted to check something. As far as I know, the `zeppelinVersion` is replaced with the actual version number during the release process, so it ends up pointing to the correct version-specific documentation.

If we change the link to always point to `latest` documentation page, I'm a bit concerned that users of older Zeppelin versions might get redirected to a newer version' documentation, which could be different from what they're using.

I'm not completely sure, so just raising the point. 
@jongyoul Curious to hear your thoughts too.

ParkGyeongTae:
How about a compromise: we keep using {{zeppelinVersion}} for released versions to ensure users get version-specific documentation, but fall back to latest only when the version is a SNAPSHOT?

PR #4987, opened 2025-07-19
[ZEPPELIN-6234] fix: table() function not working

kmularise:
### What is this PR for?
This PR introduces the ensureTableFunction() method to enhance compatibility with mongosh (MongoDB Shell) in the Apache Zeppelin Mongo interpreter context. Previously, the .table() function was attached to DBQuery.prototype, which may not be sufficient in newer MongoDB environments where command cursors or altered prototypes are used.
The new method safely injects the .table() utility function into relevant cursor prototypes at runtime, ensuring table rendering works as expected even under mongosh.

Additionally, the changes:

- Introduce safer prototype extension using a dummy cursor check.
- Preserve the original behavior for backwards compatibility.
- Avoid redundant registration in unsupported environments.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Add ensureTableFunction() for dynamic prototype injection

### What is the Jira issue?
* [ZEPPELIN-6234](https://issues.apache.org/jira/browse/ZEPPELIN-6234): MongoDB interpreter: .table() function not working

1. Use Apache Zeppelin with the MongoDB interpreter.
2. Connect to a MongoDB 8.x instance using mongosh.
3. Execute a standard query with .table():
```js
db.collection.find().table()
```

4. Verify:
- `%table` format renders properly in Zeppelin UI.
- No registration or prototype errors occur in shell.
- Legacy Mongo Shells continue to function as before.

Pasted Graphic

### Questions:
* Is there breaking changes for older versions? No. The solution maintains backward compatibility.
* Does this needs documentation? Optional. Could be mentioned in Mongo interpreter tips or mongosh compatibility notes.

### DIFF
--- a/mongodb/src/main/resources/shell_extension.js
+++ b/mongodb/src/main/resources/shell_extension.js
@@ -104,6 +104,26 @@ function printTable(dbquery, fields, flattenArray) {
     printTable(this, fields, flattenArray);
 };
 
+function ensureTableFunction() {
+    var tableFunc = function(fields, flattenArray) {
+        printTable(this, fields, flattenArray);
+        return this;
+    };
+
+    try {
+        var sampleCursor = db.getCollection('__dummy__').find({}).limit(0);
+        var proto = Object.getPrototypeOf(sampleCursor);
+
+        if (proto && !proto.table) {
+            proto.table = tableFunc;
+        }
+    } catch (e) {
+        print("Registration Error: prototype registration failed:", e.message);
+    }
+}
+
+ensureTableFunction();
+
 if (globalThis.DBCommandCursor)
     (DBCommandCursor.prototype || DBCommandCursor).table = (DBQuery.prototype || DBQuery).table;
 


tbonelee (dismissed):
LGTM. Ran some sample queries with MongoDB interpreter, and everything worked as expected

mongodb/src/main/resources/shell_extension.js
```diff
@@ -104,6 +104,26 @@ function printTable(dbquery, fields, flattenArray) {
...
+        printTable(this, fields, flattenArray);
+        return this;
+    };
+
+    try {
+        var sampleCursor = db.getCollection('__dummy__').find({}).limit(0);
+        var proto = Object.getPrototypeOf(sampleCursor);
+
+        if (proto && !proto.table) {
+            proto.table = tableFunc;
```
seung-00, line 124: Shouldn't we apply `tableLimit` here as well, like in line 102?

mongodb/src/main/resources/shell_extension.js
```diff
@@ -104,6 +104,26 @@ function printTable(dbquery, fields, flattenArray) {
...
+        }
+    } catch (e) {
+        print("Registration Error: prototype registration failed:", e.message);
+    }
+}
```
seung-00, line 129: What do you think about merging the existing table prototype logic into `ensureTableFunction` to avoid duplication?

mongodb/src/main/resources/shell_extension.js
kmularise: Thanks for the review! I missed applying the `tableLimit` check in the `tableFunc`. I'll update the code to include the tableLimit check.

mongodb/src/main/resources/shell_extension.js
kmularise: I think you're right to avoid duplication. Looking at this again, there's potential for duplicate registration, and merging into `ensureTableFunction` makes more sense. I'll move the existing table prototype logic into `ensureTableFunction`.

Thanks for the feedback!

ParkGyeongTae (approved):
👍 LGTM

ParkGyeongTae:

PR #4989, opened 2025-07-19
[ZEPPELIN-6252] Migrate frontend TypeScript configuration to `strict: true` for better type safety

tbonelee:
### What is this PR for?
This PR enables `strict: true` in the frontend Typescript configuration (`tsconfig.json`) to enhance type safety and catch potential issues at compile time.

During this migration:
- Existing logic was preserved as much as possible.
- `non-null assertion` was avoided where possible. In cases whre Angular lifecycle constraints (e.g., `ngOnInit, @Input`) made it unavoidable, it was used with caution.
- Optional properties introduced due to potential `undefined` values now throw intended errors when accessed unsafely, making error sources more explicit and enable type narrowing.
- `any` type was avoided where possible. In unavoidable cases, it was used with `tslint:disable` or `@ts-ignore` comments.

This improves type robustness and aligns with modern TypeScript best practices, while minimizing risk to existing functionality.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6252

- Ensure the frontend builds successfully.
- Manual verification of major UI flows.

### Questions:
* Is there breaking changes for older versions? No breaking changes expected

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/interpreter/item/item.component.html
+++ b/zeppelin-web-angular/src/app/pages/workspace/interpreter/item/item.component.html
@@ -135,17 +135,6 @@
       <i nz-icon nzType="down"></i>
     </a>
     <nz-dropdown-menu #perUserMenu="nzDropdownMenu">
-      <a nz-dropdown>
-        <span *ngIf="interpreterRunningOption === runningOptionMap.perUserModeName; else perNote">
-          {{ optionFormGroup.get('perUser').value }}
-        </span>
-        <ng-template #perNote>
-          <span>
-            {{ optionFormGroup.get('perNote').value }}
-          </span>
-        </ng-template>
-        <i nz-icon nzType="down"></i>
-      </a>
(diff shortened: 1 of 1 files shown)


Reamer:
I think this is a good change, but I can't review it because I don't know anything about TypeScript and Angular.

ping @jongyoul

tbonelee:
@jongyoul Thank you for carefully reviewing this long PR, I really appreciate it.

I assume you're referring to the changes where we pass generated objects as parameters instead of reading the fields directly (e.g., `renderBefore(config, chart)`, `refreshSetting(config)`, `renderAfter(config)`).
The point is that by passing the generated value, TypeScript can infer it's already defined at the call site, which removes the need for repetitive null/undefined checks. If we had relied on member fields instead, we would have needed explicit type guards to narrow them to non-null/undefined.

If this wasn't the part you meant, please let me know.

jongyoul:
@tbonelee Thank you for the explanation. That's exactly what I want to know. 👍

PR #4993, opened 2025-07-20
[ZEPPELIN-6253] fix: remove SNAPSHOT from version when building crede…

hyo-ryeong:
### What is this PR for?

Fixes a broken "Learn more" link on the credential page.

Zeppelin currently appends `-SNAPSHOT` to the version when constructing the documentation link, which results in a 404 page.  
This PR removes `-SNAPSHOT` from the version and ensures it uses a valid documentation path.

Broken link (before):  
`https://zeppelin.apache.org/docs/0.12.0-SNAPSHOT/setup/security/datasource_authorization.html`

Corrected link (after):  
`https://zeppelin.apache.org/docs/0.12.0/setup/security/datasource_authorization.html`

---

### What type of PR is it?

- [x] Bug Fix

---

### What is the Jira issue?

[JIRA: ZEPPELIN-6253](https://issues.apache.org/jira/browse/ZEPPELIN-6253)

---

1. Navigate to **Credential** page in Zeppelin.
2. Click the **"?" (Learn more)** link at the top right.
3. Confirm it opens a valid docs page without 404.

---

### Questions

- Does the license files need to update? → No  
- Is there breaking changes for older versions? → No  
- Does this need documentation? → No

### DIFF
--- a/zeppelin-web/src/app/credential/credential.controller.js
+++ b/zeppelin-web/src/app/credential/credential.controller.js
@@ -199,14 +199,11 @@ function CredentialController($scope, $rootScope, $http, baseUrlSrv, ngToast) {
   }
 
   $scope.getCredentialDocsLink = function() {
-    let currentVersion = $rootScope.zeppelinVersion || '0.12.0';
-    currentVersion = currentVersion.replace('-SNAPSHOT', '');
+    const currentVersion = $rootScope.zeppelinVersion && $rootScope.zeppelinVersion.replace('-SNAPSHOT', '') || 'latest';
 
     const isVersionOver0Point7 = currentVersion.split('.')[1] > 7;
 
-    /*
-     * Add '/setup' to doc link on the version over 0.7.0, add '/setup' to the doc URL
-     */
+    // Add '/setup' to doc URL for versions > 0.7.0
     return `https://zeppelin.apache.org/docs/${currentVersion}${
       isVersionOver0Point7 ? '/setup' : ''
     }/security/datasource_authorization.html`;
--- a/zeppelin-web/src/app/credential/credential.controller.js
+++ b/zeppelin-web/src/app/credential/credential.controller.js
@@ -199,10 +199,13 @@ function CredentialController($scope, $rootScope, $http, baseUrlSrv, ngToast) {
   }
 
   $scope.getCredentialDocsLink = function() {
-    const currentVersion = $rootScope.zeppelinVersion;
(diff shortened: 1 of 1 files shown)


zeppelin-web/src/app/credential/credential.controller.js
```diff
@@ -199,10 +199,13 @@ function CredentialController($scope, $rootScope, $http, baseUrlSrv, ngToast) {
   }
 
   $scope.getCredentialDocsLink = function() {
-    const currentVersion = $rootScope.zeppelinVersion;
-    const isVersionOver0Point7 = currentVersion && currentVersion.split('.')[1] > 7;
+    let currentVersion = $rootScope.zeppelinVersion || '0.12.0';
+    currentVersion = currentVersion.replace('-SNAPSHOT', '');
```
tbonelee: How about replacing the `-SNAPSHOT` part if present, and falling back to `'latest'` instead?
Also, I think it's preferrable to avoid  hardcoding version strings like `'0.12.0'`.
```suggestion
    const currentVersion = $rootScope.zeppelinVersion
      && $rootScope.zeppelinVersion.replace('-SNAPSHOT', '') || 'latest';
```

zeppelin-web/src/app/credential/credential.controller.js
```diff
@@ -199,10 +199,13 @@ function CredentialController($scope, $rootScope, $http, baseUrlSrv, ngToast) {
...
+
+    const isVersionOver0Point7 = currentVersion.split('.')[1] > 7;
+
     /*
-     * Add '/setup' to doc link on the version over 0.7.0
+     * Add '/setup' to doc link on the version over 0.7.0, add '/setup' to the doc URL
```
tbonelee: The add part seems to repeat the same meaning as the original comment.

tbonelee (commented):
Thanks for fixing the broken doc URL.
Just to clarify, even after removing `-SNAPSHOT`, the docs page for the current development version might still be missing, since the latest version is not released yet. So the link could still be broken unless it falls back to a stable version like `latest`.
We might also consider a unified approach in #4980 if needed.

zeppelin-web/src/app/credential/credential.controller.js
hyo-ryeong: Thanks for the review!
I updated the code to remove the hardcoded default version and added a fallback to 'latest' when zeppelinVersion is not available.
Also simplified the comment to avoid duplication.

ParkGyeongTae (commented):
I think the PR title might be unfinished — could you please review it and update if needed?

PR #4995, opened 2025-07-20
[MINOR] Bump up version to 0.13.0-SNAPSHOT

jongyoul:
### What is this PR for?
Change the version to 0.13.0-SNAPSHOT

### What type of PR is it?
Improvement

### Todos
* [x] - Change versions to 0.13.0-SNAPSHOT

The version should show 0.13.0-SNAPSHOT when building it.

### Questions:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,4 @@ tramp
 
 # python venv
 .venv
+zeppelin-zengine/conf/
--- a/.gitignore
+++ b/.gitignore
@@ -145,4 +145,4 @@ tramp
 
 # python venv
 .venv
-zeppelin-zengine/conf/
+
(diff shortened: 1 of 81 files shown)


pan3793:
the next version is 0.13 or 1.0?

Reamer:
> the next version is 0.13 or 1.0?

Based on the snapshot version 0.13.0

jongyoul:
I'm also thinking about the next version, but I just followed adding one minor version in this PR. We can start to discuss whether the next version could be 1.0 or 0.13 :-) Then, we can make a release branch for it.

.gitignore
```diff
@@ -145,3 +145,4 @@ tramp
 
 # python venv
 .venv
+zeppelin-zengine/conf/
```
Reamer: Is this change necessary?

Reamer (dismissed):
I only found one small thing.

.gitignore
jongyoul: Oh.. it's not. I will remove it.

jongyoul: @Reamer Removed it. 🙏

Reamer (approved):

PR #4998, opened 2025-07-24
[ZEPPELIN-6258] Improve Process resource management in SparkInterpreterLauncher

renechoi:
### What is this PR for?
`detectSparkScalaVersion` calls `waitFor()` before reading spark-submit's output and never drains stdout, so the process blocks once it fills a pipe buffer. There is also no timeout, exit check or cleanup.

This merges stderr into stdout, drains it on a reader thread bounded by 30s, then waits, checks the exit value and destroys a live process.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6258

`SparkInterpreterLauncherTest#testDetectSparkScalaVersionWithLargeProcessOutput` runs a fake spark-submit writing 256KB to stdout. It hangs on master, passes here.

### Questions:
* Is there breaking changes for older versions? A `spark-submit --version` over 30s now fails rather than blocking.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -20,6 +20,7 @@ package org.apache.zeppelin.interpreter.launcher;
 import java.io.FileInputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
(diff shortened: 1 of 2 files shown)


ParkGyeongTae:
Could you please rebase this branch onto `master`?

renechoi:
Sorry for the long delay, and rebased now. #5000 landed in the meantime and rewrote `detectSparkScalaVersion`, so replaying the old diff would have undone that; this is the same fix written against the current code.

Added `testDetectSparkScalaVersionWithLargeProcessOutput`, which points SPARK_HOME at a fake spark-submit writing 256KB to stdout. It hangs on master and passes here.

PR #5006, opened 2025-08-02
[ZEPPELIN-6265] Support classic Helium viz packages in new UI

tbonelee:
### What is this PR for?
This PR enables the use of external Helium visualization packages, which are previously available only in the classic UI, in the new UI.

The process of loading and importing Helium packages, as well as rendering them, closely follows the logic from the classic UI.

Since the new-style visualization classes in the new UI are not compatible with the existing ones, the implementation branches accordingly to handle both.

Classic Helium visualization packages depend on legacy technologies such as AngularJS, jQuery, and several visualization libraries, which had to be included as a result.
For Bootstrap styles, conflicts arose with existing styles, so I copied and modified the original HTML templates to prevent clashes.
If a Helium package attempts to use the original HTML template, it will be replaced with the modified version.

Additionally, I identified that the Helium visualization support classes from the discontinued development of the new UI are no longer in use, so I removed them.

### What type of PR is it?

Improvement

### Todos
* [v] - Added support for loading Helium packages
* [v] - Fixed style mismatches in Helium packages
* [v] - Other functional issues resolved

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6265

* Enables a classic Helium visualization package in the classic UI, and verify that it works in the new UI.
  - Note: The `/helium` package management page has not yet been added.

### Questions:


zeppelin-web-angular/src/app/interfaces/helium.ts
```diff
@@ -0,0 +1,28 @@
+import { HeliumPackageType } from '@zeppelin/helium';
```
ParkGyeongTae, line 13: Could you please add the Apache License header at the top of this file? This ensures compliance with ASF requirements. Thanks!

tbonelee: Thank you for checking! I've fixed it with other Rat-check violations.

zeppelin-web-angular/src/app/interfaces/helium.ts
```diff
@@ -10,4 +10,31 @@
...
+  enabled: boolean;
+}
+
+export interface HeliumBundle {
+  id: string;
+  name: string;
+  icon: string;
+  type: HeliumType;
+  // tslint:disable-next-line:no-any
+  class: any;
```
seung-00: How about use `unknown` (with type narrowing)?

zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
```diff
@@ -0,0 +1,328 @@
...
+import * as angular from 'angular';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AngularDragDropService {
+  constructor() {}
+
+  // tslint:disable-next-line:no-any
+  addDragDropDirectives(module: any): void {
```
seung-00: How about use `IModule` type from `@types/angular`?

zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
```diff
@@ -0,0 +1,328 @@
...
+    // jqyoui-draggable directive
+    // tslint:disable-next-line:no-any
+    module.directive('jqyouiDraggable', [
+      'customDragDropService',
+      // tslint:disable-next-line:no-any
+      function(dragDropService: any) {
+        return {
+          restrict: 'A',
+          // tslint:disable-next-line:no-any
+          link: function(scope: any, element: any, attrs: any) {
```
seung-00: How about removing the explicit `any`? Type inference would work correctly here.

tbonelee:
@seung-00 
Replaced `any` types with proper interfaces for classic visualizations. 

Added new interfaces instead of reusing the Visualization class from zeppelin-vis, since Helium packages are eval’d at runtime and their prototype chains don’t match. Interfaces were preferred over classes because only the shape is shared across eval’d definitions.

zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
```diff
@@ -0,0 +1,326 @@
...
+export class AngularDragDropService {
+  constructor() {}
+
+  addDragDropDirectives(module: angular.IModule): void {
+    // Drag and drop service to maintain state across directives
+    // tslint:disable-next-line:no-any
+    module.service('customDragDropService', [
+      '$parse',
+      // tslint:disable-next-line:no-any
+      function($parse: any) {
```
seung-00: Would it be hard to use the types provided by @types/angular here as well?

And I'm not fully sure, but if the TS compiler can infer the type in some places without explicitly declaring it, it might be nice to rely on inference.

zeppelin-web-angular/src/app/services/table-data-adapter.service.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.comment = updatedClassicData.comment;
+      }.bind(this),
+
+      // Refresh data from modern TableData
+      refresh: function() {
+        const updatedClassicData = this.convertToClassicFormat(modernTableData);
+        this.columns = updatedClassicData.columns;
+        this.rows = updatedClassicData.rows;
+        this.comment = updatedClassicData.comment;
+      }.bind(this)
```
seung-00: How about using an arrow function instead of `bind`? I think it's more readable and more modern way.

zeppelin-web-angular/src/app/services/table-data-adapter.service.ts
```diff
@@ -0,0 +1,115 @@
...
+      // tslint:disable-next-line:no-any
+      loadParagraphResult: function(paragraphResult: any) {
+        // Delegate to modern TableData's method
+        modernTableData.loadParagraphResult(paragraphResult);
+
+        // Update proxy data after loading
+        const updatedClassicData = this.convertToClassicFormat(modernTableData);
+        this.columns = updatedClassicData.columns;
+        this.rows = updatedClassicData.rows;
+        this.comment = updatedClassicData.comment;
```
seung-00: I'm not sure, but it looks like the proxy was supposed to be updated. However, it seems that the `TableDataAdapterService` instance, which is bound, is being updated instead.

zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
tbonelee: Thanks for checking! I'll update it to utilize TS type inference.

zeppelin-web-angular/src/app/services/table-data-adapter.service.ts
tbonelee: You're right, the `this` binding was incorrect here. So I changed it to use arrow functions and access through variable names instead.

tbonelee:
@seung-00 I've checked for places where type inference can be used as suggested in the review, and fixed the incorrect `this` bindings. I've also added proper type definitions to replace `any` usage in several places.

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
```diff
@@ -359,11 +359,11 @@ private void prepareSource(HeliumPackage pkg, String[] moduleNameVersion,
         .append(" from \"../" + mainFileName + "\"\n");
 
     loadJsRegister.append(HELIUM_BUNDLES_VAR + ".push({\n");
-    loadJsRegister.append("id: \"" + moduleNameVersion[0] + "\",\n");
-    loadJsRegister.append("name: \"" + pkg.getName() + "\",\n");
-    loadJsRegister.append("icon: " + gson.toJson(pkg.getIcon()) + ",\n");
-    loadJsRegister.append("type: \"" + pkg.getType() + "\",\n");
-    loadJsRegister.append("class: " + className + "\n");
+    loadJsRegister.append("  id: \"" + moduleNameVersion[0] + "\",\n");
```
Reamer: My suggestion is to use the following style so that no string concatenation is performed.
```suggestion
    loadJsRegister.append(idention).append("id: \"").append(moduleNameVersion[0]).apend("\",\n");
```

Of course, the idention variable must be defined before, but it also makes it easier to change it globally. It also makes the code more readable, explaining why two spaces are inserted.  I don't think these are necessary from a purely programmatic point of view, do they?

Reamer (commented):
A minor comment on the backend code. I am unable to review the frontend part.

zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java
tbonelee: Thanks for pointing that out. I've removed the indentation (as it didn't really improve readability) and replaced the string concatenation with `.append()`, following your suggestion.

tbonelee:
Rebased onto the master branch and add some Prttier exclusion rules for `*.min.css` and `*.min.js` files

ParkGyeongTae:
@tbonelee Please rebase onto the latest master to resolve the conflicts.

tbonelee:
Rebased it onto the master branch

tbonelee:
Thanks for the reviews, I've merged this into master and branch-0.12

dididy:
It looks like zeppelin-angular-web/src/app/helium-manager/public-api.ts is an empty file and is missing the license header comment, or it might not have been removed.

PR #5009, opened 2025-08-03
[ZEPPELIN-6267] Refactor JarHelper Class to Use a Logging Framework Instead of System.out.println()

hyunw9:
### What is this PR for?

Refactor JarHelper Class to Use a Logging Framework Instead of `System.out.println()`

### What type of PR is it?
Refactoring

### Todos
* [x] - Replace `println()` to `LOGGER.info()`

### What is the Jira issue?
* [here](https://issues.apache.org/jira/browse/ZEPPELIN-6267)

### Questions:
* Does the license files need to update? - no
* Is there breaking changes for older versions? - no

### DIFF
--- a/flink/flink-scala-2.12/src/main/java/org/apache/zeppelin/flink/internal/JarHelper.java
+++ b/flink/flink-scala-2.12/src/main/java/org/apache/zeppelin/flink/internal/JarHelper.java
@@ -29,7 +29,6 @@ import java.util.jar.JarEntry;
 import java.util.jar.JarInputStream;
 import java.util.jar.JarOutputStream;
 
-import org.apache.zeppelin.flink.Flink117Shims;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


flink/flink-scala-2.12/src/main/java/org/apache/zeppelin/flink/internal/JarHelper.java
```diff
@@ -29,6 +29,10 @@
 import java.util.jar.JarInputStream;
 import java.util.jar.JarOutputStream;
 
+import org.apache.zeppelin.flink.Flink117Shims;
```
Reamer: I see no reason for this import.

Reamer (changes requested):
Looks good, just a potentially incorrect import.

flink/flink-scala-2.12/src/main/java/org/apache/zeppelin/flink/internal/JarHelper.java
hyunw9: I just fixed it. Thanks for notice :)

PR #5010, opened 2025-08-04
[ZEPPELIN-6268] Fix resource leaks and add null checks for getResourceAsStream

renechoi:
## What is this PR for?
`getResourceAsStream()` in the Python interpreters was never closed, and a missing resource surfaced as an NPE inside `IOUtils`. All four sites now use try-with-resources and name the resource in the `IOException`: three in `IPythonInterpreter`, one in `PythonInterpreter.copyResourceToPythonWorkDir`.

The first revision also touched `alluxio` and `rlang`; master dropped both modules (ZEPPELIN-6420, ZEPPELIN-6430), so the diff is python-only.

## What is the Jira issue?
[ZEPPELIN-6268](https://issues.apache.org/jira/browse/ZEPPELIN-6268)

## How should this be tested?
`interpreter-test-jupyter-python (3.9, 11)` runs the module on 348bc8d: 38 tests, 0 failures, including the kernel-launching `IPythonInterpreterTest` (9) and `PythonInterpreterTest` (6). No new test: a closed stream has no observable seam, and the `IOException` needs a missing packaged resource.

### DIFF
--- a/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
+++ b/python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
@@ -132,9 +132,14 @@ public class IPythonInterpreter extends JupyterKernelInterpreter {
   }
 
   private void initPythonInterpreter(String gatewayHost, int gatewayPort) throws IOException {
-    InputStream input =
-            getClass().getClassLoader().getResourceAsStream("python/zeppelin_ipython.py");
-    List<String> lines = IOUtils.readLines(input, StandardCharsets.UTF_8);
+    List<String> lines;
+    try (InputStream input =
+            getClass().getClassLoader().getResourceAsStream("python/zeppelin_ipython.py")) {
+      if (input == null) {
+        throw new IOException("Cannot find resource: python/zeppelin_ipython.py");
+      }
+      lines = IOUtils.readLines(input, StandardCharsets.UTF_8);
+    }
     ExecuteResponse response = jupyterKernelClient.block_execute(ExecuteRequest.newBuilder()
             .setCode(StringUtils.join(lines, System.lineSeparator())
                     .replace("${JVM_GATEWAY_PORT}", gatewayPort + "")
@@ -143,9 +148,13 @@ public class IPythonInterpreter extends JupyterKernelInterpreter {
       throw new IOException("Fail to setup JVMGateway\n" + response.getOutput());
     }
 
-    input =
-            getClass().getClassLoader().getResourceAsStream("python/zeppelin_context.py");
-    lines = IOUtils.readLines(input, StandardCharsets.UTF_8);
+    try (InputStream input =
+            getClass().getClassLoader().getResourceAsStream("python/zeppelin_context.py")) {
+      if (input == null) {
+        throw new IOException("Cannot find resource: python/zeppelin_context.py");
+      }
+      lines = IOUtils.readLines(input, StandardCharsets.UTF_8);
+    }
     response = jupyterKernelClient.block_execute(ExecuteRequest.newBuilder()
             .setCode(StringUtils.join(lines, System.lineSeparator())).build());
     if (response.getStatus() != ExecuteStatus.SUCCESS) {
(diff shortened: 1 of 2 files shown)


alluxio/src/test/java/org/apache/zeppelin/alluxio/AlluxioInterpreterTest.java
```diff
@@ -224,9 +224,12 @@ void mkdirTest() throws IOException, AlluxioException {
   private void fileReadTest(String fileName, int size) throws IOException {
     File testFile = new File(PathUtils.concatPath(mLocalAlluxioCluster.getAlluxioHome(), fileName));
     FileInputStream fis = new FileInputStream(testFile);
-    byte[] read = new byte[size];
-    fis.read(read);
-    fis.close();
-    assertTrue(BufferUtils.equalIncreasingByteArray(size, read));
+    try {
```
Reamer: Please use try-with-resources. This is more elegant.

python/src/main/java/org/apache/zeppelin/python/PythonInterpreter.java
```diff
@@ -196,15 +197,21 @@ protected boolean useIPython() {
   private void copyResourceToPythonWorkDir(String srcResourceName,
                                            String dstFileName) throws IOException {
     FileOutputStream out = null;
+    InputStream in = null;
     try {
```
Reamer: Please use try-with-resources, also here

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -134,6 +134,9 @@ private void setupJVMGateway(String gatewayHost, int gatewayPort) throws IOExcep
   private void initPythonInterpreter(String gatewayHost, int gatewayPort) throws IOException {
     InputStream input =
```
Reamer: Here, closing the input stream is missing; this can be done directly with try-with-resource.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -145,6 +148,9 @@ private void initPythonInterpreter(String gatewayHost, int gatewayPort) throws I
 
     input =
```
Reamer: Here, closing the input stream is missing; this can be done directly with try-with-resource.

python/src/main/java/org/apache/zeppelin/python/IPythonInterpreter.java
```diff
@@ -161,6 +167,9 @@ private void initPythonInterpreter(String gatewayHost, int gatewayPort) throws I
 
     if (additionalPythonInitFile != null) {
       input = getClass().getClassLoader().getResourceAsStream(additionalPythonInitFile);
```
Reamer: Here, closing the input stream is missing; this can be done directly with try-with-resource.

rlang/src/main/java/org/apache/zeppelin/r/IRInterpreter.java
```diff
@@ -129,6 +129,9 @@ protected void initIRKernel() throws IOException, InterpreterException {
     String timeout = getProperty("spark.r.backendConnectionTimeout", "6000");
     InputStream input =
```
Reamer: Same please use try-with-resource do close the stream.

rlang/src/main/java/org/apache/zeppelin/r/ZeppelinR.java
```diff
@@ -97,6 +97,9 @@ public void open() throws IOException, InterpreterException {
     try {
```
Reamer: Same use try-with-resource for the input and FileOutputStream

Reamer (changes requested):
As you can see, I am a big fan of try-with-resources.
This is a JDK8 feature and should be used because it saves a lot of boilerplate code.

renechoi:
Rebased onto current master and pushed 348bc8d. The try-with-resources requests that still apply are in: `PythonInterpreter.copyResourceToPythonWorkDir`, and the three `getResourceAsStream` sites in `IPythonInterpreter`, each with a null check on the stream.

The other three files you reviewed are no longer in the diff. `alluxio` was dropped by ZEPPELIN-6420 (1a6a8e3) and `rlang` by ZEPPELIN-6430 (0f0a2d5). The PR is now python-only: 2 files, +29/-17, 18/18 checks green.

PR #5013, opened 2025-08-06
[ZEPPELIN-6238] Prevent infinite loop in NotebookRestApiTest due to interpreter startup failure

gurioh:
### What is this PR for?
The testRunNoteNonBlocking integration test in NotebookRestApiTest can enter an infinite loop when the Python interpreter fails to start. This happens because the interpreter cannot find the python executable, resulting in an IOException: error=2, No such file or directory.

The test uses httpPost with blocking=true, followed by p1.waitUntilFinished() to wait for the paragraph to complete. When the interpreter fails to start, the paragraph's status remains in a READY state instead of transitioning to ERROR. This causes waitUntilFinished() to wait indefinitely, as isTerminated() never returns true.

The fix is to modify the test logic to immediately check the paragraph's status after p1.waitUntilFinished(). By moving the assertEquals(Job.Status.FINISHED, p1.getStatus()) assertion to immediately after the wait call, the test will fail instantly if the paragraph's status is not FINISHED (e.g., ERROR). This prevents the test from proceeding to p2.waitUntilFinished() and avoids the infinite loop.

image

image

### What type of PR is it?
Bug Fix

### What is the Jira issue?
*[ZEPPELIN-6238] 

image

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
@@ -666,12 +666,12 @@ class NotebookRestApiTest extends AbstractTestRestApi {
           Paragraph p2 = note1.getParagraph(1);
           try {
             p1.waitUntilFinished();
+            assertEquals(Job.Status.FINISHED, p1.getStatus());
             p2.waitUntilFinished();
+            assertEquals(Job.Status.FINISHED, p2.getStatus());
           } catch (InterruptedException e) {
             fail(e);
           }
-          assertEquals(Job.Status.FINISHED, p1.getStatus());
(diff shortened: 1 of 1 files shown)


ParkGyeongTae:
Hello, thank you for your contribution!
Could you update the PR title to:
`[ZEPPELIN-6238] Prevent infinite loop in NotebookRestApiTest due to interpreter startup failure`
instead of the current one, `Prevent infinite loop in NotebookRestApiTest due to interpreter startup failure`?

gurioh:
I’ve updated it.
Thank you for the review.

Reamer:
The test is being tested by CI. I don't see any problems in the current master version.

ParkGyeongTae (commented):
I think this addition is useful, but I couldn’t reproduce the issue on my local machine. I’m using a MacBook Pro M4. Could you share what environment you were running when you saw this problem?

PR #5015, opened 2025-08-07
[ZEPPELIN-6269] Handle remove failures in NotebookRepoSync to prevent data inconsistency

hyunw9:
### What is this PR for?

If an `IOException` occurs while trying to remove a note from one of the repositories, the `for` loop would terminate immediately. This leaves the note in an inconsistent state where it is deleted from some repositories but not from others, leading to data inconsistency.

**Approach**

1.  The method will now attempt to remove the note from all configured repositories, even if an error occurs with one of them.
2.  Any repository that fails to remove the note is added to a `failedRepos` list.
3.  After iterating through all repositories, the method checks the `failedRepos` list. If the list is not empty, it throws a single `IOException` that includes information about all the repositories where the removal failed.

### What type of PR is it?
Refactoring

### Todos
* [x] - Refactor remove method
* [x] - Add test for modified method

### What is the Jira issue?
* [[ZEPPELIN-6269]](https://issues.apache.org/jira/browse/ZEPPELIN-6269)

* Added simple integration test in `NotebookRepoSyncTest`.

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -28,8 +28,6 @@ import org.apache.zeppelin.user.AuthenticationInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.gson.Gson;
-
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


jongyoul (dismissed):
LGTM. I have one minor comment. What do you think we can handle if some repos fail to remove notes? There might still not be in-sync status.

hyunw9:
@jongyoul It would be a nicer approach. Let me show you a draft. :)

hyunw9:
@jongyoul I just pushed modified version. How about this approach? if it fails, we can raise exception immediately, so we can avoid out-of-sync status.

PR #5016, opened 2025-08-07
[ZEPPELIN-6233] Upgrade Node.js version 18

celinayk:
### What is this PR for?
Upgrade Node.js version 18

### What type of PR is it?
Bug Fix

### 

* [x]  Upgrade Node.js to version 18.20
* [x] Upgrade npm to version 10.8.2
* [x] Upgrade Angular to version ~15.2.0
* [x] Upgrade tsc to version 4.9.5

### What is the Jira issue?
* [ZEPPELIN-6233](https://issues.apache.org/jira/browse/ZEPPELIN-6233)

### Questions:

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -86,32 +86,34 @@ jobs:
       - name: Run headless test
         run: xvfb-run --auto-servernum --server-args="-screen 0 1024x768x24" ./mvnw package -pl zeppelin-web-angular ${MAVEN_ARGS}
 
-  test-selenium-with-spark-module-for-spark-3-4:
+  test-selenium-with-spark-module-for-spark-3-5:
     runs-on: ubuntu-24.04
     defaults:
       run:
         shell: bash -l {0}
+    env:
+      ZEPPELIN_SELENIUM_BROWSER: firefox
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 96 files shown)


zeppelin-web-angular/angular.json
```diff
@@ -215,41 +217,43 @@
               "projects/zeppelin-helium/tsconfig.spec.json"
             ],
             "exclude": [
+              "**/node_modules/**",
               "**/node_modules/**"
```
ParkGyeongTae: Is there a reason for including `**/node_modules/**` twice in the exclude list? It seems redundant—one entry should be sufficient.

celinayk: Thank you for catching that! I've removed the duplicate entries from all projects in the angular.json file.

zeppelin-web-angular/.gitignore
```diff
@@ -44,4 +44,7 @@ Thumbs.db
 
 #
 .env
-yarn.lock
\ No newline at end of file
+yarn.lock
+
+# Angular cache
+.angular/cache
```
tbonelee, line 50: Please add a new line at the end of the file.

seung-00:
Could we also update the <node.version> property in root [pom.xml](https://github.com/apache/zeppelin/blob/master/pom.xml#L111)? It's used to set up the Node environment in CI.

PR #5022, opened 2025-08-10
[ZEPPELIN-6216] Show clear message when job manager is disabled instead of infinite loading

devyeony:
### What is this PR for?
This PR fixes the issue that occurs when `zeppelin.jobmanager.enable` is set to `false`. In this case, the server sends no response, causing the Job Manager page in both classic and new UIs to show an infinite loading indicator. This behavior confuses users and gives the impression that the page is broken.

To improve user experience, this PR introduces explicit server-side handling that returns an explicit "not allowed" response when the Job Manager is disabled via configuration. Additionally, both classic and new UIs have been updated to properly handle this response and display a user-friendly message: **"Job Manager is disabled in the current configuration."**

This change ensures users are informed about the disabled state instead of encountering an endless loading screen.

### What type of PR is it?
Improvement

### Todos
* [x] Implement server-side forbidden response when Job Manager is disabled
* [x] Add `JOB_MANAGER_DISABLED` WebSocket message and corresponding handling in both Classic and New UI
* [x] Add unit and integration tests covering the new behavior

**Note:** 
* Although REST API tests have been implemented, the current frontend (both Classic and New UI) interacts with the Job Manager primarily through WebSocket communication. Therefore, the functional verification and actual handling of the "Job Manager disabled" scenario have been focused on the WebSocket path. The REST API tests are prepared to ensure future compatibility if the REST endpoints are used.
* In the user-request handling method, a ForbiddenException triggers sending a clear error message to the client so the UI can inform the user that the Job Manager is disabled. However, during broadcast events where no direct client request is involved, the exception is logged at debug level and silently skipped to avoid unnecessary noise or client interruptions.

### What is the Jira issue?
[[ZEPPELIN-6216]](https://issues.apache.org/jira/browse/ZEPPELIN-6216)

* Automated tests have been added to cover the new behavior when the Job Manager is disabled:
  * Unit tests in `JobManagerServiceTest` verify that forbidden exceptions are thrown.
  * REST API tests in `NotebookRestApiTest` check that HTTP 403 responses with appropriate messages are returned.
  * WebSocket and server event handling tests in `NotebookServerTest` ensure no unexpected exceptions occur and proper notification messages are sent.
* Manual testing steps:
  1. Set `zeppelin.jobmanager.enable` to `false` via `ZeppelinConfiguration`.
  2. Access the Job Manager page in both classic and new UI.
  3. Confirm that instead of infinite loading, a clear message stating **"Job Manager is disabled in the current configuration."** is displayed.

* AS-IS (New UI)
AS-IS NEW UI

* AS-IS (Classic UI)
AS-IS CLASSIC UI

* TO-BE (New UI)
TO-BE NEW UI

* TO-BE (Classic UI)
TO-BE CLASSIC UI

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
@@ -59,6 +59,7 @@ import org.apache.zeppelin.service.AuthenticationService;
 import org.apache.zeppelin.service.JobManagerService;
 import org.apache.zeppelin.service.NotebookService;
 import org.apache.zeppelin.service.ServiceContext;
+import org.apache.zeppelin.service.exception.JobManagerForbiddenException;
 import org.apache.zeppelin.socket.NotebookServer;
 import org.apache.zeppelin.user.AuthenticationInfo;
 import org.quartz.CronExpression;
@@ -218,6 +219,14 @@ public class NotebookRestApi extends AbstractRestApi {
   private void checkIfParagraphIsNotNull(Paragraph paragraph, String paragraphId) {
     if (paragraph == null) {
       throw new ParagraphNotFoundException(paragraphId);
+    }
+  }
+
+  private void checkIfJobManagerIsEnabled() {
+    try {
+      jobManagerService.checkIfJobManagerIsEnabled();
+    } catch (JobManagerForbiddenException e) {
+      throw new ForbiddenException(e.getMessage());
     }
   }
 
@@ -1147,6 +1156,7 @@ public class NotebookRestApi extends AbstractRestApi {
   @ZeppelinApi
   public Response getJobListforNote() throws IOException, IllegalArgumentException {
     LOGGER.info("Get note jobs for job manager");
+    checkIfJobManagerIsEnabled();
     List<JobManagerService.NoteJobInfo> noteJobs = jobManagerService
             .getNoteJobInfoByUnixTime(0, getServiceContext(), new RestServiceCallback<>());
     Map<String, Object> response = new HashMap<>();
@@ -1170,6 +1180,7 @@ public class NotebookRestApi extends AbstractRestApi {
   public Response getUpdatedJobListforNote(@PathParam("lastUpdateUnixtime") long lastUpdateUnixTime)
       throws IOException, IllegalArgumentException {
     LOGGER.info("Get updated note jobs lastUpdateTime {}", lastUpdateUnixTime);
+    checkIfJobManagerIsEnabled();
     List<JobManagerService.NoteJobInfo> noteJobs =
             jobManagerService.getNoteJobInfoByUnixTime(lastUpdateUnixTime, getServiceContext(),
                     new RestServiceCallback<>());
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
@@ -18,6 +18,7 @@
 package org.apache.zeppelin.service;
 
 import jakarta.inject.Inject;
+import java.util.Collections;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
 import org.apache.zeppelin.notebook.AuthorizationService;
@@ -26,6 +27,7 @@ import org.apache.zeppelin.notebook.NoteInfo;
 import org.apache.zeppelin.notebook.Notebook;
 import org.apache.zeppelin.notebook.Paragraph;
 import org.apache.zeppelin.scheduler.Job;
+import org.apache.zeppelin.service.exception.JobManagerForbiddenException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,12 +57,26 @@ public class JobManagerService {
     this.zConf = zConf;
   }
 
+  public void checkIfJobManagerIsEnabled() throws JobManagerForbiddenException {
+    if (!zConf.isJobManagerEnabled()) {
+      throw new JobManagerForbiddenException();
+    }
+  }
+
+  private boolean isJobManagerDisabled(ServiceContext context, ServiceCallback<?> callback) throws IOException {
+    if (!zConf.isJobManagerEnabled()) {
+      callback.onFailure(new JobManagerForbiddenException(), context);
+      return true;
+    }
+    return false;
+  }
+
   public List<NoteJobInfo> getNoteJobInfo(String noteId,
                                           ServiceContext context,
                                           ServiceCallback<List<NoteJobInfo>> callback)
       throws IOException {
-    if (!zConf.isJobManagerEnabled()) {
-      return new ArrayList<>();
+    if (isJobManagerDisabled(context, callback)) {
+      return Collections.emptyList();
     }
 
     return notebook.processNote(noteId,
@@ -83,8 +99,8 @@ public class JobManagerService {
                                                     ServiceContext context,
                                                     ServiceCallback<List<NoteJobInfo>> callback)
       throws IOException {
-    if (!zConf.isJobManagerEnabled()) {
-      return new ArrayList<>();
+    if (isJobManagerDisabled(context, callback)) {
+      return Collections.emptyList();
     }
 
     List<NoteJobInfo> notesJobInfo = new LinkedList<>();
@@ -103,7 +119,7 @@ public class JobManagerService {
   public void removeNoteJobInfo(String noteId,
                                 ServiceContext context,
                                 ServiceCallback<List<NoteJobInfo>> callback) throws IOException {
-    if (!zConf.isJobManagerEnabled()) {
+    if (isJobManagerDisabled(context, callback)) {
       return;
     }
     List<NoteJobInfo> notesJobInfo = new ArrayList<>();
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/exception/JobManagerForbiddenException.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/exception/JobManagerForbiddenException.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.service.exception;
+
+/**
+ * Runtime exception thrown when the job manager is disabled.
+ */
+public class JobManagerForbiddenException extends Exception {
+
+  private static final long serialVersionUID = -8872599278254399427L;
+  public static final String MESSAGE = "Job Manager is disabled in the current configuration.";
+
+  public JobManagerForbiddenException() {
+    super(MESSAGE);
+  }
+
+  public JobManagerForbiddenException(String message) {
+    super(message);
+  }
+}
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -90,6 +90,7 @@ import org.apache.zeppelin.service.JobManagerService;
 import org.apache.zeppelin.service.NotebookService;
 import org.apache.zeppelin.service.ServiceContext;
 import org.apache.zeppelin.service.SimpleServiceCallback;
+import org.apache.zeppelin.service.exception.JobManagerForbiddenException;
 import org.apache.zeppelin.ticket.TicketContainer;
 import org.apache.zeppelin.types.InterpreterSettingsList;
 import org.apache.zeppelin.user.AuthenticationInfo;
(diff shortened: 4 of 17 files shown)


devyeony:
For reference, here are screenshots of the Job Manager page when `zeppelin.jobmanager.enable` is set to `true`. This might help with comparing the behavior between enabled and disabled states.

* New UI
Screenshot 2025-08-10 at 10 56 39 am

* Classic UI
Screenshot 2025-08-10 at 10 57 01 am

zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.zeppelin.notebook.NoteInfo;
 import org.apache.zeppelin.notebook.Notebook;
 import org.apache.zeppelin.notebook.Paragraph;
+import org.apache.zeppelin.rest.exception.ForbiddenException;
```
Reamer: Here, the exception is imported from the REST implementation. I don't like this style.

Perhaps a new RuntimeException would be better.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -1147,12 +1147,16 @@ public Response getCronJob(@PathParam("noteId") String noteId)
   @ZeppelinApi
   public Response getJobListforNote() throws IOException, IllegalArgumentException {
     LOGGER.info("Get note jobs for job manager");
-    List<JobManagerService.NoteJobInfo> noteJobs = jobManagerService
-            .getNoteJobInfoByUnixTime(0, getServiceContext(), new RestServiceCallback<>());
-    Map<String, Object> response = new HashMap<>();
-    response.put("lastResponseUnixTime", System.currentTimeMillis());
-    response.put("jobs", noteJobs);
-    return new JsonResponse<>(Status.OK, response).build();
+    try {
```
Reamer: The try-catch style is not used in this class. Please change the style of the implementation to the following structure.

```
- checkjobManagerEnabled()  --> this method can throw an RestException
- JobManager.anything()
```

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -548,45 +548,54 @@ public void unicastNoteJobInfo(NotebookSocket conn,
...
-          @Override
-          public void onSuccess(List<JobManagerService.NoteJobInfo> notesJobInfo,
-                                ServiceContext context) throws IOException {
-            super.onSuccess(notesJobInfo, context);
-            Map<String, Object> response = new HashMap<>();
-            response.put("lastResponseUnixTime", System.currentTimeMillis());
-            response.put("jobs", notesJobInfo);
-            conn.send(serializeMessage(new Message(OP.LIST_NOTE_JOBS).put("noteJobs", response)));
-          }
+    try {
```
Reamer: Here, too, I don't like the try-catch style. It's better to work on the callback and call `onFailure(...)`.

Reamer (changes requested):
Thank you for greatly improving the user interface at this point. As mentioned in the other comments, we should maintain the style of the existing code and not mix Rest and Websocket code.

devyeony:
@Reamer Thank you for your feedback and for recognizing the improvements to the user interface. 
I have updated the code based on your comments, and your thorough review has been very helpful in understanding and gaining a clearer picture of the project’s structure. 
Please feel free to share any feedback if there’s something I might have misunderstood or missed. Thank you again.

zeppelin-server/src/main/java/org/apache/zeppelin/service/exception/JobManagerForbiddenException.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.service.exception;
+
+/**
+ * Runtime exception thrown when the job manager is disabled.
+ */
+public class JobManagerForbiddenException extends RuntimeException {
+
```
Reamer, line 24: You forgot the serialVersionUID.

zeppelin-server/src/main/java/org/apache/zeppelin/service/exception/JobManagerForbiddenException.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.zeppelin.service.exception;
+
+/**
+ * Runtime exception thrown when the job manager is disabled.
+ */
+public class JobManagerForbiddenException extends RuntimeException {
```
Reamer: Maybe it's better to use and checked Exception. I would prefer a checked Exception, because we can handle the exception well, depending on whether it is thrown in the WebSocket or via Rest. What do you thing?
```suggestion
public class JobManagerForbiddenException extends Exception {
```

zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
```diff
@@ -55,11 +56,25 @@ public JobManagerService(Notebook notebook,
...
+    return false;
+  }
+
   public List<NoteJobInfo> getNoteJobInfo(String noteId,
                                           ServiceContext context,
                                           ServiceCallback<List<NoteJobInfo>> callback)
       throws IOException {
-    if (!zConf.isJobManagerEnabled()) {
+    if (isJobManagerDisabled(context, callback)) {
       return new ArrayList<>();
```
Reamer: This should be preferred.
```suggestion
      return Collections.emptyList();
```

zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
```diff
@@ -83,7 +98,7 @@ public List<NoteJobInfo> getNoteJobInfoByUnixTime(long lastUpdateServerUnixTime,
                                                     ServiceContext context,
                                                     ServiceCallback<List<NoteJobInfo>> callback)
       throws IOException {
-    if (!zConf.isJobManagerEnabled()) {
+    if (isJobManagerDisabled(context, callback)) {
       return new ArrayList<>();
```
Reamer: This should be preferred.
```suggestion
      return Collections.emptyList();
```

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -2294,7 +2305,9 @@ public class WebSocketServiceCallback<T> extends SimpleServiceCallback<T> {
     @Override
     public void onFailure(Exception ex, ServiceContext context) throws IOException {
       super.onFailure(ex, context);
-      if (ex instanceof ForbiddenException) {
+      if (ex instanceof JobManagerForbiddenException) {
```
Reamer: In my opinion, this is the wrong place.
Please override the `onFailure` method in the `JobManagerServiceCallback` class.

Reamer (changes requested):
A few suggestions regarding the backend code.
Please ensure that your formatter does not change the indentation of unmodified code.

zeppelin-server/src/main/java/org/apache/zeppelin/service/exception/JobManagerForbiddenException.java
devyeony: @Reamer I also think using a checked exception would make things clearer. Thank you for the suggestion!

devyeony:
@Reamer I’ve updated the code according to your feedback from the code review. I’d appreciate it if you could take a look. Thank you!

zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
```diff
@@ -1932,6 +1943,11 @@ public void onSuccess(List<JobManagerService.NoteJobInfo> notesJobInfo,
       connectionManager.broadcast(JobManagerServiceType.JOB_MANAGER_PAGE.getKey(),
           new Message(OP.LIST_UPDATE_NOTE_JOBS).put("noteRunningJobs", response));
     }
+
+    @Override
+    public void onFailure(Exception ex, ServiceContext context) {
+      LOGGER.debug(ex.getMessage());
```
Reamer, line 1949: The parent method should still be called so that other exceptions continue to be logged and handled.
```suggestion
      if (ex instanceof JobManagerForbiddenException) {
          LOGGER.debug(ex.getMessage());
      } else {
          super.onFailure(ex, context);
      }
```

zeppelin-server/src/test/java/org/apache/zeppelin/rest/NotebookRestApiTest.java
```diff
@@ -1187,4 +1189,72 @@ void testRunWithServerRestart() throws Exception {
...
+    }
+  }
+
+  @Test
+  void testGetUpdatedJobList_whenJobManagerDisabled() throws IOException {
+    boolean originalFlag = disableJobManagerAndBackupFlag();
+    String expectedErrorMessage = "Job Manager is disabled in the current configuration.";
+    CloseableHttpResponse response = null;
+
+    try {
```
Reamer, line 1226: Please use the try-with-resources pattern, which is more elegant.

Reamer (changes requested):
One style change and one change that does not restrict the current function but could potentially make future developments more difficult.

Unfortunately, I am unable to accept the frontend code. @tbonelee Can you help?

zeppelin-web/src/app/jobmanager/jobmanager.component.js
```diff
@@ -135,8 +137,14 @@ function JobManagerController($scope, ngToast, JobManagerFilter, JobManagerServi
...
     JobManagerService.subscribeUpdateJobs($scope, updateJobsCallback);
 
+    const jobManagerDisabledListener = $rootScope.$on('jobManagerDisabled', function(event, data) {
+      $scope.isFilterLoaded = true;
+      $scope.jobManagerEnabled = false;
+    });
+
     $scope.$on('$destroy', function() {
       JobManagerService.disconnect();
+      jobManagerDisabledListener();
```
tbonelee, line 147: Maybe we could use a variable name that makes its call-time role explicit, such as `deregistration`.

zeppelin-web/src/app/jobmanager/jobmanager.html
tbonelee, line 1: How about moving th `ng-if="jobManagerEnabled"` block inside the `ng-if="isFilterLoaded"` block?
This would make it clearer that the `ng-if="jobManagerEnabled"` block only renders once loading is complete, aligning better with the logical flow.

zeppelin-web-angular/projects/zeppelin-sdk/src/interfaces/message-data-type-map.interface.ts
```diff
@@ -75,7 +75,7 @@ import {
   RunParagraph
 } from './message-paragraph.interface';
 
-import { ListNoteJobs, ListUpdateNoteJobs } from './message-job.interface';
+import { ListNoteJobs, ListUpdateNoteJobs, JobManagerDisabled } from './message-job.interface';
```
tbonelee, line 78: Consider ordering the imports like this. While not all TSLint rules are followed at the moment, we could start aligning them with this commit and in future ones.
```suggestion
import { JobManagerDisabled, ListNoteJobs, ListUpdateNoteJobs } from './message-job.interface';
```

zeppelin-web-angular/src/app/pages/workspace/job-manager/job-manager.component.html
tbonelee, line 1: How about combining `loading` and `isJobManagerEnabled` into a single variable like this?
```typescript
status: 'loading' | 'success' | 'disabled' = 'loading';
```

This could allow us to replace multiple if/else branches with a straightforward switch statement. What do you think?
```html
  <ng-container [ngSwitch]="status">
    <nz-card *ngSwitchCase="'loading'">
      ...
    </nz-card>

    <ng-container *ngSwitchCase="'success'">
       ...
    </ng-container>

    <ng-container *ngSwitchCase="'disabled'">
       ...
    </ng-container>
  </ng-container>
```

tbonelee (changes requested):
I focused mainly on the frontend. I didn't find any functional issues and only added a few refactoing suggestions. I'd appreciate it if you could review them.

zeppelin-web-angular/src/app/pages/workspace/job-manager/job-manager.component.html
devyeony: @tbonelee Good idea, I agree. I'll refactor the code to use an ngSwitch instead of multiple if/else branches.

zeppelin-web/src/app/jobmanager/jobmanager.html
devyeony: @tbonelee Thanks for the feedback! I agree that your suggestion to nest `ng-if="jobManagerEnabled"` inside `ng-if="isFilterLoaded"` definitely makes the logical flow of the code clearer.

While testing that structure, I discovered it introduced a noticeable UI delay. The nested `ng-if` blocks the entire job list from rendering until after the data has loaded, which can make the app feel slow to the user.

To improve this, I tried keeping the blocks as siblings. This allows the browser to render the list as soon as the data arrives, creating a much smoother user experience.

Here’s a quick summary of the trade-offs:

| Approach                  | Pros                                      | Cons                                       |
| :------------------------ | :---------------------------------------- | :----------------------------------------- |
| **Nested `ng-if` (Suggested)** | Very clear, sequential logic           | Can cause a rendering delay   |
| **Sibling `ng-if`s (Current)** | Fast, non-blocking render | Code logic is parallel, not sequential |

I'm trying to find the best way to balance the code clarity you pointed out with the performance I was aiming for.

I'd love to get your thoughts on this. Perhaps there's another pattern that could give us the best of both worlds?

Please feel free to correct me if my understanding is wrong on any of this. Thank you again!

tbonelee: Thansk for the detailed explanation! From what I understand, the 'UI delay' you mentioned might not necessarily come from the `ng-if` nesting itself, but possibly from the 'digest cycle' of AngularJS (which cause UI updates) not being triggered when the state is updated inside a `Promise.then()`.

If the promise is a native ES6 `Promise` (rather than AngularJS's `$q` or `$http`), AngularJS doesn't automatically run a digest cycle. In that case, the view won't update until the next `$apply()` happens somewhere else, which can fell like a rendering delay.

One possible way to address this could be wrapping the state update in `$scope.$apply()` within the `then()` callback.

```js
asyncNotebookJobFilter(...)
  .then(() => {
    $scope.apply(() => {
      $scope.isFilterLoaded = true;
    });
  });
```

The other could be replacing `new Promise()` in `asyncNotebookJobFilter` with `$q()`Since `$q` is integrated with the digest cycle, view updates will be reflected automatically. You would need to inject `$q` into `JobManagerController`.

If this doesn't match the issue you observed, please feel free to correct me!

PR #5026, opened 2025-08-15
[ZEPPELIN-6282] Add null check condition

proceane:
### What is this PR for?
This PR was created to prevent an error that occurs in the `InterpreterSetting.getInterpreterGroup` method when `groupId` is null and is retrieved from the Set.
Normally, the groupId parameter does not seem to be passed as null.
I found it when tested service shutdown(#4965).

### What type of PR is it?
Improvement

### Todos
* [x] - Add null check condition

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6282

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
@@ -491,6 +491,9 @@ public class InterpreterSetting {
   }
 
   ManagedInterpreterGroup getInterpreterGroup(String groupId) {
+    if (groupId == null) {
+      return null;
+    }
     return interpreterGroups.get(groupId);
   }
 
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/SessionConfInterpreter.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/SessionConfInterpreter.java
@@ -49,8 +49,14 @@ public class SessionConfInterpreter extends ConfInterpreter {
       finalProperties.putAll(updatedProperties);
       LOGGER.debug("Properties for Session: {}:{}", sessionId, finalProperties);
 
-      List<Interpreter> interpreters =
-          interpreterSetting.getInterpreterGroup(interpreterGroupId).get(sessionId);
+      InterpreterGroup interpreterGroup =
+            interpreterSetting.getInterpreterGroup(interpreterGroupId);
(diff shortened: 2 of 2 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -491,6 +491,9 @@ public ManagedInterpreterGroup getInterpreterGroup(ExecutionContext executionCon
   }
 
   ManagedInterpreterGroup getInterpreterGroup(String groupId) {
+    if (groupId == null) {
+      return null;
```
ParkGyeongTae, line 495: Could you add a warn log before returning in the null check so that this situation is more visible?

proceane:
@ParkGyeongTae Thanks for review.
In my case, the problem occurred when I set it as follows.
1. When the ZeppelinServer shutdown method is executed, the `shardServiceLocator.shutdown()` method is executed.
2. Set the value of the `zeppelin.search.enable` option to `false` in the Recovery Test (if this value is true, an error occurs first in `LuceneSearch`, so it cannot be reproduced)

If you set it up as above and run the recovery test, it seems that a problem occurs where the groupId is set to null in the interpreter setting section when restarting the server.
As I mentioned at the beginning, this doesn't usually happen.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
proceane: Returning `null` could cause an NPE in the code currently calling this method (e.g., SessionConfInterpreter:53).
Therefore, I think it would be better to either not return null when groupId is null, or to modify the part that uses this method.

ParkGyeongTae: I agree with your point, and I’d like to ask for a code update.
I found that `ManagedInterpreterGroup getInterpreterGroup(String groupId)` is used in two places: line 386 of `InterpreterSettingManager.java` and line 53 of `SessionConfInterpreter.java`.

In `InterpreterSettingManager.java`, null is already handled properly, but in `SessionConfInterpreter.java` it is not. Could you please update the code to handle null in that part as well?

ParkGyeongTae (approved):
I was able to reproduce this issue, and I confirmed that the changes fix the problem.

ParkGyeongTae:

PR #5036, opened 2025-08-20
[ZEPPELIN-6287] Fix lint errors in zeppelin-web-angular modules

seung-00:
### What is this PR for?
Fix various lint errors in zeppelin-web-angular.
I think they were hidden because only lint-staged was running on staged files during pre-commit.
- add prettier check to the lint script
- add types/webpack-env for better type safety

### What type of PR is it?
Refactoring

### Todos
* [ ] - Add lint check to zeppelin-web-angular CI workflow

### What is the Jira issue?
* [ZEPPELIN-6287](https://issues.apache.org/jira/browse/ZEPPELIN-6287)

* run `npm run lint`

### Questions:


seung-00:
@tbonelee 
I plan to run the lint script in CI, so I’d rather not include any fix there. We already run lint‑staged with fixes locally, which I think should be sufficient. Would that be okay with you?
`@types/webpack-env` is needed to use NodeJS.Require type.

tbonelee:
> I plan to run the lint script in CI, so I’d rather not include any fix there. We already run lint‑staged with fixes locally, which I think should be sufficient. Would that be okay with you?

Ah, sorry, I think my earlier comment might have been unclear.
I didn't mean adding `--write` to the lint script in CI. I was suggesting running Prettier with `--write` once on
 the existing files so the codebase is already formatted, and then the lint step would pass cleanly without further fixes. Similar to how you've already fixed the ESLint errors.

> `@types/webpack-env` is needed to use NodeJS.Require type.

Got it, thanks for confirming!

seung-00:
I see. I’ve pushed the formatted code. Thank you. @tbonelee

tbonelee (approved):
LGTM. Thanks for the additional work.

seung-00:
@tbonelee Could you merge this PR? There might be some conflicts since I touched a lot of files.

tbonelee:

PR #5044, opened 2025-08-24
[ZEPPELIN-6294] Made "Clone paragraph" button work properly in New UI

voidmatcha:
### What is this PR for?
**Description:**
Clicking the paragraph control’s Clone paragraph button that trigger COPY_PARAGRAPH. Upon receiving PARAGRAPH_ADDED, the Classic UI includes data.paragraph.text as expected, but the New UI receives an empty data.paragraph.text. Additionally, after cloning, the cursor is not positioned in the cloned paragraph in the New UI.

**Expected:**
- Cloned paragraph contains the original text.
- Cursor focuses/positions inside the cloned paragraph.

**Actual (New UI):**
- data.paragraph.text is empty on PARAGRAPH_ADDED.
- Cursor not focused / not moved to the cloned paragraph.

https://github.com/user-attachments/assets/dd96ceaa-b0f1-4b1e-ad11-46560a0e2b79

**[AS-IS]**

https://github.com/user-attachments/assets/39530005-b169-4301-9416-ed14cb6888fc

**[TO-BE]**

https://github.com/user-attachments/assets/81302edb-0d52-49e9-a554-e2509a7bca18

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* [[ZEPPELIN-6294](https://issues.apache.org/jira/browse/ZEPPELIN-6294)]

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
+++ b/zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
@@ -31,6 +31,7 @@ import { NoteStatusService, ParagraphStatus } from '@zeppelin/services/note-stat
 import * as DiffMatchPatch from 'diff-match-patch';
 import { isEmpty, isEqual } from 'lodash';
 
+import { NotebookParagraphCodeEditorComponent } from '@zeppelin/pages/workspace/notebook/paragraph/code-editor/code-editor.component';
 import { NotebookParagraphResultComponent } from '@zeppelin/pages/workspace/share/result/result.component';
 import { ParagraphConfigResults, ParagraphResults } from '../../../../projects/zeppelin-sdk/src';
 import { MessageListener, MessageListenersManager } from '../message-listener/message-listener';
@@ -54,6 +55,7 @@ export abstract class ParagraphBase extends MessageListenersManager {
 
   // Initialized by `ViewChildren` in the class which extends ParagraphBase
   notebookParagraphResultComponents!: QueryList<NotebookParagraphResultComponent>;
+  notebookParagraphCodeEditorComponent!: NotebookParagraphCodeEditorComponent;
 
   constructor(
     public messageService: MessageService,
@@ -142,6 +144,8 @@ export abstract class ParagraphBase extends MessageListenersManager {
         }
         this.cdr.markForCheck();
       });
+      this.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
+      this.notebookParagraphCodeEditorComponent.getEditorSetting();
       this.cdr.markForCheck();
     }
   }
--- a/zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
+++ b/zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
@@ -55,7 +55,7 @@ export abstract class ParagraphBase extends MessageListenersManager {
 
   // Initialized by `ViewChildren` in the class which extends ParagraphBase
   notebookParagraphResultComponents!: QueryList<NotebookParagraphResultComponent>;
-  notebookParagraphCodeEditorComponent!: NotebookParagraphCodeEditorComponent;
+  notebookParagraphCodeEditorComponent?: NotebookParagraphCodeEditorComponent;
 
   constructor(
     public messageService: MessageService,
@@ -144,8 +144,10 @@ export abstract class ParagraphBase extends MessageListenersManager {
         }
         this.cdr.markForCheck();
       });
-      this.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
-      this.notebookParagraphCodeEditorComponent.getEditorSetting();
+      if (this.notebookParagraphCodeEditorComponent) {
+        this.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
+        this.notebookParagraphCodeEditorComponent.getEditorSetting();
+      }
       this.cdr.markForCheck();
     }
   }
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
@@ -163,6 +163,7 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
     }
     const definedNote = this.note;
     definedNote.paragraphs.splice(data.index, 0, data.paragraph);
+    this.cdr.detectChanges();
     const paragraphIndex = definedNote.paragraphs.findIndex(p => p.id === data.paragraph.id);
 
     definedNote.paragraphs[paragraphIndex].focus = true;
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
@@ -111,6 +111,10 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
 
   @MessageListener(OP.INTERPRETER_BINDINGS)
   loadInterpreterBindings(data: MessageReceiveDataTypeMap[OP.INTERPRETER_BINDINGS]) {
+    this.listOfNotebookParagraphComponent.forEach(item => {
+      item.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
+      item.notebookParagraphCodeEditorComponent.getEditorSetting();
+    });
     this.interpreterBindings = data.interpreterBindings;
     if (!this.interpreterBindings.some(item => item.selected)) {
       this.activatedExtension = 'interpreter';
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
@@ -163,10 +163,10 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
     }
     const definedNote = this.note;
     definedNote.paragraphs.splice(data.index, 0, data.paragraph);
-    this.cdr.detectChanges();
     const paragraphIndex = definedNote.paragraphs.findIndex(p => p.id === data.paragraph.id);
 
     definedNote.paragraphs[paragraphIndex].focus = true;
+    this.cdr.detectChanges();
     const addedParagraph = this.listOfNotebookParagraphComponent.find((_, index) => index === paragraphIndex)
       ?.notebookParagraphCodeEditorComponent;
 
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
@@ -76,6 +76,9 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
   sidebarWidth = 370;
   sidebarAnimationFrame = -1;
   isSidebarOpen = false;
+  // send > CLONE_PARAGRAPH: receive > PARAGRAPH_ADDED → PARAGRAPH → trigger EDITOR_SETTING
+  // send > INSERT_PARAGRAPH: receive > trigger EDITOR_SETTING after PARAGRAPH_ADDED
+  isTriggeredByInsertParagraph = false;
 
   @MessageListener(OP.NOTE)
   getNote(data: MessageReceiveDataTypeMap[OP.NOTE]) {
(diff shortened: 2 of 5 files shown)


tbonelee (commented):
Thanks for taking a look! From my side, data.paragraph.text appears on PARAGRAPH_ADDED in both UIs, but I might be mistaken. If there’s a case where it’s empty, I’d love the exact repro steps.
One possibility is that the New UI is skipping a valid PARAGRAPH as if it were short-circuited when another request bumps lastMsgIdSeq before the final reply—`PARAGRAPH` message (see message.ts, L187). I noticed `EDITOR_SETTING` messages being sent from the client and suspect they might be the trigger. If the only intended short-circuit is RUN_PARAGRAPH → synthetic PARAGRAPH_STATUS, would it make sense to remove the second filter() at L187 rather than adding a special-case branch at the `NotebookServer`? Happy to adjust based on your feedback.

voidmatcha:
I really appreciate about your review. 👍 👍

As you mentioned, the issue was caused by EDITOR_SETTING being triggered before PARAGRAPH, which led to short-circuit processing. However, modifying the condition in [message.ts, L187](https://github.com/dididy/zeppelin/blob/6ffd605bd71e3857f5680d2de26ad71469f6b198/zeppelin-web-angular/projects/zeppelin-sdk/src/message.ts#L186-L198) didn’t seem optimal from a performance standpoint.(I'm not entirely sure if that's really the case)

So I adjusted the timing in the relevant section of the [app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts](https://github.com/apache/zeppelin/pull/5044/commits/1bbaeaf1fbba60531e774a199e9c445b6242c99d#diff-b9e215ed5d0d59a7732256f1da6be620f4cdcc62019f8807a9046e60017551ae) instead. Because the [`getEditorSetting`](https://github.com/dididy/zeppelin/blob/6ffd605bd71e3857f5680d2de26ad71469f6b198/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts#L196) method that triggers EDITOR_SETTING is [only executed when the parameter of `setParagraphMode` is `true`](https://github.com/dididy/zeppelin/blob/6ffd605bd71e3857f5680d2de26ad71469f6b198/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts#L192-L198), and the only place it is used in this form is inside [`onDidChangeModelContent`](https://github.com/dididy/zeppelin/blob/6ffd605bd71e3857f5680d2de26ad71469f6b198/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts#L86-L99). Therefore, [I modified it so that when `setParagraphMode(true)` is called in that area](https://github.com/dididy/zeppelin/blob/6ffd605bd71e3857f5680d2de26ad71469f6b198/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts#L94), the timing is adjusted using setTimeout.

After reverted NotebookServer code, everything worked as intended. Thanks for pointing out a better way to improve it.

tbonelee (changes requested):
Nice fix! Since the `setTimeout(..., 200)` is a heuristic to dodge a race, could we add a short comment explaining (1) why it's needed and (2) a TODO o how to make this deterministic or order-independent?

voidmatcha:
> Nice fix! Since the `setTimeout(..., 200)` is a heuristic to dodge a race, could we add a short comment explaining (1) why it's needed and (2) a TODO o how to make this deterministic or order-independent?

https://github.com/apache/zeppelin/pull/5044/commits/ab073599e086c6f9fb9084122814c0b0778d970d

Ensuring the execution order of `EDITOR_SETTING` seems necessary because it can otherwise cause a bug that passes through the short-circuit condition. Although this issue does not reproduce for everyone, controlling the execution timing appears to be required.

image

EDITOR_SETTING is executed when the first paragraph component is rendered. From the Classic UI console logs above, we can see that it runs after `INTERPRETER_BINDINGS` when first entering the note. Therefore, I added a variable **editorSettingTriggerAllowed** with an initial value of false in `/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts`. The **getEditorSetting** method now only executes if this variable is true.

Then, in `/app/pages/workspace/notebook/notebook.component.ts`, I added code that sets **editorSettingTriggerAllowed** to true for all paragraphs and calls **getEditorSetting** when receiving `INTERPRETER_BINDINGS`.

스크린샷 2025-08-27 오후 11 10 40
<= Classic UI
<br />
스크린샷 2025-08-27 오후 11 13 32
<= New UI unrevised result 
스크린샷 2025-08-27 오후 11 16 59
<= New UI revised result 

The part where I noticed the issue was at the moment of receiving `PARAGRAPH_ADDED` op. When a new paragraph component is added by `PARAGRAPH_ADDED`, **getEditorSetting** is immediately executed. This triggers `EDITOR_SETTING`, but since execution is now blocked by **editorSettingTriggerAllowed**, as I fixed, it does not run at that point. Instead, **editorSettingTriggerAllowed** is set to true and **getEditorSetting** is executed when the `PARAGRAPH` op(`/app/core/paragraph-base/paragraph-base.ts`) is received.

zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
```diff
@@ -54,6 +55,7 @@ export abstract class ParagraphBase extends MessageListenersManager {
 
   // Initialized by `ViewChildren` in the class which extends ParagraphBase
   notebookParagraphResultComponents!: QueryList<NotebookParagraphResultComponent>;
+  notebookParagraphCodeEditorComponent!: NotebookParagraphCodeEditorComponent;
```
tbonelee: I think this non-null assertion should be removed and changed to optional. I found it to be `undefined` when `paragraph.config.editorHide === true` in the `%md` editor case. Maybe you could check some errors after running `%md` interpreter.
(`zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.html`)
```html
    <zeppelin-notebook-paragraph-code-editor
      *ngIf="!paragraph.config.editorHide && !viewOnly"
...
```
I mistakenly added non-null assertion for `notebookParagraphCodeEditorComponent` in `NotebookParagraphComponent` in a previous commit, but that was incorrect. Sorry for the confusion.

zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
```diff
@@ -109,6 +109,10 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
 
   @MessageListener(OP.INTERPRETER_BINDINGS)
   loadInterpreterBindings(data: MessageReceiveDataTypeMap[OP.INTERPRETER_BINDINGS]) {
+    this.listOfNotebookParagraphComponent.forEach(item => {
+      item.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
```
tbonelee: We should handle some case where `notebookParagraphCodeEditorComponent` is undefined.
Also, should we trigger editor settings for all paragraphs here? Would calling `this.setParagraphMode(true)` in `editor.onDidChangeModelContent()` be sufficient to get editor settings only to the paragraph whose interpreter has changed?

zeppelin-web-angular/src/app/core/paragraph-base/paragraph-base.ts
```diff
@@ -142,6 +144,8 @@ export abstract class ParagraphBase extends MessageListenersManager {
         }
         this.cdr.markForCheck();
       });
+      this.notebookParagraphCodeEditorComponent.editorSettingTriggerAllowed = true;
```
tbonelee: We should handle the case where `this.notebookParagraphCodeEditorComponent` is undefined

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -61,6 +61,7 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
   private monacoDisposables: IDisposable[] = [];
   height = 18;
   interpreterName?: string;
+  editorSettingTriggerAllowed: boolean = false;
```
tbonelee, line 66: Could we perhaps add a comment explaining  why this property is introduced here?

tbonelee (changes requested):
Thanks for the thorough update — I’ve left some questions and requested changes.

voidmatcha:
c90cd09
- Made `notebookParagraphCodeEditorComponent` optional.

5e35dbf
- Added a comment for `editorSettingTriggerAllowed`.

6c993a0
- There are two related actions: `CLONE_PARAGRAPH` and `INSERT_PARAGRAPH`. Each needs to be handled slightly differently.

#### Behavior
- **CLONE_PARAGRAPH**
  1. `send > CLONE_PARAGRAPH`
  2. `receive > PARAGRAPH_ADDED → PARAGRAPH → trigger EDITOR_SETTING`
- **INSERT_PARAGRAPH**
  1. `send > INSERT_PARAGRAPH`
  2. `receive > trigger EDITOR_SETTING` **after** `PARAGRAPH_ADDED`

#### Implementation
- `CLONE_PARAGRAPH` was already handled correctly.
- For `INSERT_PARAGRAPH`, we added:
  - A variable `isTriggeredByInsertParagraph` in `notebook.component.ts`.
  - Set it to `true` when `insertParagraph` in `paragraph.component.ts` is executed.
- This allows branching based on whether `PARAGRAPH_ADDED` was triggered by an `INSERT_PARAGRAPH` or a `CLONE_PARAGRAPH`.

#### Reasoning
- `CLONE_PARAGRAPH`: `EDITOR_SETTING` runs **after receiving PARAGRAPH**.
- `INSERT_PARAGRAPH`: `EDITOR_SETTING` runs **when `PARAGRAPH_ADDED` is received**.

zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
voidmatcha: In testing, the interpreter setting works correctly when `this.setParagraphMode(true)` is executed inside `editor.onDidChangeModelContent()`.

Since this runs **before** the code that executes when receiving `INTERPRETER_BINDINGS` above (`editor.onDidChangeModelContent()`), the code that runs after `editorSettingTriggerAllowed = true` was modified to directly call `getEditorSetting()` instead of using `setParagraphMode(true)` when receive `PARAGRAPH` or `INTERPRETER_BINDINGS`. 

c90cd09

Because if the interpreter is already set inside `this.setParagraphMode(true)`, `getEditorSetting()` will not be executed due to the condition, and since `setParagraphMode(true)` has already run, there is no need to execute it again.

voidmatcha:
I rebased due to conflicts caused by merged PRs.

zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
```diff
@@ -142,15 +155,19 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
...
-        focus: p.id === data.paragraph.id
-      };
-    });
-    definedNote.paragraphs = [...definedNote.paragraphs];
+    definedNote.paragraphs.splice(data.index, 0, data.paragraph);
+    const paragraphIndex = definedNote.paragraphs.findIndex(p => p.id === data.paragraph.id);
+
+    definedNote.paragraphs[paragraphIndex].focus = true;
+    const addedParagraph = this.listOfNotebookParagraphComponent.find((_, index) => index === paragraphIndex)
+      ?.notebookParagraphCodeEditorComponent;
```
tbonelee: I think something like `this.cdr.detectChanges()` should be called, so that the paragraph component is added for the new paragraph in `definedNote.paragraph`.

voidmatcha: 5aab81a c496536

Upon review, components using `ChangeDetectionStrategy.OnPush`—as you mentioned—do not detect changes in the same reference array unless the reference itself is replaced. Since I am currently modifying the array with `splice` (which does change the reference in place), Angular does not automatically recognize the change. Therefore, calling `detectChanges` or `markForCheck` is necessary.

The reason why Clone Paragraph or Add Paragraph (Insert Paragraph) in the current PR appeared to work correctly is due to Angular’s behavior of updating the view during events triggered within the component. Consequently, when executed via a button click, the focus handling works as expected, but when triggered via a shortcut key, the focus does not update properly.

To address this, after receiving `PARAGRAPH_ADDED` and performing a `splice` to update the array, we now explicitly call `detectChanges` to force change detection, which has been confirmed to work correctly.

1cdff50

Even after handling it this way, the focus event did not work correctly when triggered via shortcut keys. To address this, in `paragraph.component.ts`, I added a blur event on the current paragraph when executing `cloneParagraph` and `insertParagraph`.

---

I rebased again due to conflicts caused by merging the latest PR.

voidmatcha:
This has been improved with a better approach in #5254. For the cursor-related issue, I plan to create a separate issue and address it separately. - ZEPPELIN-6429

PR #5046, opened 2025-08-26
[ZEPPELIN-6295] Add configurations that allow LDAP users map directly to role

paul8263:
### What is this PR for?

Add configurations that allow LDAP users map directly to the role

### What type of PR is it?

Improvement

### Todos
N/A

### What is the Jira issue?
[ZEPPELIN-6295](https://issues.apache.org/jira/browse/ZEPPELIN-6295)

Added unit tests & compiled and tested manually.

### Questions:
* Does this needs documentation? Yes. And the doc has been updated.

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
@@ -1,0 +1,87 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.zeppelin.realm;
+
+import org.apache.shiro.authz.AuthorizationInfo;
+import org.apache.shiro.authz.SimpleAuthorizationInfo;
+import org.apache.shiro.realm.ldap.DefaultLdapRealm;
+import org.apache.shiro.realm.ldap.LdapContextFactory;
+import org.apache.shiro.subject.PrincipalCollection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.naming.NamingException;
+import javax.naming.ldap.LdapContext;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * RoleMappingLdapRealm is a simple LDAP realm that returns configured role for specific users.
+ * Multiple roles(quoted string, comma separated) can be specified for a user.
+ * Example configurations:
+ * ldapRealm = org.apache.zeppelin.realm.RoleMappingLdapRealm
+ * ldapRealm.rolesByUsername = "user1":"admin","user2":"role1,role2","user3":"role3"
+ * ldapRealm.defaultRole = "role1,role2"
+ */
+public class RoleMappingLdapRealm extends DefaultLdapRealm {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(RoleMappingLdapRealm.class);
+
+  private static final String ROLE_NAMES_DELIMITER = ",";
+  private String defaultRole = "role1";
+
+  private final Map<String, String> rolesByUsername = new LinkedHashMap<>();
+
+
+  public String getDefaultRole() {
+    return defaultRole;
+  }
+
+  public void setDefaultRole(String defaultRole) {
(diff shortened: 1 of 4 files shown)


zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
```diff
@@ -0,0 +1,86 @@
...
+ * Example configurations:
+ * ldapRealm = org.apache.zeppelin.realm.RoleMappingLdapRealm
+ * ldapRealm.rolesByUsername = user1:admin,user2:user,user3:user,user4:user
+ * ldapRealm.defaultRole = user
+ */
+public class RoleMappingLdapRealm extends DefaultLdapRealm {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(RoleMappingLdapRealm.class);
+
+  private String defaultRole = "admin";
```
hyunw9: Hello, thanks for the implementation. I have a quick question about the default role.
I noticed that defaultRole is initialized to "admin". If an administrator forgets to set the `ldapRealm.defaultRole property` in shiro.ini, does this mean any authenticated user not in the rolesByUsername map would get admin privileges by default? This seems like a potential security risk.
I’m wondering if this behavior is intentional.

zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public Set<String> getRoleNamesForUser(String username, LdapContext ldapContext,
+                                         String userDnTemplate) {
+    String role = rolesByUsername.get(username) == null ? defaultRole : rolesByUsername.get(username);
+
+    LOGGER.debug("For user {}, RoleMappingLdapRealm returns {} as the role", username, role);
+    Set<String> roles = new HashSet<>();
+    roles.add(role);
+    return roles;
```
hyunw9: I think this could be more concise using the Java 9 ~ API:
```
String role = rolesByUsername.getOrDefault(username, defaultRole);

LOGGER.debug("For user {}, RoleMappingLdapRealm returns {} as the role", username, role);
return Set.of(role);
```
Would you consider updating the implementation?

zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
paul8263: Hi @hyunw9 ,
Thanks for the review. I will change the default role to role1, keeping it consistent with the example in shiro.ini.template.
The examples in the doc are correct but I forgot to change the default value for "defaultRole". Sorry for that.

zeppelin-server/src/main/java/org/apache/zeppelin/realm/RoleMappingLdapRealm.java
paul8263: I will do it later

paul8263: But changing it to Set.of(role) seems to be incompatible with Java 8. Should we remain the Java 8 compatibility?

hyunw9: Sorry for the confusion. As you pointed out, we need to support JDK 1.8, so it would be better to keep using the new HashSet<>() approach.

paul8263: I reviewed the doc named how_to_build.md and its requirements chapter states that the Jave should be 1.8 (151+).

If we should remain the backward compatibility of Java 8, we'd better not use Set.of(role).

Besides, I found another compilation issue with Java 8, locating in JDBCInterpreter::validateConnectionUrl

If we should keep Java 8 working, I would like to file another issue.

Thanks. Looking forward to your reply.

hyunw9: Thank you for your support. and i agree with your opinion. 
If you don’t mind, I’d be happy to follow up together once you create the new issue. 😄

Reamer: We have a documentation issue. JDK 11 is required as a minimum.
https://github.com/apache/zeppelin/blob/8d73bfc50c5746913b762a5d83acf6d9628efc10/pom.xml#L100

Please use `Set.of(...)`

hyunw9: That could be a bit confusing. As a follow-up, it would be good to update the documentation.

paul8263: OK. I will use Set.of(...)

Reamer:
Looks good to me. Why can a user only have one role?

paul8263:
> Looks good to me. Why can a user only have one role?

Hi @Reamer ,
I referred to ldapRealm.rolesByGroup, which is one to one mapping.But activeDirectoryRealm.groupRolesMap supports one to many. If necessary I will update the code.

Reamer:
> If necessary I will update the code.

Not for me, I have a setup. Which I don't plan to change. If you don't have a requirement, then it's not a requirement.

ParkGyeongTae (approved):
Could you please write the PR title in full? The current one seems to be cut off.

paul8263:
> Could you please write the PR title in full? The current one seems to be cut off.

The title is OK now.

ParkGyeongTae:

PR #5048, opened 2025-08-26
[ZEPPELIN-6299] Refactor StaticRepl for readability, correctness, and modern Java usage

eunhwa99:
### What is this PR for?
This PR refactors the `StaticRepl` class to improve readability, maintainability, and align with modern Java best practices.  
There are no functional changes.  

- Generic type refinement with diamond operator in loop:  Adds type safety by explicitly specifying the generic type. Prevents unchecked warnings and makes the code clearer to readers and tools.
- Replace `Arrays.asList` to `List.of` :  List.of (Java 9+) produces an immutable list, which is safer and better communicates the intent that the list will not be modified. It avoids side effects of Arrays.asList, which is fixed-size and backed by the original array.
- Minor typos fixed: Corrected spelling mistakes and adjusted comments.
-  Removes redundant boolean comparison to follow standard Java coding conventions.
- Replace index-based loop with enhanced for-each to eliminate boilerplate index management, improves readability, and reduces chances of off-by-one errors. It makes the intent ("iterate all elements") clearer.
- Variable declaration moved closer to usage to improve code locality and readability by reducing variable scope. This makes the code easier to maintain and follow.

### What type of PR is it?
Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-6299](https://issues.apache.org/jira/browse/ZEPPELIN-6299)  

* Run existing unit tests to confirm behavior remains unchanged.  

### Questions:

### DIFF
--- a/java/src/main/java/org/apache/zeppelin/java/StaticRepl.java
+++ b/java/src/main/java/org/apache/zeppelin/java/StaticRepl.java
@@ -39,7 +39,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -85,8 +84,8 @@ public class StaticRepl {
     // replace name of class containing Main method with generated name
     code = code.replace(mainClassName, generatedClassName);
 
-    JavaFileObject file = new JavaSourceFromString(generatedClassName, code.toString());
-    Iterable<? extends JavaFileObject> compilationUnits = Arrays.asList(file);
+    JavaFileObject file = new JavaSourceFromString(generatedClassName, code);
+    Iterable<? extends JavaFileObject> compilationUnits = List.of(file);
 
     ByteArrayOutputStream baosOut = new ByteArrayOutputStream();
     ByteArrayOutputStream baosErr = new ByteArrayOutputStream();
@@ -101,6 +100,8 @@ public class StaticRepl {
     System.setOut(newOut);
     System.setErr(newErr);
 
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
     CompilationTask task = compiler.getTask(null, null, diagnostics, null, null, compilationUnits);
 
     // executing the compilation process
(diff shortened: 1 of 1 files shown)


java/src/main/java/org/apache/zeppelin/java/StaticRepl.java
```diff
@@ -104,14 +100,16 @@ public static String execute(String generatedClassName, String code) throws Exce
     System.setOut(newOut);
     System.setErr(newErr);
 
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
```
ParkGyeongTae, line 104: Maybe consider moving this code a bit higher?

Reamer: Why?

ParkGyeongTae: I previously suggested moving `DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();` to the top, but that was my mistake.  

What I really meant is that we should move `JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();` to the top.  

The reason is that `ToolProvider.getSystemJavaCompiler()` can return `null` if running with a JRE instead of a JDK.  
By checking this early, we can fail fast before doing unnecessary work like parsing the code or redirecting `System.out/err`.

eunaverse: Oh I see, thanks for the point!

Would it make sense to throw an exception here if `JavaCompiler` is null, since running Java code requires a JDK? 
Since this seems more related to functional stability, I feel it could be handled in a separate PR. How do you feel about that?

ParkGyeongTae: My personal opinion is that it would be good to throw an exception if `JavaCompiler` is null.
I also think it would be better to handle this in a separate PR.

ParkGyeongTae (approved):
I have checked the existing unit tests.

java/src/main/java/org/apache/zeppelin/java/StaticRepl.java
eunaverse: Thanks for your opinion 👍  I have created another Jira Issue for this. https://issues.apache.org/jira/browse/ZEPPELIN-6306

ParkGyeongTae:

PR #5049, opened 2025-08-26
[ZEPPELIN-6300] Replace deprecated StringUtils.equals to Objects.equals

eunhwa99:
### What is this PR for?
Replaces deprecated `StringUtils.equals` with `Objects.equals` in `NotebookServer` for null-safe comparison.

### What type of PR is it?
Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* [ZEPPELIN-6300](https://issues.apache.org/jira/browse/ZEPPELIN-6300)

* Run existing unit tests for `NotebookServer` to confirm behavior is unchanged. 
### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -35,6 +35,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


Reamer:
Why don't you use the following method as suggested by JavaDoc? `[Strings.CS.equals(CharSequence, CharSequence)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/Strings.html#equals(java.lang.CharSequence,java.lang.CharSequence))

eunaverse:
> Why don't you use the following method as suggested by JavaDoc? `[Strings.CS.equals(CharSequence, CharSequence)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/Strings.html#equals(java.lang.CharSequence,java.lang.CharSequence))

Thank you for the suggestion. I initially replaced the deprecated `StringUtils.equals` with `Objects.equals` to use a JDK standard null-safe comparison and avoid relying on external libraries. 
However, I agree with your point and have updated it to use `Strings.CS.equals` as recommended by the JavaDoc, ensuring null-safe comparison while following the project's Apache Commons guidelines.

Reamer (approved):

Reamer:
> I initially replaced the deprecated StringUtils.equals with Objects.equals to use a JDK standard null-safe comparison and avoid relying on external libraries.

This is a valid and good answer. The code change to the StringUtils library would not have been necessary.
But now that it exists, we can proceed with the last change.

ParkGyeongTae (approved):
@eunhwa99 
Thank you as always for your contribution!
I noticed you updated StringUtils.equals to Strings.CS.equals. Could you also update the PR title and description to reflect this change?

eunaverse:
> @eunhwa99 Thank you as always for your contribution! I noticed you updated StringUtils.equals to Strings.CS.equals. Could you also update the PR title and description to reflect this change?

Thanks for your point! I modified it :)

Reamer:

PR #5051, opened 2025-08-27
[MINOR] Update java version check that only allows java 11 or newer

paul8263:
### What is this PR for?
Update java version check that only allows java 11 or newer

### What type of PR is it?

Hot Fix

### Todos
N/A

### What is the Jira issue?
N/A

Manually tested.

### Questions:

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -76,12 +76,9 @@ function check_java_version() {
     java_ver_output=$("${JAVA:-java}" -version 2>&1)
     jvmver=$(echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR==1 {print $2}' | cut -d\- -f1)
     JVM_VERSION=$(echo "$jvmver"|sed -e 's|^\([0-9][0-9]*\)\..*$|\1|')
-    if [ "$JVM_VERSION" = "1" ]; then
-        JVM_VERSION=$(echo "$jvmver"|sed -e 's|^1\.\([0-9][0-9]*\)\..*$|\1|')
-    fi
 
-    if [ "$JVM_VERSION" -lt 8 ] || { [ "$JVM_VERSION" -eq 8 ] && [ "${jvmver#*_}" -lt 151 ]; } ; then
-        echo "Apache Zeppelin requires either Java 8 update 151 or newer"
+    if [ "$JVM_VERSION" -lt 11 ]; then
+        echo "Apache Zeppelin requires either Java 11 or newer"
         exit 1;
     fi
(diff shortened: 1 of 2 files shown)


bin/common.sh
```diff
@@ -76,12 +76,9 @@ function check_java_version() {
...
     jvmver=$(echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR==1 {print $2}' | cut -d\- -f1)
     JVM_VERSION=$(echo "$jvmver"|sed -e 's|^\([0-9][0-9]*\)\..*$|\1|')
-    if [ "$JVM_VERSION" = "1" ]; then
-        JVM_VERSION=$(echo "$jvmver"|sed -e 's|^1\.\([0-9][0-9]*\)\..*$|\1|')
-    fi
 
-    if [ "$JVM_VERSION" -lt 8 ] || { [ "$JVM_VERSION" -eq 8 ] && [ "${jvmver#*_}" -lt 151 ]; } ; then
-        echo "Apache Zeppelin requires either Java 8 update 151 or newer"
+    if [ "$JVM_VERSION" -lt 11 ]; then
+        echo "Apache Zeppelin requires either Java 11 or newer"
```
pan3793, line 81: Actually, it will fail immediately even without this check. If the error message is clear enough, I think we don't need this check.

paul8263: Hi @pan3793 ,
I tried running Zeppelin with JDK 8 and got: Exception in thread "main" java.lang.UnsupportedClassVersionError: org/apache/zeppelin/utils/CommandLineUtils has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0.

How about moving the java version check at the beginning of zeppelin-daemon.sh, making it fail before actually launching the java process?

pan3793: it seems to be called by `function getZeppelinVersion`, can you add the check inside the function instead of `zeppelin-daemon.sh`?

paul8263: If we run zeppelin-daemon.sh start we would get Exception in thread "main" java.lang.UnsupportedClassVersionError: org/apache/zeppelin/server/ZeppelinServer has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0. So I think we should add the check both in zeppelin-daemon.sh and getZeppelinVersion in common.sh

pan3793 (approved):

ParkGyeongTae:

PR #5045, opened 2025-08-25
[ZEPPELIN-6296] Modify New UI when entering a note, the note name is displayed in the page title

dididy:
### What is this PR for?
**[Appropriate action - Classic UI]**

스크린샷 2025-08-25 오후 9 18 14

In Classic UI, the tab title is displayed as “Zeppelin” on the main page. However, when entering a note, the title changes to the note’s name (left tab: main, right tab: note view).

**[AS-IS]**
스크린샷 2025-08-25 오후 9 19 04

In contrast, in the New UI the page title is fixed as “Zeppelin” on all pages. It seems necessary to modify this to behave like the Classic UI.

**[TO-BE]**
스크린샷 2025-08-25 오후 9 41 57

I fixed it to work properly by using the existing @angular/platform-browser package’s "Title".

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* [[ZEPPELIN-6296](https://issues.apache.org/jira/browse/ZEPPELIN-6296)]

### Questions:
* Does the license files need to update? N
* Is there breaking changes for older versions? N
* Does this needs documentation? N

PR #5053, opened 2025-08-27
[ZEPPELIN-6301] Add tslint custom rule about ensure the order in constructor

dididy:
### What is this PR for?
https://github.com/apache/zeppelin/pull/5045#discussion_r2300816684

Following above PR, I received feedback from a committer suggesting that constructor parameters should be ordered as public → protected → private.
I found it deeply but there was no existing TSLint rule for this, So I implemented a custom rule. 

It works as intended, and I ran lint --fix with the new rule and confirmed that it was applied correctly. The modified files are also included in this PR. 

If you want to test it just follow below command.

```sh
// Base Dir: ./zeppelin/zeppelin-web-angular
npm run lint
```
or
```sh
// Base Dir: ./zeppelin/zeppelin-web-angular
npx tslint -c tslint.json 'src/**/*.ts' --fix
```

You can build the lint rule file using the command below

```sh
// Base Dir: ./zeppelin/zeppelin-web-angular/tslint-rules
npx tsc constructorParamsOrderRule.ts
```

### What type of PR is it?
Improvement

### What is the Jira issue?
* [[ZEPPELIN-6301](https://issues.apache.org/jira/browse/ZEPPELIN-6301)]

### Questions:

### DIFF
--- a/zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
+++ b/zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
@@ -1,0 +1,106 @@
+"use strict";
+var __extends = (this && this.__extends) || (function () {
+    var extendStatics = function (d, b) {
+        extendStatics = Object.setPrototypeOf ||
+            ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
+            function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
+        return extendStatics(d, b);
+    };
+    return function (d, b) {
+        extendStatics(d, b);
+        function __() { this.constructor = d; }
+        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+    };
+})();
+var __spreadArrays = (this && this.__spreadArrays) || function () {
+    for (var s = 0, i = 0, il = arguments.length; i < il; i++) s += arguments[i].length;
+    for (var r = Array(s), k = 0, i = 0; i < il; i++)
+        for (var a = arguments[i], j = 0, jl = a.length; j < jl; j++, k++)
+            r[k] = a[j];
+    return r;
+};
+exports.__esModule = true;
+var ts = require("typescript");
+var Lint = require("tslint");
+var Rule = /** @class */ (function (_super) {
+    __extends(Rule, _super);
+    function Rule() {
+        return _super !== null && _super.apply(this, arguments) || this;
+    }
+    Rule.prototype.apply = function (sourceFile) {
+        return this.applyWithFunction(sourceFile, walk);
+    };
+    Rule.FAILURE_STRING = "Constructor parameters should be ordered: public, protected, private";
+    return Rule;
+}(Lint.Rules.AbstractRule));
+exports.Rule = Rule;
+function walk(ctx) {
+    var checkNode = function (node) {
+        if (ts.isConstructorDeclaration(node)) {
+            var params_1 = node.parameters;
+            if (params_1.length <= 1)
+                return;
+            var rankMap_1 = { public: 0, protected: 1, private: 2, none: 3 };
+            var getModifierRank_1 = function (param) { return rankMap_1[getModifier(param)]; };
+            var lastRank = -1;
+            var needFix = false;
+            for (var i = 0; i < params_1.length; i++) {
+                var currentRank = getModifierRank_1(params_1[i]);
+                if (currentRank < lastRank) {
+                    needFix = true;
+                    break;
+                }
+                lastRank = currentRank;
+            }
+            if (needFix) {
+                var sourceText_1 = node.getSourceFile().text;
+                // 파라미터 조각 추출
+                var paramSlices_1 = params_1.map(function (p, idx) {
+                    // 현재 파라미터 시작
+                    var start = p.getStart();
+                    // 다음 파라미터가 있으면 그 직전까지 포함
+                    // (즉, 현재 파라미터 끝 ~ 다음 파라미터 시작 사이의 ,와 공백/주석까지 포함)
+                    var end = idx < params_1.length - 1 ? params_1[idx + 1].getStart() : p.getEnd();
+                    var text = sourceText_1.slice(start, end);
+                    return { node: p, text: text };
+                });
+                // 정렬
+                var sorted = __spreadArrays(paramSlices_1).sort(function (a, b) { return getModifierRank_1(a.node) - getModifierRank_1(b.node); });
+                var startLine = ctx.sourceFile.getLineAndCharacterOfPosition(node.parameters.pos).line;
+                var endLine = ctx.sourceFile.getLineAndCharacterOfPosition(node.parameters.end).line;
+                var isMultiLine_1 = startLine !== endLine;
+                var indent_1 = sorted[0].text.replace(sorted[0].text.trim(), '');
+                // 조합
+                var fixText = sorted.map(function (s, index) {
+                    console.log(s.text);
+                    if (index === paramSlices_1.length - 1) {
+                        return s.text.replace(',', '');
+                    }
+                    if (s.text.includes(',')) {
+                        return s.text;
+                    }
+                    if (!s.text.includes('\n') && isMultiLine_1) {
+                        console.log('IN??');
+                        return s.text + ',' + indent_1;
+                    }
+                    return s.text + ', ';
+                }).join("").trim();
+                var fix = Lint.Replacement.replaceFromTo(params_1[0].getStart(), params_1[params_1.length - 1].getEnd(), fixText);
+                ctx.addFailureAtNode(node, Rule.FAILURE_STRING, fix);
+            }
+        }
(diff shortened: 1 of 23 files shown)


zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.ts
tbonelee, line 1: Could you prettify this file and add a Apache license header at the top of the file?

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
+"use strict";
```
tbonelee: Apache license header should be added here.

tbonelee (commented):
Thank you for writing a custom lint rule.
I've requested a few minor changes: adding an Apache license header and formatting. Also, when running `npm run lint`, it seems that the `json-visualization` under `helium-vis-example` is failing the lint check, so please take a look at that as well.

While checking the `json-visualization` case, I noticed a potential edge case. To comply with the new rule, the parameter order was adjusted, but this caused the required parameter to come after an optional parameter, which could trigger a TS error. To handle such cases, it may be necessary to enforce a lexicographical order where required parameters always come before optional ones. (In Angular classes, parameters can also be declared with the `@Optional` decorator, so this could be a possible case.)

Additionally, I wonder if it might make sense to limit this rule to Angular classes. Since Angular class constructors are not directly invoked, the parameter order could be treated purely as a matter of readability within the class itself, similar to how member variables are ordered. For non-Angular classes that can be instantiated directly, though, it may not always be ideal for constructor arguments to strictly follow visibility order, as readability from the caller's perspective might differ. Do you think this distinction would be worth considering?

tbonelee:
And could you add a brief note on what to do when the rule .ts file is changed (e.g., compilation steps)?

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
...
+                var currentRank = getModifierRank_1(params_1[i]);
+                if (currentRank < lastRank) {
+                    needFix = true;
+                    break;
+                }
+                lastRank = currentRank;
+            }
+            if (needFix) {
+                var sourceText_1 = node.getSourceFile().text;
+                // 파라미터 조각 추출
```
ParkGyeongTae: Please remove the Korean comments. Use English comments instead, or just remove them if not necessary.

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
...
+                var paramSlices_1 = params_1.map(function (p, idx) {
+                    // 현재 파라미터 시작
```
ParkGyeongTae: Please remove the Korean comment (same request as above).

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
...
+                    var start = p.getStart();
+                    // 다음 파라미터가 있으면 그 직전까지 포함
+                    // (즉, 현재 파라미터 끝 ~ 다음 파라미터 시작 사이의 ,와 공백/주석까지 포함)
```
ParkGyeongTae: Please remove the Korean comment (same request as above).

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
...
+                    var end = idx < params_1.length - 1 ? params_1[idx + 1].getStart() : p.getEnd();
+                    var text = sourceText_1.slice(start, end);
+                    return { node: p, text: text };
+                });
+                // 정렬
```
ParkGyeongTae: Please remove the Korean comment (same request as above).

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.js
```diff
@@ -0,0 +1,106 @@
...
+                var sorted = __spreadArrays(paramSlices_1).sort(function (a, b) { return getModifierRank_1(a.node) - getModifierRank_1(b.node); });
+                var startLine = ctx.sourceFile.getLineAndCharacterOfPosition(node.parameters.pos).line;
+                var endLine = ctx.sourceFile.getLineAndCharacterOfPosition(node.parameters.end).line;
+                var isMultiLine_1 = startLine !== endLine;
+                var indent_1 = sorted[0].text.replace(sorted[0].text.trim(), '');
+                // 조합
```
ParkGyeongTae: Please remove the Korean comment (same request as above).

ParkGyeongTae (commented):
Please make a small change.

voidmatcha:
Thank you for the detailed review.

First, I removed build file(**constructorParamsOrderRule.js**) and updated `.gitignore` so that any build files under `tslint-rules/*.js` are not included. I also added a script in package.json to build the files inside tslint-rules folder, and ensured that this script is executed during postinstall script in package.json. In other words, when a user runs `postinstall`, `constructorParamsOrderRule.js` file will be generated, and the custom TSLint rule will run based on that file.

command are as follows:

```sh
npm run build:tslint-rules
```

Regarding the point you raised about `optional parameters` or the `@Optional` decorator, I updated the rule to handle those cases as well. I confirmed that **json-visualization.ts** is now fixed correctly. If other edge cases come up, I’ll make sure to address them.

As for non-Angular classes—are there any being used in this project? From what I’ve seen in **zeppelin-web-angular**, constructors are only used within Angular classes. Since Zeppelin is a large open-source project, I believe having a predictable and consistent order is more important than individual convenience. That’s also how I interpreted the original proposer(@ParkGyeongTae)’s intention behind suggesting this rule. I’d be curious to hear your thoughts on this point.

I’ve also completed prettify formatting and added the license header. Please let me know if there’s anything else needed.

tbonelee:
I appreciate your hard work. For non-Angular classes, I was initially considering possible future cases. However, since there are currently very few (maybe not any?) non-Angular classes, I agree that universal consistency for the prediction makes sense.

Lastly, before I re-review could you check the failing `run-tests-in-zeppelin-web-angular` job?

voidmatcha:
Considering future-proof design also seems important, so this is probably something worth discussing further later. Thank you for the quick response.

When running `npm run build:tslint-rules`, the error `TS2688: Cannot find type definition file for 'minimatch'` occurs. I determined that this is related to the internal workings of the tslint module and can be safely ignored. Therefore, I modified the `tsc` execution for building the `tslint-rules` folder to use `--types node` to bypass this error, and it now works correctly.  

I also discovered an issue where new lint rule did not work correctly when comments were placed above, and this has been fixed as well.

zeppelin-web-angular/package.json
```diff
@@ -101,4 +102,4 @@
       "pre-commit": "lint-staged"
     }
   }
-}
+}
```
tbonelee: New lines are needed before EOF

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.ts
```diff
@@ -48,9 +48,9 @@ function walk(ctx: Lint.WalkContext<void>) {
 
         // For keeping comment
         const paramSlices = params.map((p, idx) => {
-          const start = p.getStart();
+          let start = p.getFullStart();
```
tbonelee: Let's change `let` to `const`

zeppelin-web-angular/tslint-rules/constructorParamsOrderRule.ts
```diff
@@ -0,0 +1,119 @@
...
+        }
+        lastRank = currentRank;
+      }
+
+      if (needFix) {
+        const sourceText = node.getSourceFile().text;
+
+        // For keeping comment
+        const paramSlices = params.map((p, idx) => {
+          let start = p.getFullStart();
```
tbonelee: Did you change `getStart()` to `getFullStart()` to fix the `fixText` variable result? Just curious.

ParkGyeongTae:
@dididy 
Thanks for the contribution. I agree with your point. Right now, constructors are only used in Angular classes, so applying the rule everywhere makes sense. At first, I was thinking about all constructors, but if we add non-Angular classes in the future, we can revisit this.

voidmatcha:
The reason for using **getFullStart()** is that comments may be located above rather than to the right. 

```ts
    // tslint:disable-next-line:no-any <= it was edge case
    public visualization?: any, // right side comment <= previously it works well
```

I discovered this edge-case issue when `tslint:disable-*` comments were positioned at the top—the order changed but wasn’t moved together. While **getStart()** omits preceding comments and whitespace, **getFullStart()** includes them, allowing to handle this issue according to the intended behavior.

tbonelee (approved):

ParkGyeongTae (approved):
LGTM 👍

tbonelee:
Thanks. Merged into `master` and `branch-0.12`.

PR #5055, opened 2025-08-29
[ZEPPELIN-6307] Rename fieldname in NewNoteRequest class

celinayk:
### What is this PR for?
This PR renames the field 'name' to 'notePath' in the NewNoteRequest class to better reflect its purpose. The field actually represents the note's path rather than just a name, making the code more semantically clear and addressing the TODO comment that requested this change.

### What type of PR is it?
Refactoring

### Todos
* [x] - Rename field 'name' to 'notePath' in NewNoteRequest class

### What is the Jira issue?
[ZEPPELIN-6307](https://issues.apache.org/jira/browse/ZEPPELIN-6307)

### Questions:


ParkGyeongTae (approved):
I think this change makes the variable name clearer.

ParkGyeongTae:

PR #5058, opened 2025-08-30
[ZEPPELIN-6310] When paragraph to edit, word wrap doesn't work

dididy:
### What is this PR for?
**Description:**
You can check that in the Classic UI, word wrap is applied when editing an paragraph, but in the New UI it is not, which makes editing difficult. From a user’s perspective, the lack of word wrap can make editing both inconvenient and confusing, so this issue should be addressed. Also, in the case of .md files, word wrap is applied when the file is executed, but it is not applied when editing, which seems to cause an inconsistency.

https://github.com/user-attachments/assets/064d883d-0b3e-44cb-9448-bc00848bae78

**[AS-IS]**

https://github.com/user-attachments/assets/3084b82a-a8fd-492f-8ef0-3fa421f92e99

**[TO-BE]**

https://github.com/user-attachments/assets/8b8db68b-9373-41d1-b091-57599b34a4f2

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* [[ZEPPELIN-6310](https://issues.apache.org/jira/browse/ZEPPELIN-6310)]

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
@@ -157,6 +157,7 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
       scrollBeyondLastLine: false,
       contextmenu: false,
       matchBrackets: 'always',
+      wordWrap: 'on',
       scrollbar: {
         handleMouseWheel: false,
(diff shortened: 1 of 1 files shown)


tbonelee (approved):
LGTM. For others' reference, linking the monaco-editor v0.30.1 typings:
https://github.com/microsoft/monaco-editor/blob/5a7ba61be909ae9e4889768a3453ebb0dec392e2/monaco-editor/typedoc/monaco.d.ts#L2927

tbonelee:

PR #5060, opened 2025-08-31
[ZEPPELIN-6312] Remove configuration WS usage in frontend

seung-00:
### What is this PR for?
This PR removes usage of the configuration API via WebSocket in the frontend.  
The configuration API exists in both WS and REST, but unlike other APIs, it doesn’t seem to need two separate channels.  
Maintaining both surfaces increases maintenance cost, so this PR unifies them by keeping only the REST API.  

### What type of PR is it?
Improvement

### Todos
* [ ] - [ZEPPELIN-6313](https://issues.apache.org/jira/browse/ZEPPELIN-6313) Drop configuration WS endpoint in backend

### What is the Jira issue?
* [ZEPPELIN-6312](https://issues.apache.org/jira/browse/ZEPPELIN-6312)

* Check the configuration page (/configuration)  
* Check the notebook page (/notebook/{notebook_id})

### Questions:
* Does the license files need to update? M


zeppelin-web-angular/src/app/services/configuration.service.ts
```diff
@@ -13,18 +13,29 @@
 import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 
+import { retry } from 'rxjs/operators';
+import { Configuration } from '../../../projects/zeppelin-sdk/src';
```
tbonelee: Could you import this from `@zeppelin/sdk` to align with other import statements?

zeppelin-web-angular/src/app/share/note-import/note-import.component.ts
```diff
@@ -101,8 +101,8 @@ export class NoteImportComponent extends MessageListenersManager implements OnIn
   }
 
   constructor(
+    private configurationService: ConfigurationService,
```
tbonelee: Check this lint error in CI please. (related to #5053)
```
ERROR: /home/runner/work/zeppelin/zeppelin/zeppelin-web-angular/src/app/share/note-import/note-import.component.ts:103:3 - Constructor parameters should be ordered: public, protected, private
```

zeppelin-web-angular/src/app/share/note-import/note-import.component.ts
```diff
@@ -19,8 +19,8 @@ import { NzUploadFile } from 'ng-zorro-antd/upload';
 
 import { MessageListener, MessageListenersManager } from '@zeppelin/core';
 import { MessageReceiveDataTypeMap, OP, SendNote } from '@zeppelin/sdk';
+import { ConfigurationService } from '@zeppelin/services';
```
tbonelee: When I ran the app locally, it seems this barrel import may have caused a circular dependency, which resulted in `ConfigurationService` being undefined. I initially encountered related errors on the login page.

I am trying to fix overall circular dependencies in #5065, so in here we could import from the specific paths like below.
```ts
import { ConfigurationService } from '@zeppelin/services/configuration.service';
```

tbonelee (changes requested):
Replacing WS with REST API seems reasonable since it reduce complexity.
As the login page is currently not working properly, that issue should be fixed as well.

seung-00:
I’ve updated according to the review comments.
But I found one issue: the configurations WS API has no auth check, while the REST API does. So users without permission can still access the configs via WS.

Removing WS right now would break things, so I’ll fix this in another PR before merging this one.

@tbonelee

tbonelee:
image
Could you also take a look at this? It seems that fetching configurations in `ConfigurationService.initialize()` when the user is not logged in might be causing some issues.

PR #5063, opened 2025-08-31
[ZEPPELIN-6309] Improve method by replacing JsonObject parameter

celinayk:
### What is this PR for?
Refactored the `convertPermissionsFromUsersToOwners` method in InterpreterSetting.java to improve code readability and maintainability by separating JSON parsing logic from business logic, addressing the TODO comment that identified this as "ugly code".

### What type of PR is it?
Refactoring

### Todos
  * [x] - Refactor convertPermissionsFromUsersToOwners method to remove JsonObject parameter
  * [x] - Extract JSON parsing logic into separate static helper method
  * [x] - Update all callers to use new method signature
  

### What is the Jira issue?
[ZEPPELIN-6309](https://issues.apache.org/jira/browse/ZEPPELIN-6309)

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterInfoSaving.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterInfoSaving.java
@@ -64,7 +64,7 @@ public class InterpreterInfoSaving implements JsonSerializable {
         for (InterpreterSetting interpreterSetting : infoSaving.interpreterSettings.values()) {
           JsonObject interpreterSettingJson = jsonObject.getAsJsonObject("interpreterSettings")
               .getAsJsonObject(interpreterSetting.getId());
-          List<String> users = InterpreterSetting.extractUsersFromJsonObject(interpreterSettingJson);
+          List<String> users = InterpreterSetting.extractUsersFromJsonString(interpreterSettingJson.toString());
           interpreterSetting.convertPermissionsFromUsersToOwners(users);
         }
       }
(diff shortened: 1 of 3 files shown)


zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterSetting.java
```diff
@@ -1016,22 +1016,29 @@ public void run() {
...
-            this.option.getOwners().add(user.getAsString());
+        JsonArray usersArray = option.getAsJsonArray("users");
+        if (usersArray != null) {
+          for (JsonElement userElement : usersArray) {
+            users.add(userElement.getAsString());
           }
         }
       }
     }
+    return users;
```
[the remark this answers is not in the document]
celinayk: Good catch! Updated to use LinkedList for consistency with the rest of the class. This maintains consistency with the owners field initialization on line.

zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterInfoSaving.java
```diff
@@ -62,9 +62,10 @@ public static InterpreterInfoSaving loadFromFile(Path file) throws IOException {
 
       if (infoSaving != null && infoSaving.interpreterSettings != null) {
         for (InterpreterSetting interpreterSetting : infoSaving.interpreterSettings.values()) {
-          interpreterSetting.convertPermissionsFromUsersToOwners(
-                  jsonObject.getAsJsonObject("interpreterSettings")
-                          .getAsJsonObject(interpreterSetting.getId()));
+          JsonObject interpreterSettingJson = jsonObject.getAsJsonObject("interpreterSettings")
```
Reamer, line 65: Are you sure this is what is meant by “ugly” here?
I think we should use GSON here to create objects directly.
```
GSON.fromJson(json, myFanceClass.class);
```

celinayk: > Are you sure this is what is meant by “ugly” here? I think we should use GSON here to create objects directly.
> 
> ```
> GSON.fromJson(json, myFanceClass.class);
> ```

Thank you for the feedback! You're absolutely right. I've refactored the code to use GSON.fromJson() directly
  instead of JsonObject parameter.

ParkGyeongTae (approved):
LGTM 👍

ParkGyeongTae:

PR #5064, opened 2025-09-01
[ZEPPELIN-6197/6229] Fixed the shortcuts in the New UI so that they match the Classic UI and work correctly

dididy:
### What is this PR for?
Unlike the Classic UI, the New UI either lacked implementations for several shortcuts or had shortcuts assigned differently from the Classic UI. I consider this a significant issue, and after addressing other related derivative issues, I am finally submitting it.

#### d96ae5bde - Remove the top margin in the editor search
[AS-IS]
스크린샷 2025-09-01 오후 11 26 14
[TO-BE]
스크린샷 2025-09-01 오후 11 26 23

Fixed an issue where the top area increased by 33px when search was activated. <- [referece](https://github.com/microsoft/monaco-editor/blob/5a7ba61be909ae9e4889768a3453ebb0dec392e2/monaco-editor/typedoc/monaco.d.ts#L3473)

#### 0d24eb420 - Handle focus on PARAGRAPH_MOVED receive
When performing "Move Paragraph Up" or "Move Paragraph Down," the paragraph moves up or down after receiving `PARAGRAPH_MOVED`. During this process, focus was being skipped, so I added it.

#### 1d3d294bc - Overriding shortcuts conflict with the editor
In the New UI, we use the Monaco editor, where reserved shortcuts had to be explicitly overridden. I added shortcut functionality for four actions: "Toggle Editor", "Cut the Line", "Paste the Line", and "Search Inside the Code".

#### 57f45e1e8 - Make action bar searchCode accessible to siblings
There are `paragraph.component.ts` and `action-bar.component.ts` that share `notebook.component.ts` as their parent. In the Classic UI, "Find in Code" is a shortcut that activates the search button in the action-bar. Although this functionality is currently marked as [TODO](https://github.com/dididy/zeppelin/blob/fix/ZEPPELIN-6197/6229/zeppelin-web-angular/src/app/pages/workspace/notebook/action-bar/action-bar.component.ts#L220-L222)([ZEPPELIN-6279](https://issues.apache.org/jira/secure/RapidBoard.jspa?rapidView=632&view=detail&selectedIssue=ZEPPELIN-6279)) in the New UI, I set up the connection by adding a Template Reference Variable and Event Binding in `notebook.component.html`. With this setup, `paragraph.component.ts` can invoke the **searchCode()** method of `action-bar.component.ts`.

#### b027a33 - Ensure shortcut keys work correctly
image

All shortcut orders have been updated based on the Classic UI. There was logic to handle key bindings differently for macOS, but upon review, it was unnecessary and has been removed. On macOS, typing a combination of Alt and an English letter can produce a different character, so to ensure correct behavior, this resulting character combination is included in the shortcut handling. I added several actions that were not yet implemented.

The distinction between command mode and edit mode for shortcuts was meaningless, so it was removed. The original implementation seemed intended to mimic a vi-like behavior, but this would have made the shortcut system different from the Classic UI.

In the Classic UI, "Auto Completion" had a shortcut to trigger it manually. In the New UI, auto completion is triggered automatically when typing, so "Auto Completion" shortcut was unnecessary and has been removed in this PR.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* [[ZEPPELIN-6197](https://issues.apache.org/jira/browse/ZEPPELIN-6197)]
* [[ZEPPELIN-6229](https://issues.apache.org/jira/browse/ZEPPELIN-6229)]

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
@@ -109,6 +109,72 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
     }
   }
 
+  // Handle Ctrl+Alt+E: Toggle editor show/hide
+  handleToggleEditorShow() {
+    this.toggleEditorShow.emit();
+  }
+
+  // Handle Ctrl+K: Cut current line to clipboard
+  async handleCutLine() {
+    if (!this.editor) {
+      return;
+    }
+
+    const position = this.editor.getPosition();
+    const model = this.editor.getModel();
+    if (!position || !model) {
+      return;
+    }
+
+    const lineNumber = position.lineNumber;
+    const lineContent = model.getLineContent(lineNumber);
+
+    if (!lineContent) {
+      return;
+    }
+
+    await navigator.clipboard.writeText(lineContent);
+
+    this.editor.executeEdits('cut-line', [
+      {
+        range: new monaco.Range(lineNumber, 1, lineNumber, lineContent.length + 1),
+        text: '',
+        forceMoveMarkers: true
+      }
+    ]);
+  }
+
+  // Handle Ctrl+Y: Paste from clipboard at current position
+  async handlePasteFromClipboard() {
+    if (!this.editor) {
+      return;
+    }
+
+    const text = await navigator.clipboard.readText();
+    const position = this.editor.getPosition();
+    if (position) {
+      this.editor.executeEdits('my-source', [
+        {
+          range: new monaco.Range(position.lineNumber, position.column, position.lineNumber, position.column),
+          text: text,
+          forceMoveMarkers: true
+        }
+      ]);
+    }
+  }
+
+  // Handle Ctrl+S: Show find widget
+  handleShowFind() {
+    if (this.editor) {
+      this.editor.getAction('actions.find').run();
+
+      // Focus on the find widget input field
+      const findInput = document.querySelector('.find-widget .input') as HTMLInputElement;
+      findInput.focus();
+      findInput.select();
+    }
+  }
+
   initializedEditor(editor: IEditor) {
     this.editor = editor as IStandaloneCodeEditor;
     this.editor.addCommand(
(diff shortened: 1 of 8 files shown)


zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
```diff
@@ -382,6 +388,41 @@ export class NotebookParagraphComponent extends ParagraphBase implements OnInit,
...
+        newIndex = i + 1;
+        break;
+      }
+    }
+
+    if (newIndex < 0 || newIndex >= this.note.paragraphs.length) {
+      return;
+    }
+    this.messageService.moveParagraph(this.paragraph.id, newIndex);
+  }
```
tbonelee: Could you make the “find new index” logic consistent across `moveCursorUp`, `moveCursorDown`, `moveParagraphUp`, and `moveParagraphDown`?
Any approach is fine. If you choose the declarative style: given the precondition that the corresponding paragraph exists in `this.note.paragraphs`, you can ignore the edge case where `findIndex()` returns -1 (which would make `newIndex` become 0, even though 0 wouldn't be semantically meaningful here). Nevertheless, it might be better to handle that edge case for robustness against future changes.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
```diff
@@ -576,6 +543,75 @@ export class NotebookParagraphComponent extends ParagraphBase implements OnInit,
...
+            this.removeParagraph();
+            break;
+          case ParagraphActions.InsertAbove:
+            this.insertParagraph('above');
+            break;
+          case ParagraphActions.InsertBelow:
+            this.insertParagraph('below');
+            break;
+          case ParagraphActions.InsertCopyOfParagraphBelow:
+            this.cloneParagraph('below');
```
tbonelee, line 551: This method(`this.cloneParagraph()`) doesn't seem to be working as expected. We could address this issue in a separate PR.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -120,6 +121,71 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
...
+          start(options: { forceRevealReplace: boolean; seedSearchStringFromSelection: boolean }): void;
+          getFindInputFocusElement(): HTMLElement | null;
+        };
+        if (controller) {
+          controller.start({
+            forceRevealReplace: false,
+            seedSearchStringFromSelection: true
+          });
+          controller.getFindInputFocusElement()?.focus();
+        }
```
tbonelee: Could you explain how this works to me?

zeppelin-web-angular/src/app/services/shortcut.service.ts
```diff
@@ -16,56 +16,59 @@ import { EventManager } from '@angular/platform-browser';
...
-  SwitchEnable = 'Paragraph:SwitchEnable',
-  Link = 'Paragraph:Link'
+  Clear = 'Paragraph:Clear',
+  Link = 'Paragraph:Link',
+  ReduceWidth = 'Paragraph:ReduceWidth',
+  IncreaseWidth = 'Paragraph:IncreaseWidth',
+  FindInCode = 'Paragraph:FindInCode'
 }
 
 export const ShortcutsMap = {
```
tbonelee, line 44: Could you add some comments explaining why additional non-ASCII combinations are needed? It may help future developers better understand this code.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
tbonelee, line 1: How about extracting a method for each `addCommand` callback? It would make what each command callback does clearer and make the code easier to grep for future developers.

tbonelee (commented):
Thanks for taking care of this important task. All of the key bindings you added worked well in my local environment. Perhaps next time we could split this kind of work into separate issues (e.g., fixing just the key bindings themselves and fixing the behavior of each key binding).

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
voidmatcha: 3759b33

I’ve updated all four functions to use a declarative approach. The logic in `moveParagraphUp()` and `moveParagraphDown()` was previously imperative because that’s how it was implemented in the Classic UI.

[moveParagraphUp > zeppelin-web/src/app/notebook/notebook.controller.js#L1438-L1447](https://github.com/dididy/zeppelin/blob/fix/ZEPPELIN-6197/6229/zeppelin-web/src/app/notebook/notebook.controller.js#L1438-L1447
)
[moveParagraphDown > zeppelin-web/src/app/notebook/notebook.controller.js#L1462-L1472
](https://github.com/dididy/zeppelin/blob/fix/ZEPPELIN-6197/6229/zeppelin-web/src/app/notebook/notebook.controller.js#L1462-L1472)

I’ve confirmed that both functions work correctly after converting them to a declarative approach.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
voidmatcha: #5044

If the current issue is the one where a blank paragraph appears when cloning a paragraph and only shows the cloned content correctly after a refresh, a PR for this had already been opened. The points raised in the previous review have been addressed, and I’ve also updated the code to handle additional cases.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
voidmatcha: f75faa9

This code is meant to open the editor's find widget and focus it, but it wasn’t working correctly. The `getFindInputFocusElement()` method doesn’t work in current old versions of the Monaco editor, so I changed it to use a query selector to get the element and manually trigger the focus event.

zeppelin-web-angular/src/app/services/shortcut.service.ts
voidmatcha: 13f201a

I’ve added it.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
voidmatcha: 2f5838f

I’ve added it. Thanks for always giving such helpful reviews. I’ll make sure to reflect any other feedback if needed.

voidmatcha:
I corrected the keyboard shortcuts in the dropdown that appears when clicking the settings icon at the top-right of a paragraph; they had differed from the original. The same issue existed in the Classic UI and has been fixed there as well.

e787a62
[NewUI: As-Is]
스크린샷 2025-09-06 오후 3 28 57

[NewUI: To-Be]
스크린샷 2025-09-06 오후 3 28 41

bd4952b
[Classic UI: As-IS]
스크린샷 2025-09-06 오후 3 12 05
[Classic UI: To-Be]
스크린샷 2025-09-06 오후 3 11 50

tbonelee (approved):

PR #5065, opened 2025-09-02
[ZEPPELIN-6315] Add a lint rule to prevent circular imports in zeppelin-web-angular

tbonelee:
### What is this PR for?

Circular imports in JS/TS can make imported symbols unexpectedly `undefined` at runtime and also confuse build and tooling. This PR introduces a TSLint rule to detect circular imports and refactors imports to prevent them.

- Added `tslint-no-circular-imports` and enabled the rule.
- Fixed explicit circular-import violations.
- Standardized import conventions and layering rules to avoid regressions.

#### Import conventions

To keep dependencies simple and acyclic:

1. Cross-module improts -> use the modules's barrel
Import from the module entry (its public API), not deep files.
```ts
// good
import { Foo } from '@zeppelin/some-module';
// bad
import { Foo } from '@zeppelin/some-module/feature/foo';
```
- Rationale: reduces tight coupling and ad-hoc interweaving between modules.

2. Intra-module imports -> use relative paths
Inside the same module, import relatively instead of via that module's barrel.
```ts
// good
import { Bar } from './bar/bar.component';
// bad
import { Bar } from '@zeppelin/this-module';
```
- Rationale: prevents barrel -> implementation -> barrel loops that often create cycles.

> Terminology
> A "module" here means a directory published via a single alias like @zeppelin/{module} (one slash).

#### Layering rules

Enforce a single direction of dependencies. This keeps cycles out by construction.
This interpretation aims to minimize disruption to the existing structure. If you have alternative proposals, please let me know.

- `core`: framework-agnostic types, tokens, and utilities.
  - Can be used by anyone.
  - Must not depend on `services`, `pages`, or `share`
- `services`: non-UI logics
  - May depend on `core`.
  - Must not depend on `share` or `pages`.
- `pages`: May depend on `services`, `share`, and `core`.
- `share` : presentational componets, directives
  - May depend on `core` and `service`

### What type of PR is it?
Refactoring

### What is the Jira issue?
[ZEPPELIN-6315](https://issues.apache.org/jira/browse/ZEPPELIN-6315)

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/visualizations/public-api.ts
+++ b/zeppelin-web-angular/src/app/visualizations/public-api.ts
@@ -10,11 +10,11 @@
  * limitations under the License.
  */
 
-export * from './g2.config';
-export * from './visualization.module';
-export * from './table/table-visualization';
 export * from './area-chart/area-chart-visualization';
 export * from './bar-chart/bar-chart-visualization';
+export * from './g2.config';
 export * from './line-chart/line-chart-visualization';
 export * from './pie-chart/pie-chart-visualization';
 export * from './scatter-chart/scatter-chart-visualization';
+export * from './table/table-visualization';
+export * from './visualization.module';
--- a/zeppelin-web-angular/src/app/visualizations/visualization.module.ts
+++ b/zeppelin-web-angular/src/app/visualizations/visualization.module.ts
@@ -28,20 +28,14 @@ import { NzRadioModule } from 'ng-zorro-antd/radio';
 import { NzTableModule } from 'ng-zorro-antd/table';
 import { NzTagModule } from 'ng-zorro-antd/tag';
 
-import { AreaChartVisualization } from './area-chart/area-chart-visualization';
 import { AreaChartVisualizationComponent } from './area-chart/area-chart-visualization.component';
-import { BarChartVisualization } from './bar-chart/bar-chart-visualization';
 import { BarChartVisualizationComponent } from './bar-chart/bar-chart-visualization.component';
 import { VisualizationPivotSettingComponent } from './common/pivot-setting/pivot-setting.component';
 import { VisualizationScatterSettingComponent } from './common/scatter-setting/scatter-setting.component';
 import { VisualizationXAxisSettingComponent } from './common/x-axis-setting/x-axis-setting.component';
-import { LineChartVisualization } from './line-chart/line-chart-visualization';
 import { LineChartVisualizationComponent } from './line-chart/line-chart-visualization.component';
-import { PieChartVisualization } from './pie-chart/pie-chart-visualization';
 import { PieChartVisualizationComponent } from './pie-chart/pie-chart-visualization.component';
-import { ScatterChartVisualization } from './scatter-chart/scatter-chart-visualization';
 import { ScatterChartVisualizationComponent } from './scatter-chart/scatter-chart-visualization.component';
-import { TableVisualization } from './table/table-visualization';
 import { TableVisualizationComponent } from './table/table-visualization.component';
 
 const VisualizationComponents = [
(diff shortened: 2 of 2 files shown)


zeppelin-web-angular/src/app/pages/workspace/workspace.component.ts
```diff
@@ -11,11 +11,11 @@
...
-import { isRecord } from '@zeppelin/utility/type-utility';
+import { publishedSymbol } from '@zeppelin/core';
+import { isRecord } from '@zeppelin/utility';
 import { Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 
-import { publishedSymbol } from '@zeppelin/core/paragraph-base/published';
 import { HeliumManagerService } from '@zeppelin/helium-manager';
 import { MessageService } from '@zeppelin/services';
 import { setTheme } from '@zeppelin/visualizations/g2.config';
```
voidmatcha: I think `src/app/visualizations` could also be imported as `@zeppelin/visualizations`.  
Most of the folders under `src/app` are already set up with `public-api.ts` and `index.ts`, which allows importing directly from that folder depth. I believe we could apply the same approach here as well (though I wasn’t aware of the existing context behind this setup, so I’d like to be cautious).  

https://github.com/dididy/zeppelin/commit/ce0b4b3d5417df8d5301c71a00fd20bfc7c85284

You can see an example of this in the commit above.  

I’m also aware that `projects/zeppelin-visualization` is already imported as `@zeppelin/visualization`. Since the naming might feel a bit similar, it could cause some confusion—but I’d like to suggest handling it this way for consistency with the existing convention. If there was another reason behind it, it might not need to be changed.

voidmatcha:
https://github.com/dididy/zeppelin/commit/044685dd0888c5eea42253133caa579d5cbe6e8a

There’s a weird case where imports go one level deeper than they should. This could also be fixed in this PR. You can check it in the above commit.

---

It’s just a simple suggestion, but for `index.ts` files that gather exports, it might be better to sort them alphabetically. What do you think? There should be a TSLint rule for this. It’s applied here, but there are many places overall where it’s not yet enforced.

---

`note-action.service.ts` also seems to have been moved appropriately in context. This PR helped me better understand the roles within the `zeppelin-web-angular`. The criteria for using SDK messages versus `message.service`, and situations where they were mixed, were a bit confusing, but they seem well clarified now. Thank you for creating a good PR and contributing.

tbonelee:
Thanks for the review!
I fixed the imports that were unnecessarily going up and down in the path.
I also agree with your suggestion about sorting exports alphabetically, so I reordered them manually.
It seems the current TSLint rule (`ordered-imports`) only applies to imports, not exports. Perhaps we could create or find a simple rule for exports in a separate issue.

zeppelin-web-angular/src/app/pages/workspace/workspace.component.ts
tbonelee: Thanks for pointing out the missing parts! I cherry-picked your commit and made some modifications.
Since the newly added elements in the `VisualizationModule` declarations were neither Angular components, directives, nor pipes, I reverted those. If I still missed anything, please feel free to let me know.

I also agree that the similarity in naming could cause some confusion. We might consider further refactoring in later PRs.

voidmatcha: Thank you for addressing the review comments. I haven’t checked everything, but the current changes seem to be working well.

https://issues.apache.org/jira/browse/ZEPPELIN-6325

I’ve opened an issue regarding enforcing alphabetical order in the aggregated `index.ts` exports.

tbonelee: Thanks for taking care of the follow-up issue 👍

tbonelee:
Thanks, I've merged this into `master` and `branch-0.12`.

PR #5067, opened 2025-09-04
[ZEPPELIN-6317] Fix responsive layout issue of the control action bar in New UI

dididy:
### What is this PR for?
In the Classic UI, when the window width was reduced, the top control bar area expanded vertically without breaking and did not overlap with paragraphs. However, in the New UI, the control bar layout breaks, its height is not flexible, and due to its fixed height, some paragraph buttons become unclickable.

https://github.com/user-attachments/assets/117660db-ffae-41b6-a4ac-acec5b26389e

**[AS-IS]**

https://github.com/user-attachments/assets/228ee02d-e516-4db5-bd0c-5bcb79fabf5a

**[TO-BE]**

https://github.com/user-attachments/assets/a7bfee29-8878-4c45-9e85-ddf7532ab67a

### What type of PR is it?
Bug Fix
Improvement

### What is the Jira issue?
* [[ZEPPELIN-6317](https://issues.apache.org/jira/browse/ZEPPELIN-6317)]

### Questions:


tbonelee (approved):
LGTM

PR #5070, opened 2025-09-07
[ZEPPELIN-6319] Fixing unsupported wildcard path expansion in zeppelin-web-angular

tbonelee:
### What is this PR for?
Windows cmd and does not support wildcard path expansion, which causes errors like the one below.
image

This PR fixes the issue by replacing the glob pattern path argument with a specific `tsconfig.json` file for `tslint-rules/`.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6319

### Questions:


dididy (approved):
Thank you for resolving the issue that stemmed from the changes [I made](https://ci.appveyor.com/project/ApacheSoftwareFoundation/zeppelin/builds/52656648#L4292).  👍👍

ParkGyeongTae:

PR #5074, opened 2025-09-14
[ZEPPELIN-6327] Fix wrong order of the arguments for Visualization components

tbonelee:
### What is this PR for?
Visualizations are not working because the constructor arguments were passed in the wrong order.
Since using `any` made this error hard to detect, I replaced it with specific types.
(Related to #5053)

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6327

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
@@ -16,7 +16,6 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
-  ComponentFactoryResolver,
   EventEmitter,
   Injector,
   Input,
@@ -52,9 +51,14 @@ import {
   HeliumClassicVisualizationConstructor,
   HeliumVisualizationBundle
 } from '@zeppelin/interfaces';
-import { DynamicTemplate, HeliumService, NgZService, RuntimeCompilerService } from '@zeppelin/services';
-import { ClassicVisualizationService } from '@zeppelin/services/classic-visualization.service';
-import { TableData, Visualization } from '@zeppelin/visualization';
+import {
+  ClassicVisualizationService,
+  DynamicTemplate,
+  HeliumService,
+  NgZService,
+  RuntimeCompilerService
+} from '@zeppelin/services';
+import { TableData, Visualization, VisualizationConstructor } from '@zeppelin/visualization';
 import {
   AreaChartVisualization,
   BarChartVisualization,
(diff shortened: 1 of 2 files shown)


voidmatcha:
I didn’t realize there could be such an edge case. From a quick look, I now understand that Angular allows calling constructors dynamically in that way, and in such cases, the order of the parameters must match the order in the constructor. From my quick review, it seems that there are no side effects from the changes I made that are similar to the issue caused here by my changes.(#5053) But just to be safe, I’ll take a closer look.

One thing I’m curious about: I understand that `visualizationItem.componentFactoryResolver` was removed because it’s no longer needed in Angular 9. Is that correct? It looks like it’s still being used in **visualization-component-portal.ts** and **json-visualization.ts**, and based on my review, that’s probably because the project is still using Angular 8. I was wondering if there’s a plan to upgrade to Angular 9, and if not, what challenges are preventing it.

tbonelee:
I missed that `componentFactoryResolver` could actually be passed into `JsonVisualization`. The reason I removed it was because no component in `src/app/visualizations` seemed to be using it, so I thought it was unused. For `JsonVisualization`, however, it seems better to restore it.

That said, I was also considering whether we should remove the new Helium visualization interface that relies on Angular's dynamic rendering. My reasoning is twofold:

1. The Helium visualization packages currently published on npm are still based on the old UI (AngularJS) visualization class, so I don't think the new interface is practically in use. (`JsonVisualization` exists, but I see it more as an example than something used in practice.)
2. Even if we do provide a new Helium visualization interface, I think it would be better if it could be implemented without depending on Angular. That way, we could reduce framework-specific dependencies in the future.

So my proposal is:
- Add an adapter for the old Helium visualization packages (#5006).
- Keep the new visualization logic only for the default visualizations for now (`src/app/visualizations`)
- If a new Helium visualization interface is needed later, provide a framework-agnostic class instead of one tied to Angular.

What do you think about this direction?

zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
```diff
@@ -198,6 +200,7 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit,
   }
 
   private handleVisualizationBundles(bundles: HeliumVisualizationBundle[]): void {
+    console.log('bundles', bundles);
```
voidmatcha: It looks like a console.log was accidentally left in.

Sorry for the late reply. If it’s not actually being used, I agree that gradually removing Helium visualization interface that relies on Angular's dynamic rendering makes sense. I fully agree with reducing Angular dependencies, and the direction you suggested sounds good to me.

dididy (approved):
LGTM

tbonelee:

tbonelee:

PR #5075, opened 2025-09-14
[ZEPPELIN-6328] Enable strict template modes in zeppelin-web-angular

tbonelee:
### What is this PR for?
Type checks in template checks are not fully enabled right now, which makes it harder to catch type errors.
Enabling strict type checks in Angular templates would help surface these errors at compile time.

I enabled `strictTemplates` and `fullTemplateTypeCheck` options in tsconfig files.
Then I ran `npm run build` and fixed those type errors.

### What type of PR is it?
Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6328

### Questions:


zeppelin-web-angular/src/app/pages/workspace/interpreter/item/item.component.html
```diff
@@ -125,11 +125,11 @@ <h3 class="form-title">Option</h3>
       [nzDisabled]="interpreterRunningOption === runningOptionMap.globallyModeName || mode === 'view'"
     >
       <span *ngIf="interpreterRunningOption === runningOptionMap.perUserModeName; else perNote">
-        {{ optionFormGroup.get('perUser').value }}
+        {{ optionFormGroup.get('perUser')?.value || '' }}
```
tbonelee, line 128: We can’t use strictly typed forms yet, but we’ll be able to after upgrading to Angular 14:
https://angular.dev/guide/forms/typed-forms

zeppelin-web-angular/src/app/pages/workspace/notebook-repos/item/item.component.ts
```diff
@@ -73,6 +73,10 @@ export class NotebookRepoItemComponent implements OnChanges {
     this.settingFormArray = this.fb.array(controls);
   }
 
+  getSettingControl(index: number): FormControl {
+    return this.settingFormArray.at(index) as FormControl;
```
tbonelee, line 77: An assertion for `AbstractControl` to `FormControl`.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/control/control.component.ts
tbonelee, line 1: I've removed two way bindings for `@Input` properties in this component (.ts file and .html file)

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.html
```diff
@@ -40,26 +40,27 @@
...
-      [(runOnSelectionChange)]="paragraph.config.runOnSelectionChange"
-      (tableHideChange)="commitParagraph()"
+      [enabled]="paragraph.config.enabled"
+      [tableHide]="paragraph.config.tableHide"
+      [colWidth]="paragraph.config.colWidth"
+      [titleShow]="paragraph.config.title"
+      [fontSize]="paragraph.config.fontSize"
+      [showLineNumbers]="paragraph.config.lineNumbers"
+      [editorHide]="paragraph.config.editorHide"
+      [runOnSelectionChange]="paragraph.config.runOnSelectionChange"
```
voidmatcha, line 50:
```suggestion
      *ngIf="paragraph.config?.title"
      [value]="paragraph?.title || ''"
      ...
      [something]="paragraph.config?.something"
      ...
      [config]="configs[i] ?? {}"
      ...
```
How about revising it like this for the optional case?

zeppelin-web-angular/src/app/share/page-header/page-header.component.ts
```diff
@@ -21,7 +21,7 @@ import { InputBoolean } from 'ng-zorro-antd/core/util';
 })
 export class PageHeaderComponent implements OnInit {
   @Input() title!: string;
-  @Input() description!: string | TemplateRef<void>;
+  @Input() description!: TemplateRef<void>;
```
voidmatcha, line 24: https://github.com/apache/zeppelin/pull/5075/files#diff-97564afdcc91a75edcc8c7cc380cee9de3ef14dd50ddb4e874becccb7423a15cR19

Wouldn’t this part(description) in page-header.component.html need to ensure that the value of the `ngTemplateOutlet` attribute (**extra**) converges to null?

zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
```diff
@@ -96,11 +89,8 @@ export class TableVisualizationComponent implements OnInit {
     this.filterRows();
   }
 
-  onSortChange(type: 'descend' | 'ascend' | string, key: string): void {
-    const opt = this.colOptions.get(key);
-    if (!opt) {
-      throw new Error('opt is not found');
-    }
+  onSortChange(type: 'descend' | 'ascend' | string | null, key: string): void {
```
voidmatcha:
```suggestion
  onSortChange(key: string, type?: 'descend' | 'ascend'): void {
```
Was the `null` added because of added tsconfig option? With the above modification, and by adjusting the parameter order in [`nzSortChange` within **table-visualization.component.html**](https://github.com/apache/zeppelin/blob/5734012572e5b1c7c5a76d50da75e6d4f473e94f/zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.html#L49) as [this](https://github.com/apache/zeppelin/pull/5075/files#diff-da0b297d95609bb39dfce32ab16e7d6a9cb24adf5eb8897891bb5ab231f73309R77), it seems cleaner — though it needs to be confirmed whether this actually passes the type check.

tbonelee: Could you please check the attached commit links (in this comment and other comments)?
They don't seem to be working on my end.

voidmatcha: I thought it was possible to share a link to a specific line in the commit, but it seems that’s not the case. I’ve updated the link.

tbonelee: I tried your suggestion, but `nzSortChange` emits an event of following type, so the handler needs to accept `null`
```
export declare type NzTableSortOrder = string | 'ascend' | 'descend' | null;
```
Instead, I imported this type and used it in `onSortChange`. What do you think?

zeppelin-web-angular/src/app/share/page-header/page-header.component.ts
tbonelee: I thought `description` was a required property, so the component caller should provide it with a defined type (`TemplateRef<void>` type), ensuring that the property is non-null.

zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
voidmatcha: It seems unavoidable since `nzSortChange` has its own type. Making it optional probably won’t pass the type check either. I think it should be enough to just align the parameter order—matching it with `onTermChange` should suffice.

tbonelee: I've reordered them as you suggested.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.html
tbonelee: Could you give me a more detailed explanation of which property you mean by "optional", or provide a more specific code example?

zeppelin-web-angular/src/app/share/page-header/page-header.component.ts
voidmatcha: I missed that part, Then it’s fine to just leave it as is.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.html
voidmatcha: https://github.com/apache/zeppelin/pull/5075/files#diff-b51f1242e607c0c7567f1b1d0667919f687bd0ef86a65ce9be1d5a58768882bbR22-R24

I noticed that you added optional handling for paragraph and paragraph.config in paragraph.component.html. Since config and title in paragraph are optional types, I think this should be reflected. My earlier suggestion may have seemed a bit unclear. If this change isn’t correct, it’s totally fine to just leave it as is.

tbonelee: Ah, that's because unlike `PublishedParagraphComponent`, `NotebookParagraphComponent` has `paragraph` defined as a required non-null property. Does that explanation make sense? Or did I perhaps misunderstand something?

voidmatcha: That’s true, but even if `paragraph` is non-null, `config` or `title` are optional, so they might not exist. Typically, the template type check does not enforce the attribute types when the parent is non-null, and since the current type check passes fine, this can probably be ignored.

https://github.com/apache/zeppelin/pull/5075/files#diff-487607158ec7cd3af649d3a9a23b74fec125bda0bde28ac52c089fbf001bda19L72

Following up on the [comment link](https://github.com/apache/zeppelin/pull/5075/files#diff-b51f1242e607c0c7567f1b1d0667919f687bd0ef86a65ce9be1d5a58768882bbR22-R24) I left earlier, I noticed that `id` is being treated as non-null here, which also feels a bit strange. Looking at `published/paragraph/paragraph.component.ts`, it receives `OP.NOTE`. After receiving it, it makes sense logically to find the paragraph whose ID matches the current URL before performing any actions. Therefore, if `zeppelin-notebook-paragraph` is also handled with `*ngIf` so that it only operates when the paragraph exists, then treating it as non-null becomes meaningful.

Screenshot

In the case of `PublishedParagraphComponent`, this refers to the screen displayed when accessing a link generated via “Link this paragraph.” Previously, I observed that if I access the link without the paragraph being executed, the results do not appear, leaving a blank screen until the paragraph is actually run, which seems awkward. I previously raised an issue about this ([ZEPPELIN-6324](https://issues.apache.org/jira/browse/ZEPPELIN-6324)). I’m currently working on modifying this part to make it execute paragraph automatically when a user accesses a shared link and the paragraph has not been executed.

If a user accesses shared link that paragraph does not exist, it seems more correct to redirect to the home screen and display an error modal indicating that the paragraph does not exist. If you agree, I'll create a new issue for this behavior.

tbonelee: As far as I know, `title` is optional and is only  used by `zeppelin-elastic-input`, while `config` is not optional.
Could it be that we're looking at different parts of the code?
(I checked `ParagraphItem` in `message-paragraph.interface.ts` and `notebook/paragraph/paragraph.component.ts`)

tbonelee: For the `PublishsedParagraphComponent`(`published/paragraph/paragraph.html`), it makes sense to render `zeppelin-notebook-paragraph-result` only when `paragraph` is defined. I've updated it so that it's rendered conditionally.

tbonelee: Additionally, regarding the automatic run for "Link this paragraph", how about showing a confirmation message so that users can decide whether to run the selected paragraph if it hasn'been executed yet?
In some notebooks, running paragraphs might cause side effects on external databases, so I think it would be better to ask the user before executing. What do you think?

tbonelee: I agree that it makes sense to handle cases where a paragraph does not exist on the published page.

voidmatcha: > As far as I know, `title` is optional and is only used by `zeppelin-elastic-input`, while `config` is not optional. Could it be that we're looking at different parts of the code? (I checked `ParagraphItem` in `message-paragraph.interface.ts` and `notebook/paragraph/paragraph.component.ts`)

I was actually referring to the internal properties of `paragraph.config`, more precisely to `ParagraphConfig`.  
As for `paragraph.title`, as you mentioned, it’s treated as optional in `zeppelin-elastic-input`, so it should be fine to leave it as is without making any changes.

voidmatcha: > For the `PublishsedParagraphComponent`(`published/paragraph/paragraph.html`), it makes sense to render `zeppelin-notebook-paragraph-result` only when `paragraph` is defined. I've updated it so that it's rendered conditionally.

I’ve reviewed it and can confirm. Thank you.

> Additionally, regarding the automatic run for "Link this paragraph", how about showing a confirmation message so that users can decide whether to run the selected paragraph if it hasn'been executed yet? In some notebooks, running paragraphs might cause side effects on external databases, so I think it would be better to ask the user before executing. What do you think?

That’s a great idea. I’ll add a comment about this to the issue. Thanks for the valuable suggestion.
Also, I plan to have this process run only when the result doesn’t exist.

> I agree that it makes sense to handle cases where a paragraph does not exist on the published page.

Okay I'll add this issue.

---

I just thought of something— it might be a good first step to try converting only the page that users land on via “Link this paragraph” to a micro-frontend with React. This page is originally an isolated part of the app, and later it would be easy to extend it from the paragraph to the notebook page.

tbonelee: > I was actually referring to the internal properties of paragraph.config, more precisely to ParagraphConfig.
As for paragraph.title, as you mentioned, it’s treated as optional in zeppelin-elastic-input, so it should be fine to leave it as is without making any changes.

Got it. For those optional properties in `paragraph.config`, I made the parameters optional and set default values in `zeppelin-notebook-paragraph-control` (`control.component.ts`). If they are not optional in `control.component.ts` and given values are optional, then the compiler will emit errors.

tbonelee: > I just thought of something— it might be a good first step to try converting only the page that users land on via “Link this paragraph” to a micro-frontend with React. This page is originally an isolated part of the app, and later it would be easy to extend it from the paragraph to the notebook page.

I personally support on migrating from Angular to React, as it could help more frontend developers contribute to our project.
That said, since this is not a small decision, perhaps we should discuss it more broadly with the community. What would be the best way to move this idea forward? @jongyoul

dididy:
LGTM 👍

dididy (approved):
LGTM 👍

tbonelee:
Thank you for your reviews. I've merged this into master and branch-0.12

PR #5078, opened 2025-09-20
[ZEPPELIN-6323] Apply dark mode to the new UI

dididy:
### What is this PR for?
This PR adds dark mode support and system theme integration for the Zeppelin New UI. There were multiple demands such as [ZEPPELIN-5062](https://issues.apache.org/jira/browse/ZEPPELIN-5602) and [ZEPPELIN-4024](https://issues.apache.org/jira/browse/ZEPPELIN-4024).

#### Example: Follow system theme + other parts

https://github.com/user-attachments/assets/2159d54e-6403-4f80-91f0-3f66a93881e1

#### Example: Change with button + notebook

https://github.com/user-attachments/assets/59bdf1bc-86a3-42e1-a3d0-1d89d955ed7d

#### Automatic System Theme Detection & Sync  
- Automatically detect OS-level dark/light mode settings  
- Real-time detection and application of system theme changes  
- Theme cycle pattern: `auto(system) → opposite theme → original theme → auto`  

#### Comprehensive Dark Mode UI Support  
- Applied dark mode styles across all major components  
- Added dark mode overrides for Ant Design components  
- Full Monaco Editor dark theme support  
- Consistent color scheme and visual hierarchy  

#### Enhanced User Experience  
- Eliminated FOUC (Flash of Unstyled Content) with logic handled in `index.html`  
- Easy theme switching via a toggle button  
- Persisted user preferences in local storage  
- Theme state maintained after page reloads  

With dark mode support and system theme integration, Zeppelin delivers a modern, user-friendly experience. Users can either rely on system theme settings for automatic adaptation or manually select their preferred theme.  

### What type of PR is it?
Improvement

### Todos
* [ ] I created a dark mode [background image](https://github.com/dididy/zeppelin/blob/5d078a40c17e0561202e809da0761016516b86f2/zeppelin-web-angular/src/assets/images/bg-dark.png) used for login and loading with ChatGPT, but I need to verify whether there are any copyright issues.
* [ ] Due to the limited environment setup, I wasn’t able to check all the cases where graphs are rendered. I think we can leave this as a follow-up issue to work on later.

### What is the Jira issue?
* [[ZEPPELIN-6323](https://issues.apache.org/jira/browse/ZEPPELIN-6323)]

### Questions:


tbonelee:
Could you check failing E2E tests first?
I think we might need to extract the host based on the `CI` environment variable (or handle it another way) to fix this.
image

zeppelin-web-angular/src/app/share/theme-toggle/theme-toggle.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.themeService.setTheme(theme);
+  }
+
+  getThemeIcon(): string {
+    if (this.currentTheme === 'system') {
+      return '🤖';
+    }
+    return this.currentTheme === 'dark' ? '🌙' : '☀️';
+  }
+}
```
tbonelee: Could we switch from Unicode emoji to SVG icons? Emoji rendering depends on OS/browser font support and can vary or fail, while SVGs provide consistent visuals across environments.

zeppelin-web-angular/src/app/services/theme.service.ts
```diff
@@ -0,0 +1,197 @@
...
+      if (this.systemStartedWith === 'light') {
+        this.setTheme('system');
+        this.systemStartedWith = null;
+      } else if (this.systemStartedWith === 'dark') {
+        this.setTheme('dark');
+      } else {
+        this.setTheme('system');
+      }
+    }
+  }
```
tbonelee, line 128: I think this makes thes state changes feel unexpected when the user's system is set to dark mode.

(Parenthesis indicate the repeating cycle.)

If starting in light:
```
light -> system -> (light -> dark -> system) -> (light -> dark -> system) -> ...
```

If starting in dark:
```
dark -> light -> system -> (light -> dark -> system) -> (light -> dark -> system) -> ...
```

zeppelin-web-angular/src/app/services/theme.service.ts
tbonelee, line 1: I noticed the theme service manually manages MediaQuery listeners with `initSystemThemeDetection()` and `removeSystemThemeListener()`. This works, but it's quite verbose and the cleanup logic is easy to miss.

The core issue is that `effectiveTheme` depends on two inputs (user preference `theme$` and OS preference (MediaQuery) ), but this relationship is currently maintained imperatively with manual `.next()` calls scattered across `setTheme()` and the listener callback.

What if we made this relationship explicit using RxJS? Something like:

I noticed the theme service manually manages MediaQuery listeners with imperative `.next()` calls in multiple places. Since `effectiveTheme` depends on both user preference (`theme$`) and OS preference (MediaQuery), what if we made this relationship more explicit using RxJS?

```typescript
const systemTheme$ = fromEvent<MediaQueryListEvent>(mediaQuery, 'change').pipe(
  map(e => e.matches ? 'dark' : 'light'),
  startWith(mediaQuery.matches ? 'dark' : 'light')
);

combineLatest([theme$, systemTheme$]).pipe(
  map(([theme, sys]) => theme === 'system' ? sys : theme)
).subscribe(effectiveTheme => {
  effectiveThemeSubject.next(effectiveTheme);
  applyTheme(effectiveTheme);
});
```

This way, the data flow becomes:
```
[User Preference]              [OS Preference]
     theme$                     systemTheme$
        ↓                            ↓
        └────→ combineLatest ←───────┘
                    ↓
              effectiveTheme
                    ↓
               applyTheme()
                    ↓
            [DOM + Monaco Updates]
```

Whenever either source changes, `effectiveTheme` automatically recalculates. This would simplify the manual listener management, make cleanup automatic through RxJS subscriptions, and make the state dependencies more explicit. What do you think?

Example implementation : https://github.com/tbonelee/zeppelin/commit/2fbe5aa62bc826417b673c355e95f568b8f7f4a5

tbonelee:
In dark mode, `antd-dark.less` appears to override markdown.less or github-markdown.css, causing Markdown spacing/typography to differ from light mode. Could you also check this issue?

https://github.com/user-attachments/assets/179b4177-eefa-4c06-90c4-0313ac1445ed

zeppelin-web-angular/src/app/share/theme-toggle/theme-toggle.component.ts
voidmatcha: 09b40bd
I modified it using Material Icons, and since it’s under the Apache 2.0 license, it should be fine to use.
🤖 to  [Smart Toy](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:smart_toy:FILL@0;wght@700;GRAD@0;opsz@24&icon.query=toy&icon.size=24&icon.color=%231f1f1f)
🌙 to [Dark Mode](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:dark_mode:FILL@0;wght@700;GRAD@0;opsz@24&icon.query=light_&icon.size=24&icon.color=%231f1f1f)
☀️ to [Light Mode](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:light_mode:FILL@0;wght@700;GRAD@0;opsz@24&icon.query=light+mode&icon.size=24&icon.color=%231f1f1f)
https://developers.google.com/fonts/docs/material_icons?hl=en#licensing

zeppelin-web-angular/src/app/services/theme.service.ts
voidmatcha: b8715d2

Your suggestion does look more concise and easier to maintain, just like you said. I cherry-picked your commit and applied it.

zeppelin-web-angular/src/app/services/theme.service.ts
voidmatcha: 4818168
I wanted to control the flow based on the system theme, but it got complicated in many ways, so I decided to ignore the system theme and just fix the direction. The sequence is now light -> dark -> system.

zeppelin-web-angular/src/app/share/theme-toggle/theme-toggle.component.ts
tbonelee: Could you internalize those assets? We may need to account for environments like intranet. Also, we might need to add a line to the LICENSE file (to be confirmed). If the font files need to be excluded from the RAT check, please refer to `/pom.xml`.

voidmatcha: b40eb61
I internalized it — I hadn’t really realized that part, so thanks for pointing it out!

zeppelin-web-angular/src/app/share/share.module.ts
```diff
@@ -37,6 +37,7 @@ import { NzToolTipModule } from 'ng-zorro-antd/tooltip';
 import { NzTreeModule } from 'ng-zorro-antd/tree';
 import { NzUploadModule } from 'ng-zorro-antd/upload';
 
+import { ThemeToggleComponent } from '@zeppelin/share/theme-toggle/theme-toggle.component';
```
tbonelee: Could you change this to relative path import?

voidmatcha: 3d115a4
I applied it.

5d078a4
I found that the markdown.less styles weren’t being applied properly in light mode, so I fixed that part.


https://github.com/user-attachments/assets/70bfadf6-cdcc-41c8-84fc-cd00abad60d1

tbonelee:

PR #5080, opened 2025-09-21
[ZEPPELIN-6332] Unify keyboard event handling between Angular and Monaco editor

tbonelee:
### What is this PR for?

Keyboard event handling logic is currently split between in `paragraph.component.ts`(Angular) and `code-editor.component.ts` (Monaco editor) in new UI.

Since both Angular and the Monaco editor capture key events, the auctual handler depends on where the focus is.

Refactoring this logic would improve cohesion and type safety, making the code more readable and less prone to bugs.

##### Extracted keybinding logic into a separate module
- Keybinding logic involves low-level event handling details, but it is not a core responsibility of the components.
- To improve separation of concerns and reusability, I extracted the logic into a dedicated module: `@zeppelin/key-binding`.

##### Added a keybinding coverter
- Introduced a converter from Angular keybindings to Monaco editor keybindings.
- This allows keybindings defined once  in `shortcuts-map.ts` to be reused for both Angular and Monaco editor handlers.

##### Unified keybinding handling in a single class
- Initialization of keybinding handlers for both Angular and Monaco editor is now managed by a new `KeyBinder` class.
- A `KeyBinder` instance is created in each `ParagraphComponent`.
- Angular handlers are registerd during `ParagraphComponent` initialization, while Monaco handlers are registered when the editor is initialized (triggered by the `initKeyBindings` emitter in `NotebookParagraphCodeEditorComponent`).
- All handlers simpley emit the appropriate action key to an RxJS `Subject` within `KeyBinder`.

##### Categorized handler for each action between Angular and Monaco editor at type level.
- All actions are mapped to corresponding handler method name via `ParagraphActionToHandlerName`. Each actions must be explicitly specified as a key at the type level, and each value must be one of the method names from `NotebookParagraphKeyboardEventHandler`.
- The `MonacoKeyboardEventHandler` type defines method names for some actions, while the rest are defined in `AngularKeyboardEventHandler`. Both are subtypes of `NotebookParagraphKeyboardEventHandler` and the methods in each type are mutually exclusive.
- `NotebookParagraphCodeEditorComponent` implements the Monaco handler, and `ParagraphComponent` implements the Angular handler.
- Both handler implementation receives an `action` parameter via` handleKeyEvent(action, event)`, then look up the corresponding method in `ParagraphActionToHandlerName` using the `action` value.

### What type of PR is it?
Refactoring

### What is the Jira issue?
[ZEPPELIN-6332](https://issues.apache.org/jira/browse/ZEPPELIN-6332)

Since E2E tests are not added, we could check key shortcuts manually while running Zeppelin.

### Questions:


zeppelin-web-angular/src/app/helium-manager/public-api.ts
tbonelee, line 1: Removing this file that was mistakenly left behind in the previous PR.

zeppelin-web-angular/src/app/key-binding/key-binder.ts
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+import { ElementRef } from '@angular/core';
+import { editor as MonacoEditor } from 'monaco-editor';
+import { from, Subject } from 'rxjs';
+import { map, mergeMap, takeUntil } from 'rxjs/operators';
+
+import { ShortcutService } from '@zeppelin/services';
+import * as _ from 'lodash';
```
voidmatcha: I noticed that in other places lodash is being used with named imports. Would it make sense to align with that convention? That said, keeping it as-is also has the advantage of making it clear that the method comes from lodash, so I think either way could work. What do you think?

zeppelin-web-angular/src/app/key-binding/key-code-converter.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+import { isNil } from 'lodash';
+import { KeyCode, KeyMod } from 'monaco-editor';
+
+function isAscii(ch: string): boolean {
+  if (ch.length !== 1) {
+    throw new Error('Only single character is allowed');
+  }
+  return ch.charCodeAt(0) < 128;
```
voidmatcha: How about extracting 128 as a constant to clarify its meaning? While it can be inferred from the function name, I just wanted to suggest it.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -242,6 +218,56 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
     });
   }
 
+  handleKeyEvent(action: ParagraphActions, event: NullableKeyboardEvent) {
+    const handlerName = ParagraphActionToHandlerName[action];
+    const handlerFn = handlerName && handlerName in this && this[handlerName as keyof MonacoKeyboardEventHandler];
+    if (!handlerFn) {
```
voidmatcha:
```suggestion
    if (!handlerFn && typeof handleFn !== 'function') {
```
How about handling it a bit more safely, just in case?

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -112,42 +120,22 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
...
-    const lineNumber = position.lineNumber;
-    const lineContent = model.getLineContent(lineNumber);
-
-    if (!lineContent) {
-      return;
+  handleMoveCursorDown() {
+    if (this.editor) {
+      this.editor.trigger('keyboard', 'cursorDown', {});
     }
+  }
```
voidmatcha, line 128:
```suggestion
  handleMoveCursorUp() {
    if (this.editor) {
      this.editor.trigger('keyboard', 'cursorUp', null);
    }
  }

    const lineNumber = position.lineNumber;
    const lineContent = model.getLineContent(lineNumber);

    if (!lineContent) {
      return;
  handleMoveCursorDown() {
    if (this.editor) {
      this.editor.trigger('keyboard', 'cursorDown', null);
    }
  }
```
I know the third parameter payload of this.editor.trigger is typed as any, so it technically doesn’t matter, but I’d like to suggest that passing null might be a bit cleaner.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -242,6 +218,56 @@ export class NotebookParagraphCodeEditorComponent implements OnChanges, OnDestro
     });
   }
 
+  handleKeyEvent(action: ParagraphActions, event: NullableKeyboardEvent) {
```
voidmatcha: `event: NullableKeyboardEvent` 
It seems to be defined but unused, so it could probably be removed.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
```diff
@@ -490,146 +473,46 @@ export class NotebookParagraphComponent extends ParagraphBase implements OnInit,
...
-                return {
-                  event,
-                  action: action as ParagraphActions
-                };
-              })
-            )
-        );
-      });
-    });
+  private handleKeyEvent(action: ParagraphActions, event: NullableKeyboardEvent) {
```
voidmatcha, line 548: I’m just curious. In `shortcut.service.ts`, the `bindShortcut` method uses `KeyboardEvent` as the type for its handler parameter. But here, `NullableKeyboardEvent` is being used. It seems like the type was adjusted starting from `keyBinderService` to match, but I’m wondering if the `null` handling is actually necessary.

dididy (commented):
Some comments seem to be defined below the lines they refer to. Is there a particular reason for this? If not, it might be better to keep the placement consistent.

Separating the logic into the key-binding folder, as you did, is a very appropriate approach. It makes the code cleaner and easier to maintain. Thank you for doing such a great job. 👍👍

zeppelin-web-angular/src/app/key-binding/key-binder.ts
tbonelee: Good point. I switched lodash to named imports for consistency. I’m keeping castArray/isNil, and I use chain when we benefit from lazy evaluation to avoid intermediate arrays. Previously, lodash’s map conflicted with RxJS’s map, which blocked named imports. Since stand-alone _.map() is equivalent to native array .map() for our use, I replaced it with the native method. That removes the collision, lets us use named imports.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
tbonelee: Thanks! I added the function type guard you suggested. I also replaced the unnecessarily narrow type assertion with a minimal one that simply conveys what the compiler can’t infer, keeping behavior the same while reducing unsound narrowing.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
tbonelee: This is for the Monaco case. `editor.addCommand()` handlers don’t receive an event argument, so the handler could not pas `KeyboardEvent` to `events$` property. That’s why I introduced `NullableKeyboardEvent`. You can see this in `src/app/key-binding/key-binder.ts:initKeyBindingsOnMonaco`.

tbonelee:
Thanks for the review!

I've habitually put comments inside the condition blocks because it keeps things consistent with a following `else if` block.
Instead of,
```typescript
// Comment for if block
if (...) {
  ...
  // Comment for else-if block
  // (formatter aligns indentation differently from the if-block comment,
  // and since it's placed insde the if block it visually blends with the if block's content)
} else if (...) {
  ...
}
```

I've been doing:
```typescript
if (...) {
  // Comment for if block
  ...
} else if (...) {
  // Comment for else-if block
  ...
}
```

But while writing the comments and testing in my IDE, I realized the pattern below also works well and it plays nicely with Prettier 😂
```typescript
// Comment for if block
if (...) {
  ...
}
// Comment for else-if block
else if (...) {
  ...
}
```

Anyway, there are no standalone `else` branches here, so I'm fine placing the comments outside the blocks.

Just wondering, how do you usually write comments in this kind of case?

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/paragraph.component.ts
voidmatcha: Oh, I see. That’s unavoidable then. I think it’s fine to leave it as is. Thanks for the explanation. I’ve reviewed all the changes mentioned above comments.

voidmatcha:
I haven’t thought too deeply about this before, but the approach you mentioned seems reasonable. I looked through Zeppelin to see if there were similar cases of using comments like this, but I couldn’t find any. Usually, I place comments immediately above the relevant code, which I think works fine because it maintains consistency for anyone reading the code. However, varying the position of comments based on their purpose could help emphasize different points. (Actually, the part I intended to modify was the TODO comment about `as const` in `/notebook-paragraph-keyboard-event-handler.ts`.)

Could you add an e2e test for the keyboard shortcuts?

tbonelee:
I've been trying to add E2E tests for key bindings, but there's an issue that scenarios involving 'creating test note' harder.
If browser 'A' opens `Create New Note` modal to create new test note and browser 'B' actually creates the note, a `NOTES_INFO` message arrives, `this.cdr.markForCheck()` is called, and the modal on browser 'A' is closed. As a result, 'A' fails to click the 'Create' button.

I'll create an issue for this and plan to add the keybinding E2E tests after it's fixed. Tests that rely on creating a fresh note will likely hit the same problem, so we might have to address this first.

tbonelee:

PR #5081, opened 2025-09-22
[ZEPPELIN-6336] Fix failing zengine tests caused by invalid interpreters.json

tbonelee:
### What is this PR for?
Tests in the `zeppelin-zengine` module were failing. The main reason was that the `interpreter.json` in test resources were invalid: the value object did not have an `id` field matching its corresponding key.
As a result, the deserialized `InterpreterSetting` instance had an auto-genarated `id` field, causing the key and value.id to be inconsistent.
Before #5063, invalid settings were simply skipped. However, after that change, they caused a NPE.

This PR fixes the invalid JSON files and adds minor validation logic in the deserialization method so that such issues can be detected early by users.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6335

* Check `core-modules` - `zeppelin-zengine` tests in CI

### Questions:

### DIFF
--- a/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterInfoSaving.java
+++ b/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/InterpreterInfoSaving.java
@@ -93,7 +93,20 @@ public class InterpreterInfoSaving implements JsonSerializable {
   }
 
   public static InterpreterInfoSaving fromJson(String json) {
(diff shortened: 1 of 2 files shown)


dididy (approved):
LGTM 👍

pan3793 (approved):

tbonelee:
Thanks for the reviews, merged into master and branch-0.12.

PR #5084, opened 2025-09-25
[ZEPPELIN-6338] Allow separate frontend modules to use different Node/NPM versions

tbonelee:
### What is this PR for?
Moved the `node.version` and `npm.version` properties from the parent `pom.xml` to each module's `pom.xml`.
This allows `zeppelin-web` and `zeppelin-web-angular` to manage their Node/NPM versions independently.

Note: Each module installs and manages its own Node/NPM, so this change simply makes that independence explicit without adding extra maintenance burden or increasing package size.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6338

Check CIs if Node and npm is installed for both modules.

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,9 +107,7 @@
     <scalatest.version>3.2.19</scalatest.version>
     <scalacheck.version>1.18.1</scalacheck.version>
 
-    <!-- frontend maven plugin related versions-->
-    <node.version>v16.20.2</node.version>
-    <npm.version>8.19.4</npm.version>
+    <!-- frontend maven plugin version-->
     <plugin.frontend.version>1.12.1</plugin.frontend.version>
 
     <!-- common library versions -->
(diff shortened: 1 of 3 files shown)


tbonelee:
@dididy Thanks for checking! I think the Livy test failure last night was due to temporarily external issues (e.g., network problems). The license-check issue is being addressed in #5075.

Reamer:
I like the PR. Please also note that we maintain a node version here.

https://github.com/apache/zeppelin/blob/3e9e89565d311bfbaaafc9932b3adb5ce6104b1d/zeppelin-zengine/src/main/java/org/apache/zeppelin/helium/HeliumBundleFactory.java#L73-L76

tbonelee:
@Reamer 
Oh, I missed that part, thanks for pointing it out.
I checked, and that section is for building third-party Helium packages. Since Helium packages are based on the old UI source code, I think it might be better to keep it aligned with that side.
My thought is that using the older version for bundling would likey provide better compatibility across both the old and the new UI.

tbonelee:
Thanks, merged it into master and branch-0.12

PR #5086, opened 2025-09-28
[ZEPPELIN-6233] Upgraded the Node.js version of zeppelin-web-angular from 16 to 18

dididy:
### What is this PR for?
Upgraded the Node.js version of zeppelin-web-angular from **16.20.2** to **18.18.0**

#### [[zeppelin-web-angular/package.json](https://github.com/apache/zeppelin/compare/master...dididy:zeppelin:test/node18?expand=1#diff-c7de4df5c0afd29d2acd136879ab99f0fec5aeb696a4423885124e93bd1a986f)]
OpenSSL legacy provider: Added NODE_OPTIONS='--openssl-legacy-provider' to maintain compatibility with libraries that rely on legacy cryptographic algorithms, which are no longer enabled by default in Node 18 (OpenSSL 3).

I upgraded Husky from version 3 to version 9 and added `"prepare": "husky"` to the `package.json`. In Husky v9, the `prepare` script runs automatically when a user executes `npm install`, which ensures that Git hooks are installed without any additional manual steps. Unlike `postinstall`, which runs after package installation, `prepare` runs during the installation process, so the hooks are ready immediately after `npm install`.

Husky's Git hook management has moved from `package.json` to the `.husky/` directory. All hook scripts that were previously defined in `package.json` have been migrated to individual files under `.husky/`, following the new Husky v9 structure. This change ensures a clearer separation of configuration and hook scripts and aligns with the recommended setup for Husky v9.

#### [[zeppelin-web-angular/proxy.conf.js](https://github.com/apache/zeppelin/compare/master...dididy:zeppelin:test/node18?expand=1#diff-2a24ea2db081bdc01ecadf2293530e5997cdd179f9dc47820627776fc1b5e591)]
Updated the proxy target from `localhost` to `127.0.0.1` following the Node.js 18 upgrade.

In Node.js 18, `localhost` may resolve to IPv6 (`::1`) first, which can cause connection issues with servers bound to IPv4 (`127.0.0.1`).  
Using `127.0.0.1` ensures a consistent IPv4 connection and improves stability.

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-6233
### Questions:

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,2 @@
+cd zeppelin-web-angular
+lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,21 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
 cd zeppelin-web-angular
 lint-staged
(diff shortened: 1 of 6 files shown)


tbonelee:
How about using latest of version 18 (v18.20.8)?
https://nodejs.org/en/download/archive/v18.20.8

voidmatcha:
Thanks for suggestion. I applied it(b95231a). 

## [node 16.20.2]
스크린샷 2025-09-28 오후 3 43 33

## [node 18.20.8]
스크린샷 2025-09-28 오후 2 03 42


Also, the above npm warnings have been occurring since Node 16, so I'll ignore them for now.(I’ll leave this as an additional issue) Keeping them here for historical reference. With the version upgrade, the npm warnings related to Playwright are no longer present.

tbonelee:
Thanks, merged this into master and branch-0.12

PR #5088, opened 2025-09-28
[ZEPPELIN-6336] Enable Conditional Login Test Based on shiro.ini Presence in Zeppelin

dididy:
### What is this PR for?
Currently, Zeppelin(zeppelin-web-angular)’s E2E authentication tests require the presence of a `shiro.ini` file to run.  
However, in certain build or CI environments, this file may not exist.  
In such cases, login tests may fail or behave unpredictably.  

To improve flexibility, the test framework should support both scenarios:  
- **Auth mode (`shiro.ini` exists)** → Run all tests, including authentication/login tests  
- **Anonymous mode (`shiro.ini` does not exist)** → Skip authentication/login tests, but run all other tests  

#### 1. GitHub Actions Workflow (Matrix Mode)
- Added `strategy.matrix.mode: [anonymous, auth]`
- In `auth` mode, copy `shiro.ini.template → shiro.ini`
- In `anonymous` mode, skip `shiro.ini` setup to simulate a no-auth environment

#### 2. Playwright Global Setup / Teardown
- **`global-setup.ts`**
  - Added `LoginTestUtil.isShiroEnabled()` to detect presence of `shiro.ini`
  - If enabled → load credentials & run login tests  
  - If disabled → skip login tests, log message  
- **`global-teardown.ts`**
  - Added environment cleanup (e.g., reset cache)

#### 3. Authentication Utility (`login-page.util.ts`)
- `isShiroEnabled()`: Checks if `shiro.ini` is accessible via `fs.access`
- `getTestCredentials()`: Parses credentials only when `shiro.ini` exists
- `resetCache()`: Clears cached values between test runs

#### 4. Test Code Updates
- **`app.spec.ts`**  
  - Conditionally checks whether login page or workspace should be visible, based on `isShiroEnabled()`
- **Other Playwright tests**  
  - Authentication-related tests are skipped when `shiro.ini` is not present

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-6336

### Questions:

### DIFF
--- a/zeppelin-web-angular/pom.xml
+++ b/zeppelin-web-angular/pom.xml
@@ -262,7 +262,6 @@
       <properties>
         <web.e2e.disabled>false</web.e2e.disabled>
         <web.e2e.enabled>true</web.e2e.enabled>
-        <zeppelin.shiro.config>conf/shiro.ini</zeppelin.shiro.config>
         <test.groups>authentication</test.groups>
       </properties>
     </profile>
--- a/zeppelin-web-angular/pom.xml
+++ b/zeppelin-web-angular/pom.xml
@@ -247,22 +247,6 @@
         <web.e2e.enabled>true</web.e2e.enabled>
       </properties>
     </profile>
-    
-    <profile>
-      <id>e2e-anonymous</id>
-      <properties>
-        <web.e2e.disabled>false</web.e2e.disabled>
-        <web.e2e.enabled>true</web.e2e.enabled>
-      </properties>
-    </profile>
-    
-    <profile>
-      <id>e2e-auth</id>
-      <properties>
-        <web.e2e.disabled>false</web.e2e.disabled>
-        <web.e2e.enabled>true</web.e2e.enabled>
-      </properties>
-    </profile>
   </profiles>
 
 </project>
--- a/zeppelin-web-angular/pom.xml
+++ b/zeppelin-web-angular/pom.xml
@@ -253,7 +253,6 @@
       <properties>
         <web.e2e.disabled>false</web.e2e.disabled>
         <web.e2e.enabled>true</web.e2e.enabled>
-        <test.groups>anonymous</test.groups>
       </properties>
     </profile>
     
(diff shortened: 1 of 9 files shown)


zeppelin-web-angular/pom.xml
```diff
@@ -247,6 +247,25 @@
...
+    
+    <profile>
+      <id>e2e-auth</id>
+      <properties>
+        <web.e2e.disabled>false</web.e2e.disabled>
+        <web.e2e.enabled>true</web.e2e.enabled>
+        <zeppelin.shiro.config>conf/shiro.ini</zeppelin.shiro.config>
+        <test.groups>authentication</test.groups>
+      </properties>
+    </profile>
```
tbonelee: What do the `test.groups` and `zeppelin.shiro.config` properties do?

voidmatcha: It was a leftover from when I had considered a different approach in the past, but it seemed unnecessary, so I removed it.

tbonelee: Those two profiles look the same now. Do you have plans to differentiate them by profile in the future? If not, how about merging them into a single `e2e` profile here, and also using just `-Pe2e` in frontend.yml?

voidmatcha: 75981b6
Right, it turned out to be unnecessary after all. I removed that part, and everything has been unified under web-e2e.

tbonelee: Sorry, but just one last request 🙏 — in the frontend.yml, for the step “Run headless E2E test with Maven”, wouldn’t it be possible to merge them without an if/else as well? (That’s the only additional change I was hoping for.)

If there was another intention behind keeping them separate, please let me know.

voidmatcha: 6fa5dd8
Oh, I totally missed that part — thanks for catching it!

tbonelee:

PR #5089, opened 2025-09-30
[ZEPPELIN-6350] Fix Selenium test by Replacing Firefox with Edge

tbonelee:
### What is this PR for?
This PR stabilizes the `test-selenium...` job by switching the Selenium browser from Firefox to Microsoft Edge.
In #4941, we worked around a Linux ChromeDriver issue (https://github.com/SeleniumHQ/selenium/issues/15358) by moving from Chrome to Firefox, but that workaround also seems to be not working well.

I use Edge driver instead, since it does not have flakiness like Firefox driver and is more aligned with Chrome since it is also based on Chromium.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[[ZEPPELIN-6350]](https://issues.apache.org/jira/browse/ZEPPELIN-6350)

- Check `test-selenium...` job in `frontend`

### Questions:

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -104,28 +104,32 @@ jobs:
       run:
         shell: bash -l {0}
     env:
-      ZEPPELIN_SELENIUM_BROWSER: firefox
+      ZEPPELIN_SELENIUM_BROWSER: "edge"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Install Firefox (non-snap) and geckodriver
-        run: |
-          sudo snap remove firefox || true
-          sudo add-apt-repository -y ppa:mozillateam/ppa
-          sudo tee -a /etc/apt/preferences.d/mozilla-firefox <<EOF
-          Package: *
-          Pin: release o=LP-PPA-mozillateam
-          Pin-Priority: 1001
-          EOF
-          sudo apt-get update
-          sudo apt-get install -y firefox
-          GECKODRIVER_VERSION=$(curl -s https://api.github.com/repos/mozilla/geckodriver/releases/latest | grep tag_name | cut -d '"' -f 4)
-          wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          tar -xzf geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          chmod +x geckodriver
-          sudo mv geckodriver /usr/local/bin/
-          geckodriver --version
-          firefox --version
+      - name: Tune Runner VM
+        uses: ./.github/actions/tune-runner-vm
+      - name: Set up Edge browser
+        uses: browser-actions/setup-edge@v1
+      - name: Print Edge version
+        run: msedgedriver --version
+      - name: Set up JDK 11
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 11
+      - name: Cache local Maven repository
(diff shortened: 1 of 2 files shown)


.github/workflows/frontend.yml
```diff
@@ -104,28 +104,32 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.m2/repository
+            !~/.m2/repository/org/apache/zeppelin/
+            ~/.spark-dist
+            ~/.cache
+          key: ${{ runner.os }}-zeppelin-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-zeppelin-
```
tbonelee, line 132: I reverted this part to align it with other jobs

.github/workflows/frontend.yml
```diff
@@ -104,28 +104,32 @@ jobs:
...
-          wget -q https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          tar -xzf geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
-          chmod +x geckodriver
-          sudo mv geckodriver /usr/local/bin/
-          geckodriver --version
-          firefox --version
+      - name: Tune Runner VM
+        uses: ./.github/actions/tune-runner-vm
+      - name: Set up Edge browser
+        uses: browser-actions/setup-edge@v1
```
tbonelee, line 114: Used official browser actions to setup Edge driver.

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
```diff
@@ -89,6 +90,14 @@ private WebDriver constructWebDriver(int port) {
         return null;
       }
     };
+    Supplier<WebDriver> edgeDriverSupplier = () -> {
```
pan3793, line 93: please also update the comment at line 49

.github/workflows/frontend.yml
```diff
@@ -136,19 +140,15 @@ jobs:
...
-          geckodriver --version
-      - name: Run integration tests with Firefox
+      - name: run tests
         run: |
-          xvfb-run --auto-servernum --server-args="-screen 0 1600x1024x16" \
-            ./mvnw verify -DfailIfNoTests=false \
-            -pl zeppelin-integration \
-            -Pweb-classic -Pintegration -Pspark-scala-2.12 -Pspark-3.5 -Pweb-dist -Pusing-source-tree \
-            ${MAVEN_ARGS}
+          xvfb-run --auto-servernum --server-args="-screen 0 1600x1024x16" ./mvnw verify -DfailIfNoTests=false -pl zeppelin-integration -Pweb-classic -Pintegration -Pspark-scala-2.12 -Pspark-3.5 -Pweb-dist -Pusing-source-tree ${MAVEN_ARGS}
```
pan3793: what's wrong with the previous multi-line style?

zeppelin-integration/src/test/java/org/apache/zeppelin/WebDriverManager.java
tbonelee: I've added it, thanks!

.github/workflows/frontend.yml
tbonelee: That was due to copy-pasting when reverting to the previous version, not intentional.
I've updated it back to the multi-line style as you suggested.

Reamer (approved):
LGTM

tbonelee:
Thanks for the reviews. I've merged this into master and branch-0.12

PR #5090, opened 2025-10-01
[ZEPPELIN-6352] Prevent information disclosure in WebApplicationExceptionMapper

chadongmin:
### What is this PR for?
  This PR fixes a security vulnerability in `WebApplicationExceptionMapper` that exposes sensitive information through exception serialization. When exceptions occur, the current implementation
  serializes the entire exception object to JSON and returns it to clients, potentially leaking stack traces, internal paths, class names, and other sensitive details. This PR replaces the detailed
  exception response with a generic error message while maintaining proper server-side logging for debugging purposes.

  ### What type of PR is it?
  Improvement

  ### Todos
  * [x] - Replace exception serialization with generic error message
  * [x] - Maintain server-side error logging

  ### What is the Jira issue?
  * https://issues.apache.org/jira/browse/ZEPPELIN-6352

  * **Automated testing**: Unit tests should verify that non-WebApplicationExceptions return a generic error message instead of detailed exception information
  * **Manual testing**:
    1. Trigger an exception in the application
    2. Verify that the client receives only the generic error message: `{"status":"error","message":"Internal server error"}`
    3. Check server logs to confirm the full exception details are still logged

  N/A

  ### Questions:
  * Does the license files need to update? **No**
  * Is there breaking changes for older versions? **No** - Only changes the error response format for better security
  * Does this needs documentation? **No**

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/WebApplicationExceptionMapper.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/WebApplicationExceptionMapper.java
@@ -17,25 +17,18 @@
 
 package org.apache.zeppelin.rest.exception;
 
-import com.google.gson.GsonBuilder;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.ExceptionMapper;
 import jakarta.ws.rs.ext.Provider;
 
-import org.apache.zeppelin.rest.message.gson.ExceptionSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Provider
 public class WebApplicationExceptionMapper implements ExceptionMapper<Throwable> {
+    
   private static final Logger LOGGER = LoggerFactory.getLogger(WebApplicationExceptionMapper.class);
-
-  public WebApplicationExceptionMapper() {
-    GsonBuilder gsonBuilder = new GsonBuilder().enableComplexMapKeySerialization();
-    gsonBuilder.registerTypeHierarchyAdapter(
-            Exception.class, new ExceptionSerializer());
-  }
 
   @Override
   public Response toResponse(Throwable exception) {


zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/WebApplicationExceptionMapper.java
Reamer: With this change, GSON is no longer used in this class and must therefore be removed.

Reamer (changes requested):
I'm not sure what to think about the change yet. I understand the reason for the changes, but the information could be helpful during development.
The exception is also in the log file, which should be sufficient, even for development.

What do the other maintainers think? I'm fine with this change.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/WebApplicationExceptionMapper.java
chadongmin: @Reamer Thanks for pointing out what I missed. I removed the dependency on Gson from this class.

chadongmin:
@Reamer  For security reasons, I decided it would be better to return a generic response rather than a detailed exception when a 500 error occurs. If other maintainers disagree, I will close this PR.

tbonelee:
I also support this change. Exposing raw exception details can be risky, and in Zepplin's case backend and frontend develompment are not fully isolated, so devlopers can still check the server logs when debugging.

ParkGyeongTae:
I also agree with this change. It improves security and logs are enough for debugging.

zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/WebApplicationExceptionMapper.java
```diff
@@ -32,13 +31,10 @@
 public class WebApplicationExceptionMapper implements ExceptionMapper<Throwable> {
   private static final Logger LOGGER = LoggerFactory.getLogger(WebApplicationExceptionMapper.class);
 
-  private final Gson gson;
-
   public WebApplicationExceptionMapper() {
     GsonBuilder gsonBuilder = new GsonBuilder().enableComplexMapKeySerialization();
```
Reamer: The GsonBuilder is also no longer needed.

chadongmin: I removed GsonBuilder based on your review.

tbonelee:

PR #5092, opened 2025-10-01
[ZEPPELIN-6351] Close modals only on their own user actions

tbonelee:
### What is this PR for?
On the home page, an open modal (e.g., `Import Note`, `Create New Note`) closes when a  `NOTES_INFO` message is received.

This is unexpected and makes running parallel E2E tests that create new notes unreliable.

#### Root cause
Both modals treated any `NOTES_INFO` message as the result of *their* own submit action.
Since `NOTES_INFO` message is broadcast for various events, this led to false positives.

#### Fix
Close modals only in response to messages addressed to the submitting client: `IMPORT_NOTE` and `NEW_NOTE`. The server already sends `NEW_NOTE` (used by the old UI), but not `IMPORT_NOTE`. I added a server-sent `IMPORT_NOTE` and replaced `broadcastNote(note)` with sending `IMPORT_NOTE` only to the caller. The previously broadcast `NOTE` is mainly useful to users already on that note page; for a newly imported note, no user is on that page yet, so the broadcast is unnecessary.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
[[ZEPPELIN-6351]](https://issues.apache.org/jira/browse/ZEPPELIN-6351)

- Open two browser windows (A and B).
- In A, open a modal (e.g., Import Note or Create New Note).
- In B, perform a submit that triggers a note update. Verify the modal in A does not close.
- In A, submit the modal. Verify it closes and the note list updates.

### Questions:


zeppelin-web-angular/src/app/share/note-import/note-import.component.ts
```diff
@@ -92,7 +92,7 @@ export class NoteImportComponent extends MessageListenersManager implements OnIn
         // @ts-ignore
         result.name = this.noteImportName;
       }
-      this.messageService.importNote(result as SendNote);
+      this.messageService.importNote(result as ImportNote['note']);
```
voidmatcha, line 95:
```suggestion
  processImportJson(data: ImportNote['note']) {
    ...
    if (result.paragraphs && result.paragraphs.length > 0) {
      if (!this.noteImportName) {
        this.noteImportName = result.name;
      } else {
        result.name = this.noteImportName;
      }
      this.messageService.importNote(result);
    } else {
    ...
  }
```
How about revising it like this?

dididy (commented):
Thanks for jumping on this! Closing the modal only on the right messages makes a big difference — this was definitely needed. Appreciate you fixing it 👍

zeppelin-web-angular/src/app/share/note-import/note-import.component.ts
tbonelee: Are you suggesting we change the `data` argument type to `ImportNote['note']`?
Since `data` comes from outside the Zeppelin application(user-provided file or HTTP request) and the method already validates its shape(though somewhat superficially), I believe keeping it as `unknown` is more appropriate. What do you think?

voidmatcha: Contextually, that makes sense — the multiple conditions in the method are probably there to validate that `data` has the correct shape. I left that comment on impulse because the `ts-ignore` caught my eye, but it doesn’t really seem like something that needs to be changed.

pan3793 (approved):

PR #5093, opened 2025-10-02
[CI-Hotfix] Remove not allowed action in Selenium test

tbonelee:
I attempted to create a related JIRA issue, but the Apache JIRA site isn't working for an hour. Therefore, I’m proceeding without creating the issue.

### What is this PR for?
Currently, our CI workflow uses browser-actions/setup-edge@v1 to install Microsoft Edge. However, this action is not allowed in the apache/zeppelin repository because only actions from approved/verified sources can be used.

image

To resolve this, I removed the external action and replace it with a direct installation step.

### What type of PR is it?
Bug Fix

- Check if CI works

### Questions:

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -118,12 +118,12 @@ jobs:
           sudo apt-get install -y microsoft-edge-stable
       - name: Install msedgedriver
         run: |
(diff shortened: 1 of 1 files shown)


tbonelee:
Thanks! Merged into master and branch-0.12

PR #5096, opened 2025-10-05
[ZEPPELIN-6357] Fixed missing login prerequisite in each test

dididy:
### What is this PR for?
https://github.com/apache/zeppelin/actions/runs/18239525409/job/51939323284?pr=5095

I recently found that the `run-playwright-e2e-tests (auth)` step wasn’t running properly during the E2E tests.  
The root cause was that the tests required a setup (tear-up) process.  
Since handling login in `globalSetup` within `playwright.config` made it difficult to account for all browser contexts,  
I added a method in `utils` to perform the login when a `shiro.ini` file exists, and made each test execute it in `beforeEach`.

During this process, I also merged the `helper` into `utils`, as they essentially served the same purpose.

### What type of PR is it?
Bug Fix
Improvement

### What is the Jira issue?
ZEPPELIN-6357

### Questions:


tbonelee:

PR #5097, opened 2025-10-06
[ZEPPELIN-6356] Fix ZeppelinClientIntegrationTest

tbonelee:
### What is this PR for?

After #5090, we started hiding unhandled error details from clients, which caused some tests that depended on those details to fail. The afffected cases are note imports/creations that should fail when the target note path already exists.

This PR introduces a `ConflictException` and updates `NotebookRestApi` to return HTTP 409 (Conflict) when `NotebookService` throws or returns `NotePathAlreadyExistsException`.

I chose to do this exception-to-HTTP mapping in `NotebookRestApi` because `NotebookService` is also used outside REST contexts, and assigning HTTP status codes belongs in the REST layer.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6356

- Check if `zeppelin-integration-test` and other tests pass

image

### Questions:


tbonelee:
@Reamer When you have a moment, could you please take a look at this?

zeppelin-server/src/main/java/org/apache/zeppelin/rest/NotebookRestApi.java
```diff
@@ -512,7 +522,15 @@ public Response createNote(String message) throws IOException {
             defaultInterpreterGroup,
             request.getAddingEmptyParagraph(),
             getServiceContext(),
-            new RestServiceCallback<>());
+            new RestServiceCallback<>() {
```
Reamer: I don't like overriding the onFailure method here just to handle the exception differently.

tbonelee:
Understood. So you prefer letting exceptions be thrown at the source and then handling them centrally in `WebApplicationExceptionMapper` wherever possible? I'm generally used to preventing service-layer exceptions from bubbling past the controller (except in unexpected cases), which is why I initially did it that way. But I don't have a strong preference.
I've updated the code as you suggested. Thanks for the thoughtful feedback!

zeppelin-server/src/main/java/org/apache/zeppelin/rest/exception/ConflictException.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.zeppelin.rest.exception;
+
+import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.core.Response;
+import org.apache.zeppelin.utils.ExceptionUtils;
+
+/**
+ * ConflictException handler for WebApplicationException
+ */
+public class ConflictException extends WebApplicationException {
```
Reamer: I think this class can now be removed.

Reamer:
> be thrown at the source and then handling them centrally in WebApplicationExceptionMapper wherever possible?

Yes, I hope this simplifies the code and leads to a uniform API response.

tbonelee:
@Reamer I removed that class. I also realized I merged the previous PR, which caused the test failure, into `branch-0.12`. Since it is a breaking change, should I revert that commit to keep the 0.12 line stable? If you agree, I will open a revert PR targeting branch-0.12. Sorry for the oversight.

tbonelee:
I noticed the tests aren't fixed yet. I'll correct them and re-request review. Sorry for the confusion.

Reamer:
> @Reamer I removed that class. I also realized I merged the previous PR, which caused the test failure, into `branch-0.12`. Since it is a breaking change, should I revert that commit to keep the 0.12 line stable? If you agree, I will open a revert PR targeting branch-0.12. Sorry for the oversight.

No, that's not necessary. We'll merge this change as well and accept the breaking change.

tbonelee:
Thansk, I merged this into master and also the branch-0.12 as you suggested

PR #5147, opened 2026-02-02
[ZEPPELIN-6395] Fix NumberFormatException for numeric interpreter properties in New UI

kevinjmh:
#### Problem:

The New UI (Angular) saves numeric interpreter properties as JavaScript number types, which get deserialized to Java Double by Gson. When these properties are converted to strings (e.g., 60000.0), parsing them as Long or Integer fails with NumberFormatException.

#### Solution:

Convert all property values to strings in the New UI before sending to the backend, matching the behavior of the Old UI which uses strings for all property values.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6395

### Questions:


kevinjmh:
See #5265

PR #666, opened 2016-01-22
Shell interpreter improvements

Karuppayya:
Creating new PR with the changes from 
https://github.com/apache/incubator-zeppelin/pull/615
Please check the above PR for prior discussions.

### What is this PR for?
*Provide ability to to run  shell commands in parallel
*Provide ability to cancel shell command
*Propagate the error from shell commands to UI

### What type of PR is it?
Improvement

### Todos
NA

### Is there a relevant Jira issue?
No

### How should this be tested?
*To check parallelism, run more than 10 shell commands concurrently.
*To verify whether error is propagate to UI, execute a shell command which will error out(simplest being cd of a non existent directory )
*To verify the cancel functionality, try cancelling a shell command that is running.

### Screenshots (if appropriate)
NA

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #5218, opened 2026-04-19
[ZEPPELIN-6411]  Semantic search for Zeppelin

kkalyan:
### What is this PR for?

 Added `EmbeddingSearch` — a new `SearchService` implementation that enables natural language search across Zeppelin notebooks using ONNX-based  sentence embeddings (all-MiniLM-L6-v2). 
Disabled by default, enabled with `zeppelin.search.semantic.enable = true`.

 **The problem**: 
Zeppelin's built-in search uses Lucene's keyword matching, which works well for exact terms but falls short for the way analysts actually search.
   A user looking for "yesterday's spending" gets zero results — even though their notebooks contain SELECT sum(cost) WHERE date = current_date -
   interval '1' day. The words don't match, so Lucene can't find it.

   This PR adds EmbeddingSearch, an alternative SearchService that uses sentence embeddings (all-MiniLM-L6-v2 via ONNX Runtime) to match by meaning
   instead of keywords. It runs entirely in-process with no external services required.

   Beyond semantic matching, EmbeddingSearch addresses other gaps in notebook search:

   - Indexes paragraph output — table results and text output become searchable, not just the code
   - Extracts SQL table names — FROM/JOIN references are extracted and used to boost related paragraphs in a two-phase ranking
   - Strips interpreter prefixes — %spark.sql, %python etc. are removed so they don't pollute search results
   - Live indexing — new or updated paragraphs are searchable immediately, no restart needed
 ### What type of PR is it?
 Feature

 ### Todos
 - [x] EmbeddingSearch core implementation (ONNX inference, mean pooling, cosine similarity)
 - [x] Table name extraction from SQL (FROM/JOIN regex) with two-phase search boosting
 - [x] Paragraph output indexing (TABLE, TEXT results)
 - [x] Versioned binary persistence (v3 format)
 - [x] Live indexing (new paragraphs searchable immediately)
 - [x] Angular UI: render search results with separate code/output/tables blocks
 - [x] Classic UI: same improvements
 - [x] 11 unit tests including semantic validation
 - [x] Documentation

 ### What is the Jira issue?
 - https://issues.apache.org/jira/browse/ZEPPELIN-6411

 **Automated tests:**
 ```bash
 # Embedding search tests (requires ~86MB model download, one-time)
 ZEPPELIN_EMBEDDING_TEST=true mvn test -pl zeppelin-zengine -Dtest=EmbeddingSearchTest

 # Verify no regressions to existing Lucene search
 mvn test -pl zeppelin-zengine -Dtest=LuceneSearchTest

 Manual testing:

 1. Set zeppelin.search.semantic.enable = true in zeppelin-site.xml
 2. Restart Zeppelin
 3. Search for natural language queries like:
   - "yesterday's spending" (Lucene: 0 results → Semantic: finds spend queries)
   - "how much do drivers earn" (finds taxi tip analysis)
   - "late deliveries" (finds shipping performance queries)
   - "airport rides" (both work — keyword match exists)
  ```

 Semantic Search with New UI
image
 Semantic Search with Classic UI
image

### Questions:

 - Does the license files need to update? 
 - Yes — NOTICE updated with ONNX Runtime (MIT) and DJL Tokenizers (Apache 2.0) attribution.
 - Is there breaking changes for older versions? 
 - No. Disabled by default. Existing LuceneSearch behavior is unchanged.

### DIFF
--- a/bin/install-search-model.sh
+++ b/bin/install-search-model.sh
@@ -26,24 +26,53 @@ MODEL_NAME="all-MiniLM-L6-v2"
 MODEL_REVISION="c9745ed1d9f207416be6d2e6f8de32d1f16199bf"
 BASE_URL="https://huggingface.co/sentence-transformers/${MODEL_NAME}/resolve/${MODEL_REVISION}"
 
+# Expected SHA256 checksums for integrity verification
+MODEL_SHA256="6fd5d72fe4589f189f8ebc006442dbb529bb7ce38f8082112682524616046452"
+TOKENIZER_SHA256="be50c3628f2bf5bb5e3a7f17b1f74611b2561a3a27eeab05e5aa30f411572037"
+
 INDEX_PATH="${1:-/tmp/zeppelin-index}"
 MODEL_DIR="${INDEX_PATH}/models/${MODEL_NAME}"
 
 mkdir -p "${MODEL_DIR}"
 
+verify_sha256() {
+  local file="$1" expected="$2"
+  local actual
+  if command -v sha256sum >/dev/null 2>&1; then
+    actual=$(sha256sum "${file}" | cut -d' ' -f1)
+  elif command -v shasum >/dev/null 2>&1; then
+    actual=$(shasum -a 256 "${file}" | cut -d' ' -f1)
+  else
+    echo "WARNING: Neither sha256sum nor shasum found, skipping integrity check for ${file}"
+    return 0
+  fi
+  if [ "${actual}" != "${expected}" ]; then
+    echo "ERROR: SHA256 mismatch for ${file}"
+    echo "  Expected: ${expected}"
+    echo "  Actual:   ${actual}"
+    rm -f "${file}"
+    return 1
+  fi
+  echo "SHA256 verified: ${file}"
+}
+
 download() {
-  local url="$1" dest="$2"
+  local url="$1" dest="$2" expected_sha="$3"
   if [ -f "${dest}" ]; then
-    echo "Already exists: ${dest}"
-    return
+    if verify_sha256 "${dest}" "${expected_sha}"; then
+      echo "Already exists and verified: ${dest}"
+      return
+    fi
+    echo "Existing file failed verification, re-downloading..."
   fi
   echo "Downloading ${url} ..."
   curl -fSL --connect-timeout 30 --max-time 300 -o "${dest}.tmp" "${url}"
   mv "${dest}.tmp" "${dest}"
+  verify_sha256 "${dest}" "${expected_sha}"
   echo "Saved: ${dest}"
 }
 
-download "${BASE_URL}/onnx/model.onnx" "${MODEL_DIR}/model.onnx"
-download "${BASE_URL}/tokenizer.json" "${MODEL_DIR}/tokenizer.json"
+download "${BASE_URL}/onnx/model.onnx" "${MODEL_DIR}/model.onnx" "${MODEL_SHA256}"
+download "${BASE_URL}/tokenizer.json" "${MODEL_DIR}/tokenizer.json" "${TOKENIZER_SHA256}"
 
 echo "Model installed to ${MODEL_DIR}"
--- a/bin/install-search-model.sh
+++ b/bin/install-search-model.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Downloads the sentence-transformer model required for semantic search.
+# Run this once before starting Zeppelin with zeppelin.search.semantic.enable=true.
+#
+# Usage: bin/install-search-model.sh [INDEX_PATH]
+#   INDEX_PATH defaults to /tmp/zeppelin-index (matches zeppelin.search.index.path)
+
+set -euo pipefail
+
+MODEL_NAME="all-MiniLM-L6-v2"
+MODEL_REVISION="c9745ed1d9f207416be6d2e6f8de32d1f16199bf"
+BASE_URL="https://huggingface.co/sentence-transformers/${MODEL_NAME}/resolve/${MODEL_REVISION}"
+
+INDEX_PATH="${1:-/tmp/zeppelin-index}"
+MODEL_DIR="${INDEX_PATH}/models/${MODEL_NAME}"
+
+mkdir -p "${MODEL_DIR}"
+
+download() {
+  local url="$1" dest="$2"
+  if [ -f "${dest}" ]; then
+    echo "Already exists: ${dest}"
+    return
+  fi
+  echo "Downloading ${url} ..."
+  curl -fSL --connect-timeout 30 --max-time 300 -o "${dest}.tmp" "${url}"
+  mv "${dest}.tmp" "${dest}"
+  echo "Saved: ${dest}"
+}
+
+download "${BASE_URL}/onnx/model.onnx" "${MODEL_DIR}/model.onnx"
+download "${BASE_URL}/tokenizer.json" "${MODEL_DIR}/tokenizer.json"
+
+echo "Model installed to ${MODEL_DIR}"
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -8,7 +8,7 @@ replacement for `LuceneSearch` that understands meaning, not just keywords.
 
 **Example**: Searching "yesterday's spending" finds paragraphs containing
 `SELECT sum(cost) FROM analytics.daily_sales WHERE date = current_date - interval '1' day`
-— something keyword search cannot do.
+— something keyword search cannot do (returns 0 results with LuceneSearch).
 
 ## Motivation
 
@@ -41,27 +41,29 @@ finding the right query becomes a significant productivity bottleneck.
 │  │ (DJL)        │  │ (CPU)        │  │ ConcurrentHashMap│  │
 │  └──────────────┘  └──────────────┘  └────────┬─────────┘  │
 │                                                │            │
-│  Query: embed → brute-force cosine sim → top-20│            │
-│  Index: embed paragraph text+title+output      │            │
+│  Two-phase query:                              │            │
+│  1. Embed query → cosine sim → find tables     │            │
+│  2. Re-rank with table boost → top-20          │            │
 │                                                ▼            │
-│                                    embedding_index.bin      │
-│                                    (persisted to disk)      │
+│  Index: text + title + output + tables   embedding_index.bin│
+│         (persisted to disk, versioned)                      │
 └─────────────────────────────────────────────────────────────┘
 ```
 
 ### Model
 
 - **all-MiniLM-L6-v2**: 384-dimensional sentence embeddings
-- 22MB ONNX model (86MB fp32, quantized version available)
+- 86MB ONNX model (quantized version available at 22MB)
 - Downloaded on first use to `zeppelin.search.index.path/models/`
 - Runs on CPU via ONNX Runtime (~5ms per paragraph)
 
 ### Index
 
 - In-memory `ConcurrentHashMap<String, IndexEntry>` with `ReadWriteLock`
-- Each entry: 384 floats (1.5KB) + metadata strings
+- Each entry stores: embedding (384 floats), notebook name, paragraph text,
+  title, extracted SQL table names, and paragraph output
 - 10K paragraphs ≈ 15MB RAM, 50K paragraphs ≈ 75MB RAM
-- Persisted as single binary file (`embedding_index.bin`)
+- Persisted as versioned binary file (`embedding_index.bin`, currently v3)
 - Brute-force cosine similarity: < 50ms for 50K paragraphs
 
 ### What gets indexed (vs. LuceneSearch)
@@ -70,17 +72,31 @@ finding the right query becomes a significant productivity bottleneck.
 |---------|:---:|:---:|
 | Paragraph text | ✓ | ✓ |
 | Paragraph title | ✓ | ✓ |
-| Notebook name | ✓ | ✓ |
+| Notebook name | ✓ | ✓ (in embedding context) |
 | Paragraph output (TABLE, TEXT) | ✗ | ✓ |
+| SQL table names (FROM/JOIN) | ✗ | ✓ (extracted + boosted) |
 | Interpreter prefix stripped | ✗ | ✓ |
+
+### Two-Phase Search
+
+1. **Phase 1 — Table Discovery**: Run cosine similarity, collect SQL table names
+   from top-20 results weighted by rank
+2. **Phase 2 — Table Boost**: Re-score results, boosting paragraphs that reference
+   the discovered tables (+0.05 per matching table)
+
+This helps queries like "click funnel analysis" surface all paragraphs that query
+the same tables, even if their SQL text is very different.
 
 ## Configuration
 
 Disabled by default. Enable with a single property:
 
-```properties
-# In zeppelin-site.xml or zeppelin-env.sh
-zeppelin.search.semantic.enable = true
+```xml
+<!-- In zeppelin-site.xml -->
+<property>
+  <name>zeppelin.search.semantic.enable</name>
+  <value>true</value>
+</property>
 ```
 
 Requires `zeppelin.search.enable = true` (already the default).
@@ -96,17 +112,34 @@ Requires `zeppelin.search.enable = true` (already the default).
 ## Changes
 
 ### New files
-- `zeppelin-zengine/.../search/EmbeddingSearch.java` — Core implementation
-- `zeppelin-zengine/.../search/EmbeddingSearchTest.java` — Tests (gated behind env var)
+- `zeppelin-zengine/.../search/EmbeddingSearch.java` — Core implementation (~700 lines)
+- `zeppelin-zengine/.../search/EmbeddingSearchTest.java` — 11 tests including semantic validation
+- `docs/embedding-search.md` — This document
 
-### Modified files
+### Modified files — Backend
 - `zeppelin-zengine/pom.xml` — Add `onnxruntime` and `djl-tokenizers` dependencies
 - `zeppelin-zengine/.../conf/ZeppelinConfiguration.java` — Add `ZEPPELIN_SEARCH_SEMANTIC_ENABLE`
 - `zeppelin-server/.../server/ZeppelinServer.java` — Wire `EmbeddingSearch` based on config
+- `NOTICE` — Attribution for ONNX Runtime and DJL
+
+### Modified files — Frontend
+- `zeppelin-web-angular/.../result-item/` — Render search results with separate
+  code block, output block, and table name display (replaces Monaco editor)
+- `zeppelin-web/src/app/search/` — Same improvements for Classic UI
+- Various TypeScript build fixes (`tsconfig`, type annotations)
 
 ### Dependencies added
 - `com.microsoft.onnxruntime:onnxruntime:1.18.0` (~50MB, Apache 2.0 compatible)
-- `ai.djl.huggingface:tokenizers:0.28.0` (~2MB, Apache 2.0)
+- `ai.djl.huggingface:tokenizers:0.28.0` (~2MB, Apache 2.0, JNA excluded to
+  avoid version conflict with Zeppelin's existing JNA 4.1.0)
+
+## Search Result Display
+
+Both Angular and Classic UIs now render search results with:
+- **Code block**: SQL/Python code with syntax-appropriate styling
+- **Output block**: Paragraph execution results (table data, text output)
+- **Table names**: Extracted SQL table names highlighted with 📊 icon
+- **Language badge**: `sql`, `python`, `md`, etc.
 
 ## Design Decisions
 
@@ -114,30 +147,22 @@ Requires `zeppelin.search.enable = true` (already the default).
 
 ONNX Runtime is the standard inference engine for transformer models. It supports
 the exact same model files used by Python (HuggingFace, ChromaDB, etc.), ensuring
-embedding compatibility. DJL and other Java ML libraries either don't support
-sentence-transformers or require significantly more code.
+embedding compatibility.
 
 ### Why brute-force instead of HNSW/ANN?
 
 For Zeppelin's scale (typically < 50K paragraphs), brute-force cosine similarity
-on normalized vectors is:
-- **Fast enough**: < 50ms for 50K entries (384-dim dot product)
-- **Exact**: No approximation error
-- **Zero complexity**: No graph construction, no tuning parameters
-- **Tiny memory**: Just a flat float array
-
-HNSW would add ~3x memory overhead and code complexity for negligible latency gain.
+on normalized vectors is fast enough (< 50ms), exact (no approximation error),
+and adds zero complexity.
 
 ### Why download model on first use instead of bundling?
 
-The ONNX model is 86MB (fp32). Bundling it would bloat the Zeppelin distribution.
+The ONNX model is 86MB. Bundling it would bloat the Zeppelin distribution.
 Downloading on first use keeps the distribution lean and allows users to swap models.
 
 ### Why not use Lucene's vector search (since 9.0)?
 
 Zeppelin uses Lucene 8.7.0. Upgrading to 9.x is a separate, larger effort.
-Even with Lucene 9.x vector search, you'd still need the ONNX model for embedding
-generation — so the dependency footprint is similar.
 
 ## Testing
 
@@ -150,16 +175,17 @@ ZEPPELIN_EMBEDDING_TEST=true mvn test -pl zeppelin-zengine \
 mvn test -pl zeppelin-zengine -Dtest=LuceneSearchTest
 ```
 
-### Key test: `semanticSearchFindsRelatedConcepts`
+### Key tests
 
-This test validates the core value proposition — that a natural language query
-("yesterday's spending") correctly ranks a SQL spend query above an unrelated
-user count query, even though neither contains the word "spending" or "yesterday".
+- `semanticSearchFindsRelatedConcepts` — validates that "yesterday's spending"
+  ranks a SQL spend query above an unrelated user count query
+- `newParagraphIsLiveIndexed` — validates that newly added paragraphs are
+  immediately searchable without restart
 
 ## Future Work
 
 - [ ] Quantized model support (22MB INT8 vs 86MB FP32)
 - [ ] Hybrid search: combine embedding similarity with keyword matching
-- [ ] Frontend: show similarity scores in search results
-- [ ] Configurable model path for air-gapped environments
+- [ ] Configurable model URL for air-gapped environments
 - [ ] Batch embedding during initial index rebuild
+- [ ] Similarity score display in search results
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -7,7 +7,7 @@ search across Zeppelin notebooks using ONNX-based sentence embeddings. This is a
 replacement for `LuceneSearch` that understands meaning, not just keywords.
 
 **Example**: Searching "yesterday's spending" finds paragraphs containing
-`SELECT sum(cost) FROM click_funnel WHERE date = current_date - interval '1' day`
+`SELECT sum(cost) FROM analytics.daily_sales WHERE date = current_date - interval '1' day`
 — something keyword search cannot do.
 
 ## Motivation
@@ -17,7 +17,7 @@ Lucene's `StandardAnalyzer`. This has several limitations for notebook search:
 
 1. **No semantic understanding** — "yesterday's spend" won't find `current_date - 1`
 2. **Poor SQL tokenization** — `StandardAnalyzer` breaks on underscores and dots in
-   table names like `eq_analytics_prod.click_funnel_raw`
+   table names like `analytics_db.daily_sales`
 3. **No output indexing** — query results (table data, text output) are not searchable
 4. **Exact match only** — users must guess the exact terms used in notebooks
 
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -1,0 +1,165 @@
+# ZEPPELIN-6411: Semantic Search for Notebooks using Sentence Embeddings
+
+## Summary
+
+Add `EmbeddingSearch` — a new `SearchService` implementation that enables natural language
+search across Zeppelin notebooks using ONNX-based sentence embeddings. This is a drop-in
+replacement for `LuceneSearch` that understands meaning, not just keywords.
+
+**Example**: Searching "yesterday's spending" finds paragraphs containing
+`SELECT sum(cost) FROM click_funnel WHERE date = current_date - interval '1' day`
+— something keyword search cannot do.
+
+## Motivation
+
+Zeppelin's current search (`LuceneSearch`) uses keyword-based full-text search with
+Lucene's `StandardAnalyzer`. This has several limitations for notebook search:
+
+1. **No semantic understanding** — "yesterday's spend" won't find `current_date - 1`
+2. **Poor SQL tokenization** — `StandardAnalyzer` breaks on underscores and dots in
+   table names like `eq_analytics_prod.click_funnel_raw`
+3. **No output indexing** — query results (table data, text output) are not searchable
+4. **Exact match only** — users must guess the exact terms used in notebooks
+
+For teams with hundreds or thousands of notebooks (common in data/analytics teams),
+finding the right query becomes a significant productivity bottleneck.
+
+## Architecture
+
+```
+                    SearchService (abstract)
+                    ├── LuceneSearch        (existing, keyword-based)
+                    ├── EmbeddingSearch     (new, semantic)
+                    └── NoSearchService     (existing, no-op)
+
+┌─────────────────────────────────────────────────────────────┐
+│  EmbeddingSearch                                            │
+│                                                             │
+│  ┌──────────────┐  ┌──────────────┐  ┌──────────────────┐  │
+│  │ HuggingFace  │  │ ONNX Runtime │  │ In-Memory Index  │  │
+│  │ Tokenizer    │→ │ Inference    │→ │ float[][] + meta │  │
+│  │ (DJL)        │  │ (CPU)        │  │ ConcurrentHashMap│  │
+│  └──────────────┘  └──────────────┘  └────────┬─────────┘  │
+│                                                │            │
+│  Query: embed → brute-force cosine sim → top-20│            │
+│  Index: embed paragraph text+title+output      │            │
+│                                                ▼            │
+│                                    embedding_index.bin      │
+│                                    (persisted to disk)      │
+└─────────────────────────────────────────────────────────────┘
+```
+
+### Model
+
+- **all-MiniLM-L6-v2**: 384-dimensional sentence embeddings
+- 22MB ONNX model (86MB fp32, quantized version available)
+- Downloaded on first use to `zeppelin.search.index.path/models/`
+- Runs on CPU via ONNX Runtime (~5ms per paragraph)
+
+### Index
+
+- In-memory `ConcurrentHashMap<String, IndexEntry>` with `ReadWriteLock`
+- Each entry: 384 floats (1.5KB) + metadata strings
+- 10K paragraphs ≈ 15MB RAM, 50K paragraphs ≈ 75MB RAM
+- Persisted as single binary file (`embedding_index.bin`)
+- Brute-force cosine similarity: < 50ms for 50K paragraphs
+
+### What gets indexed (vs. LuceneSearch)
+
+| Content | LuceneSearch | EmbeddingSearch |
+|---------|:---:|:---:|
+| Paragraph text | ✓ | ✓ |
+| Paragraph title | ✓ | ✓ |
+| Notebook name | ✓ | ✓ |
+| Paragraph output (TABLE, TEXT) | ✗ | ✓ |
+| Interpreter prefix stripped | ✗ | ✓ |
+
+## Configuration
+
+Disabled by default. Enable with a single property:
+
+```properties
+# In zeppelin-site.xml or zeppelin-env.sh
+zeppelin.search.semantic.enable = true
+```
+
+Requires `zeppelin.search.enable = true` (already the default).
+
+### Configuration matrix
+
+| `search.enable` | `search.semantic.enable` | Result |
+|:---:|:---:|---|
+| true | false (default) | LuceneSearch (existing behavior) |
+| true | true | EmbeddingSearch (semantic) |
+| false | any | NoSearchService |
+
+## Changes
+
+### New files
+- `zeppelin-zengine/.../search/EmbeddingSearch.java` — Core implementation
+- `zeppelin-zengine/.../search/EmbeddingSearchTest.java` — Tests (gated behind env var)
+
+### Modified files
+- `zeppelin-zengine/pom.xml` — Add `onnxruntime` and `djl-tokenizers` dependencies
+- `zeppelin-zengine/.../conf/ZeppelinConfiguration.java` — Add `ZEPPELIN_SEARCH_SEMANTIC_ENABLE`
+- `zeppelin-server/.../server/ZeppelinServer.java` — Wire `EmbeddingSearch` based on config
+
+### Dependencies added
+- `com.microsoft.onnxruntime:onnxruntime:1.18.0` (~50MB, Apache 2.0 compatible)
+- `ai.djl.huggingface:tokenizers:0.28.0` (~2MB, Apache 2.0)
+
+## Design Decisions
+
+### Why ONNX Runtime instead of a Java ML library?
+
+ONNX Runtime is the standard inference engine for transformer models. It supports
+the exact same model files used by Python (HuggingFace, ChromaDB, etc.), ensuring
+embedding compatibility. DJL and other Java ML libraries either don't support
+sentence-transformers or require significantly more code.
+
+### Why brute-force instead of HNSW/ANN?
+
+For Zeppelin's scale (typically < 50K paragraphs), brute-force cosine similarity
+on normalized vectors is:
+- **Fast enough**: < 50ms for 50K entries (384-dim dot product)
+- **Exact**: No approximation error
+- **Zero complexity**: No graph construction, no tuning parameters
+- **Tiny memory**: Just a flat float array
+
+HNSW would add ~3x memory overhead and code complexity for negligible latency gain.
+
+### Why download model on first use instead of bundling?
+
+The ONNX model is 86MB (fp32). Bundling it would bloat the Zeppelin distribution.
+Downloading on first use keeps the distribution lean and allows users to swap models.
+
+### Why not use Lucene's vector search (since 9.0)?
+
+Zeppelin uses Lucene 8.7.0. Upgrading to 9.x is a separate, larger effort.
+Even with Lucene 9.x vector search, you'd still need the ONNX model for embedding
+generation — so the dependency footprint is similar.
+
+## Testing
+
+```bash
+# Run embedding search tests (requires model download, ~86MB first time)
+ZEPPELIN_EMBEDDING_TEST=true mvn test -pl zeppelin-zengine \
+  -Dtest=EmbeddingSearchTest
+
+# Run existing Lucene tests (should still pass, no changes)
+mvn test -pl zeppelin-zengine -Dtest=LuceneSearchTest
+```
+
+### Key test: `semanticSearchFindsRelatedConcepts`
+
+This test validates the core value proposition — that a natural language query
+("yesterday's spending") correctly ranks a SQL spend query above an unrelated
+user count query, even though neither contains the word "spending" or "yesterday".
+
+## Future Work
+
+- [ ] Quantized model support (22MB INT8 vs 86MB FP32)
+- [ ] Hybrid search: combine embedding similarity with keyword matching
+- [ ] Frontend: show similarity scores in search results
+- [ ] Configurable model path for air-gapped environments
+- [ ] Batch embedding during initial index rebuild
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -1,3 +1,17 @@
+<!--
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
 # ZEPPELIN-6411: Semantic Search for Notebooks using Sentence Embeddings
 
 ## Summary
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -147,12 +147,32 @@ Requires `zeppelin.search.enable = true` (already the default).
 - `ai.djl.huggingface:tokenizers:0.28.0` (~2MB, Apache 2.0, JNA excluded to
   avoid version conflict with Zeppelin's existing JNA 4.1.0)
 
-## Search Result Display
+## Search Result Response Contract
 
-Both Angular and Classic UIs now render search results with:
+Both `LuceneSearch` and `EmbeddingSearch` return `List<Map<String, String>>` with
+these keys:
+
+| Key | LuceneSearch | EmbeddingSearch |
+|-----|-------------|-----------------|
+| `id` | `noteId` or `noteId/paragraph/paragraphId` | Same |
+| `name` | Notebook title | Notebook title |
+| `snippet` | Highlighted paragraph text (`<B>` tags) | Paragraph text (no highlighting) |
+| `text` | Full paragraph text | Full paragraph text |
+| `header` | Highlighted paragraph title (`<B>` tags) | Paragraph title (plain) |
+| `title` | Same as `header` | Paragraph title (plain) |
+| `tables` | `""` (empty) | Space-separated SQL table names |
+| `output` | `""` (empty) | Paragraph output (truncated to 300 chars) |
+
+The `title`, `tables`, and `output` fields are dedicated structured fields. The
+`header` field preserves backward compatibility — for `LuceneSearch` it contains
+the highlighted paragraph title, for `EmbeddingSearch` it contains the plain title.
+
+### Frontend Display
+
+Both Angular and Classic UIs render search results with:
 - **Code block**: SQL/Python code with syntax-appropriate styling
-- **Output block**: Paragraph execution results (table data, text output)
-- **Table names**: Extracted SQL table names highlighted with 📊 icon
+- **Output block**: Paragraph execution results (from `output` field)
+- **Table names**: Extracted SQL table names (from `tables` field)
 - **Language badge**: `sql`, `python`, `md`, etc.
 
 ## Design Decisions
--- a/docs/embedding-search.md
+++ b/docs/embedding-search.md
@@ -140,7 +140,6 @@ Requires `zeppelin.search.enable = true` (already the default).
 - `zeppelin-web-angular/.../result-item/` — Render search results with separate
   code block, output block, and table name display (replaces Monaco editor)
 - `zeppelin-web/src/app/search/` — Same improvements for Classic UI
-- Various TypeScript build fixes (`tsconfig`, type annotations)
 
 ### Dependencies added
 - `com.microsoft.onnxruntime:onnxruntime:1.18.0` (~50MB, Apache 2.0 compatible)
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
@@ -85,7 +85,13 @@ public class EmbeddingSearch extends SearchService {
   private static final String MODEL_NAME = "all-MiniLM-L6-v2";
   private static final int EMBEDDING_DIM = 384;
   private static final int MAX_SEQ_LENGTH = 256;
+  /** Maximum number of candidates returned from {@link #query(String)}. */
   private static final int MAX_RESULTS = 20;
+  /**
+   * Cosine similarity floor for a candidate to be considered a match.
+   * Tuned empirically against all-MiniLM-L6-v2: values below this are effectively noise
+   * for short-query / long-paragraph comparisons. See embedding-search.md for details.
+   */
   private static final float MIN_SIMILARITY = 0.25f;
   private static final int MAX_TEXT_LENGTH = 1500;
 
@@ -94,9 +100,28 @@ public class EmbeddingSearch extends SearchService {
   /** Regex to extract qualified table names from SQL (e.g. schema.table). */
   private static final Pattern TABLE_RE =
       Pattern.compile("(?:FROM|JOIN)\\s+([a-zA-Z_]\\w*\\.[a-zA-Z_]\\w*)", Pattern.CASE_INSENSITIVE);
+  /**
+   * Additive score boost applied to a candidate for each relevant table it references.
+   * Chosen small enough that it only breaks ties among already-similar candidates
+   * and cannot promote semantically unrelated results past {@link #MIN_SIMILARITY}.
+   */
   private static final float TABLE_BOOST = 0.05f;
+  /**
+   * Fraction of the top table's weight used as the cutoff for "relevant" tables in Phase 1
+   * of {@link #query(String)}. Tables below this share are dropped from the boost set
+   * to avoid amplifying incidental mentions.
+   */
+  private static final float TABLE_WEIGHT_THRESHOLD_RATIO = 0.2f;
   private static final long FLUSH_INTERVAL_SECONDS = 5;
+  /**
+   * Hard upper bound on deserialized entry count to protect against a corrupted/tampered
+   * index file causing unbounded allocation on startup. 10M paragraphs is well beyond any
+   * plausible deployment (~18 GB of vectors alone at 384 floats/entry).
+   */
   private static final int MAX_INDEX_ENTRIES = 10_000_000;
+  private static final String INDEX_FILE_NAME = "embedding_index.bin";
+  /** Binary format version written by {@link #saveIndex()} and required by {@link #loadIndex()}. */
+  private static final int INDEX_VERSION = 3;
   private static final String EXPECTED_MODEL_SHA256 =
       "6fd5d72fe4589f189f8ebc006442dbb529bb7ce38f8082112682524616046452";
 
@@ -153,9 +178,8 @@ public class EmbeddingSearch extends SearchService {
       throw new IOException("Failed to initialize embedding model", e);
     }
 
-    loadIndex();
-    if (zConf.isIndexRebuild()
-        || !Files.exists(indexPath.resolve("embedding_index.bin"))) {
+    boolean indexLoaded = loadIndex();
+    if (shouldBootstrapIndex(zConf, indexLoaded)) {
       notebook.addInitConsumer(this::addNoteIndex);
     }
     flushScheduler.scheduleWithFixedDelay(this::flushIfDirty,
@@ -178,9 +202,8 @@ public class EmbeddingSearch extends SearchService {
         throw new IOException("Failed to initialize embedding model", e);
       }
     }
-    loadIndex();
-    if (zConf.isIndexRebuild()
-        || !Files.exists(indexPath.resolve("embedding_index.bin"))) {
+    boolean indexLoaded = loadIndex();
+    if (shouldBootstrapIndex(zConf, indexLoaded)) {
       notebook.addInitConsumer(this::addNoteIndex);
     }
     flushScheduler.scheduleWithFixedDelay(this::flushIfDirty,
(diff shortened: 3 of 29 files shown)


jongyoul:
@kkalyan Thank you for the contribution. Could you please check the CI first and fix it? Moreover, I will review it but I have a simple question. Do we download the model when we start the server?

kkalyan:
> @kkalyan Thank you for the contribution. Could you please check the CI first and fix it? Moreover, I will review it but I have a simple question. Do we download the model when we start the server?

Hi @jongyoul  - Yes, the model (~86MB) is downloaded on first start when `zeppelin.search.semantic.enable=true` (disabled by default). It's cached at  `{zeppelin.search.index.path}/models/all-MiniLM-L6-v2/` and reused on subsequent starts. I'll fix the CI issues — ESLint brace-style in the classic  UI controller and the missing ASF license header on the docs file.

jongyoul:
I think it's better to have it by default as we need to assume the environment not to download it dynamically. Moreover, don't we need to wait until it's downloaded when starting the server?

kkalyan:
> I think it's better to have it by default as we need to assume the environment not to download it dynamically. Moreover, don't we need to wait until it's downloaded when starting the server?

Thank you @jongyoul. You're right — downloading at startup is problematic for production/air-gapped environments.
 A couple of questions so I get this right:
 1. Should the model be bundled in the distribution (~86MB), or would a configurable local path (`zeppelin.search.model.path`) where admins
   pre-stage the model be better?
 2. If the model isn't found, should Zeppelin fall back to `LuceneSearch` with a warning, or fail fast?
 3. Would an optional helper script (`bin/install-search-model.sh`) be acceptable for the download convenience?
 
Happy to implement whichever direction you prefer.

hyunw9:
This looks like a useful feature.
I have a couple of questions on this PR :
1. Model version pinning: The model URL points to `resolve/main/`, so the weights could silently change if the upstream model is updated. Since `embedding_index.bin` doesn't seem to store which model version generated the vectors, how would we detect a mismatch between an existing index and the current model? Pinning to a specific commit hash might be safer.

2. JNA compatibility: Looks like the DJL tokenizer's JNA dependency is excluded and the runtime falls back to `zeppelin-server`'s JNA 4.1.0. This seems to work for now, but could this break if DJL needs to be upgraded to a version that requires a newer JNA? Just wondering if this is something we should keep in mind.

kkalyan:
@jongyoul:

   Updated in the latest push:

   - Added `bin/install-search-model.sh` — downloads the model (~86MB) and tokenizer from a pinned HuggingFace commit. Run once before enabling semantic
   search.
   - Removed all runtime download logic from `EmbeddingSearch`. Server now fails fast with a clear error message if the model isn't pre-installed.
   - No network I/O at startup.

   Usage:
   ```bash
   bin/install-search-model.sh            # uses default /tmp/zeppelin-index
   bin/install-search-model.sh /my/path   # custom index path

kkalyan:
Thanks for the review @hyunw9!

   1. Model version pinning: Good catch. The URLs are now pinned to a specific HuggingFace commit (c9745ed1d9f2) in bin/install-search-model.sh, and the
   runtime download logic has been removed entirely from EmbeddingSearch.java. The model is pre-installed via the helper script, so there's no risk of
   silent drift. Upgrading the model becomes an explicit, reviewable change to the script.

   2. JNA compatibility: You're right to flag this. DJL tokenizers 0.28.0 pulls JNA 5.14.0, which we exclude to avoid conflicting with Zeppelin's JNA
   4.1.0 (from Hadoop). This works because DJL only uses basic Native.load() which is stable across JNA versions. If a future DJL upgrade requires newer
   JNA APIs, it would fail immediately at startup (tokenizer init), not silently. At that point the fix would be upgrading Zeppelin's JNA globally —
   same pattern other modules follow. Added a comment in the pom.xml exclusion to document this.

kkalyan:
@jongyoul @hyunw9 checking in on this. any other feedback?

zeppelin-zengine/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+  }
+
+  @Override
+  public void deleteNoteIndex(String noteId) {
+    if (noteId == null) {
+      return;
+    }
+    indexLock.writeLock().lock();
+    try {
+      index.entrySet().removeIf(e -> e.getKey().startsWith(noteId));
```
jongyoul: 🐛 **Prefix collision** — if two note IDs share a prefix (e.g. `2A123` and `2A1234`), this removes entries for both. Beyond the functional bug, this means a user deleting their own note can incidentally evict another user's index entries (low-grade availability impact; `LuceneSearch` is not affected since it uses `Term`-based matching).

Since the docId format is `<noteId>/paragraph/<pId>`, suggest:

```java
removeIf(e -> e.getKey().equals(noteId) || e.getKey().startsWith(noteId + "/"));
```

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+  }
+
+  // ---- Persistence ----
+
+  /**
+   * Save index to a binary file.
+   * Format: [int:version=3][int:count] then for each entry:
+   *   [utf:docId] [utf:noteName] [utf:text] [utf:title] [utf:tables] [utf:output] [float[384]:embedding]
+   */
+  private void saveIndex() throws IOException {
```
jongyoul, line 852: This rewrites the **entire** binary index on every mutation, and it's called from every add/update/delete path (`addParagraphIndex`, `updateParagraphIndex`, `addNoteIndex`, `updateNoteIndex`, `deleteNoteIndex`, `deleteParagraphIndex`). For a 50K-paragraph deployment that's ~75MB written to disk on **every paragraph save** — not viable in production.

Options:
- Debounced/scheduled flush (e.g. flush every N seconds if dirty)
- Per-entry on-disk storage
- Append-only log + periodic compaction

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+    Path file = indexPath.resolve("embedding_index.bin");
+    Path tmpFile = indexPath.resolve("embedding_index.bin.tmp");
+    indexLock.readLock().lock();
```
jongyoul, line 858: Holding `indexLock.readLock()` across disk I/O means index mutations are blocked for the entire flush duration. Query reads still proceed (correct semantics), but writers wait. Suggest serializing to a `byte[]` (or `ByteArrayOutputStream`) buffer under the lock, then writing to disk after releasing it.

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+    Files.createDirectories(indexPath);
+
+    try {
+      initModel();
+    } catch (Exception e) {
+      throw new IOException("Failed to initialize embedding model", e);
+    }
+
+    if (zConf.isIndexRebuild()) {
+      notebook.addInitConsumer(this::addNoteIndex);
```
jongyoul, line 193: Initial indexing is only registered when `isIndexRebuild()` is true. If `embedding_index.bin` is missing or partial (first run with `index.rebuild=false`, or corrupted file), there's no bootstrap path — the index will simply stay empty. Suggest also bootstrapping when the index file is absent, e.g.:

```java
if (zConf.isIndexRebuild() || !Files.exists(indexPath.resolve("embedding_index.bin"))) {
  notebook.addInitConsumer(this::addNoteIndex);
}
```

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+      this.output = output;
+    }
+  }
+
+  @Inject
+  public EmbeddingSearch(ZeppelinConfiguration zConf, Notebook notebook) throws IOException {
+    super("EmbeddingSearch");
+    this.notebook = notebook;
+    this.indexPath = Paths.get(zConf.getZeppelinSearchIndexPath());
+    Files.createDirectories(indexPath);
```
jongyoul, line 182: 🔒 **Restrict permissions on the index directory.** `embedding_index.bin` stores paragraph text and **paragraph output** (up to 1000 chars from `extractOutput`) in plaintext, which may include customer PII or financial values. The default index path is `/tmp/zeppelin-index` (world-traversable), and `Files.createDirectories` uses the umask default (typically 755).

This is the same threat class as the existing notebook-on-disk directory, but the new feature broadens it by including execution **output**. Suggest:

```java
Files.createDirectories(indexPath);
if (Files.getFileStore(indexPath).supportsFileAttributeView("posix")) {
  Files.setPosixFilePermissions(indexPath,
      PosixFilePermissions.fromString("rwx------"));
}
if (indexPath.toAbsolutePath().startsWith("/tmp")) {
  LOGGER.warn("zeppelin.search.index.path is under /tmp ({}); "
      + "paragraph text and output will be readable by other local users. "
      + "Consider setting it to a private directory.", indexPath);
}
```

Also apply `0600` to `embedding_index.bin` itself in `saveIndex()` after the atomic move. Disk-level encryption (LUKS, FileVault) remains the operator's responsibility; this is just the in-app baseline.

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+    if (!Files.exists(modelFile) || !Files.exists(tokenizerFile)) {
+      throw new IOException(
+          "Embedding model not found at " + modelDir + ". "
+              + "Run bin/install-search-model.sh before enabling semantic search.");
+    }
+
+    ortEnv = OrtEnvironment.getEnvironment();
+    OrtSession.SessionOptions opts = new OrtSession.SessionOptions();
+    opts.setIntraOpNumThreads(Runtime.getRuntime().availableProcessors());
+    ortSession = ortEnv.createSession(modelFile.toString(), opts);
```
jongyoul, line 260: 🔒 **Re-verify model integrity at load time.** Even if `install-search-model.sh` validates SHA256 at install time, anyone with write access to the index dir (see the permissions comment above) can swap `model.onnx` between Zeppelin restarts. Loading an attacker-supplied ONNX model into ORT has historically been an RCE vector.

Suggest hardcoding the expected SHA256 of `model.onnx` and `tokenizer.json` (matching the commit pinned in `install-search-model.sh`) and refusing to start if they don't match.

zeppelin-zengine/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+    if (!Files.exists(file)) {
+      return;
+    }
+    try (DataInputStream in = new DataInputStream(Files.newInputStream(file))) {
+      int first = in.readInt();
+      int version;
+      int count;
+      if (first >= 2 && first <= 3) {
+        version = first;
+        count = in.readInt();
```
jongyoul: 🔒 **Bound the entry count from the index file.** A corrupted or tampered `embedding_index.bin` with `count = Integer.MAX_VALUE` (or any large value) drives this loop into per-entry allocation until OOM. The surrounding `try { } catch (IOException)` does **not** catch `OutOfMemoryError`, so this is a startup DoS.

Suggest a sanity bound, e.g.:

```java
if (count < 0 || count > 10_000_000) {
  LOGGER.warn("embedding_index.bin reports implausible entry count {}; ignoring file", count);
  return;
}
```

zeppelin-zengine/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,706 @@
...
+      // Truncate to max sequence length
+      int seqLen = Math.min(inputIds.length, MAX_SEQ_LENGTH);
+      long[] ids = new long[seqLen];
+      long[] mask = new long[seqLen];
+      long[] tokenTypeIds = new long[seqLen];
+      System.arraycopy(inputIds, 0, ids, 0, seqLen);
+      System.arraycopy(attentionMask, 0, mask, 0, seqLen);
+
+      long[] shape = {1, seqLen};
+      OnnxTensor idsTensor = OnnxTensor.createTensor(ortEnv, LongBuffer.wrap(ids), shape);
```
jongyoul: 🪲 **Native tensor leak on partial allocation failure.** Three `OnnxTensor.createTensor` calls happen sequentially before the `try { ... } finally { close() }` block; if the second or third call throws, the earlier tensors are never closed (native memory leak).

Suggest creating each tensor inside its own try-with-resources, or accumulating successfully-created handles in a list and closing all in `finally`.

bin/install-search-model.sh
```diff
@@ -0,0 +1,49 @@
...
+mkdir -p "${MODEL_DIR}"
+
+download() {
+  local url="$1" dest="$2"
+  if [ -f "${dest}" ]; then
+    echo "Already exists: ${dest}"
+    return
+  fi
+  echo "Downloading ${url} ..."
+  curl -fSL --connect-timeout 30 --max-time 300 -o "${dest}.tmp" "${url}"
```
jongyoul, line 69: 🔒 **Verify SHA256 of downloaded files.** The script pins a HuggingFace commit SHA, which protects against repository content drift, but it does not verify the bytes received. ORT 1.18.x has had RCE/DoS CVEs around model deserialization, so the following scenarios remain exploitable:

- HuggingFace storage/CDN compromise
- MITM with a valid-looking certificate (corporate proxy, captured CA)
- Internal mirror/cache poisoning

Suggest hardcoding the expected SHA256 of `model.onnx` and `tokenizer.json` and verifying after download:

```bash
echo "<sha256>  ${dest}" | sha256sum -c -
```

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -75,110 +47,49 @@ export class NotebookSearchResultItemComponent implements OnChanges, OnDestroy {
...
-  }
-
-  setHighlightKeyword(): void {
-    let mergedStr = this.result.header ? `${this.result.header}\n\n${this.result.snippet}` : this.result.snippet;
 
-    const regexp = /<B>(.+?)<\/B>/g;
-    const matches = [];
-    let match = regexp.exec(mergedStr);
+    // snippet = SQL/code, header = tables + output
+    this.codeText = (this.result.snippet || '').replace(/<\/?B>/gi, '');
```
jongyoul: Stripping `<B>` tags here removes the keyword highlighting that `LuceneSearch` provides — so users who never enable semantic search will lose highlighting in search results too. Since `EmbeddingSearch` is opt-in (default off), this is a regression for the default user.

Could you preserve highlight rendering for the Lucene path? E.g. convert `<B>...</B>` into `<mark>` via sanitized `innerHTML`, or render it manually via a small parsing pass.

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -75,110 +47,49 @@ export class NotebookSearchResultItemComponent implements OnChanges, OnDestroy {
...
+    if (!text) {
+      return '';
     }
-  }
-
-  setLanguage() {
-    const model = this.editor?.getModel();
-    if (!model) {
-      throw new Error('Editor model is not defined.');
+    if (/select|insert|create|from|where/i.test(text)) {
```
jongyoul: `/select|insert|create|from|where/i.test(text)` will misclassify any paragraph containing the word "create" as SQL — including markdown like "Click Create to continue" or python comments mentioning "insert". Suggest:

1. Check the interpreter prefix (`%spark.sql`, `%md`, …) first — this is reliable
2. Fall back to the keyword heuristic only if no prefix is present

The backend `IndexEntry` could also include the resolved interpreter name to avoid duplicating detection in the frontend.

zeppelin-web-angular/tsconfig.base.json
```diff
@@ -12,6 +12,8 @@
     "outDir": "./dist/out-tsc",
     "sourceMap": true,
     "strict": true,
+    "noImplicitAny": false,
```
jongyoul: These two flags relax type safety **project-wide** and aren't required by the semantic search feature. The accompanying `as any` casts in `credential.component.ts`, `code-editor.component.ts`, `save-as.service.ts`, `run-scripts.directive.ts` look unrelated too.

Could you split the tsconfig changes (and the related casts) into a separate PR? It would be much easier to review the underlying type issues independently, and avoids permanently lowering the project's type-safety bar as a side effect of a search feature.

jongyoul (changes requested):

jongyoul:
Hello, I checked logic in details and found some improvement points. Some were critical issues and others were not, but I left all of them for your reference. Please check the list below:

## Critical

1. **Unrelated TypeScript strictness regression** — `tsconfig.base.json` and `zeppelin-sdk/tsconfig.json` add `noImplicitAny: false` and `skipLibCheck: true`, plus several `as any` casts in `credential.component.ts`, `code-editor.component.ts`, `save-as.service.ts`, `run-scripts.directive.ts`. None of these are related to semantic search and they permanently lower project-wide type safety. Please split these out — ideally fix the underlying type issues in a separate PR.

2. **`deleteNoteIndex` prefix collision bug** (`EmbeddingSearch.java:548`) — `startsWith(noteId)` will also remove entries whose note IDs share the prefix (e.g. `2A123` vs `2A1234`). Beyond the functional bug, this is also a low-grade availability concern: a user deleting their own note can incidentally remove another user's index entries (existing `LuceneSearch` is not affected since it uses `Term`-based matching). See inline.

3. **`saveIndex()` rewrites the entire binary on every mutation** (`EmbeddingSearch.java:635`) — called from every add/update/delete path. With 50K paragraphs that's ~75MB written to disk on every paragraph save. Not viable for production. See inline.

4. **Initial indexing only registered when `isIndexRebuild`** (`EmbeddingSearch.java:137-138`) — if `embedding_index.bin` is missing or partial, there's no bootstrap path. See inline.

5. **UI regression for default (Lucene) users** (`result-item.component.ts:55`) — the new UI strips `<B>` highlight tags, so users who never enable semantic search lose keyword highlighting in search results. See inline.

## Major

- **`LICENSE` file not updated for bundled dependencies** — both [ONNX Runtime (MIT)](https://central.sonatype.com/artifact/com.microsoft.onnxruntime/onnxruntime) and DJL Tokenizers (Apache-2.0) are Category A, so they are license-compatible (no `dev@`/Legal review needed; see [LEGAL-230](https://issues.apache.org/jira/browse/LEGAL-230) for the precedent on JNI native binaries like Snappy-Java). However, per the [ASF licensing-howto](https://infra.apache.org/licensing-howto.html) bundled dependencies should also be listed in `LICENSE` (not just `NOTICE`), with the bundled artifact path, license type, source URL, and license text or a link to it.
- **Index file/directory should be created with restrictive permissions (security)** — `EmbeddingSearch` indexes paragraph **output** (`extractOutput`, up to 1000 chars) into `embedding_index.bin` in plaintext. Output may contain customer PII, monetary values, etc. The default index path is `/tmp/zeppelin-index` (world-traversable), and `Files.createDirectories` uses the umask default (typically 755). This widens the existing notebook-on-disk surface — same threat class as the notebook directory, but the new feature adds output text to it. Suggest creating the index dir/file with mode `0700` and warning if the configured path starts with `/tmp`. Disk-level encryption remains the operator's responsibility. See inline.
- **Model file integrity not verified after download (supply chain)** — `bin/install-search-model.sh` pins to a HuggingFace commit SHA but doesn't verify SHA256 of the downloaded files. ONNX models can trigger RCE/DoS via deserialization (multiple CVEs in ORT 1.18.x), so MITM/cache poisoning/repository compromise scenarios should be hardened. See inline.
- **Model file integrity not re-verified at runtime** — `ortSession = ortEnv.createSession(modelFile.toString(), opts)` loads whatever is on disk. Combined with the permission issue above, anyone with write access to the index dir can swap the model. A SHA256 comparison at startup would provide defense-in-depth.
- **Index file deserialization is unbounded (DoS)** — `loadIndex` reads `count = in.readInt()` and loops `count` times, allocating per-entry memory. A corrupted/tampered index file with `count = Integer.MAX_VALUE` causes startup OOM (caught only as `IOException` — `OutOfMemoryError` is not caught). Suggest a sanity bound (e.g. 10M entries). See inline.
- **`saveIndex` performs disk I/O while holding `indexLock.readLock()`** (`EmbeddingSearch.java:638`) — query reads still proceed (correct), but all index mutations block until the disk flush completes. Serialize to a buffer first, then write outside the lock.
- **Synchronous embedding on the notebook event thread** — `addParagraphIndex` runs ONNX inference (~5ms) plus a full disk write on the same thread that delivers paragraph events. Consider an async worker queue with batching.
- **CI does not run the new tests** — `@EnabledIfEnvironmentVariable("ZEPPELIN_EMBEDDING_TEST", "true")` means default `mvn test` skips them entirely. Please add a CI job (or Maven profile) that exercises `EmbeddingSearchTest` with the model downloaded/cached.
- **`canIndexAndQuery` doesn't actually validate semantic behavior** — the assertion `r.get("text").contains("all")` is a substring check that keyword search would also pass. `semanticSearchFindsRelatedConcepts` is the only true semantic test; consider adding more cases like that.

## Minor

- **`detectInterpreter` false positives** (`result-item.component.ts:78` and the matching `detectLang` in classic UI) — `/select|insert|create|from|where/i` will tag any markdown containing the word "create" as SQL. Suggest checking the interpreter prefix (`%spark.sql`, `%md`, …) first and falling back to heuristics.
- **`extractTables` only matches `schema.table` form** — by design, but the limitation should be documented. Single-name tables (`FROM customers`) and dialect quoting are missed, so the table-boost phase is effectively a no-op for transactional-DB style queries. A proper SQL parser (JSqlParser, Calcite) would be more accurate but is a reasonable follow-up rather than a blocker.
- **Native tensor leak on partial allocation failure** — three `OnnxTensor.createTensor` calls in `embed()` happen outside the try-with-resources block; if the second/third call throws, the earlier tensors are never closed. Wrap each tensor allocation so all successfully-created handles are released.
- **Magic numbers** scattered through `EmbeddingSearch` (`MIN_SIMILARITY=0.25`, `TABLE_BOOST=0.05`, threshold `0.2`, text limits 1500/2000/500/300). Consider exposing the tunables as `ConfVars`.
- **`ImmutableMap.of(...)` with 5 entries** in `query()` is at Guava's positional-overload limit; a builder or `Map.of(...)` is safer for future field additions.
- **Classic UI inline styles** (`result-list.html`) — long `style="..."` attributes; please move into the existing CSS file.
- **ONNX Runtime 1.18.0 / DJL 0.28.0** are slightly behind current stable; Java APIs are stable across minor versions, but bumping picks up security/native fixes. Worth a separate dependency-bump PR.
- **Test uses `Files.createSymbolicLink`** to share the model dir across tests — fails on Windows without admin/dev mode. Cleaner solution: make `EmbeddingSearch` accept the model directory as a separate config so the test can point both `indexDir` and `modelDir` independently.

## Suggested follow-ups (non-blocking)

- Configurable model URL for air-gapped environments (already in your "Future Work" — good)

jongyoul:
@tbonelee @dididy Could you please help review for the FE part?

@kkalyan Could you please check the CI as well?

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
```diff
@@ -360,7 +360,7 @@ export class NotebookParagraphCodeEditorComponent
       return;
     }
     const text = model.getValue();
-    const newDecorations = [];
+    const newDecorations: any[] = [];
```
voidmatcha:
```suggestion
    const newDecorations: MonacoEditor.IModelDeltaDecoration[] = [];
```

zeppelin-web-angular/src/app/pages/workspace/credential/credential.component.ts
```diff
@@ -146,7 +146,7 @@ export class CredentialComponent {
     this.credentialService.getCredentials().subscribe(data => {
       const controls = [...Object.entries(data.userCredentials)].map(e => {
         const entity = e[0];
-        const { username, password } = e[1];
+        const { username, password } = e[1] as any;
```
voidmatcha:
```suggestion
        const { username, password } = e[1] as CredentialForm;
```

zeppelin-web-angular/tsconfig.base.json
voidmatcha: I agree with this review. In the same vein as `skipLibCheck`, I think it would be better to remove these configuration changes as well.

zeppelin-web-angular/src/app/share/run-scripts/run-scripts.directive.ts
```diff
@@ -32,10 +32,10 @@ export class RunScriptsDirective implements OnChanges {
     if (!this.scriptsContent.toString()) {
       return;
     }
-    this.ngZone.onStable.pipe(take(1)).subscribe(() => {
+    (this.ngZone.onStable as any).pipe(take(1)).subscribe(() => {
```
voidmatcha:
```suggestion
    (this.ngZone.onStable).pipe(take(1)).subscribe(() => {
```

zeppelin-web-angular/projects/zeppelin-sdk/tsconfig.json
```diff
@@ -5,6 +5,8 @@
     "target": "es2015",
     "declaration": true,
     "inlineSources": true,
+    "skipLibCheck": true,
+    "noImplicitAny": false,
```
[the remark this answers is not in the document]
voidmatcha: https://github.com/apache/zeppelin/pull/5218/changes#r3152807288
Same reason applies.

voidmatcha (changes requested):
I left a small type-related review. I’ll continue reviewing the frontend changes as the PR progresses.

https://github.com/dididy/zeppelin/commit/7ae095f9bc1f99a5183c7f02651ccc996e88f13c

I verified that tsc passes locally after applying the suggested changes.

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.html
```diff
@@ -12,11 +12,18 @@
...
-  ></zeppelin-code-editor>
+  <div *ngIf="codeText" class="code-block">
+    <pre>{{ codeText }}</pre>
+  </div>
+  <div *ngIf="outputText" class="output-block">
+    <pre>{{ outputText }}</pre>
+  </div>
+  <div *ngIf="tablesText" class="tables-block">
+    📊 {{ tablesText }}
+  </div>
```
voidmatcha:
```suggestion
  <div *ngIf="tablesText" class="tables-block">📊 {{ tablesText }}</div>
```
The CI(run-playwright-e2e-tests) is failing due to lint issues, so this part needs to be fixed. Running `npm run lint:fix` in `zeppelin-web-angular` resolves the problem.

voidmatcha: run-playwright-e2e-tests passed after applied two commits.

https://github.com/dididy/zeppelin/commit/7ae095f9bc1f99a5183c7f02651ccc996e88f13c
https://github.com/voidmatcha/zeppelin/commit/15fc35e1a8aca5c9ef6aceb853e8907eff6131ff

https://github.com/voidmatcha/zeppelin/actions/runs/25116480167/job/73604863894

zeppelin-web-angular/src/app/pages/workspace/credential/credential.component.ts
tbonelee: I think `e` is already properly typed as a proper tuple type without type assertion(`as`). So we could just remove `as any` instead of changing it to `as CredentialForm`.

zeppelin-web-angular/tsconfig.base.json
tbonelee: I think all `tsconfig` changes and type assertions (`as`) are not necessary. Also those type assertions could be handled with explicit typings or implicit type-narrowings.

zeppelin-web-angular/src/app/share/run-scripts/run-scripts.directive.ts
tbonelee:
```suggestion
    this.ngZone.onStable.pipe(take(1)).subscribe(() => {
```
A minor point, but I think the parentheses are unnecessary as well.

zeppelin-web-angular/projects/zeppelin-sdk/tsconfig.json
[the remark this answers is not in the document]
tbonelee: Maybe we could just revert those added lines to keep the git history clean, since their default values are already `skipLibCheck=false` and `noImplicitAny=true`.

zeppelin-web-angular/src/app/pages/workspace/notebook/notebook.component.ts
```diff
@@ -321,7 +321,7 @@ export class NotebookComponent extends MessageListenersManager implements OnInit
     this.securityService.getPermissions(note.id).subscribe(data => {
       this.permissions = data;
       this.isOwner = !(
-        this.permissions.owners.length && this.permissions.owners.indexOf(this.ticketService.ticket.principal) < 0
+        this.permissions?.owners?.length && this.permissions.owners.indexOf(this.ticketService.ticket.principal) < 0
```
tbonelee:
```suggestion
        this.permissions.owners.length && this.permissions.owners.indexOf(this.ticketService.ticket.principal) < 0
```

Both `this.permissions` and `.owners` are guaranteed to be defined here: `this.permissions` is assigned from `data` on the line above, and `Permissions.owners` is typed as non-optional `string[]`. The `?.` chaining isn't needed. Could we revert this? Looks unrelated to semantic search.

zeppelin-web-angular/src/app/pages/workspace/notebook/paragraph/code-editor/code-editor.component.ts
tbonelee: It makes sense but we'd better only handle the changes related to new feature in this PR. So I think it would be better to revert this change.

```suggestion
    const newDecorations = [];
```

zeppelin-web-angular/src/app/services/save-as.service.ts
```diff
@@ -19,7 +19,7 @@ export class SaveAsService {
   saveAs(content: string, filename: string, extension: string) {
     const BOM = '\uFEFF';
     const fileName = `${filename}.${extension}`;
-    const binaryData = [];
+    const binaryData: string[] = [];
```
tbonelee:
```suggestion
    const binaryData = [];
```
Could you roll back this so that this PR only handles the Semantic search-related things?

zeppelin-web-angular/src/app/share/run-scripts/run-scripts.directive.ts
tbonelee, line 1: Could you also roll back this, since it does not affect the new feature implementation?

zeppelin-web-angular/src/app/utility/get-keyword-positions.ts
tbonelee, line 1: I alse suggest rolling back this file for the same reason as before.

tbonelee (changes requested):
Thanks for working on this feature. I focused my review on the frontend changes.

To keep this PR focused on what it sets out to do, would you consider reverting the tsconfig changes and the related type fixups that aren't tied to semantic search? Splitting them into a separate PR would make this one easier to evaluate and avoid weakening the project-wide strict settings as a side effect of a feature.

I also noticed that highlighting that previously appeared in LuceneSearch results now renders as plain text in the UI. The `<B>...</B>` markup that Lucene's `SimpleHTMLFormatter` produces is currently stripped on the frontend, so existing search results lose visual emphasis on the matched terms. It would be good to preserve the existing behavior for LuceneSearch responses.

The API contract between backend and frontend has always been implicit, but this PR makes it noticeably more complex (two backends with different `header` shapes, a multi-line layout, the `📊` prefix, optional highlight tags, and `snippet` meaning different things across implementations). It would help future contributors a lot if this were written down explicitly in one place, for example a "Response contract" section in `embedding-search.md` listing the keys, the per-backend differences, and the `header` line format.

Using an emoji as a machine-readable delimiter for table names also seems fragile in practice. The same `📊 ` prefix is produced in `EmbeddingSearch.java:504` and parsed in two frontends (`result-item.component.ts:64`, `result-list.controller.js`), with no shared constant or schema tying them together. The current parser splits `header` by newline and classifies any line starting with `📊` as a table, which means a paragraph whose output happens to start with `📊` (for example a `%md` cell titled `📊 Sales summary`, or printed status text containing the emoji) would appear in the UI as if it were a SQL table reference. The same parser is also brittle to newlines in user-provided fields like paragraph titles, which the backend currently appends to `header` verbatim. It would be worth exploring an alternative approach that doesn't rely on parsing user-visible content as a structural delimiter.

kkalyan:
@jongyoul: Thank you for the thorough review — Here's what's addressed in the latest push:

**Critical:** Reverted all unrelated tsconfig/type changes. Fixed `deleteNoteIndex` prefix collision. Replaced per-mutation `saveIndex()` with a  debounced 5s flush (dirty flag + `ScheduledExecutorService`). Index now also bootstraps when `embedding_index.bin` is missing. Lucene `<B>`  highlighting is preserved (converted to `<mark>` tags).

  **Security:** Index directory created with `0700`, files with `0600`, warning logged if path is under `/tmp`. SHA256 verification added to both   `install-search-model.sh` and runtime `initModel()`. `loadIndex` now has a 10M entry sanity bound. `saveIndex` serializes to buffer under lock,   writes to disk outside lock.
   **Other:** Fixed OnnxTensor leak on partial allocation. Replaced `📊` emoji delimiter with `[TABLES]` prefix. Updated `LICENSE` file. Improved  test assertions for semantic search behavior. `NoteEventAsyncListener` already handles async dispatch, so no separate executor needed.
   
Still on the list for follow-up PRs: CI job for embedding tests, configurable model URL, magic number extraction to `ConfVars`, dependencyversion bumps.

kkalyan:
@tbonelee   Thank you for the detailed frontend review. All tsconfig changes and unrelated type fixups have been reverted. Lucene `<B>` highlighting is now  preserved — converted to `<mark>` tags via `innerHTML` binding in Angular and `ng-bind-html` in Classic. The `📊` emoji delimiter has been replaced   with a structured `[TABLES]` prefix to avoid collision with user content. I'll add a "Response contract" section to `embedding-search.md` in a  follow-up.

kkalyan:
@voidmatcha: Thanks for catching the lint issues and suggesting the type fixes. All unrelated type changes have been reverted per the consensus. The Angular  template lint issue (`📊` whitespace) is resolved since we switched to `[TABLES]` prefix with a single-line `<div>`. The `run-scripts.directive.ts`  and `credential.component.ts` are back to their original forms.

jongyoul (commented):
Thanks @kkalyan for the quick turnaround on the previous round — the fixes all look good:
- Debounced flush + dirty flag with serialize-then-write-outside-lock
- `deleteNoteIndex` prefix collision (`startsWith(noteId + "/")`)
- Bootstrap path also triggers when `embedding_index.bin` is missing
- POSIX `0700`/`0600` permissions on index dir/file + `/tmp` warning
- `OnnxTensor` lifecycle in `embed()`
- SHA256 verification in both `install-search-model.sh` and at runtime
- Reverted unrelated `tsconfig` / `as any` changes
- Preserved Lucene `<B>` → `<mark>` highlighting

A few more rounds of feedback below — mostly architectural/follow-up rather than blocker.

---

### Sharding (saveIndex linear cost)

The debounce helps with frequency, but `saveIndex()` still re-serializes the entire index every flush. One paragraph edit rewrites everything, so cost grows with total index size. Could we shard by note (e.g. `index/notes/<noteId>.bin`) so an edit only rewrites that note's file? WDYT?

---

### In-memory duplication

Do we need to keep paragraph text/title/output in `IndexEntry`? `NoteManager` already has an LRU cache and lazy-loads notes via `notebook.processNote()`. Caching the same data here doubles the memory footprint and adds a consistency surface on edits. Could we keep just the embedding + docId in memory, and pull paragraph fields from `Notebook` when building the top-K results?

---

### Authorization

Authorization is a concern here. `query()` truncates to top-20 before returning, and the auth filter in the REST layer runs after that. So if the top-20 are all notes the caller can't read, they'll see "no results" even though there are matching notes they're authorized for further down — a regression vs. the existing keyword search.

The table boost phase has a similar issue: `relevantTables` is computed from the unfiltered top-20, so table names from inaccessible notes end up shaping the ranking the caller sees.

I think `query()` should take user/roles (or a readability predicate) and apply the filter before the top-K cutoff.

---

### `updateNoteIndex` re-embeds every paragraph

Tracing the call path: `NotebookService.updateNote` → `Notebook.fireNoteUpdateEvent` → `SearchService.handleNoteUpdateEvent` → `updateNoteIndex`. That covers note rename and note config changes (cron, etc.) — paragraph edits go through a separate event path.

Right now `updateNoteIndex` re-embeds every paragraph in the note via `indexNote(note)`. For a config-only change (e.g. updating a cron schedule) that's pure waste, and even for a rename it's heavy. `LuceneSearch.updateNoteIndex` only refreshes the note-name field for this reason.

Could this just update the `noteName` field on existing `IndexEntry`s in place and skip the embed loop? Note name is a small part of `buildParagraphText`, so the slight embedding staleness self-heals on the next paragraph touch.

---

### `flushIfDirty` loses dirty state on failure

```java
if (indexDirty.compareAndSet(true, false)) {
  try {
    saveIndex();
  } catch (IOException e) {
    LOGGER.error("Failed to flush embedding index to disk", e);
  }
}
```

If `saveIndex()` throws, the dirty flag was already cleared, so the failed write is silently dropped until the next mutation re-sets dirty. Worth adding `indexDirty.set(true)` in the catch so the next tick retries.

---

### Dead version-handling in `loadIndex`

`loadIndex` branches on v1/v2/v3 of the binary format:

```java
if (first >= 2 && first <= 3) {
  version = first;
  ...
} else {
  version = 1;
  ...
}
```

But this PR introduces the format — there's no v1/v2 in the wild. Worth dropping the legacy branches and keeping just v3 to reduce surface area.

---

### Magic constants

A few tuning knobs are hard-coded with no rationale or way to override:
- `MIN_SIMILARITY = 0.25f`
- `TABLE_BOOST = 0.05f`
- top-K cutoff of `20` (also duplicated in the Phase-1 table-collection loop and `MAX_RESULTS`)
- `MAX_INDEX_ENTRIES = 10_000_000`

Could these be config keys (or at least documented constants with a comment on how they were chosen)? The `20` literal in particular should reuse `MAX_RESULTS`.

---

### Partial-load bootstrap gap

If `loadIndex()` hits an `IOException` mid-read, it logs a warning and returns with the in-memory index in whatever state it ended up in. But the bootstrap trigger only fires when the file is **missing** (`!Files.exists(...)`), not when it's present-but-corrupt. So a partially-loaded or torn index file will sit there indefinitely. Could the bootstrap also trigger on load failure (e.g. delete the bad file or set a flag that forces re-index)?

kkalyan:
@jongyoul Thanks for the second pass. Addressed the smaller items in this push (bc3d252); pulling the three architectural ones out to a follow-up PR with
  rationale below.

  Landed in this push

  - flushIfDirty losing dirty state on failure — Good catch. Now re-sets indexDirty in the catch and logs "will retry on next tick" so the next scheduled
  flush picks it up instead of waiting on the next mutation.
  - updateNoteIndex re-embedding every paragraph — Rewrote to mirror LuceneSearch: only refreshes the noteName field on existing IndexEntrys under the write
  lock, skips entries that already have the new name, and calls markDirty() only if anything actually changed. Cron-only edits are now free; a rename is
  O(paragraphs-in-note) in-memory with zero ONNX calls. Embedding drift self-heals on the next paragraph touch, as you noted.
  - Dead v1/v2 branches in loadIndex — Dropped. loadIndex now only handles INDEX_VERSION = 3, returns boolean for success, and unknown versions are treated
  as load failure (fed into the bootstrap path below).
  - Partial-load bootstrap gap — New shouldBootstrapIndex(zConf, loaded) helper triggers rebuild on any of: isIndexRebuild(), missing file, or load failure.
  On load failure it also Files.deleteIfExists the bad file before rebuild so the new index is written fresh rather than appending to a torn one.
  - Magic constants — Hoisted INDEX_VERSION, INDEX_FILE_NAME, and TABLE_WEIGHT_THRESHOLD_RATIO to named constants; replaced the 20 literal in Phase-1 table
  collection with MAX_RESULTS; added Javadoc on MIN_SIMILARITY, TABLE_BOOST, and MAX_INDEX_ENTRIES explaining how they were chosen.

  Deferred to follow-up PR

  These three are real, but each changes either the SearchService contract or the persistence layer in ways I don't want to fold into this PR:

  - Sharding (index/notes/<noteId>.bin) — Agreed the linear rewrite cost is the long-term concern. Done properly this needs a per-note lock strategy, a
  manifest/index-of-indices for load, and a compaction story for deletes. The current debounced global flush is bounded by the 5s interval and happens off
  the event thread, so it's acceptable as a v1. I'd rather do the sharding redesign on its own PR where we can also decide whether to move to an append-only
  log.
  - In-memory text/title/output duplication vs. NoteManager LRU — You're right that IndexEntry duplicates data that Notebook.processNote() can produce on
  demand. Switching to "embedding + docId only, rehydrate on read" is the right direction, but it changes the hot path: every query() result becomes
  MAX_RESULTS calls to processNote (which can load from disk under LRU pressure). Wanted to benchmark that against the current in-memory path before
  committing, and handle the consistency/eviction edges. Follow-up.
  - Authorization filter ordering — This one's the most important of the three. Agreed that truncating to top-20 before the REST auth filter runs is a
  regression vs. LuceneSearch, and that the table-boost phase contaminating ranking with inaccessible notes is worse than no-results. The right fix is a
  query(String, AuthenticationInfo) overload (or a Predicate<String> canRead parameter) applied before the Phase-1 table collection and before the top-K
  cut. That touches the SearchService interface and both LuceneSearch and EmbeddingSearch — cleaner as a dedicated PR with a test that exercises the auth
  filter specifically.

  Happy to file the three follow-ups against ZEPPELIN-6411 (or split JIRAs) so they don't get lost — just let me know which you'd prefer.

zeppelin-server/src/main/java/org/apache/zeppelin/search/EmbeddingSearch.java
```diff
@@ -0,0 +1,909 @@
...
+  /** Load index from disk if it exists. Supports v1/v2/v3 formats. */
+  /**
+   * Load the index from disk.
+   *
+   * @return {@code true} if the index loaded successfully (or file was absent);
+   *         {@code false} if the file was present but failed to load or was corrupt,
+   *         signalling the caller to trigger a bootstrap rebuild.
+   */
+  private boolean loadIndex() {
+    Path file = indexPath.resolve("embedding_index.bin");
```
jongyoul: It should be INDEX_FILE_NAME.

jongyoul:
@kkalyan Thank you for handling a lot of comments promptly. The changes looks good to me. Moreover, your suggestion looks reasonable as well. To track them clearly, could you please create new tickets for those issues and leave comments with `TODO(ZEPPELIN-XXXX): short summary`? Moreover, passing CI is also a bar to merge this PR. Please triage them if they are flaky.

kkalyan:
Thanks @jongyoul! Addressing the two remaining items.

  **Tickets & TODO markers** — filed and wired up:

  - ZEPPELIN-6412 — EmbeddingSearch: shard persistence by note
  - ZEPPELIN-6413 — EmbeddingSearch: reduce IndexEntry in-memory duplication
  - ZEPPELIN-6414 — EmbeddingSearch: apply authorization filter before top-K cutoff

  `TODO(ZEPPELIN-XXXX): <summary>` comments are in place at the three call sites in `EmbeddingSearch.java`
  (`IndexEntry` → 6413, `query()` → 6414, `saveIndex()` javadoc → 6412). Also rebased onto current master to
  pick up the recent dependabot merges.

  **CI triage** — four reds on the last push; triaged each against the last 20 master commits and the earlier
  pushes on this branch:

  | Check | Fails on master (last 20) | Earlier pushes here | Verdict |
  |---|---|---|---|
  | `test-selenium-with-spark-module-for-spark-3-5` | **14/20** | fails on every push | Pre-existing master
  breakage, not caused by this PR |
  | `run-playwright-e2e-tests (auth, 3.9)` | 5/20 | passed on 7cede8e, failed on bc3d252; `anonymous` variant
  passes on the same push | Flaky |
  | `run-e2e-tests-in-zeppelin-web` | 1/20 | passed on 7cede8e, failed on bc3d252 | Flaky (classic UI e2e) |
  | `spark-test (3.9, 17)` | 0/20 | passed on every earlier push | Infra flake — failed at `Setup conda
  environment` (`conda failed with exit code 1`), before any test code ran |

  None of them point to code introduced in this PR. The rebase push will re-trigger CI; I'll ping again if
  anything still fails after a fresh run.

docs/embedding-search.md
```diff
@@ -0,0 +1,205 @@
...
+- `zeppelin-zengine/pom.xml` — Add `onnxruntime` and `djl-tokenizers` dependencies
+- `zeppelin-zengine/.../conf/ZeppelinConfiguration.java` — Add `ZEPPELIN_SEARCH_SEMANTIC_ENABLE`
+- `zeppelin-server/.../server/ZeppelinServer.java` — Wire `EmbeddingSearch` based on config
+- `NOTICE` — Attribution for ONNX Runtime and DJL
+
+### Modified files — Frontend
+- `zeppelin-web-angular/.../result-item/` — Render search results with separate
+  code block, output block, and table name display (replaces Monaco editor)
+- `zeppelin-web/src/app/search/` — Same improvements for Classic UI
+- Various TypeScript build fixes (`tsconfig`, type annotations)
```
tbonelee:
```suggestion
```

tbonelee (changes requested):
Thanks for switching away from the emoji. `[TABLES]` does address the most obvious user-content collision case.

While reading through the code I noticed a few things I wanted to surface together:

1. **Title appears to leak into the output area even on a normal response.** The parser in `result-item.component.ts` classifies lines starting with `[TABLES]` as `tablesText` and everything else as `outputText`, with no dedicated branch for title. So even on a well-formed EmbeddingSearch response like `"Sales Q4\n[TABLES]analytics.daily_sales\n\nrevenue:1.2M..."`, I confirmed that `Sales Q4` (the user-set title) ends up classified as `outputText` and rendered in the output area.

2. A user-supplied title or output starting with `[TABLES]` looks like it could trigger a false positive. Could you double-check this? For example, a paragraph titled `[TABLES] used in this query`, or a cell whose output is `print("[TABLES] in this dataset:")`, would have that line classified as a SQL table list and rendered as a fake table in the UI.

3. The same parser handles both `LuceneSearch` and `EmbeddingSearch` responses, but I don't see a backend signal in the response itself. `LuceneSearch`'s `header` on master means "highlighted paragraph title" (single line), so when it goes through the current parser it doesn't start with `[TABLES]` and ends up classified as `outputText`, meaning the title shows up in the output area in LuceneSearch mode too, if I'm reading the flow correctly. Could you confirm?

---

@jongyoul The kind of user-input collision in (2) probably isn't common in normal usage. But putting it together with (1) and (3), where the classification goes off even on the well-formed flow, I'm wondering if the in-band marker + line-split approach is fundamentally trying to carry two responsibilities on the same channel. So I'd like to suggest splitting `title` / `tables` / `output` into separate response fields in this PR. What do you think? That would remove the in-band marker entirely, drop the need for the FE to disambiguate the two backends, and as a side benefit preserve master's `header` meaning ("highlighted single-line title") so we avoid the subtle shape change for any external consumer parsing the existing response.

On timing, once this ships externally, reshaping the response later gets more expensive, so I'd lean toward folding this into the current PR if possible. Would appreciate your thoughts.

kkalyan:
@tbonelee Thanks for the detailed review — all three points were spot on. I've addressed them in the latest push:

Separate response fields: I replaced the in-band [TABLES] marker approach with dedicated title, tables, and output fields in the search  response. Both LuceneSearch and EmbeddingSearch now emit the same set of keys, LuceneSearch returns empty strings for tables and output, while EmbeddingSearch populates them. The header field is preserved for backward compatibility (contains the paragraph title in both backends).
This eliminates all three issues you flagged:
 1. Title no longer leaks into the output area — title is a dedicated field, not mixed into header lines.
 2. No more user-content collision — there's no in-band marker to collide with. A paragraph titled [TABLES] used in this query is just a title string.
3. No need to disambiguate backends on the frontend — both backends return the same structured keys, so the frontend reads result.tables and result.output directly without parsing.

 I also added a "Search Result Response Contract" table in docs/embedding-search.md documenting all keys and per-backend differences.

The frontend parsing in both Angular and Classic UIs is significantly simplified — the line-by-line [TABLES] parser is gone, replaced by  direct field reads.

voidmatcha (changes requested):
While reviewing this PR I spotted two frontend issues that I think are worth addressing before merge. Both are small and isolated.

---

**Original** - base:master

스크린샷 2026-05-11 오전 12 35 44
스크린샷 2026-05-11 오전 12 35 58


**As-Is** - base:https://github.com/apache/zeppelin/pull/5218

스크린샷 2026-05-11 오전 12 35 50

스크린샷 2026-05-11 오전 12 36 03

**To-Be** - base:https://github.com/voidmatcha/zeppelin/commits/review-pr-5218-new

스크린샷 2026-05-11 오전 12 35 47

스크린샷 2026-05-11 오전 12 36 01

---

#### 1. Paragraph `title` is dropped on the UI

`LuceneSearch.formatResult` puts the highlighted paragraph title into the `title` field for paragraph-level hits (with Lucene `<B>...</B>` markup), but
neither the Angular nor the Classic frontend reads or renders it. So the backend populates the field and the UI silently drops it.

Suggested fix: render it next to the note name on paragraph hits, converting `<B>` → `<mark>` like `snippet` already does.

**`zeppelin-web-angular/.../result-item.component.ts`**

```ts
titleHtml = '';

// inside parseResult(), after displayName is set:
const titleRaw = hasParagraph ? (this.result.title || '') : '';
this.titleHtml = titleRaw.replace(/<B>/gi, '<mark>').replace(/<\/B>/gi, '</mark>');
```

**`result-item.component.html`** — inside the header, between the note link and the interpreter badge:

```html
<span *ngIf="titleHtml" class="paragraph-title" [innerHTML]="titleHtml"></span>
```

**`zeppelin-web/src/app/search/result-list.controller.js`** — inside the paragraph branch:

```js
let titleHtml = (note.title || '').replace(/<B>/gi, '<mark>').replace(/<\/B>/gi, '</mark>');
note.titleHtml = titleHtml;
```

**`result-list.html`** — next to the existing note name link:

```html
<span ng-if="note.titleHtml" ng-bind-html="note.titleHtml" style="color:#8c8c8c;font-size:13px;font-weight:normal"></span>
```

#### 2. Dark mode regression in the Angular result item

`zeppelin-web-angular/.../result-item.component.less` doesn't follow the `.themeMixin({...})` wrapper convention used by 28 of the 31 sibling `.less` files
  under `pages/workspace/**`. Hardcoded literals (`#f6f8fa`, `#fafbfc`, `#24292e`, `#586069`, `#e6f7e6`, `#389e0d`, `#fff7e6`, `#d48806`, `#e6f0ff`, `#1890ff`,
`#22863a`, `#e1e4e8`, `#d1d5da`, `#e8e8e8`, `#666`, `#fff3bf`) leak through under the dark theme — backgrounds stay light, low-contrast text becomes hard to
  read.

Suggested fix: port the rules into `.themeMixin({...})` and replace literals with the existing tokens used elsewhere in the workspace:

- `@background-color-light`, `@background-color-base`, `@border-color-split`
- `@text-color`, `@text-color-secondary`
- `@green-6/7`, `@gold-6/7`, `@blue-6/7` for the badge variants
- `@code-family` for the monospaced font
I've prepared local patches for both. Happy to push them as a follow-up commit on this branch if helpful.

3fbeb109a92de13c989f9cbcb2f93f7965d984a0
941683a3a7907149e953f3ff3e33f9dec7fed641

https://github.com/voidmatcha/zeppelin/commits/review-pr-5218-new/

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -75,110 +48,47 @@ export class NotebookSearchResultItemComponent implements OnChanges, OnDestroy {
...
-      }
-      match = regexp.exec(mergedStr);
-    }
-
-    mergedStr = mergedStr.replace(regexp, '$1');
-    this.mergedStr = mergedStr;
-    const keywords = [...new Set(matches)];
-    this.highlightPositions = getKeywordPositions(keywords, mergedStr);
+    const tables = this.result.tables || '';
+    this.tablesText = tables.trim().split(/\s+/).filter(t => t).join(', ');
```
voidmatcha: The `frontend/run-playwright-e2e-tests` job is failing on `prettier --check`. Only one line is at fault:


```suggestion
    this.tablesText = tables
      .trim()
      .split(/\s+/)
      .filter(t => t)
      .join(', ');
```

kkalyan:
@voidmatcha Thank you for catching both the title gap and the dark mode regression.

**Paragraph title**: `titleHtml` is now parsed from `result.title` (with `<B>` → `<mark>` conversion) and rendered as a `.title-block` div above the code block in both Angular and Classic UIs.

**Dark mode**: Rewrote `result-item.component.less` to use `themeMixin()` with design tokens (`@background-color-light`, `@text-color-secondary`, `@border-color-split`, `@green-1`/`@green-7`, `@gold-1`/`@gold-7`, `@blue-1`/`@blue-6`). All hardcoded color literals are gone.

**Prettier**: Reformatted the `tablesText` chain to multi-line — CI `prettier --check` passes.

kkalyan:
Pushed additional improvements based on hands-on testing with a production-scale S3 notebook repository (1,300+ notebooks via `S3NotebookRepo`):

**Hybrid search (keyword boost)**: Pure embedding similarity missed exact identifier matches — e.g., searching "TETRIS" wouldn't find notebooks containing `TETRIS_VIDEO_SINGLE_MEDIA` in SQL because the cosine similarity fell below the 0.25 threshold. Added a keyword boost (`KEYWORD_BOOST = 0.30f`) that promotes results where the query string appears literally in the indexed text. Semantic ranking for natural language queries is unaffected.

**Query term highlighting**: `EmbeddingSearch` now wraps query term matches in `<B>` tags (matching Lucene's convention), so the frontend highlights them identically in both backends.

**Clickable result cards**: The entire search result card is now clickable and navigates to the paragraph. Text selection is preserved — dragging to copy won't trigger navigation.

**Deep link highlighting**: Wired the `term` query parameter to `onParagraphSearch()` in `notebook.component.ts`, so clicking a search result highlights the matching terms in the Monaco editor.

**Search history**: Recent searches are persisted in `localStorage` and surfaced via a native `<datalist>` dropdown (up to 20 entries).

**InterpreterFactory NPE fix**: Added a null check in `InterpreterFactory.getInterpreter()` that throws a proper `InterpreterNotFoundException` instead of NPE when the interpreter group isn't installed. This prevents error dialogs when opening notebooks whose interpreters aren't configured locally.

kkalyan:
@jongyoul @tbonelee @voidmatcha gentle reminder to take a look at the changes.

jongyoul:
@tbonelee @voidmatcha Could you please kindly check it again after running CI?

@kkalyan Thank you for handling it. I recently almost fixed flaky tests, so it would be better to rebase your changes with the latest master if CI is still red.

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -21,136 +21,66 @@ function SearchResultCtrl($scope, $routeParams, searchService) {
...
       }
-
       note.id = note.id.replace('paragraph/', '?paragraph=') +
         '&term=' + $routeParams.searchTerm;
 
+      // Preserve Lucene <B> highlighting by converting to <mark>
+      let codeHtml = (note.snippet || '').replace(/<B>/gi, '<mark>').replace(/<\/B>/gi, '</mark>');
+      let code = (note.snippet || '').replace(/<B>/g, '').replace(/<\/B>/g, '');
+
+      let tables = (note.tables || '').trim().split(/\s+/).filter(function(t) { return t; }).join(', ');
```
tbonelee: Please modify this line like this to make it pass the lint.
```suggestion
      let tables = (note.tables || '').trim().split(/\s+/).filter(function(t) {
        return t;
      }).join(', ');
```

tbonelee: I’ll check the remaining parts tomorrow or so. Sorry for keeping you waiting.

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.less
```diff
@@ -10,10 +10,130 @@
...
+
+.result-card {
+  cursor: pointer;
+  user-select: text;
+}
+
+.result-header {
+  display: flex;
+  align-items: center;
+  gap: 8px;
```
tbonelee, line 28: We might need some gap between the card header and body.
```suggestion
  gap: 8px;
  margin-bottom: 8px;
```

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -10,175 +10,105 @@
...
 
-  setDisplayNameAndRouterLink(): void {
-    const term = this.router.snapshot.params.queryStr;
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.result) {
+      this.parseResult();
+    }
+  }
+
+  navigate(event: MouseEvent): void {
```
tbonelee: I prefer the more specific method name like `navigateToResult()`.
It aligns with the other method name (`parseResult()`).

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -10,175 +10,105 @@
...
+    const selection = window.getSelection();
+    if (selection && selection.toString().length > 0) {
+      return;
+    }
+    event.preventDefault();
```
tbonelee: This event handler is attached to `div` element, so `event` argument and `preventDefault()` calling might be unnecessary.

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -10,175 +10,105 @@
...
-            range: new Range(line, character, line, character + highlight.length),
-            options: {
-              className: 'mark',
-              stickiness: 1
-            }
-          };
-        })
-      );
-      this.cdr.markForCheck();
+  private detectInterpreter(text: string): string {
```
tbonelee, line 97: Could you please check the format of this input argument? When I logged it, `text` did not seem to include the interpreter prefix.

 Also, if the prefix is not available here, the fallback heuristic seems fragile. For example, Python code with `from ... import ...` can be detected as `sql` because `from` matches the SQL keyword check. We may make this detection logic more conservative.

Image

zeppelin-web/src/app/search/result-list.controller.js
```diff
@@ -21,136 +21,68 @@ function SearchResultCtrl($scope, $routeParams, searchService) {
   $scope.searchTerm = $routeParams.searchTerm;
   let results = searchService.search({'q': $routeParams.searchTerm}).query();
 
+  function detectLang(text) {
```
tbonelee, line 24: The same point applies to `zeppelin-web`

zeppelin-web/src/app/search/result-list.html
tbonelee, line 1: Could you remove inline style codes and add it into CSS file?

tbonelee (changes requested):

zeppelin-web-angular/src/app/pages/workspace/notebook-search/result-item/result-item.component.ts
```diff
@@ -10,175 +10,106 @@
...
+      this.queryParams = { paragraph, term };
     }
     this.displayName = this.result.name ? this.result.name : `Note ${noteId}`;
-  }
 
-  setHighlightKeyword(): void {
-    let mergedStr = this.result.header ? `${this.result.header}\n\n${this.result.snippet}` : this.result.snippet;
+    const snippet = this.result.snippet || '';
+    // Preserve Lucene <B> highlighting by converting to <mark>
+    this.codeHtml = snippet.replace(/<B>/gi, '<mark>').replace(/<\/B>/gi, '</mark>');
```
voidmatcha: The snippet/title go through only `<B>` to `<mark>` before hitting `[innerHTML]` (New UI) / `ng-bind-html` (Classic), so any `<` in paragraph source (most commonly `%sql` like `WHERE id < 100`) is parsed as DOM and disappears from the search card.

Not XSS - `DomSanitizer` / `ngSanitize` already strip the dangerous subset. Functional regression only. This is the sanitize step missing from @jongyoul's earlier suggestion for `sanitized innerHTML + <mark>`. Not asking to bring Monaco/Ace back.

#### Both UIs

> Classic: Original(master) / As-Is / To-Be

Image

Image

Image

> New UI: Original(master) / As-Is / To-Be

Image

Image

Image

Fix (cherry-pick): [voidmatcha/zeppelin@69b96612](https://github.com/voidmatcha/zeppelin/commit/69b966126)

```ts
const escapeAndMark = (s?: string): string =>
  (s || '')
    .replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;')
    .replace(/&lt;(\/?)B&gt;/gi, '<$1mark>');

this.codeHtml = escapeAndMark(this.result.snippet);
this.titleHtml = escapeAndMark(this.result.title);
```

Same shape in Classic UI's `result-list.controller.js`.

voidmatcha (changes requested):
Re-review on HEAD 5eee762c6 - earlier feedback all addressed. One frontend regression flagged inline (snippet/title raw-HTML rendering swallows `<`).

voidmatcha (approved):
LGTM on the part I reviewed.

The snippet/title rendering regression is fixed (source is now escaped before highlighting, so raw `<` in code renders correctly in both UIs). 

The Playwright CI flakiness isn't specific to this PR. I'm fixing the e2e auth/modal stability separately, so those checks should settle. (https://issues.apache.org/jira/browse/ZEPPELIN-6422)

tbonelee (approved):
Thanks for handling those feedbacks. I've reviewed focusing on the frontend parts and they looks good to me.

jongyoul (approved):

jongyoul:
Merged into master (3e470121).

jongyoul:
@kkalyan Thank you for your patience during the long review process. Your first contribution has been successfully merged! We look forward to your continued contributions to Apache Zeppelin.

kkalyan:
thank you @jongyoul. look forward to contribute more in the future. I saw new rc cut for 0.12.1 - can we try to include this CR there?

jongyoul:
@kkalyan The minor version only has bug fixes and minor improvements. Your feature is not a small one, so it will be included in a new major version. We plan to release a minor version within two weeks and a major version within the next month.

PR #5228, opened 2026-05-06
[MINOR] Drop ticket value from WebSocket debug log statements

jongyoul:
### What is this PR for?

Removes the WebSocket auth ticket value from three `LOGGER.debug` call sites in `NotebookServer.onMessage`. The ticket is a per-session UUID and adds no debugging value beyond the principal that owns it; emitting the raw value makes it visible to anyone with access to log files or downstream log collectors.

The three call sites and the change applied to each:

- **RECEIVE block** — drops the `RECEIVE TICKET` column. The remaining `op` / `principal` / `roles` / `data` columns are sufficient to identify the message.
- **"no ticket on file" branch** — logs the principal that has no entry instead of echoing back the rejected ticket.
- **"ticket mismatch" branch** — logs the principal whose ticket did not match, rather than both raw values.

`Message.toString()` does not include the ticket field, so the surrounding `LOGGER.trace("RECEIVE MSG = " + receivedMessage)` already does not leak it.

### What type of PR is it?

Improvement

### What is the Jira issue?

N/A — minor logging hygiene change, no behavioral or API change.

Diff is self-evident. The three changed sites stay on the existing branches; behavior (what is returned to the client, what is rejected) is unchanged. Existing `NotebookServerTest` continues to exercise these paths.

N/A

### Questions:

- Does the license files need to update? No

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/socket/NotebookServer.java
@@ -279,7 +279,6 @@ public class NotebookServer implements AngularObjectRegistryListener,
       if (receivedMessage.op != OP.PING) {
         LOGGER.debug("RECEIVE: " + receivedMessage.op +
             ", RECEIVE PRINCIPAL: " + receivedMessage.principal +
-            ", RECEIVE TICKET: " + receivedMessage.ticket +
             ", RECEIVE ROLES: " + receivedMessage.roles +
             ", RECEIVE DATA: " + receivedMessage.data);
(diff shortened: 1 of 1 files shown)


ParkGyeongTae (approved):
LGTM. Sensible logging hygiene fix — the ticket is a session credential and has no business sitting in debug logs. Keeping principal is enough to identify the message. No behavior change, existing tests cover the paths.

tbonelee (approved):

jongyoul:
Merged into master (e1e59bc7).
Cherry-picked into branch-0.12.

PR #5209, opened 2026-04-06
[ZEPPELIN-6409] Fix Selenium integration test flakiness after ZEPPELIN-6400

jongyoul:
## Summary

- **`authenticationUser()`**: Replace `sleep(1000)` with explicit wait for the navbar user dropdown to appear (proves login succeeded and AngularJS digest is done), then force-dismiss any lingering Bootstrap modal backdrop via jQuery
- **`logoutUser()`**: Use `clickableWait()` instead of raw `findElement()` for robust element interaction; wrap modal close button in try-catch
- **`testAngularRunParagraph()`**: Fix race condition where `waitForParagraph("FINISHED")` matched the *previous* run's state. Use `stalenessOf` to detect paragraph output refresh, then `visibilityWait` for the new element, and JavaScript click to bypass ng-click overlay issues

## Context

After ZEPPELIN-6400 moved `ZeppelinConfiguration` from `zeppelin-interpreter` to `zeppelin-zengine`, `RemoteInterpreterServer.init()` changed from loading config via `ZeppelinConfiguration.load()` + overlay to pure `Properties`-based initialization. This subtle timing change in interpreter startup exposed pre-existing Selenium test flakiness in the `test-selenium-with-spark-module-for-spark-3-5` CI job.

The CI failures were:
- `InterpreterModeActionsIT.testPerUserIsolatedAction` — `ElementClickInterceptedException` because login modal was still visible when clicking the navigation dropdown
- `ZeppelinIT.testAngularRunParagraph` — `TimeoutException` due to race condition: after re-running an Angular paragraph, the old output element was matched before the new one rendered

## Test plan

- [x] CI: `frontend.yml` — `test-selenium-with-spark-module-for-spark-3-5` job passes
- [x] CI: `frontend.yml` — All other E2E jobs remain green (Playwright auth failure is unrelated Firefox keyboard shortcut flakiness)

PR #5250, opened 2026-05-17
[ZEPPELIN-1836] Make testAngularRunParagraph wait for rendered Angular output

jongyoul:
### What is this PR for?

`testAngularRunParagraph` has been failing on every `frontend.yml` `test-selenium-with-spark-module-for-spark-3-5` run on master since the last green build at 6353224c on 2026-04-23. The failure is always at `ZeppelinIT.java:348`:

```
org.openqa.selenium.TimeoutException: Expected condition failed: waiting for visibility of element located by By.xpath: (//div[@ng-controller="ParagraphCtrl"])[1]//div[@id="angularRunParagraph"] (tried for 30 second(s) with 500 milliseconds interval)
```

Tracing the server log confirms the angular paragraph re-run completes successfully on the server side (paragraph FINISHED, result broadcast) and the `stalenessOf(oldAngularDiv)` check before `visibilityWait` does succeed — so the old `angularRunParagraph` div is detached as expected. The new div, however, is not detected as visible within 30s.

The race lives on the frontend side of the re-render: AngularJS' `renderAngular()` does `elem.html(generated)` and then `$compile(elem.contents())(paragraphScope)`. Between those two calls the result div can be momentarily detached or empty, which is exactly the window `visibilityWait` polls into. Once it gets a stale reference, the wait does not refetch.

This builds on the `stalenessOf` + JavaScript-click fix from [ZEPPELIN-6409](https://issues.apache.org/jira/browse/ZEPPELIN-6409) (#5209). That fix proved the *old* output was being torn down; this PR makes the wait for the *new* output equally tolerant of mid-`$compile` DOM churn.

### What type of PR is it?
Bug Fix

### Todos
* [x] Replace single `visibilityWait` with a content-aware polling wait that re-finds the element and ignores `StaleElementReferenceException`
* [x] Mirror the content-based pattern already used after the first run (`waitForText(\"Run second paragraph\", ...)`)

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-1836 (long-standing flaky test ticket, open since 2016)

* CI `frontend.yml` → `test-selenium-with-spark-module-for-spark-3-5` should run `ZeppelinIT.testAngularRunParagraph` to completion without timing out at line 348.
* Local repro (slow, requires `-Pweb-classic` and Spark 3.5):
  ```
  ./mvnw verify -DfailIfNoTests=false -pl zeppelin-integration \
    -Pweb-classic -Pintegration -Pspark-scala-2.12 -Pspark-3.5 -Pweb-dist -Pusing-source-tree \
    -Dit.test=ZeppelinIT#testAngularRunParagraph
  ```

N/A

### Questions:

### DIFF
--- a/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
+++ b/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
@@ -30,6 +30,7 @@ import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.OutputType;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.TakesScreenshot;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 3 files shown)


jongyoul:
@voidmatcha @tbonelee Does it make sense?

tbonelee (approved):
Diagnosis and fix both look right. There's probably a cleaner path, but for test code against the legacy classic-web UI, landing it at "passes reliably" is fine by me.

Minor: the `fast-uri` lock bump looks accidental, worth double-checking.

voidmatcha (approved):
Approve 👍 

The `testAngularRunParagraph` fix is the real win. Re-enabling the test and waiting on the actual rendered output properly closes the render race.

Scope note: this also hardens the shared `authenticationUser()`, but the target test runs in `ZeppelinIT` (auth off, no login). That method is only used by the shiro-enabled `AuthenticationIT` / `PersonalizeActionsIT` / `InterpreterModeActionsIT`, whose login flakiness #5335 fixes at the root via REST, dropping this method from the call path. Both are good; cleaner to let #5335 own login. After merge, I'll leave a note on #5335 to track the overlap.

tbonelee:
Merged to master

PR #5257, opened 2026-05-24
[MINOR] chore(zeppelin-web-angular): remove @antv/g2plot and migrate to chart.js to fix npm audit

jongyoul:
### What is this PR for?
This PR completely removes the `@antv/g2plot` dependency and migrates the charting visualization to `chart.js` in order to resolve the `npm-audit` failures caused by the critical malware advisories for `@antv/color-util` and `@antv/adjust` (GHSA-rh6v-hwr4-6jcp / GHSA-qcp2-qp9h-qprg).

### What changes are proposed?
1. **Remove `@antv/g2plot`**: Uninstalled `@antv/g2plot` and its transitive dependencies to clean up the security alerts.
2. **Install `chart.js`**: Installed `chart.js` (and upgraded `webpack-dev-server` to `^5.2.4` to remediate other moderate vulnerabilities).
3. **Migrate Visualization**: Rewrote `TableVisualization.tsx` using `chart.js` to draw the 5 types of charts (Bar, Line, Pie, Scatter, Area) on a HTML5 Canvas context. The visualization layout and color scheme (`#1890ff`) are preserved.

Verify that the `npm-audit` job passes in GitHub Actions.

### DIFF
--- a/zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
+++ b/zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
@@ -12,10 +12,10 @@
 
 import { useState, useEffect, useMemo, useRef } from 'react';
 import { Table } from 'antd';
-import type { Column, Line, Pie, Scatter } from '@antv/g2plot';
 import { VisualizationControls } from './VisualizationControls';
 import { parseTableData, exportFile } from '@/utils';
 import type { ParagraphConfigResult, ParagraphIResultsMsgItem, VisualizationMode } from '@zeppelin/sdk';
+import type { Chart, ChartConfiguration } from 'chart.js';
 
 interface TableVisualizationProps {
   result: ParagraphIResultsMsgItem;
@@ -66,7 +66,8 @@ export const TableVisualization = ({ result, config }: TableVisualizationProps) 
   };
 
   useEffect(() => {
-    if (!chartRef.current || !tableData || tableData.rows.length === 0 || currentMode === 'table') return;
+    const container = chartRef.current;
+    if (!container || !tableData || tableData.rows.length === 0 || currentMode === 'table') return;
 
     const data = tableData.rows.map((row, idx) => ({
       category: row[0] || `Row ${idx + 1}`,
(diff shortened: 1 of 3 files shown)


jongyoul:
@tbonelee @ParkGyeongTae @voidmatcha I heard that @antv had a security issue, so it's been fully deprecated. So npm-audit failed, so I removed them and used the vanilla one. Could you please check it?

zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
```diff
@@ -75,72 +74,132 @@ export const TableVisualization = ({ result, config }: TableVisualizationProps)
...
-            point: {
-              size: 3,
-              shape: 'circle'
+          chartConfig = {
+            type: 'line',
+            data: {
+              labels: data.map(d => d.category),
+              datasets: [{
+                label: 'Value',
+                data: data.map(d => data.value),
```
voidmatcha:
```suggestion
                data: data.map(d => d.value),
```

> Original(master)
Image

> As-Is
Image

> To-Be
Image

The callback references the outer data array instead of the element d, so data.value is undefined for every row and the Area Chart renders empty.

voidmatcha (commented):
- npm audit on master: 21 advisories (10 moderate / 5 high / 6 critical, incl. the @antv/adjust & @antv/color-util malware advisories)
- Diff review: no new dangerouslySetInnerHTML / eval / dynamic Function introduced, and chart labels/data are drawn on a 2D canvas context. So the migration doesn't add any XSS surface.

LGTM from a supply-chain perspective.

zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
```diff
@@ -75,72 +74,132 @@ export const TableVisualization = ({ result, config }: TableVisualizationProps)
       y: parseFloat(row[1] || '0') || 0
     }));
 
-    let chart: Column | Line | Pie | Scatter | null = null;
+    chartRef.current.innerHTML = '';
+
+    let chart: any = null;
```
tbonelee: **issue: `let chart: any` and `let chartConfig: any` violate `@typescript-eslint/no-explicit-any` (set to `error` in this project)**

`npm run lint` fails on both lines. Beyond the lint rule, the migration also loses the previous `Column | Line | Pie | Scatter` discrimination. Chart.js 4 ships full generics (`ChartConfiguration<TType>`, `Chart<TType>`), so we can satisfy the rule and tighten the original type safety in one go.

Concretely: drop the `chartConfig: any` intermediate variable and call `new Chart<TType>(ctx, {...})` inline per case. The literal `type: 'bar'` narrows `data.datasets[N]` and `options` to the bar-specific shape, so typos like `data.map(d => data.value)` would be caught at compile time.

```ts
import type { Chart as ChartJS } from 'chart.js';

// ...

let chart: ChartJS | null = null;

import('chart.js/auto').then(module => {
  // ...
  switch (currentMode) {
    case 'multiBarChart':
      chart = new Chart<'bar'>(ctx, {
        type: 'bar',
        data: {
          labels: data.map(d => d.category),
          datasets: [{ label: 'Value', data: data.map(d => d.value), backgroundColor: '#1890ff' }]
        },
        options: { responsive: true, maintainAspectRatio: false }
      });
      break;
    case 'lineChart':
      chart = new Chart<'line'>(ctx, { /* ... */ });
      break;
    // pieChart -> Chart<'pie'>, scatterChart -> Chart<'scatter'>,
    // stackedAreaChart -> Chart<'line'>
  }
});
```

Same pattern for the remaining four cases. Tested locally: no runtime change, both `any` lint warnings go away, and the per-case dataset shape is now type-checked.

zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
```diff
@@ -75,72 +74,132 @@ export const TableVisualization = ({ result, config }: TableVisualizationProps)
...
     });
 
     return () => {
       cancelled = true;
       if (chart) {
         chart.destroy();
       }
+      if (chartRef.current) {
+        chartRef.current.innerHTML = '';
+      }
```
tbonelee: **issue: cleanup references `chartRef.current`, which violates `react-hooks/exhaustive-deps` (set to `error` in this project)**

`npm run lint` fails on the cleanup line. In this specific component the ref is unlikely to actually diverge between effect runs (the chart `<div>` is identity-stable across chart-mode switches, and on switching to the `table` mode the ref simply goes to `null`, which the existing `if (chartRef.current)` guard handles), so this is not a runtime bug. But the project has the rule at `error`, so it still needs to be fixed.

The canonical fix is to capture the ref to a local `const` at effect start and use it everywhere (initial guard, the dynamic `import().then()` callback, and the cleanup return). This satisfies the rule without an `eslint-disable` and makes the lifecycle binding explicit.

```ts
useEffect(() => {
  const container = chartRef.current;
  if (!container || !tableData || tableData.rows.length === 0 || currentMode === 'table') return;

  container.innerHTML = '';

  let chart: ChartJS | null = null;
  let cancelled = false;

  import('chart.js/auto').then(module => {
    if (cancelled) return;
    // ... use `container.appendChild(canvas)` instead of `chartRef.current.appendChild(...)`
  });

  return () => {
    cancelled = true;
    if (chart) chart.destroy();
    container.innerHTML = '';
  };
}, [currentMode, tableData]);
```

tbonelee (changes requested):

zeppelin-web-angular/projects/zeppelin-react/src/components/visualizations/TableVisualization.tsx
jongyoul: Thanks for the review! I've captured the `chartRef.current` reference into a local `container` variable at the beginning of the effect and updated the effect and cleanup function to use it. This resolves the `react-hooks/exhaustive-deps` warning.

tbonelee:
@jongyoul Could you rebase this onto master?

voidmatcha (approved):
LGTM

tbonelee (approved):

tbonelee:
**praise:** Lint is clean now. `Chart | null` / `ChartConfiguration | null` resolves both `no-explicit-any` errors, and the union still narrows per-case at the `{ type: 'bar', ... }` literal assignment, so each case's dataset/options shape stays type-checked. Verified locally with the project ruleset: errors 4 → 0 on this file.

**thought (non-blocking):** If you ever want tighter per-case typing, `Chart<'bar'>` / `ChartConfiguration<'bar'>` etc. drop the union entirely and narrow without relying on the assignment site. Current shape is good; completely optional follow-up.

**nitpick (non-blocking):** Now that `container` is captured at effect start, the `!container` check inside `import('chart.js/auto').then(...)` and the `if (container)` guard in the cleanup are unreachable. `container` was already non-null past the early return, and closure capture pins it for the effect's lifetime. Safe to simplify to `if (cancelled) return;` and `container.innerHTML = '';`. Pure tidy-up, no behavior change.

PR #5259, opened 2026-05-30
[ZEPPELIN-6423] Reload note when switching notebooks in the Angular UI

tbonelee:
### What is this PR for?
Navigating between notes in the Angular UI changed the URL but left the page showing the previously opened note.

The note fetch in `NotebookComponent` was bound only to the WebSocket `connectedStatus$` stream (introduced in ZEPPELIN-6387). Because Angular reuses `NotebookComponent` across `:noteId` route changes — so `ngOnInit` does not re-run — and the WebSocket stays connected, selecting another note from the header notebook list never re-fetched the note. The URL/route params updated, but `getNote()` was never called for the new `noteId`, so the page kept rendering the old note.

This PR drives the fetch from a `combineLatest` of the connection status and the route params, so it fires on **both** a WebSocket (re)connect **and** a `noteId`/`revisionId` change. The reconnect-reload behavior from ZEPPELIN-6387 is preserved; `distinctUntilChanged` on the connection stream avoids a redundant fetch on init.

### What type of PR is it?
Bug Fix

### Todos
* [x] Re-fetch the note on `noteId`/`revisionId` route changes
* [x] Preserve WebSocket reconnect-reload behavior
* [x] Add e2e regression test

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6423

* **Automated:** `zeppelin-web-angular/e2e/tests/notebook/main/notebook-navigation.spec.ts` opens one note, then navigates to a second note via the header "Notebook" dropdown and asserts the displayed note title (not just the URL) updates. Verified failing before the fix and passing after, against a live backend.
* **Manual:**
  1. Open a notebook.
  2. From the header **Notebook** dropdown, click a different note.
  3. The page content should switch to the newly selected note (previously it stayed on the old note while the URL changed).

N/A

### Questions:

### DIFF
--- a/zeppelin-web-angular/e2e/tests/notebook/main/notebook-navigation.spec.ts
+++ b/zeppelin-web-angular/e2e/tests/notebook/main/notebook-navigation.spec.ts
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


jongyoul:
@tbonelee Could you please rebase it?

jongyoul (approved):

tbonelee:

PR #3496, opened 2019-10-31
[ZEPPELIN-4407] Add the ability to copy/paste result table

amakaur:
### What is this PR for?
This PR provides the ability to copy/paste the entire result table from a paragraph.

Currently, there's an option to download the result table into a csv or tsv file, but it creates a new file for each result table per paragraph. In order to copy multiple result tables from same and different notebooks we've to download each file, copy the dataset, and paste it into one external app. 

With this new approach there's no need to download multiple files. Simply click on either CopyAll (copies data in TSV format) or pick the options from drop down menu to download or copy table as TSV/CSV. This makes it easier to copy dataset of multiple tables into one external app without creating and opening multiple files.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4407

### How should this be tested?
 - Create a dataframe 
- use z.show(df)
- Click on CopyAll
- Paste into external app

### Screenshots (if appropriate)
#
image
## Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #5261, opened 2026-06-02
[ZEPPELIN-4407] Add copy to clipboard (TSV/CSV) for table results

kkalyan:
## What is this PR for?

I've seen users downloading CSV and opening in spreadsheet viewer and copying the text. 
Adding way to copy CSV/TSV directly. This is orginally implemented by @amakaur #3496 but it was closed due to lack of tests.  I'm picking it up now. 

### Changes

**Angular UI (`zeppelin-web-angular`)**
- `result.component` — paragraph toolbar dropdown: renamed existing items to "Download as CSV/TSV", added divider, then "Copy as TSV" and "Copy as CSV"
- `table-visualization.component` — inner table Export menu: added "Copy all data as TSV/CSV" and "Copy visible data as TSV/CSV" (mirrors the existing "Export visible" scope)

**Classic AngularJS UI (`zeppelin-web`)**
- `result-chart-selector.html` — same dropdown restructure: Download / divider / Copy
- `result.controller.js` — new `$scope.copyToClipboard(delimiter)` function

### Behaviour
- Header row (column names) is always included in the copied text
- Cell values containing the delimiter, double-quotes, or newlines are RFC 4180 quoted
- Uses `navigator.clipboard.writeText` with a `document.execCommand('copy')` fallback for older browsers

## What type of PR is it?
Feature

## What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-4407

## How should this be tested?

1. Run a paragraph that outputs a TABLE (e.g. `%sh printf "col1\tcol2\na\t1\nb\t2\n"`)
2. Click the **▾** next to the download button in the paragraph toolbar
3. Verify the menu shows: **Download as CSV**, **Download as TSV**, *(divider)*, **Copy as TSV**, **Copy as CSV**
4. Click **Copy as TSV** → paste into a spreadsheet app or text editor — expect headers + rows, tab-delimited
5. Click **Copy as CSV** → paste → expect headers + rows, comma-delimited
6. Test with a cell value containing a comma, e.g. `"hello, world"` → the CSV copy should quote it correctly

## Tests

- **Classic UI (Karma/Jasmine):** `zeppelin-web/src/app/notebook/paragraph/result/result.controller.test.js` — 4 new specs covering TSV copy, CSV copy, delimiter quoting, and double-quote escaping
- **Angular UI (Playwright E2E):** `zeppelin-web-angular/e2e/tests/notebook/paragraph/copy-to-clipboard.spec.ts` — 3 new specs (skipped on CI, require a live interpreter)

## Questions
- Does the license file need update? No
- Is there a breaking change for older versions? No — existing Download as CSV/TSV behaviour is unchanged
- Does this need documentation? No

## screenshots
image

image

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
@@ -286,7 +286,8 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit, 
       ...rows.map(row => columns.map(col => escape(row[col])).join(delimiter))
     ];
     const text = lines.join('\n');
-    navigator.clipboard.writeText(text).catch(() => {
+    // TODO: Refactor the duplicated copy-to-clipboard logics
+    const fallbackCopy = () => {
       const el = document.createElement('textarea');
       el.value = text;
       el.style.position = 'absolute';
@@ -295,7 +296,14 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit, 
       el.select();
       document.execCommand('copy');
       document.body.removeChild(el);
-    });
+    };
+    // navigator.clipboard is undefined in non-secure contexts (e.g. plain HTTP),
+    // where writeText would throw synchronously before the catch could run.
+    if (navigator.clipboard) {
+      navigator.clipboard.writeText(text).catch(fallbackCopy);
+    } else {
+      fallbackCopy();
+    }
   }
 
   switchMode(mode: VisualizationMode) {
--- a/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
@@ -271,6 +271,33 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit, 
     }
   }
 
+  copyToClipboard(type: 'tsv' | 'csv'): void {
+    if (!this.tableData || !this.tableData.rows) {
+      return;
+    }
+    const delimiter = type === 'tsv' ? '\t' : ',';
+    const { columns, rows } = this.tableData;
+    const escape = (value: unknown): string => {
+      const str = String(value ?? '');
+      return str.includes(delimiter) || str.includes('"') || str.includes('\n') ? `"${str.replace(/"/g, '""')}"` : str;
+    };
+    const lines = [
+      columns.map(escape).join(delimiter),
+      ...rows.map(row => columns.map(col => escape(row[col])).join(delimiter))
+    ];
+    const text = lines.join('\n');
+    navigator.clipboard.writeText(text).catch(() => {
+      const el = document.createElement('textarea');
+      el.value = text;
+      el.style.position = 'absolute';
+      el.style.left = '-9999px';
+      document.body.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      document.body.removeChild(el);
+    });
+  }
+
   switchMode(mode: VisualizationMode) {
     if (!this.config) {
       throw new Error('config is not defined');
--- a/zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
+++ b/zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
@@ -75,6 +75,30 @@ export class TableVisualizationComponent implements OnInit {
     writeFile(wb, `export.${type}`);
   }
 
+  copyToClipboard(type: 'tsv' | 'csv', all = true) {
+    const delimiter = type === 'tsv' ? '\t' : ',';
+    const sourceRows = all ? this.rows : [...this.nzTable.data];
+    const escape = (value: unknown): string => {
+      const str = String(value ?? '');
+      return str.includes(delimiter) || str.includes('"') || str.includes('\n') ? `"${str.replace(/"/g, '""')}"` : str;
+    };
+    const lines = [
+      this.columns.map(escape).join(delimiter),
+      ...sourceRows.map(row => this.columns.map(col => escape(row[col])).join(delimiter))
+    ];
+    const text = lines.join('\n');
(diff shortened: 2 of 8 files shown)


zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
```diff
@@ -271,6 +271,33 @@ export class NotebookParagraphResultComponent implements OnInit, AfterViewInit,
...
+    navigator.clipboard.writeText(text).catch(() => {
+      const el = document.createElement('textarea');
+      el.value = text;
+      el.style.position = 'absolute';
+      el.style.left = '-9999px';
+      document.body.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      document.body.removeChild(el);
+    });
```
tbonelee:
```suggestion
    // TODO: Refactor the duplicated copy-to-clipboard logics
    const fallbackCopy = () => {
      const el = document.createElement('textarea');
      el.value = text;
      el.style.position = 'absolute';
      el.style.left = '-9999px';
      document.body.appendChild(el);
      el.select();
      document.execCommand('copy');
      document.body.removeChild(el);
    };
    // navigator.clipboard is undefined in non-secure contexts (e.g. plain HTTP),
    // where writeText would throw synchronously before the catch could run.
    if (navigator.clipboard) {
      navigator.clipboard.writeText(text).catch(fallbackCopy);
    } else {
      fallbackCopy();
    }
```

zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
```diff
@@ -75,6 +75,30 @@ export class TableVisualizationComponent implements OnInit {
...
+    navigator.clipboard.writeText(text).catch(() => {
+      const el = document.createElement('textarea');
+      el.value = text;
+      el.style.position = 'absolute';
+      el.style.left = '-9999px';
+      document.body.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      document.body.removeChild(el);
+    });
```
tbonelee:
```suggestion
    // TODO: Refactor the duplicated copy-to-clipboard logics
    const fallbackCopy = () => {
      const el = document.createElement('textarea');
      el.value = text;
      el.style.position = 'absolute';
      el.style.left = '-9999px';
      document.body.appendChild(el);
      el.select();
      document.execCommand('copy');
      document.body.removeChild(el);
    };
    // navigator.clipboard is undefined in non-secure contexts (e.g. plain HTTP),
    // where writeText would throw synchronously before the catch could run.
    if (navigator.clipboard) {
      navigator.clipboard.writeText(text).catch(fallbackCopy);
    } else {
      fallbackCopy();
    }
```

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
```diff
@@ -914,6 +914,49 @@ function ResultCtrl($scope, $rootScope, $route, $window, $routeParams, $location
     saveAsService.saveAs(dsv, exportedFileName, extension);
   };
 
+  $scope.copyToClipboard = function(delimiter) {
+    let text = '';
+    for (let titleIndex in tableData.columns) {
+      if (tableData.columns.hasOwnProperty(titleIndex)) {
+        text += tableData.columns[titleIndex].name + delimiter;
```
tbonelee: We should escape not only the row value but the header value also. You could extract the `escape` function inside the closure and use that for this.

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
```diff
@@ -914,6 +914,49 @@ function ResultCtrl($scope, $rootScope, $route, $window, $routeParams, $location
...
+      }
+    }
+    text = text.substring(0, text.length - 1) + '\n';
+    for (let r in tableData.rows) {
+      if (tableData.rows.hasOwnProperty(r)) {
+        let row = tableData.rows[r];
+        let dsvRow = '';
+        for (let index in row) {
+          if (row.hasOwnProperty(index)) {
+            let stringValue = (row[index]).toString();
```
tbonelee: Null check is needed at here.
```suggestion
            let stringValue = (value === null || value === undefined) ? '' : String(value);
```

tbonelee (changes requested):
Thanks for the suggestion.
Overall logics looks good to me, and I only left a few feedbacks.
You could also add a test cases which handles double quote delimiter in both `zeppelin-web` and `zeppelin-web-angular`

zeppelin-web-angular/src/app/pages/workspace/share/result/result.component.ts
kkalyan: Done — extracted `fallbackCopy` as a named function and added the `if (navigator.clipboard)` guard before calling `writeText`, so the fallback triggers correctly in non-HTTPS contexts too. Applied the same pattern to both `result.component.ts` and `table-visualization.component.ts`.

zeppelin-web-angular/src/app/visualizations/table/table-visualization.component.ts
kkalyan: Done — same change applied here in `table-visualization.component.ts`: extracted `fallbackCopy` and added the `navigator.clipboard` existence guard.

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
kkalyan: Fixed — extracted an `escape()` closure that is now shared by both the header row and cell values, so header names containing the delimiter, double quotes, or newlines are properly RFC 4180 quoted.

zeppelin-web/src/app/notebook/paragraph/result/result.controller.js
kkalyan: Fixed — the `escape()` function now does `(value === null || value === undefined) ? '' : String(value)` before checking for special characters.

tbonelee (approved):

PR #5262, opened 2026-06-02
[ZEPPELIN-6422] Stabilize flaky Playwright E2E tests

voidmatcha:
### What is this PR for?
Three races caused flaky `frontend / run-playwright-e2e-tests`:

1. Per-test login raced on the shared session cookie under parallel workers -> moved to a single `setup` project + `storageState`.
2. `locator.fill` on Ant modal inputs landed before Angular bound the form-control -> new `BasePage.fillAndVerifyInput()` retries via `expect.toPass` until the input value sticks.
3. Modal/dropdown/theme/logout transitions had no explicit wait -> targeted waits added at each boundary.

Inline comments on the diff for the non-obvious bits.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
ZEPPELIN-6422

### Questions:

### DIFF
--- a/zeppelin-web-angular/e2e/global.setup.ts
+++ b/zeppelin-web-angular/e2e/global.setup.ts
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import { test as setup, expect } from '@playwright/test';
+import { LoginTestUtil } from './models/login-page.util';
+import { performLoginIfRequired, waitForZeppelinReady } from './utils';
+
+// Resolved against the Playwright project rootDir (zeppelin-web-angular/).
+// Must match the `storageState` value declared for browser projects in playwright.config.js.
+export const STORAGE_STATE = path.join('playwright', '.auth', 'user.json');
+
+setup('authenticate', async ({ page }) => {
+  fs.mkdirSync(path.dirname(STORAGE_STATE), { recursive: true });
+
+  const isShiroEnabled = await LoginTestUtil.isShiroEnabled();
+  if (!isShiroEnabled) {
+    // Auth variant disabled — write an empty storage state so dependent projects load,
+    // then exit. This keeps the setup-project pattern uniform across CI matrix variants.
+    await page.context().storageState({ path: STORAGE_STATE });
+    return;
+  }
+
+  await page.goto('/');
+  await waitForZeppelinReady(page);
+
+  await performLoginIfRequired(page);
+
+  // Verify we are authenticated. Don't rely on performLoginIfRequired's return value —
+  // it returns false both for "no work to do" and "login attempt failed".
+  await expect(page.locator('zeppelin-login')).toBeHidden({ timeout: 30000 });
+  await expect(page.getByRole('heading', { name: 'Welcome to Zeppelin!' })).toBeVisible({ timeout: 30000 });
+
+  await page.context().storageState({ path: STORAGE_STATE });
+});
--- a/zeppelin-web-angular/e2e/models/base-page.ts
+++ b/zeppelin-web-angular/e2e/models/base-page.ts
@@ -104,10 +104,22 @@ export class BasePage {
     await expect(locator).toBeVisible({ timeout });
     await expect(locator).toBeEnabled({ timeout: 5000 });
 
-    // Click first so Angular's form control is focused and its initial setValue cycle
-    // has completed before we overwrite it.  Then fill() atomically sets the value.
-    await locator.click();
-    await locator.fill(value);
-    await expect(locator).toHaveValue(value, { timeout: 10000 });
+    // Ant-modal autofocus + Angular form initialization race: any of fill / type /
+    // pressSequentially can land BEFORE the form-control's initial value sync,
+    // after which Angular silently resets the input back to the model's initial
+    // value (placeholder). ng-dirty is set but the visible value is wrong.
+    await expect(async () => {
+      await locator.click();
+      await locator.fill(value);
+      await locator.evaluate((el: HTMLInputElement) => {
+        el.dispatchEvent(new Event('input', { bubbles: true }));
+        el.dispatchEvent(new Event('change', { bubbles: true }));
+      });
+      // Verify the value stuck — if Angular reset it, this throws and toPass retries.
+      const actual = await locator.inputValue();
+      if (actual !== value) {
+        throw new Error(`fillAndVerifyInput retry: expected "${value}" got "${actual}"`);
+      }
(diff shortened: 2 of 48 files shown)


zeppelin-web-angular/e2e/models/folder-rename-page.ts
```diff
@@ -77,12 +67,11 @@ export class FolderRenamePage extends BasePage {
   }
 
   async enterNewName(name: string): Promise<void> {
-    await this.renameInput.fill(name);
+    await this.fillAndVerifyInput(this.renameInput, name);
```
voidmatcha, line 70: pressSequentially reproduced the same race. Angular's FormControl initializes after any key/value input, so every input method hits the same reset window. Routed this and the other modal-input specs through BasePage.fillAndVerifyInput instead.

zeppelin-web-angular/e2e/models/node-list-page.ts
```diff
@@ -41,15 +41,12 @@ export class NodeListPage extends BasePage {
     await this.createNewNoteButton.click();
   }
 
-  private getNoteByName(noteName: string): Locator {
-    return this.page.locator('nz-tree-node').filter({ hasText: noteName }).first();
+  noteLinkByName(noteName: string): Locator {
+    return this.nodeListContainer.getByRole('link', { name: noteName, exact: true });
```
voidmatcha, line 45: Master's substring hasText + .first() non-deterministically picked one of several TestNotebook_* nodes coexisting in the shared workspace. getByRole('link', { name, exact: true }) is exact-match and the Playwright-recommended locator.

zeppelin-web-angular/e2e/global.setup.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  await performLoginIfRequired(page);
+
+  // Verify we are authenticated. Don't rely on performLoginIfRequired's return value —
+  // it returns false both for "no work to do" and "login attempt failed".
+  await expect(page.locator('zeppelin-login')).toBeHidden({ timeout: 30000 });
+  await expect(page.getByRole('heading', { name: 'Welcome to Zeppelin!' })).toBeVisible({ timeout: 30000 });
+
+  await page.context().storageState({ path: STORAGE_STATE });
+});
```
voidmatcha, line 45: This setup project is the core change in the PR. The 29 specs that previously called login in their own beforeEach raced on the shared session cookie under parallel workers. Consolidating to a single login + storageState removes that race.

tbonelee:
Could you rebase onto the master branch first? There have been many dependency updates, so I think the rebase needs to happen before anything else.

zeppelin-web-angular/e2e/utils.ts
```diff
@@ -327,121 +404,125 @@ export const navigateToNotebookWithFallback = async (
...
-  await dropdownTrigger.click();
+  const paragraphId = noteJson.body?.paragraphs?.[0]?.id;
+  if (!paragraphId || !paragraphId.startsWith('paragraph_')) {
+    throw new Error(`Create notebook REST response did not include paragraph id: ${JSON.stringify(noteJson.body)}`);
+  }
 
-  const paragraphLink = page.locator('li.paragraph-id a').first();
-  await paragraphLink.waitFor({ state: 'attached', timeout: 15000 });
+  return { noteId, paragraphId };
+};
```
voidmatcha, line 481: UI create-note is already covered by note-create-modal.spec.ts. Reusing it for setup would re-run the same Ant-modal + ngModel race in every spec's beforeEach. REST is deterministic and keeps that coverage in one spec.

---

Rebased onto the latest master.

voidmatcha:
c8b6026 from #5263

Note: #5263 currently fails CI on the same root cause that's fixed here. - c80421e
This PR going green covers it. If #5263 lands first, I'll rebase afterwards.

---

c80421e from #5259 

**As-Is**

01-default-viewport-dropdown

**To-Be**
02-widened-viewport-dropdown


CI fail was `Element is outside of the viewport`. The notebook dropdown overflows the default 1280×720 page.  Widening the viewport keeps the target within Playwright's actionable area, so the routerLink click lands. 

The dropdown likely just needs a bounded max-height with internal overflow scrolling. Happy to bundle that change here (removing the viewport workaround) or open a separate follow-up issue & PR. whichever  reviewers prefers.

tbonelee:
Although this PR is mainly for fixing the flaky test, I think it's okay to include the Notebook dropdown fix here if it's needed. If the more proper long-term fix would make this PR too large, we can go with a smaller fix in this PR and leave a TODO comment for the follow-up refactor.

voidmatcha:
c8baf60

https://github.com/user-attachments/assets/34c30b49-7753-40ff-b2d1-bf1277902d9e



Thanks for the flexibility! I've folded the dropdown fix into this PR: a max-height + internal scroll on the overlay scoped via nzOverlayClassName, with the E2E viewport workaround removed from the spec in the same commit.

zeppelin-web-angular/e2e/tests/home/home-page-notebook-actions.spec.ts
```diff
@@ -54,7 +53,20 @@ test.describe('Home Page Notebook Actions', () => {
...
+          await input.fill(specialInput);
+          await input.evaluate((el: HTMLInputElement) => {
+            el.dispatchEvent(new Event('input', { bubbles: true }));
+            el.dispatchEvent(new Event('change', { bubbles: true }));
+          });
+          const actual = await input.inputValue();
+          if (actual !== specialInput) {
+            throw new Error(`filterInput retry: got "${actual}"`);
+          }
+        }).toPass({ timeout: 15000, intervals: [200, 500, 1000, 2000] });
```
tbonelee: This logic seems to overlap with the earlier code. How about extracting it into a shared/common function? I thought it served the same purpose in terms of intent as well. What do you think?

zeppelin-web-angular/e2e/tests/notebook/action-bar/action-bar-functionality.spec.ts
```diff
@@ -46,7 +47,19 @@ test.describe('Notebook Action Bar Functionality', () => {
...
+      await titleInputField.fill(notebookName);
+      await titleInputField.evaluate((el: HTMLInputElement) => {
+        el.dispatchEvent(new Event('input', { bubbles: true }));
+        el.dispatchEvent(new Event('change', { bubbles: true }));
+      });
+      const actual = await titleInputField.inputValue();
+      if (actual !== notebookName) {
+        throw new Error(`titleInput retry: got "${actual}"`);
+      }
+    }).toPass({ timeout: 15000, intervals: [200, 500, 1000, 2000] });
```
tbonelee: Same as the comment above.

zeppelin-web-angular/e2e/models/base-page.ts
```diff
@@ -104,10 +104,22 @@ export class BasePage {
...
+      await locator.evaluate((el: HTMLInputElement) => {
+        el.dispatchEvent(new Event('input', { bubbles: true }));
+        el.dispatchEvent(new Event('change', { bubbles: true }));
+      });
+      // Verify the value stuck — if Angular reset it, this throws and toPass retries.
+      const actual = await locator.inputValue();
+      if (actual !== value) {
+        throw new Error(`fillAndVerifyInput retry: expected "${value}" got "${actual}"`);
+      }
+    }).toPass({ timeout: 15000, intervals: [200, 500, 1000, 2000] });
```
tbonelee, line 123: Same as the below comment.

tbonelee (changes requested):
LGTM. Just suggesting some minor refactoring.

tbonelee (approved):

tbonelee:
Merged into master

PR #5263, opened 2026-06-03
[ZEPPELIN-6425] Migrate zeppelin-react ESLint config to v9 flat format

voidmatcha:
### What is this PR for?

This regression surfaced after ZEPPELIN-6421 (zeppelin-react lint integration) and ZEPPELIN-6424 (Angular 13 to 21 upgrade) merged in close succession. Each PR passed CI in isolation, but together they broke `npm run lint:react`:

- ZEPPELIN-6421 added `projects/zeppelin-react/package.json` scripts that use the ESLint 8 `--ext .ts,.tsx` flag and rely on `projects/zeppelin-react/.eslintrc.json`.
- ZEPPELIN-6424 bumped the root install to ESLint 9, which removed `--ext` and dropped `.eslintrc.*` support entirely in favor of flat config (`eslint.config.{js,mjs,cjs}`).

The two changes never collided in their own PR CIs because each was tested against a master that didn't yet contain the other. Once both landed, master's `frontend / run-playwright-e2e-tests (auth, 3.9)` and `(anonymous, 3.9)` started failing at the `npm lint` step before Playwright ran.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
ZEPPELIN-6425

### Questions:


tbonelee (changes requested):
LGTM. We only need to correct the reference for the `.eslintrc.json` in `zeppelin-react/README.md`

tbonelee (approved):

tbonelee:
Merged into master

PR #5264, opened 2026-06-03
[ZEPPELIN-6426] Remove redundant eslint-disable comments in zeppelin-web-angular

tbonelee:
### What is this PR for?
zeppelin-web-angular has several eslint-disable comments that no longer suppress any violation under the current ESLint configuration (e.g. no-invalid-this, id-blacklist, no-eval, jsdoc/no-types, @typescript-eslint/naming-convention, @angular-eslint/component-class-suffix). These unused directives are misleading and add noise, so this PR removes them. No functional change.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6426

* Run `npm run lint` in zeppelin-web-angular and confirm it still passes.

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/core/message-listener/message-listener.ts
+++ b/zeppelin-web-angular/src/app/core/message-listener/message-listener.ts
@@ -19,7 +19,6 @@ import { Message, MessageReceiveDataTypeMap, ReceiveArgumentsType } from '@zeppe
   template: '',
   standalone: false
 })
-// eslint-disable-next-line @angular-eslint/component-class-suffix
 export class MessageListenersManager implements OnDestroy {
   __zeppelinMessageListeners__?: Array<() => void>;
   __zeppelinMessageListeners$__: Subscriber<unknown> | null = new Subscriber();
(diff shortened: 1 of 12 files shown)


zeppelin-web-angular/src/app/core/message-listener/message-listener.ts
```diff
@@ -43,18 +42,15 @@ export function MessageListener<K extends keyof MessageReceiveDataTypeMap>(op: K
...
         throw new Error('__zeppelinMessageListeners$__ is not defined');
       }
-      // eslint-disable-next-line no-invalid-this
+
       this.__zeppelinMessageListeners$__.add(
-        // eslint-disable-next-line no-invalid-this
         this.messageService.receive(op).subscribe(data => {
           // @ts-ignore
-          // eslint-disable-next-line no-invalid-this
+
```
voidmatcha:
```suggestion
```
Suggestion: blank lines left behind where comments were deleted

voidmatcha (approved):
Suggestion: add `linterOptions: { reportUnusedDisableDirectives: 'error' }` to eslint.config.js so this stays enforced by default it's only a warning and `ng lint` never fails on warnings.

LGTM

tbonelee:
Thanks @voidmatcha! Added it in 746751d2f, and applied the same to the `zeppelin-react` config (baf33cd64) since it has the identical gap. Lint passes on both with no unused directives.

voidmatcha (approved):

pan3793 (approved):

tbonelee:
Thanks. Merged into master.

PR #5265, opened 2026-06-03
[ZEPPELIN-6427] Convert interpreter setting form to typed reactive forms

tbonelee:
### What is this PR for?

Numeric interpreter properties edited in the Angular UI are sent as JSON numbers, which Gson deserializes as `Double` (`60000` → `"60000.0"`), breaking `Long`/`Integer` parsing in interpreters such as JDBC (ZEPPELIN-6395). The server-side workaround from ZEPPELIN-6131 only covers the update path, not create.

This PR fixes it on the client by converting `InterpreterItemComponent` from `UntypedFormBuilder` to typed reactive forms:

- Non-checkbox values are sent as strings (like the classic UI); checkbox values stay real booleans, and `"true"`/`"false"` strings from corrupted data are normalized back on save.
- New request DTOs mirror the fields `InterpreterOption.java` actually reads; the UI-only `session`/`process` fields (dead since ZEPPELIN-1210) are no longer sent.
- Fixes wrong `Properties.value`/`type` interface types; response option fields that Gson omits when null are now optional.

Alternative to #5147, which stringifies checkbox booleans too — persisting `"false"` makes an unchecked checkbox render as checked on reload. Credit to @kevinjmh for the original diagnosis.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6427 (fixes ZEPPELIN-6395)

No unit test infra exists in `zeppelin-web-angular` (Playwright e2e only), so verified by `ng build` (strict, 0 errors), lint/prettier, and manually: numeric property saves as `"60000"` in `interpreter.json` (create and update), JDBC paragraph runs without `NumberFormatException`, unchecked checkbox stays unchecked after reload, and a regression pass over create/edit/cancel, property and dependency CRUD, and interpreter binding mode options.

N/A

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/interfaces/interpreter.ts
+++ b/zeppelin-web-angular/src/app/interfaces/interpreter.ts
@@ -65,11 +65,46 @@ interface SnapshotPolicy {
 interface Properties {
   [key: string]: {
     name: string;
-    value: boolean;
-    type: string;
+    value: string | number | boolean | null;
+    type: InterpreterPropertyTypes;
     defaultValue?: string;
     description?: string;
   };
+}
+
+export interface InterpreterPropertyValue {
+  name: string;
+  value: string | boolean;
+  type: InterpreterPropertyTypes;
+}
+
+/**
+ * Request shape for creating/updating an interpreter setting.
+ * Mirrors the fields the server actually reads (InterpreterOption.java) —
+ * UI-only state such as `session`/`process` must not be sent.
+ */
+export interface InterpreterSettingOption {
+  isExistingProcess: boolean;
+  isUserImpersonate: boolean;
+  owners: string[];
+  perNote: string;
+  perUser: string;
+  /** null is accepted by the server and kept as the default -1 (unset) */
+  port: number | null;
+  host: string;
+  remote: boolean;
+  setPermission: boolean;
+}
+
+export interface InterpreterSettingRequest {
+  name: string;
+  group: string;
+  option: InterpreterSettingOption;
+  properties: Record<string, InterpreterPropertyValue>;
+  dependencies: Array<{
+    groupArtifactVersion: string;
+    exclusions: string[];
+  }>;
 }
 
 interface InterpreterGroupItem {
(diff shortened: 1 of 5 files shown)


zeppelin-web-angular/src/app/pages/workspace/interpreter/item/item.component.ts
```diff
@@ -86,31 +122,38 @@ export class InterpreterItemComponent extends DestroyHookComponent implements On
...
+    const { isExistingProcess, isUserImpersonate, owners, perNote, perUser, port, host, remote, setPermission } =
+      formData.option;
+    const setting: InterpreterSettingRequest = {
+      name: formData.name,
+      group: formData.group,
+      option: { isExistingProcess, isUserImpersonate, owners, perNote, perUser, port, host, remote, setPermission },
+      properties,
+      dependencies: formData.dependencies.map(({ groupArtifactVersion, exclusions }) => ({
+        groupArtifactVersion,
+        exclusions: exclusions.split(',').filter(s => s !== '')
```
voidmatcha:
```suggestion
        exclusions: exclusions
          .split(',')
          .map(s => s.trim())
          .filter(s => s !== '')
```
nit: no per-entry `.trim()`, so `"org.foo:bar, org.baz:qux"` stores `" org.baz:qux"` with a leading space, which won't match as a Maven exclusion.

zeppelin-web-angular/src/app/interfaces/interpreter.ts
```diff
@@ -65,13 +65,48 @@ interface SnapshotPolicy {
 interface Properties {
   [key: string]: {
     name: string;
-    value: boolean;
-    type: string;
+    value: string | number | boolean | null;
+    type: InterpreterPropertyTypes;
     defaultValue?: string;
```
voidmatcha:
```suggestion
    defaultValue?: string | boolean;
```
nit: `defaultValue` arrives as a real JSON boolean for checkbox properties. The setting templates use unquoted booleans (jdbc/src/main/resources/interpreter-setting.json:115 `"defaultValue": false`)

jongyoul:
@tbonelee Could you please share screenshots of your changes if changed?

tbonelee:
@jongyoul It changes the parts which use untyped Angular form API into typed form API (This does not change previous behaviors).
Additionally, I ensure number type `value` fields to be serialized into JSON string value to fix [ZEPPELIN-6395](https://issues.apache.org/jira/browse/ZEPPELIN-6395)(This changes the behavior).

When a user changes the value of the field whose type is `number`,
### Previously
`value` is saved as JSON number (in `conf.interpreter.json`).

image

### Currently
`value` is saved as JSON string (numeric string).

image

voidmatcha (approved):
LGTM 👍👍

pan3793 (approved):

tbonelee:
Thanks. Merged into master.

PR #5276, opened 2026-06-25
[ZEPPELIN-6208] Enable DuckDB support in JDBC Interpreter by excluding incompatible default properties

hyunw9:
What is this PR for?

  The JDBC interpreter forwards all of an interpreter's prefixed properties (default.*) straight to the JDBC driver when opening a connection. Many of these keys are consumed by Zeppelin itself or by the DBCP connection pool — e.g. driver, url, precode, statementPrecode,
  completer.ttlInSeconds, validationQuery, maxIdle — and are not valid JDBC connection properties.

  Lenient drivers (PostgreSQL, MySQL) silently ignore unknown properties, so this went unnoticed. But strict drivers reject the connection outright. For example, DuckDB fails with:

  SQLException: Invalid Input Error: The following options were not recognized:
                completer.ttlInSeconds, url, driver, maxIdle

  This currently makes it impossible to use DuckDB (and other strict drivers such as MS SQL Server) with the JDBC interpreter.

  Until now this was worked around with a driver-specific whitelist that only applied when the driver class was Presto/Trino (PRESTO_PROPERTIES). That approach is hard to maintain: every new strict driver needs its own if branch and allow-list, and it also strips any legitimate driver
  property the user added but that wasn't on the list.

  This PR replaces that with a generic, driver-agnostic deny-list:

  - Introduces NON_DRIVER_PROPERTIES, a single source of truth listing only Zeppelin-internal and DBCP pool keys (user/password are deliberately kept, since they are standard JDBC properties).
  - Adds toDriverProperties(), which returns a filtered copy of the properties handed to the driver, leaving the original untouched (the previous Presto code mutated the shared per-user Properties in place — a latent bug).
  - Removes the Presto/Trino special-case branch and the PRESTO_PROPERTIES whitelist; the generic filter subsumes it.
  - Adds an optional escape hatch, zeppelin.jdbc.driver.excludeProperties (comma-separated), so operators can exclude additional keys for future drivers without a code change.

  Result: DuckDB, MS SQL Server, Trino/Presto, and any future strict driver work through one consistent rule, while genuine driver properties (e.g. SSL, useSSL, sslmode) still pass through unchanged.

  What type of PR is it?

  Bug Fix

  Todos

  - [x] Replace Presto/Trino whitelist with a generic internal-property deny-list
  - [x] Keep user/password and arbitrary driver properties (e.g. SSL) flowing to the driver
  - [x] Add unit tests for the filtering logic
  - [x] Add end-to-end tests against real strict drivers (DuckDB embedded, Trino)

| Property setting (1) | Property setting (2) |
  |:---:|:---:|
  | 스크린샷 2026-06-25 오후 9 20 23 | 스크린샷 2026-06-25 오후 9 20 29 |

  **2. Result — Before / After**

  | Before — DuckDB connection error | After — connected successfully |
  |:---:|:---:|
  | before-duckdb-error | after-success
  

  What is the Jira issue?

  - [ZEPPELIN-6208](https://issues.apache.org/jira/browse/ZEPPELIN-6208)

  How should this be tested?

  Automated tests added in JDBCInterpreterTest:

  - testToDriverProperties — internal/pool keys are removed; user, password, and arbitrary driver props (SSL) are kept; the original Properties is not mutated.
  - testToDriverPropertiesWithUserDefinedExcludes — zeppelin.jdbc.driver.excludeProperties strips additional user-specified keys.
  - testDuckDbConnectionWithInternalProperties — end-to-end: configures the interpreter with internal keys present and runs CREATE/INSERT/SELECT against an embedded DuckDB (no server needed). Fails on the old code, passes here.
  - testTrinoConnectionWithInternalProperties — end-to-end against a real Trino coordinator; auto-skipped via assumeTrue when none is reachable on localhost:8080.

  mvn -pl jdbc -am test -Dtest=JDBCInterpreterTest

  Manual: add a JDBC interpreter with default.driver=org.duckdb.DuckDBDriver, default.url=jdbc:duckdb:, add the org.duckdb:duckdb_jdbc dependency in the interpreter settings, leave an internal key such as default.completer.ttlInSeconds=120, and run a query — it now connects successfully.

  Questions:

  - Does the license files need to update? - X
  - Is there breaking changes for older versions? - X
  - Does this need documentation? - Optional

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -43,6 +43,8 @@
 
     <!--test library versions-->
     <mockrunner.jdbc.version>1.0.8</mockrunner.jdbc.version>
+    <duckdb.jdbc.version>1.3.1.0</duckdb.jdbc.version>
+    <trino.jdbc.version>481</trino.jdbc.version>
   </properties>
 
   <dependencies>
@@ -64,6 +66,20 @@
       <groupId>com.mysql</groupId>
       <artifactId>mysql-connector-j</artifactId>
       <version>${mysql.connector.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.duckdb</groupId>
+      <artifactId>duckdb_jdbc</artifactId>
+      <version>${duckdb.jdbc.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.trino</groupId>
+      <artifactId>trino-jdbc</artifactId>
+      <version>${trino.jdbc.version}</version>
       <scope>test</scope>
     </dependency>
 
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -147,14 +147,29 @@ public class JDBCInterpreter extends KerberosInterpreter {
   private static final String DBCP_STRING = "jdbc:apache:commons:dbcp:";
   private static final String MAX_ROWS_KEY = "zeppelin.jdbc.maxRows";
 
-  private static final Set<String> PRESTO_PROPERTIES = new HashSet<>(Arrays.asList(
-          "user", "password",
-          "socksProxy", "httpProxy", "clientTags", "applicationNamePrefix", "accessToken",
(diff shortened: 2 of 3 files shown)


jongyoul (approved):
LGTM! Thanks for enabling DuckDB support in JDBC Interpreter by excluding incompatible default properties.

jongyoul:
Merged into master (206b6a31).

PR #5288, opened 2026-07-11
[ZEPPELIN-6481] Restore interrupt status after InterruptedException in JDBC completion loading

HwangRock:
### What is this PR for?

The `catch (InterruptedException e)` block around `awaitTermination(...)` in `createOrUpdateSqlCompleter(...)` in `JDBCInterpreter` logs a warning and closes the connection, but never restores the interrupt status. Catching `InterruptedException` clears the thread's interrupt flag, so the interrupt is swallowed: callers on the completion request path can no longer detect it, and cooperative cancellation and shutdown signals are silently dropped. This is the classic swallowed-interrupt anti-pattern.

This restores the interrupt with `Thread.currentThread().interrupt()` as the last statement of the block, after the connection-close attempt. Placing it last keeps the existing resource-release behavior running first, then leaves the interrupt status intact when the method returns so the calling thread can detect it.

### What type of PR is it?

Bug Fix

### Todos

* [x] Restore interrupt status in the `InterruptedException` catch block
* [x] Add a unit test covering the restored interrupt status

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-6481

* `./mvnw test -pl jdbc -Dtest=JDBCInterpreterTest`
* New test `testCreateOrUpdateSqlCompleterRestoresInterruptStatusOnTimeout` pre-interrupts the calling thread so `awaitTermination(...)` throws `InterruptedException` immediately, drives the completer through the catch path, and asserts `Thread.currentThread().isInterrupted()` is `true` on return. Before the fix the flag is cleared and the assertion fails; after the fix it passes. 32 tests in the class pass.

N/A

### Questions:

### DIFF
--- a/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
+++ b/jdbc/src/main/java/org/apache/zeppelin/jdbc/JDBCInterpreter.java
@@ -328,6 +328,7 @@ public class JDBCInterpreter extends KerberosInterpreter {
           LOGGER.warn("Error close connection", e1);
         }
       }
+      Thread.currentThread().interrupt();
     }
     return completer;
   }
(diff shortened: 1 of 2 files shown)


ParkGyeongTae (approved):
LGTM. +1

jongyoul:
Merged into master (dfd8cf7d).

PR #5289, opened 2026-07-11
[ZEPPELIN-6453] Handle secondary NotebookRepo remove failures consistently

move-hoon:
### What is this PR for?
In `NotebookRepoSync`, secondary storage failures were already handled for `save()` and `move()`, but `remove()` still used a generic `for` loop with a `TODO` comment for handling secondary remove failures.

This PR makes the `remove()` failure handling consistent with the existing `save()` and `move()` behavior:
* **Primary Storage (`getRepo(0)`)**: A failure during removal stops the operation and propagates the `IOException`.
* **Secondary Storage (`getRepo(1)`)**: A failure during removal is caught and explicitly logged, preventing silent inconsistencies without interrupting the overall operation.

To achieve this, the `for` loop in `remove(String noteId, String notePath, AuthenticationInfo subject)` was replaced with explicit `getRepo(0)` and `getRepo(1)` calls, and the outstanding `TODO` was removed.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6453

Tested via `NotebookRepoSyncTest` using Mockito to explicitly verify the primary vs. secondary failure behaviors:
* `testRemoveSucceedsWhenSecondaryRepoFails`: Verifies that an `IOException` from the secondary repo is caught and does not stop the removal process.
* `testRemoveFailsWhenPrimaryRepoFails`: Verifies that an `IOException` from the primary repo halts the operation and propagates the exception.
* All existing tests in `NotebookRepoSyncTest` continue to pass.

N/A

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/repo/NotebookRepoSync.java
@@ -208,10 +208,15 @@ public class NotebookRepoSync implements NotebookRepoWithVersionControl {
 
   @Override
   public void remove(String noteId, String notePath, AuthenticationInfo subject) throws IOException {
-    for (NotebookRepo repo : repos) {
-      repo.remove(noteId, notePath, subject);
+    getRepo(0).remove(noteId, notePath, subject);
+    if (getRepoCount() > 1) {
+      try {
+        getRepo(1).remove(noteId, notePath, subject);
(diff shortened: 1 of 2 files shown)


jongyoul (approved):
Reviewed the primary/secondary removal failure handling and its unit tests. The behavior is consistent with save() and move().

jongyoul:
Merged into master (7bb88cab).

PR #5292, opened 2026-07-12
[ZEPPELIN-6345] Fire NoteRemove event for each note when deleting a folder

HwangRock:
### What is this PR for?

Deleting a folder left its notes searchable after removal. `removeFolder(String, AuthenticationInfo)` in `Notebook` called `noteManager.removeFolder` first, which detached the notes from the tree, so the following `removeNote(noteId)` loaded a `null` note and skipped `fireNoteRemoveEvent`. `SearchService.deleteNoteIndex` was therefore never invoked and the Lucene search index kept stale documents for the deleted notes, so they still showed up in search results after emptying the trash or removing a folder.

This PR loads the notes non-destructively via a new `NoteManager.getNoteInfoRecursively`, runs the same per-note cleanup as `removeNote(Note, AuthenticationInfo)` (`setRemoved`, `removeNoteAuth`, `fireNoteRemoveEvent`) for each note, and only then deletes the folder. This also removes the pre-existing `NotebookRepo.remove is called twice` TODO, since the repo remove now happens once.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6345

https://github.com/user-attachments/assets/3bce191f-6f12-443a-bd75-2a0ada28712c

* Added `NotebookTest#testRemoveFolderFiresNoteRemoveEventForEachNote`: creates two notes under `/folder1`, registers a `NoteEventListener` counting `onNoteRemove`, calls `removeFolder("/folder1", ...)`, and asserts the event fired once per note. Fails before the change (count `0`), passes after.
* Manual: create notes, move them to trash, empty the trash, then search for a deleted note. Before the fix it still appears in results; after the fix it is gone.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/NoteManager.java
@@ -289,6 +289,17 @@ public class NoteManager {
   }
 
   /**
+   * Returns the NoteInfo of all notes under the given folder, without removing them.
+   *
+   * @param folderPath
+   * @return
+   * @throws IOException
+   */
+  public List<NoteInfo> getNoteInfoRecursively(String folderPath) throws IOException {
+    return getFolder(folderPath).getNoteInfoRecursively();
+  }
+
+  /**
    * Remove the folder from the tree and returns the affected NoteInfo under this folder.
    *
    * @param folderPath
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/Notebook.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/notebook/Notebook.java
@@ -552,11 +552,22 @@ public class Notebook {
 
   public void removeFolder(String folderPath, AuthenticationInfo subject) throws IOException {
     LOGGER.info("Remove folder {}", folderPath);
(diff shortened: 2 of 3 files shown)


hyunw9:
LGTM :)

ParkGyeongTae (approved):
LGTM. The root cause analysis is accurate — moving `noteManager.removeFolder()`
to after the per-note cleanup ensures `processNote()` can still find each note
in `notesInfo` and fire `NoteRemoveEvent` correctly.

While reviewing this, I noticed that `removeFolder` does not evict note cache
entries via `noteCache.removeNote()`, which is a pre-existing issue unrelated
to this fix. I've filed a separate Jira ticket for it — feel free to take a
look if you're interested.
https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6529

HwangRock:
@ParkGyeongTae Thanks for filing ZEPPELIN-6529 and letting me know! Confirmed the bug and sent a PR here: https://github.com/apache/zeppelin/pull/5296

ParkGyeongTae:
Thanks for the quick follow-up! I'll take a look at #5296.

ParkGyeongTae:
Merged into master (0c4b30f8).

PR #5300, opened 2026-07-14
[ZEPPELIN-6531] Fail fast when ZeppelinWebSocketClient cannot connect

HwangRock:
### What is this PR for?
`ZeppelinWebSocketClient.connect()` blocks forever when the websocket handshake fails. It waits on `connectLatch.await()` with no timeout, and the latch is only counted down in the success callback (`onConnect`). Pre-handshake failures — connection refused, TLS error, rejected upgrade — are not delivered through the annotated callbacks at all; Jetty completes the `CompletableFuture` returned by `WebSocketClient.connect()` exceptionally instead. The original code discards that future, so the failure never reaches the waiting thread and `connect()` hangs.

This affects any `zeppelin-client` (`ZSession`) user that passes a `MessageHandler` to `start()`/`reconnect()`: if the target Zeppelin server is down or unreachable, the calling thread blocks indefinitely.

Fix: wait on the future with a bounded timeout and surface handshake failures as `IOException`. On failure, stop the Jetty client so its threads are not leaked. `connectLatch` is unused after this change, so it is removed together with its unused `getConnectLatch()` getter.

Note on approach: adding `connectLatch.countDown()` to `onError` does not fix this. Against jetty-websocket 11, pre-handshake failures reach only the future, never `@OnWebSocketError`, so the latch would still never be released. The future is the only path that carries the failure.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Follow-up: make the connect timeout configurable via `ClientConfig` (currently a 30s constant)

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6531

Added `ZeppelinWebSocketClientTest` (the first unit test in `zeppelin-client`): connecting to a closed port must fail within a bounded time instead of hanging.

Also verified end-to-end against a docker-exposed port. Ran `nginx` on a port, which accepts the TCP connection but rejects the websocket upgrade (HTTP 404 on `/ws`), and pointed the client at both that port and a closed port.

Before (original code, same docker port):
```
[INFO] Running org.apache.zeppelin.client.websocket.ManualE2EVerifyTest
```
Hangs here — no result line. The forked surefire JVM stayed alive past 45s and had to be killed.

After (this PR):
```
[E2E] closed-port(1) failed-fast in 295ms -> IOException: Failed to establish websocket connection to ws://127.0.0.1:1/ws
[E2E] open-port-non-ws(18080) failed-fast in 45ms -> IOException: Failed to establish websocket connection to ws://127.0.0.1:18080/ws
```
The docker probe was a throwaway check and is not part of the PR; the committed test uses a closed port and needs no docker.

### Questions:
* Is there breaking changes for older versions? `getConnectLatch()` is removed. It is public but has no callers in the repo, and it only exposed `connect()`'s internal latch, which no longer exists.

### DIFF
--- a/zeppelin-client/src/main/java/org/apache/zeppelin/client/websocket/ZeppelinWebSocketClient.java
+++ b/zeppelin-client/src/main/java/org/apache/zeppelin/client/websocket/ZeppelinWebSocketClient.java
@@ -33,8 +33,11 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 
 /**
@@ -45,8 +48,8 @@ import java.util.concurrent.TimeUnit;
 public class ZeppelinWebSocketClient {
   private static final Logger LOGGER = LoggerFactory.getLogger(ZeppelinWebSocketClient.class);
   private static final Gson GSON = new Gson();
+  private static final long DEFAULT_CONNECT_TIMEOUT_MS = 30_000;
 
-  private CountDownLatch connectLatch = new CountDownLatch(1);
   private CountDownLatch closeLatch = new CountDownLatch(1);
 
   private Session session;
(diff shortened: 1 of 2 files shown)


tbonelee (approved):
LGTM

tbonelee:
Merged to master

PR #5301, opened 2026-07-14
[ZEPPELIN-6475] Throw clear error for missing InfluxDB token

gyowoo1113:
### What is this PR for?
`InfluxDBInterpreter.open()` read the `influxdb.token` property and immediately called `toCharArray()` on the returned value. When the token was not configured, `getProperty(...)` returned `null`, causing a `NullPointerException` without explaining that the authentication token was missing.

This PR validates the token before building the InfluxDB client. Missing, empty, and blank token values now throw an `InterpreterException` with a clear configuration message instead of a `NullPointerException`.

Unit tests were added for absent, empty, and whitespace-only token values. Other InfluxDB properties and client creation behavior are unchanged.

### What type of PR is it?
Bug Fix

### Todos
* [x] Validate missing, empty, and blank `influxdb.token` values
* [x] Throw `InterpreterException` with a clear error message
* [x] Add unit tests for each invalid token case

### What is the Jira issue?
[ZEPPELIN-6475](https://issues.apache.org/jira/browse/ZEPPELIN-6475)

`./mvnw test -pl influxdb` passes successfully.

To run only the interpreter test:

`./mvnw test -pl influxdb -Dtest=InfluxDBInterpeterTest` passes successfully.

Missing, empty, and blank `influxdb.token` values throw `InterpreterException` instead of `NullPointerException`.

N/A

### Questions:

### DIFF
--- a/influxdb/src/main/java/org/apache/zeppelin/influxdb/InfluxDBInterpreter.java
+++ b/influxdb/src/main/java/org/apache/zeppelin/influxdb/InfluxDBInterpreter.java
@@ -169,12 +169,17 @@ public class InfluxDBInterpreter extends AbstractInterpreter {
 
 
   @Override
-  public void open() {
+  public void open() throws InterpreterException {
 
     if (this.client == null) {
+      String token = getProperty(INFLUXDB_TOKEN_PROPERTY);
+      if (token == null || token.trim().isEmpty()) {
+        throw new InterpreterException("influxdb.token property is not set. Please configure the InfluxDB auth token.");
+      }
+
       InfluxDBClientOptions opt = InfluxDBClientOptions.builder()
           .url(getProperty(INFLUXDB_API_URL_PROPERTY))
-          .authenticateToken(getProperty(INFLUXDB_TOKEN_PROPERTY).toCharArray())
+          .authenticateToken(token.toCharArray())
           .logLevel(LogLevel.valueOf(
               getProperty(INFLUXDB_LOGLEVEL_PROPERTY, LogLevel.NONE.toString())))
           .org(getProperty(INFLUXDB_ORG_PROPERTY))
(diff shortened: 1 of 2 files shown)


influxdb/src/main/java/org/apache/zeppelin/influxdb/InfluxDBInterpreter.java
```diff
@@ -169,12 +169,17 @@ private QueryApi getQueryApi() {
 
 
   @Override
-  public void open() {
+  public void open() throws InterpreterException {
 
     if (this.client == null) {
+      String token = getProperty(INFLUXDB_TOKEN_PROPERTY);
+      if (token == null || token.trim().isEmpty()) {
```
hyunw9: nit :
How about `token.isBlank()` instead? it covers same situations but it reads more clearly.

hyunw9 (commented):
LGTM

jongyoul (dismissed):

influxdb/src/main/java/org/apache/zeppelin/influxdb/InfluxDBInterpreter.java
gyowoo1113: Good point, thanks! I’ve updated it to use isBlank().

ParkGyeongTae (approved):
Thanks for fixing this, @gyowoo1113! I pulled the branch and verified locally:
- Confirmed the NPE reproduces on master when influxdb.token is unset (getProperty(...).toCharArray() on a null value).
- The open() signature change to throws InterpreterException is safe — the parent Interpreter.open() already declares throws InterpreterException, so no callers are affected.
- String.isBlank() is fine here since the project targets Java 11.
- Ran ./mvnw test -pl influxdb -Dtest=InfluxDBInterpeterTest: all 5 tests pass (2 existing + 3 new for missing/empty/blank token).

ParkGyeongTae:
Merged into master (a46096fe).

PR #5302, opened 2026-07-14
[ZEPPELIN-6539] Fix DockerClient / container resource leaks on interpreter failure paths

hyunw9:
### What is this PR for?

The Docker interpreter launcher (DockerInterpreterProcess) can leak resources on error paths when interpreters are started/stopped repeatedly by notebook workloads.

1. stop() may leak the DockerClient. docker.close() was called after the try/catch that kills/removes the container. If killContainer/removeContainer throws an unexpected (unchecked) exception, it is not caught and docker.close() is skipped, leaking the underlying HTTP socket / file descriptors.
2. start() may leave an orphaned container. start() runs pull → create → start → copy files → exec → wait-for-register. If preparation fails after the container is started (e.g. copyRunFileToContainer / execInContainer), the running container is not cleaned up 

### What type of PR is it?
Bug Fix

### Todos
- [x] Fix stop() to always close the DockerClient
- [x] Fix start() to clean up a started container on preparation failure
- [x] Add unit tests (Mockito interaction tests)

### What is the Jira issue?
* [[ZEPPELIN-6539]](https://issues.apache.org/jira/browse/ZEPPELIN-6539)

Automated unit tests added in DockerInterpreterProcessTest (DockerClient is mocked):
- stop_alwaysClosesDockerClient_evenWhenKillContainerFails
- start_removesContainer_whenContainerPreparationFails 

### Questions:

### DIFF
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -394,12 +394,20 @@ public class DockerInterpreterProcess extends RemoteInterpreterProcess {
   private void cleanupContainerQuietly() {
     try {
       docker.killContainer(containerName);
-      docker.removeContainer(containerName);
     } catch (InterruptedException e) {
-      LOGGER.warn("Interrupted while cleaning up container {}", containerName, e);
+      LOGGER.warn("Interrupted while killing container {} during cleanup", containerName, e);
       Thread.currentThread().interrupt();
     } catch (DockerException e) {
-      LOGGER.warn("Failed to clean up container {} after start failure", containerName, e);
+      LOGGER.warn("Failed to kill container {} during cleanup", containerName, e);
+    }
+
+    try {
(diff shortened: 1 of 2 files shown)


zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
```diff
@@ -363,10 +385,22 @@ public void stop() {
...
+      docker.close();
     }
+  }
 
-    // Close the docker client
-    docker.close();
+  // Best-effort removal of a container that was (partially) created during start().
+  private void cleanupContainerQuietly() {
+    try {
+      docker.killContainer(containerName);
```
tbonelee, line 396: **Suggestion (non-blocking):** In `cleanupContainerQuietly()`, `killContainer` and `removeContainer` share one `try`, so if `killContainer` throws (e.g. the container was created but never started, so it isn't running), `removeContainer` is skipped and the container is still leaked. Could we run them independently so `removeContainer` always fires?

ParkGyeongTae (approved):
LGTM +1

ParkGyeongTae:
Merged into master (79a6530e).

PR #5305, opened 2026-07-15
[ZEPPELIN-6473] Replace deprecated substr() with slice() in note-create

kimyenac:
### What is this PR for?
The New UI note-create dialog generates note names using `String.prototype.substr()`, an Annex B (legacy) API the language spec recommends against and flags for eventual removal. This PR migrates the three `substr()` calls in `note-create.component.ts` to the recommended `slice()`.

The two methods are not blindly interchangeable — `substr(start, length)` takes a length, while `slice(start, end)` takes an end index — so each call site was checked individually to confirm behavior is preserved:

- `newNoteName()`: `substr(15)` → `slice(15)` — single-arg, runs to end of string (equivalent).
- `cloneNoteName()`: `substr(0, lastIndex)` → `slice(0, lastIndex)` — equivalent specifically because the start is `0`, so `length` and the end index coincide.
- `cloneNoteName()`: `substr(lastIndex)` → `slice(lastIndex)` — single-arg, runs to end of string (equivalent).

Generated names for both new notes (`Untitled Note N`) and cloned notes are unchanged from current behavior.

### What type of PR is it?
Improvement

### Todos
* [x] - Replace all three `substr()` calls with `slice()`
* [x] - Confirm no remaining `substr()` usage in the file

### What is the Jira issue?
[ZEPPELIN-6473](https://issues.apache.org/jira/browse/ZEPPELIN-6473)

This repository's frontend has no unit-test infrastructure, so verification is via lint plus a production build:
```
cd zeppelin-web-angular
npm run lint && npm run build:angular
```
Both pass.

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
+++ b/zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts
@@ -74,8 +74,8 @@ export class AngularDragDropService {
               const identifierTokens = argsString ? argsString.split(',') : [];
               const args = identifierTokens.map(item => $parse(item.trim())(scope));
 
-              const constructorName =
-                _callbackStr.indexOf('.') !== -1 ? _callbackStr.substr(0, _callbackStr.indexOf('.')) : null;
+              const dotIndex = _callbackStr.indexOf('.');
+              const constructorName = dotIndex !== -1 ? _callbackStr.slice(0, dotIndex) : null;
               // @ts-ignore
               const constructorCandid = constructorName && scope[constructorName];
               const constructor =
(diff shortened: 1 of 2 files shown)


voidmatcha (changes requested):
The note-create changes look behavior-preserving to me.

One small scope suggestion: although ZEPPELIN-6473 was filed around note-create, if this PR is intended as a New UI `substr()` cleanup, it may be worth replacing the remaining Angular UI usage as well:

https://github.com/apache/zeppelin/blob/master/zeppelin-web-angular/src/app/services/angular-drag-drop.service.ts#L78

```ts
const dotIndex = _callbackStr.indexOf('.');
const callback = dotIndex !== -1 ? _callbackStr.slice(0, dotIndex) : null;
```

This keeps the existing behavior while avoiding the deprecated substr() call.

kimyenac:
Thanks for the review, @voidmatcha !

Good point extended the cleanup to the remaining `substr()` in `angular-drag-drop.service.ts` in the latest commit. Cached `indexOf('.')` in a local and used `slice(0, dotIndex)`, which keeps the same behavior including the `null` branch when there's no dot. Verified with `npm run lint`.

ParkGyeongTae (approved):
LGTM +1

voidmatcha (approved):
LGTM 👍

voidmatcha:
Merged into master (09355ef9).

PR #5308, opened 2026-07-16
[ZEPPELIN-6541] Pin maven-checkstyle-plugin version so it is actually applied

ParkGyeongTae:
### What is this PR for?
The root `pom.xml` declares a `plugin.checkstyle.version` property (`2.17`) intended to pin the `maven-checkstyle-plugin` version used across all modules, but the plugin's `<pluginManagement>` declaration never referenced it via a `<version>` tag. As a result Maven silently resolved a different, non-deterministic plugin version regardless of what the property said — confirmed locally: the property said `2.17`, but the build actually ran `checkstyle:3.4.0:check`. This PR adds the missing `<version>${plugin.checkstyle.version}</version>` tag and bumps the property to `3.6.0` (the current latest release) so the pinned version is the one that actually runs.

### What type of PR is it?
Improvement

### Todos
* [x] Add `<version>${plugin.checkstyle.version}</version>` to the `maven-checkstyle-plugin` block in `<pluginManagement>`
* [x] Bump `plugin.checkstyle.version` from `2.17` to `3.6.0`

### What is the Jira issue?
[ZEPPELIN-6541](https://issues.apache.org/jira/projects/ZEPPELIN/issues/ZEPPELIN-6541)

```bash
./mvnw -pl shell verify -DskipTests
```
Confirm the log shows checkstyle:3.6.0:check (not an arbitrary/older version) and the build completes with BUILD SUCCESS. No module overrides the plugin version locally, so all modules using maven-checkstyle-plugin inherit the pinned version consistently.

Screenshots (if appropriate)

N/A

Questions:

- Does the license files need to update? No.
- Is there breaking changes for older versions? No — build-time only change, no runtime behavior affected.
- Does this needs documentation? No.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,6 @@
     <plugin.avro.version>1.7.7</plugin.avro.version>
     <plugin.buildhelper.version>1.7</plugin.buildhelper.version>
     <plugin.buildnumber.version>1.4</plugin.buildnumber.version>
-    <plugin.checkstyle.version>2.17</plugin.checkstyle.version>
     <plugin.cobertura.version>2.7</plugin.cobertura.version>
     <plugin.download.version>1.6.0</plugin.download.version>
     <plugin.exec.version>1.6.0</plugin.exec.version>


tbonelee:
One thing worth weighing before wiring it back up: that `<version>` reference wasn't just never added, it was removed on purpose in #4884 ([ZEPPELIN-6138] Update Apache Parent), which un-pinned ~10 plugins at once to defer their versions to the parent pom. It just left the `plugin.checkstyle.version` property behind as leftover cruft.

So re-pinning would special-case checkstyle away from the parent-managed version, while the other plugins from that commit stay parent-managed. Unless there's a concrete reason to need `3.6.0`, would just deleting the orphaned property fit the ticket's intent better? Curious what you think.

ParkGyeongTae:
Good catch, thanks for digging into the #4884 history — you're right.

I re-checked: the `<version>` tag was removed intentionally in ZEPPELIN-6138 as part of deferring ~10 plugins to the Apache parent pom's managed versions, and `plugin.checkstyle.version` was just leftover cruft from that change, not a missing reference that needed to be restored.

I've updated this PR (and the Jira description) to remove the orphaned property instead of re-pinning it, so maven-checkstyle-plugin stays consistent with the other plugins from #4884 that are parent-managed. Verified locally that the build still runs checkstyle (now inherited from the parent pom, currently 3.4.0) and passes.

PTAL, thanks again for catching this.

tbonelee (approved):

tbonelee:
Merged into master

PR #5310, opened 2026-07-16
[ZEPPELIN-6537] Close DockerClient in a finally block in DockerInterpreterProcess.stop()

HwangRock:
### What is this PR for?
`DockerInterpreterProcess.stop()` calls `docker.close()` after the `killContainer` / `removeContainer` try-catch, but `close()` is outside that block and the `catch` only handles `InterruptedException` and `DockerException`:

```java
try {
  docker.killContainer(containerName);
  docker.removeContainer(containerName);
} catch (InterruptedException e) { ...; Thread.currentThread().interrupt(); }
catch (DockerException e) { ... }
docker.close();   // skipped if kill/remove throws anything else
```

If `killContainer` / `removeContainer` throws an exception that is neither of those two (e.g. an unchecked `RuntimeException` from the client), the exception propagates out of `stop()` and `docker.close()` is skipped. The `DockerClient` holds an HTTP connection to the Docker daemon, so its socket and file descriptors leak. Interpreters are started and stopped repeatedly, so the leak accumulates and can exhaust file descriptors ("Too many open files") on long-running servers.

Fix: move `docker.close()` into a `finally` block so the client is always closed regardless of how `killContainer` / `removeContainer` exit.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - none

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6537

Added `testStopClosesDockerClientEvenWhenKillContainerThrows` in `DockerInterpreterProcessTest`: it injects a mock `DockerClient`, stubs `killContainer` to throw a `RuntimeException`, calls `stop()`, and verifies the exception still propagates (`assertThrows`) while `docker.close()` is still invoked (`verify(mockDocker).close()`).

On the old code this fails with "Wanted but not invoked: dockerClient.close()"; with the fix it passes. Deterministic (mock-based), no flakiness. Full `DockerInterpreterProcessTest` passes.

### Questions:
* Is there breaking changes for older versions? No — behavior only changes on the failure path (close now always runs); the exception still propagates as before.

### DIFF
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -363,10 +363,10 @@ public class DockerInterpreterProcess extends RemoteInterpreterProcess {
       Thread.currentThread().interrupt();
     } catch (DockerException e) {
       LOGGER.error(e.getMessage(), e);
+    } finally {
+      // Close the docker client
+      docker.close();
     }
-
-    // Close the docker client
-    docker.close();
   }
 
   // Because docker can't create a container with the same name, it will cause the creation to fail.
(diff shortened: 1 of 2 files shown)


tbonelee:
Thanks for the fix! This same change (moving `docker.close()` into a `finally` block in `stop()`) already landed in `main` via #5302. See the current `DockerInterpreterProcess.stop()`. That PR also handles the `start()` orphaned-container case, so this looks like a duplicate. Mind closing it? Thanks for spotting this!

HwangRock:
@tbonelee You're right, thanks for the pointer. #5302 already covers this in `stop()` and additionally handles the `start()` orphan case that mine doesn't, so this is a duplicate. Closing it — thanks for catching it.

PR #5311, opened 2026-07-16
[ZEPPELIN-6311] Support JDK 17

pan3793:
## What is this PR for?

Enable Zeppelin to build, test, and run on JDK 17. This involves two areas of work:

- Flink Scala version alignment
- JPMS module-access flags

Cassandra test migration is handled separately in #5412.

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/ZEPPELIN-6311

## How should this be tested?

CI passes on both JDK 11 and JDK 17 across all modules. Verified locally:

- `flink/flink-scala-2.12` module: 7 `FlinkInterpreterTest` tests pass on both JDK 11 and JDK 17
- `bin/common.sh` syntax-checked; `JPMS_JAVA_OPTS` verified in `JAVA_OPTS` and `JAVA_INTP_OPTS`

## Questions

- Does the licenses file need update? No

### 1. Flink: Scala 2.12.7 -> 2.12.20

Flink 1.19/1.20 binary tgz bundles `flink-scala_2.12-*.jar` which shades `scala-library`, `scala-compiler`, and `scala-reflect` 2.12.7. Scala 2.12.7's `JrtClassPath.asURLs()` crashes on JDK 17+ with:

```
java.lang.RuntimeException: /packages cannot be represented as URI
    at java.base/jdk.internal.jrtfs.JrtPath.toUri(JrtPath.java:175)
    at scala.tools.nsc.classpath.JrtClassPath.asURLs(DirectoryClassPath.scala:204)
```

This was fixed in Scala 2.12.9 ([scala/bug#11608](https://github.com/scala/bug/issues/11608)).

Changes:
- Change `flink.scala.version` from `2.12.7` to `2.12.20` in `flink/pom.xml` (removes the `jdk17+` profile). The compile-time version must match the runtime version due to binary incompatibility in `Settings.usejavacp()` between 2.12.7 and 2.12.20.
- Add `maven-antrun-plugin` to `flink-1.19` and `flink-1.20` profiles that patches the downloaded Flink tgz: strips `scala/*` classes from `flink-scala_2.12-*.jar` via `zip -d`, then copies `scala-library`/`scala-compiler`/`scala-reflect` 2.12.20 jars into `FLINK_HOME/lib`. Scoped to Flink 1.x only.
- Add `patchFlinkScala()` in `DownloadUtils` for integration tests that download Flink via `DownloadUtils.downloadFlink()` to `~/.cache/`. Guards on `scalaVersion == "2.12"` and `flinkVersion < 2.0`.

Experimental support for Java 17 was added in Flink 1.18. ([FLINK-15736](https://issues.apache.org/jira/browse/FLINK-15736))

Flink 1.19/1.20 sticky with Scala 2.12.7 due to [FLINK-12461](https://github.com/apache/flink/pull/14780), upgrade Scala is a necessary step for JDK 17 support.

### 2. JPMS args and JDK 17 CI

JDK 17 enforces strong module encapsulation. Many libraries (Netty, Mockito, etc.) require `--add-opens` flags to access internal JDK APIs. These flags were already needed for surefire tests but were either missing or duplicated per-module.

Assisted-by: GLM 5.2

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
     steps:
       - name: Checkout
         uses: actions/checkout@v5
@@ -90,7 +90,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
     env:
       INTERPRETERS: 'hbase,jdbc,file,flink-cmd,cassandra,elasticsearch,bigquery,livy,groovy,java,neo4j,sparql,mongodb,influxdb,shell'
     steps:
@@ -136,7 +136,7 @@ jobs:
       fail-fast: false
       matrix:
         python: [ 3.9 ]
-        java: [ 11 ]
+        java: [ 11, 17 ]
     steps:
       - name: Checkout
         uses: actions/checkout@v5
@@ -181,7 +181,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
     steps:
       # user/password => root/root
       - name: Start mysql
@@ -285,7 +285,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
+        java: [ 11, 17 ]
     steps:
       - name: Checkout
         uses: actions/checkout@v5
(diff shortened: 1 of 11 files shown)


pan3793:
@jongyoul, the current changes in this PR should make Zeppelin compatible with JDK 17, the real hacky part is the patch for Flink - Flink does not work out of the box until 2.0, let me know your thoughts about this part. (I may not be able to add support for Flink 2.0, and I hope others can take this if you think we should not hack Flink 1.19/1.20 but upgrade to Flink 2.0 first)

pan3793:
I split the Cassandra part change into ZEPPELIN-6639

jongyoul (approved):
Thanks for working through this, Cheng! I'm happy to move forward with the Flink 1.x workaround.

It would be helpful to document that running the Flink interpreter on JDK 17 assumes the Flink distribution pointed to by `FLINK_HOME` is also compatible with JDK 17, including its bundled Scala compiler and runtime. A follow-up PR would be perfectly fine; this doesn't need to block this PR.

Separately, what do you think about offering both JDK 11 and JDK 17 builds of Zeppelin for the upcoming major release to ease the transition?

pan3793:
thanks @jongyoul, will update docs about Flink binary requirements later.

for binary release tarball, it's unnecessary to provide separate builds for JDK 11 and 17, the single artifact built with JDK 11 should be compatible with both JDK 11 and 17 - use Spark as an example: the [spark-3.5.9-bin-hadoop3.tgz](https://dlcdn.apache.org/spark/spark-3.5.9/spark-3.5.9-bin-hadoop3.tgz) can run with JDK 8/11/17

PR #5312, opened 2026-07-16
[ZEPPELIN-6467] Propagate serialization failures in Resource.serializeObject() instead of swallowing them

gyowoo1113:
### What is this PR for?
`Resource.serializeObject(Object)` caught serialization exceptions, printed the stack trace, and still returned a `ByteBuffer` from the partially written output. When serialization failed, callers could receive truncated or empty data instead of the original `IOException`.

This PR removes the exception-swallowing catch block and uses try-with-resources for `ObjectOutputStream`. Serialization failures now propagate as the already-declared `IOException`, allowing the existing caller-side error handling to run and preventing invalid buffers from being returned.

A unit test was added with a `Serializable` object that throws `IOException` during serialization. The method signature and behavior for valid serializable objects are unchanged.

### What type of PR is it?
Bug Fix

### Todos
* [x] Remove the exception-swallowing catch block
* [x] Remove `printStackTrace()`
* [x] Propagate serialization failures as `IOException`
* [x] Use try-with-resources for `ObjectOutputStream`
* [x] Add a regression test for serialization failure

### What is the Jira issue?
[ZEPPELIN-6467](https://issues.apache.org/jira/browse/ZEPPELIN-6467)

`./mvnw test -pl zeppelin-interpreter -Dtest=ResourceTest` passes successfully.

To verify that the shaded interpreter JAR includes the change:

`./mvnw clean package -pl zeppelin-interpreter,zeppelin-interpreter-shaded -DskipTests` passes successfully.

N/A

### Questions:

### DIFF
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/resource/Resource.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/resource/Resource.java
@@ -390,17 +390,12 @@ public class Resource implements JsonSerializable, Serializable {
     if (o == null || !(o instanceof Serializable)) {
       return null;
     }
-
+    
     ByteArrayOutputStream out = new ByteArrayOutputStream();
-    try {
-      ObjectOutputStream oos;
-      oos = new ObjectOutputStream(out);
+    try (ObjectOutputStream oos = new ObjectOutputStream(out)) {
       oos.writeObject(o);
-      oos.close();
-      out.close();
-    } catch (Exception e) {
-      e.printStackTrace();
     }
+
     return ByteBuffer.wrap(out.toByteArray());
   }
 
--- a/zeppelin-interpreter/src/main/java/org/apache/zeppelin/resource/Resource.java
+++ b/zeppelin-interpreter/src/main/java/org/apache/zeppelin/resource/Resource.java
@@ -390,7 +390,6 @@ public class Resource implements JsonSerializable, Serializable {
     if (o == null || !(o instanceof Serializable)) {
       return null;
     }
-    
     ByteArrayOutputStream out = new ByteArrayOutputStream();
     try (ObjectOutputStream oos = new ObjectOutputStream(out)) {
       oos.writeObject(o);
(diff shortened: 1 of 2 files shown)


zeppelin-interpreter/src/main/java/org/apache/zeppelin/resource/Resource.java
```diff
@@ -390,17 +390,12 @@ public static ByteBuffer serializeObject(Object o) throws IOException {
     if (o == null || !(o instanceof Serializable)) {
       return null;
     }
-
+
```
tbonelee: Nit: trailing whitespace.

```suggestion

```

zeppelin-interpreter/src/test/java/org/apache/zeppelin/resource/ResourceTest.java
```diff
@@ -90,4 +92,19 @@ void testInvokeMethod_shouldAbleToInvokeMethodWithClass() throws ClassNotFoundEx
     Resource r = new Resource(null, new ResourceId("pool1", "name1"), "object");
     assertEquals(true, r.invokeMethod("startsWith", new Class[]{ java.lang.String.class }, new Object[]{"obj"}));
   }
+
+  @Test
+  void testSerializeObject_shouldPropagateIOException() {
+    IOException exception = assertThrows(IOException.class, () -> Resource.serializeObject(new FailingSerializable()));
+    assertEquals("Serialization failed", exception.getMessage());
+  }
+
```
tbonelee: Nit: trailing whitespace.

```suggestion

```

tbonelee (dismissed):
LGTM. Removing the swallowing `catch` and switching to try-with-resources means serialization failures now propagate as the declared `IOException` instead of returning a truncated buffer. Callers already handle it, and the `FailingSerializable` regression test is a good fit. Thanks!

tbonelee (approved):

tbonelee:
Merged into master

tbonelee:
Thanks for the fix! One small process note for next time: please prefix the PR title with the Jira key, e.g. `[ZEPPELIN-6467] ...`, following the contribution convention. Our merge tooling picks up the Jira issue from the title, so without it the Jira follow-up (resolve, fixVersion, comment) does not happen automatically. Already handled manually for this one, so nothing to do here. Thanks again!

PR #5314, opened 2026-07-17
[ZEPPELIN-6435] Remove frontend build dependency from IndexHtmlServletTest

gyowoo1113:
### What is this PR for?
`IndexHtmlServletTest` included a disabled Angular HTML addon test because it depended on the generated file at `zeppelin-web-angular/dist/zeppelin/index.html`. That artifact is not built during normal `zeppelin-server` tests, so the test could not run independently.

This PR replaces the frontend build dependency with a temporary, test-owned `index.html` file and re-enables `testZeppelinWebAngularHtmlAddon`.

The `zeppelin-web` index and the Angular-style index exercise different branches in `IndexHtmlServlet`. The `zeppelin-web` index does not contain explicit `</head>` and `</body>` closing tags, so the servlet falls back to inserting the head addon before `<body>` and the body addon before `</html>`. The Angular-style index contains both closing tags, so the re-enabled test verifies that the configured head and body addons are inserted immediately before `</head>` and `</body>`.

### What type of PR is it?
Improvement

### Todos
- [x] Remove the dependency on `zeppelin-web-angular/dist/zeppelin/index.html`
- [x] Create a temporary, test-owned Angular-style `index.html`
- [x] Re-enable `testZeppelinWebAngularHtmlAddon`
- [x] Verify addon insertion before `</head>` and `</body>`

### What is the Jira issue?
[[ZEPPELIN-6435]](https://issues.apache.org/jira/browse/ZEPPELIN-6435)

`./mvnw test -pl zeppelin-server -Dtest=IndexHtmlServletTest` passes successfully.

N/A

### Questions:

### DIFF
--- a/zeppelin-server/src/test/java/org/apache/zeppelin/server/IndexHtmlServletTest.java
+++ b/zeppelin-server/src/test/java/org/apache/zeppelin/server/IndexHtmlServletTest.java
@@ -25,11 +25,12 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.net.URL;
-
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 import org.apache.zeppelin.conf.ZeppelinConfiguration;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 import jakarta.servlet.ServletConfig;
 import jakarta.servlet.ServletContext;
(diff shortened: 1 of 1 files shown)


zeppelin-server/src/test/java/org/apache/zeppelin/server/IndexHtmlServletTest.java
```diff
@@ -43,8 +44,9 @@ class IndexHtmlServletTest {
     private final static String TEST_HEAD_ADDON = "<!-- bar -->";
 
     private final static String FILE_PATH_INDEX_HTML_ZEPPELIN_WEB = "../zeppelin-web/dist/index.html";
-    private final static String FILE_PATH_INDEX_HTML_ZEPPELIN_WEB_ANGULAR = "../zeppelin-web-angular/dist/zeppelin/index.html";
-
+
```
tbonelee:
```suggestion

```

tbonelee:
Since the PR's goal is to make `IndexHtmlServletTest` independent of the frontend build, could we apply the same treatment to `testZeppelinWebHtmlAddon` in this PR? It still hardcodes `../zeppelin-web/dist/index.html`, another git-ignored build artifact. It only passes in CI because the core job builds `zeppelin-web` via `-Pweb-classic`; a clean local `./mvnw test -pl zeppelin-server` would fail it for the exact reason you're fixing on the Angular side. Reusing the `@TempDir` pattern here (with a document that omits the closing tags, so it exercises the fallback branch in `IndexHtmlServlet`) would make both tests fully self-contained and complete the stated goal.

gyowoo1113:
Thanks for catching this. I updated `testZeppelinWebHtmlAddon` to use `@TempDir` as well and to verify the fallback behavior using an HTML document without closing tags.

tbonelee (approved):

PR #5304, opened 2026-07-14
[ZEPPELIN-6536] Stabilize notebook keyboard shortcut e2e editor seeding

voidmatcha:
### What is this PR for?
This PR stabilizes the notebook keyboard-shortcut e2e suite by avoiding keyboard-driven fixture seeding in the Monaco editor.

In the Firefox Playwright job, setup could trigger Monaco suggestions while typing fixture content. That let Enter accept a suggestion before the shortcut assertion ran, corrupting expected text, for example `line` becoming `license`.

The helper now seeds fixture text through the editor textarea without shortcut-style key events. The shortcut assertions still use real keyboard events.

This also re-enables the clone-content shortcut coverage now that https://github.com/apache/zeppelin/pull/5254 has been merged. The skip is no longer needed, and keeping it would leave that shortcut path without regression coverage.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6536

### Screenshots (if appropriate)
N/A

### Questions:
* Does the license files need to update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #5315, opened 2026-07-17
[ZEPPELIN-6544] Harden keyboard-shortcut e2e helpers against silent passes

voidmatcha:
### What is this PR for?
Follow-up to ZEPPELIN-6536 (#5304). A cross-review of the merged keyboard-shortcut suite found four silent-pass risks in the helpers; this PR closes them:

- `pressShortcutFromHostUntil` now polls for the press's effect before retrying, so a slow toggle is not double-pressed and the loop cannot return on the transient state between two in-flight effects.
- The clear-output test asserts the result is visible before pressing; previously a no-output run let the helper skip the press entirely.
- InsertAbove/InsertBelow emptiness assertions now gate on the editor being rendered (`waitForEditorRendered`); an unrendered editor reads as `''` and matched vacuously.
- `setCodeEditorContent` compares normalized equality instead of containment, so stale fixture text fails at seeding time.

Test-only change; no product code touched.

### What type of PR is it?
Improvement

### What is the Jira issue?
ZEPPELIN-6544

### Questions:

### DIFF
--- a/zeppelin-web-angular/e2e/models/notebook-keyboard-page.ts
+++ b/zeppelin-web-angular/e2e/models/notebook-keyboard-page.ts
@@ -10,7 +10,7 @@
  * limitations under the License.
  */
 
-import test, { expect, Locator, Page } from '@playwright/test';
+import { expect, Locator, Page } from '@playwright/test';
 import { navigateToNotebookWithFallback } from '../utils';
 import { ShortcutsMap } from '../../src/app/key-binding/shortcuts-map';
 import { ParagraphActions } from '../../src/app/key-binding/paragraph-actions';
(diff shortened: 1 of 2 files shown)


voidmatcha:
Ran the full keyboard-shortcut suite 10x across all browser projects (retries=0), all green:
https://github.com/voidmatcha/zeppelin/actions/runs/29631108745

Shortcuts are still driven by real key events; only the fixture text is seeded through the notebook REST API (the same way notebooks are already created). Typing directly into Monaco let Firefox's autocomplete corrupt the seeded text, so this also removes the previous per-browser branching and unifies seeding into one path. Real typing itself is still exercised by the cursor-move, tab-indentation, and autocomplete tests. Content assertions read persisted server state to avoid Monaco's non-deterministic whitespace rendering.

tbonelee (approved):

tbonelee:
Merged into master

PR #5316, opened 2026-07-17
[ZEPPELIN-5876] Implement DockerInterpreterProcess.isAlive() using container state

hyunw9:
### What is this PR for?

The Docker launcher's isAlive() isn't really implemented. It just calls isRunning(), which only checks whether the interpreter's Thrift port accepts a connection (checkIfRemoteEndpointAccessible). That tells you the port is open, not that the process behind it is alive. A container can be OOMKilled or already gone and still look alive if the port happens to answer for a moment. It also doesn't match the InterpreterClient contract, which says isAlive should reflect process status and stay separate from running. The method still had //TODO(ZEPPELIN-5876): Implement it more accurately sitting on it.

While I was there I filled in getErrorMessage(), which always returned null. When the container isn't running it now says why: OOMKilled, or a non-zero exit code. That logic lives in a small describeContainerFailure() helper to keep getErrorMessage() short.

One change is test-only: the docker field is now 'VisibleForTesting', so tests can inject a mock DockerClient and skip needing a real daemon.

### What type of PR is it?
Feature

### Todos
- [x] Implement isAlive() from the container's actual state (running/paused)
- [x] Implement getErrorMessage() (OOMKilled / exit code)
- [x] Unit tests with a mocked DockerClient

### What is the Jira issue?
[[ZEPPELIN-5876]](https://issues.apache.org/jira/browse/ZEPPELIN-5876)

Automated unit tests in DockerInterpreterProcessTest :
- isAlive_trueWhenContainerRunning - running == true → alive
- isAlive_falseWhenContainerExitedOrOomKilled — running == false → not alive
- getErrorMessage_reportsOomKilled - oomKilled == true → message contains the reason

### Questions:
* Does the license files need to update? - no
* Is there breaking changes for older versions? - no

### DIFF
--- a/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
+++ b/zeppelin-plugins/launcher/docker/src/main/java/org/apache/zeppelin/interpreter/launcher/DockerInterpreterProcess.java
@@ -46,6 +46,7 @@ import com.spotify.docker.client.exceptions.DockerException;
 import com.spotify.docker.client.messages.Container;
 import com.spotify.docker.client.messages.ContainerConfig;
 import com.spotify.docker.client.messages.ContainerCreation;
+import com.spotify.docker.client.messages.ContainerState;
 import com.spotify.docker.client.messages.ExecCreation;
 import com.spotify.docker.client.messages.HostConfig;
 import com.spotify.docker.client.messages.PortBinding;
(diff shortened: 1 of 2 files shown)


tbonelee:
A couple of points:

**1. (should address) An unrelated existing test was removed**

`start_removesContainer_evenWhenKillFailsDuringCleanup` was deleted in this PR. That test covered a distinct rollback-resilience case (`removeContainer` must still fire even when both `startContainer` and `killContainer` throw during cleanup), which is unrelated to the `isAlive` / `getErrorMessage` change here. The remaining `start_removesContainer_whenContainerPreparationFails` only exercises the kill-succeeds path, so this scenario is no longer covered. Looks like it may have been dropped accidentally while adding the new tests. Could you restore it?

**2. (minor) Test name is broader than what it verifies**

`isAlive_falseWhenContainerExitedOrOomKilled` only sets `running=false` and never touches `oomKilled`. Since `isAlive` only looks at `running`/`paused` and does not distinguish exited vs OOMKilled, the behavior is fine, but the name promises more coverage than exists. Renaming to something like `isAlive_falseWhenContainerNotRunning` would match the actual assertion. The OOMKilled path is already covered by `getErrorMessage_reportsOomKilled`.

tbonelee (approved):

tbonelee:
Merged into master

PR #2543, opened 2017-08-17
[ZEPPELIN-2860] Invalid last job execution time

byung-u:
### What is this PR for?
Fixed invalid last job execution time

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
* [ZEPPELIN-2860](https://issues.apache.org/jira/browse/ZEPPELIN-2860)

### How should this be tested?
Run Notebook and check job page execution time

### Screenshots (if appropriate)
image

### Questions:
* Does the licenses files need update? NO
* Is there breaking changes for older versions? NO
* Does this needs documentation? NO

PR #5319, opened 2026-07-18
[ZEPPELIN-6546] Restore Math.max accumulation for NoteJobInfo.unixTimeLastRun

HwangRock:
### What is this PR for?
Job Manager reports a wrong "last run" time for any note whose most recently executed paragraph is not the last paragraph in the note.

`NoteJobInfo(Note)` in `JobManagerService` overwrites `lastRunningUnixTime` on every loop iteration, so the last paragraph always wins:

```java
lastRunningUnixTime = getUnixTimeLastRunParagraph(paragraph);
```

Consequences:
- Both UIs render and sort by `unixTimeLastRun` (`jobmanager.filter.js` in zeppelin-web, `job-manager.component.ts` in zeppelin-web-angular), so the displayed relative time and the Recently-Update sort order are wrong.
- The incremental path `getNoteJobInfoByUnixTime()` filters with `unixTimeLastRun > lastUpdateServerUnixTime`, so a note that just ran can be silently dropped from `LIST_UPDATE_NOTE_JOBS` pushes when an older paragraph sits at the bottom.

This is a regression of ZEPPELIN-2860, which fixed exactly this in 2017 with `Math.max` (#2543). The ZEPPELIN-3737 refactor moved the logic from NotebookServer into JobManagerService and dropped the `Math.max` (001c621c7). This PR restores it.

Verified on a live server: a two-paragraph note whose first paragraph finished today and whose second paragraph was created 8 days ago and never ran showed "8 days ago" in Job Manager before the fix and the correct run time after.

https://github.com/user-attachments/assets/4508d689-2d6a-4998-bbbc-326a6006a9d9

### What type of PR is it?
Bug Fix

### Todos
* [x] Restore `Math.max` accumulation in `NoteJobInfo`
* [x] Regression tests

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6546

`mvn test -pl zeppelin-server -Dtest=JobManagerServiceTest` — the two new tests pin `unixTimeLastRun` to the max paragraph timestamp via the `getNoteJobInfoByUnixTime` filter boundary and fail without the fix.

Manually: create a note with two paragraphs, run only the first one, open the Job Manager page. Before the fix the note shows the second paragraph's creation date as its last run and sorts accordingly; after the fix it shows the actual run time.

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/service/JobManagerService.java
@@ -194,7 +194,7 @@ public class JobManagerService {
         }
         // get data for the job manager.
         ParagraphJobInfo paragraphItem = new ParagraphJobInfo(paragraph);
-        lastRunningUnixTime = getUnixTimeLastRunParagraph(paragraph);
+        lastRunningUnixTime = Math.max(lastRunningUnixTime, getUnixTimeLastRunParagraph(paragraph));
         paragraphs.add(paragraphItem);
       }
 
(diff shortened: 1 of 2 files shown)


tbonelee (approved):
LGTM

tbonelee:
Merged into master

PR #5323, opened 2026-07-19
[ZEPPELIN-6441] Replace outdated type assertions with satisfies

kimyenac:
### What is this PR for?
The Angular UI now uses TypeScript ~5.9.3, so the pre-4.9 workaround TODOs that asked to replace broad type assertions with `satisfies` are obsolete. This replaces those outdated assertions with `satisfies`, which improves compile-time checking without changing runtime behavior.

- `app.module.ts`: `as JoinedEditorOptions` → `satisfies JoinedEditorOptions`, so unsupported editor options are caught at compile time.
- `create-repository-modal.component.ts`: `as Record<...>` → `satisfies Record<...>`, so every form control key is validated against `CreateInterpreterRepositoryForm`.
- `notebook-paragraph-keyboard-event-handler.ts`: `as const` → `as const satisfies ...` for both the action→handler map and the Monaco-handled action list. `as const` is intentionally kept so the downstream `typeof`-based literal indexing still narrows correctly, while `satisfies` now validates keys/values (e.g. invalid handler names are rejected).

All three changes are compile-time-only annotations; runtime behavior is unchanged.

### What type of PR is it?
Improvement

### What is the Jira issue?
[ZEPPELIN-6441](https://issues.apache.org/jira/browse/ZEPPELIN-6441)

- `cd zeppelin-web-angular && npm run lint` passes (0 errors).
- Verified the `satisfies` checks are active: injecting an unsupported editor option is caught as TS2353, and an invalid handler name is caught as TS2418.

### Questions:
* Does this need documentation? No

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zeppelin-web-angular/src/app/app.module.ts
+++ b/zeppelin-web-angular/src/app/app.module.ts
@@ -60,8 +60,7 @@ registerLocaleData(en);
       useValue: {
         defaultEditorOption: {
           scrollBeyondLastLine: false
-          // TODO: Change 'as' to 'satisfies' when typescript version is over 4.9 to detect unsupported editor options at compile time.
-        } as JoinedEditorOptions,
+        } satisfies JoinedEditorOptions,
         onLoad: loadMonaco
       }
     },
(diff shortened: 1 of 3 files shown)


tbonelee (approved):
LGTM. Nice cleanup of the obsolete pre-4.9 TODOs.

Ran `tsc --noEmit -p src/tsconfig.json` with the changes applied: 0 errors. Also confirmed the `satisfies` checks are actually enforced (injecting an invalid handler name fails with TS2418), and the `Record<ParagraphActions, ...>` passing means the full enum is covered. `as const satisfies` keeps the downstream `typeof` narrowing intact. Compile-time only, no runtime change.

jongyoul (approved):

tbonelee:
Merged into master

PR #5326, opened 2026-07-19
[ZEPPELIN-6464] Throw IllegalArgumentException for an unsupported Scala version in SparkInterpreterLauncher

miinhho:
### What is this PR for?
This PR updates `SparkInterpreterLauncher` to throw `IllegalArgumentException` when the Scala version parsed from `spark-submit --version` is outside the supported range.

Unsupported Scala versions are invalid input values for this validation path, so `IllegalArgumentException` better describes the failure than a generic `Exception`.

The exception message is intentionally unchanged, and the `detectSparkScalaVersion(...)` method signature still declares `throws Exception` because the method can still raise checked exceptions from process execution, stream reading, and fallback Scala version detection.

This is behavior-preserving for callers: the existing caller catches `Exception` and wraps it in an `IOException`, so the surfaced behavior and public API remain unchanged.

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6464

* Build and run the module tests:
```  
./mvnw test -pl zeppelin-server --am
```

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -285,7 +285,7 @@ public class SparkInterpreterLauncher extends StandardInterpreterLauncher {
       } else if (scalaVersion.startsWith("2.13")) {
         return "2.13";
       } else {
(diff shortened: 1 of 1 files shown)


jongyoul (approved):

tbonelee:
Merged into master

PR #5329, opened 2026-07-20
[ZEPPELIN-6555] Avoid deadlock in ManagedInterpreterGroup.close() by not holding the group lock while closing session interpreters

hyunw9:
### What is this PR for?
`ManagedInterpreterGroup.close()` can deadlock with a concurrent `RemoteInterpreter.open()` because the two paths take the same two monitors in the opposite order.

Two monitors are involved: the interpreter-group monitor and an individual interpreter's monitor.

- `open()` takes the interpreter monitor first (`synchronized(this)`) and then the group monitor (via `getOrCreateSession()` and the angular-registry push).  Order: interpreter → group.
- `close(String)` is a `synchronized` method, so it holds the group monitor while it spawns the per-interpreter close threads and `join()`s them. Each close thread runs `interpreter.close()`, which takes the interpreter monitor. Order: group → interpreter.

So when an interpreter is opened while its session is concurrently closed (for example, restarting or shutting down an interpreter while a paragraph on that session is still starting up), the two orders form a circular wait.

### What type of PR is it?
Bug Fix

### Todos
* [x] Stop holding the group monitor while closing session interpreters in `close(String)`
* [x] Keep only the session-map removal and last-session teardown under the lock
* [x] Add regression test `ManagedInterpreterGroupTest#close_doesNotDeadlockWithConcurrentOpen`

### What is the Jira issue?
[[ZEPPELIN-6555]](https://issues.apache.org/jira/browse/ZEPPELIN-6555)

- Added a deterministic regression test,`ManagedInterpreterGroupTest#close_doesNotDeadlockWithConcurrentOpen`.
### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/ManagedInterpreterGroup.java
@@ -101,22 +101,29 @@ public class ManagedInterpreterGroup extends InterpreterGroup {
    * Close all interpreter instances in this session
    * @param sessionId
    */
-  public synchronized void close(String sessionId) {
-    LOGGER.info("Close Session: {} for interpreter setting: {}", sessionId, interpreterSetting.getName());
-    close(sessions.remove(sessionId));
+  public void close(String sessionId) {
+    LOGGER.info("Close Session: {} for interpreter setting: {}",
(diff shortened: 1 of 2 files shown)


jongyoul (approved):
LGTM! Thanks for resolving the deadlock in ManagedInterpreterGroup.close() by not holding the group lock while closing session interpreters.

jongyoul:
Merged into master (3d71dfef).

PR #5335, opened 2026-07-21
[ZEPPELIN-6525] Authenticate via REST for Selenium login preconditions

HwangRock:
### What is this PR for?

`test-selenium-with-spark-module-for-spark-3-5` fails on ~88% of master runs. Every failure lands in `AbstractZeppelinIT.authenticationUser`: the Bootstrap 3 login modal races its fade animation and the WebSocket `session_logout` auto-open, so the login click gets intercepted, the field is not interactable, or the modal never reaches its shown state. Across eight failing master runs the exception rotates between those three, which is why hardening any single step never held on its own. Since that helper is the shared login precondition for `AuthenticationIT`, `PersonalizeActionsIT` and `InterpreterModeActionsIT`, one racy modal destabilizes the whole job.

Test-only, no production changes.

- Add `authenticationUserViaRest`, which logs in with a synchronous XHR `POST /classic/api/login` issued from inside the browser's own JS context, so the browser session itself becomes authenticated. Reloading afterwards lets `app.js`'s pre-bootstrap `GET /security/ticket` render the logged-in state. No modal opens, so none of its races apply. Used wherever login is only a precondition (35 call sites across the three ITs).
- Keep `authenticationUser` as hardened by ZEPPELIN-1836, and keep `AuthenticationIT.testSimpleAuthentication` on it, so the classic login modal keeps its automated coverage.
- Add `extractNoteIdFromCurrentUrl`, which strips any `?query`/`#fragment` after the last path segment. After a reload the SPA appends `?ref=...` to note URLs, which otherwise leaks into the extracted note id and breaks the `@href` lookups.
- Correct `PersonalizeActionsIT.testDynamicFormAction` (see below).

An earlier approach injected a JSESSIONID obtained from a separate `java.net.http` client via `addCookie`; that session never transferred to the browser — the server logged a successful login while the browser stayed anonymous. Issuing the request inside the page authenticates the real session, which the server logs confirm (`LoginRestApi` returns the expected principal and `SecurityRestApi` then reports it for the browser).

Rebased onto current `master` so it sits on top of #5250.

### Why `testDynamicFormAction` changes

The test asserted that a non-owner's dynamic-form edit is ignored (`Status: Before`). The server does the opposite: `NotebookService.runParagraph` applies the caller's params to that user's own paragraph copy, `Note.run` executes that copy, and `GUI.textbox` prefers the params value over the default. Nothing filters a non-owner's params.

The same wrong expectation was already corrected for `testGraphAction` in 238e03242 (ZEPPELIN-4355), which also added the owner-side check that proves isolation; the dynamic-form case was missed at the time and still has no such step. The assertion has not been revisited since 2017, and its message (`"The output of graph mode is changed"`) is a leftover from the graph test.

It passed until now only because a late WebSocket broadcast replaces `paragraph.settings` and reverts the typed form value before the run. The test now waits for the input to actually hold the new value, so that race fails loudly instead of passing silently.

Adding the owner-side isolation step would fail on a real production bug — in personalized mode the caller's params are written to the shared master paragraph before the personalized branch is taken, and then persisted. Filed separately as ZEPPELIN-6556; not addressed here.

### Scope and related issues

- **ZEPPELIN-6525, login modal** — fixed, though not the way the ticket proposed. Driving the modal defensively (retrying open/fill/submit, tolerating `ElementClickInterceptedException`) was tried first and made things worse: one error became six, and ignoring the interception threw away the diagnostic the raw exception carried. Taking the modal out of the precondition path addresses the class of failure rather than one instance of it.
- **ZEPPELIN-6525, `ParagraphActionsIT` dynamic-form sleeps** — out of scope here. `ParagraphActionsIT` reports `Tests run: 21, Failures: 0, Errors: 0` on master and on this branch, so it is not what keeps the job red. Those fixed sleeps remain a latent race and deserve a follow-up.
- **ZEPPELIN-6526, REST session instead of per-test UI logins** — implemented, with one deviation: the proposed `addCookie` injection does not work, for the reason above, so the request is issued from inside the page instead. The `?ref=` risk the ticket flagged did materialise and is handled by `extractNoteIdFromCurrentUrl`.
- **ZEPPELIN-6556** — production isolation bug found while investigating this; filed, not fixed here.

### What type of PR is it?

Bug Fix (test infrastructure)

### What is the ticket?

https://issues.apache.org/jira/browse/ZEPPELIN-6525

Re-run `test-selenium-with-spark-module-for-spark-3-5`. Locally:

```
ZEPPELIN_SELENIUM_BROWSER=chrome ./mvnw -o failsafe:integration-test failsafe:verify \
  -pl zeppelin-integration -Dit.test=AuthenticationIT -DfailIfNoTests=false \
  -Pintegration -Pusing-source-tree -Pweb-classic -Pspark-scala-2.12 -Pspark-3.5
```

`AuthenticationIT` (3/3, including the modal-based `testSimpleAuthentication`) and `PersonalizeActionsIT#testDynamicFormAction` pass locally.

### Questions

- Does this need documentation? No.

### DIFF
--- a/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
+++ b/zeppelin-integration/src/test/java/org/apache/zeppelin/AbstractZeppelinIT.java
@@ -50,6 +50,7 @@ abstract public class AbstractZeppelinIT {
   protected static final long MAX_IMPLICIT_WAIT = 30;
   protected static final long MAX_BROWSER_TIMEOUT_SEC = 30;
   protected static final long MAX_PARAGRAPH_TIMEOUT_SEC = 120;
+  private static final String CLASSIC_LOGIN_PATH = "/classic/api/login";
 
   protected void authenticationUser(String userName, String password) {
     WebElement loginModal = manager.getWebDriver().findElement(By.id("loginModal"));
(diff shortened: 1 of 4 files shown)


voidmatcha:
Thanks @HwangRock 🙏 Now that #5250 is merged, `master` conflicts with this PR in `AbstractZeppelinIT.authenticationUser()` — #5250 hardened that method, while this PR replaces it with `authenticationUserViaRest()`. Could you rebase onto latest `master` and resolve it like this:

- **Keep the REST login** — it's the right root-cause fix for the precondition call sites, so take this PR's side there.
- **Don't delete `authenticationUser()`.** Keep it and have only `AuthenticationIT.testSimpleAuthentication` still use it, since #5250's hardened version is now the only automated coverage of the classic login modal (the Playwright suite only exercises the new UI login).

So the end state: `authenticationUser()` stays, used only by `testSimpleAuthentication`; everything else uses REST. That clears the conflict and keeps both fixes. If you'd rather drop classic-modal coverage entirely, let me know.

HwangRock:
@voidmatcha Thanks for the detailed resolution — made it really easy to follow!

Applied it: `authenticationUser()` untouched, used only by `testSimpleAuthentication`, the other 30 call sites on `authenticationUserViaRest()`. And good catch on the Playwright claim — you were right, it only covers the new UI. Fixed the description. CI is green, selenium included 🎉

One thing caught my eye while I was in there, and you'd know it better than anyone.

`run-playwright-e2e-tests (auth)` is 53m and drives the whole hour-long `frontend` workflow. Of the five browser projects in `playwright.config.js`, three are Chromium — the bundled build plus the `chrome` and `msedge` channels. ZEPPELIN-6314 sets the goal as "Chromium / Firefox / WebKit", and nothing in `e2e/` touches either channel; every browser branch is `webkit` or `firefox`. Dropping the two would leave engine coverage untouched and cut that job to ~27m — half the critical path from a config-only change!

WebKit stays, so ZEPPELIN-6450 is unaffected. JupyterLab runs Chromium only and marimo has WebKit commented out, so three engines is already generous.

Were the channels deliberate? Happy to open a ticket and PR if not 🙂

voidmatcha (approved):
LGTM 👍

Agreed on dropping the redundant Chrome/Edge channels. It'd be great if you open an issue and take it forward!

jongyoul:
Merged into master (a31975a0).

PR #5337, opened 2026-07-24
[ZEPPELIN-6551] Fix Job Manager crash on removal broadcast for a note not in the viewer's list

kimyenac:
### What is this PR for?

When a note is permanently deleted, the backend broadcasts a field-less `NoteJobInfo` stub (`noteName=null`, `isRemoved=true`) to **every** Job Manager subscriber, while each viewer's initial job list is owner-filtered (`JobManagerService#getNoteJobInfoByUnixTime`).

For a note the viewer does not own, `updateJobs` fell into the `currentJobIndex === -1` branch and pushed the stub into `this.jobs`. `filterJobs` then ran `job.noteName.match(noteNameReg)` and threw `TypeError: Cannot read properties of undefined (reading 'match')`. Because the stub persists in `this.jobs`, every subsequent `filterJobs` (each filter keystroke / job update) rethrew, so the page's filter/sort stayed broken until re-navigation.

This affects any user viewing `/jobmanager` when another user permanently deletes a note they don't own.

### What does this PR do?

- `updateJobs`: do not add removal stubs that aren't already in the list (`if (!updateJob.isRemoved)`). The guard is kept nested inside `if (currentJobIndex === -1)` rather than merged into that condition, so a removal stub can never fall through to the `else` branch's `splice(currentJobIndex, 1)` — merging would call `splice(-1, 1)` and silently drop the last job.
- `filterJobs`: guard `job.noteName?.match(...)` with optional chaining as defense in depth.

### What type of PR is it?

Bug Fix

### What is the Jira issue?

* https://issues.apache.org/jira/browse/ZEPPELIN-6551

* With two users: open `/jobmanager` as user A; as user B permanently delete a note A does not own; confirm A's page keeps filtering without a console `TypeError`.

### Questions:

### DIFF
--- a/zeppelin-web-angular/e2e/models/job-manager-page.ts
+++ b/zeppelin-web-angular/e2e/models/job-manager-page.ts
@@ -1,0 +1,44 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Locator, Page } from '@playwright/test';
+import { waitForZeppelinReady } from '../utils';
+import { BasePage } from './base-page';
+
+export class JobManagerPage extends BasePage {
+  readonly searchInput: Locator;
+  readonly jobItems: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.searchInput = page.locator('input[placeholder="Search jobs..."]');
+    this.jobItems = page.locator('zeppelin-job-manager-job');
+  }
+
+  async navigate(): Promise<void> {
+    await this.navigateToRoute('/jobmanager', { timeout: 60000 });
+    await this.page.waitForURL('**/#/jobmanager', { timeout: 60000 });
+    await waitForZeppelinReady(this.page);
+  }
+
+  jobItemByName(noteName: string): Locator {
+    return this.jobItems.filter({ hasText: noteName });
+  }
+
+  async filterByNoteName(noteName: string): Promise<void> {
+    await this.fillAndVerifyInput(this.searchInput, noteName);
+  }
+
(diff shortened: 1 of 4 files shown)


jongyoul:
@kimyenac Can we add test cases for this issue? :-)

kimyenac:
@jongyoul Thanks for the review! Added e2e regression tests in `job-manager-removal-broadcast.spec.ts` (5 cases).

Reverting the fix makes two of them fail with the reported `TypeError: Cannot read properties of undefined (reading 'match')` and a filter that stops responding to input. A third guards against folding the removal check into the `currentJobIndex === -1` condition, which would `splice(-1, 1)` the last job out of the list.

The Job Manager's WebSocket feed is stubbed in the test because `zeppelin.jobmanager.enable` defaults to false and `getNoteJobInfoByUnixTime` is owner-filtered, so the cross-user removal broadcast cannot be produced from a single session. All other WebSocket traffic still goes to the real server.

voidmatcha (commented):
Thanks for feedback! One suggestion.

test 1 and test 3 assert toHaveCount(2) right after the removal, but the list already has 2 items, so they pass on a state that is true before the removal is processed. On a regressed build filterJobs throws and the DOM stays frozen at 2, so the assertion stays true for the whole timeout and never catches it. The error check is one-shot too.

Sending a sentinel job after the removal (WebSocket delivery is FIFO) and awaiting its render guarantees the removal ran first, so both tests actually exercise the regression. (test 2 already has a barrier via the search input.)

a3afb7415

kimyenac:
@voidmatcha Good catch. With the fix reverted, the old spec passed "no listed job is dropped" outright, and the runtime-error test was a coin flip: run alone it passed in 760ms, before the `pageerror` landed.

Fixed in [8007d2247](https://github.com/apache/zeppelin/pull/5337/commits/8007d2247c91dfccfca6149e393b57520c28efd5). Both tests now broadcast a sentinel job after the removal and wait for it to render. Same socket, FIFO, same listener, so the sentinel appearing proves the removal was handled. On a regressed build `filterJobs` keeps throwing, so it never appears at all. The counts became `toHaveCount(3)` (Alpha + Beta + sentinel), which also pins down that the stub was dropped. The filter test keeps its search-box barrier; only its redundant count assertion is gone.

Verified on chromium: 5/5 pass with the fix, tests 1-3 fail with it reverted, and the same 3 fail if the guard is folded into `currentJobIndex === -1`.

voidmatcha (approved):
LGTM 👍

voidmatcha:
Merged into master (2ad49b3d).

PR #5340, opened 2026-07-25
[ZEPPELIN-6548] Use specific exceptions for Spark Scala fallback detection

miinhho:
### What is this PR for?

This PR refines exception types in `SparkInterpreterLauncher` when the fallback Spark Scala version detection validates the `SPARK_HOME/jars` layout.

Previously, `detectSparkScalaVersionByReplClass(...)` threw generic `Exception` for expected validation failures such as missing or duplicate `spark-repl` jars, or an unsupported `spark-repl` Scala suffix. This PR replaces those generic throw sites with more specific exception types while keeping the existing messages and caller behavior unchanged.

Missing or duplicate `spark-repl` jars now throw `IOException`, and an unrecognized Scala suffix now throws `IllegalArgumentException`.

This PR is a follow-up to [ZEPPELIN-6464](https://issues.apache.org/jira/browse/ZEPPELIN-6464)

### What type of PR is it?
Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6548

- Build and run the module tests:
```
./mvnw test -pl zeppelin-server --am
```

### Questions:

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java
@@ -318,10 +318,10 @@ public class SparkInterpreterLauncher extends StandardInterpreterLauncher {
     }
     
     if (sparkReplJars.isEmpty()) {
-      throw new Exception("No spark-repl jar found in SPARK_HOME: " + sparkHome);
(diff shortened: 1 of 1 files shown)


jongyoul (approved):
LGTM! Thanks for replacing raw Exception with specific exception types in SparkInterpreterLauncher.

jongyoul:
Merged into master (2147d79a).

PR #5343, opened 2026-07-26
[ZEPPELIN-6447] Add Module Federation typing for the published paragr…

JangAyeon:
### What is this PR for?
`PublishedParagraphComponent.loadReactWidget()` read the global Module Federation container via `window.reactApp` behind a `@ts-ignore` comment, so the compiler couldn't check the `container.get(...)` call that follows it.

`ReactRemoteLoaderService` already declares this container's shape as `RemoteContainer` and augments `Window.reactApp` with it globally. This PR exports that existing interface and reuses it here instead of adding a second, possibly conflicting `Window.reactApp` typing.

### What type of PR is it?
Improvement

### Todos
  * [x] Remove the `@ts-ignore` comment near `window.reactApp`
  * [x] Type the container with the existing `RemoteContainer` interface (covers the `get` API used here)
  * [x] Preserve the `window.reactApp not available` error when the container is missing

### What is the Jira issue?
 [ZEPPELIN-6447](https://issues.apache.org/jira/browse/ZEPPELIN-6447)

  * `cd zeppelin-web-angular && npm run lint` passes with no new errors.
  * Load a published paragraph with `?react=true` and confirm the React widget still mounts; without the remote entry script, the "window.reactApp not available" error still surfaces.

### Questions:

### DIFF
--- a/zeppelin-web-angular/src/app/pages/workspace/published/paragraph/paragraph.component.ts
+++ b/zeppelin-web-angular/src/app/pages/workspace/published/paragraph/paragraph.component.ts
@@ -30,6 +30,7 @@ import {
   ParagraphIResultsMsgItem
 } from '@zeppelin/sdk';
 import { HeliumService, MessageService, NgZService, NoteStatusService } from '@zeppelin/services';
+import { RemoteContainer } from '@zeppelin/share';
 import { SpellResult } from '@zeppelin/spell';
 import { isNil } from 'lodash';
 import { NzModalService } from 'ng-zorro-antd/modal';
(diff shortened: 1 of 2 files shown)


voidmatcha (approved):
`tsc` passes locally, so removing the `@ts-ignore` works.

One thing that isn't obvious from the issue: the type declaration ZEPPELIN-6447 asks for already exists. 52a638e5b (ZEPPELIN-6428, Jun 17) added a `declare global` block at `react-remote-loader.service.ts:22` declaring `Window.reactApp` as `RemoteContainer | undefined`, and it applies program-wide. The issue was filed on Jun 27, so criterion 2 ("the container type includes at least the `get` API") was already satisfied by the time it was written, and the `@ts-ignore` was the only part left.

Which means the import and the `export` don't add any typing. I reverted the import, the annotation and the `export` and ran `tsc` again, and the output was the same (exit 0, nothing printed). To rule out the type having fallen back to `any` I added `container?.nope()` with no import, and it still errored with `TS2339: Property 'nope' does not exist on type 'RemoteContainer'`.

f5ab35f31b5751fff19ab38528443f4a6fa44adc

Either way works, just flagging it.

tbonelee:
Merged into master

PR #5346, opened 2026-07-26
[ZEPPELIN-6561] Restore default SIGINT handler so python paragraph cancel works under daemon launch

HwangRock:
### What is this PR for?
Cancelling a running `%python` paragraph has no effect when Zeppelin is started via `zeppelin-daemon.sh`: the user code runs to completion and the result is recorded as SUCCESS while the job status becomes ABORT.

The cancel plumbing itself works. `cancel()` in `PythonInterpreter` sends SIGINT to the correct python pid, and the interpreter log shows it. The problem is signal disposition inheritance: `zeppelin-daemon.sh` starts the server with `nohup ... &` from a non-interactive shell, so per POSIX the whole process chain (ZeppelinServer JVM, interpreter JVM, python) inherits SIGINT=SIG_IGN, and CPython keeps SIGINT ignored instead of installing the KeyboardInterrupt handler when it starts with the signal already ignored. The SIGINT sent by `cancel()` is then a no-op. Running `signal.getsignal(signal.SIGINT)` inside an affected interpreter prints `Handlers.SIG_IGN`.

This cannot be fixed in the shell scripts, since POSIX forbids a non-interactive shell from resetting a signal that was ignored on entry. The fix restores the default SIGINT handler at the top of `zeppelin_python.py` when the inherited disposition is SIG_IGN. Starting Zeppelin in the foreground with `bin/zeppelin.sh` was never affected, which is why cancellation appears to work in some environments and not in others.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/ZEPPELIN-6561

* Automated: `testSigintDefaultHandlerRestoredWhenInheritedIgnored` in `PythonInterpreterTest` launches the interpreter through a shell wrapper that ignores SIGINT before exec'ing python, reproducing the disposition of a daemon launch, and asserts the default handler is restored inside the interpreter process. Without the fix the assertion fails with `Handlers.SIG_IGN`. Unlike the disabled `testCancelIntp`, it does not depend on timing.
* Manual: start Zeppelin with `bin/zeppelin-daemon.sh start`, run a `%python` paragraph such as `for i in range(1, 50): print(i); time.sleep(0.5)`, and cancel it a few seconds in. Before the fix it runs to 49 and stores SUCCESS. After the fix it stops immediately with a KeyboardInterrupt traceback and ERROR.

#### Before
https://github.com/user-attachments/assets/b9c6f5b4-24f6-4eb9-9019-37dab52526f0

#### After
https://github.com/user-attachments/assets/7b35d549-d97e-45cb-9d95-4ae5c543e0c0

### Questions:

### DIFF
--- a/python/src/main/resources/python/zeppelin_python.py
+++ b/python/src/main/resources/python/zeppelin_python.py
@@ -15,12 +15,19 @@
 # limitations under the License.
 #
 
-import os, sys, traceback, json, re
+import os, signal, sys, traceback, json, re
 
 from py4j.java_gateway import java_import, JavaGateway, GatewayClient
 from py4j.protocol import Py4JJavaError
 
 import ast
+
+# When Zeppelin is started via zeppelin-daemon.sh (nohup ... &), this process
+# inherits SIGINT=SIG_IGN and CPython keeps it ignored instead of installing the
+# KeyboardInterrupt handler, so PythonInterpreter.cancel()'s SIGINT would be a
+# no-op. Restore the default handler to keep paragraph cancellation working.
+if signal.getsignal(signal.SIGINT) == signal.SIG_IGN:
+  signal.signal(signal.SIGINT, signal.default_int_handler)
 
 class Logger(object):
   def __init__(self):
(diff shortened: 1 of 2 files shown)


jongyoul (approved):

jongyoul:
Merged into master (21b845a0).

PR #5347, opened 2026-07-26
[ZEPPELIN-6342] Enable the endOfLine option in Prettier and cover tsx files

kimyenac:
### What is this PR for?

Two related gaps in the `zeppelin-web-angular` Prettier setup.

`.prettierrc` did not declare `endOfLine`, so the line ending the formatter enforces was never stated in the project's own config. Prettier 3 defaults to `lf` and `.editorconfig` already declares `end_of_line=lf`, so the effective behavior does not change. Declaring it makes `.prettierrc` authoritative instead of dependent on those fallbacks.

React `.tsx` sources were covered by no formatter at all. The `prettier` glob in `lint` and `lint:fix` matched `{ts,js,json,css,html}` only, and `lint:react` runs ESLint alone, so nothing formatted the tsx files under `projects/zeppelin-react`. Adding `tsx` to both globs closes that gap. This PR also includes the resulting `prettier --write` output for the three files that were not formatted, `ImageRenderer.tsx`, `TableVisualization.tsx` and `PublishedParagraph.tsx`. Those changes are formatting only, with no logic change.

`scss` and `less` are deliberately left out. There are no scss files, and `.less` is intentionally listed in `.prettierignore`.

### What type of PR is it?
Improvement

### Todos
* [x] - Declare `"endOfLine": "lf"` in `.prettierrc`
* [x] - Add `tsx` to the `lint` and `lint:fix` prettier globs
* [x] - Apply `prettier --write` to the three unformatted tsx files

### What is the Jira issue?
* https://issues.apache.org/jira/browse/ZEPPELIN-6342

* `cd zeppelin-web-angular && npm run lint` passes, including `prettier --check` over the widened glob. The remaining `member-ordering` warnings are pre-existing and unrelated to this change.
* `npm run lint:react` passes, and the zeppelin-react vitest suite (14 tests) still passes after the reformat.
* `prettier --check "**/*.tsx"` on master flags exactly `ImageRenderer.tsx`, `TableVisualization.tsx` and `PublishedParagraph.tsx`, and passes on this branch.

N/A

### Questions:


jongyoul (approved):

voidmatcha (approved):
LGTM 👍

tbonelee:
Merged into master

PR #5348, opened 2026-07-26
[ZEPPELIN-6560] Lint the new UI e2e suite and document its conventions

voidmatcha:
### What is this PR for?

The new UI e2e suite has never been linted. `angular.json` sets `lintFilePatterns` to `src/**/*.ts` and `src/**/*.html`, so nothing under `e2e/` is checked beyond what the TypeScript compiler catches. `zeppelin-web-angular` has no unit or component test target either: no `test` target in `angular.json`, no `npm test` script, no spec files under `src/`. The e2e suite is the app's only automated coverage, so a check in it that quietly passes goes unnoticed. Two are in the tree today. `e2e/tests/notebook/paragraph/paragraph-functionality.spec.ts:184` is meant to confirm that a cancelled paragraph stops running:

```ts
await expect(page.locator('.paragraph-control .fa-spin')).not.toBeVisible({ timeout: 15000 });
```

`.fa-spin` is defined in the vendored FontAwesome stylesheet but applied by no Angular template, so the locator matches no element, and `not.toBeVisible` passes when a locator resolves to no DOM node. Separately, `e2e/tests/notebook/paragraph/copy-to-clipboard.spec.ts:25` attributes coverage to `PAGES.SHARE.SHARE_RESULT`, a key that does not exist: `SHARE_RESULT` lives under `PAGES.WORKSPACE`. Playwright transpiles without type checking, so the expression is `undefined` and that describe's coverage attribution is silently void. This PR fixes that one, since it is a single key. The `.fa-spin` assertion needs new page-object locators to assert anything real, so it is repaired in a follow-up change rather than here.

This PR adds `e2e/**/*.ts` to `lintFilePatterns` and applies `eslint-plugin-playwright`'s recommended flat config to that path. The pre-commit hook runs `eslint --fix` through lint-staged, and `--fix` applies fixes at any severity including warnings, so an autofixable rule rewrites test code without review. `playwright/no-useless-not` rewrites `.not.toBeVisible()` into `.toBeHidden()`; both forms pass when the locator matches nothing, but the rewritten one reads as a deliberate hidden-state check, so it would make the assertion above harder to spot. That rule is off. `prefer-web-first-assertions` and `no-networkidle` are lowered from the preset's `error` to `warn`, because the suite has 13 `waitForLoadState('networkidle')` calls and 9 assertions made on extracted values, and at `error` those would block every commit until the tests are rewritten. `ng lint` exits 0 on warnings, so they record the debt without gating the build. Effective severities are 36 rules at 12 error, 21 warn, 3 off, and every error-level rule passes on the current suite.

Auditing the tree against the rules the doc states turned up a few more, fixed here: an unused catch binding in `login-page.util.ts`, `eslint-disable` markers carrying the tracking key on the four ZEPPELIN-6379 skips, reasons on the `waitForTimeout` calls in `utils.ts` and `react-footer.spec.ts`, reasons on the two `eslint-disable` lines in `notebook-keyboard-page.ts`, a message on the unexplained WebKit skip in `dark-mode.spec.ts`, and the missing `PAGES` annotation in `user-menu-navigation.spec.ts`.

`e2e/AGENTS.md` is updated too. It drops the `Tooling: Use e2e-skills` section, which pointed contributors at an external personal repository, along with the vendor-specific setup notes in the preamble. In their place it gains an escape-hatch contract (`// JUSTIFIED:` and `// eslint-disable-next-line ... -- <why>`) and a locator hierarchy that explains why `exact: true` matters and when a CSS selector is acceptable. Four rules that claimed more than the tree delivers were brought back in line with it: the spec path pattern, where a POM is constructed, how POM files split, and web-first assertions. The doc also records what lint does not enforce, since the plugin has no rule for always-true assertions like the one above.

### What type of PR is it?
Improvement

### Todos
* [x] Run `npm run lint` with the e2e patterns added
* [x] Check every autofixable rule for one that could change a test's meaning through the pre-commit hook
* [x] Compare each override against the preset severity so the config states only real deviations
* [x] Confirm every `PAGES` reference resolves

### What is the Jira issue?
ZEPPELIN-6560

This is a lint and documentation change; no test behaviour changes.

```bash
cd zeppelin-web-angular
npm run lint
```

Expect 0 errors. The warnings are the debt described above. To see the new coverage on its own, run `npx eslint 'e2e/**/*.ts'`; before this PR `ng lint` reported nothing for `e2e/` because the path was not in `lintFilePatterns`. Adding a violation of any error-level rule, for example `test.describe.skip('x', () => {})` in a spec, should now fail the lint.

The coverage-key fix is visible in the run's annotations: `npm run e2e:fast -- tests/notebook/paragraph/copy-to-clipboard.spec.ts` attributes that describe to the `share/result` component instead of `undefined`. That spec needs a running shell interpreter, so it skips on CI.

### Questions:

### DIFF
--- a/zeppelin-web-angular/angular.json
+++ b/zeppelin-web-angular/angular.json
@@ -151,7 +151,7 @@
         "lint": {
           "builder": "@angular-eslint/builder:lint",
           "options": {
-            "lintFilePatterns": ["src/**/*.ts", "src/**/*.html"]
+            "lintFilePatterns": ["src/**/*.ts", "src/**/*.html", "e2e/**/*.ts"]
           }
         }
       }
--- a/zeppelin-web-angular/e2e/AGENTS.md
+++ b/zeppelin-web-angular/e2e/AGENTS.md
@@ -17,81 +17,67 @@ limitations under the License.
 
 # AGENTS.md
 
-> E2E (Playwright) conventions for `zeppelin-web-angular/e2e/`. A scoped companion
-> to the repository-root AGENTS.md, loaded only when working under `e2e/`.
-> See [AGENTS.md specification](https://github.com/agentsmd/agents.md).
+> E2E (Playwright) conventions for `zeppelin-web-angular/e2e/`. A scoped companion to the repository-root AGENTS.md, loaded only when working under `e2e/`. See [AGENTS.md specification](https://github.com/agentsmd/agents.md).
 
-Config: `zeppelin-web-angular/playwright.config.js` (Angular UI) and
-`playwright.classic.config.js` (legacy classic UI), sharing `playwright.shared.js`.
-This document is the shared source of truth for E2E conventions; Codex and
-agents.md-native tools read it directly.
-Claude Code / Gemini users can symlink `CLAUDE.md` / `GEMINI.md` to it locally
-(both gitignored, personal, not committed).
-
-## Tooling: Use e2e-skills
-
-Generate, review, and debug with [e2e-skills](https://github.com/voidmatcha/e2e-skills)
-instead of ad-hoc prompts. It encodes the rules below and adds a deterministic
-silent-pass scanner.
-
-```bash
-npx skills add voidmatcha/e2e-skills -g --all   # or -a <agent>
-```
-
-| Task | Skill |
-| --- | --- |
-| Generate new Playwright coverage | `playwright-test-generator` |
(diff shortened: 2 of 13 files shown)


jongyoul (approved):

tbonelee (approved):

tbonelee:
Merged into master

PR #5349, opened 2026-07-26
[ZEPPELIN-6543] Handle invokeMethod serialization failure as InterpreterRPCException

gyowoo1113:
### What is this PR for?
This PR follows up on ZEPPELIN-6467 / PR #5312.

Following PR #5312, `Resource.serializeObject()` propagates serialization failures as `IOException`. As a result, `RemoteInterpreterEventServer.invokeMethod()` can receive an exception while re-serializing a remote resource invocation result. The existing handler logged the exception and returned a null `ByteBuffer`, causing the generated Thrift client to report a missing result instead of preserving the original serialization failure.

The Jira issue identified this behavior through code analysis, but the server-side deserialize-and-re-serialize failure path had not yet been reproduced.

This PR adds a regression test using a serializable object that succeeds during the initial serialization and fails during the server-side second serialization. It then changes `invokeMethod()` to propagate the failure as `InterpreterRPCException`, allowing the original error message to reach the caller instead of being converted into an unrelated Thrift missing-result error.

The behavior for successfully serialized results is unchanged.

### What type of PR is it?
Bug Fix

### Todos
* [x] Reproduce the server-side deserialize-and-re-serialize failure path with a regression test
* [x] Propagate serialization failures as `InterpreterRPCException`
* [x] Verify that the propagated exception contains the original failure message

### What is the Jira issue?
[[ZEPPELIN-6543](https://issues.apache.org/jira/browse/ZEPPELIN-6543)]

`./mvnw test -pl zeppelin-server -Dtest=RemoteInterpreterEventServerTest` passes successfully.

N/A

### Questions:
* Code inspection suggests that `Resource.serializeObject()` may return `null` when the result is not serializable, but this path is not covered by the regression test in this PR. Should this case also be handled in this PR, or should it be addressed separately?

### DIFF
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/RemoteInterpreterEventServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/RemoteInterpreterEventServer.java
@@ -440,6 +440,7 @@ public class RemoteInterpreterEventServer implements RemoteInterpreterEventServi
       try {
         obj = Resource.serializeObject(ret);
       } catch (IOException e) {
+        LOGGER.error("invokeMethod failed", e);
         throw new InterpreterRPCException(e.toString());
       }
     }
--- a/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/RemoteInterpreterEventServer.java
+++ b/zeppelin-server/src/main/java/org/apache/zeppelin/interpreter/RemoteInterpreterEventServer.java
@@ -440,7 +440,7 @@ public class RemoteInterpreterEventServer implements RemoteInterpreterEventServi
       try {
         obj = Resource.serializeObject(ret);
       } catch (IOException e) {
-        LOGGER.error("invokeMethod failed", e);
+        throw new InterpreterRPCException(e.toString());
       }
     }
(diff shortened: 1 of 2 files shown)


tbonelee:
The exception propagation in `invokeMethod` now matches the pattern in `getResource()` right above it, which is a good cleanup. I ran the test locally and it passes.

One thing: removing `LOGGER.error` means the stack trace is gone on both the server and the client side. Only `e.toString()` (class name plus message) crosses the wire, so there is no way left to find out what actually failed to serialize, i.e. which field is not `Serializable`. I would suggest keeping the log and throwing:

```java
} catch (IOException e) {
  LOGGER.error("invokeMethod failed", e);
  throw new InterpreterRPCException(e.toString());
}
```

gyowoo1113:
Thanks for the suggestion! I restored the server-side error logging while keeping the InterpreterRPCException propagation, so the stack trace is still available for debugging.

jongyoul (approved):
LGTM! Thanks for handling invokeMethod serialization failure as InterpreterRPCException.

jongyoul:
Merged into master (427bf3aa).

PR #5350, opened 2026-07-26
[ZEPPELIN-6562] Fix specs that silently pass or flake because they target DOM the new UI no longer renders

voidmatcha:
### What is this PR for?

Several new UI specs target DOM that the previous UI rendered. Those selectors match no element, and the result splits two ways: some checks pass without proving anything, and some waits never resolve and time out. Both symptoms come from the same cause.

The silent passes are the heavier half. `waitForParagraphExecution` in `notebook-keyboard-page.ts` waits on `.paragraph-control .fa-spin, .running-indicator, .paragraph-status-running`. None of those classes appear in any new UI template, and `.fa-spin` is only a class definition in the vendored FontAwesome stylesheet that is never applied. The wait resolves immediately, so callers move on while the paragraph still reads `READY`. The keyboard suite uses this helper throughout.

The same selector is used as an assertion. `paragraph-functionality.spec.ts:184` is meant to confirm that a cancelled paragraph stopped running, but the locator matches nothing and `not.toBeVisible` passes whether or not execution stopped. This PR asserts that `.status` reaches `ABORT` instead. Asserting that the cancel control disappears is not enough, because that control only renders while the paragraph is `PENDING` or `RUNNING`, so it also disappears on natural completion. For the same reason the cancel has to happen after the run starts: the control is visible during `PENDING` too, so the existing code clicked it while the interpreter was still starting and exercised something other than what the test claims.

The flaky half comes from three places. The trash folder's "Empty" anchor collapses when the row loses hover: its bounding box is `{width:0, height:0}` unhovered and `{width:12, height:17}` hovered, so splitting reveal and click clicks empty space. A click that lands while a dialog is still running its open animation is not registered at all. And paragraphs arrive over the WebSocket, so `waitForLoadState('networkidle')` can resolve before any paragraph has rendered.

Strict-mode collisions sit on top of that. `getByRole('link', { name: 'Job' })` matches the accessible name as a substring, so it also selects note links whose title contains "Job". The header and user-menu locators now pass `exact: true`, checked against the names in the templates.

Fallback selector chains move into the Page Object and narrow to the element that exists. The cancel button was `.cancel-para, [nz-tooltip*="Cancel"], [title*="Cancel"], button:has-text("Cancel"), i[nz-icon="pause-circle"], .anticon-pause-circle`; only `.cancel-para` is in the new UI. The export dropdown is declared without `nzTrigger`, so ng-zorro opens it on hover rather than click, which is now encapsulated as `openExportMenu()`. The clipboard spec produced a TEXT result because `%sh` output has no `%table` marker, so the export control never rendered at all; adding the marker gives it a real TABLE result.

### What type of PR is it?
Bug Fix

### Todos
* [x] Confirm the removed selectors appear in no Angular template and in no React component
* [x] Confirm each replacement exists and means what the test assumes
* [x] Check that every changed assertion still fails when the feature is broken
* [x] Run the cancel test, measure the status transitions, and set the timeout from the measurement
* [x] Lint with the rules added in ZEPPELIN-6560 and confirm no new violations

### What is the Jira issue?
ZEPPELIN-6562

```bash
cd zeppelin-web-angular
npm run e2e:fast -- tests/notebook/ tests/home/ tests/share/ tests/workspace/
```

The clipboard spec needs a shell interpreter and skips on CI. The cancel test runs on CI against the Python interpreter; measured locally, the status goes `PENDING` to `RUNNING` and reaches `ABORT` about ten seconds after the click, which is what the assertion timeout is set from.

The dead selectors can be confirmed directly: `grep -rn "fa-spin\|running-indicator\|paragraph-status-running" zeppelin-web-angular/src` matches only the vendored FontAwesome stylesheet and no template.

### Questions:

### DIFF
--- a/zeppelin-web-angular/e2e/models/header-page.ts
+++ b/zeppelin-web-angular/e2e/models/header-page.ts
@@ -44,20 +44,26 @@ export class HeaderPage extends BasePage {
     this.notebookMenuItem = page.locator('[nz-menu-item]').filter({ hasText: 'Notebook' });
     this.notebookDropdownTrigger = page.locator('.node-list-trigger');
     this.notebookDropdown = page.locator('zeppelin-node-list.ant-dropdown-menu');
-    this.jobMenuItem = page.getByRole('link', { name: 'Job' });
+    // A global getByRole('link', { name: 'Job' }) matches by substring.
+    // It also picks up note-list links (e.g. a note named "...Job..."), which is a strict-mode violation.
+    // Scoping to the menu item and matching exactly selects the single header entry.
+    this.jobMenuItem = page.locator('[nz-menu-item]').getByRole('link', { name: 'Job', exact: true });
     this.userDropdownTrigger = page.locator('.header .user .status');
     this.userBadge = page.locator('.header .user nz-badge');
     this.searchInput = page.locator('.header .search input[type="text"]');
     this.themeToggleButton = page.locator('zeppelin-theme-toggle button');
 
+    // Same substring and strict-mode risk as jobMenuItem above.
+    // Scope these links to the dropdown overlay and match exactly.
+    const userMenu = page.locator('.zeppelin-user-menu');
     this.userMenuItems = {
-      aboutZeppelin: page.getByText('About Zeppelin', { exact: true }),
-      interpreter: page.getByRole('link', { name: 'Interpreter' }),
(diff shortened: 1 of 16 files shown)


tbonelee (approved):

tbonelee:
Merged into master

PR #5352, opened 2026-07-26
[HOTFIX] Dedupe minimatch and bump postcss in zeppelin-react

voidmatcha:
### What is this PR for?

The `npm-audit` job fails on master with seven high findings, all in `devDependencies`:

- `brace-expansion <=5.0.7`, GHSA-mh99-v99m-4gvg (DoS via unbounded expansion)
- `postcss <=8.5.17`, GHSA-r28c-9q8g-f849 (path traversal in source map auto-loading)

`minimatch@10.2.5` is already hoisted at the top of `zeppelin-react`'s tree. The vulnerable `brace-expansion@1.1.16` comes from four *nested* copies of `minimatch@3.1.5`, under `eslint`, `@eslint/eslintrc`, `@eslint/config-array` and `eslint-plugin-react`. An `overrides` entry for `minimatch` collapses those onto the version already present, removing `minimatch@3` from the tree, and `brace-expansion` then moves 5.0.7 to 5.0.8.

`postcss` moves 8.5.15 to 8.5.23 in the lockfile (`npm audit fix`, no `package.json` change; `nanoid` follows as its dependency). That is the same bump as #5341, which this supersedes.

The result is `found 0 vulnerabilities`. The gate itself is unchanged: still `--audit-level=high` over the full dependency tree, so `devDependencies` keep their CI coverage.

Upgrading the eslint side is not an alternative here. `eslint-plugin-react@7.37.5` is the latest release, declares `minimatch@^3.1.2` itself, and does not run on eslint 10 (`TypeError: contextOrFilename.getFilename is not a function` at `lib/util/version.js:31`), so no version bump removes the nested copy. Overriding `brace-expansion` alone does not work either, because 5.0.8 drops the CJS default export that `minimatch@3.1.5` calls, which gives a green audit and `TypeError: expand is not a function` on lint. Overriding both avoids that, since `minimatch@10` does not use the removed export.

The lockfile shrinks by 129 lines, dropping the four nested `minimatch` / `brace-expansion` / `balanced-match` trees and `concat-map`.

### What type of PR is it?
Hot Fix

The `npm-audit` job on this PR. Locally, from a deleted `node_modules` in `zeppelin-web-angular/projects/zeppelin-react`:

```
$ npm ci --ignore-scripts && npm audit --audit-level=high
found 0 vulnerabilities      # exit 0 (master: 7 high, exit 1)
$ npm run lint               # exit 0
$ npm run build              # exit 0
$ npm test                   # 14 passed
```

### Questions:

### DIFF
--- a/zeppelin-web-angular/projects/zeppelin-react/package-lock.json
+++ b/zeppelin-web-angular/projects/zeppelin-react/package-lock.json
@@ -501,37 +501,6 @@
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
-    "node_modules/@eslint/config-array/node_modules/balanced-match": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.2.tgz",
-      "integrity": "sha512-3oSeUO0TMV67hN1AmbXsK4yaqU7tjiHlbxRDZOpH0KW9+CeX4bRAaX0Anxt0tx2MrpRpWwQaPwIlISEJhYU5Pw==",
-      "dev": true,
-      "license": "MIT"
-    },
-    "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-      "version": "1.1.16",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.16.tgz",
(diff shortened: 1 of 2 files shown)


tbonelee (approved):

tbonelee:
Merged into master

